PR #77, opened 2018-11-20
add meteor effect

viknet365:
add meteor effect from https://www.tweaking4all.com/hardware/arduino/arduino-all-ledstrip-effects-in-one/#Sources


Aircoookie:
I'm still quite new to contribution management and stuff. It's really weird, I just had full write access to your branch until I merged it, and now I don't have any in order to re-add the deleted file. Seems strange that the system would give me write access to your branch in the first place...

PR #90, opened 2018-12-17
Use time/time.h from local dependencies

cboltz:
This fixes a compile issue, which can be
a) file not found or
b) (after installing the Time library) redefinition of a variable

Note: I'm new to Arduino and WLED, so please double-check before merging ;-)


Aircoookie:
Thank you! I'll check it out later!

LindezaGrey:
I did the same fix, looks good

Aircoookie:
Works well for me, too! Merging.

PR #91, opened 2018-12-17
Fix include filename for WiFiUdp.h

cboltz:
I didn't find a WiFiUDP.h, but WiFiUdp.h (with lowercase "dp") exists ;-)

Note: I'm new to Arduino and WLED, so please double-check before merging ;-)

### DIFF
--- a/wled00/wled00.ino
+++ b/wled00/wled00.ino
@@ -51,7 +51,7 @@
 #endif
 
 #include <EEPROM.h>
-#include <WiFiUDP.h>
+#include <WiFiUdp.h>
 #include <DNSServer.h>
 #ifndef WLED_DISABLE_OTA
  #include <ArduinoOTA.h>


Aircoookie:
Thank you! The change is already in the dev branch (https://github.com/Aircoookie/WLED/pull/89)! Just out of interest, was there a software update of the Arduino IDE or ESP core that causes this? It is really interesting to see that you and two other users came across this in a matter of hours while no one else did for a year :)

cboltz:
I don't know - as I already said, I'm new to Arduino and WLED, so the only thing I can tell you is that it was already broken when I first tried it some days ago ;-)  (maybe I would have been the first if I'd done the pull request instantly ;-)

The more likely (even if it sounds unlikely) reason is that you finally got some Linux users ;-) - Linux has case-sensitive filenames.

Issue #131, opened 2019-03-14
"Sync-Lights"-state not saved


YeonV:
Hello :)

First of all:
Thank you sooo much for your AWESOME WORK!!!

After reboot the "Sync-Lights"-button defaults back to "on". Is it possible to save the "off"-state upon Power-loss and power-on? Or could i change the default value to off in the arduino code (preferably in the usermod-section)

A small code example would be soo nice :)

P.S.:

I included the wled in "Home Assistant" with all effect-, color- and brightness-change functionality.
Everything controlable via Google Home.
With a small component its now also possible to animate over multiple strips via udp within homeassistant itself.
If its of any use for you, let me know...

Aircoookie:
Hi!

Thank you, I'm very glad that you like it :)

You can set the boot default in the web settings -> Sync Settings -> WLED broadcast! I recommend just unchecking the "Send notification on direct change" checkbox, that way you can easily temporarily re-enable it with the "Sync Lights" button.

Would be awesome if you shared your config! (if it's no extra trouble for you). I bet there are a lot of HA users who would benefit from it!

Mariu86:
I want to start using home assistant and anything that can help me it's very good. Thanks

YeonV:
> You can set the boot default in the web settings -> Sync Settings -> WLED broadcast! I recommend just unchecking the "Send notification on direct change" checkbox, that way you can easily temporarily re-enable it with the "Sync Lights" button.

Perfect mate!!
That was what i was looking for!!

> Would be awesome if you shared your config! (if it's no extra trouble for you). I bet there are a lot of HA users who would benefit from it!

I dont know if its the right place to share some of my things, but here you go... 
(it got a bit longer than I tought, sorry for that)

WLED in HA:
![strips](https://user-images.githubusercontent.com/28861537/54463771-779bfa00-4774-11e9-9a2a-9056546b25e4.png)


Here is the HA-configuration using the official mqtt light:
```
light:
  - platform: mqtt
    name: "LED-Strip_1"
    state_topic: "blade/ledx/strip1/c"
    command_topic: "blade/ledx/strip1"
    rgb_state_topic: "blade/ledx/strip1/c"
    rgb_command_topic: "blade/ledx/strip1/col"
    brightness_command_topic: "blade/ledx/strip1"
    brightness_state_topic: "blade/ledx/strip1/g"
    brightness_value_template: "{{ value }}"
[...]

PR #133, opened 2019-03-18
HA Light Auto Discovery

debsahu:
- Send HA MQTT Discovery message ~2.4kB based on input by @YeonV from here: https://github.com/Aircoookie/WLED/issues/131

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -75,7 +75,7 @@ build_flags =
   -D WLED_DISABLE_INFRARED
 
 [common:esp8266_512k]
-platform = ${common:esp8266.arduino_core_2_4_0}
+platform = espressif8266@1.7.0
 build_flags =
   -D PIO_FRAMEWORK_ARDUINO_LWIP2_HIGHER_BANDWIDTH_LOW_FLASH
   ; -D WLED_DISABLE_MOBILE_UI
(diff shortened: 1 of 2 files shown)


debsahu:
Fixed PIO/TravisCI as well https://travis-ci.com/debsahu/WLED/builds/104834062

Aircoookie:
Thank you very much for your work! So if I understood this correctly, HA is now able to detect WLED devices without further user setup (creating config as in #131 manually)?
For now, I disabled it by default, because the string constants use some 2kB of global memory, and I'm really running low at the moment with high LED counts. But I agree, I will re-implement the getModeName() with the flash strings from the original WS2812FX and then I should be able to enable this by default! Also thanks for PIO and travis fixes!

debsahu:
yes that is the idea, HA should be able to detect WLED automatically. You will reclaim ~2.4kB memory back after the function call.

debsahu:
Just pushed getmodeName() and getModeCount() additions.

YeonV:
OMG, thanks so much! Feeling so honored that you make my noobish things useful <3
Also thanks for the mention :)
Do we need to add YYYY (discovery topic) and XXXX (device name) to the web interface so one could set that from the frontend?

You did a great job mate!

debsahu:
@YeonV Both YYYY and XXXX are already on web-interface. YYYY can be set on Sync Settings and XXXX is the friendly name you provide in the settings as well.

PR #136, opened 2019-03-20
Flash Strings for Modes, Downgrade PIO to 2.4.2 core and more

debsahu:
- Using a for loop for generating HA autodiscovery topic. Uses getModeCount() and getModeName()
- Tested works on latest HA 0.89.2
- Fix crash on 2.4.2
- Use String instead of Flash Characters for JSON devlopment
- Move Arduino Core to 2.4.2
- added WLED_ENABLE_HOMEASSISTANT_AUTODISCOVERY to compilation


Aircoookie:
Sounds very good, thanks! Taking a look now.

Aircoookie:
Finally had time to check this out. Unfortunately your changes don't seem to have fixed it for me. flashmodenames branch still crashing on Wemos D1 (4M 1M spiffs) using 2.4.2, no matter if using LwiP 1 or 2, using DHCP or static, or using ArduinoIDE or PIO. 2.5.0 no issues as before. I'm sure that I have merged your changes correctly. Still assuming this is a heap fragmentation issue due to the many Strings involved. Not seeing any way we could workaround that though. Since I don't have a HA setup, I use `broker.hivemq.com` as my test MQTT broker, that might be a factor.

On a related note, since the travis fix, PIO will always compile (and if selected, try to upload) for each and every supported board one after the other [nodemcuv2, d1_mini, esp01_1m, esp01, esp32dev]. Do you have an idea what's causing this? I'm assuming it's just an user error on my side and I need to specify my target board somewhere within the IDE.

debsahu:
Ok let me check with hivemq and maybe I can provide a binary file to check.

I removed the default environment from `platformio.ini`, which means it will compile for all environments. That's a good thing as the ESP32 settings were outdated. All one has to do is uncomment the top default environment matching one's hardware environment.

debsahu:
see if the changes make sense.

PR #139, opened 2019-03-23
fix flash strings

debsahu:
- fix for Flash Strings core v2.4.2
- fix Animation names as well


Aircoookie:
Thanks for the quick fix!
I decided however that the flash strings and separate implementations for each core are not optimal. I've rewritten your code a bit, it now only uses 2k flash and 200b global memory! With this usage I'm fine with enabling the feature by default for all installations, so the ENABLE flag is no longer neccessary. Also causes (next to) no heap fragmentation. Tested working with 2.4.2, 2.5.0 and esp32. It would however be cool if you did a small test yourself to see if it still works fine with HA! Changes are in the latest master. This doesn't use the flash strings, but instead creates the fx_list in the char buffer from the JSON_mode_names (just prepending the [FX=YY]). Also, for esp01 I disabled SPIFFS for now in the platformio.ini, so now it can fit in flash with more features enabled on core 2.4.2. Let me know what you think!

debsahu:
kudos, works flawlessly!

debsahu:
For some reason the web-responses are very slow.

Aircoookie:
Was able to reproduce. Circumstances are very mysterious however.

- Core v2.4.2, MQTT state doesn't (seem to) matter
- Only occurs when compiling with PIO, not with ArduinoIDE (regardless of LwiP)
- Only occurs on PIO with LwiP v1, not with v2 (higher bandwidth respectively)
- Only occurs on ESPs that have the slowness issue with core 2.5.0 as well (tested on 6 units to make sure)

Any ideas? Workaround is easy, we just default to LwiP 2 in PIO. Still, I'll investigate on this some more. What bugs me the most is that it doesn't happen in arduino IDE with LwiP v1. Firmware binary compiled with PIO (same settings+flags!) is also about 2kb larger for some reason.

debsahu:
I did a few changes, converted all `const ... PROGMEM = ....` to `static const .... PROGMEM = .... ` in **html_classic.h**, **html_mobile.h**, **html_other.h** and **html_settings.h**. I also erased and used the correct partition table in Arduino IDE (4M/1M) and reuploaded sketch. Seems to have fixed the slowdown. I agree that we need to migrate to LWIP2 from 1.4.

PR #152, opened 2019-04-15
Fixed MQTT color response

YeonV:
Leading zeros are not trimmed on /c topic anymore :)
Before blue: #FF
After blue: #0000FF


Aircoookie:
Thank you so much! I can't believe I missed this when I added MQTT and it went unnoticed for so long!

PR #160, opened 2019-05-01
Added support for APA102 LEDs.

T-Arens:
Uncomment "#define USE_APA102" in NbpWrapper.h.
Connect clock to GPIO 0 and data to GPIO 2.
Choose color order "GRB" in the settings, do not check RGBW.
Tested on ESP-01 with 301 LEDs.


Aircoookie:
Thank you!

PR #166, opened 2019-05-09
Set a LED color by API command XE={position}.

robi772:
Set a LED color by API command XE={position}.
ex: /win&XE=149&XC=h0000FF


Aircoookie:
Thank you for your pull request!
I'm sorry for not getting back to you.
Unfortunately I need to close this because I just removed support for pixel locking to save memory. The new way to do this is to use segments or `handleOverlayDraw()`
Happy new year!

PR #237, opened 2019-09-29
Add online status via MQTT LWT

mw-white:
This PR adds an online/offline indication for MQTT using the last will & testament functionality.  The availability status is published at **mqttDeviceTopic/o**.

When the device connects to the broker, it sets a last will to publish "offline" to the availability status topic.  It then publishes "online" to the availability status topic.  

I've also added the availability topic into the homeassistant mqtt discovery packet.

I find this very handy since it gives me almost immediate feedback in HA if the device goes offline.

### DIFF
--- a/wled00/wled17_mqtt.ino
+++ b/wled00/wled17_mqtt.ino
@@ -45,6 +45,11 @@ void onMqttConnect(bool sessionPresent)
     mqtt->subscribe(subuf, 0);
   }
 
+  // publish online status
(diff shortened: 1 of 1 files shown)


mw-white:
Failed on ESP-01.  I'll try to get it to fit...

mw-white:
I just found issue 207 about the mqtt rewrite...looks like it will include availability.  Closing this one...

PR #344, opened 2019-11-12
Add support for Saskatchewan Timezone

ericyanush:
This PR adds support for the Saskatchewan Timezone (America/Regina) UTC-6 (CST no DST).


Aircoookie:
Thank you! It took so long to merge this because new timezones can't be added in the middle of the list (although that would make sense to have the list in a logical order).
For now I added Saskatchewan to the end of the list and will add a method soon that sorts the list alphabetically.

PR #364, opened 2019-11-21
fix #361

### LINKED ISSUE #361: Docs for compiling WLED for esp32 in PIO
Hi @Aircoookie !

First of all: What a library! Judging from the sheer amount of useful features it's just everything that I ever dreamed of! 

As I was not able to get the binaries running on my esp32 here (flashed it with the esptool [is that even correct?], but it never created an softAP), I was attempting to compile your lib in PlatformIO, which was not successful either. Although I installed all libraries according to your quickstart, uncommented env_default = esp32dev in the platformio.ini and hit on Build, several errors were thrown at me, the first being:

> 
> Assembler messages:
> Fatal error: can't create .pio/build/esp32dev/lib744/NeoPixelBus_ID547/internal/NeoEsp8266UartMethod.cpp.o: No such file or directory
> *** [.pio/build/esp32dev/lib744/NeoPixelBus_ID547/internal/NeoEsp8266UartMethod.cpp.o] Error 1

Is there any documentation on howto build the library with OpenplatformIO or is it heavily experimental?

Thanks very much,
Tom

debsahu:
- update ESP32 core to fix https://github.com/Aircoookie/WLED/issues/361

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -91,7 +91,7 @@ build_flags =
   -D WLED_DISABLE_INFRARED
 
 [common:esp32]
-platform = espressif32@1.7.0
+platform = espressif32@1.11.1
 build_flags =
   -D PIO_FRAMEWORK_ARDUINO_LWIP2_HIGHER_BANDWIDTH
   -D ARDUINO_ARCH_ESP32


Aircoookie:
Thanks!

PR #368, opened 2019-11-22
Fix Visual Studio Project Structure and add 3 new Effects 

stockklauser:
- Fix the Visual Studio Project Structure with the e131 lib change to e131 async lib
- Add 3 New Effects: Police, Police All, Multi Dynamic


Aircoookie:
Thank you very much :)
Will merge tomorrow, but I'll add a little change to the police effect code to make it more memory efficient. Multi Dynamic in fact is already part of WLED, it is regular Dynamic with maximum intensity selected!

Issue #194, opened 2019-07-28
Adalight not working


ViFi-connected:
Using **Arduino Nano** i can stream from PC to **ws2812b** via **Prismatik** with no problems at all. But when using WLED on **NodeMCU**, the stream is delayed and just not working properly. Is it an **esp8266** problem? Should i connect the **Nano** before **NodeMCU** to grab and pass the data? Thanks for any advice.

Aircoookie:
Hi! Sorry that there are issues with Adalight!
Can you send me the code that you use with the Arduino Nano?
Maybe the serial baudrate of something is off. I will check if it is a problem with WLED!

No, WLED can't process the raw pixel data coming from Arduino nano. You could use an Arduino relay module and connect it to GPIO12 of NodeMCU, then connect the led data lines from both the Nano and NodeMCU to it. This can automatically switch between the nano ambilight and WLED.

Ideally thought, WLED should work nicely with Adalight. How many WS2812b leds are you using?

ViFi-connected:
I am using 116 leds. Baudrate is set to 115200 in Prismatik (works with Nano, same USB port). 

[Adalight_ArduinoNano.txt](https://github.com/Aircoookie/WLED/files/3453014/Adalight_ArduinoNano.txt)

codamo35:
Hey Aircoookie, great stuff cooking in here. Thoroughly enjoying it!! I had a similar issue a while back with a Prismatik/Adalight setup, but didn't extensively troubleshoot it. Just spent a bit of time today looking back into it and here's what I'm running into:

-I have 400 addressable points in my ceiling, but tested this down to just 20 LED points (changing length had no effect).
-I have the WLED 0.8.4 binary loaded onto a NodeMCU, which works fine and dandy with controlling all 400 points. I have the Prismatik 5.11.2.19 binary application loaded to my Mac OS 10.14.5, and the application functions fully with no issues. 
-Baudrate set to 115200 in the Prismatik application and I have my WLED device hooked to my computer via USB (with correct COM port set in Prismatik). 
-With Prismatik running WLED sees incoming commands and stops producing it's own colors, and then produces a frozen color projection across all LEDs. If I display a moving color-wheel video on my computer screen under the marked pixel-points, the LEDs change color every 7-45 seconds, although sometimes not at all. It's not predictable and not repetitive. Each instance when it changes it is a stationary color change, no movements, just frozen.
[...]

PR #391, opened 2019-11-26
rewrite adalight parser

badbadc0ffee:
Fix for #194

### DIFF
--- a/wled00/wled04_file.ino
+++ b/wled00/wled04_file.ino
@@ -16,12 +16,13 @@ enum class AdaState {
 void handleSerial()
 {
   static auto state = AdaState::Header_A;
-  static unsigned int count = 0;
-  static unsigned int pixel = 0;
+  static uint16_t count = 0;
+  static uint16_t pixel = 0;
   static byte check = 0x00;
   static byte red   = 0x00;
   static byte green = 0x00;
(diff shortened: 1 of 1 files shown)


Aircoookie:
Thank you for the fix! Really appreciated :)

@raptusis are you building from source and can test this? If not, no worries, but I'll be able to test it in two days time at the earlierst.

badbadc0ffee:
BTW: the change was needed to connect our office x-mas tree to our Jenkins build farm. The LEDs flicker softly and change their color based on the result of our latest build...
I have also verified the function with AmbiBox on Windows and Hyperion on Linux.
tree
Merry christmas.

raptusis:
> Thank you for the fix! Really appreciated :)
> 
> @raptusis are you building from source and can test this? If not, no worries, but I'll be able to test it in two days time at the earlierst.

OK thanks! I will try as soon as possible! :)

Aircoookie:
Verified working for me with 54 LEDs! And way better than before, no flickers or lags. Thanks again :)

Mariu86:
How do you use the adalight Aircoookie?

raptusis:
Ok , all work very well ! 
Thanks badbadc0ffee ! 
Thanks Aircoookie !

PR #400, opened 2019-11-29
Adding new setting to disable fixed number of LED

ironosity:
Under LED Preferences a user can specify 0-254 LEDs to disable.  Disabling 1, for example will make every other LED off/black in all effects.  

Note: this doesn't apply to cronxie and possibly other things that you can do with api such as segments.

Also added a new effect based on static pattern that just uses third color option (or white if not set).  Intensity slider can set pattern to duplicate LEDs from 1-8 of each color before going to next color.


Aircoookie:
Hi and thank you for the huge PR! I will merge this as I believe it is a feature that deserves being in 0.9.0! Later I will rewrite it so that you can set it for each segment individually :)

PR #438, opened 2019-12-09
Grouping and Spacing

spitsw:
Basic grouping and spacing.

The segment.stop is now calculated and adjusted based on segment.rawLength, segment.group and segment.spacing

Save and loading from EEPROM for segment[0].

I haven't updated the segments UI.

This is what I use for grouping LEDs for my Christmas lights.

I would like to see something that is feature compatible with this


Aircoookie:
Thank you, amazing work! Will do some minor adjustments and merge soon :)

spitsw:
Thanks.

It probably needs some testing. I'm not sure what impact it has on Cronixie.

But it seems to work for me.

Warren

fhscholl:
I think I'm reading this correctly, but I'm new to WLED so I want to be sure. Does this basically abstract pixels into groups and apply effects at the group level? Setting group size to 1 and spacing to 0 would result in the behavior today, but you could take a string of 192 lights with a group size of 12 and it would treat the whole string like it had 16 pixels?

If so, this is awesome and exactly what I need for a use case I started to design last week for my landscape lights.

spitsw:
Yes, that is correct. This enables grouping pixels into virtual pixels.

Aircoookie:
Fixed the merge conflicts. Will now make a few little changes and then merge, thanks :)

spitsw:
Note: I have observed an issue where the grouping/settings revert back to 1. I haven't had the time to investigate or fix.

The steps to reproduce are:
- Set up a macro to turn on / off LED's.
- Set up a daily timer to trigger on/off macros.

The LED's will turn on with a grouping of 1 on the 2nd day.

I don't think it's related to the macro's / timing, rather the segments aren't being initialised correctly in some circumstances. There is probably a quicker way to reproduce the problem.

PR #443, opened 2019-12-10
add bouncing balls effect

fishbone-git:
used to svn, new to git.  Worked this out on my home build so thought I'd share since I see you have it on your todo list.  hope the diff is done correctly.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -961,10 +961,6 @@ uint16_t WS2812FX::mode_halloween(void) {
  * Random colored pixels running.
  */
 uint16_t WS2812FX::mode_running_random(void) {
-  uint32_t cycleTime = 25 + (3 * (uint32_t)(255 - SEGMENT.speed));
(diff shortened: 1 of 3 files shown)


Aircoookie:
Thank you for porting these awesome effects! I will check it out soon :)

fishbone-git:
The sinelon effect had prev_pos used to bridge the gap as the faster sinelon may skip some LEDs.  Was it intentional to go without that? I had hoped to use setRange to cover the gap, but it didn't work as I had expected so an awkward for loop was used

Aircoookie:
Addressing skipped LEDs is a good point i didn't consider!
I just assumed it was a redundant artifact from porting, sorry!
prev_pos is also not a perfect solution though because I had some occasions where the sinelon wouldn't travel all the way to the edge when the direction is inverted (say pos and prev_pos are both 1 since 0 was skipped). I'll think about what we could do about it :)

PR #445, opened 2019-12-11
Closes Aircoookie/WLED#444

spitsw:
This corruption was created when the files were converted from UTF-16 to UTF-8


Aircoookie:
The Chinese had taken over! Thank you :)

PR #481, opened 2019-12-18
Support for analog RGBW led strips

Def3nder:
Hi Aircoookie,

I tried to separate the changes to reflect only the analog stips.

After implementing the PIN driver for 4 channel LEDs, I realized that Alexa did produce completly wrong colors - so I debugged what Alexa does send when selecting one of the five white color tones: Alexa sends 5 single CT values.
So I added a switch-case structure for these values where I  1) assign 100% on the white channel  and  2) add some yellow or some white from the RGB leds (this is in EspalexaDevice.cpp and in WLED12_alexa.ino). Finally there is an error in the RGBtoRGBW function: the maximum saturation was 255 and it should be 100 to get correct RGBW values - this is the change in WLED14_colors.ino.

At the end I added three common boards in the platformio.ini: 
1.   the MagicHome Wifi Controller
2.   the H801 Wifi Controller (for use with 4 channel stripes and with 5 channel stripes)

Looking forward to hearing from you soon, 
Def3nder


Aircoookie:
Hi Def3nder,

awesome, thank you for your hard work!
Will merge in a few minutes :)

Aircoookie:
Ok, I had to create a new pull request #487 as I somehow can't push updates to this one (aside from fixing the merge conflict). Don't worry, you are still the owner of the commits you made and will be listed as collaborator!

Aircoookie:
And done :) Thanks once again for your contribution!

Def3nder:
pls. have a look at the comments I made in #481 -  from my point of view we need the changes in EspalexDevcice.cpp, too.

Aircoookie:
That's right, but I did in fact merge them... Here: https://github.com/Aircoookie/WLED/blob/master/wled00/src/dependencies/espalexa/EspalexaDevice.cpp#L125

Is it not working as intended anymore? I've moved the code for the analog LEDs iin `NpbWrapper.h` from `Show()` to `SetPixelColor()`. The reason is that the color `GetPixelColor()` gives is often not accurate (because of the brightness calculation having to be reversed, which doesn't work perfectly. Could be that this broke this part of the code: https://github.com/Aircoookie/WLED/blob/master/wled00/NpbWrapper.h#L200
because you need slightly different hardcoded values since the "brightness error" is removed.
Sorry in case I broke it :(

Def3nder:
...I did not try it yet - let me do this on both RGBW and RGB CW/WW stripes.

I let you know the result :-)

Issue #269, opened 2019-10-18
Support for solid RGBW LED strips


Def3nder:
Hi Aircooookie,

thx fir this awesome project - it's amazing! 

I have quite a lot of non-digital LED strips attached to MagicHome Adapters re-flashed with Tasmota. 
Unfortunately beside a simple "color-change" there is no single effect available.

I think some of the effects in WLED would look good on single colour strips as well (like rainbow or breath etc.).

Do you think it would be possible to use WLED with an analogue 3 pin RGB strip (+ the common cathode) ?

The NeoPixelBus library does hide so much so that I could not find the point in the code where the colour(s) is passed to the PIN.

I would love to use WLED for the analogue RGB strips.

Greetings from Berlin,
Def3nder

Philje123:
Same for me too! I have 3 Magic home controllers with fixed RGB strips (non-addressable) running some custom MQTT code. But would love them all to be Wled. I really like it.

theshadowsvk:
Guys you can't have effects like rainbox on non-addressable strips. Imagine that on electricity level. You can only control each color on whole strip, not single diode without led-IC that controls it.
Just buy addressable strips and you are done, they are pretty cheap.
WLED project is only for addressable led strips, so the neopixelbus library, so the ws2812fx library.

Def3nder:
Sure, you can. Just define a WLED strip with a single pixel - all you can do with that can get done with an non-addressable strip, too.

Def3nder:
I think the way implementing this could bei tweaking the NeoPixelBus Library to offer another hardwsre type and add the logic to drive the three pins there...
What do you think?

theshadowsvk:
That's not the way it works. It can behave like single pixel cause it is analog. It lacks the IC which controls the chip, wled project,neopixelbus library etc sends data to this IC, it is data stream. And one IC cant control the whole led strip if you add one external like ws2811.

creativeretail:
This will do what you want:-

http://www.freetronics.com.au/freenfet3

Schematic / PCB files here:-

https://github.com/freetronics/FREENFET3

brentbrooks70:
@Def3nder this is a Duplicate of #58 Please see that issue for info on to make this work for you

Def3nder:
> @Def3nder this is a Duplicate of #58 Please see that issue for info on to make this work for you

@brentbrooks70 Thx - this looks exactly what I was looking for - I will try it and post my results here for reference.

[...]

PR #487, opened 2019-12-19
Add Analog/Solid RGB(W) strip support

Aircoookie:
Thank you @Def3nder!
(details in original PR #481)


Def3nder:
Hi Aircoookie,

the case structure for the 5CH leds in NpbWrapper.h uses the values for R,G,B & W that are set in EspalexaDevice.cpp.  This change there was necessary, because when slecting a color temperature value in the Alexa App, the white channel should be 100% all the time - only the tone should be changed using either the RGB colors to add orange or white-blue (or in case we have 5 channels this 5 distinct values are 'translated' into CW + WW values in the #ifdef WLED_USE_5CH switch/case structure.)

Does this make sense ?
This was the only way to get real white colors (without RGB) with the 4 channel strip with Alexa.

GeorgeIoak:
Thanks @Def3nder for your work and @Aircoookie for merging! In #269 I mentioned the Novastella RGBWW flood light. Unfortunately it has a slightly different pin definition than the H801 controller but I believe it is the same other than the pin assignment(s).

What is the best approach to try and get this merged in? I provided some details in the other thread.

HansWurst90:
Hi @Def3nder,

I found this PR when I was looking for a solution to control Analog LED Controllers like the MagicHome Controller via e1.31. It seems like this PR modifies WLED to work with analog LEDs.
How would you go about to flash a MagicHome controller or a NodeMCU to work with analog LEDs  instead of addressable LEDs? Is there a certain setting before or after flashing?
I'm sorry if this is not the right place to ask but I could not find a better one.

PR #507, opened 2019-12-23
More framerate conversions, etc

fishbone-git:
Some conversions for framerate along with other enhancements:
tri-chase using intensity for a width setting
oscillate with 3 colours and intensity for width
merge larson scanners (dual is boolean)


Aircoookie:
Thank you! :)
I'll quickly check out the modified effects, then merge!

Merry christmas!

Aircoookie:
Awesome job on this, everything looks great! Merging! (also don't mind travis, seems like Platformio is having server issues)

fishbone-git:
platformio was blocking local building for a while too.  Workaround is to go offline, which obviously won't work for travis . I have some more code for setting width of scan (I saw your note).  Ran into an issue I can't seem to track locally, so will file that separately (tertiary colour isn't saved in presets).  Happy Holidays

PR #510, opened 2019-12-23
Sizing for scan and dual_scan

fishbone-git:
add some sizing to the scan and dual_scan modes


Aircoookie:
Awesome, thanks! Making Scan support more than one pixel was on my list for a long time!
There was just one thing i didn't quite like, which was that the lit sections would "overshoot" (get smaller at the edges and overflow into adjacent segments). Fixed that! The second section of dual scan also supports the tertiary color now.

Also noticed that in your earlier PR the `SEGMENT.start` offsets were missing from the Scanner effects, so I added them back :) Otherwise the effect would always run from LED 0 on!

fishbone-git:
I noted the change.  I followed the oscillator style but this makes more sense for the scan mode. Thanks  for the correct.  I don't have multiple segments, so totally missed that.  Noted for future.

PR #524, opened 2019-12-27
1d fireworks

fishbone-git:
from: http://www.anirama.com/1000leds/1d-fireworks/
ported to wled with some ability to adjust for different segment sizes.


fishbone-git:
Nice.  SEGENV.data is very handy.

PR #527, opened 2019-12-28
add multi fireworks starburst effect

fishbone-git:
based on a video and c# code from here:
https://www.reddit.com/r/arduino/comments/c3sd46/i_made_this_fireworks_effect_for_my_led_strips/


Aircoookie:
Thank you so much for porting that over! Haven't taken a look yet in person, but from the video it looks absolutely amazing! Hopefully I can get it merged before New Year's!

There is one small problem (which is also the reason I haven't merged your Fireworks1D and BouncingBalls PRs yet), which is static variables. They wouldn't be an issue on a system with a small number of effects or large amount of memory, but since WLED has less than 20 kilobytes of free RAM on ESP8266, it's hard to justify using more than a thousand bytes just for the two fireworks effects.
Another issue is that running the same effect on multiple segments will likely lead to weirdness, as they share the static variables. 

Since they look so mesmerizing though we really need to add them! I'll try finding a way to allocate some memory that is released once the effect is no longer running, so that it can be re-used for other features.

Currently we do have `_locked[i]`, a byte array the size of the LED count. A few effects already use it and i was initially planning to use it for this too. The problem with that is that each segment should only access part of the array, from seg.start to seg.stop. With small segments and many floats (4 bytes each) needed, that won't be enough memory for fireworks unless the segments are huge.
I'll let you know once I come up with a solution!

fishbone-git:
I knew the statics and floats would be an issue.  All these effects come standalone sketches.  I've stripped it down where I could and it works well enough on my setup (not as good as that video, sadly, but a reasonable approximation.  It's tricky with these types of effects with overlapping animations and state changes. I had pondered some sort of added parameter space in the segment class that could be wiped on mode changes, but that's beyond my tinker-level skills. Let me know if I can help

Aircoookie:
Wow I just merged without fully implementing the memory logic stuff (facepalm). Will not revert, but add that tomorrow!

PR #550, opened 2020-01-04
ESP-07 module support

srg74:
ESP-07 module support plus 2 library dependency for Dallas sensor

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -10,6 +10,7 @@ lib_dir = ./wled00/src
 ; env_default = nodemcuv2
 ; env_default = esp01
 ; env_default = esp01_1m
+; env_default = esp07
 ; env_default = d1_mini
 ; env_default = esp32dev
 ; env_default = esp8285_4CH_MagicHome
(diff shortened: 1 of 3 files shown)


platformio.ini
```diff
@@ -43,7 +44,9 @@ lib_deps_external =
   #Timezone@1.2.1
   #For use SSD1306 0.91" OLED display uncomment following
   #U8g2@~2.27.2
-
+  #For Dallas sensor uncomment following 2 lines
+  #DallasTemperature@~3.8.0
+  #OneWire@~2.3.5
```
srg74: Dallas sensor libraries added for temperature sensor as Quinled pub feature

platformio.ini
```diff
@@ -156,6 +159,18 @@ build_flags =
...
+platform = ${common:esp8266.platform}
+monitor_speed = ${common.monitor_speed}
+upload_speed = ${common.upload_speed}
+framework = ${common.framework}
+build_flags =
+  ${common.build_flags}
+  ${common:esp8266.build_flags}
+lib_deps =
+  ${common.lib_deps_external}
+
```
srg74: ESP-07 and ESP-07S module support for external antenna feature

platformio.ini
```diff
@@ -10,6 +10,7 @@ lib_dir = ./wled00/src
 ; env_default = nodemcuv2
 ; env_default = esp01
 ; env_default = esp01_1m
+; env_default = esp07
```
srg74: ESP-07 and ESP-07S module support for external antenna feature

Aircoookie:
Thank you! I do believe that turning off the display is a good idea to save power and prolong lifetime.
However, I don't believe permanently turning off the display 5 minutes after boot is a bit overkill. Displayed values could be needed after that point. I've changed that so that it turns off 3 minutes after any displayed information changed, and back on once it changes again.
I hope you agree :)

PR #564, opened 2020-01-07
Local include of IRremoteESP8266 library to save 11k program space

Def3nder:
Hi @Aircoookie , if the IRremoteESP8266.h is changed slightly (the #defines) a lot of not needed program code can be excluded. To do this it needs to be included localy.

I did this and excluded all code for sending IR and all code for the various Air Conditioner remotes:
[IRremoteESP8266.h](https://github.com/Aircoookie/WLED/files/4029931/IRremoteESP8266.h.txt)

This saves 11k of program space.

I used the latest library in version 2.7.2.

PLATFORMIO.INI:
The statement "env_default" is depreciated and will not be supported starting with the next release of PlatformIO. The "to-be-used" statement is "default_envs" and can have multiple comma separated enties.


Aircoookie:
Hi and thank you :)
Unfortunately doesn't seem to compile quite yet. I will look into it :)
Also I don't get why we are including all the air-conditioner files still. Since you removed the includes referencing them, we should be able to delete them, right?

Def3nder:
The files are all included and compiled.
In every .c file there are some #ifdefs that define if several codelines are included or not.
So the functions are there anyway - only the code itself isn't included in the library then.

Aircoookie:
Ok, I manually removed everything that seemed unneccessary. Program size decreased by another 13k I believe. Compiles fine! Could you test the latest commit to see if it still works as intended with actual IR hardware?

Def3nder:
Yes, sure - I will test this in the evening when I'm back home 😄

Def3nder:
Hi @Aircoookie , ripping out all the ir_modules does not work.
Now the result is always a code like 0x3195A31F instead of 0xF7C03F and it does change every single time. So, the ir_something.h and .cpp are needed.

Def3nder:
@Aircoookie Do you work on this your own ?  // can I close this PR ?

...I did not find a solution to change the #defines after including IRremoteESP8266.h (I tried to #undef all the unneeded moduled and #define them all with "false", but this did not have any effect on the compile.)

So, from my point of view including the library localy is the only way reducing the program space.

Aircoookie:
@Def3nder yep, let's close this for now. Found another potential pitfall, which is that the LGPL license of the library. Technically I am not allowed to include it into WLED.

I've found out that if you change all the air conditioner defines to false (not undeffing or deleting anything), about 25k of program memory is saved, almost the entire size of the library (28k). I will open an issue/pull request with them to change the air conditioner defines from wled00 or platformio.ini, so we wouldn't need to include the libary locally for that.

Mariu86:
I have a question if you change this library it will de-code other remote controls ? I use the 4 colored buttons used for teletext from my tv remote to control Wled with macros after you change the library I can still use the tv remote?

Def3nder:
@Mariu86 no, this library has different "protocols" implemented so it "undestands" combination of different data steams.

If you can control WLED with your TV reomte right now it uses one of the supported protocols.

@Aircoookie tried to remove all the protocols (and the code for sending IR) to save space but unfortunately this is too much.... but anyway: the license prevented to localy include + modify the files, so for the upcoming releases everything will stay the same.

Def3nder:
@Aircoookie : one last thing to this: the get ESP32 "enabled" the library version must be higher than v2.6.2.

I would suggest to stick with the actual version and change the platformio.ini **lib_deps_external** to
`IRremoteESP8266@2.7.2`

Mariu86:
Thanks for clarifying this ! Keep the good work guys and thanks for this awesome project!

PR #565, opened 2020-01-07
More IR remotes (24/40/44-keys) and select them in the settings

Def3nder:
Hi @Aircoookie , 

I added the logic for three more IR remotes that can be bought for small money at AliExpress (less than 2 EUR) and changed the settings so that the actual remote type can be selected instead of just turning all ON/OFF.

If you like I can add pictures of the remotes and a link to AliExpress to the wiki:

24-key: 
image

40-key: 
image

44-key:
image


Def3nder:
I forgot one line in WLED00.ino:

**byte whiteLast = 128;                         //white channel before turned off. Used for toggle function**

...right after:
**byte briIT = 0;
byte briLast = 128;**

I added this to the branch now but I do not know if this will be used for the PR...

Def3nder:
Hi @Aircoookie , I will do the same here: extract the changes without the IRremotreESP8266 library :-)

PR #566, opened 2020-01-07
ESP32 support for solid (analog) RGB strips

Def3nder:
Hi @Aircoookie , I changed NpbWrapper.h to use the ledc funtion with the ESP32 architecture.
This works very well and does flicker less because of the higher PWM frequency.

I saw examples where the PWM frequency was raised up to 12kHz with a resolution of 8 bit, but I thought 5kHz would be enough and would lead to less "stress" for the MCU.


Aircoookie:
Awesome!
Unfortunately this PR also contains the IR library integration, so I cannot merge it...
Would it be possible to create a new branch on your fork with just the NpbWrapper.h changed, then submitting a new PR?
Thanks and sorry for the inconvenience :)
(Same goes for the new IR codes PR #565)

I would just copy out the relevant parts and put them in master manually, but I want to make sure you get the commit attribution you deserve for your hard work!

Def3nder:
ok - I create a new one ... let's try 😄

PR #568, opened 2020-01-07
ESP32 support for solid (analog) RGB(W) stripes

Def3nder:
Uses ledc funtion to drive analog RGB(W) stripes with 3, 4 or 5 channels.

...this time only this single change :-)

Also important for the last PR:
the define in platformio.ini needs to be "WLED_USE_ANALOG_LEDS" - I changed this in the PR, too.


Aircoookie:
Awesome, thanks!

Issue #534, opened 2019-12-30
E1.31 does not support White channel


AndrewApi:
Hi
Just tested new code for "Fix analog write range" release "WLED 0.9.0-b2 (build 1912291)", works great now full brightness on LED Strip.

I am using RGBWW analogue strip, WLED controls all 4 channels when RGBW is selected, but in E1.31 mode I can only control the 3 RGB channels, I can not get the white channel (CH4) to light at all. Is this a possibility to correct please.

Thank you for all your hard work on this great project.

Def3nder:
Hi @AndrewApi ,

did you try to set the option „ Auto-calculate white channel from RGB: “ in the LED Settings ?
Does this change anything ?

AndrewApi:
Hi Def3nder

Thanks for the reply.
This takes away the White channel fader in GUI, it does not change anything when trying to control the White channel via E1.31.

Aircoookie:
Hi!
Unfortunately, any channels but RGB are not supported in E1.31 mode currently.
The reason for that is that most E1.31 sending software assumes 3 channels (RGB) per LED.
I might consider adding an "E1.31 layout" or similar option, that would allow you to control the white channel via E1.31.

pille:
i'd like to implement E1.31 RGBW modes analog to those introduced with #665.
does it make sense to wait until you figured out emulation in #573?

Def3nder:
Hi @pille, is there any software that can send RGBW via E1.31 ?

pille:
it should be possible to just add another channel for the white value.
i'm using QLC+ with my own fixture definition.

Aircoookie:
DDP now supports white channel, but there is basically no software that supports it (yet)

PR #573, opened 2020-01-08
RGBW support for E1.31 (should close issue #534)

Def3nder:
if "enable RGBW" is selected in the LED settings, the function colorRGBtoRGBW(col) is called after receiving new DMX pixels.

This should close issue #534


Def3nder:
...instead of "useRGBW" the if statement could also query "autoRGBtoRGBW" - I cannot say which one is better ❓

Aircoookie:
Hi again and thank you!
Unfortunately, this will not work because it computes the white value of the primary color, not the pixel data that was received. Now, we could easily change this so that `colorRGBtoRGBW` calculates the white for every received pixel, but I am worried about performance. The function uses floating point maths and would be executed up to 25000 times a second (500 LEDs at 50fps, about the maximum WLED E1.31 can do). I will run a test to seehow long the function takes to calculate. I would say it needs to be able to execute those 25000 times in less than 200ms in order to not lead to slowdowns, and i doubt that's achievable, but I will try it :)

Aircoookie:
Ok, executing `colorRGBtoRGBW` 25000 times takes 480 milliseconds on ESP8266 with 80Mhz cpu clock. That is unfortunately too long. I was able to optimize the function a bit to the point where it can execute 25000 times in 350ms. That's an improvement, but may still significantly impact the E1.31 performance. I'll have to think about it. Maybe we can also run the ESP at 160Mhz, which would cut the calculation time in half, but I'm not sure what side effects that may have.

Aircoookie:
Actually managed to optimize that function even more! It runs more than twice as fast now hehe. 230ms for 25000 passes. This is getting into territories where we could consider using it :)

Def3nder:
Hi @Aircoookie, may I ask what you did change for this dramatic speed improvement? 
...and can we use this in the existing RGB2RGBW conversion, too?

Aircoookie:
Well, unfortunately my changes caused the function to have an imprecise or outright wrong result.
Haven't had the time to look into the cause yet unfortunately yet, but here is my test sketch:
https://gist.github.com/Aircoookie/714f8feb8664dd1702b4fbfa6fede844

Def3nder:
Hi @Aircoookie, I did read some articles about RGB to RGBW conversion and the most interesting one was [this](https://blog.athrunen.dev/experimenting-with-efficiently-combining-rgb-and-true-white/).

...and an implementation of the function [here](https://github.com/ManiacalLabs/BiblioPixel/issues/31).

The conversion works quite different: it calculates the "grey part" of the colors, set's the white channel to this number and substracts it from all channels. This way "pure" colors do not lead to lighting the white channel (what has been so in the old version).

And the best thing: **it takes 6ms for 25.000 passes !** 

...this way we could a) add it to E.131 and b) replace the old version 😄 

What do yoou think ?

Aircoookie:
Wow that is amazing! Good job 😄 

I will take a closer look tomorrow and compare the resulting colors of the new function to the old one and plain RGB. This function is so fast, if it indeed creates pleasant colors, we could even add it to the `setPixelColor()` driver, so that every WLED interface (including Ambilight, E1.31 and even effects using desaturated FastLED palettes) can take advantage of the white channel!

Def3nder:
Hi @Aircoookie, the old function did have wrong results.

some exaples: 
- `rgb 128,128,128` is **50% white** - this should translate to `0,0,0,128`
- `rgb 0,255,0` is **100% green** - this should not get any white part at all (`0,255,0,0`)
- `rgb 128,128,255` is **50% white and 50% red** - this should translate to `0,0,127,128`

This is what the function does in theorie.

But you are right: we should test it in real life.
I will catch my RGBW strips (I got one SK6812 RGBWW last week) and compare the colors.
If everytjhing works as expected the result should be a "warmer" color, because the WW is about `3000k` and the sum of RGB normally is about `6500k`.

I will come back with the results ...

Def3nder:
Hi @Aircoookie,

so I've tested the function and it works as expected (and more accurate than the other one):
- old function, pure RED: 
image (there is white included)
- new function, pure RED: 
image (no white)

- OLD, pure green: 
image (there is white included)
- NEW, pure green:
image (no white)

- OLD, pure blue: 
image (there is white included)
- NEW, pure blue:
image (the 'strip-logic' does add some white to every blue to even the brightness)

- and as test the Police effect with "Pastel Color" as background to show the speed:
[20200129_202811.zip](https://github.com/Aircoookie/WLED/files/4130424/20200129_202811.zip)

...but I discovered **another issue**:
When enabling `Auto-calculate white channel from RGB` the color[ ] array will be changed directly and when the UI gets updated, only the RGB-part will be used to display the color.
**Example:**
- I select this color:
image
- after UI refresh it will be displayed like this:
image

- A solution could be to use the `rgb2rgbw function` directly within `setPixelColor()`. Then we would need to doublicate the color array to leave the input untouched.
- Or the color selector should stay in RGBW mode. Then the above selection would become this:
image

However: the function itselfs seems to be a good change 😄 - what do you think ?

Aircoookie:
Awesome, thanks for your testing!
I was quite busy today and couldn't check it out myself yet, but it looks very good!

Ah yes, makes sense that that would happen since the function modifies the input RGB data. If we'd put it into setPixelColor, that would be a non-issue though (or just making a second array, passing that to strip.setColor() instead).
Another thing I was thinking about is if we really need to or should subtract the white from the RGB channels. I'm sure it will yield cleaner and more uniform colors (white has almost the same brightness as pure colors), but keeping the RGB fully lit would increase peak light output. Maybe this is something I could make user configurable :)

Def3nder:
Hi @Aircoookie, 
yes, the peak output would be higher when the white channel isn't substracted,
However "normaly" the people buy an more expensive RGBW strip, because they want to get rid of the very cold 100% white they get from the mix of R,G and B (around `6500 kelvin`).
A RGBWW strip mostly has a white channel with around `3000 kelvin` and creates this "warm" white.

Using both RGB and white channel would result in a white tone of `4800 kelvin` (or even higher if the brightness of the three LEDs could be higher than the one of the white channel).

I think a **color temperature slider** would do the trick.
Then we could choose wether to use "pure RGB" (=cold) or substracting the white part in total (=warm) or do something in the middle (=natural) and mix this from 0 to 255 😄 .

Then the color model would need to be R,G,B,W,CT (like in some commercial solutions).
This way the user would have total control over the color and over the white (via R,G,B,W) and could use the desired tones in all effects (when enabling `Auto-calculate white channel from RGB`).

henrygab:
Taking the grey portion makes sense, but it seems like it would change brightness because it does not take into account perceived brightness.  E.g. Two LEDS at  127 are perceived brighter than a single one at 255.

Can they grey removal apply the perceived brightness table correction?  E.g. So `BR(old red - new red) == BR(white)'?

This does not affect "pure" colors, only those with leftover R, G, or B.  So, a test sketch  to see which is better might apply this to groups of pixels, or show every  pixel twice (once with,once without brightness correction)?

Def3nder:
Hi @henrygab, no, this is not the case. 

The white LED is like 5 times bigger than the R,G or B one. (have a look at the pictures above)
  


The emitted light is as bright as all three together.

henrygab:
> Hi @henrygab, no, this is not the case.
> The white LED is like 5 times bigger than the R,G or B one. (have a look at the pictures above)
> The emitted light is as bright as all three together.

@Def3nder ... Ok, I can accept that it's output at 255 will match the output of the three primary-color LEDs all being set to 255.  In fact, I can accept that if all three primary-color LEDs are set to the same value N, then setting the white LED to N will match brightness.  However, this was not my concern ... Let me try walking through with a specific example.  I'll number the thoughts, so it'll be easier if you find a flaw in my understanding (which I hope occurs!):

Consider the following original RGB value:
`RGB == { 254, 191, 127 }`
1. Per the above discussion, this would convert to:
    `RGBW == { 127, 64, 0, 127 }`
2. Thus, 127 was removed from each of RGB, and the W was set to 127; the W is thus output light at a level equivalent to `RGB == {127, 127, 127}`
3. The red LED is still individually lit at level 127
4. The green LED is still individually lit at level 64
5. Thus, the total red output ~= 2x a red LEDs at 127
6. Thus, the total green output ~= 1x green LED at 127 + 1x green LED at 64
7. Gamma correction for `RGBW { 127,  64,   0, 127 } == {  36,   5,  0, 36 }`
8. Gamma correction for `RGBW { 254, 191, 127,   0 } == { 252, 114, 36,  0 }`

Thus, except for the trivial cases where all three primary color values are identical, because the calculation is not taking into account the gamma values, would the perceived mix of the colors change by simply subtracting pre-gamma values?

Would a more conversion that ensures the gamma values of the converted values combine be more accurate?
Using the above example:
1. detect common quantity in RGB
    `W =  127`
2. lookup the gamma for that quantity in GammaT table
    `Wg = 36`
3. for each of R, G, and B, first temporarily convert using GammaT table
    `Rg = 252`
    `Gg = 114`
    `Bg = 36`
4. Next, subtract the white gamma value to find the target R/G/B gamma
    `R'g = 216`
    `G'g = 78`
    `B'g = 0`
5. If wanting to convert back to RGB values, then reverse-lookup using GammaT table
    `R' = 240`
    `G' = 167`
    `B' = 0`

I do not have RGBW leds to check this against.  Also, the gamma correction would likely differ for each primary color, depending on the temperature of the white LED.  But, that's just a few extra constant tables … is my question more clear with this example?

Thanks!

Def3nder:
@Aircoookie, correct me if I'm wrong, but I cannot see any gamma correction active in WLED.

There is a function `WS2812FX::gamma32` that reads `gammaCorrectCol` but this function is not called from any other. 
So from my point of view enabling / disabling color- or brightness gamma correction has not effect.
Did you switch this off at some point in time ?

If working on the color model, should I use the gamma32 function then ?
What is your suggestion ?

Aircoookie:
@Def3nder gamma correction is indeed a point we need to consider.
Gamma correction is a bit hidden, it works via the SEGCOLOR macro used by effects:
```cpp
#define SEGCOLOR(x)      gamma32(_segments[_segment_index].colors[x])
```
What I would propose is adding the RGBW conversion function after the gamma correction, which would be the case if we added it to setPixelColor(). I will try that out once I have time on Thursday :)

Aircoookie:
Hi @Def3nder !
I finally made some progress on this and added the auto white routine to `setPixelColor()`.
Right now there are probably too many modes choosable in LED settings, I should probably cut them down because it could be confusing, but this is just for testing. Here are my thoughts on each one:

None: This does not auto calc the white channel, instead we have the separate white slider. Straightforward.

Brighter: This is the new algorithm without subtracting the white from the RGB channels. I like it a lot, but ideally I would like something that outputs more to the white channel. It is very nice for pure white, but once you try to choose something a bit warmer the brightness quickly drops off. You can easily test this with the new color quick selectors by toggling between the Fairy white (slightly warmer) and pure white.

Accurate: At least for now, the name accurate doesn't seem to fit for this mode at all. This is your original proposal, with the white channel being subtracted from the RGB. It doesn't seem to work well at all unfortunately, the fairy white color doesn't look white at all, more like a weird greenish. If we want to make it work, we probably need to add it before the gamma correction after all, which would also mean that it wouldn't work with whites from palettes at all.

Dual: This is just for testing, Brighter mode with a White slider. You can either set a white manually or let the algorithm handle it if the slider is set to 0.

Legacy: This is the old algorithm. It produces a bit brighter non-pure whites than the Brighter mode, which I like. Together with Brighter, it is also closest in making the RGBW strip look the same color as other RGB only strips in the room, but as discussed earlier it has the issues of adding a little bit of white to saturated colors as well as being too slow for a per-pixel calculation.

Let me know what you think about those modes!

I find myself still preferring setting a saturated RGB color (like orange) manually and then adding full 255 white channel to it. This gives a very nice warm white with almost peak brightness. Of course the auto mode can never really replicate that as it wouldn't make sense to turn on full white channel on a saturated RGB color. Maybe this is the point I need to come up with a good color temperature algorithm that covers this use case!

Def3nder:
Hi @Aircoookie, I saw the code and I like options to choose from 😄 
I will have a closer look this evening.

What I have on my "to-do-list" is to work on a color-model, that will introduce a value for the color temperature and a kind of "features" byte, where we can store what the physical LED's are capable of.

For example: we could have RGB-only in the UI (e.g. with auto-RGBW enabled) and have an RGBW strip.

The strip has two possibilities of redering whites:
1. RGB channel at equal levels
2. the white channel

...each having different color temperatures.

Then in the settings we could enter the color temperatures of the white channel(s) - e.g 3000k for W1 and 6500k for W2 and we can assume a value for the CT of the RGB lights.

Then we can introduce a color temperature slider that gives a target CT and we calculate which channels (RGB, W and W2) we need to mix in what manner.

So basically we need an internal color-model that can hold the values for the biggest color-space and the calculate all others from there.

Can we close this PR ?

Aircoookie:
Yeah, I agree, we will need something that handles color temperature differently depending on the connected hardware, a feature byte sounds like a great idea!
The lights controllable with WLED are starting to have really different capabilities:
- Classic 3-channel RGB
- 4-channel RGBW (with 3 different color temperatures of white available, WW, NW and CW)
- 5-channel RGB + WW + CW
- 3-channel Amber + WW + CW WS2812 strips
- in the future maybe single channel analog white LED or WW+CW

Probably the first two are the most important focus for now, but the ideal situation would be to use all the capabilities of each hardware while trying to have different hardware stay close to each other in visual appearance and actual Kelvin color temperature.

For RGB, I believe getting an acceptable Kelvin spectrum could work like this:
Normal fade: Orange -> Warm White -> Full RGB white -> Blueish white

For RGBW, we could do something like this:
RGB-only orange -> Dim up white channel -> While white channel at full brightness, shift RGB channels from orange over white to blue -> Dim down white channel -> RGB-only blueish

Amber + WW + CW is probably the most straightforward:
For uniform brightness, just fade:
Amber only -> Amber 50% WW 50% -> WW only-> WW 50% CW 50% -> CW only
For peak brightness at around 3000k (WW):
Amber only -> Dim up WW -> Dim up CW (all on, peak light output) -> Dim down Amber -> Dim down WW -> CW only

I am not sure if there is a good one-size-fits all color model, RGB is good as a base model, but Kelvin temperature is probably the go-to for good white support.

Thank you for working on it, I'll give it some thoughts as well!

PR #582, opened 2020-01-10
Include new branch of IRremoteESP8266 to save 21k

Def3nder:
Hi @Aircoookie , I did open an issue and talked with the author of the IRremoteESP8266 library.

He created a new branch "compile_flags" where the usage of the different modules can by selected at compile time with -D build flags.

So: no code changes at all - only platformio.ini changes that result in 21024 bytes smaller code 😄


Aircoookie:
Awesome, thank you for sorting that out! :)

PR #585, opened 2020-01-12
FX intensity slider to desaturate Color in Colorloop FX

Def3nder:
Suggestion: 
with this little change the intensity slider has a function in the Colorloop Effect:

- With default (128) value and bigger the saturation is 100% (like now).
- With values below 128 of the intensity slider the saturation decreases

So the Colorloop Effect can be controlled to show saturated rainbow colors or pastel colors.

What do you think ?


Aircoookie:
This is very cool!

PR #587, opened 2020-01-13
Add WS2815 brightness limiter power model, using "magic value" of 255mA.

stringandstickytape:
Not sure if this is up to scratch - my C++ isn't that great - but does work and gives answers very close to what my ammeter says.  I used the "magic" mA value of 255 when I saw that you'd already assigned some slightly-"magic" behaviour to certain other values - but I suppose would be done better by adding a "use WS2815 power model" switch instead, but I didn't go that far.


stringandstickytape:
I made a bit of a mess of the commit.  Am done fiddling with it now :)

stringandstickytape:
Just realised it shouldn't display "(for most effects, ~1.5A is enough)" under these circumstances :/

stringandstickytape:
Hm. Maybe this should work using a specific LED type instead of via the mA setting? Views appreciated.

Aircoookie:
Thank you for this! Looks quite good to me :) Have you verified that the actual current consumption is now close to the estimated one?

For now the "magic value" approach is good, because we don't need another variable. Long term it might be easier to have a second variable for LED type, but this is not something we need right now.

I will merge your PR as soon as I add the Grouping support :)

stringandstickytape:
> Have you verified that the actual current consumption is now close to the estimated one?

Using the ammeter, I can see that at peak, each LED draws 13ma (3.94A total over 300 LEDs), but 12ma gives the right answers in WLED's calculations.

Having inspected the code, I now realise this is because of the 1ma standby per LED which is automatically deducted from the power budget.  I'm not sure why standby usage is relevant to this calculation, but presumably I have either reached the correct answer of 12ma, or it should be 13ma and the 1ma standby usage should be excluded from the calculation?

Issue #578, opened 2020-01-09
Effect "Police All" works different with >100 LEDs


Def3nder:
With less then 100 Pixels the effect does fill the whole strip solid with RED / BLUE...
...but with more Pixels it's working different:

See this video: https://www.youtube.com/watch?v=wDoUTj43Q94

I tried to change the build flags from PIO_FRAMEWORK_ARDUINO_LWIP2_HIGHER_BANDWIDTH (v2) to PIO_FRAMEWORK_ARDUINO_LWIP_HIGHER_BANDWIDTH (v1.4) but this did not have any effect.

I also tried to add segments: e.g, segment0 = 0...91, segment1 = 92...316 
After this the segment0 does display as expected (solid fill colors) but the 2nd segment stil has this strange behaviour.

Any suggestion how to change this ?

Aircoookie:
Hi!
Thank you for making me aware of this! This problem happens when the effect speed and/or LED count is high enough to move the effect by more than one pixel every frame. That will lead to some of the pixels not getting filled (and then potentially getting filled with the second color, which effectively makes the effect purple). Will fix soon!

Aircoookie:
also happens with Sinelon by the way, and Juggle (although the effect is quite nice with juggle ang i wouldn't neccessarily change that)

Dreamoffice:
@Aircoookie 
i have another issue with Police All Effect. with 100 LED in the preferences it works like a charm but with more (i have 816) it works like on the video.
[https://youtu.be/ESaA0FjHbnU](url)

Def3nder:
closed with PR #634

PR #599, opened 2020-01-16
Fix gaps in "Police All" and the three "Sinelon" effects (fixes issue #578)

Def3nder:
Fixes gaps in the effects "Police All" and the three "Sinelon" effects.

This iteration works with SEGENV.aux0 to remember the last position and a loop filling the colors in that range.

This fixes issue #578


Def3nder:
I've updated the code from a "trial & error" aproach to a more exact one.

Works very well with both effects.
Juggle does work a lot different so that I did not find an easy way correcting this (and as you said: juggle does not look bad even with high speed values 😄 )

Aircoookie:
Hi, just quickly asking if you closed this intentionally? I was going to maybe implement something else for police all (filling both areas each frame, that will also improve color transitions that happen while the effect is running). For sinelon it seemed like a good fix though!

Def3nder:
Hi Aircoookie, sorry, no I did not do this intentionally 😢 

...what I did is: I changed the fork-name, because it now does contain the sinelon, too.
Did not think about a new fork will auto-close this ...

I open another PR...

PR #600, opened 2020-01-17
Plasma effect

Def3nder:
The speed slider influences the length of the waves  (more speed = smaller waves / more waves).
The intensity slider influences the hue variation from the primary color that was choosen - far left = nearly no variation, far right = +/- 30% variation in hue value.


Aircoookie:
Hi,
thank you, this effect is awesome! I've removed the automatic palette changing for now as I plan to implement that for all effects soon (and is was taking up static memory)

PR #603, opened 2020-01-19
added support for 21-key remote

recliq:
Hi,

I got some 21key remotes and added the codes, maybe it's worth adding it to WLED.
It's this remote: https://images-na.ssl-images-amazon.com/images/I/51NMA0XucnL.jpg


Aircoookie:
Thank you!

PR #604, opened 2020-01-19
Added percent display effect.

recliq:
This is not really an effect but usefull for status displays of any kind.
Intensity values from 0 - 100 represent the pertentage amount of LED that light up.
It uses col0/palette for on and col1 for off.

PS: I'm not very familiar with pull requests so things might go sideways.. :/


Aircoookie:
Looks good!

Issue #615, opened 2020-01-23
Home Assistant Native Integration - Slow to update state changes vs MQTT - Polling issue?


baremoney:
Hello, I am running a **home assistant** automation that is triggered when the nightlight **attribute state changes**. The problem is the **native integration is slow** and can on average take anywhere from a couple of seconds to a few seconds to reflect the state change which triggers the automation. While using **mqtt before**, the changes were **updated much faster as if instantly**. **Does the native integration use some kind of polling interval to reduce overhead and is there a work around or way to have mqtt publish the nighlight state?** (By testing manually my first guess is that the bottleneck is not the reaction time of the esp wled device or the automation, but only how long home assistant takes to acknowledge the change in state of the nightlight via the home assistant wled native integration).

Def3nder:
Hi @baremoney, this is a question belonging to the logic used in home assistant.
@frenck was the one that wrote that awesome WLED integration. I cannot really read the python code, but it looks like using a timer (polling).

The WLED itself does only send MQTT notification on any change in the UI as far as I know.

So it looks like the WLED integration is using the JSON API rather than MQTT ?

frenck:
Yes, the WLED integration uses the native API and does not require or rely on MQTT. There is no streaming kinda endpoint available on WLED, which makes it a polling integration.

However, since the device is local, we use a relatively high polling rate. For most use cases, it should not matter that much. Changes from HA are send instantly and applied instantly by WLED and internally instantly reflected withing HA itself, however, if changed from an outside source, HA needs max 9 seconds to catch up.

Def3nder:
@baremoney, does this answer your question ? If you need the nightlight change even faster than this 9 seconds, you could create an MQTT sensor for this in HA. 

The problem with this is that the MQTT response does contain XLM and no JSON 😢 
(and an easy way to create a sensor from an MQTT topic does rely on JSON...

@Aircoookie: is there a reason why you used XML in the `/v` topic in MQTT ?
...could that be changed to deliver the JSON part that is in the `state` tree ?

baremoney:
@Def3nder @frenck Thanks for the helpful overview about the integration! Having looked at the xml mqtt published I see the entries:
`<ns>0</ns>
<nr>0</nr>
<nl>0</nl>
<nf>0</nf>
<nd>1</nd>
<nt>1</nt>`
The wiki says mqtt is being overhauled by another user so I guess they have added the timelight option i was looking for , ie, nl ?

[...]

PR #617, opened 2020-01-23
MQTT JSON respone for topic "/v"

Def3nder:
As a result of #615 I added a compile option to used JSON in the response to the "/v" topic instead of XML.
As the JSON "state" serialization does also include all segments, this response could grow to about 4k (and the segments are not included in the XML) so I changed the `serializeState` function to have an argument `bool includeSegments` to switch between "with/without segments".

I would suggest to add this to the MQTT settings later...

What do you think ?


Def3nder:
Hi @Aircoookie ,

the implementation does do what is expected, however **the heap constantly decreases** 😢 
So I kind of mirrored the code from the WLED19_JSON.INO, but it does allocate memory everytime.

**So, please do not commit.**

Aircoookie:
Hi @Def3nder !
It is a really good idea to support the JSON API via MQTT and something I should have added support for long ago.
Good thing you noticed the memory leak, I will try to fix it soon :) Maybe we can also find a way to include the segments, maybe in a different topic.

Def3nder:
Hi @Aircoookie, I wonder why the memory leak is only there in this "add-on" and not in the JSON API.
I tested the JSON API with a flood of requests and the heap did stay at the samd level.

Def3nder:
Hi @Aircoookie, I deleted to allocated JSON-object after using it - now the leak is away 😄 
Should be safe to commit now.

...something similar should be included in the WLED19_JSON.INO as well, shouldn't it ?

It would suggest in function `serveJson`:

change from:
```
[...]
  response->setLength();
  request->send(response);
}
```
to:
```
[...]
  response->setLength();
  request->send(response);
  delete response; response = NULL;
}
```
?

I wounder why this was necessary, because up to now I thought that objects would be deleted automatically if the scope has been left ... strange 😣

Def3nder:
Hi @Aircoookie, I just thought about publishing the segments information:

if we push this to the MQTT broker each single UI change, this could lead to a delay.
I generated a dummy JSON with the max. amount of segments and with (looking into the future) 5 bytes per color. This is **2244 bytes** long (2030 of this are the segments):
[WLED_Status_incl_segments.json](https://github.com/Aircoookie/WLED/files/4117701/WLED_Status_incl_segments.json.txt)

What do you think ? Should I write some code to include the segment information maybe as "/seg" in the MQTT response ?

Aircoookie:
Hi @Def3nder thank you for trying that out! Yes, I believe over 2k might be too much. I remember the MQTT homeassistant autodiscovery, it was 2.4kB and caused reboot loops and/or MQTT disconnects...

Another possible approach would be to have a topic for every single segment and only publish to the a segment's topic if that segment was updated. This might be more viable memory wise, but we would need to test if subscribing to so many topics at once could cause issues.

Def3nder:
Hi @Aircoookie, or we do not publish segment changes and only offer the posibility to request the segment via MQTT. So the other system could send and /WLED/light-name/seg[0..9] and will get the content of one single segment as JSON in return.

What about this ?

wled00/wled19_json.ino
```diff
@@ -191,8 +191,12 @@ void serializeSegment(JsonObject& root, WS2812FX::Segment& seg, byte id)
 	root["rev"] = seg.getOption(1);
 }
 
-
 void serializeState(JsonObject root)
+{
+  serializeState(root, true);
```
psi-4ward, line 196: I had to move this call below it's definition to work

Aircoookie:
Closing since outdated.
Still I hope to get around to adding MQTT JSON sending soon, one topic for info, one for state without segments, and one for each (updated) segment :)

PR #634, opened 2020-01-27
Fix Police All, Sinelon and Lighthouse (Comet) effects

Def3nder:
2nd time ...I'm still learning **Github** 😄 

I noticed yesterday that the `Lighthouse` effect has the same problem, too.

...and there were some typos in the `NpbWrapper.h` ...


fishbone-git:
Have you tested the Sinelon/Dual Sinelon options?  I can't seem to get a single trail, even when setting only one colour in the UI.

Def3nder:
Hi @fishbone-git, thank for the testing - I did split contents inside the `if (dual)` into parts and then forgot to double the statement as well ...

Is corrected now and tested 😄

fishbone-git:
gaps seem to be a common problem with the beatsin* functions.  Oddly, few solutions come up, so good to have something in WLED

PR #643, opened 2020-01-29
Correct start of Lighthouse effect at high speeds

Def3nder:
Currently when using a high value of the FX speed, the start of the effect (1st active pixel) moves from the 1st pixel to the 2nd. then the 3rd. etc. (the first pixels remain at the color(1).

This change covers this situation so that the effect always does start at pixel 0.


Aircoookie:
Thank you for catching that!

PR #659, opened 2020-02-04
Fix Chase Effects at high speeds + "Two Dots" Effect at low speeds

Def3nder:
Hi @Aircoookie, the Chase Effect (and most of the effects that use the WS2812FX::chase function did show gaps at high speed values.

This fixes the gaps and fixes the "random" variant, too (as this relies on SEGENV.step from 0.8.6 version of the effect that was not present in the actual one.

...later I noticed that the efect "Two Dots" had some flicker at low speeds.
This was introduced from the gap-calculation in "police_base" and does occur, if the idexR did not change from one to the other call. I've added an if-statement to check that - it's okay now.


Aircoookie:
Thank you!

PR #664, opened 2020-02-08
HTML API "/url" for the current effect settings

Def3nder:
I added "/url" as request to wled18_server.ino the returns a HREF link with all the parameters used in the current running effect.

E.g. I'm runing "Ripple" with Palette "Party" dimmed to brightness of 36 with a reddish color(0), speed at 1/4 and intensity at 3/4. This results in the following URL:

[http://192.168.178.143/win&A=36&CL=hFF2F05&C2=h000000&FX=79&SX=71&IX=171&FP=6](http://192.168.178.143/win&A=36&CL=hFF2F05&C2=h000000&FX=79&SX=71&IX=171&FP=6)  

The idea behind this is to create a section in the wiki with the "best-off" effects and the settings used for them (as the settings make a huge difference for a lot of the effects).

A second idea would be to add maybe the "TOP10" of them to the favourites page (or somewhere else), so that a new user can get a good starting point.


Def3nder:
Hi @Aircoookie,

yesterday there were some people asking about a posibility to kind of backup the settigns.

Wouldn't it be an interims solution to included this PR as a sort of "backup the preset of the running effect" - so the users could activate their presets one-on-one and call http://[WLED-IP]/url and get an URL that can be saved to a text file.

If - for some reason - they loose the settings, they can easily paste the URLs in a browser and get all their custom effects back 😄

Aircoookie:
Hi @Def3nder !
Initially I thought this PR was neat but not really a neccessary feature as you could get all the values from the normal `/win` XML response.
What I realized though is that even if you know the HTTP API commands by heart (which certainly not everybody does), you have to convert them to a valid request URL by hand, which is not convenient. Also 3 people have thumbed up this feature so it seems like it would be well received :)

Merging!

PR #665, opened 2020-02-08
DMX refactoring

pille:
implement different DMX modes for E1.31:

  DISABLED: don't act on packets
  SINGLE_RGB: treat all LEDs the same and controll them using 3 channels: RGB
  SINGLE_DRGB: as above, but has an additional first channel for Dimmer
  EFFECT: not a realtime mode. disables fadeTransition to reduce delay. just exposes parameters as 11 channels to trigger effects locally: Dimmer Effect Speed Intensity Palette PriRed PriGreen PriBlue SecRed SecGreen SecBlue
  MULTIPLE_RGB: legacy mode. address each LED individually with 3 channels: RGB.
  MULTIPLE_DRGB as above, but has an additional first channel for master Dimmer

add client stats (IP + user agent)
add support for DMX address (so you don't need a whole universe per fixture)
skip out-of-order packets
add DMX settings (this increases EEPROM layout version)

### DIFF
--- a/wled00/data/settings_sync.htm
+++ b/wled00/data/settings_sync.htm
@@ -88,6 +88,14 @@
 		Multicast mode: <input type="checkbox" name="EM"><br>
 		E1.31 start universe: <input name="EU" type="number" min="1" max="63999" required><br>
 		<i>Reboot required.</i> Check out <a href="https://github.com/ahodges9/LedFx" target="_blank">LedFx</a>!<br><br>
+		DMX start address: <input name="DA" type="number" min="1" max="510" value="1" required><br>
+		DMX mode: <input name="DM" type="radio" value="0"> disabled<br>
+		<input name="DM" type="radio" value="1"> Single RGB (3 Channels for all LEDs: Red Green Blue)<br>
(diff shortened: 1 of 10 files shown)


Aircoookie:
Thank you very much, those new modes are really neat!
I've made some minor adjustments, like making sure that the Multi RGB is the default mode when people update and adding white channel support to the 11 channel mode :)
I believe that the out-of-order packet check and DMX address option are really helpful even for people who only run it in Multi RGB mode!

pille:
i'm mostly fine with your changes.
but why did you remove the E1.31 user-agent string?

Aircoookie:
I just don't think it's that useful. Even IP is not really needed since you will know the controller the stream comes from anyways in 99% of cases. Also not all software seems to send it. In my testing with LedFx it just said "default ua string" or something along those lines. Lastly, the string parsing could slow down the e131 handler a bit or cause heap fragmentation if it changes often.

pille:
the string is static and helpful for debugging, since you can have multiple sources.
that sounds, like ledfx needs a pull request ;) QLC+ reports a proper UA.
i did not experience speed-related problems.

PR #671, opened 2020-02-10
Simplify Code for Gradient / Loading effect

Def3nder:
You mentioned sometime to simplify the code - what do you think of this try ?


Aircoookie:
Thank you, this looks great!

Issue #613, opened 2020-01-22
Effect in 2 different length strips to have the same duration


sirs2k:
Congrats on this fantastic project Aircoookie!

I have 2 different length strips running parallel, one with 150leds and one with 130leds

I'd like for wled to automatically adjust the speed of both so that a "wipe" effect for example can start and finish on both at the same time.

That would be great!
Thanks all

Aircoookie:
Hi,
are you running these on separate ESPs? If yes, you can sync them up via UDP by pressing the sync button in the UI.
If they are wired in series, you can create two segments and both set them to the same effect.
If they are wired in parallel, unfortunately the first 130 LEDs of both strips will always be identical.
Also make sure you have WLED 0.9.0 installed, I believe it hasn't been too long ago that I added Wipe having the same speed everywhere regardless of LED count :)

sirs2k:
Thanks for the prompt reply.
Yes they are on 2 different ESP's. If u do connect them in series, which pins do I use? D4 and which? 

Yeah I meant to have the opposite effect. 

Yeah that confirms my fear. So I'll have to set their speeds manually to achieve them both starting/finishing at the same time. That feature would be very handy though as in many situations you have different lengths of strips that you want to have the duration of wipe across.

Another question, when I do sync effects via UDP, most times they take effect on each strip with massive delay between.
For example, it effects strip 1 and then after 5 seconds the changes take effect on strip 2, many times it doesn't even sync. How can I fix that?

Def3nder:
Hi @sirs2k, connection "in series" does mean that you connect the last `DOUT pin` from the first strip to the first `DIN pin` of the second strip:
![image](https://user-images.githubusercontent.com/33399267/72969792-86a4fd00-3dc6-11ea-8aab-6227d868c44e.png)

After having done this (only one ESP with both strips connected to each other) you need to change in the **LED preferences** the `LED count` to `280` and then on in the main UI on the **Segments tab** change the `Stop LED` of segmant 0 to `150` add another segment 1 with `Start LED` set to `150` and `Stop LED` set to `280` (remember to click on the ✔️ to accept the LED changes every time you change numbers here).

Most effects have the same total duration regardless of the length - so they should be both start AND finish at the same time (WIPE is one of them: `ledIndex = (prog * SEGLEN) >> 15;` )

[...]

PR #672, opened 2020-02-10
Disable WiFi Sleep Mode

Def3nder:
Hi @Aircoookie,

in issue #613 @sirs2k reported a "non-working" UDP sync and I tried to solve the issue with him.
We somehow got stuck, but he mentioned that his UniFy access point reports the ESP as having "Power Saving Mode" on.

I have one "test-ESP" (an ESP32) connected to USB-power all the time and did get reset (with reason 12) several times a day. After changing the WiFi Power Mode to "None", this seems to be gone (no more reset since then).

Do you want to give this a try ?

I included the WiFi Power Safe Mode ("Sleep") in the JSON-info 
and added the "Reset Reason", too.

Feel free to in/exclude this 😄


Aircoookie:
Hi,
thank you, very good to know that sleep can cause issues!
I will include this, but I will make it optional, because it will of course increase power consumption.

sirs2k:
Thanks guys, I'm happy to test a version without sleep mode if you like 🙂

Def3nder:
thx - can be closed then ...

PR #688, opened 2020-02-16
try pio workaround for pio4.2.0

fishbone-git:
From the link:
https://community.platformio.org/t/pio-4-2-0-complile-error-for-asynctcp-id1826/12025/4

"Sorry, this is our bug and has been fixed in https://github.com/platformio/platformio-core/commit/5cc9a328ab4d64c42f3cda714f65e150eca31128 4

We plan to publish bug fix release of PIO Core 4.2.1 next week, A temporary solution is to use lib_ignore = AsynTCP or lib_compat_mode = strict in platformio.ini."

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -127,6 +127,8 @@ build_flags =
   ${common:esp8266.build_flags}
 lib_deps =
   ${common.lib_deps_external}
+lib_compat_mode = strict
+lib_ignore = AsynTCP
 
 [env:d1_mini]
 board = d1_mini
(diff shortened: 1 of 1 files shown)


Aircoookie:
Awesome, thanks for finding out! Also good to know that the cause will be fixed by the next release :)

debsahu:
@Aircoookie Just checking in, this is missing a ‘c’: AsyncTCP and currently it’s AsynTCP. I’ll do a PR to help, just tag me next time so that I get notified.

fishbone-git:
oops, looks like I propagated a typo.  This probably means that the lib_ignore part of this fix isn't needed since the current master no longer fails (or Travis is already using pio4.2.1)

PR #702, opened 2020-02-21
Refactor callMode

Def3nder:
Hi @Aircoookie,

the use of const.h makes the code much more "readable" - thank you for that !

...I catched that idea and did the same with the variable `callMode` that is used in the functions 
`notify` and `colorUpdated`.


Aircoookie:
Thank you!
When I added the const.h I knew there was something I had forgot to refactor 😅

PR #713, opened 2020-02-24
Correct sun rise

Def3nder:
sorry I must have been somehow confused - using the col[] array (+ colSec[] ) for the SunRise effect in the NightLight.


Def3nder:
The build is somehow too big for an ESP01 @ 512k flash.

Aircoookie:
I know :) Had that problem as well in master, because I changed the Arduino core to use for 512k as well, so don't worry.

PR #721, opened 2020-02-25
Refactor platformio.ini

Def3nder:
Hi @Aircoookie,

I just couldn't get the IR remote from one of my strips working this morning...
...until I noticed that the -D WLED_DISABLE_INFRARED was in the build_flags 😢 

...so I decided to refactor the platformio.ini based on the one from ESPurna.

The environments are very small now:

```
[env:d1_mini]
board = d1_mini
platform = ${common.platform_latest}
board_build.ldscript = ${common.ldscript_4m1m}
build_flags = ${common.build_flags_esp8266}
``` 

Do you like this approach ?


Aircoookie:
I love this style! It is a lot more readable and easy to use!

Just two things before I merge:
- There are special Travis environments in the bottom to test against different arduino core versions. That can be useful, but I don't think we need them for now. The 2.3.0 causes the build to fail, as WLED doesn't run on Core earlier than 2.4.2.

- It seems like you accidentally commited wled00.ino.cpp?

Other than that, this is awesome! Thank you :)

Def3nder:
Hi @Aircoookie,

agree - did just a copy&paste for the travis section - I removed this section as well as the accidently pushed wled00.ino.cpp 😄

Aircoookie:
Awesome!

PR #722, opened 2020-02-25
Add switch pin

zewelor:
Add switch pin option. Also fix ifdefs when button pin is -1

### DIFF
--- a/wled00/NpbWrapper.h
+++ b/wled00/NpbWrapper.h
@@ -17,6 +17,10 @@
 #define BTNPIN  0  //button pin. Needs to have pullup (gpio0 recommended)
 #endif
 
+#ifndef SWITCHPIN
+#define SWITCHPIN  -1  //switch pin.
+#endif
+
 #ifndef IR_PIN
 #define IR_PIN  4  //infrared pin (-1 to disable)  MagicHome: 4, H801 Wifi: 0
 #endif
--- a/wled00/wled00.ino
+++ b/wled00/wled00.ino
@@ -330,6 +330,9 @@ bool buttonLongPressed = false;
 unsigned long buttonPressedTime = 0;
 unsigned long buttonWaitTime = 0;
 
+//switch
+bool switchStateBefore = LOW;
+
 //notifications
 bool notifyDirectDefault = notifyDirect;
 bool receiveNotifications = true;
--- a/wled00/wled05_init.ino
+++ b/wled00/wled05_init.ino
@@ -96,8 +96,12 @@ void beginStrip()
   // Initialize NeoPixel Strip and button
   strip.setShowCallback(handleOverlayDraw);
 
-#ifdef BTNPIN
+#if BTNPIN >= 0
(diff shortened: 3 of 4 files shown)


Def3nder:
Hi @zewelor, what should be the use case of this switch ?
...right now it does exactly the same as the button, doesn't it ?

Another thing: there is nothing about debouncing in your code - so this could lead to more than one "switch action" at a time.

Aircoookie:
I believe adding switch support is indeed a good idea - another user on the forum has requested it as well. However, I would rather implement it as an extension to the button code rather than a new pin + logic. Then we could also run different macros when the switch is turned on/off. And it will make it easier to implement multi button/switch support in the future.
But thank you for the PR @zewelor , this is a good starting point!

zewelor:
By switch I understand something like this https://www.google.com/search?q=rocket+switch&sxsrf=ALeKk03Hy7usfUTNGZ00En6maR1Dep74JA:1582714666057&source=lnms&tbm=isch&sa=X&ved=2ahUKEwirwM3Lh-_nAhUVAhAIHTVZDakQ_AUoAXoECA8QAw&biw=1536&bih=879&dpr=1.25 . It differs from button, it stays in new position, button only changes state for short period of time.

Right some debounce code would be good. I can add it if it will be accepted ?

Def3nder:
Hi @zewelor, hmm - okay - in that case a rebounce would not be needed because it will stay "for a long time" in that position. 

Suggestion: what about adding another state to the button ?

Now we have `shortPress` and `LongPress` (> 0,6 sec).
We could add the status `hold` with e,g. > 2 sec.

If you then attach your switch to the button PIN, you get your state with the new "hold-action".

@Aircoookie: what do you think of this ?

Aircoookie:
@Def3nder I like the hold-action idea, the only drawback though is that it would take those e.g. 2 seconds after switched until the action is carried out. What I'd like about that solution though is that the short-press action still works by quickly operating the switch.

My idea was adding a "release" action. If it is set, long and double press are disabled and short press action will be fired as soon as the button/switch is pressed (except debounce of course) and the release action is fired when it is released.

zewelor:
I think mixing button with switch could end up with some conflicts. In exmaple with "release" action. Rocket switch doesn't really have any realease action. Its always switched, one position or another. Also any double click etc, its much harder to do witch switch, and if switch is hard to operate, it will be difficult to do any multiple clicks.

There is also some conflict with pins, that you can use for button vs for switch. For switch you cannot really use gpio0 / gpio2 / gpio15. If you use it for switch, you have to be sue to leave switch in correct position, to allow esp to boot properly ( https://cdn.instructables.com/FT4/3DPW/ICE0BBHM/FT43DPWICE0BBHM.LARGE.jpg ). I think it would be better to forbid using that pins for switch at all. When used as button its not a problem ( unless button get stuck etc ).

pepeEL:
Any news... Is any chance to add switch pin for switch monostable ?

zewelor:
Can we merge it?

PR #730, opened 2020-02-27
Added heltec_wifi_kit_8 board

srg74:


Aircoookie:
Thank you for the PR! I don't believe there is a reason to include it though as it will provide the same binary as d1_mini environment, so it will just increase time neccessary to build all env's :)

Actually I would like to shorten travis build duration so that it only does nodemcu + esp01 + esp01_1m + esp32. Then the build would only take 5 minutes.

srg74:
Not a problem. I agree with you.

PR #736, opened 2020-02-28
Add all release environments to platformio.ini

Def3nder:
Hi @Aircoookie,

I added the custom configurations (like LEDPIN or WS2801 and APA102) to the environments so that we get all configured in the PIO and the users can build everything like distributed in the releases.


fishbone-git:
As a broder "solution" why not expand the platformio_override example to have all of the disable, enable, pin_option flags so that it's a one-file point for users to be able to tweak the build for the interface-level customizations?  It already has the disables, but LEDPIN, IR_PIN, BTN_PIN, RLYMDE, etc can be added

Def3nder:
like this ?

fishbone-git:
i would go with copying the comments too so it's clear what each means. so, if all you want to do it make WLED work with your favourite flavour of ESP board and breakout, then you can follow lots of online how-tos for setting up platformIO in various IDEs and then just make a platformio_override with your tweaks.  At least until any of these are configurable in a UI.

Aircoookie:
Hmm does anybody know how to exclude some environments from travis? 
I really like this PR but ideally i'd like travis to only build the `d1_mini`, `esp32_dev`, `esp01` and `esp01_m` environments so that it doesn't take 15 minutes, but only 5.

`default_envs` is ignored by CI as it seems...

fishbone-git:
perhaps change the travis.yaml to use platformio run instead of platformio ci in the script section.  the run command follows platformio.ini (and you'll need to change the platformio.ini to uncomment the release build line).  Travis does allow some custom files, but I've used this for a secrets file so don't know if platformio.ini is also a possible option.

fishbone-git:
ah...something like this in travis.yml:
before_install:
  - mv travis_platformio.ini platformio.ini

The residual risk being an issue in the platformio.ini file for other builds or forgetting to sync up changes in the common sections.

Def3nder:
...unfortunately I cannot compile with -D DEBUG_ESP_CORE enabled 😢 
This does lead to errors in debug.h that I just cannot resolve.

...I'm just trying an expplicit "include" of environments in Travis.Ci ...

Aircoookie:
Awesome, thanks!
The debug not working is unfortunate, but I don't think it is a major issue for now, we can look into it later.

Ready for merge?

Def3nder:
DEBUG is building now, but the "include" in the YAML did not work 😢 

but: **ready to merge.**

Will have a closer look into `How to only compile specific builds with Travis.CI` ...

Aircoookie:
One solution would be to have a `platformio_travis.ini` file that contains just the four envs. I don't really like that approach though because it essentially duplicates the environments and any change to the regular platformio.ini would also need to be mirrored there. There has to be a better solution 😅

platformio.ini
```diff
@@ -88,7 +89,7 @@ debug_flags = -D DEBUG=1 -D WLED_DEBUG -DDEBUG_ESP_CORE -DDEBUG_ESP_WIFI -DDEBUG
 #     TLS_RSA_WITH_AES_256_CBC_SHA / AES256-SHA
 #  This reduces the OTA size with ~45KB, so it's especially useful on low memory boards (512k/1m).
 # ------------------------------------------------------------------------------
-build_flags = -g -w -DMQTT_MAX_PACKET_SIZE=1024 -DPIO_FRAMEWORK_ARDUINO_LWIP2_HIGHER_BANDWIDTH 
+build_flags = -g -w -DMQTT_MAX_PACKET_SIZE=1024 -DPIO_FRAMEWORK_ARDUINO_LWIP_HIGHER_BANDWIDTH
```
zewelor: Is this change intentional ? Also maybe it would be better to try this: https://github.com/Aircoookie/WLED/pull/732#issuecomment-592110043

PR #738, opened 2020-02-28
Fix 24-key IR remote

Def3nder:
I've found the error - the notification was in the wrong scope (one bracket too early).

This way the `colorUpdated(NOTIFIER_CALL_MODE_BUTTON)` was only called for repeating changes (like brightness) or for the remotes that send code > 0xFF0000.


Aircoookie:
Thank you very much for finding that one!
I didn't because I didn't notice that `IR_24` and `IR_24_OLD` are different types 😄

PR #739, opened 2020-02-28
Travis.CI builds only for default_envs

Def3nder:
Hi @Aircoookie,

I found the solution 😄 - took a while but then it worked.

Instead of calling `platformio ci --project-conf=./platformio.ini` we call `platformio run` (the same as one a local PC). Then it does build the environments specified in `default_envs` only.

I've changed the PIO to use 
```
default_envs = d1_mini, esp01, esp01_1m, esp32dev
```
Travis build-time is reduce to around 5 minutes again.

AND: (bonus) I added build_cache_dir to the PIO.ini and in the Travis cache section.
Now if you push only little changes to Git, the travis build will only compile the change files and will retrieve the rest from the build-cache,

A 2nd compile of all 4 envs did take just 55 seconds.


Aircoookie:
Amazing work, thank you! I didn't even know this caching mechanism was usable in Travis.

PR #746, opened 2020-03-02
New IRremote functions

Def3nder:
Hi @Aircoookie,

two of the remotes can change effect-speed and effect-intensity.
These settings don't do anything when the Effect "Solid" is running.

In that case the color (actual the hue of the color) will be changed with speed increase / decrease.
and the saturation will be changed with the intensity increase / decrease.
(the sliders will stay where they are in these cases).


Aircoookie:
Hi @Def3nder ,
thank you, I like these additions!
I recently removed the `gGradientPaletteCount`, that's the reason the build fails. Instead of `maxPaletteIndex` I'd use `strip.getPaletteCount() -1`. I will quickly make this change and merge :)

Def3nder:
Thx - what about the conversion warnings ?

Are they show at your side, too ?

Def3nder:
hmm - let me rework this - sorry - was submitted too fast

Aircoookie:
This looks good, thanks for the quick fix!

PR #748, opened 2020-03-02
sync compiler values to the ones in the Wiki

Def3nder:
...so we get nearer to the target that if a user clones the repo and compiles without changes, that he gets exactly the same binaries.

Did I overlook something ?
Should be quite close right now.

One question: how do you create the release binaries ?


Def3nder:
...found a last one:

the environment for ESP32 with LEDPIN 16 😄

Aircoookie:
You are right about the gpio16 esp32 version! Let's stay on core `2.6.3` though, I will update the wiki. (0.9.1 release binaries were compiled with 2.6.2, and 2.6.3 is supposed to be even a bit more stable). It was only left on 2.5.2 in the wiki because I hadn't yet done enough 2.6 testing two months ago and after that forgot to update my recommendation 😄

Def3nder:
okay - you are right - will you add the ESP32-PIN16 version ? I will close this, okay ?

...saw you added the compiler flags for the IR library - this I had forgotten - sorry.

Aircoookie:
No worries, thanks :) I just added the pin 16 environment!

mike2nl:
@Aircoookie @Def3nder 
So core is now 2.6.3, right?
When 2.7.x is out i will do a lot of tests. Then i will report but i think it will mot before end of April so the guys at esp8266/Arduino get it right.

Aircoookie:
@mike2nl yep, PIO is using 2.6.3 and I also changed the wiki recommendation to that. 0.9.1 builds are still using 2.6.2 and Jason recommended we use 2.6.1. So far 2.6.3 seems good though. Excited for 2.7.0!

PR #759, opened 2020-03-06
Usermod: Project Cars rpm indicator and shift lights

jwingefeld:
turns WLED lights into a rev light and shift indicator for the game Project Cars.


Mariu86:
It's working with PS4 ?

jwingefeld:
> It's working with PS4 ?

yes, works with PS4 as far as i know. I only tested xbox. But the game should be the same.

PR #766, opened 2020-03-10
fixes #742

### LINKED ISSUE #742: E1.31 Issues
I started having an issue with new FW release. Transmitting E1.31 from LEDFX to my 300 LEDs running on ESP8266 causes issues with what appears to be the last 150 LEDs. It causes them to turn a solid color and only randomly change. I did not have this on my previously compiled version before the PR of the newest version. I did revert back to a compiled bin I had from January 10th with no issues.

pille:
* track E1.31 seqcuence numbers for each universe individually
* add E1.31 packet out-of-sequence config option (on/off)
  * extends EEPROM format

fixes #742

### DIFF
--- a/wled00/data/settings_sync.htm
+++ b/wled00/data/settings_sync.htm
@@ -85,6 +85,7 @@
 		<h3>Realtime</h3>
 		Receive UDP realtime: <input type="checkbox" name="RD"><br><br>
 		E1.31 (sACN)<br>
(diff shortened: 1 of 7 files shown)


Aircoookie:
Awesome! I was suspecting the universes to have unique sequence numbers, so good to know that was indeed the cause of the issue.

Issue #462, opened 2019-12-15
Integration with DMX


raresserban:
Hello,
I want to make a usermod to integrate this library with a DMX light, similar to this:
https://www.youtube.com/watch?v=4PjBBBQB2m4

What I was thinking is to have the light as another pixel in the strip, but I am not sure how I can do that. If that is not possible, then I would like at least for the effects to be applied to the DMX light as well.

How can I have access in the `userLoop` function to the current effect, or the value of a pixel from the strip so I can update my DMX light?

Aircoookie:
Hi!
That sounds like a really cool idea! You can get the color of any pixel quite easily:
Just use `strip.getPixelColor(i)`. That will give you the color in uint32_t format. If you need the R,G,B components, you can do:
```cpp
uint16_t pixel = 0;
uint32_t in = strip.getPixelColor(pixel);
byte r = in >> 16 & 0xFF;
byte g = in >> 8  & 0xFF;
byte b = in       & 0xFF;
```

Looking forward to what you come up with and let me know if you need anything else :)

jwingefeld:
> I want to make a usermod to integrate this library with a DMX light, similar to this:
> https://www.youtube.com/watch?v=4PjBBBQB2m4

Hello raresserban,

i am currently working on this. It already works, i just have to make it "user servicable". Let me know if you want to try it out.

/Jan

raresserban:
@jwingefeld It would be great if you could share the code!I tried something but with not luck so currently I am waiting for a new DMX cable, didn't have time to do this for christmas.

jwingefeld:
I will let you know when it is usable. Should be a matter of days.

Please note that you will need a MAX485 chip connected to your ESP8266.

abyssdj:
> 
> 
> I will let you know when it is usable. Should be a matter of days.
> 
> Please note that you will need a MAX485 chip connected to your ESP8266.

I would also be interested in this. I've recently discovered WLED and it fits my requirements perfectly for LED strips, but would also be great to have a DMX counterpart, so that all the nodes use the same software. Extra bonuses would be "effects" for pan/tilt, etc to be able to run a basic display as a fallback when not receiving sacn.

abyssdj:
Extending on the above, would it be possible to have Artnet as an alternative option to sACN? I find it faster and more suitable in my case.

Aircoookie:
@abyssdj yes, ArtNet support has also already been requested in https://github.com/Aircoookie/WLED/issues/417

[...]

PR #809, opened 2020-03-28
DMX: Ability to work together with LEDs and set segment and start LED for fixtures

raresserban:
Hello, like I said in #462 , I added ability for DMX fixtures to work together with LEDs.
This is achieved by adding a setting for DMX Segment and DMX Start LED, so DMX fixtures can be part of a particular segment, and fixtures can also start from a specific LED inside that segment.

This allows more powerful control to position DMX fixtures where you want on a strip. For example, you could have an LED strip, followed by 2 fixtures, which can be set to begging at the LED the strip ends, so the color effects will continue on to them. Or you can have the DMX fixtures in the middle of the strip, synchronized with what happens there.

This is my first pull request to an Arduino related project but hopefully it can be merged so others can use this also :)
If merged, I will updated the Wiki afterwards.


raresserban:
So any opinions on this?

Aircoookie:
Hi! Sorry for the delay on this!
I am currently in the process of refactoring the codebase to .cpp files! Will merge this for sure in a few days! I believe being able to choose the LED used for the DMX fixture is very useful :) Maybe I will only add the start LED part though, as effectively does the same as segment selection, but is even more powerful.

PR #817, opened 2020-03-31
Switched DMX-library

EckPhi:
For whatever reason ESP-DMX did not work for me with an ESP32, after some research I discovered some other libraries that had an implementation for ESP8266 and ESP32. I used LXESP32DMX for ESP32 and LXESP8266DMX for ESP8266. Since those are better supported I thought this might be of interest for the whole project as these libraries also support DMX-RDM (not implemented), which might be useful for someone.

According to https://github.com/cansik/esp-dmx-max485 the libraries I use should also be faster.

I did this mainly for myself, so no big deal if the memory usage gain is a deal breaker.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "lib/LXESP8266DMX"]
+	path = lib/LXESP8266DMX
+	url = https://github.com/claudeheintz/LXESP8266DMX.git
+[submodule "lib/LXESP32DMX"]
+	path = lib/LXESP32DMX
+	url = https://github.com/claudeheintz/LXESP32DMX.git
--- a/wled00/html_settings.h
+++ b/wled00/html_settings.h
@@ -186,6 +186,8 @@ function GCH(num) {
 }
 function mMap(){
   d=document;
+  directionPin=document.Sf.DP.value;
+  outputPin=document.Sf.OP.value;
   numCh=document.Sf.CN.value;
   numGap=document.Sf.CG.value;
   if (parseInt(numCh)>parseInt(numGap)) {
@@ -214,6 +216,8 @@ function S(){GCH(15);GetV();mMap();}function H(){window.open("https://github.com
 
 <i>Number of fixtures is taken from LED config page</i><br>
 
+MAX485 direction pin: <input type="number" min="0" max="39" name="DP" maxlength="2" onchange="mMap();"><br />
+MAX485 serial output pin: <input type="number" min="0" max="39" name="OP" maxlength="2" onchange="mMap();"><br />
 channels per fixture (15 max): <input type="number" min="1" max="15" name="CN" maxlength="2" onchange="mMap();"><br />
 start channel: <input type="number" min="1" max="512" name="CS" maxlength="2"><br />
 spacing between start channels: <input type="number" min="1" max="512" name="CG" maxlength="2" onchange="mMap();"> [ <a href="javascript:alert('if set to 10, first fixture will start at 10,\nsecond will start at 20 etc.\nRegardless of the channel count.\nMakes memorizing channel numbers easier.');">info</a> ]<br>
--- a/wled00/wled00.ino
+++ b/wled00/wled00.ino
@@ -23,20 +23,23 @@
 //#define WLED_DISABLE_INFRARED    //there is no pin left for this on ESP8266-01, saves 12kb
 #define WLED_ENABLE_MQTT           //saves 12kb
 #define WLED_ENABLE_ADALIGHT       //saves 500b only
-//#define WLED_ENABLE_DMX          //uses 3.5kb
+//#define WLED_ENABLE_DMX            //uses 10.7kb
 
 #define WLED_DISABLE_FILESYSTEM    //SPIFFS is not used by any WLED feature yet
 //#define WLED_ENABLE_FS_SERVING   //Enable sending html file from SPIFFS before serving progmem version
 //#define WLED_ENABLE_FS_EDITOR    //enable /edit page for editing SPIFFS content. Will also be disabled with OTA lock
 
 //to toggle usb serial debug (un)comment the following line
-//#define WLED_DEBUG
+#define WLED_DEBUG
 
 //library inclusions
 #include <Arduino.h>
 #ifdef WLED_ENABLE_DMX
-  #include <ESPDMX.h>
-  DMXESPSerial dmx;
+  #ifdef ESP8266
+    #include <LXESP8266DMX.h>
+  #else
+    #include <LXESP32DMX.h>
+  #endif
 #endif
 #ifdef ESP8266
   #include <ESP8266WiFi.h>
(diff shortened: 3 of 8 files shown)


Aircoookie:
@jwingefeld sorry if I overread it somewhere, but what do you think about this library switch? I'd say ESP32 compatiblility and being able to change the pin would be quite nice, especially given that the espdmx pin is the current default LED pin 2.

@EckPhi do you know what the extra direction pin is about?

Happy easter holidays :)

jwingefeld:
Ohai,

i've found a few issues with that PR.

1. doesn't compile with the arduino ide. The library is named LXESP8266UARTDMX, the include looks for LXESP8266DMX. Minor problem.
2. You are calculating the brightness for each individual channel. Every lamp i've seen so far have a dedicated brightness channel. Even the cheap ones from AliExpress. They are often called shutter, brightness, dimmer etc. This would be a double inclusion of the brightness value resulting in a logarithmic scaling. Ideal would be a check whether there's a brightness channel set or not. And then calculate brightness in the color channels, if there's none for that specific fixture.
3. Like you mentioned, the library is three times the size. But then again this is only ~6kb more.
4. On ESP8266 this does not work at all since you rely on several functions of the ESP32 library, which doesn't get included in the first place.

So in this form, this would break a lot. I would love to see this implemented, just for the performance gain. But on the other hand, the ESPDMX library is simple, straight forward and works great. Even on ESP32 from what i've seen. Maybe we can just improve this one. For example setting the pin without modifying the library itself.

Regarding DMX-RDM: can you give an example what this would be useful for? Lamps usually don't report that much back which could be useful for use inside WLED. I've seen overheat protection or color calibration issues. But you won't see this feature under $1000 USD for a single lamp.

What do you think? You can also contact me on discord. I'm JvPeek#8906 there.

EckPhi:
> @EckPhi do you know what the extra direction pin is about?

@Aircoookie I think this pin controls RE/WE of the RS485 transceiver. 



> doesn't compile with the arduino ide. The library is named LXESP8266UARTDMX, the include looks for LXESP8266DMX. Minor problem.

@jwingefeld To be honest, I didn't have ESP8266 available when I implemented this (I think I tried to compile it, and for whatever reason it succeeded). But I've now fixed it and tested it on an ESP8266.



> You are calculating the brightness for each individual channel. Every lamp i've seen so far has a dedicated brightness channel. Even the cheap ones from AliExpress. They are often called shutter, brightness, dimmer etc. This would be a double inclusion of the brightness value resulting in a logarithmic scaling. Ideal would be a check whether there's a brightness channel set or not. And then calculate brightness in the color channels, if there's none for that specific fixture. (TL;DR: If !brightnesschannel, then calculateBrightness)

Actually I implemented a search in the "DMXFixtureMap" for a shutter (5) and I only adjust the color channels if none was found, do I have to search for something else?



> On ESP8266 this does not work at all since you rely on several functions of the ESP32 library, which doesn't get included in the first place.

You are right, I have not excluded the semaphore part for ESP8266. (Should be fixed now).



> Regarding DMX-RDM: can you give an example what this would be useful for? Lamps usually don't report that much back that could be useful for use inside WLED. I've seen overheat protection or color calibration issues. But you won't see this feature under $1000 USD for a single lamp.

Right, it's pretty useless for lamps... Actually, my intention was that if it is needed in the future (because such lamps become cheaper or something like that), no additional library is needed and that simply the existing library can be used.

Happy Easter :)

pille:
@Aircoookie: did you already made your mind up about merging this? i want to implement a DMX proxy and the new lib provides more flexibility regarding pin configuration.

Aircoookie:
@pille in my opinion this can be merged. But I would still like a go-ahead from @jwingefeld for the improved version as I don't have the hardware and expertise to actually test the DMX output :)

pille:
i can do the testing.

pille:
641016038048fdbdbb7ab8ed8bdc3a75e652f9e4 still fails to compile because the DMX lib is not found (ESP8266/ESP32)

```
**************************************************************************
* Looking for LXESP8266UARTDMX.h dependency? Check our library registry!
*
* CLI  > platformio lib search "header:LXESP8266UARTDMX.h"
* Web  > https://platformio.org/lib/search?query=header:LXESP8266UARTDMX.h
*
**************************************************************************

     #include <LXESP8266UARTDMX.h>
                                  ^
compilation terminated.
*** [.pio/build/d1_mini_dmx/src/wled00.ino.cpp.o] Error 1
```

```
********************************************************************
* Looking for LXESP32DMX.h dependency? Check our library registry!
*
* CLI  > platformio lib search "header:LXESP32DMX.h"
* Web  > https://platformio.org/lib/search?query=header:LXESP32DMX.h
*
********************************************************************

compilation terminated.
*** [.pio/build/esp32dev/src/wled00.ino.cpp.o] Error 1
```

EckPhi:
> [6410160](https://github.com/Aircoookie/WLED/commit/641016038048fdbdbb7ab8ed8bdc3a75e652f9e4) still fails to compile because the DMX lib is not found (ESP8266/ESP32)
I did include the libraries manually (for ArduinoIDE [https://www.arduino.cc/en/Guide/Libraries#toc5](url) and for platformio [https://docs.platformio.org/en/latest/librarymanager/ldf.html#storage](url)). However I added the libs as submodule, have a look if it works for you (I'm not using platformio).

pille:
i pulled them in via
```
git submodule init
git submodule update --recursive
```
but they don't seem to be picked up automatically in platformio.

when i put the files inside the libs `src` directory into `wled00/src/dependencies/` it compiles.

Aircoookie:
@pille th LXESPDMX libraries are bsd 3 clause licensed. This allows us to include it in the `src/dependencies`, so let's go that route :) Also makes compilation easier for ArduinoIDE users as they don't have to install the library!

pille:
i did it for ESP8266 in my branch https://github.com/pille/WLED/tree/dmx-lib-switch-fix-esp8622 (based upon 40ec180b6b7139ee8a18f74d682aae69e80366e4) so that it compiles.

when i wanted to do the same for ESP32, it got a lot messier, as they bundle `rdm` lib in different versions and rely on their own libs without bundling them.

note: this PR is still on a branch, before ino-to-cpp-refactoring.

pille:
tested 2294e7e60062ca309530051d3be1b64f0e06e8fb on ESP8266:
* basically works
* can't change PIN (setting reset on reboot)
* occasional flashes (on changing colors)

pille:
decided to base my work for #862 on the proven implementation.

EckPhi:
Hi, sorry for the late feedback, but I was pretty busy in real life. Is there any need to use the LXESP libraries? If so, reimplementing the code should not be a big problem (including the fixes). However, if it is not needed and the ESPDMX-lib works fine with esp8266 and esp32, I would discard this pr.

Aircoookie:
Closing in favor of #2652

Issue #772, opened 2020-03-13
E131 wastes first packet for RGB modes


mjdieter:
With ver 9.1 the arlsLock() call in handleE131Packet() is after the pixels are set in various modes. arlsLock() then blanks all pixels (after the first packet set them) thereby wasting the first packet data. The previous version called arlsLock() before setting pixels.

mjdieter:
When I move the arlsLock() before the select in handleE131Packet() the whole system is more responsive to E131 packets and responds more like I would expect it to.

Aircoookie:
Thank you for finding this one, that was indeed not intended. Fixed!

PR #833, opened 2020-04-10
cleansup #772

pille:
return not needed anymore with relocation of arlsLock()


Aircoookie:
I still left that return in because otherwise it would `e131NewData = true`. We don't really want that as it causes extra calls to `show()` we only want in realtime mode.

pille:
you're right.

PR #850, opened 2020-04-15
Added BME280 sensor

srg74:
Added files to usermod directory and changes to platform.ini to support BME280 sensor.


srg74:
@Aircoookie could you please take a look what Codacy is don't like it.

Aircoookie:
I've checked codacity, either I don't understand the issue or it is incorrect, but it is nothing to worry about.

PR #858, opened 2020-04-20
travis: add own build targets that enable all features

pille:
* added travis builds (ESP8266 / ESP32)
* travis explicitly calls those
* reenabled default release builds


pille:
@Aircoookie: any reason for this not being merged?

pille:
the goal of explicitly calling the `travis_*` builds is to deliver a `platformio.ini` that doesn't build "random" targets when people checkout the project.
in case you'd like to test more, let's create more targets ;)

Aircoookie:
Indeed, it might make sense to just select a single environment (maybe `d1_mini` since most frequently used) as default and handle the travis environments in the travis.yml. Still I would like the flexibility of selecting the travis environments in platformio.ini since we might be able to have travis auto-create some "nightly" builds in the future. Will think about that. For now merging this as is, since I like travis building fully featured for testing purposes.

PR #862, opened 2020-04-21
implement DMX proxy mode via MAX485

pille:
output a configurable universe (E1.31 or ArtNet) via MAX485 module that connects its DI to D4 on d1_mini.
as this function competes with DMX output for the same hardware, the universe to proxy acts as a switch.
0       = DMX OUTPUT (of configured WLED LEDs, which is the classic behaviour)
1-63999 = DMX PROXY  (selected universe, ArtNet is limited up to 32768)

## New Features
* Added Proxy Universe setting to forward E1.31 data directly to DMX output
* Configure Proxy Universe value in Sync settings (0–63999 range; 0 = disabled)
* Setting is automatically saved and persists across device restarts


pille:
@Aircoookie: would be nice to see this in 0.10.0. any reason for this not being merged?

softhack007:
@coderabbitai review

PR #863, opened 2020-04-21
E1.31 and ArtNet universes and DMX addresses start at 1

pille:


Aircoookie:
Hi,
thanks for the PR! This is certainly true for E1.31 - you will notice that if you select that in the settings the universe and address will be set to 1 automatically.
For ArtNet, documentation seems to indicate however that 0 is valid for both DMX universes and addresses, and xLights and Jinx both have 0 as default universe value when creating a new device, therefore I would like to keep that possibility :)

pille:
sure, though i can't find a reliable source of ArtNet documentation.

PR #964, opened 2020-06-04
adds support for 9-key remote and intensity/speed repeat actions

kareem613:
Made a small refactor of how the repeat signal was handled for brightness. If I followed the existing patterns, I would have added a third 'or' to a couple of if statements. That pattern would quickly get out of hand with more remote codes to handle. 
Instead I used a similar pattern to how the lastValidCode is tracked. It now also tracks the last valid action that was taken and it's corresponding value. 
Actions are all defined in the ir_codes.h file. Power on/off and changes to brightness, intensity, speed. 
I didn't have to touch much else except for breaking out an explicit changeBrightness function similar to the existing changeEffectxxx functions. 

Pardon new remote support and repeat actions in a single PR. I wasn't planning on doing the repeat actions until I was in too deep already. :)

### DIFF
--- a/wled00/fcn_declare.h
+++ b/wled00/fcn_declare.h
@@ -57,8 +57,10 @@ void onHueData(void* arg, AsyncClient* client, void *data, size_t len);
 
 //ir.cpp
 bool decodeIRCustom(uint32_t code);
+void applyRepeatActions();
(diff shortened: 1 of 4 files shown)


Aircoookie:
Yep, I agree this change makes it easier to maintain. Thank you!

PR #973, opened 2020-06-07
New readme design

WoodyLetsCode:
- added emojis
- changed the first sentence (information about WS2801)
- added images
- added discord banner

I added this, because I thought for a long time that WLED doesn't support WS2801.
I didn't read the wiki. And so I think this could be useful for many other people.


Aircoookie:
I love this! The emojis really help the project look a lot more inviting!
The pictures are also mega cool, just as in a commercial product ad!
Just one question to make sure before I merge - do we have the usage rights to these images? I just don't want to get in trouble by infringing on the photographer's or Apple's copyright or trademarks. 
Also before merging I would scale them down/reduce the JPEG quality a bit to reduce the file size to 1MB. That is good practice because otherwise any readme load would take way to much bandwidth. Hope that's alright for you :)

WoodyLetsCode:
I made these pictures with Mockuuups Studio. They are free to use. You can find informations about the license [here](https://mockuuups.studio/license/).
I also compressed the images. Now they have a small size.

Aircoookie:
Awesome, thanks!

WoodyLetsCode:
Can  you also change the description of your project to something like this: _Control WS2801, WS2812B, WS2811, SK6812, APA102 and much more RGB LEDs with an ESP8266/ESP32 over WiFi!_
I think this is more accurate. 😁

Aircoookie:
Changed it to `Control WS2812B and many more types of digital RGB LEDs with an ESP8266 or ESP32 over WiFi!`. I think we shouldn't list too many, it should be short and to the point :) But you are right that pointing out the compatibility is a great idea! The old description was the same since 2016 haha

WoodyLetsCode:
Nice, thanks!

mike2nl:
@WoodyLetsCode 
very well done.

- one question/idea about the design.
In my 2 cents i would make the discord link/image 50% smaller

WoodyLetsCode:
Okay. I made a pull request for this: #976

PR #992, opened 2020-06-14
Usermods: PIR sensor switch, v2 unreachable net services

gegu:
After I discovered the usermods v2, I wanted to try them right away and I think the possibilities are great - thanks a lot. Especially, placing entries on the info page in the web interface is very cool. 👍  This also allows the implementation of simple user interactions in conjunction with the JSON state values.

This PR contains a usermod to use a PIR sensor with WLED and a v2 usermod for the "fix unreachable net services".

A small wish for a future version: It would be great if state values could also be saved persistently.


Aircoookie:
Awesome, thank you! Yes, I agree that persistant values would be a good idea, it should be possible with the upcoming filesystem support!

PR #995, opened 2020-06-14
Add Digital Clock option to Time options in overlay.cpp

MSeven:
This adds a simple 7 segment digital clock output option to the existing time overlay functionality.

Requires exactly 30 Pixels in a 7,7,2,7,7 setup displaying hh:mm.

The intended wiring direction per digit is g,f,a,b,c,d,e. (while wiring from the reverse!)
7segment

A number of different 3dprintable frame designs are available [on thingiverse](https://www.thingiverse.com/search?q=7+segment)


adamoberley:
What would it take to add functionality to work with this clock? https://www.thingiverse.com/make:789125

The Arduino code is here https://github.com/groemi/DigitalClockSmartShelving

MSeven:
> What would it take to add functionality to work with this clock? https://www.thingiverse.com/make:789125
> 
> The Arduino code is here https://github.com/groemi/DigitalClockSmartShelving

After a quick look i would say, the digit display should work in general (as long as the segments are wired up in the correct order). The only thing is, that the current code assumes one (seven segment digit) segment to be one pixel. You can either hardcode a hack for this in a custom version. The clean solution would be to use (Wled)segments/groups to define physical pixels per segment. (i currently lack the hardware to properly test this).

adamoberley:
> After a quick look i would say, the digit display should work in general (as long as the segments are wired up in the correct order). The only thing is, that the current code assumes one (seven segment digit) segment to be one pixel. You can either hardcode a hack for this in a custom version. The clean solution would be to use (Wled)segments/groups to define physical pixels per segment. (i currently lack the hardware to properly test this).

This was actually my plan to do but WLED currently only supports 10 segments. Do you know how to hardcode more segments with a usermod? If I could create them all as segments, then I could use NodeRed and Home Assistant to control each segment so I can toggle clock mode, have it turn on with motion, etc.

lupus6:
Is this updated in the pre flash bin files?

lupus6:
And to add more led to the segments I just wired them in parallel in each segment.

NeariX67:
Here is my code to make that clock from DIY Machine to work:
https://github.com/nearix67/wled_clock

Aircoookie:
This would be best suited as a usermod now, and I believe there is already at least one that accomplishes 7 segment support :)

Issue #1033, opened 2020-07-08
WLED submits invalid DNS hostnames to DHCP


garyd9:
**Describe the bug**

WLED appears to submit the "Server description" text field as a hostname when registering via DHCP.  The "server description" often will contain characters that are invalid machine names and if the DHCP entry is passed along to DNS, a DNS server can't properly handle the submitted server name.

**To Reproduce**

In a new WLED instance running on a wifi network that uses DHCP and where DHCP forwards names to DNS, enter config, select "User Interface" -> and type a "Server description" that contains spaces.  For example, "Front Porch Lights."  Save.  Go to WiFi Setup and ensure no static IP information is provided (or that it's all 0.0.0.0.)

**Expected behavior**

WLED would submit a hostname with DHCP that meets the requirements of RFC 952 and 1123.   (from 1 to 63 characters long, letters from a to z, the digits from 0 to 9, and the hyphen (-).  A hostname may not start with a hyphen)

One alternative might be to replace invalid characters in the "server description" with hyphens.  Another might be to use a configured mDNS name.  As a fallback, use the MAC address.

**WLED version**

 - Board: esp8266
 - Version: 0.10.0
 - Format:  binary from WLED site

**Additional context**

I validated the issue using Windows Server 2016 DHCP that forwards to Windows Server 2016 DNS (which drops the DNS entry that DHCP attempts to create and logs an error.)

garyd9:
I'd love to propose a fix, but I'm not familiar with what libs are available for use.  It appears that std c++ strings are, however.  I just don't know what version of C++...  So, perhaps in wled.cpp, ~line326 (before the #ifdef ESP8266):

```
// NOTE:  this is off the top of my head and should most certainly be tested if it even compiles.

// mangle the serverDescription to be a valid DNS hostname by replacing invalid chars with hyphens.
String hostname(serverDescription);
for (auto it = hostname.begin(); it != hostname.end(); ++it)
    if (!::isalnum(*it))  *it = '-';   // replace any non alpha-numerics with a hyphen.. yes, a hyphen replaces itself
// ensure the first character isn't a hyphen
if ((0 < hostname.size()) && ('-' == hostname[0]))  hostname.erase(0,1);

```
then, there are two lines of code that follow that called "WiFi.hostname(serverDescription)".  Just change them to pass in hostname instead.

garyd9:
[...]

PR #1035, opened 2020-07-09
Use string derived from serverDescription for wifi.hostname()

garyd9:
The code was sending illegal hostname strings to WiFi.hostname() (which is then submitted to DHCP and often times to DNS.)  A valid hostname contains only alphanumeric characters and hyphens (though it can't start with a hypen.)  This change simply alters the value passed to wifi.hostname() by replacing all non alphanum chars with hyphens while ensuring the first char is never a hyphen.  If the resulting hostname is empty, it uses the escapedMac value (which I'm assuming is initialized by the time this code executes.)

This change would result issue #1033

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -147,7 +147,7 @@ shared_libdeps_dir = ./wled00/src
 framework = arduino
 board_build.flash_mode = dout
 monitor_speed = 115200
-upload_speed = 921600
+upload_speed = 115200
 lib_extra_dirs =
     ${common.shared_libdeps_dir}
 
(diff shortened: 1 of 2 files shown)


garyd9:
Duh.  I'm an idiot.  The "change upload speed default" wasn't supposed to be in the pull request branch.  I'll try to figure out how to get rid of it, but if I don't manage to, please ignore that commit.  I'm sure you have a very good reason to leave the default at 115200.

Aircoookie:
Merged in #1048 :)

PR #1076, opened 2020-08-03
Fix multi-universe ArtNet output

tschundler:
The existing code was very buggy - a 512 channel ArtNet packet sent data to
171 LEDs, not 170. Also the second ArtNet universe was off-by-one.

The math was very confusing with so many branches. The reduces the
amount of code and amount of branching.

Tested with Resolume + my own ArtNet software + 252 LEDs for ArtNet.
Tested with LedFx for e1.31

I suspect further refinement is warranted, arguably the start channel for ArtNet is also "1" (not "0"), at the first byte, while the e1.31 buffer has channel 1 at the second byte due to the start code.


Aircoookie:
Thank you very much for the fix and style improvements, it was much needed as the code duplication in the switch block was getting out of hand.
I just quickly renamed the local `ledCount` variable to `ledsTotal` to avoid confusion with the global `ledCount`.

PR #1082, opened 2020-08-06
Change product name on JSON page

JeffResc:
This PR introduces that the product name on the JSON page is the actual board type and environment instead of displaying "FOSS." I've made it match the "Environment" row of the table found on the info tab. I find this to be a much more accurate description of the device that's running WLED.

Here's how it's displayed on the HomeAsssistant devices menu:
HomeAssistant Screenshot


henrygab (approved):
This looks like a good, simple, and useful fix.

PR #1090, opened 2020-08-09
fix json state variable typo for sunset offset

garyd9:
Only a visual bug, but a bug nevertheless.


garyd9:
folding this fix in another PR with another bug fix.

Issue #1084, opened 2020-08-06
Alternative color picker: Enter RGB(W) decimal or hex code


ezcGman:
**Is your feature request related to a problem? Please describe.**
A perfectionism problem I would say: I use WLED to provide ambient light to my shelf holding my console collection. Every console gets a background light with their companies main color, And e,g, SEGA and PlayStation have a very similar blue. So it is really hard to just pick them with the color picker.

**Describe the solution you'd like**
Instead of just a color picker, it could be an option to switch between the picker and manual RGB(W) code text fields. LIke all image applications for example do.

**Describe alternatives you've considered**
A way to edit the config behind manually maybe?

**Additional context**
N/A

Thank you for your ideas for making WLED better!

huggy-d1:
Not sure manually is what I would describe it, but you can create URLs on another device to select very specific color values.  Perhaps you could iterate over a few URLs until you find the exact colors you want, then store them in presets.
Basically you would be manually color mapping requested color to actual visual color output which relates to room lighting, time of day (if outside light can enter the room), your visual acuity (which is perfectly suited to your needs !), the strip's performance (delivering proper current), and the LEDs used on the strip.
Look at the WLED Wiki for URL requests that can help you find the right colors.

gman-php:
Yah, JSON API is a good idea. I can use that to exactly set the color codes, thx! :) Still, I think it would be really convenient to "just" have three/four text inputs in the UI!

The JSON API now will fix it for me though, thx!

JeffResc:
Hey @gman-php, I just created pull request #1093. If you have a minute, let me know what you think of the addition.

gman-php:
Hey @JeffResc,

that looks awesome, thanks for the fast solution!
I didn't comment on the PR itself, as I wanted to leave that for code/code style/potential bug comments. And comment here if that solves my feature request :)

Pretty cool, thx!
Greetings,

Andy!

blazoncek:
Implemented way back in 2020. Closing.

PR #1093, opened 2020-08-10
Add hex color input

JeffResc:
This PR relates to issue #1084 and adds a hex color input to the main index page of the UI. If the hex color is not supported, a small message will appear saying "Invalid hex color." Otherwise, the new hex color will be set in the color picker and will be set on the LEDs as well.

Edit: I've also added code to disable the web browser's spell check in the text box as you could see in the GIF.

Here's a small GIF of what was added:
GIF of new hex input

### DIFF
--- a/wled00/data/index.htm
+++ b/wled00/data/index.htm
@@ -935,6 +935,8 @@ input[type=number]::-webkit-outer-spin-button {
 				<div class="sliderdisplay"></div>
 			</div><br>
 		</div>
+		<input name="HC" type="text" placeholder="#000000" onkeyup="hcUpdate()" />
+		<p class="labels" style="display: none;" id="invalidHexColor">Invalid hex color</p>
 		<div id="wwrap">
 			<p class="labels">White channel</p>
 			<div class="sliderwrap il">
@@ -2120,6 +2122,22 @@ function mergeDeep(target, ...sources) {
 	}
 	return mergeDeep(target, ...sources);
 }
+
+function hcUpdate() {
+	const regex = /^#([0-9a-f]{6})$/i;
+	if (regex.test(document.getElementsByName("HC")[0].value)) {
+		document.getElementById("invalidHexColor").style.display = "none";
+		cpick.color.set(document.getElementsByName("HC")[0].value);
+		cpick.forceUpdate();
+		setColor(1);
+	} else {
+		document.getElementById("invalidHexColor").style.display = "block";
+	}
+}
+
+cpick.on('color:change', function(color) {
+	document.getElementsByName("HC")[0].value = color.hexString;
+});
 
 size();
 _C.style.setProperty('--n', N);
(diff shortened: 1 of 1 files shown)


wled00/data/index.htm
```diff
@@ -2121,6 +2123,22 @@
 	return mergeDeep(target, ...sources);
 }
 
+function hcUpdate() {
+	const regex = /^#([0-9a-f]{6})$/i;
+	if (regex.test(document.getElementsByName("HC")[0].value)) {
+		document.getElementById("invalidHexColor").style.display = "none";
+		cpick.color.set(document.getElementsByName("HC")[0].value);
```
pbolduc, line 2130: Since I have been working on JavaScript minimization, here are some tips to make this code better. Not sure it will improve the chance this is merged. 

Use the `d` variable instead if `document` to keep JavaScript size down. Also avoid accessing the DOM multiple of times. Save the value of `document.getElementsByName("HC")` and reuse. ie

```javascript
function hcUpdate() {
	const regex = /^#([0-9a-f]{6})$/i;
	const hexcolor = document.getElementsByName("HC");
	if (regex.test(hexcolor[0].value)) {
		...
		cpick.color.set(hexcolor[0].value);

```

JeffResc: Thanks @pbolduc, but I'm not sure if this PR is necessary anymore. I read in the [0.10.2 release notes](https://github.com/Aircoookie/WLED/releases/tag/v0.10.2) that it adds "Option for HEX color entry and RGB sliders." I don't currently have a WLED instance up and running, is anyone able to confirm that this is the same feature that's in this PR? If so, I can go ahead and close this.

pbolduc: Seems you are correct, I found this code but I have not validated on a running WLED instance,

https://github.com/Aircoookie/WLED/blob/503b8cbf349158ecfaf1fa7969e745809a53d142/wled00/data/index.htm#L943

pbolduc: Here is from the latest source:
image
clicking on the icon,
image

Aircoookie:
@JeffResc sorry for not getting back to you yet! Indeed the hex entry field was already added to the UI by me in June, but the option for enabling it was just added yesterday (if you don't count clicking the palette icon which was only meant as debug).
Sorry if I have caused you unneccessary work implementing this (it was in the changelogs for build 2006220 but arguably quite hidden) The spellcheck is a great idea though, we should definitely add it!

@pbolduc slightly off-topic, but what browser are you using? The hex input field should be styled (dark, rounded corners, bigger, white font) and I'm wondering why that isn't working... or is this only when building this PR?

pbolduc:
> @pbolduc slightly off-topic, but what browser are you using? The hex input field should be styled (dark, rounded corners, bigger, white font) and I'm wondering why that isn't working... or is this only when building this PR?

I think that may be a regression in my changes. I will have to double check. I was using the new Edge browser (dev channel) for the previous screen shots, But just pulled the latest master from this repo and it looks like this:

image

Aircoookie:
Alright, that is the way it is supposed to look! Confirmed on my end that it is rendered correctly in Chrome, Firefox and the new chromium-based Edge.

JeffResc:
Since everything seems to check out, I'm going to close this PR. Thanks for your help @pbolduc and @Aircoookie!

PR #1099, opened 2020-08-18
initial word clock matrix usermod

bwente:
Based on code provided by @Aircoookie I added my code to create a word clock using a 8x8 RGB matrix.

https://www.hackster.io/bwente/word-clock-with-just-two-components-073834

image

This a bare bones working clock based on my particular layout. There are some things in there that I tried and it didn't work. I honestly have know idea what I am doing. I have gone as far as I can and would appreciate any help to make it better and of more use to others. I can add svg files for laser cutting and stl fils for printing.


bwente:
This is great! It is official. Now I can report issues and bugs that I don't how to fix! ;)

PR #1100, opened 2020-08-19
Improve TPM2.NET code

mickabrig7:
Add support for receiving TPM2.NET packets out of order, and hopefully prevent crashes if faced with protocol errors.

Not yet tested as I'm currently at work, but I should be able to tonight using my 40x30 matrix (if no one tests it until then) 😄


mickabrig7:
Tested and confirmed working perfectly 👌
https://photos.app.goo.gl/21yNDN6gpLsSG2hVA

Feel free to merge when you feel like it !

PR #1109, opened 2020-08-23
Allow overriding CLKPIN/DATAPIN

xxv:
Allows `CLKPIN` and `DATAPIN` to be overridden on a per-board basis.


fishbone-git:
ideally this should also include adding examples of use in platform_override.ini.example so that use of the examples is consistent in showing what options are configurable at compiletime

PR #1112, opened 2020-08-25
design updates

bwente:
update missed localTime var update
removed brightness schedule, use timed macros instead
add design files for laser or 3d prining


Aircoookie:
Thank you!
I don't understand why this was giving merge conflicts, but I resolved them :)

PR #1133, opened 2020-09-01
Feature/minimize javascript

pbolduc:
Here is the initial PR for review.  Please do not merge yet. I have to review why files like `wled00/html_settings.h` are showing changes when I didn't expect changes to those files. I may need to merge upstream changes.

The main index page is down to 28145 compressed bytes

- From pre v0.10.2 : down from 29325 (delta: 1880)
- From v0.10.2 : down from 28167 (delta: 22)

I should note that I tried to minimize the changes to the javascript so that I did not introduce any bugs. There are additional improvements that can reduce the size further. There are also potentially other webpack magic that can be applied such as removing extra white space and tree shaking to remove dead/unused code in iro or RangeTouch

See: https://github.com/Aircoookie/WLED/issues/1127

One bonus of these changes beyond sizes, is that we can make variables and functions more readable because we dont need to worry about size.  Also, we can add comments and formatting.

**Edit 2020-09-09**: This PR will most likely not be merged until other changes around how the settings pages get their data are complete. There are additional development life cycle enhancements that need to be in place to avoid friction.
- wait for the `GetV()` changes to be implemented so that the settings pages can be compressed and served the same way as index.htm
- extend the processing to all of the `.htm` pages
- enable `webpack-dev-server` to test all of the pages together
- add watch and hot module reloading (HMR) to auto re-build and refresh browser as changes are made to the source
- update tools/cdata.js to compress all processed htm files


Aircoookie:
Thank you, great work! Will not merge yet as wished.

About the the changes to `html_settings.h`, the `cdata.js` NPM script does not seem to minify/remove tabs from javascript that is in front of the `GetV()` function. For the release I removed the indentations manually but they are of course reintroduced when running the build script. I'm considering moving the JS in `settings_ui.h` to the bottom of the body as done in the index as a workaround.

pbolduc:
> About the the changes to `html_settings.h`, the `cdata.js` NPM script does not seem to minify/remove tabs from javascript that is in front of the `GetV()` function. For the release I removed the indentations manually but they are of course reintroduced when running the build script. I'm considering moving the JS in `settings_ui.h` to the bottom of the body as done in the index as a workaround.

That explains it. I have updated my PR to revert the changes to html_settings.h and we should be good to merge.  I will update my issue with things that are done after work (I am in Pacific time zone)

pbolduc:
This change introduces a regression on the hex color entry. I will need to resolve before this can be merged.

image

pbolduc:
Determined the reason for the regression is that on the input field, the `type="text"` was dropped in my transform. Therefore, the associated style was not applied. I will review the webpack plugins do see if there is a reason it got dropped on the transform. The webpack plugins should not remove any attributes on the html markup. It should only remove extra white space.

pbolduc:
> Determined the reason for the regression is that on the input field, the `type="text"` was dropped in my transform. 

The `html-webpack-plugin` has configuration setting `removeRedundantAttributes` which is true. Explicit setting of this to false resolved the issue.

pbolduc:
Only thing I have noticed is the page doesnt load and format on IE 11.  The issue is that it can't find the onLoad function.  IE 11 is such an old dog and even Microsoft is stop supporting it on Office 365 and Microsoft Teams.  I tested on Chrome Edge (new Edge), Chrome and Firefox.

Please review and provide any feedback or issues.

Aircoookie:
Just reviewed your PR and unfortunately it doesn't seem to work for me. Running the NPM script after making changes to `index.htm` or `index.js` does not update the `html_ui.h` file. The usual 
```
Inlined 86876 characters
Compressed 29248 bytes
Writing wled00/html_ui.h
```
at the end of the NPM output is also missing.
The log is the following:
```
> node tools/cdata.js

Reading wled00/data/dist/index.html
Reading wled00/data/style.css as PAGE_settingsCss
Reading wled00/data/settings.htm as PAGE_settings
Reading wled00/data/settings_wifi.htm as PAGE_settings_wifi
Reading wled00/data/settings_leds.htm as PAGE_settings_leds
Reading wled00/data/settings_dmx.htm as PAGE_settings_dmx
Reading wled00/data/settings_ui.htm as PAGE_settings_ui
Reading wled00/data/settings_sync.htm as PAGE_settings_sync
Reading wled00/data/settings_time.htm as PAGE_settings_time
Reading wled00/data/settings_sec.htm as PAGE_settings_sec
Writing 34384 characters into wled00/html_settings.h
Reading wled00/data/usermod.htm as PAGE_usermod
Reading wled00/data/msg.htm as PAGE_msg
Reading wled00/data/dmxmap.htm as PAGE_dmxmap
Reading wled00/data/update.htm as PAGE_update
Reading wled00/data/welcome.htm as PAGE_welcome
Reading wled00/data/liveview.htm as PAGE_liveview
Reading wled00/data/favicon.ico as favicon
Writing 8839 characters into wled00/html_other.h
```

It seems to want to read `index.html` from a non-existant `dist` subfolder. I tried renaming the index from `.htm` to `.html` and moving it to said folder alongside with the `index.js`, which results in the script compressing the HTML (+CSS), but not recognizing the JS file...

Aircoookie:
Another question I have is this portion of the new `index.js` file:
```javascript
import iro from '@jaames/iro';
import RangeTouch from "rangetouch";
```

Where does it get these from? Does it download the latest master of these libraries and then embeds them into the final C array or are the libraries in a different file somewhere?
Thank you for the PR and sorry that it is not working right away for me!

Aircoookie:
Found two more issues :(
First one is minor, the horizontal spacing of the quick color selectors is too low... Maybe something got stripped there:
master:

minimize-javascript:


Second, which is more problematic, the segment controls don't work at all anymore. Likely just a matter of adding more of the `window.xxx = xxx` linkages though:
```
(index):1 Uncaught ReferenceError: makeSeg is not defined
(index):1 Uncaught ReferenceError: selSeg is not defined
(index):1 Uncaught ReferenceError: expand is not defined
```

pbolduc:
In the data folder I added a README.md file (in caps only to match file name and not for emphasis). 

```
npm install
npm run webpack
npm run build
```

The webpack one creates the file in dist folder which is not checked in.

In the future these can combine into single step/task.

I will look at issues you reported above. I appreciate your patience. I know how unsatisfying it is to receive people's PR and having to spend time test and troubleshooting.

I think I should have created some automated UI tests first.

PS I also noticed I may have introduced a JavaScript error around access to 'pwr' or something.

I am going to rollback almost all JavaScript 'optimizations' and ensure all required functions are exported and do more testing.

The imports of iro and rangetouch at the top pull from the installed npm packages. The babel transpiler pulls them in and inlines into the JavaScript.

Aircoookie:
No worries, thank you so much for contributing and that errors happen with such a large structural change is inevitable :)

pbolduc:
> the horizontal spacing of the quick color selectors is too low... 

This is really odd, it comes down to the white space in the markup

This looks normal:
image
image

But when the whites space between the divs is removed,
image
image

I still have to compare the styles in the dev tools, but cant explain it yet.

pbolduc:
Firefox devtools shows the reason, the #text white space takes up room, 4.71667 pixels to be exact.. Maybe I have to review the style sheet for hints

image

Aircoookie:
Indeed strange behavior, I wouldn't have expected the browser to insert a whitespace on newline.
Rather than disabling the optimizations to keep the whitespace, as unconventional as it seems, putting the `</div>` closing tag in the next line immediately in front of the next div so that the linebreak is inside the div will get rid of the spaces ([reference](https://stackoverflow.com/questions/1097006/)) 
Then just changing the margin on the `.qcs` to `2px 4.5px` restores the proper spacing :)

pbolduc:
You could do that.  From what I found the behavior because of the `display: inline-block;` styling.  There is a minimizer option that resolves this. 

> Remove spaces between tags is not safe when you use display: inline-block

I found if I use the recommendation in this issue https://github.com/webpack-contrib/html-loader/issues/166 and add `conservativeCollapse: false` we dont get this behavior. 

You can mess with the styles, but we could always be chasing our tails.  I have a few more changes to clean up tonight and we should be good to go.

Aircoookie:
Hey, today I finally had time to review the new changes!
I can confirm that the issues due to the missing references as well as the spacing are now resolved!
Also happy to see that the webpack command is integrated in the `npm run build` command now.

There is just a single nitpick I still have - the UI has a simple local mode where you can test as you make changes just by providing a WLED instance IP and refreshing the browser tab after each change. This still works perfectly with the combined `dist/index.html` file, but of course you'd need to execute the `npm run webpack` after each change for that to update.
It would be perfect if either  
- `index.htm` would properly import `index.js` even when not packed (best case, requires no npm for local development, but I guess it would not be that easy due to the JS dependencies now being included by the webpack)
- there would be a dev webpack command that detects file changes to `index.htm` and `index.js` and run the webpack in that case

I still have a lot to learn regarding tools like NPM - I love it because it allows for powerful scripting and automatic composing - but I am frankly not sure if I could add that functionality myself😄

pbolduc:
Going to close this for now. I will still have the changes in my repo. It is pointless to keep it open now cluttering up things everywhere. If I find the motification to get this to somewhere close to where it can be merged, I will open another PR.

PR #1136, opened 2020-09-02
Change default access point name

JeffResc:
This proposed pull request changes the default "WLED-AP" access point name to "WLED-xxxxxx" where "xxxxxx" is the last four characters of the device's MAC address. This will allow users to differentiate their devices before setting them up and this change stops multiple devices from broadcasting on the same SSID, which may be a problem if you have multiple instances.

### DIFF
--- a/wled00/set.cpp
+++ b/wled00/set.cpp
@@ -270,7 +270,12 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
     if (request->hasArg("RS")) //complete factory reset
     {
       clearEEPROM();
-      serveMessage(request, 200, "All Settings erased.", "Connect to WLED-AP to setup again",255);
+      escapedMac = WiFi.macAddress();
+      escapedMac.replace(":", "");
+      escapedMac.toLowerCase();
+      strcpy(apSSID, "WLED-");
+      sprintf(apSSID + 5, "%*s", 6, escapedMac.c_str() + 6);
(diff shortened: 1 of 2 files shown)


Aircoookie:
Thank you!  
Indeed making the SSID unique would solve the issue with multiple devices broadcasting on the same SSID, which does indeed happen (for example if the router does not boot in time after a power outage), so it would be very useful for that!  
There are two related drawbacks though, if setting up multiple devices, you'll be required to re-enter the password for each of the instances. The second problem is that not knowing the SSID ahead of time makes it a lot harder to, for example, generate a QR code for quick connection or auto-connecting to the AP programmatically (which would be required for an auto-setup process in the future, since Android and iOS actually don't allow you to access the list of nearby WiFis anymore, just provide one to connect to).
I believe I will merge this with a compile flag to disable it in case a future use case requires all devices to have the same SSID.

pille:
:+1: for compile-time-flag. i like the approach of initially configuring new devices in a semi-automatic way, while i can understand the idea to avoid confusion.

JeffResc:
So for the auto-connect/quick-connect QR code, would it be possible to have a second SSID be broadcasted as hidden under `WLED-AP`: `wled1234` SSID and password? That way you could still connect to the original AP as intended, but quick-connect will still work on that second SSID. I know that WiFi adapters on computers and routers are capable of this, but I've never tried it with an ESP8266 or ESP32.

PR #1156, opened 2020-09-07
enable sliders with a different max value

bigbear3001:
this enable reading the max value from the max attribute of the sliders and enables sliders with max values other than 255


Aircoookie:
Thank you, this will make it easier to add sliders with other ranges in the future!
Will merge after #1133 :)

PR #1163, opened 2020-09-08
Firefox Custom Scrollbar

JeffResc:
Firefox supports [limited scrollbar customization](https://drafts.csswg.org/css-scrollbars-1/), but currently, WLED does not support them as the CSS differs between Chrome and Firefox. I've added the code to enable the limited customization that Firefox supports, which is not all that Chrome does, but the new scrollbar definitely looks better than the default.

**Before**:
Old Scrollbar

**After**:
New Scrollbar

### DIFF
--- a/wled00/data/index.htm
+++ b/wled00/data/index.htm
@@ -60,6 +60,8 @@ body {
 				-ms-user-select: none;
 						user-select: none;
 	-webkit-tap-highlight-color: transparent;
+  scrollbar-width: 6px;
+  scrollbar-color: var(--c-sb) transparent;
 }
 
 html,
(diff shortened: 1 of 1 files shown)


trinityEon:
This is nice. I like the scrollbar for Firefox. Have you noticed that on android chrome when you are in the effects tab then scroll all the way to the bottom, you can't scroll back to the top because it conflicts with scroll to refresh feature in chrome

Aircoookie:
This is cool! Awesome that there is a way to make the scrollbar a bit nicer on Firefox too (to be honest some way should be CSS standard so it works in all browsers)

Currently waiting on merging #1133 before making any changes to the `index.htm`, but then I will merge this :)

PR #1164, opened 2020-09-09
Akemi on the welcome screen

JeffResc:
Hello again!

This pull request proposes to replace the smiley face on the welcome screen with the new cheerful Akemi mascot. At first, I was concerned about storage since it was an image, but it is not actually much larger than the current image anyway. I provided my testing results below along with screenshots of the UI before and after.

|                | Flash Before  | Flash After   | Flash Change        |
|----------------|---------------|---------------|---------------------|
| travis_esp8266 | 704692 bytes  | 704868 bytes  | +176 bytes (+0.02%) |
| travis_esp32   | 1222562 bytes | 1222718 bytes | +156 bytes (+0.01%) |

**Before**
Before

**After**
After


Aircoookie:
Made a little change by using the base64 of the 32x32 sprite together with the `pixelated` and `crisp-edges` image rendering css attributes and that actually brought the size down by an entire kilobyte, wouldn't have believed that!

|                | Flash Before  | Flash After   | Flash Change        |
|----------------|---------------|---------------|---------------------|
| travis_esp8266 | 704692 bytes  | 703668 bytes  | -1024 bytes (-0.14%) |

Also #1167 was just submitted and now I don't know which one to go for as I love both 😅

Aircoookie:
Hey, as you might have seen I had a bit of a decisive crisis which artwork to use and even the community was rather split, with 51 voting for the logo and 46 for Akemi. Even though I would absolutely love to have Akemi on the welcome page since she is very fitting, and despite the logo using quite a bit more flash (704916 bytes on travis_esp8266), I believe it represents WLED better especially for first-time users who might have gotten WLED from other sources like pre-flashed boards (or, god-forbid, viralleds) and despite having more pixels, the appearance is still very good on mobile devices. We can also move to Akemi at any point if we find a better place for the logo :)

JeffResc:
Sounds good, thank you for considering my PR 👍

Aircoookie:
Actually I'm merging your PR (because I wanted to stick to the base64 format for the imagen), just added a commit to change to the logo :)

pfeerick:
I think it's quite appropriate Akemi be used for the welcome screen, since the initial colour used in a new install is orange. :)

Aircoookie:
Me too! I actually tried putting both Akemi and the logo but that looked too noisy :)

PR #1165, opened 2020-09-09
New WLED logo on info page

JeffResc:
Hello again!

This pull request proposes to replace the old WLED on the info screen with the new one. I provided screenshots of the UI before and after below.

**Before**
Before

**After**
After


trinityEon:
Why? This new version is too busy and distracting with all the colors and more unreadable for the interface. To be honest the other one looks a lot cleaner

WoodyLetsCode:
Hey!
I like it! Can you use this file and than upscale it?
This uses less space.
[wled_akemi_original.zip](https://github.com/Aircoookie/WLED/files/5195998/wled_akemi_original.zip)

pbolduc:
It would be great if the logo was available in SVG, it could possibly make the file smaller. Can we ask the original designer.?

WoodyLetsCode:
I don't think so. I think PNG is the best file format for pixel art. This file format works best with images with areas of solid color.
In SVG you only have rectangles and rendering vectors is quite expensive.

JeffResc:
@WoodyLetsCode Thanks, I've added that and it looks like it saved a lot of space. By chance do you have the compressed image of the Cheerful Akemi for #1164?

Aircoookie:
@JeffResc I agree, PNG is better suited for pixel art! Currently working on #1164, managed to get the size down by 400byte more actually by using a 32x32 version! Will commit that in a moment!

Aircoookie:
Although I absolutely prefer my new design over the old blue one with tags (that was on the readme), it is made to be more of an eye-catching splash art and might distract from the data displayed on the Info page, which the monochrome design does not. So in my opinion @trinityEon 's point is absolutely valid and I still have to think about it but I would tend to say that I like the monochrome version a tad more in places where other important things are displayed on screen, like the info page or the top bar of the WLED mobile app.

pbolduc:
> I don't think so. I think PNG is the best file format for pixel art. This file format works best with images with areas of solid color.
> In SVG you only have rectangles and rendering vectors is quite expensive.

I am not convinced that SVG only has rectangles (https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/alphachannel.svg). Rendering vectors is not done on the micro-controller but rather on the user's browser.

But I am fine with your statement that PNG is better for this case.

PR #1167, opened 2020-09-09
New welcome page

WoodyLetsCode:
Hello,
i updated the welcome page. I replaced the old smiley new with the new WLED Logo.
Also I added some animations and hover effects for the buttons.
Here is a little preview:
grafik
If you want to see the animations you can have a preview here: [welcome.zip](https://github.com/Aircoookie/WLED/files/5195928/welcome.zip)
By the way, i changed the code formatting of welcome.htm. If this is problem for you, just write me back. :smile:


Aircoookie:
Thank you, it looks great!
Now I have to decide between your PR and #1164 😅 Need to think a bit about which one to go with now, since I love both designs...

WoodyLetsCode:
Yes, thats quite funny that @JeffResc and me had the same idea. Actually i wanted to make this PR yesterday, but than there was a bug with the animations in chrome... :sweat_smile: 
It's your decision, wich you like more. :smile: If you want i can change the image to another Akemi.

Aircoookie:
Hey, as you know even the community was quite split on which artwork to use (51/46 discord vote).
Now I decided to ultimately settle on the logo for now simply for the better recognizability of the software!
I am going to merge #1164 because I prefer the image in base64 for now (ESP8266 sometimes doesn't like multiple successive requests for page resources), but I took inspiration from your idea of adding animations, so I added a simple fade-in :)
I hope you like the new page, personally I think it's miles better than the old smiley!

PR #1176, opened 2020-09-12
Clean up index.htm

WoodyLetsCode:
Hello,
approx. two month ago i wrote you about the idea to split the index.htm file into two other files index.css and index.js.
Because it isn't very easy to work in the index file anymore, spliting it up would it make more clear and easier to edit the file.
This would be very helpful for future updates.
I created two new files: index.js and index.css
**index.js:** contains all the javascript
**index.css:** contains all the css code
**indexhtm:** only contains html
I hope you like it. 😄


WoodyLetsCode:
Hey @Aircoookie ,
are you interested in this at all? Because i haven't heard anything from you. If you are interested in it and like it I will resolve the merge conflicts. Otherwise please tell me what you don't like. 😄

Aircoookie:
Hi!
Sorry for not getting back to you!
Splitting the CSS, Javascript and HTML to separate files is an idea I really like as it would make developing the UI more accessible. The problem is that your code serves the css and js in separate requests. That is standard practice on larger servers and due to caching, could even reduce loading times/bandwidth. ESPs however can occasionally crash if they receive multiple requests in short succession, so that should be avoided in our special case.
There is another PR, #1133 similar to yours, but it splits the javascript to a different file and then merges them for the gzipped version that is added to `html_ui.h`. It got a bit complicated though with an additional `npm webpack` command, and local testing didn't work, since the browser couldn't find the .js file - that's why that is not yet merged either.

Also, I made many (not yet published) additions to the `index.htm` for the upcoming v0.11 - so please don't bother fixing the conflicts for now. Your work was not in vain though - I believe combining it with #1133 could make adjustments to the UI easier in the long term 😄

WoodyLetsCode:
Hi!
Thank you for getting back to me!
I didn't  know that there are currently some open PR about the same idea.
Maybe next time i should check the other PR before makeing an PR. 😄 
But now its happy to know that sb ist working on it.

pbolduc:
I just closed my PR because I wasn't working on it anymore and as Aircoookie stated it got complicated for non-web devs trying to use webpack. One area I was thinking about using HTTP caching on some of these static requests. On serving the js, css or any other static page, we can send a precomputed ETag representing the hash of the file. The etag hash could be computed when WLED is compiled. Similar how to git does it, the first 6 characters of the sha1 hash should be good enough. I may have to run through some things to think this through.  If the browser can cache these requests, then the server just needs to return a small 304 Not Modified response.

WoodyLetsCode:
I think I will also close this PR, because it got outdated. But I think I will make a new PR, that will not split the files for esp. I have an idea how to realize this, but i haven't tested it yet. If everything works well i will make an PR.
Or are there currently any other projects running that are trying to split the html files?

gsieben:
I currently had to make a few changes to the index.htm for my usermod (additional relays). Index.htm does not support usermods. If one ever wants to work on a splitt, one should think about it. I am so reluctant to interfere with the main code when I only want to write an extension. This could be considered for such a splitt - maybe in the future.

WoodyLetsCode:
So you are looking for an easier way to implement to add own functionality in web ui?
Do I understand this right?

gsieben:
Yes absolutely. 

 

I have implemented an additional switch at the bottom and an additional "main column". Also I need a refresh for the status button. The switches can also be controlled via MQTT/Alexa and HTML. In the end I have implemented an additional page, with which one can control further relays.  For the refresh of the button status I am looking for the best solution without interfering too much in the main code. 

 

You will see what I mean when I will provide my code.

PR #1178, opened 2020-09-12
Async temperature usermod

pbolduc:
This PR improves the DallasTemperature usermod. The improvements include:

- Use non-blocking read of sensor. Previous version would block for ~94 ms on each time the temperature is read. See [WaitForConversion](https://github.com/milesburton/Arduino-Temperature-Control-Library/blob/master/examples/WaitForConversion/WaitForConversion.ino) example.
- does not report `0 Sensor Error` before the first temperature is read. Now will display the number of seconds until the first sensor read.
- if sensor is not found during setup, the usermod is disabled. It will not report Temperature in the info and will not continue to try to read the sensor.
- will not report super low temperature that is beyond the senor's specs. The DallasTemperature  library will return really low values for temperature to indicate an error
- add example platformio_override.ini
- allow customization of the temperature read interval and number of seconds after boot to read sensor via compile time #define 
- changed most preprocessor directives to start with `USERMOD_DALLASTEMPERATURE_` to avoid conflicts with other mods or WLED source.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -227,24 +227,6 @@ upload_speed = 921600
 board_build.ldscript = ${common.ldscript_4m1m}
 build_flags = ${common.build_flags_esp8266} 
 
-; move to platformio_override.ini before PR
-; lib_deps were copied from above, but should use a variable
-[env:d1_mini_usermod_dallas_temperature_C]
-board = d1_mini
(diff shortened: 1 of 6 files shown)


Aircoookie:
Tested and works very well! This increases the quality of the mod immensely. Thank you so much!

PR #1184, opened 2020-09-15
ESP8266: Allow auto modem sleep if power off and sleep permitted

### LINKED ISSUE #1144: Bug: Missing delay to activate sleep mode on ESP8266
According to the ESP8266 docs you need delay(1) or more to enter DTIM-based wifi sleep (light sleep). Uploading the current version to a d1 mini without connecting anything else, the power consumption is very constant at 90mA, even after the 10s sleep timeout (turning LEDs off in web gui doesn't change that).

However if I change wled.cpp in line 89 to:
if(offMode)delay(1);else strip.service();

and then turn off the LEDs in the web gui the power consumption jumps between 35mA and 55mA (while the behaviour with LEDs on is unchanged)

I didn't make a pull request because I have no idea what timing limitations I broke with that delay statement

pfeerick:
Closes #1144. 

Adds a 1ms delay _only_ when main power is off and the `noWifiSleep` option is not active, and _only_ for the ESP8266. This is needed for the automatic modem sleep to operate on the ESP8266 between DTIM beacons. The ESP32 appears to have a different mechanism, via the `esp_wifi_set_ps()` API call, but that a problem for another issue/PR. 

I'm leaving this as a draft for now, until I've been able to test it for a few days for unforeseen side effects. ;)

To show the effect of this on my Wemos D1 Mini, I powered it via USB and put a calibrated power meter on it. The LED Strip and relay is still being powered via the main PSU. 

[Pre PR video - ~150ma continous](https://drive.google.com/file/d/1qxgkezETsACZFKeU4QbLtIcT25076euF/view?usp=sharing)

[Post PR - rolling 90-120ma](https://drive.google.com/file/d/1r2apWgB8K0hsdP80Ym6kVgxmejDOsFzI/view?usp=sharing)

[Post PR - rolling 80-140ma - second meter](https://drive.google.com/file/d/1r14A52LPN_nJZGMMwX72nC6COJwpHkWP/view?usp=sharing)

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -84,10 +84,12 @@ void WLED::loop()
 
     handleHue();
     handleBlynk();
-
     yield();
+
     if (!offMode)
       strip.service();
+    else if (offMode && !noWifiSleep)
+      delay(1);
   }
   yield();
 #ifdef ESP8266
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -89,8 +89,8 @@ void WLED::loop()
     if (!offMode)
       strip.service();
 #ifdef ESP8266
-    else if (offMode && !noWifiSleep)
-      delay(1);
+    else if (!noWifiSleep)
+      delay(1); //required to make sure ESP enters modem sleep (see #1184)
 #endif
   }
   yield();
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -88,8 +88,10 @@ void WLED::loop()
 
     if (!offMode)
       strip.service();
+#ifdef ESP8266
     else if (offMode && !noWifiSleep)
       delay(1);
+#endif
   }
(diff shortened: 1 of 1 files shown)


Aircoookie:
Thank you :) Let me know once you believe it is ready for review!

pfeerick:
I think it's time to call it cooked. Just short of 5 days uptime with zero stability issues. Driving a 507 led strip in four segments, with static and animated effects, this seems to be mostly working - sometimes it seems to still get hung up in 'full power' mode, but if left alone seems to revert to sleepy modem mode. 

image

I'll look as spinning up a ESP32 and see if I can see what's happening over there re: modem sleep, etc. I'll probably poke around the guts of WLED a bit more and see what's keep the ESP8266s attention, but I suspect it's the amount of network traffic / something on the network waking it up, not WLEDs fault.

PR #1185, opened 2020-09-15
Loxone support added

m0fa:
Hello,

this pull request adds loxone support. It was implemented with the help of the Loxforum ([https://www.loxforum.com/forum/german/software-konfiguration-programm-und-visualisierung/260948-wled-projekt](https://www.loxforum.com/forum/german/software-konfiguration-programm-und-visualisierung/260948-wled-projekt)). 

Loxone is a home automation solution that uses self-coded commands for RGB, brightness and color temperature. To make the connection of WLED to Loxone as easy as possible, WLED was adapted and extended. Thus the Loxone commands are interpreted and applied.

A separate UDP socket was added. This allows a faster transfer than HTTP because the overhead is minimized. In the web interface the socket can be activated and the port can be chosen. The settings are stored in the EEPROM. The UDP socket uses the HTTP-Api. 

Loxone offers two commands. One for RGB values and one for brightness and color temperature. For these commands a parser was added to interpret them. The parser was added to the HTTP-Api and the JSON-Api.
For this the parameter `LX` is used. The values are applied to the primary color. `LY` uses the same implementation for the secondary color. The HTTP-Api expects `&LX` and `&LY` in the request. The JSON api expects `{"seg":{"lx":<v>}}`.

Loxone RGB values have the structure `BBBGGGRRR`. Each color is specified in the range from 0 to 100%. For example, "100" stands for red (#ff0000).
Also a conversion of color temperature and brightness to RGB was added.  This command has the structure `20bbbtttt`. "bbb" stands for the brightness in the range 0-100%. "tttt" realizes the color temperature in the range 2700 to 6500 Kelvin. The original implementation is from: [https://github.com/somebuddy87/FASTLED_Loxone_Edition](https://github.com/somebuddy87/FASTLED_Loxone_Edition).


platformio.ini
```diff
@@ -366,3 +366,45 @@ build_flags = ${common.build_flags_esp8266} ${common.debug_flags} ${common.build
...
 build_flags = ${common.build_flags_esp32} ${common.debug_flags} ${common.build_flags_all_features}
+
+# ------------------------------------------------------------------------------
+# codm pixel controller board configurations
+# ------------------------------------------------------------------------------
+
+[env:codm-controller-0.4]
+board = esp_wroom_02
+upload_port = /dev/cu.SLAB_USBtoUART
+monitor_port = /dev/cu.SLAB_USBtoUART
```
codmpm, line 377: Should be removed...

Aircoookie:
Hi and thank you for the pull request!

I have a few points to address before merging:
- The API UDP socket: I believe this is a great idea! While HTTP doesn't have that much of an overhead in my testing at least, for sending commands in quick succession this is a very good alternative. Is Loxone limited to port 7000 or is that selectable? If so, I would propose adding the API functionality to the existing notifierUdp socket - that would decrease the memory usage a bit.

- The color formats used by Loxone appear rather arbitrary - I do not believe these fields would be used by any other software using the WLED API. If I understood correctly, Loxone doesn't have the means/free configurability in order to support the default 0-255 range color channels and I also find it rather strange that the white temperature command supports brightness, but the RGB one doesn't. I am a bit hesistant to add redundant functionality to the APIs - here multiple means to set an RGB color - so I personally believe the parser would best fit as a usermod, but I also get that you would like the default build of WLED to work with Loxone out of the box without requiring a custom compilation. We could also add a compile flag, e.g. `#define WLED_ENABLE_LOXONE` and enable that only by default for the custom boards you added. What do you think?

- The white color temperature to RGB conversion is something I wanted to add to WLED for a long time - looks awesome, very happy about that! If you don't mind, I will also add a way to set the temperature in either in kelvin or in CT/mireds, the system Philips uses with their hue products.

Thanks again!

codmpm:
Thanks @Aircoookie, great to get your feedback. Much appreciated. 
I tried to reach you on the wled discord (@prego), maybe you can answer there as well.

The current UI for the UDP socket is located under "Sync Interfaces" and has the port configurable. 
Bildschirmfoto 2020-09-22 um 15 48 29

I totally agree with you regarding the Loxone format for color information. We did not know about usermods, though.
We also stumbled about the non existing brightness information regarding RGB. They include it in the UI but only output the RGB 0-100% values.

Bildschirmfoto 2020-09-22 um 15 53 40 Kopie
(You can change the color picker to WW/CW on the lower right, which then gives you - when configured - the color temperature and brightness)

Regarding the color Temperature to RGB conversion we are working on a better solution. This one currently is just taken from the FastLED fork to be able to "understand" these values.

@m0fa can give deeper info on the implementation 👍

m0fa:
Hello,

we can add the API functionality to the notifierUdp socket. It would be useful to activate the API in the web interface. Or should it be permanently active?

I don't like the redundant functionality either. Unfortunately, Loxone has its own color formats and the conversion to RGB has to be implemented.
We'd like to have an easy way to update the system. The Loxone functionality should be retained. A custom compilation would be bad. A compile flag for the custom boards is better. The users will mainly update via the web interface. What do you think?

Aircoookie:
Thank you for the feedback!
In that case, going with a compile flag is likely the best option! I don't believe the parser will use a lot of resources, so I'd be open to enable it by default.

Awesome, let's go with notifierUdp then :) It also handles the wled sync modes (first byte 0), some realtime modes (first byte 1-5) and TPM2.net (first byte 0x9c/156). These do not include any ascii characters, so we can safely add the API functionality in the bottom of `handleNotifications()`. Having it always enabled is best, that way we reduce the number of settings :)

Do you want to make these changes or should I update the PR before merging?

m0fa:
Hey and thank you. That sounds great!

If you like, you can update the pull request. Otherwise I can make the changes too. What do you prefer?

Aircoookie:
Awesome! Alright, I will make the adjustments and merge soon :)

Aircoookie:
Changes are done! Unfortunately I have no write access to your fork and thus this PR :(
@m0fa There should be a little checkbox titled "Allow edits from maintainers" in the right toolbar next to the PR. It would be awesome if you can enable it!  
Otherwise I would have to make a new PR which was quite difficult to do while retaining all commits made last time I had to do it 😅

Aircoookie:
Merged and working well :)

m0fa:
Great. Thank you very much!

PR #1208, opened 2020-09-27
Added washing machine FX

Phunkafizer:


Aircoookie:
Thank you! This is a cool effect :)

PR #1215, opened 2020-09-29
Enable explicit sunrise or sunset

### LINKED ISSUE #1214: SUNRISE -- Cannot enable via single HTTP API / Macro
## **Describe the bug**

It's desired to have a single macro / HTTP API call for each of the following:
* enable the sunrise nightlight effect
* enable the sunset nightlight effect

## **To Reproduce**

1. Manually set the power off
2. Use HTTP API with macro named 'Sunrise / Sunset' (_see below_)
3. Verify sun rises over the next one minute (_good!_)
4. Manually set the power on
5. Use HTTP API with macro named 'Sunrise / Sunset'
6. Verify sun sets over the next one minute (_good!_)
7. _Now, try to get force a sunrise, even if the power is on, with single API call:_
8. Manually set the power on
9. Use HTTP API with macro named 'Force Sunrise'
10. Notice that the LEDs turn off ... and then stay off (sunrise fails to work)

## **Expected behavior**

A single HTTP API call should be available that will deterministically select SUNRISE or SUNSET exists.
SUNRISE/SUNSET should not be forced to depend on the current power state of the LEDs.

## **Root Cause**

The code parsing the HTTP API...
1. runs linearly, not based on order parameters are sent.
2. parsing of nightlight parameters `NL=` occurs before parsing of master power `&T=`
3. parsing of master power `&T=` explicitly sets `nightlightActive = false`

Therefore, it's impossible, in a single HTTP API call / single macro, to:
* both ensure the LED power is OFF and enable nightlight for sunrise
* both ensure the LED power is ON and enable nightlight for sunset

## **WLED version**
 - Board: Wemos D1 (any)
 - Version: 0.10.2 "Fumikiri"
 - Format: Official released binary from Github

## **Additional context**

The HTTP API allows enabling a sunrise/sunset nightlight effect using, as an example, the following parameters:
* `&NL=1` to set one-minute total sunrise time (for testing purposes)
* `&NT=255` to set target nightlight brightness to 255
* `&NF=3` to set nightlight mode 3 (sunrise/sunset)

The selection of whether the nightlight will go to sunrise or sunset mode depends on the **_current power setting_**.  Thus, it would make sense that, by also including the master power setting, sunrise or sunset could be effectively selected by setting that as the first parameter:
* `T=0` to set power off should be the very first parameter.

## **Macro List**
Name | Parameters 
-----|-----
Sunrise/Sunset | `&NL=1&NT=255&NF=3`
Power Off | `&T=0`
Power On | `&T=1`
Force Sunrise | `&T=0&NL=1&NT=255&NF=3`
Force Sunset | `&T=1&NL=1&NT=255&NF=3`

henrygab:
Adds explicit nightlight setting for sunrise and explicit setting for sunset.

This allows nightlight function to deterministically start sunrise or sunset (e.g., from HTTP or JSON API).

Fixes #1214

### DIFF
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -109,6 +109,8 @@
 #define NL_MODE_FADE              1            //Fade to target brightness gradually
 #define NL_MODE_COLORFADE         2            //Fade to target brightness and secondary color gradually
 #define NL_MODE_SUN               3            //Sunrise/sunset. Target brightness is set immediately, then Sunrise effect is started. Max 60 min.
+#define NL_MODE_SUNRISE           4            //Explicit choide of sunrise, else identical to NL_MODE_SUN
+#define NL_MODE_SUNSET            5            //Explicit choide of sunset,  else identical to NL_MODE_SUN
 
 //EEPROM size
 #define EEPSIZE 2560  //Maximum is 4096
--- a/wled00/data/settings_leds.htm
+++ b/wled00/data/settings_leds.htm
@@ -169,7 +169,9 @@
 			<option value="0">Wait and set</option>
 			<option value="1">Fade</option>
 			<option value="2">Fade Color</option>
-			<option value="3">Sunrise</option>
+			<option value="3">Sunrise/Sunset</option>
+			<option value="4">Sunrise</option>
+			<option value="5">Sunset</option>
 		</select>
 		<h3>Advanced</h3>
 		Palette blending:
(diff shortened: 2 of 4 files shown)


henrygab:
Got it compiling using VSCode + PlatformIO extension.  Wow, the dependency handling is so much cleaner than Arduino.  Anyways, I will now be able to test the change.  Would still appreciate early code review, as WLED has some complexity.

henrygab:
Resolved via an alternate method by 3107943af81ae5b64267e83b36726d2e2e397f88.

Issue #1210, opened 2020-09-27
Control ESP01S builtin LED


sasukebinbin:
Hi,

I like this so much after compare with ESPHome Fastled. 

I use it with ESP01S and implement in Home assistant. 

My problem is no way to control the builtin LED on ESP01S.

It is always on. I hope there is an extra button I can use to turn on or off the builtin LED.  And it will be recognised in Home assistant as an entity. 

Thanks
Rex

Aircoookie:
Hi,
thank you!
Unfortunately the builtin LED uses the same pin (GPIO 2) as the LED data and therefore it is always on as turning it off would break the data output.
Here is some more info: https://github.com/Aircoookie/WLED/wiki/FAQ#how-to-turn-off-the-bright-blue-onboard-led

huggy-d1:
Handy black nail polish or electrical tape works.  Desoldering the on-board LED is also good if you simply want it off.
If you *need* on-board LED to be separately controllable by HA, you can modify the LED pin in WLED code, compile for the ESP01S, and finally, move the data pin wire to your new pin.
If you want a button input, WLED already has a BTNPIN option.  It also has a RELAYPIN option.
With the BTNPIN, you can wire up a pushbutton that I believe controls RELAYPIN output, which you can define to be the on-board LED pin.
This may not work exactly as described, but perhaps by looking at the wiki (linked above by @Aircoookie, you may find other details that might work for your specific needs.

Aircoookie:
The builtin LED is now turned off when the addressable LEDs are turned off thanks to #1245 :)

PR #1245, opened 2020-10-11
Disable onboard LED when LEDs are not in use

JeffResc:
This pull request seeks to disable the onboard LED when the LEDs are not in use, closing issue #18 and #1210 🎉

As of now, please do not merge this pull request as I am seeking input on how to go about enabling this option. Currently, it is on by default, which shouldn't be an issue, but should this be a build flag or enabled in the UI instead? Please either leave a review suggestion for me to accept or let me know and I can try to implement it myself.

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -650,7 +650,6 @@ class WS2812FX {
     void handle_palette(void);
 
     bool
-      firstRun = true,
       shouldStartBus = false,
       _useRgbw = false,
       _skipFirstMode,
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -650,6 +650,8 @@ class WS2812FX {
     void handle_palette(void);
 
     bool
+      firstRun = true,
+      shouldStartBus = false,
       _useRgbw = false,
       _skipFirstMode,
       _triggered;
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -396,14 +396,18 @@ void WS2812FX::setBrightness(uint8_t b) {
     {
       _segments[i].setOption(SEG_OPTION_FREEZE, false);
     }
-    if (!shouldStartBus)
-      shouldStartBus = true;
+    #if LEDPIN == LED_BUILTIN
+      if (!shouldStartBus)
+        shouldStartBus = true;
+    #endif
   } else {
-    if (shouldStartBus) {
-      shouldStartBus = false;
-      const uint8_t ty = _useRgbw ? 2 : 1;
-      bus->Begin((NeoPixelType)ty, _lengthRaw);
-    }
+    #if LEDPIN == LED_BUILTIN
+      if (shouldStartBus) {
+        shouldStartBus = false;
+        const uint8_t ty = _useRgbw ? 2 : 1;
+        bus->Begin((NeoPixelType)ty, _lengthRaw);
+      }
+    #endif
   }
   if (SEGENV.next_time > millis() + 22 && millis() - _lastShow > MIN_SHOW_DELAY) show();//apply brightness change immediately if no refresh soon
 }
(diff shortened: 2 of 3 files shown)


Aircoookie:
Thank you very much, this is great!

As for enabling it, first off all it should only be enabled/compiled in if `LEDPIN == LED_BUILTIN`. Other than that, I agree that the deallocation and immediate re-initialization of the bus shouldn't cause any issues. If the bus was destroyed at the time of turning off the onboard LED, the released memory could be used by network handlers etc. and thus might cause the bus to be reallocated in a different location, causing heap fragmentation. The way you solved it that shouldn't occur though 👍 
We could add a runtime setting but if this works reliably indeed (will test also on ESP32) I don't think it is needed.

wled00/FX.h
```diff
@@ -650,6 +650,8 @@ class WS2812FX {
     void handle_palette(void);
 
     bool
+      firstRun = true,
```
Aircoookie, line 653: What does this do? Doesn't seem like its value is used anywhere.

JeffResc: I had a `firstRun` variable setup because I thought that my code running on boot was causing a weird flash as I described on Discord [here](https://discordapp.com/channels/473448917040758787/718943978636050542/764759086843428874). It's not needed anymore and has been removed in the latest commit.

JeffResc:
I removed the unused `firstRun` variable and made it only compile if `LEDPIN == LED_BUILTIN`. Let me know if there are any other necessary edits.

JeffResc:
I changed the variable used in the `digitalWrite()` from `LEDPIN` to `LED_BUILTIN`. Although they are checked to be the same and this is purely cosmetic, I believe the code will make more sense to someone reading it this way.

Aircoookie:
Thank you, I tested it on multiple units and it appears to work very well!

Monacoslo:
Sorry, I do not understand the answer. What is the solution now to have D1 led turned off at least when the strip is off?
+

Aircoookie:
@Monacoslo wait until release 0.11.0 and flash that when it comes out :) Will be in 3-4 days at the latest.

Otherwise, check https://github.com/Aircoookie/WLED/wiki/FAQ#how-to-turn-off-the-bright-blue-onboard-led

Monacoslo:
> @Monacoslo wait until release 0.11.0 and flash that when it comes out :) Will be in 3-4 days at the latest.
> 
> Otherwise, check https://github.com/Aircoookie/WLED/wiki/FAQ#how-to-turn-off-the-bright-blue-onboard-led

Great, thank you! I will wait :)

Monacoslo:
It works as expected, when strip is off, also arduino is off

Thank you for another amazing update!

Riza-Aslan:
Is that Something I have to activate in the settings? Cause it isn't working for me with the WS2801. Or is there something different with the WS2801 bin file?

JeffResc:
@Riza-Aslan This isn't a setting activated in the GUI. On the WS2801, is the `LEDPIN` the same as the `LED_BUILTIN` pin? I'm not sure about it as I don't use WS2801, but if `LEDPIN` is not equal to `LED_BUILTIN`, this feature will have no effect.

luca336:
I have encountered an issue. As i have very dirty mains since there isn't a continuous output from the D1 when leds are off and you switch things on and off around the house there are ghost led's.  Can it be an option in the ui to disable and enable this feature.

lloydpick:
It would be nice there could be a setting to just permanently turn off the onboard LED no matter what the state of the strip.

JeffResc:
@lloydpick — Unfortunately, the pin on the microcontroller that controls the onboard LED is the same exact pin that controls the "normal" LEDs. Therefore, if you want any LEDs on at all, the onboard LED must be on. It is just a design feature/flaw of the ESP8266.

DerT94:
I have to ask it :) 
The NodeMCU use GPIO2 for the led. So it can‘t be disabled As long as I control my led strips with it.

but since version 0.12.0 GPIO1 can also be used to control a LED stripe. 
If I would use this, is it possible to disable the status led of the NodeMCU?

Aircoookie:
Yes, if you use GPIO1 the blue LED should stay off permanently.
If it is still on somehow, try rebooting or power cycling WLED.

DerT94:
Perfect, it works

PR #1264, opened 2020-10-17
Add status led

JeffResc:
This pull request adds a status led to WLED. If Wi-Fi is disconnected, the status LED will blink 2 times a second. If MQTT disconnects, the LED will blink 4 times per second. The descriptions below document the variable changes I made:

| Build Flag | Optional | Description |
|-|-|-|
| STATUSLED | Yes, but not defining will disable the functionality | Set this to any pin number besides LEDPIN in order to enable the status LED functionality |
| STATUSLEDINVERTED | Yes | Define this if the status pin should be inverted |

| Global Variable | Description |
|-|-|
| ledStatusLastMillis | Stores the last LED update time from `millis()` |
| ledStatusType | Currently displayed status type. This corresponds to the number of blinks per second. 0 for OK, 2 for Wi-Fi disconnected and 4 for MQTT disconnected. |
| ledStatusState | Current LED state, 0 for `LOW` and 1 for `HIGH` |

### DIFF
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -509,7 +509,7 @@ WLED_GLOBAL UsermodManager usermods _INIT(UsermodManager());
 // Status LED
 #if STATUSLED && STATUSLED != LEDPIN
   WLED_GLOBAL unsigned long ledStatusLastMillis _INIT(0);
-  WLED_GLOBAL int ledStatusType _INIT(0); // current status type - corresponds to number of blinks per second
+  WLED_GLOBAL unsigned short ledStatusType _INIT(0); // current status type - corresponds to number of blinks per second
   WLED_GLOBAL bool ledStatusState _INIT(0); // the current LED state
 #endif
 
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -506,6 +506,13 @@ WLED_GLOBAL WS2812FX strip _INIT(WS2812FX());
 // Usermod manager
 WLED_GLOBAL UsermodManager usermods _INIT(UsermodManager());
 
+// Status LED
+#if STATUSLED && STATUSLED != LEDPIN
+  WLED_GLOBAL unsigned long ledStatusLastMillis _INIT(0);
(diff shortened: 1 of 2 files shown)


wled00/wled.h
```diff
@@ -506,6 +506,13 @@ WLED_GLOBAL WS2812FX strip _INIT(WS2812FX());
 // Usermod manager
 WLED_GLOBAL UsermodManager usermods _INIT(UsermodManager());
 
+// Status LED
+#if STATUSLED && STATUSLED != LEDPIN
+  WLED_GLOBAL unsigned long ledStatusLastMillis _INIT(0);
+  WLED_GLOBAL int ledStatusType _INIT(0); // current status type - corresponds to number of blinks per second
```
JeffResc, line 512: As for the type used (currently `int`), I'm sure something else could be used. This is an integer that can currently be 0, 2 or 4, so something like an `unsigned short`, should be possible, but I could not get it to work as I kept getting a redefinition error. Any pointers here would be great.

JeffResc:
In the future, you may want to integrate other sync interface connectivity issues, but I did not do that yet at least for this initial PR.

To do this just set `ledStatusType = x` when the issue starts where x is the number of blinks per second and when the issue has been resolved, set `ledStatusType = 0`. You can theoretically do this either in the function itself as long as the global variables are accessible, or inside of `WLED::handleStatusLED()`.

jezmck:
Another option would be to use the first LED as the status indicator, since it's already possible to use the 2nd LED as the beginning of the string.

JeffResc:
@jezmck I didn’t think about that, but that will help users who only have one onboard LED, which is on pin 2, such as the D1 mini.

@Aircoookie What do you think?

JeffResc:
I optimized the code so that it is 1.) shorter and 2.) uses less memory by changing ledStatusType from an `int` to an `unsigned short`. Let me know if you can think of any other changes, but other than that, this should be good to go :+1:

Aircoookie:
Awesome, thanks! Seems to work well for me too when enabling it in my platformio override :)

PR #1266, opened 2020-10-18
Add gzip support for full OTA possibilitys for 1M build

Jason2866:
refactor PlatformIO to automatically name firmwares and add gz compressed file.
Copy compiled firmwares in folder `build_output/firmware`
Do some (small) optimizations to save flash (build_flags)
Switch off disable off all warnings
Added latest and well tested core 2.7.4.7 (flicker fix for analog PWM. Arduino PR 7231)
Added in Platformio_override.ini.example 1M build which can be flashed directly from Tasmota-lite
Updated ESP32 to Platformio 2.0
Added the possibility to use a user_config_override.h fille. Will be used if it is there.
If none is provided a dummy will be generated
Added gitpod support to compile WLED with

Greeetings from Tasmota crew :-)


Aircoookie:
Will get to your PR soon, sorry for taking so long :)

Jason2866:
There is no rush 🖖

probonopd:
> Added in Platformio_override.ini.example 1M build which can be flashed directly from Tasmota-lite

Tasmota 8.5.1 (minimal) is installed on my device.
When I try to flash `WLED_0.12.0_ESP8266_1M.bin` via the Tasmota OTA web interface, I get:

> Upload Failed
> Upload buffer miscompare

Is it supposed to work? If so, how?

blazoncek:
> Is it supposed to work? If so, how?

No. WLED image size is now too big even gzipped.

PR #1268, opened 2020-10-19
Add status pixel

JeffResc:
This PR builds upon my last PR, #1264, but instead of using a GPIO LED, the user can now use the first pixel as a status LED.

New build flags:
| Build Flag | Optional | Description |
|-|-|-|
| STATUSLED | Yes, but not defining will disable the functionality | Set this to -1 in order to enable status pixel. |

New global variables:
| Global Variable | Description |
|-|-|
| lastActive | Stores if the status was last active so when we disable it, we only disable it once instead of setting `currentOverlay` to `lastOverlay` over and over again. |
| lastOverlay | Same as `currentOverlay`, but it stores the last overlay before enabling the status pixel overlay. That way we can go back to whatever overlay mode the user previously set. |

### DIFF
--- a/wled00/overlay.cpp
+++ b/wled00/overlay.cpp
@@ -121,18 +121,20 @@ void _overlayAnalogCountdown()
 
 void _overlayStatusPixel()
 {
-  if (briT == 0) {
-    strip.setRange(1, ledCount-1, 0);
-    #ifdef STATUSPIXELBRI
-      strip.setBrightness(STATUSPIXELBRI);
+  #if STATUSLED == -1
+    if (briT == 0) {
+      strip.setRange(1, ledCount-1, 0);
+      #ifdef STATUSPIXELBRI
+        strip.setBrightness(STATUSPIXELBRI);
+      #else
+        strip.setBrightness(255);
+      #endif
+    }
+    #ifdef STATUSPIXELCOL
+      strip.setPixelColor(0, (ledStatusState ? STATUSPIXELCOL : 0));
     #else
-      strip.setBrightness(255);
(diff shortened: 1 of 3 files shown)


JeffResc:
Please do not merge this PR yet. It is almost working, but it currently does not work when the LEDs are off. I will add another comment once this issue is fixed.

JeffResc:
Alright, everything should be fixed now. I also added a few more build flags for further customization:
| Build Flag | Optional | Description |
|-|-|-|
| STATUSPIXELBRI | Yes | If set, this will be the brightness of the status pixel when the rest of the strip is off. Defaults to 255. |
| STATUSPIXELCOL | Yes | Sets the color of the status pixel when on. Defaults to 255. |

JeffResc:
@Aircoookie Have you been able to take a look at this? It’s pretty similar to #1264.

otisrhw:
pinMode (2, output)? D4 shares board led.

JeffResc:
@otisrhw Could you please let me know which file and line you found that on? I did a quick search and couldn’t find an edit I made involving `pinMode`.

otisrhw:
Sorry, I didn't check code, just threw it out there. On another project, I didn't account for the board led and was unaware it shared d4, this caused status light issues. I very novice user, just trying to help.

tony-fav:
Great idea! Especially in setups using a sacrificial pixel. I'm sure some updates are necessary to get this functional in latest, but is there any particular issue keeping this from being merged?

blazoncek:
Due to changes from 0.11 to 0.12, where npbwrapper was removed and bus_manager introduced, this kind of code would need to go into bus manager itself.
This PR became obsolete.

PR #1269, opened 2020-10-19
Add status message as toast in UI

JeffResc:
This PR further expands on #1268, so please take a look at that one before this.

In this PR, I added custom error messages to each error situation displayed on the LED. Then, it writes that message to the info subsection of the JSON API. From there, when JSON is fetched, it will check if that message is not `OK` and display it in the UI as a toast message. As soon as the JSON API displays `OK` again, the toast message will disappear.

This doesn't make the most sense for a Wi-Fi status issue, but I added it now for consistency. The MQTT status works as expected. In another PR, I plan to add more status modes for the other sync interfaces available so that the status LED will reflect the status of those as well.

Please note: Build flag `STATUSLED` must be set to a valid option for this PR to work, otherwise the message will always read as `OK`.

Image:
UI Image


Aircoookie:
Hi again!
I have yet to fully review the status pixel PR - but wanted to give a little feedback here first!
The idea of also displaying the status in the UI is great - it's already something I've been doing on the new filesystem branch to indicate errors and there is already a numerical variable for it, `errorFlag`. Here are the error codes I've assigned so far: https://github.com/Aircoookie/WLED/blob/filesystem/wled00/const.h#L109
The JSON property for it is just `error` in the state object, though I should move that over to info.

Maybe we could switch this over to use that instead -  for example error 30 for wifi disconnected and error 31 for MQTT. The actual error string would then be selected by a javascript switch() in the UI code (so far I haven't done that - it just says `WLED error XX` in the toast if the flag is non-zero)

Thank you for the many PRs :) I highly appreciate it!

JeffResc:
Since those changes are on the filesystem branch, should I push to there or wait until filesystem is merged to master?

huksley:
@Aircoookie Contributors still try to change html_*.h files directly instead of modifying src/data files. Even having the "Autogenerated warning at the top". Consider deleting html_*.h files and gitignoring them to make sure you have a single source of truth for HTML UI.

JeffResc:
@huksley You know, I know that that exists, but I’ve seen a few people PR it and I wasn’t sure so I included it as well

Aircoookie:
@huksley thank you!
Yes, indeed I do not merge/will correct beforehand if a PR introduces a new setting only in `html_settings.h`.

In this PR it is an exception which I am to blame for, and that is that for some reason the npm script does not compress the javascript in the UI page of html_settings.h so I removed tabs from it manually. Those are reintroduced with any PR that uses the npm script to make changes to any HTML page. Not an ideal situation at all, I need to look into how to compress that JS automatically soon.

@JeffResc sorry, it does not seem like I responded to your question. Keeping the base as master is easiest for me for now. You may just hard-code a numeric error code and I will add the define in `const.h` for it later :)

huksley:
@Aircoookie Take a look here https://github.com/Aircoookie/WLED/pull/1287, figured out the problem why js was not minifed properly

JeffResc:
I’ll close this for now and I’ll redo this using errorFlag

PR #1287, opened 2020-10-26
Use html-minifer-terser for ES6+ support

huksley:
Fix fixes minification issue with `wled00/data/settings_ui.htm` which is caused by html-minifer (and uglify-js) not supporting ES6 / ES2015 modern syntax, like default function parameter values, multiline-strings, etc. 

See here for more info: https://github.com/kangax/html-minifier/issues/1040

⚠️ Minification result have not been tested (I don't have access to ESP right now), needs to be tested on real machine first.


Aircoookie:
Thank you so much for tracking down the issue with the minification. I would have not realized that the minifier does not support ES6. Tested on device and working well! 👍

PR #1297, opened 2020-10-29
Mention analog strips in README

franga2000:
Hi!

I know the main focus of this project are addressable strips, but I have found it immensely useful with my traditional "analog" strips and I have a feeling many people don't realize it can be used for that too (I was one of those people, as I didn't realize this a year ago when I was looking for control software and found this project).

My guess is many people find this project and turn around just like I did because the analog support isn't mentioned on the front page and I though adding a least a bit about it in the README might help.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -73,6 +73,7 @@ TM1814 | 12v | RGBW
 WS2811 | 12v | 3-LED segments
 WS2815 | 12v | 
 GS8208 | 12v |
+Analog/non-addressable | any | Requires additional circuitry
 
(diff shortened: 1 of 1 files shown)


ulte:
this is a good pt. I wasnt aware either. btw any prebuilt hardware suggestions (i know we need mosfets)

franga2000:
Mine is custom, using IRBL8721 mosfets and a Wemos Mini clone, but I know some people have managed to flash WLED onto H801 boards (some strange Chinese esp8285-based rgbw controller). You can get them for < 10€, but I haven't used them before so I can't say anything about their stability.

PR #1316, opened 2020-11-04
WIP: Ethernet Support

tbnobody:
As discussed here https://github.com/Aircoookie/WLED/issues/1027 this PR is adding support for the ethernet interface. It's maybe not yet 100% tested but ethernet is working and it still works with ESP8266 and ESP32 without ETH.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -252,7 +252,7 @@ lib_ignore =
 [env:esp32_poe]
 board = esp32-poe
 platform = espressif32@1.12.4
-build_flags = ${common.build_flags_esp32} ${common.debug_flags}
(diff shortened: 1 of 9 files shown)


tbnobody (commented):
You are right, makes absolutly sense. I think otherwise e.g. the ETH.begin() would make some strange things without the ethernet hardware

Aircoookie:
Actually I believe ETH.begin() would be fine. The WiFiEvent would just never be called.
However, I observed that the ethernet stack seems to use ~43kB of extra heap memory.
Even though we have lots of heap on ESP32, compiling it in on non-ethernet boards doesn't really serve any purpose.

In my testing this seems to work very well, thank you! Will merge tomorrow if no further issues are reported :)

PR #1366, opened 2020-11-22
Fix code for long LED strips (>256 LEDs long)

Def3nder:
Hi @Aircoookie,

I just flashed the most recent  version on my son's strripes and played around with some effects.
As I've synced all three WLED instances I could see that the "Phased" effect did behave other on the long strip than it does on the other (short) ones. 
I looked into the code and found the reason:

.... 256 / SEGLEN ...   assigned to an integer 😄 this will be 0️⃣ if the SEGLEN is bigger then 256.

So I fixed the code with this PR.

...many greatings from Berlin - I hope to have more time in the future to work with coding, again - it's quite fun ⚡ 

Ralf aka Def3nder


Aircoookie:
Awesome, thanks!

PR #1383, opened 2020-11-27
mutiple RGBW leds support for DMX control

sunbowch:
RGBW support for individual leds through DMX control. (Tested with E1.31 & Artnet)
-Added option 'Multi RGBW' to the settings
-Modified e131.cpp accordingly
-Tuned other parameters to make it work
To do: Same for other DMX options

### DIFF
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -262,7 +262,7 @@ WLED_GLOBAL uint16_t e131ProxyUniverse _INIT(0);                  // output this
 #endif
 WLED_GLOBAL uint16_t e131Universe _INIT(1);                       // settings for E1.31 (sACN) protocol (only DMX_MODE_MULTIPLE_* can span over consequtive universes)
 WLED_GLOBAL uint16_t e131Port _INIT(5568);                        // DMX in port. E1.31 default is 5568, Art-Net is 6454
-WLED_GLOBAL byte DMXMode _INIT(DMX_MODE_MULTIPLE_RGBW-1);            // DMX mode (s.a.)
+WLED_GLOBAL byte DMXMode _INIT(DMX_MODE_MULTIPLE_RGB);            // DMX mode (s.a.)
 WLED_GLOBAL uint16_t DMXAddress _INIT(1);                         // DMX start address of fixture, a.k.a. first Channel [for E1.31 (sACN) protocol]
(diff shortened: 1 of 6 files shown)


wled00/wled.h
```diff
@@ -262,7 +262,7 @@ WLED_GLOBAL uint16_t e131ProxyUniverse _INIT(0);                  // output this
 #endif
 WLED_GLOBAL uint16_t e131Universe _INIT(1);                       // settings for E1.31 (sACN) protocol (only DMX_MODE_MULTIPLE_* can span over consequtive universes)
 WLED_GLOBAL uint16_t e131Port _INIT(5568);                        // DMX in port. E1.31 default is 5568, Art-Net is 6454
-WLED_GLOBAL byte DMXMode _INIT(DMX_MODE_MULTIPLE_RGB);            // DMX mode (s.a.)
+WLED_GLOBAL byte DMXMode _INIT(DMX_MODE_MULTIPLE_RGBW-1);            // DMX mode (s.a.)
```
Aircoookie, line 265: This sets the default mode to `DMX_MODE_MULTIPLE_DRGB`. I don't think that was your intention. Shouldn't it stay `DMX_MODE_MULTIPLE_RGB`? Would make the change myself, just asking for the motivation.

Other than that, this is ready to merge, thank you!

sunbowch: I also spotted this after the the merge request... It was left over when I was looking why the new option wasn't taken into account.

Obviously the RGB must stay default.  Please cancel this mod.

I don't need more DMX control than that at the moment... But it would be logical to add RGBW and DRGBW support for single option.

Not hard to do roughly... But needs more reflection to have a cleaner implementation (redefine MAX_LED_PER_UNIVERSE -> variable).

Will try to do that when I'll be in need.

Aircoookie:
Made this a little neater by integrating the RGBW code into the Multi RGB case block since most code is identical. Give it a try if you like :) Also added the Multi RGBW option into the dropdown, seems like it wasn't added to the `html_settings.h` yet

Aircoookie:
Multi DRGBW would indeed make sense to add, will think about that. Single RGB/DRGB actually support the white channel already (despite the name), just pass an extra channel at the end

PR #1395, opened 2020-11-29
Fix IR-Remote type to cfg.json

Def3nder:
...instead of defaulting it to 0 

The IR-Remote was disabled after each reboot (because in the JSON Object of the IR got hardcoded a "0").
I changed this to the IR-Remote type (which is stored in the byte `irEnabled`).


Aircoookie:
And the tradition of finding a very annoying bug literally minutes after a major release continues 😉 thank you for the fix though!

PR #1416, opened 2020-12-02
Add default build task for VSCode

danecreekphotography:
This PR adds a `tasks.json` file that defines a default VSCode build tasks that runs the following two build commands in sequential order:

* `npm run build`
* `PlatformIO: Build`

*Rationale*

I was pulling my hair out trying to figure out why my changes to `index.html` weren't taking effect in the compiled binary. Then I discovered there's a second build command to generate the compressed file for inclusion in the binary.

With this change doing `ctrl-shift-B` in VSCode will automatically run both steps and ensure a binary is produced with the updated index.html.

There are no changes to the existing `PlatformIO: Build` step, and triggering PlatformIO build using any of the existing methods (Checkmark in the bottom status bar, ctrl-alt-B, or from the command palette) still does only the PlatformIO build.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,33 @@
+{
+  "version": "2.0.0",
+  "tasks": [
+    {
+      "type": "PlatformIO",
+      "task": "Build",
+      "group": {
+        "kind": "build",
+        "isDefault": true,
+      },
+      "problemMatcher": [
+        "$platformio"
+      ],
+      "presentation": {
+        "panel": "shared",
+      },
+      "dependsOn": [
+        "npm: build"
+      ],
+    },
+    {
+      "type": "npm",
+      "script": "build",
+      "group": "build",
+      "problemMatcher": [],
+      "label": "npm: build",
+      "detail": "npm run build",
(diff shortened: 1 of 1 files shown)


danecreekphotography:
Hmmmm of course something isn't working right after I open this :( Never mind, will re-open if I get the bug fixed.

Aircoookie:
This is a cool idea, thanks!
Perhaps we need to also run `npm install`? At least I have to do every time i clone a fork, else `npm run build` will fail.

danecreekphotography:
So I hit two issues:

1. The .bin file isn't getting dropped into the output folder. Do you happen to know where the magic is that causes that final output to get generated?
2. This does impact the base PlatformIO: Build command, it also runs the `npm run build` command. In practice I think that's ok, the npm part doesn't take long at all and running it on every build means it will always have the latest HTML.

`npm install` takes too long and shouldn't be run on every build, it should be part of grabbing the initial fork. I can add it to the .devcontainer stuff I sent along earlier. That means anyone using dev containers for development would have npm install run when the container gets created by VSCode.

PR #1421, opened 2020-12-02
Add VSCode build tasks

danecreekphotography:
This PR adds a `tasks.json` file that defines a default VSCode build tasks that runs the following two build commands in sequential order:

* `npm run build`
* `PlatformIO: Build`

*Rationale*

I was pulling my hair out trying to figure out why my changes to `index.html` weren't taking effect in the compiled binary. Then I discovered there's a second build command to generate the compressed file for inclusion in the binary.

With this change doing `ctrl-shift-B` in VSCode or running `PlatformIO: Build` will automatically run both steps and ensure a binary is produced with the updated index.html.

**Warning**: This alters the default behaviour of the PlatformIO build and anyone doing a fresh clone of this repo who fails to do `npm install` after cloning will get a build error because the HTML build step will fail. If you choose to merge this the steps at https://github.com/Aircoookie/WLED/wiki/Compiling-WLED should get updated to specifically say running `npm install` is required.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -2,8 +2,19 @@
   "version": "2.0.0",
   "tasks": [
     {
+      "label": "Build: HTML and binary",
+      "dependsOn": [
+        "Build: HTML only",
+        "Build: binary only"
+      ],
+      "dependsOrder": "sequence",
+      "problemMatcher": [
+        "$platformio",
+      ],
+    },
+    {
       "type": "PlatformIO",
-      "label": "Build: binary",
+      "label": "Build: binary only",
       "task": "Build",
       "group": {
         "kind": "build",
@@ -21,7 +32,7 @@
       "script": "build",
       "group": "build",
       "problemMatcher": [],
-      "label": "Build: HTML",
+      "label": "Build: HTML only",
       "detail": "npm run build",
       "presentation": {
         "panel": "shared"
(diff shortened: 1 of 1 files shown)


Aircoookie:
Thank you very much!

Do you know of a way to run the build (compilation) in the default configuration successfully even if `npm run build` failed? I would not like to add an extra step to the process for newer users, especially not running a terminal command, as simple as it seems for us. I want to encourage new users to build from source instead of using the binaries, and most of them do not modify the HTML.

Another option might be a way to run `npm install` only on the first build (quite similar to how it is in the devcontainer, though I understand that container and local compile are completely separate environments) and somehow keep track of whether it was run already)

Otherwise I would believe keeping the current behavior is also acceptable - you can run the npm script with a single click in VS Code and if you are doing a lot of changes/testing to the HTML, there is still `npm run dev` which watches the HTML files and re-does the HTML setup.

danecreekphotography:
Good feedback, thanks! I updated the PR and now there are three build tasks:

image

The default is `Build: Binary only` which leaves the behaviour unchanged for users. The `Build: HTML and binary` task will do both the HTML and the binary but only if manually selected by the user. For people who are actively editing HTML they can change it to the default build task themselves in `tasks.json` if they prefer.

Aircoookie:
Tested and working well!

PR #1430, opened 2020-12-04
Added Temperature Palette

k7bbr:
Added cpt-city temperature_gp palette. Palette is nice combination of blues to yellows to reds. Reduced original cpt-city palette to 18 lines to fit WLED constraints. Tested palette and it visually is consistent with original cpt-city colors and look and works in WLED.


Aircoookie:
Thank you very much, this palette looks great!

PR #1445, opened 2020-12-06
Adding a candy cane mode

munsterlander:
It was a common request when installing WLED to add a candy cane mode as a saved mode, so figured would add it to a PR.


sansillusion:
Effect not selectable, will run Washing Machine instead. I compiled usin pio and cant select the fx through html page or api. the effect shows up but it's washing machine that get selected.

fishbone-git:
#define MODE_COUNT  114 was nto updated.  It should be 115 after adding Candy Cane

Issue #1442, opened 2020-12-06
Cannot build with usermod PIR sensor switch


marulitua:
**Describe the bug**
I want to add PIR to my setup, so I follow [this documentation](https://github.com/Aircoookie/WLED/tree/v0.11.0/usermods/PIR_sensor_switch), but cannot pass the linking process

**To Reproduce**
`cp usermods/PIR_sensor_switch/usermod_PIR_sensor_switch.h wled00/.` and register the usermod

**Expected behavior**
Should be able to build

**WLED version**
 - Board: esp32 devkit v1
 - Version v0.11.0
 - Format self-compiled

**Additional context**
This is the error message
```
Processing esp32dev (board: esp32dev; platform: espressif32@2.0; framework: arduino)
--------------------------------------------------------------------------------------------------------------------------------------------------------------
Verbose mode can be enabled via `-v, --verbose` option
CONFIGURATION: https://docs.platformio.org/page/boards/espressif32/esp32dev.html
PLATFORM: Espressif 32 (2.0.0) > Espressif ESP32 Dev Module
HARDWARE: ESP32 240MHz, 320KB RAM, 4MB Flash
DEBUG: Current (esp-prog) External (esp-prog, iot-bus-jtag, jlink, minimodule, olimex-arm-usb-ocd, olimex-arm-usb-ocd-h, olimex-arm-usb-tiny-h, olimex-jtag-tiny, tumpa)
PACKAGES: 
 - framework-arduinoespressif32 3.10004.201016 (1.0.4) 
 - tool-esptoolpy 1.20600.0 (2.6.0) 
 - toolchain-xtensa32 2.50200.80 (5.2.0)
Converting wled00.ino
LDF: Library Dependency Finder -> http://bit.ly/configure-pio-ldf
LDF Modes: Finder ~ chain, Compatibility ~ strict
Found 32 compatible libraries
Scanning dependencies...
Dependency Graph
|-- <FastLED> 3.3.2
|   |-- <SPI> 1.0
|-- <NeoPixelBus> 2.6.0
|   |-- <SPI> 1.0
|-- <AsyncTCP> 1.0.3
|-- <IRremoteESP8266> 2.7.3
|-- <LittleFS_esp32> 1.0.5+sha.b941415
|   |-- <FS> 1.0
|-- <ESP Async WebServer> 2.0.0+sha.3336275
|   |-- <AsyncTCP> 1.0.3
|   |-- <FS> 1.0
[...]

PR #1448, opened 2020-12-07
Update usermods: FixUnreachableNetServices  and PIRsensorSwitch

gegu:
- Fix ESP32 compilation issue for PIRsensorSwitch. #1442 
- Configurations in the web UI of both usermods have been extended and changed values are saved permanently.
- PIRsensorSwitch supports the external change of the sensor state. For example from another usermod.


gegu:
@guilhermebera:
I removed section **Usermod installation (advanced mode)** from the README again, because in the current version the sensor state can be changed from outside and no global variable is needed anymore.
The section also contained some references to line numbers and things change over time 😃

marulitua:
Please accept this man PR :satisfied:

Issue #1449, opened 2020-12-07
Analog clock settings not stored


gegu:
**Describe the bug**
With the settings for the analog clock some values are not stored in cfg.json

- First LED
- Last LED
- 12h LED
- Show 5min marks
- Seconds (as trail)

The others I have not tested.

part of cfg.json:
```json
{
"ol":{
  "clock":1,
  "cntdwn":false
}
```

**To Reproduce**
Change values in web UI **Time & Macros** and reboot

**Expected behavior**

The values should be stored.

**WLED version**
 - Board: Wemos D1, mini
 - Version 0.11.0
 - Format self-compiled

Aircoookie:
Thank you for the report and fix. Merged :)

PR #1455, opened 2020-12-08
Store analog clock settings

gegu:
Solves #1449.


Aircoookie:
Thanks very much :)
Will merge with your other PR #1448 as it is also included there (probably by accident), but that way I can avoid a merge conflict!

PR #1459, opened 2020-12-08
Fix #1400: Using sliders / colorwheel sometimes slides the whole UI

### LINKED ISSUE #1400: Slider Annoyance on Phone
**Describe the bug**
Using color slider changes webpage

**To Reproduce**
Slide the color slider

**Expected behavior**
To choose colors with slider without it going it going to home page or effects page if sliding to the right.

**WLED version**
 - Board: Doesn't matter
 - Version 11
 - Format Binary

**Additional context**
Possible to have the bottom row, color, effects etc to not accept "swipes" out of their command zone (don't know how else to describe it sorry), when not hovering over them, any swipes on the screen do not interfere with them.

Thank you for your help!

allgoewer:
Hi @Aircoookie,

this is an attempt to fix #1400 .

iro.js has been updated in an earlier commit which changed the names of the iro-classes.
This led to the function check() in wled00/data/index.htm no longer correctly detecting whether the touch event originated from one of the sliders or the colorwheel.

My initial attempt to simply update the class names did not work as expected.
While this reduced the issue, it did not completely remove it.

Turns out some of iro.js's elements (such as the ring that makes up the slider / colorwheel knob) do not have a class name set.

Therefore we now check not only for classnames starting with "Iro" in the element the event originated from but also it's parent, if necessary.

Also I'm not sure whether we should notify the iro.js maintainers of this issue.


Aircoookie:
Thank you very much, your fix works well for me!

jaames:
> Also I'm not sure whether we should notify the iro.js maintainers of this issue
iro.js maintainer here, please do!

PR #1462, opened 2020-12-09
Add preset number to display

danecreekphotography:
A simple change to show the preset slot number next to the name of the preset in the preset list. This really helps when trying to program preset cycles or set the default preset.

image


Aircoookie:
Thank you very much - I agree it would make getting the ID for the boot preset and timed presets a lot easier!

Most likely I will give the ID a separate div tomorrow, make it grey and a bit smaller, and place it on the left of the preset, so that the actual name is still centered, then will merge :)

mike2nl:
@Aircoookie 
you wrote: -> Most likely I will give the ID a separate div tomorrow
Means on Fryday right? Because i ask i will update the fireplace animation for the kids fireplace out of hart paper. They love it.

Aircoookie:
Not yet sure if I love it... Makes everything a bit noisy but the ID is essential info after all. Maybe it needs to be a setting or shown dynamically, but in any case the benefit of having the ID available without expanding the presets might outweight that drawback :)  

presetnumbers

danecreekphotography:
I think your screenshot looks good, it's off on the left and a lighter colour. Not too noisy at all. My only worry is it might not be enough contrast for people with vision impairment.

PR #1466, opened 2020-12-10
Add utility function to expose if strip is being updated

pbolduc:
This PR addresses issue #1461. In some areas, like the dallas one wire temperature usermod, we must avoid doing things that could impact the precise timing required by the WS2801 protocol. Issues can be observed that due to timing issues, the colors sent to the strips get messed up.  This can be seen in the logic analyzer capture below.

ESP32
WLED 0.10.2 or 0.11.0
WLED Effect: Solid Blue 0x000038

When delays are introduced, the colors bits are moving around sending other color codes causing the lights to go crazy when the sensor is read.

pulseview_zqMDhmmNZ6

This PR lays the ground work to allow the temperature usermod to find a time between frame updates to read the sensor.

PR #1478, opened 2020-12-12
Update esp32_multistrip usermod to fix compile errors

broccoliboy:
Update my usermod with CanShow function to track core NpbWrapper.h and fix compile errors caused by #1466. Have not tested on hardware yet.


pszafer:
hi, with this branch second pin is not working for me and first it's a mess. half of the strip is working, second half is flickering, it seems like data from second strip is trying to overwrite data from first strip.

andyshinn:
Do you have any other usermods or modifications? I just applied these changes manually to `master` and testing multiple segments and single segment seems fine.

pszafer:
@andyshinn , no I don't.
I just adding
```
RgbColor rgb2 = RgbColor(c.G, c.R, c.B);
```
because I have two different strips.
I will try tonight to erase_flash and check if with "fresh" esp it's working

pszafer:
with pre-erased flash it's working!

PR #1495, opened 2020-12-15
Wlan on LAN

maverick1975:
hello, is possible install  wlan software on lan device like arduino with lan shield? tanks


Aircoookie:
Hello! First of all I don't think you meant to open a pull request :)

LAN is now supported, however not with the Arduino boards/shields, only with the ESP32. One board that supports it is the `Olimex ESP32 POE` which I highly recommend.
More info in https://github.com/Aircoookie/WLED/issues/1027

maverick1975:
I'm very sorry but I'm a new user and I still don't understand what the rules are like
thanks for the reply i will definitely try.

PR #1497, opened 2020-12-15
fix alexa discovery

lucaf86:
Hello,

Alexa discovery is broken in Wled 0.11.0 but it work fine for me on another project using Fauxmo esp library on a wemos d1 mini .
I Analyzed the messages exchaged between my echo show and the d1 and I compared them against the Fauxmo lib.
I've found that the 7 digit unique id calculated by Wled is not correct, it must be 12 character long so I ported the Fauxmo id calculation into Wled and fixed also the checks on M-SEARCH message since on my echo show upnp:rootdevice was used instead of ssdp:all.
Everything tested with Wled on my d1 mini and Alexa discovery now works fine. I can controll the on/off, bri, and color from Alexa.

### DIFF
--- a/wled00/src/dependencies/espalexa/Espalexa.h
+++ b/wled00/src/dependencies/espalexa/Espalexa.h
@@ -60,6 +60,7 @@
 
 #include "EspalexaDevice.h"
 
+#define DEVICE_UNIQUE_ID_LENGTH 12
 
 class Espalexa {
 private:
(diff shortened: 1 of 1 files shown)


Aircoookie:
Hi!
Thank you so much, your fix is a godsent! It instantly made one of my devices I've been having trouble with discover without issue!

Would you like to open a PR on my [Espalexa library](https://github.com/Aircoookie/Espalexa) itself as well? I would like to get that to work again (and bump the version to 2.5.0) and then integrate it to WLED and release the fixed version 0.11.

The MD5 calculation is a great idea to make sure that the IDs are truly unique, but i see a potential issue if the device is renamed (not in the Alexa app, but directly in e.g. WLED), that would change the ID. What I would propose is an uniqueid `aaaaaaaaaaii`, where `a` are the last 10 digits of the mac address and `i` is the device number/index in Espalexa in hex format. That should still ensure completely unique IDs, solve the name issue and reduce calculations/string manipulations needed in `encodeLightId()`. I will give it a try, or do you think that approach is flawed?

Thanks again :)

lucaf86:
@Aircoookie what you propose is similar to what I thought of implementing but I'm not very familiar with string manipulation in c/c++ so in the end it was easier for me to take the Fauxmo implementation and integrate it here.
However I think I can try to do it and then open the PR on Espalexa lib

Thanks

PR #1528, opened 2020-12-19
Added css property for momentum-scrolling IOS

GrumpyMeow:
On my iPhone there  was never any momentum-scrolling on the effects-tab.
With this single css property, it makes navigating the effects so much more pleasent.


Aircoookie:
Hi and thanl you :)
Would it be okay if i included the line manually? The compressed HTML is not the most recent version, so I can't merge this easily.

GrumpyMeow:
Ofcourse!

Aircoookie:
Thanks very much!
I will take a look at your other PRs soon, so you'll get contributor status either way :)

Aircoookie:
Added! Ah, now I also see how the HTML could be outdated, this PR was against the 'dev' branch (which actually makes sense, I need to merge in the changes from master more often)

GrumpyMeow:
Yes indeed. I did branch of the "dev" branch.

PR #1529, opened 2020-12-19
Added liveviewws-page for liveview via websockets

GrumpyMeow:
Hi,
The live-preview experience always felt somewhat sluggish to me. I noticed the original implementation was kind-of "dated" which caused some overhead at the ESP and browser.
This PR is for using Websockets and requestAnimationFrame to achieve a more optimal experience.

Without the WLED_ENABLE_WEBSOCKETS-define the original implementation is still used.


Aircoookie:
This appears to work great, thank you!

PR #1532, opened 2020-12-19
Implemented svg->woff

GrumpyMeow:
Extended the build-process to automatically generate a font by using the svg-files from the icons-folder.
During build the base64-encoded-woff-font-data will replace "[ICONFONT_REPLACE]"  in the index.html.
The svg-filenames can/should be prefixed with the desired unicode to use for that icon.
I've now added the svg-files which were used to generate the [current iconfont](https://user-images.githubusercontent.com/21045690/95679241-be143d80-0bd1-11eb-84b7-b92d045eb04b.PNG).
During the build a list will be displayed with unicode<->filename.


GrumpyMeow:
Hmmm.. I rebased my feature-branch on your Dev-branch. I'm not too sure if this went ok.
Please let me know if it would be better for me to recreate this feature-branch...

PR #1536, opened 2020-12-20
Fixes in settings.htm

WoodyLetsCode:
Hello,
I fixed the following errors / warnings in settings.htm:
grafik

PR #1540, opened 2020-12-20
Change to utf-8 to support localization

Master-MiShutka:


Aircoookie:
Already fixed in master by #1536 :)

PR #1542, opened 2020-12-20
split up index.htm

WoodyLetsCode:
Hey,
this splits up index.htm in 5 files:
**index.htm:** main html code
**index.js:** page javascript code
**index,css:** page css
**iro.js:** color picker
**rangetouch.js:** range touch library

While i split this up I saw that a new version of RangeTouch is available. So i updated it to the lastest verson (v2.0.1).

Detailed informations can be found here: https://github.com/pbolduc/WLED/issues/3#issuecomment-748610769


Aircoookie:
Hey,
just tested this and it appears to work great! No difference in the "baked" HTML at all, just as it should be.
Also, the local file mode for testing changes to the UI without uploading to the ESP each time still works (which is very important to me!).
Separating the libraries into their own files is also great!
I'm very happy that you have come up with a solution to split the file without any apparent drawback :)

I was just about to complain that the page JS is not minified properly in the baked HTML, then I realized it isn't either in master, so this is not a new problem from the split. I suppose we can fix it by applying the fix from #1287 to the page JS as well, but I think that is a different topic and does not need to be addressed here.

There is just one minor problem left: The four new files `index.css`, `index.js`, `iro.js` and `rangetouch.js` are added to the `html_settings.h` file. This is not _really_ a problem since it goes unused and thus is not included in the final binary by the compiler, but it would be nicer to not have it in that file.
It appears to be fixed for me by reverting your additions to the `cdata.js` file. Do you think we need those? It appeared to still merge the css and js correctly without it, but I want to make sure it doesn't break anything before pushing a commit removing it.
Looking forward to finally merging this :)

WoodyLetsCode:
Thanks for telling me this! Actually I made these changes in `cdata.js`, because I thought I need them and I woudn't work without them. But now I tested it without these changes and everything works fine. Also the script minified the css code.
So yes, I think you can merge this finally. 😄

blazoncek:
@WoodyLetsCode would it be possible to revert embedding iro.js and rangetouch.js into index.htm?
I have built a substitute index.htm (named simple.htm) with simplified UI. Now I have iro.js and rangetouch.js included twice in header files, but would prefer to have them as separate const char* and served by WLED web server on request.

WoodyLetsCode:
@blazoncek Theoretically, it should be possible. But why would you do this? Aircoookie told me that having one request instead of several requests is better, because multiple requests can overload the esp. Also, I do not understand why you have iro.js and rangetouch.js included twice in header files? It should only be included once for every time you include it in the header.

blazoncek:
Hi, indeed one request is better than many.
Said that, I have created a suplement simple.htm that also requires iro.js and rangetouch.js since it is intended to replace index.htm (I also added a UI setting to enable swiching between the two).
I have modified cdata.js to create header file for simple.js in the same way as was for index.htm and since iro and rangetouch are inlined into index.htm and simple.htm they are uding twice as much flash space.

WoodyLetsCode:
As far as I know I think you don't have to edit anything in cdata.js. Did you try it without editing anything in cdata.js?

blazoncek:
> As far as I know I think you don't have to edit anything in cdata.js. Did you try it without editing anything in cdata.js?

Oh, yes. I added:

`writeHtmlGzipped("wled00/data/simple.htm", "wled00/html_simple.h", 'simple');`

and changed

`function writeHtmlGzipped(sourceFile, resultFile, page)`

WoodyLetsCode:
So does it work now like you expected?

blazoncek:
No. iro.js and rangetouch.js are inlined into index.htm and simple.htm doubling the flash space used.
Unfortunately I do not know enough to correct cdata.js to gzip individual files without inlining JS files (or CSS).

WoodyLetsCode:
Oh, yes, you are right. The tool is embedding iro.js and rangetouch.js every time you call it, and so also for your simple.htm. Maybe you talk with Aircoookie about this and ask if these changes would be okay, because importing iro.js and rangetouch.js through a request would save flash space.
Currently, I also don't know how to disable this in cdata.js 😢, so I also would have to do some research here.

Issue #938, opened 2020-05-24
Feature request: Fake TV LED effect


fribse:
Hi @Aircoookie 

Now that the summer is coming up, people go on holiday (not much international travel this year though).
A lot of people buy these fake TV led gizmo's that they set up, to portray a living tv, and illustrate that people are home. Most of them are just individually addressable LED's in a matrix of some sort and often in a very poor quality. Maybe this could be a thing for WLED to do?
It wouldn't be that difficult to take a string place it on a platter of some sort, and attach a WLED to it.
Then it could all be controlled like the rest of the light in the house.

Legsmaniac:
Just curious but isn't the use of WLED a bit overkill for fake TV?  I mean, isn't this the sort of thing usually created stand-alone, especially since there are already heaps of projects out there for fake TV effects?   🤔

Incidentally, I chanced upon [this](https://learn.adafruit.com/fake-tv-light-for-engineers/overview) very interesting project and just tried it on a spare ESP I had.  Apparently, it's supposed to faithfully reproduce the effect of watching "Wreck It Ralph" and it's certainly very impressive.  🙂

Aircoookie:
That project is quite cool, perhaps I am going to implement it as an effect! 25k of data are still quite overkill though, even with 2k of data we should be getting about 10 minutes of non-repeating colors.
WLED is overkill for any one single effect. Just take an Arduino Uno and paste the effect function into an empty sketch. It's power comes from being able to change the "project" at any time just by a phone tap or automation :)

bphick:
I had this on my agenda to ask about a while back.  Using such an effect when WLED is already running in soffits / crown molding is a great thing.  Also for "away mode" of the house.  Curtains pulled, TV room looks occupied.

fribse:
> Also for "away mode" of the house. Curtains pulled, TV room looks occupied.

Exactly my idea :-)

oz9els:
> Just curious but isn't the use of WLED a bit overkill for fake TV? 
isolated: Yes - Probarbly. If you already have your house stuffed with WLED, then it is just a added bonus :-)
So instead of having the WLED changing to a mood when you are home, it changes to the fake-tv, when you are NOT at home.. GENIUS...

Legsmaniac:
It's a fair point.
[...]

PR #1548, opened 2020-12-21
TV Simulator Effect

Def3nder:
Hi @Aircoookie,

there had been wishes for an TV-effect for a long time and I looked at various solutions with NeoPixels ...
(e.g. #938 opened by fribse) 

The most of them looked quite ugly ...
...but then I stumbled upon the version from Adafruit: https://learn.adafruit.com/fake-tv-light-for-engineers

This one was way better and I adapted it to work with WLED.

It's based on color-scans from real movies - thus it does look way more realistic :-)

...greetings from Berlin,
Ralf


Aircoookie:
This effect looks quite realistic when viewing through a window or room door indeed :)
At some places it is too dark I believe - but movies are also dark and switching between solid and TV mode once put it at a different offset that was brighter.
I did add some defines because the mode uses about 18kB of flash, which is quite much but of course we have enough space to have it enabled by default :)

poblabs:
I think this is great! Thanks for this! One question - it only displays 1 color on the segment - where real tv (and the fake tv itself) is multiple colors. Does it make sense to have some more colors among the segment?

Def3nder:
Hi @Aircoookie ,
good idea (with the define) - I thought about a kind of algorithm that simlulates the color changes without the need of the "real color data" - maybe we can have two versions: one with the big data from tv_colors.h and another with "only" calcculated ones...

Hi @poblabs ,
yes this uses the whole stripe in one color - target is to simulate a room with a TV inside - so you must not see the TV / the LED-stripe directly - this could easily be seen as "fake TV" then. 
...and: if you do not see the light-source directly, it does not matter whether you have one single color or a bunch of them - they will mix to the target color anyway.

As "color-source" I used the pictures from here: https://thecolorsofmotion.com/films
I choose 43 "interesting looking ones" from there and combinded them in one Photoshop picture (each movie one pixel-line) and then cropped a bitmap with 18 movies (500 pixel each) forr the tv_color data.

A python script does read the pixels and converts the 8-bit RGB color to a 2-byte value (5/6/5 bits for RGB).
The effect does convert the colors back and displays them with channging duration (0,25 to 2,5 sec) and with changinf fading (0 to total-duration) and inserts with an 1/3 chance a "direct cut" (no fading at all.

This is basically it. The brightness needs to be adjusted depending on how mach pixels the LED-strip has.
I've tested it with the 316-pixel strip of my son in his room and this was way too bright at 50% brightness.

poblabs:
Thanks for the explanation! I actually own the [proper fake tv](https://faketv.com/) device and can't wait to check this out - would much rather have the control WLED offers!

Def3nder:
Hi @Aircoookie ,

I looked at your define - this disables only the effect code, but does not prevent the tv_colors.h from being included.
I think we need another ifdef around the "include "tv_color.h" in the header of FX.cpp, don't we ?

beltracri:
Hello everyone, I'm not skilled in programming/compiling, is there anyone that can help me to find a .bin file witch include the TV simulator effect?
This maybe a stupid request, I hope that someone can help me!

Legsmaniac:
@beltracri The TV Sim Effect is now built in.

beltracri:
Can't find it in pre-compiled release

beltracri:
> @beltracri The TV Sim Effect is now built in.
Can't find it in pre compiled .bin release!

Legsmaniac:
@beltracri I'm sure Aircookie posted a bin somewhere but then again, it depends what board you want it for.

beltracri:
> @beltracri I'm sure Aircookie posted a bin somewhere but then again, it depends what board you want it for.

D1 mini, ok, I'll try to search

Legsmaniac:
@beltracri Just popped one onto WeTransfer for you - https://we.tl/t-81hBmKXj85

beltracri:
> @beltracri Just popped one onto WeTransfer for you - https://we.tl/t-81hBmKXj85

Really great!!! Many thanks!

Legsmaniac:
@beltracri You're welcome.
Just realised that was the Dev version (alpha) I sent you.  It shouldn't be a problem, it does work.
Just in case, I have uploaded the latest Master V12 Beta for you as well - https://we.tl/t-LfcSVgOY6h
Any problems, just holler.
Let us know how you get on.

beltracri:
> @beltracri You're welcome.
> Just realised that was the Dev version (alpha) I sent you. It shouldn't be a problem, it does work.
> Just in case, I have uploaded the latest Master V12 Beta for you as well - https://we.tl/t-LfcSVgOY6h
> Any problems, just holler.
> Let us know how you get on.

Many thanks...
I actually run 11.0, could I update via manual OTA without losing HA and preset settings?

Legsmaniac:
You should be able to, yes.
I certainly did.

beltracri:
> You should be able to, yes.
> I certainly did.

I did it... Works like a charm!!!

Legsmaniac:
@beltracri Excellent!  😁  👍

Issue #1114, opened 2020-08-25
Dynamic fade effect


benblb:
**tl;dr is bold below**

I went through all of the effects and couldn't seem to figure out how to do the following, but apologies if it's possible today.

Essentially I would like to see a way to have each group in a strip that uses the grouping feature (i.e. 100 LEDs on the strip in 10 groups of 10 pixels) fade between random RGB colors.

**The "Dynamic" effect would be perfect since it seems each section is a random color, but it jumps from color to color rather than fading. What I'm looking for is no predictable pattern. Essentially totally randomize each color and fade from one to the next.**

If I'm being really needy it would also be cool to have a dynamic fade effect where the different pixels change at different times.

I am currently in the process of building a Nanoleaf clone using WLED and this is my ideal go-to look since each light panel would appear to be behaving on its own. Thanks for reading and considering, and I'm a huge fan of your platform!

benblb:
My mistake, dynamic already has randomized times and is perfect! I downgrade my full ask to a fade of exactly how dynamic is today. 

PS thanks for getting eyes on this so quickly!

benblb:
It's been a while since I've done C++, but I took a stab at making dynamic fade which I pasted below. I hacked together the fade effect and the dynamic effect, so hopefully I didn't mess it up too badly! 

I didn't update the .h file (I assume this is just one line) or the UI. Not sure if the UI is automatically updated or not. 

`/*
 * Lights every LED in a random color. Changes all LED at the same time
// * to new random colors.
 */
uint16_t WS2812FX::mode_dynamic_fade(void) {
  if (!SEGENV.allocateData(SEGLEN)) return mode_static(); //allocation failed

  if(SEGENV.call == 0) {
	for (uint16_t i = 0; i < SEGLEN; i++) SEGENV.data[i] = random8();
  }

  uint32_t cycleTime = 50 + (255 - SEGMENT.speed)*15;
  uint32_t it = now / cycleTime;
  uint16_t fadedur = (cycleTime * SEGMENT.intensity) >> 8;
  
  uint32_t fade = 255;
  if (fadedur) {
    fade = (rem * 255) / fadedur;
    if (fade > 255) fade = 255;
  }

  if (it != SEGENV.step && SEGMENT.speed != 0) //new color
  {
	for (uint16_t i = 0; i < SEGLEN; i++) {
		SEGENV.aux[i] = SEGENV.data[i];
[...]

PR #1550, opened 2020-12-21
Dynamic fade effect

Def3nder:
Hi @Aircoookie ,

I worked on the Projects section and implemented the "Dynammic Fade Effect" based on the original one as "base" and then a "normal" and a "smooth" version that calles the base.

This should close issue #1114 

As this adds another effect item, this only to be added "directly" when PR #1548 is committed.
...hmmm - just looking in the PR files below - it seems as this PR does include the TV Simulator, too 😢 

Anyway - I'm saving this - let me know (e.g. on Discord) if I can change the PR to another base...

many greetings from Berlin,
Ralf


Aircoookie:
Hi!
No worries - I don't even think it is possible/good to make 2 pull requests adding two effects - they would both add to the end of the file and thus result in a merge conflict.
Let's merge in the TV simulator first and then just fix this PR afterwards :)

Greetings from Hamm and merry Christmas!

Aircoookie:
This looks super cool! Thank you :)

benblb:
I'm so excited to see this effect get added! Merry Christmas to you both and thanks so much for continuing to improve WLED! Thought you might like to know that I'm actually giving WLED as a gift in the form of a light-up cloud on Friday. Can't wait for the next update. 

Warm wishes from California, cheers!

PR #1552, opened 2020-12-21
Random Background image

WoodyLetsCode:
Hey,
this adds an option to UI Settings to enable random background images.
A new background image will be generated after every reload.

grafik


Aircoookie:
Why not 😄 Those random images from the link are quite nice!

PR #1583, opened 2020-12-27
Add support for WT32-ETH01 ethernet board and make ethernet support configurable

lonestriker:
I've added support for the cheap WT32-ETH01 board as well as making the ethernet support itself configurable.  It should now be possible to build in general ethernet support to the base ESP32 image, without needing a special build just for ethernet boards.  You just need to access WLED-AP after flashing and select the ethernet board you're using.  The default is 'None' which disables ethernet support.

I've tested on both my ESP32-POE as well as the WT32-ETH01 boards and things behave as expected.  Please let me know if you have any suggestions for changes or have any questions.

Thanks!


lonestriker:
@Aircoookie I've updated the PR to include @k7bbr's  the BTNPIN change that fixes reported issues with the WT32-ETH01 boards. When you have a moment, can you take a look at the PR and let me know if there are any suggestions or changes needed to be merged into your mainline?

Aircoookie:
Hi!
Sorry, I completely missed this before I made my proposal of adding a new environment! Thank you for making this :)
I'm not yet sure if Ethernet is a good option to have configurable at runtime though - one issue is that the ETH library uses quite a bit of extra RAM compared to WiFi only (we could afford those 30k though if need be). Secondly, having the Ethernet setting on boards that don't support it could lead to confusion. As an alternative, we could make binaries with Ethernet compiled in available in future releases. This would make a lot of sense once LED type + pin is runtime selectable (as that is not dependant on the board type itself, unlike the ethernet hardware)

lonestriker:
The build right now limited to the _poe target.  That target supports both ESP32-POE and WT32-ETH01 now (or None as well, but you can grab the normal build if you want Wi-Fi only.)  My understanding is that the ESP32 has a lot of flash and PSRAM in general, so an extra 30KB shouldn't really be too onerous if you were to make a single Wi-Fi + Ethernet target.

I might suggest that we change the _poe target though and rename it to _eth if you decide to merge this PR in.

k7bbr:
I like the idea to rename the target to _eth if the PR is merged. As @lonestriker Having a single WiFi + Ethernet build would be nice in the future if the RAM overhead isn't too much. If not, the idea of a compiled binary, as @Aircoookie mentioned, in future releases would be fine for now too. If that was the route taken, it wouldn't be too hard to point Ethernet board users to the _eth binary if they wanted to use the board.

Aircoookie:
Apologies, I didn't realize that it already compiled in only on the _poe target! (also in favor of renaming to esp32_eth)
I think I'll add a little JS that hides the settings option itself if ethernet is not compiled in and then I can merge :)

Aircoookie:
Ok, made a few little adjustments :)
- Prevent compile error if "WLED_USE_ETHERNET" not defined (which disables the `ethernetType` setting, but did not do so in xml.cpp and set.cpp)
- `ethernetType` defaults to ESP32_POE instead of None. If ethernet is compiled in, that pretty much guarantees it is used on an Ethernet board (at least for now), so for ESP32_POE users ethernet will work without having to connect to WLED-AP just to enable ethernet.
- As mentioned before, hide the setting if ethernet is not compiled in.

Give it a test if you'd like, will merge tomorrow!

k7bbr:
Thanks! I'll give it a test tonight.

lonestriker:
@Aircoookie works fine for me on ESP32-POE and WT32-EHT01 boards

PR #1589, opened 2020-12-28
Add Aurora Borealis effect

Mazn1191:
Added an aurora borealis effect based on my Arduino project: https://github.com/Mazn1191/Arduino-Borealis
I also added a second aurora palette since the original one doesn't quite felt right.


Mazn1191:
I noticed I handle memory very poorly. Will need to do some rework.

Aircoookie:
Thank you for the PR :)
Yes, dynamically reallocating memory all the time isn't so good.
If you'd like, you can request a certain amount of memory for your effect using SEGENV.data (in this case that would be 20*the wave class size) and cast that pointer to your class.
Take a look at the `mode_starbust` or `mode_dynamic` effects to see how to use it :)

Mazn1191:
Thank you for the response :)
It took me a little longer to fully understand the concept and get an overview of all the functionality offered.
But I think now I've got it. I think I'll only allocate memory needed for the current amount of waves currently set by the slider. Always hogging memory for 20 waves seems a bit selfish :D 20 waves should need about 640byte of memory.

Mazn1191:
Memory issues should be sorted out now ;)

obar:
I was so impressed with the Aurora effect that I had to look up this pull request to share a compliment. I love how this blends in so nicely when changing between color schemes. Well done @Mazn1191!

PR #1609, opened 2021-01-01
Re-added countdown setup

SamPUG:
I think that in the transition to FS storage, the countdown timer setup has been forgotten. On the current version, when the countdown is set, `countdownTime` is not set, and remains at the default. 

`setCountdown()` used to be called in `saveSettingsToEEPROM()` so I thought that the `serializeConfig()` function was the best place to re-add the functionality.


Aircoookie:
Hi! Cool to see that I am not the only one to use the countdown feature for new years and thank you for contributing the fix!
I had already fixed it as well in https://github.com/Aircoookie/WLED/commit/f5ed710c0b731076afdd1c5f0d75558f9faac222
Sorry for not finding the issue earlier :)

Issue #1605, opened 2020-12-31
Need ALEXA and MQTT in usermod [asap ;-)]


gsieben:
Now I'm stuck with my project a little bit. I have made an integration for relays in the webfrontend/APP and JSON. Now I would like to do the same for ALEXA and MQTT. But now I have a problem. Because there is no user mod option for ALEXA like there is for JSON (e.g. void addToJsonState(JsonObject& root)) and Config.

The relays are configured in the setup of the webfrontend. The relays can be added flexibly without having to recompile. But since I don't know the number of relays, I can't use alexa.cpp as a workaround. 

If you are interested, you can have a look at the current state in my fork. It is up to date with wled/master. You can also quickly see what else would be quite useful for usermods. I wrote as little as possible in the main code. Nevertheless there are numerous interventions (see the readme.md) of my usermod. 

https://github.com/gsieben/WLED

Cheers Gabriel (And a happy new year)

![Webpage](https://user-images.githubusercontent.com/2237858/103388757-2f95c500-4b0b-11eb-8bd0-4ae3295ced6d.png)
![Webpage2](https://user-images.githubusercontent.com/2237858/103388882-b185ee00-4b0b-11eb-8d01-7072595cc4eb.png)
![Webpage3](https://user-images.githubusercontent.com/2237858/103388761-30c6f200-4b0b-11eb-88fc-5bd953ca6db2.png)
![Webpage4](https://user-images.githubusercontent.com/2237858/103388762-315f8880-4b0b-11eb-8b00-2bd590a74d4b.png)
![Webpage5](https://user-images.githubusercontent.com/2237858/103388763-315f8880-4b0b-11eb-80a4-5790d6acbb31.png)

PR #1614, opened 2021-01-03
Relay Control Webpage/APP HTML API JSON API

gsieben:
Since I absolutely wanted to have a control for relays, I programmed this without further delay. At first I wanted to program the whole LED control. I already have some finished projects that I just wanted to put together. But then I came across this wonderful WLED. So why should I do the work. The project can do more than my other projects combined. Thanks for this @Aircoookie.

But the integration via Alexa or MQTT which already exist was not enough for me. I also wanted to control the Relais via Webpage/App and with JSON. As flexible as possible. But this is an intermediate state. Because the Alexa and MQTT integration is not yet done. That still follows. 

Unfortunately, I had to intervene in some places in the main code. There is more of a manual patch than just a usermod, so to speak. The pullrequest targets the user mod: `GeoGab-Relays`, not more. The Alexa and MQTT integration I catch up when the integration is better supported. I want to avoid making such a complicated code as done in usermod  mqtt_switch_v2 and I want to avoid interfering even more with the main code. Therefore enhancement request #1605. 

Greetings Gabriel

Pleas find more informations at: https://github.com/gsieben/WLED/tree/master/usermods/GeoGab-Relays


blazoncek:
Hi, I have recently started implementing user configurable usermods and have (among other things) created a multi-relay usermod that serves a similar purpose as yours.
The difference being that relays in my usermod cannot be switched from outside (only WLED's on/off controls them) and yours being controlled from outside only (and no WLED integration), but I can see it beneficial to add those two functionalities together.
Since you implemented settings_relay.htm to do both configuring and changing this needs a _hack_ into existing core WLED which serves _only_ this usermod, which I think is not best approach.
I propose that we combine our usermods (no need for separate settings page from mine and JSON changing state from yours) to make universal multi-relay usermod that can be configured as a user wants during runtime.
What do you think?

PR #1617, opened 2021-01-03
Allow the temperature sensor pin to be configured

jeremypoulter:
I have a custom board that I needed to configure the pin to a different one than the default so this change allows the the pin to be configured by setting `USERMOD_DALLASTEMPERATURE_PIN`.

### DIFF
--- a/usermods/Temperature/usermod_temperature.h
+++ b/usermods/Temperature/usermod_temperature.h
@@ -4,11 +4,15 @@
 
 #include <DallasTemperature.h> //DS18B20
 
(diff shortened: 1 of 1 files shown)


blazoncek:
The latest apha version in my fork dev branch has the ability to change the usermod parameters on the fly during runtime.
It is expected to be available in next release.
You are welcome to give it a try and report back with any issues you may encounter.

jeremypoulter:
Can confirm this PR is no longer needed

PR #1637, opened 2021-01-11
Search bar and palettes preview

Moustachauve:
This changes the interface for palettes and effects selection.
For both of them, there is now a search bar at the top to find the item of your choice.
For palettes, there is now also a color preview under the name of the palette.

Here is what the palettes look like:

Here is a gif of the search bar in action:


Moustachauve:
I'm still working on this. For example, I still need to add support for the fastled palettes that are not defined in `const byte* const gGradientPalettes[]`

Moustachauve:
I think this is as ready as I can make it!

andyshinn:
Adding my feedback from Discord here as well:

* The palette list doesn't show colors until I click. Bug?
* Color palette icon maybe could be aligned with text?
* Previous effect list pinned "Solid" up top to make it easily accessible. Do we want to keep this?
* The chosen effect isn't quickly viewed without scrolling. Do we want to pin it as well?

Moustachauve:
Hey @andyshinn. as discussed on discord, the first point "The palette list doesn't show colors until I click. Bug?" This only happens in the merge of the multistrip branch. It should not happen in this PR.

All the other points are good feedback. I'll implement the solid button shortcut and the color palette alignment. 

The palette pinning will be part of a future improvement

Moustachauve:
All bugs should be fixed now and all suggestions implemented!

Moustachauve:
I changed the checkboxes to be radio buttons

PR #1653, opened 2021-01-13
Added Window Matrix Effect

CBaker007:
Created a new Window Matrix (15 wide x 20 tall) effect with choices for Christmas (tree, candy cane, star, snowflake) and upcoming Valentines Day (heart). (Depending on space, would create more over time for all holidays, sport teams, etc.) 

The idea was to create a year round matrix for a window that can be seen from a distance while keeping the matrix cost down.   While high density matrix mats are available, to make them large enough would require several mats costing over $100/window with much of the matrix resolution being wasted.  Using 2 of the 30 pixels/meter strips or 300 pixels, you can create a 18" (45.7mm) x 25" (63.5mm) for under $40/window.  

I wanted to avoid running a full scale presentation software to drive the matrix. I wanted to have a small, self contained microprocessor that I could place at the matrix that had a web interface that would allow me to modify the displayed image.  WLED seemed like the perfect solution.

User Interface / Options: Images are selected using the Segment Intensity slider.  Most images use Primary colors 1 and 2 (segcolor 0 and 1) for some foreground color choices and color 3 (segcolor 2) for background color.  Using this, some animation is possible by changing some background and or accent colors.  Note: Segment must be at least 300 pixels or the first few pixels are set to red indicating an error. 

**Possible Upgrades/Issues:**  
* There seems to be a push for space optimization. Each image choice is currently 300 bytes (1 byte/pixel) of flash.  Only need 1/2 a byte for each pixel to store the color code.  This could be optimized by some form of encoding the pixels or maybe to somehow create a 4 bit unsigned int (not sure if possible).   Current config allows you to visualize the image in the array.  Encoding would make it much less intuitive.   
* An ideal solution would allow users to create their own array and upload it to the flash (or load from web) without requiring you to compile the code. This is beyond my current coding ability.
* Another space saving solution, if necessary, might be to add a build flag creating an "optional build" for matrices and requiring a larger memory microprocessor board.
* I would like to use pallets more and create animation on background pixels like ever changing background colors.
* It would be interesting to use color codes 8 and 9 to indicate pixel turning on / off of primary 1 and 2.
* It would be interesting to add an animation affect to rotate the image using formulas at 22.5 degree increments.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -4021,3 +4021,284 @@ uint16_t WS2812FX::mode_aurora(void) {
   
   return FRAMETIME;
 }
+
+/*
+ * Custom FX - Window Matrix
+ * by Christopher Baker
+ * 12/24/2020 - v1.0
+ *
+ * Called from mode_custom_matrix15 to display actual array on maxtrix
+ * Passed the color array for the display array, Display array, and the
+ * last argument = pixel array processing order: 0 for sequencial, 1 for visual 
+*/ 
+uint16_t WS2812FX::custom_matrix15(uint32_t pix_color[], uint8_t pix_img[], uint8_t seq_order) {
+
+  /* 15x20 = 300 pixels */
+  if (SEGLEN >= 300) {
+    /* 300 pixels but index starts at 0, so 0-299 */
+    if (seq_order == 1) {
+      // Select Visual Order (Img Array Arranged Visualy 0-299 = L to R / T to B)
+      bool a_row_odd = 0;
+      uint16_t a_pix = 0;
+      uint16_t p_pix = 0;
+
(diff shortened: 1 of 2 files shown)


embedded-creations:
You might be interested in the 2D support in WLED Soundreactive, or the 2D/3D ideas being discussed and prototyped in the #2d-3d-mapping channel on Discord: https://discord.gg/RjDJuhgT

Also, check out [bitbank2's Arduino libraries](https://github.com/bitbank2) for image storage, there's options for JPG, and AnimatedGIF (I'm not sure how applicable AnimatedGIF storage is to a static image, but I know it has compression).  I plan to port the AnimatedGIF library into WLED as a usermod after 2D support is further along, and I'll post about it in that Discord channel.

PR #1683, opened 2021-01-22
Auto node discovery.

blazoncek:
Automatic WLED node discovery so it is easy to jump from one WLED node to another using web GUI.

Sample (2 nodes discovered with button to jump to their GUI)
Screenshot 2021-01-23 at 20 36 22


Aircoookie:
Thank you for making a separate tab for the Node discovery. I'm putting on some finishing touches and will finally merge :)

Issue #1686, opened 2021-01-23
NpbWrapper.h throws lots of warnings (treated as errors) unless suppressed


embedded-creations:
Unless `-Werror=reorder` is added to `build_unflags` in PlatformIO, WLED won't compile.  WLED won't compile with compiler warnings set above default (due to this issue and other issues).

Sample of errors, WLED master [0266370]as of 2021-01-23 with `-w` build flag removed:

```
In file included from wled00/FX.h:36:0,
o/packages/framework-arduinoespressif32/tools/sdk/includ                 from wled00/fcn_declare.h:98,
e/s                 from wled00/wled.h:113,
martconfig                 from wled00/FX.h:28,
_                 from wled00/FX_fcn.cpp:27:
wled00/NpbWrapper.h: In constructor 'NeoPixelWrapper::NeoPixelWrapper()':
awled00/NpbWrapper.h:427:53: error: 'NeoPixelWrapper::_pGrbw' will be initialized after [-Werror=reorder]
ck -I/Use   NeoPixelBrightnessBus<PIXELFEATURE4,PIXELMETHOD>* _pGrbw;
                                                     ^
rs/louis/.platformio/pawled00/NpbWrapper.h:423:16: error:   'NeoPixelType NeoPixelWrapper::_type' [-Werror=reorder]
ckages/framework-ardu   NeoPixelType _type;
                ^
wled00/NpbWrapper.h:191:3: error:   when initialized here [-Werror=reorder]
inoe   NeoPixelWrapper() :
   ^
In file included from wled00/FX.h:36:0,
s                 from wled00/fcn_declare.h:98,
                 from wled00/wled.h:113,
                 from wled00/FX.h:28,
                 from wled00/FX.cpp:27:
pwled00/NpbWrapper.h: In constructor 'NeoPixelWrapper::NeoPixelWrapper()':
rwled00/NpbWrapper.h:427:53: error: 'NeoPixelWrapper::_pGrbw' will be initialized after [-Werror=reorder]
   NeoPixelBrightnessBus<PIXELFEATURE4,PIXELMETHOD>* _pGrbw;
                                                     ^
wled00/NpbWrapper.h:423:16: error:   'NeoPixelType NeoPixelWrapper::_type' [-Werror=reorder]
essif32/tools/sdk/in   NeoPixelType _type;
clude/soc -I/Users/louis/.platformio/packages/framework-                ^
[...]

PR #1687, opened 2021-01-23
Reenable compiler warnings for ESP32 platform on PlatformIO

embedded-creations:
- remove "-w" ("Suppress all warnings, including those which GNU CPP issues by default.") from esp32.build_flags
- change build_unflags to suppress errors in wled00/NpbWrapper.h
- only define CONFIG_LITTLEFS_FOR_IDF_3_2 if not already defined to avoid another error

I opened a separate issue #1686 to cover fixing the errors in NpbWrapper.h, so hopefully `-Werror=reorder` can be removed in the future

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -122,7 +122,7 @@ build_flags =
 
 build_unflags =
   -Wall
-  -Wreorder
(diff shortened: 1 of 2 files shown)


Aircoookie:
Superseded by #1744?

PR #1689, opened 2021-01-23
DHT temperature and humidity sensor usermod

fernandoeng:
### DHT temperature and humidity sensor usermod

I've added a usermod for ESP32 using the DHT11 based on the Dallas one just changing a fell things, it might be good to help others to be able to get DHT sensors integrated with WLED.

It adds temperature and humidity to the info and send mqtt data to the topics "*/temperature" and "*/humidity"

it's tested with my DHT11 using a TTGO Esp32 and ADA102 rgb led strip.
Also tested with a D1 mini with DHT11 shield connected to the pin D4 and ADA102 rgb led strip.


fernandoeng:
There is a new pull request that may do the same #1719

fernandoeng:
#1719 got merged already

PR #1719, opened 2021-02-02
DHT22/DHT11 humidity/temperature sensor usermod

b3nj1:
Hi,

Please consider this PR which adds support for a DHT22 sensor. I welcome any inputs you have on the PR.

The sensor is read using the non-blocking DHT library. I have tested it with an ESP8266. I have an ESP32 on order, so will test and update if needed when that arrives.

I put my power supply and QuinLED-Dig-Uno in an outdoor enclosure. After a couple of weeks, I opened it up to find condensation had pooled up in the enclosure. Luckily, the pooling was away from any components, and everything seems fine. I'm upgrading my enclosure to have breathers which I hope will address the condensation issue. I'm adding the DHT sensors to monitor enclosure humidity.

Thanks,
Benjamin

### DIFF
--- a/usermods/DHT/platformio_override.ini
+++ b/usermods/DHT/platformio_override.ini
@@ -6,7 +6,6 @@
 ; USERMOD_DHT_CELSIUS              - define this to report temperatures in degrees celsious, otherwise fahrenheit will be reported
 ; USERMOD_DHT_MEASUREMENT_INTERVAL - the number of milliseconds between measurements, defaults to 60 seconds
 ; USERMOD_DHT_FIRST_MEASUREMENT_AT - the number of milliseconds after boot to take first measurement, defaults to 20 seconds
-; USERMOD_DHT_FIRST_MEASUREMENT_AT - the number of milliseconds after boot to take first measurement, defaults to 20 seconds
 
 [env:d1_mini_usermod_dht_C]
 extends = env:d1_mini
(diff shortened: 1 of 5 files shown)


fernandoeng:
Nice :D

b3nj1:
I received my ESP32 today, so I was able to test and update for supporting that. The version of esp32DHT that platformIO picked up by default crashed. The newest API better matches the ESP8266 version also. So, I'm pointing directly to the git repo, https://github.com/bertmelis/esp32DHT .

- Benjamin

b3nj1:
The esp32DHT library failed to collect 90+% of samples, so I rewrote to use https://github.com/alwynallan/DHT_nonblocking . It seems to work very reliably now on both ESP8266 and ESP32 boards.

I'm done tinkering :)

- Benjamin

A7F:
@b3nj1 can you please provide us a demo setting for let's say a dht11 that reporte in celsius? Asking because I tried to setup mine but it did not work. I copied the override.ini file in the same directory as platformio.ini file but I have no idea on what else to do apart from editing the override file. 
 
Also, what happens when the usermod can't find the sensor? I mean, what exactly is a "disabled" usermod?

b3nj1:
@A7F , sorry for the delay. I'm trying to get the usermod working on the latest (0.13-b2), but I'm not getting any temp readings yet.

The usermod's are compiled in, so you have to change the default_envs in platform.ini to include one of the envs in the override. For example:

`default_envs = nodemcuv2, esp01_1m_full, esp32dev, esp32_eth, custom32_LEDPIN_16_usermod_dht_C`

However, the base custom32_LEDPIN_16 isn't present anymore, so I'll have to push an update if/once I get things working.

PR #1721, opened 2021-02-04
Ability to lookup Usermod by id so Usermods can use other Usermods

kdorff:
This allows a Usermod to get a pointer to another Usermod, if it is found. One Usermod can then call functions in the other Usermod, if it found, to provide additional functionality.

I couldn't immediately find a way to do this, so I wrote my own. If there is a better way to do this, please let me know.

I have written a v2 Four Line Display Usermod (based on the ssd1306 v1 usermod) which includes a clock during input timeout. I have also written a v2 Rotary Encoder UI Usermod (based on the existing v1 rotary encoder usermods, but much more capable, supporting brightness, effect, effect speed, effect intensity, and palette). Both will happily work stand-alone, but if the Rotary Encoder UI Usermod finds the Four Line Display Usermod, it can provide additional functionality. 

Both usermod submissions coming after this is merged.

### DIFF
--- a/wled00/fcn_declare.h
+++ b/wled00/fcn_declare.h
@@ -221,6 +221,7 @@ class UsermodManager {
     void readFromConfig(JsonObject& obj);
 
     bool add(Usermod* um);
+    Usermod* lookup(uint16_t mod_id);
     byte getModCount();
 };
 
--- a/wled00/um_manager.cpp
+++ b/wled00/um_manager.cpp
@@ -15,6 +15,18 @@ void UsermodManager::readFromJsonState(JsonObject& obj) { for (byte i = 0; i < n
 void UsermodManager::addToConfig(JsonObject& obj)       { for (byte i = 0; i < numMods; i++) ums[i]->addToConfig(obj); }
 void UsermodManager::readFromConfig(JsonObject& obj)    { for (byte i = 0; i < numMods; i++) ums[i]->readFromConfig(obj); }
 
+/*
+ * Enables usermods to lookup another Usermod.
+ */
+Usermod* UsermodManager::lookup(uint16_t mod_id) {
+  for (byte i = 0; i < numMods; i++) {
+    if (ums[i]->getId() == mod_id) {
+      return ums[i];
+    }
+  }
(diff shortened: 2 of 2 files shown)


embedded-creations:
Seems useful, but can you share a snippet of code where one Usermod is actually interacting with the other?

Aircoookie:
I agree, this could lay the groundwork for some usermods expanding upon others! Thank you for the PR!

To be useful, a usermod would need to `#include` the "dependency usermod" as well as type-cast the usermod pointer returned by `lookup()` to the actual child class implementation in order to access its functions and variables, since the ones provided by the Usermod parent class are not really useful for expanding functionality. `lookup()` is also helpful for finding out if a given usermod is installed in the first place though, so I believe I will merge this soon :)

kevindorff:
I imagined people would want to see it in action, so I have a draft PR (which I've updated this morning, but I'm still cleaning and optimizing) that demonstrates my RotaryEncoderUIUsermod **optionally** making use of my FourLineDisplayUsermod.  I think this is the most complex case - generally speaking, if you want two Usermods to work in concert you probably want to always have them. I wanted my RotaryEncoderUI to be easy to modify and extend to meet personal needs. Additionally I wanted the display to be integral but completely optional. 

See `usermod_rotary_encoder_ui.h` within (updated again this morning) in my Draft PR:

    https://github.com/Aircoookie/WLED/pull/1722

Important lines
* `38: #define USE_FOUR_LINE_DISPLAY` controls if FourLineDisplayUsermod will be available in the code (`#include`d within `usermods_list.cpp`)
* `91-95` define the display object in the case where we USE_FOUR_LINE_DISPLAY or not This is where the casting from `Usermod*` to `FourLineDisplayUermod*` occurs (so we can make calls into it)
* Elsewhere in the code it's OK to check `display` against `nullptr` but any code that calls methods on the display class needs to be gated by `#ifdef USE_FOUR_LINE_DISPLAY`. This is only because I wanted to make the display Usermod optional. Some people may just want the encoder functionality.

Aircoookie:
Merged with #1722 !

PR #1722, opened 2021-02-05
Rotary Encoder, Four Line Display, and Auto Save Usermods

kdorff:
My Usermods to provide extensive rotary encoder UI support (`RotaryEncoderUIUsermod`) as well as optionally working in concert with a 128x32 or 128x64 OLED display (`FourLineDisplayUsermod`). `RotaryEncoderUIUsermod` relies on an extension to um_manager.cpp which has it's own PR (but is included here because the other isn't merged).

Additionally I've added AutoSaveUsermod. This will apply the AUTOSAVE_PRESET_NUM preset in first loop. It will then start watching for changes to `brightness`, `effect speed`, `effect intensity`, `mode (effect)`, and `palette`. If a change happens, it will wait AUTOSAVE_SETTLE_MS seconds for additional changes to come in (the window will extend with each change).  Then it will save changes with a preset name of "Auto save MM-DD HH:MM:SS" (date and time) to preset number AUTOSAVE_PRESET_NUM.

AutoSaveUsermod is standalone, but if FourLineDisplayUsermod is installed, it will notify the user of the saved changes.

Note: I don't love that WLED doesn't respect the brightness of the preset being auto loaded, so the AutoSaveUsermod will set the AUTOSAVE_PRESET_NUM preset in the first loop, so brightness IS honored. This means WLED will effectively ignore `Default brightness` and `Apply N preset at boot` when the AutoSaveUsermod is installed.


kdorff:
After reviewing other usermods and updating these to be in line with how the Temperature usermod worked (which seemed like a great example) I'm thinking this is just about finished for initial release.

Opinions, comments, and suggestions are welcome.

kdorff:
Some additional fixes and addition of AutoSaveUsermod to automatically save settings if settings change, after a quiet period of no settings changes.

My only question is if it loads "preset 1" by default (an option in LED) and preset 1 defines a brightness, that brightness is ignored because startup brightness seems to be non-optional.

PR #1724, opened 2021-02-05
Playlist handling.

blazoncek:
Memory allocation fix, empty playlist handling, randomizing playlist, optimizations.

Current implementation of playlist has no ability to stop an endless playlist and upon submitting empty one sometimes crashes ESP. I fixed both issues and as a bonus added a randomization of playlist if it is an "endless" playlist with repeats==-1.

### DIFF
--- a/wled00/json.cpp
+++ b/wled00/json.cpp
@@ -279,7 +279,7 @@ bool deserializeState(JsonObject root)
 
   JsonObject playlist = root[F("playlist")];
   if (!playlist.isNull()) {
-    loadPlaylist(playlist); return stateResponse;
+    loadPlaylist(playlist); //return stateResponse;
   }
 
   colorUpdated(noNotification ? NOTIFIER_CALL_MODE_NO_NOTIFY : NOTIFIER_CALL_MODE_DIRECT_CHANGE);
--- a/wled00/json.cpp
+++ b/wled00/json.cpp
@@ -279,7 +279,8 @@ bool deserializeState(JsonObject root)
 
   JsonObject playlist = root[F("playlist")];
   if (!playlist.isNull()) {
-    loadPlaylist(playlist); //return stateResponse;
+    loadPlaylist(playlist);
+    noNotification = true; //do not notify both for this request and the first playlist entry
   }
 
   colorUpdated(noNotification ? NOTIFIER_CALL_MODE_NO_NOTIFY : NOTIFIER_CALL_MODE_DIRECT_CHANGE);
--- a/wled00/playlist.cpp
+++ b/wled00/playlist.cpp
@@ -74,9 +74,19 @@ void serializePlaylist() {
 }
 */
 
+void unloadPlaylist() {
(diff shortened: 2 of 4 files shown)


wled00/json.cpp
```diff
@@ -279,7 +279,7 @@ bool deserializeState(JsonObject root)
 
   JsonObject playlist = root[F("playlist")];
   if (!playlist.isNull()) {
-    loadPlaylist(playlist); return stateResponse;
+    loadPlaylist(playlist); //return stateResponse;
```
pierrebeitz, line 282: is this change intentional? if so: what does it do?

wled00/playlist.cpp
```diff
@@ -10,19 +10,73 @@ typedef struct PlaylistEntry {
...
+void deserializePlaylist() {
+  DynamicJsonDocument doc(JSON_BUFFER_SIZE);
+
+  DEBUG_PRINTLN(F("Reading playlist from /playlist.json..."));
+
+  if (!readObjectFromFile("/playlist.json", nullptr, &doc)) return; //if file does not exist just exit
+
+  JsonObject playlist = doc[F("playlist")];
+  if (!playlist.isNull()) loadPlaylist(playlist);
+}
```
pierrebeitz, line 49: is this lengthy comment needed? if so: why do we want it to be here?

wled00/playlist.cpp
```diff
@@ -72,26 +126,30 @@ void loadPlaylist(JsonObject playlistObj) {
   currentPlaylist = 0; //TODO here we need the preset ID where the playlist is saved
 }
 
-void handlePlaylist()
-{
+
+void handlePlaylist() {
```
pierrebeitz, line 130: i think it would be nice to stay with the prevalent code style of this repository. by that i mean leaving the curly braces where they've been. 
that also makes it easier to review this PR as there'd be less changes. :)

pierrebeitz (commented):
thanks for investing your time here! 
in order to be respectful of aircookie's time i think it would be nice if you could address my comments and add a bit to the description of the PR in order to explaing the motivation behind it.

NB: i have absolutely no authority in this repository and most of my comments are more like "linter-comments". I just opened this and felt that i could contribute to WLED by reviewing a bit.

wled00/json.cpp
blazoncek: It is, otherwise WLED is not notified by the playlist change and behaves "strange".
For example: if "on":false is specified in json, the power button changes to off but the effect keeps playing.

wled00/playlist.cpp
blazoncek: It is a code remnant if you'd want to have loading and saving of playlists to FS. It can be deleted.

wled00/playlist.cpp
blazoncek: Got that. But the code has all sorts of indentations and curly braces positions so I used the one that is more natural to me.

PR #1738, opened 2021-02-12
Dynamic LED map creation from JSON file

blazoncek:
/ledmap.json in format {"map":[4,3,2,1,...]}.

Used for rearranging LEDs (matrices, awkward placement, ...)

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -24,6 +24,8 @@
   Modified for WLED
 */
 
+#include "wled.h"
+
 #ifndef WS2812FX_h
 #define WS2812FX_h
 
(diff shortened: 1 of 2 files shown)


JakeShirley:
@blazoncek I was working on a recent change to allow me to "ignore" one of my LEDs (I over-extended one strip so a single LED in the middle of my whole run is occluded).  This change should allow for that right?

blazoncek:
> 
> @blazoncek I was working on a recent change to allow me to "ignore" one of my LEDs (I over-extended one strip so a single LED in the middle of my whole run is occluded). This change should allow for that right?

Not sure if I understand your issue correctly, but this functionality is not meant for "ignoring" pixels but for rearranging them on physical layer.

JakeShirley:
Hypothetically you could omit an LED in the index list right?

bazzer2:
I've just found this, looks to be just what I need!
I created /ledmap.json with the defaults from the example in FX_fcn.cpp (the second example doesn't have the closing wiggly brace) but I cannot get wled to read the map at all. 

Am I missing something more obvious?

Legsmaniac:
You could omit a mid-LED using Segments.
Say you had 64 LEDs,  -
Set Segment 0 Start led 0 Stop led 32
Set Segment 1 Start led 33 Stop led 64
The 33rd LED will then remain off.

PR #1741, opened 2021-02-13
Create the sorted index of modes and palettes at startup

kdorff:
Sort modes and palettes upon startup with the new ModeSortUsermod so I don't have to manually maintain the static index array for the RotaryEncoderUIUsermod. Can easily be used by other usermods.

Also provides a list of the mode and palette names within JSON_* for direct access to lookup mode and palette names instead of having scan through the JSON string every time.

Updated FourLineDisplayUsermod and RotaryEncoderUIUsermod to use ModeSortUsermod.


kdorff:
Rebased.

PR #1742, opened 2021-02-13
IR: change to non-linear brightness steps

WJCFerguson:
Currently, brightness steps are mostly ±10, which doesn't make a whole lot of sense as the jump from 5 to 15 is profound, but the jump from 200 to 210 is barely detectable.  

This change makes the steps a geometric progression, so each successive brightness is a multiple of the previous (1.28×, as it happens, which makes for 16 steps from 5..255).  I think this geometric progression fits better with how humans perceive brightness changes.

I'm not sure if this has already been considered, and/or there would be resistance to changing the IR remote brightness-adjustment experience that users are already used to... but it's here for your consideration anyway.

While it could be done dynamically as a relative jump (incrementing by bri/4 and decrementing bri/5 works quite well), it makes for irregular behavior when moving back and forth or after clipping at the ends.

### DIFF
--- a/wled00/fcn_declare.h
+++ b/wled00/fcn_declare.h
@@ -76,7 +76,6 @@ bool decodeIRCustom(uint32_t code);
 void applyRepeatActions();
 void relativeChange(byte* property, int8_t amount, byte lowerBoundary = 0, byte higherBoundary = 0xFF);
 void changeEffectSpeed(int8_t amount);
-void changeBrightness(int8_t amount);
 void changeEffectIntensity(int8_t amount);
 void decodeIR(uint32_t code);
 void decodeIR24(uint32_t code);
--- a/wled00/ir.cpp
+++ b/wled00/ir.cpp
@@ -21,6 +21,54 @@ uint16_t irTimesRepeated = 0;
 uint8_t lastIR6ColourIdx = 0;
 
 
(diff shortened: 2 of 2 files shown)


WJCFerguson:
I now realize that there are the separate White-brightness adjustment buttons for RGBW strips, that should match these.  I can add those changes too, but would first like to hear whether they're desired and might be merged.

Aircoookie:
Thank you, I like this change a lot!
I am inclined to merge it soon if there is no negative feedback from IR users :)

WJCFerguson:
Glad to hear you like it.  

Would you like me to try to put in the RGBW W-brightness fix as well?  I have something right now, but I got carried away and made some knock-on changes (the W-brightness code didn't use the ACTION_* convention for repeats and so I brought it in line).  The other fun issue is I don't have any RGBW strips to properly test it, though I could get fairly confident by modifying `col[2]` or something instead of `col[3]` to make sure it almost definitely works.

Aircoookie:
I'd believe for color channels (including W) it is less necessary, since we are already using a geometric distribution with the gamma correction feature (which is by default enabled by default for color, but disabled for brightness). White levels 5 and 7 would map to 0 for example. You could give it a try if you'd like, but I believe you will find that for colors the linear change should quite closely match the perceived brightness :)

PR #1744, opened 2021-02-14
Fix Warnings

embedded-creations:
PlatformIO.ini suppresses all warnings, which is making it difficult to find actual errors in the code.  As is, dev has 48 warnings if all the warning suppression build flags are removed.  I fixed all but four of them.  Most of the fixes are simple and don't require testing.  A few are also simple but code was changed, so its worth reviewing and probably testing the changes.  Two need some clarification on intent before fixing.  Two need a PR to an external library to fix.

Warnings yet to fix:

- [ ] warning: "CONFIG_LITTLEFS_FOR_IDF_3_2" redefined - we still see this, but only once
  - We create that define (in both WLED.h and in PlatformIO config), but with IDF versions 3.3 and earlier, LittleFS creates the define internally too.
  - Does this need to be defined even if WLED is compiled with a more recent IDF version?
- [ ] lxparser - parseLx() - "second operand of conditional expression has no effect [-Wunused-value]"
  - `constrain(tmpBri, 0, 255);`
  - remove (as it's doing nothing right now, or fix: `tmpBri = constrain(tmpBri, 0, 255);`

## To Test

Should be fine, but as code was actually changed, it's best to test the changes to:

- [x] e131.cpp
- [ ] ESPDMX.cpp
```
wled00/src/dependencies/dmx/ESPDMX.cpp:83:14: warning: statement has no effect [-Wunused-value]
   dmxStarted == false;
```
- [x] xml.cpp
```
wled00/xml.cpp:315:44: warning: ISO C++ forbids converting a string constant to 'char*' [-Wwrite-strings]
       sappends('m',SET_F("(\"pow\")[0]"),"");
```

## Warnings from PlatformIO Managed Library

These need to be fixed with PRs to the library
```
.pio/libdeps/esp32_m5stack_atom_lite_custom2/IRremoteESP8266/src/IRsend.cpp: In member function 'bool IRsend::send(decode_type_t, uint64_t, uint16_t, uint16_t)':
.pio/libdeps/esp32_m5stack_atom_lite_custom2/IRremoteESP8266/src/IRsend.cpp:663:12: warning: unused variable 'min_repeat' [-Wunused-variable]
   uint16_t min_repeat = std::max(IRsend::minRepeats(type), repeat);
            ^
.pio/libdeps/esp32_m5stack_atom_lite_custom2/IRremoteESP8266/src/IRac.cpp: In member function 'bool IRac::sendAc(stdAc::state_t, const stdAc::state_t*)':
.pio/libdeps/esp32_m5stack_atom_lite_custom2/IRremoteESP8266/src/IRac.cpp:1292:9: warning: unused variable 'degC' [-Wunused-variable]
   float degC = desired.celsius ? desired.degrees
         ^
```

## Warnings fixed in dependencies

Should these be updated somewhere upstream, not just in this repo?

- [ ] AsyncMqttClient.cpp
```
wled00/src/dependencies/async-mqtt-client/AsyncMqttClient.cpp: In constructor 'AsyncMqttClient::AsyncMqttClient()':
wled00/src/dependencies/async-mqtt-client/AsyncMqttClient.cpp:40:61: warning: format '%x' expects argument of type 'unsigned int', but argument 3 has type 'uint64_t {aka long long unsigned int}' [-Wformat=]
   sprintf(_generatedClientId, "esp32%06x", ESP.getEfuseMac());
                                                             ^
wled00/src/dependencies/async-mqtt-client/AsyncMqttClient.cpp:40:61: warning: format '%x' expects argument of type 'unsigned int', but argument 3 has type 'uint64_t {aka long long unsigned int}' [-Wformat=]
```

- [ ] Blynk warnings: "ntohl" redefined, warning: "ntohs" redefined, warning: "htonl" redefined, warning: "htons" redefined
  - This has been fixed in Blynk already, what's the best way to apply it to the repo?


wled00/file.cpp
```diff
@@ -34,7 +34,6 @@ File f;
 //wrapper to find out how long closing takes
 void closeFile() {
   DEBUGFS_PRINT(F("Close -> "));
-  uint32_t s = millis();
```
Aircoookie, line 37: This is needed if FS debugging is enabled (see 2 lines below). Rather than removing we should enclose it in an `#ifdef WLED_DEBUG_FS`

wled00/json.cpp
```diff
@@ -134,7 +134,6 @@ void deserializeSegment(JsonObject elem, byte it)
           if (sz == 0 && sz > 4) break;
 
           int rgbw[] = {0,0,0,0};
-          byte cp = copyArray(icol, rgbw);
```
Aircoookie, line 137: byte `cp` is unused, but `copyArray(icol, rgbw);` needs to remain

wled00/wled.cpp
```diff
@@ -116,8 +116,6 @@ void prepareHostname(char* hostname)
 //handle Ethernet connection event
 void WiFiEvent(WiFiEvent_t event)
 {
-  char hostname[25] = "wled-";
```
Aircoookie, line 119: This needs an `#ifdef WLED_USE_ETHERNET` instead

wled00/wled.h
```diff
@@ -306,7 +308,7 @@ WLED_GLOBAL bool huePollingEnabled _INIT(false);           // poll hue bridge fo
 WLED_GLOBAL uint16_t huePollIntervalMs _INIT(2500);        // low values (< 1sec) may cause lag but offer quicker response
 WLED_GLOBAL char hueApiKey[47] _INIT("api");               // key token will be obtained from bridge
 WLED_GLOBAL byte huePollLightId _INIT(1);                  // ID of hue lamp to sync to. Find the ID in the hue app ("about" section)
-WLED_GLOBAL IPAddress hueIP _INIT((0, 0, 0, 0));           // IP address of the bridge
+WLED_GLOBAL IPAddress hueIP _INIT_N((( 0, 0, 0, 0))); // IP address of the bridge
```
Aircoookie, line 311: Very minor nitpick. There is an extra space after `(((`

Jason2866:
Wled uses a old version of IRremoteESP8266. The actual version (2.7.15) probably do not generate a warning.

Aircoookie (commented):
Thank you! This is hugely appreciated!

> Does this need to be defined even if WLED is compiled with a more recent IDF version?

IIRC, if this is not defined, support for file time modified is provided. I haven't tested if that has any impact on performance. We might consider only defining it on SDK 3.2

> lxparser - parseLx() - "second operand of conditional expression has no effect [-Wunused-value]"

Seems like an oversight by the original contributor. We should apply the fix.

Tested E1.31 and DDP, still works fine as expected :)
I lack the hardware to test DMX, but it appears as if the `==` was an oversight as well.
The changed line in xml.cpp works fine.

MQTT could indeed be fixed upstream.
With Blynk the current codebase included in WLED is years old - we might consider updating to upstream. I remember pruning a lot of code that was unnecessary for WLED from the Blynk library, which reduced RAM and/or flash usage (cannot remember exactly which)

embedded-creations:
Sorry I haven't gotten back to this PR, I've been very focused on the DMA SPI work with my limited time as of late.

Aircoookie:
No worries, implementing the results of my own review is the least I can do :) DMA SPI has way more priority, too!
Just wanted to get this merged before it causes too many conflicts.

PR #1767, opened 2021-02-21
Made PWM frequency configurable; changed ESP32 default PWM frequency

ezcGman:
Hey there,

I've just compiled WLED for using it on the QuinLED-Quad with some analog RGB strips I have. After switching it on it made a *horrible* beeping sound and many colors and everything below 100% brightness. After a little chat in #quinled-analog in Quindors Discord, somebody recommended to use 19531Hz, which apparently is also the max suggested in the ESPHome docs: https://esphome.io/components/output/ledc.html#recommended-frequencies

With this frequency (but still on 8bit resolution), I have zero beeping noises on any color and any brightness level. So I added this as default for the ESP32 and also made it configurable with a build flag :)

Greetings,

Andy!

### DIFF
--- a/wled00/NpbWrapper.h
+++ b/wled00/NpbWrapper.h
@@ -112,6 +112,15 @@
   #endif
   #undef RLYPIN
   #define RLYPIN -1 //disable as pin 12 is used by analog LEDs
+  
+  #ifndef PWM_FREQUENCY
+    #ifdef ARDUINO_ARCH_ESP32
+    #define PWM_FREQUENCY 19531
+    #else
+    #define PWM_FREQUENCY 880
+    #endif
+  #endif
+  
 #endif
 
 //automatically uses the right driver method for each platform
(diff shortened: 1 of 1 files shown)


wled00/NpbWrapper.h
```diff
@@ -258,7 +267,7 @@ class NeoPixelWrapper
           #endif
         }
         analogWriteRange(255);  //same range as one RGB channel
-        analogWriteFreq(880);   //PWM frequency proven as good for LEDs
+        analogWriteFreq(PWM_FREQUENCY);   //PWM frequency proven as good for LEDs
```
WJCFerguson, line 270: just a drive-by comment as I saw this mentioned on discord, but the comment on this line now seems out of place, and perhaps a comment should be at the definition?  (I'm not a WLED dev, just browsing.)

Aircoookie:
Hi!
Thank you for the PR and sharing this peculiar frequency :) In 0.12, the entire LED driving code is changing to enable configurable multi output! In implementing that, I have already made the frame rate configurable using the `WLED_PWM_FREQ` define.
I have updated the default value for ESP32 to 19531Hz. Feel free to give the dev branch a try, but note that currently only the Red and Green channels of an RGB PWM strip work as only two pins are changed. Will change that shortly :)

Greetings from me too!

ezcGman:
Hey @WJCFerguson and @Aircoookie,

sorry for the late reply!

Thx @WJCFerguson for the suggestion, totally overlooked that: Would have fixed but now I'm too late ;)

@Aircoookie: That is awesome news! I didn't check if the dev branch has changes for this, I should have done before sending in the PR...
I'll give the dev branch a try and also try to make the necessary changes to support a third channel ;)

Thanks again and greetings,

Andy!

ezcGman:
@Aircoookie Just compiled the esp32dev build, flashed to the ESP32 on the QuinLED Quad and it works peeeeerfectly fine! Set it to "PWM RGB", specified the pins and that's it :) Frequency I saw it also updated, as you said, so perfect!

One question though: What does the "index" do? 🤔 I left that on 0 and just works for me. Still curious, what that does

Aircoookie:
@ezcGman Awesome! Index is only relevant if you have digital LEDs too, or multiple PWM strips. If you e.g. have 30 digital LEDs and you set an effect that turns on each LED in sequence (wipe), setting the PWM index to 0 will make it mirror the first digital LED, 15 the one in the middle and 30 will make it turn on after all LEDs on the digital strip :)

ezcGman:
Ooooh, ok :) wow, that's some advanced tech / thinking ahead there! Nice!
Anyways, thx for this! That's all (and more) the PR was about :D

PR #1783, opened 2021-02-27
Fix wv XML response when using RGBW_MODE_MANUAL_ONLY

DerFetzer:
This should fix the problem in #888


Aircoookie:
Fixed, thank you for pointing me in the right direction. `useRGBW` was replaced with `strip.isRgbw` in the meantime, thus I updated it manually instead of merging. Thank you :)

PR #1808, opened 2021-03-19
Realtime UDP Interval Smoothing

jdtsmith:
This implements "realtime smoothing" (RTS).  RTS works to overcome the effects of network latency _jitter_ in UDP realtime streams.  These cause artifacts which make playback appear "jumpy" or "flickery", especially when compared to normal WLED animations.

RTS works by buffering a configurable number of frames in a ring.  It automatically infers the packet arrival `deltaT` interval statistically from frame arrival times, and uses active feedback control to _smooth_ realtime playback, delivering the received strip data at precise time intervals and keeping the buffer at target (~half) fullness.  This leaves room to accommodate late arriving frames, or multiple "bunched up" frames.

RTS is a pure compile-time option, and if omitted, it shouldn't add any memory or processing overhead to WLED.  When enabled, it implements a configurable number of buffers (5 by default) of the maximum UDP packet length (although this can be configured as well for fixed smaller strip sizes).  It includes detailed optional logging which can be used to diagnose network conditions and tune RTS parameters, although the defaults should work in most cases.

Tested down to 15ms delivery intervals on ESP32 with all available realtime modes.  Multi-packet frames are supported, but for more than 2-3 packets per frame at high frame rates, the current polling method misses 10% or more of the packets.  TPM2.net is not supported but possible.

See the [fork's Wiki](https://github.com/jdtsmith/WLED/wiki) for more details as well as a movie of Realtime Smoothing in action.

Also implemented is the proposed `IRGBW` (protocol 6, like WARLS+W, with 16bit indexes) for individual pixel updates.


wled00/udp.cpp
```diff
@@ -83,9 +88,16 @@ void realtimeLock(uint32_t timeoutMs, byte md)
 
   realtimeTimeout = millis() + timeoutMs;
   if (timeoutMs == 255001 || timeoutMs == 65000) realtimeTimeout = UINT32_MAX;
-  realtimeMode = md;
 
-  if (arlsForceMaxBri && !realtimeOverride) strip.setBrightness(scaledBri(255));
+  // Turn on if it was off
```
jdtsmith, line 92: This turns the strip on if it's off when realtime smoothing starts (it will turn off afterwards).

blazoncek: Comparing this to the similar code in wled_serial.cpp (`if (!realtimeMode && bri == 0) strip.setBrightness(briLast);`) it may be unnecessary to scale briLast and check for realtime mode change.
Also it would be good to return state back to off once realtime ends if it was off when realtime started (I can't seem to find the part where your code does that).
Last thing, your code only executes arlsForceMaxBri override only on mode change whereas original did it always. Is this intentional?

wled00/udp.cpp
```diff
@@ -160,16 +254,66 @@ void handleNotifications()
...
+	}
+      } while (packetSize);
+      return;
+    }
+  }
 
   // WLED nodes info notifications
   if (isSupp && udpIn[0] == 255 && udpIn[1] == 1 && len >= 40) {
-    if (!nodeListEnabled || notifier2Udp.remoteIP() == Network.localIP()) return;
+    if (!nodeListEnabled || realtimeIP == Network.localIP()) return;
```
blazoncek, line 316: Why this change?
It is inappropriate in this case.

wled00/udp.cpp
jdtsmith: Thanks.  [Here](https://github.com/Aircoookie/WLED/blob/1f871690655a280c4a74bc681f7c0cb2d4d79247/wled00/udp.cpp#L211) is where the existing code _resets_ the strip to former brightness (using `bri` as the "saved value") once the UDP stream times out.  This also turns it _off_ if it was off beforehand, which is quite nice.  

For `arlsForceMaxBri`, I moved that inside the test for realtime mode turning on, since it seems resetting it to max brightness each and every time a packet arrives wouldn't be desired (or efficient).  For example, if the user decided it was too bright after realtime started, and slid the brightness slider down, you'd be surprised when it reset. But I'm not entirely clear how this variable is intended to be used, so could certainly go back to setting it back to max bright on every packet.

In terms of `scaledBri`, unless I'm mistaken, everywhere brightness is set this function is used so as to respect the maximum current limit and save your power supply.  So I suspect this is a bug in `wled_serial`.

wled00/udp.cpp
jdtsmith: If you look [just above](https://github.com/Aircoookie/WLED/blob/1f871690655a280c4a74bc681f7c0cb2d4d79247/wled00/udp.cpp#L265), you can see `realtimeIP` will be set to `notifier2Udp.remoteIP()` if `isSupp` is true.  So these are equivalent.  Since I already tested for "self-sent updates" above in the realtime block, this is just a minor optimization.  But the name is certainly misleading I agree.  Perhaps we should call it `remoteIP` and set it once.  I now see a superfluous setting of it below in the TMP2.net block.  Or for code clarity just revert this change.

blazoncek: I agree about variable rename. It has to be clear it is notifier*2*Udp and not notifierUdp.

But still you are not correct about its use. realtimeIP is assigned inside if statements (and remains set that way until a new trigger arrives). Node notifications rely only on 2nd UDP port notifications.

wled00/udp.cpp
blazoncek: 1. reviewing the code, I can se your point now since relay reacts to strip.getBrightness() and not bri as I initially thought.
2. arlsForceMaxBri (as it is most likely intended now) is unconditional override since it is a setting (as opposed to parameter)
3. you may have a point! :)

wled00/udp.cpp
jdtsmith: The if statement only selects which UDP port (1st or 2nd) is involved, so one branch of it will have reliably executed each time the WLED Node if block is executed.  But for readability, easy to switch.

blazoncek: Reviewing the code (for the third time, even though the node broadcast code was coded by me) I can see you are wrong.
Notifier relies only on 2nd UDP port broadcasts and not both. realtimeIP (packets) can be transmitted over both ports and therefore realtimeIP can have value not intended for node broadcasts.
I still disapprove the change.

jdtsmith: Reverted.

psy0rz:
I was looking into this, but I have to say i'm surprised how well WLED handles 300 leds via DRGB at 60fps.

I'm doing this using a ESP32 and a SK6812 ledstrip.

I'm considering to create a modified WARLS protocol that uses a 2 byte index, so i can adres more than 255 leds in an effcient manner.

Still your smoothing patch sounds awesome and very welcome.

jdtsmith:
Thanks @psy0rz; check the [Wiki](https://github.com/jdtsmith/WLED/wiki) for a video showing with and without smoothing (in a "heartbeat" torture test).  Since realtime smoothing is a moderately substantial addition, I'm waiting to hear from the developer(s) whether it's something they are interested in including.  

Of course the WARLS+W/IRGBW 2 byte protocol could be added entirely separately, and is a fairly straightforward addition.  And the best solution is probably to get one of the new Quinn-LED ethernet boards, but that will limit applications.

psy0rz:
Thanks, 

I decided to go for direct control from the raspberry for my next project. I found someone who figured out how to do 16 channels in parallel at 60fps. I'm making a nodejs module for it: https://github.com/psy0rz/rpi-ws281x-smi

jdtsmith:
Was this PR closed due to the dev branch being removed in anticipation of 0.13 release?  Is there enough interest to re-target master once the release drops?   I've been using it regularly and it performs well and really help with realtime UDP over typical home WiFi.

blazoncek:
This may need rebasing since there were quite a few changes recently.
You will need to make a new PR into master, all other branches may disappear eventually.

PR #1832, opened 2021-03-28
Missing ) when IR enabled

pfeerick:
Credit for find this goes to Quindor while testing 0.12 b2 2103230 ;)


pfeerick:
Fixed in e46078286d051850d4455d15a8f11fd52a3a0a8a

PR #1844, opened 2021-03-30
Release engineering improvements

andyshinn:
Some release engineering improvements.

* Moved version to be set in pre-script which grabs version from `package.json`. One less place to modify version number. Open to suggestions on better managing the version.
* Scripts didn't seem to necessarily run in order of listing and caused race conditions in gzip firmware and renaming. Since both scripts were relying on `build_output` folder structure they have been combined and more generalized.
* Release firmwares have a new build option for the build name which gets used by the new `output_bins.py` script to generate release binaries in `build_output/release` if the env variable `WLED_RELEASE` is set to something (such as `WLED_RELEASE=1`).
* Add `requirements.in` and `requirements.txt` to help version the PlatformIO environment. The `requirements.in` file is used by `pip-tools` to update `requirements.txt`.
* Runs CI action in a matrix using default environments to build more in parallel.
* CI action has a release job that will run on tag push or release creation and push the release named binaries (ones using `WLED_RELEASE_NAME`) to the release. When commiting and then pushing first using `git push --tags` the release gets created as a draft which uploads the assets but gives opportunity to edit before publishing. You can also publish a release tag in the interface and it will run and upload the binaries to that existing release based on same tag.

We also have `VERSION` (the dated version build code) and `WLED_COENAME` in C++ headers. Thoughts about removing these and making them available in `platformio.ini` instead so they are more readily available to the build process and pipeline?


jaredgisin:
Nice! Will be nice to have the build output scripts working more reliably.

platformio.ini
```diff
@@ -151,10 +151,10 @@ build_flags = -g
...
 
 [scripts_defaults]
-extra_scripts             = pio-scripts/name-firmware.py
-                            pio-scripts/gzip-firmware.py
-                            pio-scripts/strip-floats.py
-                            pio-scripts/user_config_copy.py
+extra_scripts = pre:pio-scripts/set_version.py
+  post:pio-scripts/output_bins.py
+  post:pio-scripts/strip-floats.py
+  pre:pio-scripts/user_config_copy.py
```
pfeerick, line 157: My OCD is saying would this be better formatted like...
```
extra_scripts = 
  pre:pio-scripts/set_version.py
  pre:pio-scripts/user_config_copy.py
  post:pio-scripts/output_bins.py
  post:pio-scripts/strip-floats.py
  ```

Comment (andyshinn) [reply platformio.ini]: I modified this part. But there are definitely many parts of the configs that could use some additional formatting and cleanup.

**Comment by Aircoookie** [comment]
This is a whole slew of great improvements!
Will review, test and hopefully merge tomorrow.

The codename is changed infrequently and could use the same logic as `WLED_VERSION` - that is being defined in `package.json` and auto-populated. It's fine in `wled.h` though as it does not need to be used in multiple locations. The same is true for `VERSION`, although that is/should be bumped with every push. One potential minor problem I see is compiling with Arduino IDE (I don't know who would prefer that over PIO, but technically it's a possibility), which would need to fallback to meaningless defaults for all version identifiers.

Once again, awesome work 👍

**Comment by Aircoookie** [comment]
Works very well overall!
The only part I haven't figured out yet is how to conveniently set and unset the `WLED_RELEASE` environment variable. Would it be more convienient to set it with a cpp define or a var in `package.json` or am I missing an easy way to set `WLED_RELEASE`?

**Comment by andyshinn** [comment]
The idea was that it is always set for the 4 common envs we release now (via https://github.com/Aircoookie/WLED/blob/master/.github/workflows/wled-ci.yml#L55-L57) but you can set it via shell if you want a release binary name for one-offs with something like:

```
WLED_RELEASE=1 pio run -e d1_mini
```

Issue #1519, opened 2020-12-17
TM1814 support might not be working correctly


wonderslug:
**Describe the bug**
When trying to run a set of TM1814 chips leds im only getting a what looks like a random pattern on the first LED and what looks like a default color cycle on all the other LEDs.

Ive been able to use a modified NeoPixelBus example to work and it seems to behave fine.

Ive noticed that one of the requirments for the [NeoWrgbTm1814Feature](https://github.com/Makuna/NeoPixelBus/wiki/NeoPixelBus-object#tm1814-features) is that the SetPixelSettings needs to be set with a NeoTm1814Settings:
`strip.SetPixelSettings(NeoTm1814Settings(165,165,165,165));`

I am using an ESP8266 so using pin 3 on a QuinLEd-Uno.

Ive tried to compile this into WLED myself but still get the same behaviour .

My changes to the NeoPixelBus Cylon example to get it working were only the Method and Feature def
`NeoPixelBus<NeoWrgbTm1814Feature, NeoTm1814Method> strip(PixelCount, PixelPin);`

A slight change to an RgbColor to an RgbwColor, and adding the above SetPixelSettings right after the `Begin()`.

**WLED version**
 - Board: d1mini
 - Version 0.11.0
 - Format Binary, self-compiled, WBLD bot compiled, 

**Additional context**
Anything else you'd like to say about the problem?

Thank you for your help!

esev:
I've seen the same issue. In addition to the SetPixelSettings mentioned above, it doesn't look like `TYPE_TM1814` is handled in the code below:
https://github.com/Aircoookie/WLED/blob/d603a8a9be0d9d25917ab04a4f9fef5ca10c139e/wled00/bus_wrapper.h#L855-L863

esev:
@wonderslug if you are still interested in this bug, could you test with #1847?

Also, just curious, in your examples you were configuring a max of 16.5 mA per LED. Was there a specific reason for that, or was this based on another example?

wonderslug:
Howdy @esev 

I was just using that as the example from https://github.com/Makuna/NeoPixelBus/wiki/NeoPixelBus-object#tm1814-features as an example.  Im actually not sure what the proper values should be.

wonderslug:
As well I will be giving #1847 a try.

mlangner84:
I was able to get the TM1814 string lights I'm using working on an ESP32 with multiple pins being used (following and adopting the recommendations here.)
[...]

PR #1847, opened 2021-04-01
Initialize PixelSettings for TM1814 strips

esev:
TM1814 strips need to have PixelSettings configured. This PR configures the TM1814 chip to allow the max current for the LED. Current can then be limited via the LED settings as normal. This PR also enables support for TYPE_TM1814.

https://github.com/Makuna/NeoPixelBus/wiki/NeoPixelBus-object#tm1814-features

Related to: #1519

### DIFF
--- a/wled00/bus_wrapper.h
+++ b/wled00/bus_wrapper.h
@@ -197,7 +197,8 @@ class PolyBus {
     T tm1814_strip = static_cast<T>(busPtr);
     tm1814_strip->Begin();
     // Max current for each LED (38.0 mA).
-    tm1814_strip->SetPixelSettings(NeoTm1814Settings(380, 380, 380, 380));
+    const uint16_t max = NeoTm1814Settings::MaxCurrent;
+    tm1814_strip->SetPixelSettings(NeoTm1814Settings(/*R*/max, /*G*/max, /*B*/max, /*W*/max));
   }
   static void begin(void* busPtr, uint8_t busType, uint8_t* pins) {
     switch (busType) {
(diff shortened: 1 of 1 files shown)


wonderslug:
Howdy 

Thanks for working on this @esev 

I applied the PR to master and seems to be better but still having some significant issues on handling of things.

I have a [video here](https://s3.amazonaws.com/wonderslug/snaps/IMG_3297.MOV) showing whats happening.

In a solid color mode it always goes to the proper color on the fist pixel but its blinking but the 2nd pixel on goes to a solid red color.  When turning off it revers to its no connectivity mode for the chip (cycling rainbow and white channel)

esev:
Hey @wonderslug. 

That's no good!

Some other things to try:
1. I'm using GPIO2 with my TM1814. Is it possible to use GPIO2 with the QuinLED-Uno? That might rule-out and differences there.
2. Can you take a screenshot of the Config -> LED Preferences page? Mine looks like this.
Screenshot 2021-04-01 3 02 24 PM
3. This is kind of a long-shot, but if you're on GPIO3, try changing `NeoEsp8266DmaTm1814Method` to `NeoTm1814Method` to match what you tried in #1519 (I haven't checked that this will even compile) https://github.com/Aircoookie/WLED/blob/d603a8a9be0d9d25917ab04a4f9fef5ca10c139e/wled00/bus_wrapper.h#L121

wonderslug:
So I tested out the change to FX.cpp that @mlangner84 suggested in #1519 with your PR branch without the change you suggested above. It works great.  Im able to run my string of 15 LEDs on it just fine.  

https://github.com/Aircoookie/WLED/blob/48c0360877a9a658f4669850e48d9ebba4ee103f/wled00/FX.cpp#L36-L39

Changing the 500 to FRAMETIME worked just for the test.  It probably needs to tied into the Bus Type stuff but Im not sure how to make that happen right now.

PR #1850, opened 2021-04-02
Support of simple gestures for VL53L0X laser sensor

eg321:
Hi,

That usermod implements support of simple hand gestures with VL53L0X laser sensor: on/off and brightness correction.
It can be useful for kitchen strips to avoid any touches.
 - on/off - just swipe a hand below your sensor ("shortPressAction" is called and can be customized through WLED macros)
 - brightness correction - keep your hand below sensor for 1 second to switch to "brightness" mode.
                           Configure brightness by changing distance to the sensor (see parameters below for customization).
                           "macroLongPress" is also called here.

it's working fine for me and maybe will be useful for anyone else.
I've adapted my changes to usermods v2 structure. Let me know, if I forgot something.

### DIFF
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -577,7 +577,7 @@ WLED_GLOBAL UsermodManager usermods _INIT(UsermodManager());
 #else
   #define DEBUG_PRINT(x)
   #define DEBUG_PRINTLN(x)
-  #define DEBUG_PRINTF(x)
+  #define DEBUG_PRINTF(x...)
 #endif
 
 #ifdef WLED_DEBUG_FS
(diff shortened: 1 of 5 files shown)


wled00/wled.h
```diff
@@ -577,7 +577,7 @@ WLED_GLOBAL UsermodManager usermods _INIT(UsermodManager());
 #else
   #define DEBUG_PRINT(x)
   #define DEBUG_PRINTLN(x)
-  #define DEBUG_PRINTF(x)
+  #define DEBUG_PRINTF(x...)
```
eg321, line 580: it looks I'm first who tried to use DEBUG_PRINTF... :)

PR #1864, opened 2021-04-05
Ledmap multisegment fix.

blazoncek:
Fix for using ledmap.json with multiple segments.


andyshinn:
It looks like there is still a minor issue with mirroring on multiple segments. Single seems fine. But the mirror of the remapped LEDs seem empty (off).

blazoncek:
Fixed for mirrored segments too.

Aircoookie:
Thank you!
I totally did not notice that the remapped index would fail the `if (indexSet >= SEGMENT.start && indexSet < SEGMENT.stop)` check if the segment does not cover the entire strip.

You have changed `indexSet` back to `int16_t`, I've changed it to `int` in my latest commit https://github.com/Aircoookie/WLED/commit/ecdc0a38001d5a815b0f774f061c016f21bab198 to prevent a crash on ESP32 when an effect function provides an index out of segment bounds if the segment is reversed. I don't think that has any impact on this fix, so I will change it back to prevent regression. Please let me know if the fix still works for you (it does in my testing)

blazoncek:
Yes, I am sorry about int16_t, unfortunatelly I used an older commit as a base.

blazoncek:
While on topic, I would encourage you to consider moving "skip first led" from FX into bus manager.
It makes alot cleaner code in both sides.
I can make PR from my repo but requires a few changes.

Aircoookie:
No worries!
I will move "Skip first LED" to bus manager once I add per-strip current limiting. There will be a new per-strip setting for it, since users might not want/need a sacrificial pixel on every output. I agree that it is cleaner than in `FX_fcn.cpp`.

blazoncek:
> No worries!
> I will move "Skip first LED" to bus manager once I add per-strip current limiting. There will be a new per-strip setting for it, since users might not want/need a sacrificial pixel on every output. I agree that it is cleaner than in `FX_fcn.cpp`.

I may actually already have this code in place but is currently used as a RGBW override, which (turns out) is not really necessary. :)

PR #1873, opened 2021-04-08
Added Digital Clock overlay option for 7-segment style clocks

dinamitemic:
Hi, I've taken the changes made in [PR#995](https://github.com/Aircoookie/WLED/pull/995) for what concern the UI and html side and then I added my code for driving 7-Segment style Digital Clock.
This implementation can be customized in order to meet different configurations in terms of number of led per segment, number of led per dots and how they are arranged.
Detailed description of the code can be found [here](https://www.hwtronics.com/2021/04/08/7-segment-digital-clock-firmware-wled-powered/)
Hope you enjoy it!


NeariX67:
I really like the compactness of your code. If you look [at my fork](https://github.com/NeariX67/WLED_Clock) it took me many many lines of code to do this. And i did not use the overlay function but created an effect for it.

I like it! I'm gonna test this later today ;)

danielhoherd:
I just stumbled across this and love it! Great blog post on how to use it, because I was originally thinking it was a binary clock or something.

loverboyritesh:
Thankyou Aircoookie For everything 
Thank you  dinamitemic for your 7 segment digital clock feature ..
now to the point .....WLED supports multiple segments now and also multiple outputs pins to connect many neopixel strands ..now i was thinking is there a way to use more than one clock at same time ...means one segment /or seperate neopixel strand displays analogue clock ..and other can display the digital clock ..s that both clocks are displayed together ....right now i am using 2 seperate d1mini - one controles 60led analogue clock and other 30led 7 segment clock[hwtronics]....any help to save esp boards and combine these 2 features in one ...Thank you ...
{and why WLED??????? because its the best and i want all the pellete features and artnet features..]

Aircoookie:
@loverboyritesh That can be accomplished with two easy code changes:
- In line 77 of `overlay.cpp` (using code from this PR), add 60 to `int offset`. 
- Starting from line 195, replace the switch block with just `_overlayDigitalClock(); _overlayAnalogClock()`, which will enable both overlays.

Since there is significant interest for this feature and I really like the approach of the code implementing this as an overlay in this PR, I think I'll rewrite this to be added as a usermod (even though there are at least 2 other 7-segment usermods).

Aircoookie:
I'm pretty sure this is now covered by https://github.com/Aircoookie/WLED/tree/main/usermods/seven_segment_display_reloaded, so I will close this PR.

loverboyritesh:
With The new 0.14.0 Everything is possible , with usermods, with segments and Also with DDP and art-net without any physical connection , and not just analogue and 7-seg Clock, Scrolling matrix clock and many others, and also possibility to enable multi-usermods .... All thanks to Aircoookie... Its like a dream come true Matrix and Digiclocks.

Raj800:
I am trying to make the same same thing and I do feel that I can get it to work as a clock. 

But I also want it to work as a 2D Matrix so that I can display effects on it as if it was a two dimensional display. I am planning to put LED strips in all the sections of the 3D printed design for the effects.

This person has made it with a complete custom software, and it looks glorious!!!
https://www.printables.com/model/201156-shelf-clock
https://youtu.be/hknXNxCtM_E

I want to make this, with WLED.

The problem is that 2D Matrix requires the LEDs to be in sequential rows, like row number 1, row number 2, row number 3.
In this scenario the LEDs will not be in that sequence as they will be connected as per 7 segment requirements.
How can we configure or how can we manipulate it so that the WLED software thinks that it's a matrix and will display the effects properly.
For example roll number 1 will be the top segment of all 7 segments so it will be filled with all the LEDs
Row number 2 will have one LED then 9 missing a LEDs then again one LED then 9 missing LEDs and so on. 

Basically this will be the vertical segments but we will make WLED software think that it's a matrix with a d's absent in those positions does it will let us display effects

This may or may not me the best way to do it please do let me know

I am in process of 3D printing the parts right now, so I have not tested the code yet using the seven segment usermod. Please let me know if this functionality is already implemented.

loverboyritesh:
Bro its complicated . even though you manage to get it working the effects will not be so effective because of the missing leds in matrix setup . even in linear strip you can get awesome effects considering it continuous strip if you group leds in a single segment to create 7 virtual leds per digit... and that too you have not mentioned how many leds you are using per segment of SSD. lets assume that you are making exact of the links you provided then also your calculations are wrong i feel , 1 row - top segments [ generally 'A' may be different in you  case] of all digits. 2 row 1 led 7 gap {you said 9] 1led 7gap and so on . even in top row you should consider that 1 led is missing after each 7 leds .... that means 1gap 7leds 1gap 7leds so on ...

Assuming 7 leds per segment of SSD and gaps [missing leds] as 0 and real leds as 1.

011111110111111101111111011111110111111101111111011111110
100000001000000010000000100000001000000010000000100000001
100000001000000010000000100000001000000010000000100000001
100000001000000010000000100000001000000010000000100000001
100000001000000010000000100000001000000010000000100000001
100000001000000010000000100000001000000010000000100000001
100000001000000010000000100000001000000010000000100000001
100000001000000010000000100000001000000010000000100000001
011111110111111101111111011111110111111101111111011111110
100000001000000010000000100000001000000010000000100000001
100000001000000010000000100000001000000010000000100000001
100000001000000010000000100000001000000010000000100000001
100000001000000010000000100000001000000010000000100000001
100000001000000010000000100000001000000010000000100000001
100000001000000010000000100000001000000010000000100000001
100000001000000010000000100000001000000010000000100000001
011111110111111101111111011111110111111101111111011111110

so this this would be probably your setup as matrix . only thing is the connection and data line is messed up , again that wouldnt  be a problem coz pixels can be mapped and gapped using led.json [see documentation] , but would need high amount of calculation as to map which led to which position of the strip , 

WLED is a universe in itself , to play around and experiment , dont hesitate to build the hardware part thinking that some thing might go wrong .. everything can be solved from the programme it-self .. or i would suggest to just create a crude setup just with strips and experiment yourself ...

or just go for the linear strip effects grouping 7 leds as 1 virtual led as mentioned earlier, these effects also wont let you down.
and you can shift to better setup if in future if you get a better solution or update  without altering the physical setup .you can do it all from software side ..

blazoncek:
Please use WLED [forum](https://wled.discourse.group) or Discord for help and support questions.

PR #1889, opened 2021-04-12
Sunrise and sunset calculation and sunrise/sunset triggered presets.

blazoncek:
A new PR due to old one being screwed up by soundreactive PR.


Aircoookie:
Outstanding work!

Aircoookie:
Alright, noticed just now that `getSunriseUTC()` increases RAM usage by 640 bytes :(
Based on the vars added for the sunrise functionality, I'd expect at most an increase by 32 bytes, considering you even optimized an extra string.

Scenario | Memory used as reported after compilation
| --- | --- |
master (PR merged) | 40228
master with `9bfe27d` reverted | 39584 (-644)
master with contents of `getSunriseUTC()` replaced by `return 0;` | 39588 (-640)

`getSunriseUTC()` is definitely the culprit, all your other changes, including the new variables, only use 4 bytes net extra memory. Seems like either constants or some of the floating point functions use an absurd amount of memory. I'll try to localize the specific function (if a single one) that causes the high usage, but unfortunately, the math is what it is, so I'm not sure we can optimize the memory usage. I definitely want to keep the awesome timers though, even though a 600b memory hit on 8266 is quite significant.

Aircoookie:
It's `sin/cos/tan`. (`fmod/asin/acos/atan` seem mostly fine, getting 39764b with them still in, so only an extra 180 bytes)

Aircoookie:
This stuff works. I wish I was a math genius like these guys 😛 
No mem usage, and still accurate to the minute for me:

```cpp
//https://web.eecs.utk.edu/~azh/blog/cosine.html (cos_taylor_literal_6terms)
#define modd(x, y) ((x) - (int)((x) / (y)) * (y))

double cost(double x)
{
    x = modd(x, (2*PI));
    char sign = 1;
    if (x > PI)
    {
        x -= PI;
        sign = -1;
    }
    double xx = x * x;

    return sign * (1 - ((xx) / (2)) + ((xx * xx) / (24)) - ((xx * xx * xx) / (720)) + ((xx * xx * xx * xx) / (40320)) - ((xx * xx * xx * xx * xx) / (3628800)) + ((xx * xx * xx * xx * xx * xx) / (479001600)));
}

double sint(double x) {
  return cost((PI/2) - x);
}

double tant(double x) {
  double c = cost(x);
  if (c==0.0) return 0;
  return sint(x) / c;
}
```

39764b with these trigonometric functions. 180bytes lost is still bad, but definitely a vast improment to 640bytes. I hope this doesn't break with other lat + longs 😅

blazoncek:
hmmmm.
https://dsp.stackexchange.com/questions/20444/books-resources-for-implementing-various-mathematical-functions-in-fixed-point-a/20482#20482

Aircoookie:
Currently testing: 
```cpp
//https://stackoverflow.com/questions/3380628
// Absolute error <= 6.7e-5
float acos_t(float x) {
  float negate = float(x < 0);
  x = std::abs(x);
  float ret = -0.0187293;
  ret = ret * x;
  ret = ret + 0.0742610;
  ret = ret * x;
  ret = ret - 0.2121144;
  ret = ret * x;
  ret = ret + HALF_PI;
  ret = ret * sqrt(1.0-x);
  ret = ret - 2 * negate * ret;
  return negate * PI + ret;
}

float asin_t(float x) {
  return HALF_PI - acos_t(x);
}

//http://nghiaho.com/?p=997
double atan_t(double x)
{
  return (HALF_PI/2)*x - x*(std::abs(x) - 1)*(0.2447 + 0.0663*std::abs(x));
}
```

asin, acos, atan are not nearly as bad as sin, cos, tan, I'm seeing an additional 80 byte decrease to 39684 with the approximations, but the sunset has lost a minute of accuracy, so still testing.

Aircoookie:
Update: The minute of precision loss is negated with a better `atan` approximation:

```cpp
//https://stackoverflow.com/a/42542593
#define A 0.0776509570923569
#define B -0.287434475393028
#define C ((HALF_PI/2) - A - B)

double atan_t2(double x) {
  double xx = x * x;
  return ((A*xx + B)*xx + C)*x;
}
```

Seems like we can go with that, I will definitely make a beta for #beta-testing so we can see how much the calculated time differs from the actual sunset/sunrise time for a variety of users/locations.

blazoncek:
My head hurts from math, but I got some boards from @srg74 today and soldering will clear my head. :D

Aircoookie:
Changes pushed.

Legsmaniac:
Found this, looks easier?

https://www.esrl.noaa.gov/gmd/grad/solcalc/solareqns.PDF

blazoncek:
> Seems like we can go with that, I will definitely make a beta for #beta-testing so we can see how much the calculated time differs from the actual sunset/sunrise time for a variety of users/locations.

Today it differed 3 minutes at my place: 3 in the morning (late) and 3 in the evening (early).
IMO it is still acceptable.

Aircoookie:
Sunset a minute early for me today, sunrise a minute late.
Would agree, anything +/- 5 min is acceptable. For fine tuning there is always the minute offset.

carelinus:
Today the error was 38 minutes for me (21:31 WLED vs. 20:53 real). The difference was much smaller last time I checked... Seems to differ over the year?

blazoncek:
I can confirm that calculated times vary quite a bit sometimes.
Will look into it.

blazoncek:
The error has been found, proper fix will take some time.
For the time being I would recommend replacing atan_t() with atan() or using `if (x>1) return atan(x);` within atan_t() function.

carelinus:
Thank you very much for the quick help, the workaround works well!

carelinus:
Uh, I rejoiced to early... Since summer solstice times are completely off. (eg. sunrise ~11 am, sunset 3 am)

blazoncek:
> Edit: my bad; latest commit already seems to work fine

Indeed it does. :)

carelinus:
Is it possible that the function is again broken in 0.13.0-b3?
At least for me sunrise and sunset are many hours off... :-/

blazoncek:
> 
> 
> Is it possible that the function is again broken in 0.13.0-b3? At least for me sunrise and sunset are many hours off... :-/

Fixed in first commit since b3 was released. Get @srg74 bins or compile from source.

carelinus:
The fix works. Thanks for the quick reply!

PR #1890, opened 2021-04-12
Easier holiday/special events handling for background images.

blazoncek:
Enjoy adding new special events for nice background images.


blazoncek:
There is a comma mising in line 168.

blazoncek (commented):
I chose to display background a day early, but your change makes sense.

PR #1905, opened 2021-04-18
Use 22.5 for the TM1814 max current

esev:
Configure the TM1814 constant current at 22.5 mA as this seems to be a common value for several different RGBW LED chips.

I had originally intended this PR to use the `strip.milliampsPerLed` value, but was unable to find a way to make that work with the different voltages used for the TM1814 strip types (see comment below).

Related to: #1519

### DIFF
--- a/wled00/bus_wrapper.h
+++ b/wled00/bus_wrapper.h
@@ -2,9 +2,6 @@
 #define BusWrapper_h
 
 #include "NeoPixelBrightnessBus.h"
-#include "FX.h"
-
-extern WS2812FX strip;  // Provides access to global settings for LEDs.
 
 //Hardware SPI Pins
 #define P_8266_HS_MOSI 13
@@ -199,8 +196,8 @@ class PolyBus {
   static void beginTM1814(void* busPtr) {
     T tm1814_strip = static_cast<T>(busPtr);
     tm1814_strip->Begin();
-    const uint16_t mA = strip.milliampsPerLed * 10;
-    tm1814_strip->SetPixelSettings(NeoTm1814Settings(/*R*/mA, /*G*/mA, /*B*/mA, /*W*/mA));
+    // Max current for each LED (22.5 mA).
+    tm1814_strip->SetPixelSettings(NeoTm1814Settings(/*R*/225, /*G*/225, /*B*/225, /*W*/225));
   }
   static void begin(void* busPtr, uint8_t busType, uint8_t* pins) {
     switch (busType) {
--- a/wled00/bus_wrapper.h
+++ b/wled00/bus_wrapper.h
@@ -2,6 +2,9 @@
 #define BusWrapper_h
 
 #include "NeoPixelBrightnessBus.h"
+#include "FX.h"
+
+extern WS2812FX strip;  // Provides access to global settings for LEDs.
 
 //Hardware SPI Pins
 #define P_8266_HS_MOSI 13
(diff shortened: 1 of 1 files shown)


esev:
I'm thinking this may not be the right way to handle this setting. It looks like the `milliampsPerLed` value is intended to be the number of milliamps consumed by the entire LED module, not one single LED within the module. For the TM1814 this is likely supposed to be around (4*20mA)=80mA.

However this value doesn't play nicely with the "power units" calculations in `WS2812FX::show`. Those calculations seem to be based on a 5V RGB (3 LED) module that uses 53mA. Since the TM1814 modules are typically 12v/24v, these calculations won't match correctly if 80mA is used.

It might just be best to hard-code the constant current setting for the TM1814 modules at 20-22.5mA, since that seems to be the common current values supported in the 5040 RGBW LEDs.

@mlangner84 WDYT?

I'm going to mark this PR as a 'Draft' for now.

makerglenn:
Thanks for your efforts with the TM1814. I just found this thread and (in the background of my life) been trying to get the Costco Atomi String lights to work. I'm just mentioning this because this string of TM1814 LEDs are run at 5v not 12v or 24v. Hopefully you will take that into consideration with your calculations/development efforts.  Hoping solid TM1814 support shows up in WLED soon. Thx again...

esev:
@makerglenn do you know which LEDs are in the Atomi strip?

@davepl are these the same LEDs that [you have](https://www.youtube.com/watch?v=Z0krTGssXFA)? If so, would you be able to capture the [constant current](https://github.com/Aircoookie/WLED/issues/1519#issuecomment-817387540) values used for these?

esev:
> @davepl are these the same LEDs that you have? If so, would you be able to capture the constant current values used for these?

Never mind. Found it. Seems like this is the same 22.5mA.
https://www.youtube.com/watch?v=Z0krTGssXFA&t=20m5s

@makerglenn 22.0 should likely work for these too.

makerglenn:
Yes those are the same LEDs as in the Dave's Garage video. I've cut off their proprietary controller and trying to drive it with WLED. I was trying to get the TM1814 string to work several weeks ago but to no avail - so I took it off my workbench for a while since I don't have a scope (s/w guy). I guess its time to dust them off and give your mods a try. Thx for the current info...

esev:
@makerglenn I think there is still going to be an issue with the [refresh rates](https://github.com/Aircoookie/WLED/issues/1519#issuecomment-817113044). I have a TM1814B which doesn't have a built-in pattern and doesn't require a higher refresh rate.

As an aside, I also ordered some of those Atomi lights. My plan is to add a level shifter on GPI02 of the internal wifi controller and bypass the HC89 chip, going directly from the level shifter to the green `DO` wire like [shown here](https://learn.adafruit.com/assets/64121) with a Pi. That's assuming I can flash WLED onto the internal wifi controller.

makerglenn:
Yes the Atomi string lights use the non-B variant TM1814 that requires the refresh rate per your previous conversation. I successfully flashed the TYWE3S with the WLED ESP866_1M version. The 4 basic serial connections (Tx, Tx, 3.3v, Gnd) are nicely available and labeled on the edge of the Atomi PCB board. Easy to solder programming wires to then to the TYWE3S pins (ESP-like chip). I only had to ground out GPIO0 at the chip pin on powerup to get the TYWE3S into program mode. No issues programming the chip.... but that avenue does not work (as I found out later at the end of Dave's Garage video where he found out that there's a SC92 uP chip that actually drives the TM1814s! The TYWES3 only tells the SC92 what pre-programmed light sequence to play! I decided to breadboard WLED and try and get the string working without the Atomi logic in the way - then build my own controller (I have 6 strands of lights) or wait until the new QuinLED board comes out.

esev:
Is GPIO2 (TXD1) on the TYWE3S connected to the SC92, or GPIO1 (TXD0)? My plan was going to be to use GPIO2 and connect to the outgoing green wire (`DO`) through a level shifter.

Original: TYWE3S (GPIO1) -> SC92 -> On-board TM1814 -> Green (`DO`)
Planned: TYWE3S (GPIO2) -> Level Shifter -> Green (`DO`)

I guess I'll see more when I actually get mine and crack open the case.

esev:
@Aircoookie sorry for the side-discussions on this PR. I think the PR itself should be good to merge (if it's okay with you). 22.5 is the safer default value here as that appears to be what the native controller uses for several different TM1814 LED strips.

Aircoookie:
Thank you :)
No worries about discussion, after all it is 100% related.

PR #1912, opened 2021-04-22
Add IP Address to JsonInfo

acshef:
The value is added to the JSON only if the device is connected to the network, and uses the JSON key `"sip"` to match [wled00/xml.cpp](wled00/xml.cpp#L249). The overarching goal of this is to expose the IP Address to the Home Assistant WLED Integration, so that Home Assistant can provide a link to the WLED device (either directly through the Integration/Device page 🤞 or *ad hoc* in Lovelace).

Note: The WLED wiki would need to be updated to reflect this addition.

### DIFF
--- a/wled00/json.cpp
+++ b/wled00/json.cpp
@@ -549,13 +549,13 @@ void serializeInfo(JsonObject root)
   root[F("brand")] = "WLED";
   root[F("product")] = F("FOSS");
   root["mac"] = escapedMac;
+  char s[16] = "";
   if (Network.isConnected())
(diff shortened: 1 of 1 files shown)


Aircoookie:
Thank you, I believe this is very useful!
I've made a tiny change, renaming the field to `ip` (`sip` was meant to be `status: ip`, and ip is a bit shorter) and made the field present even if not connected, in which case it is an empty string. This aligns it with the behavior of the `lip` (Live IP) field :)

PR #1922, opened 2021-04-26
TM1814 requires refreshes to remain powered off and a slightly faster refresh speed

### LINKED ISSUE #1519: TM1814 support might not be working correctly

esev:
Several of us have tested and found that the current refresh rate of 380 isn't sufficient for the TM1814. It requires a refresh rate of 350. The faster refresh rate is needed to prevent the TM1814 from entering its default flashing cycle.

Also the TM1814 requires that the strip be refreshed/serviced even when the LED is off. If this doesn't happen then the LEDs enter the default flashing cycle. Just like @Aircoookie had thought in https://github.com/Aircoookie/WLED/issues/1519#issuecomment-817113044

Fixes #1519


wonderslug:
Hey @esev.  So im just now getting back to this and I have things set to TM1814 but im still getting the idle mode blinking when off.  I get this even when im forcing a service to in the wled.cpp like you talked about [here](https://github.com/Aircoookie/WLED/issues/1519#issuecomment-827078046).

The interesting thing is if I have the control set to off (and its in its TM1814 idle blink mode) and I go into LED settings and Save it actually turns off the LEDs,  Then when I turn the string on and off again it reverts to the TM1814 idle blink.

Any thoughts on whats going on?

PR #1932, opened 2021-04-28
Added support for reading usermodxml[1-3] and writing to usermodxml[1-3]_s via XML API

marrobHD:
**Added support for reading usermodxml[1-3] and writing to usermodxml[1-3]_s via XML API**

With this pull request, you're able to view variable usermodxml[1-3] and write usermodxml[1-3]_s with the XML API.

"VS" displays usermodxml1, "VT" displays usermodxml2 and "VU" displays usermodxml3.
"&VS" writes to usermodxml1_s, "&VT" writes to usermodxml2_s and "&VU" writes to usermodxml2_s

This feature is only enabled by uncommenting `-D WLED_ENABLE_USERMODXML` inside platformio.ini:
```ini
# platformio.ini
...
[platformio]
...
[common]
...
build_flags =
...
  #uncomment this to enable USERMODXML
  ; -D WLED_ENABLE_USERMODXML 
...
```

As an example, I've updated the QuinLED_Dig_Uno_Temp_MQTT usermod to support the XML API. 39 is the readout of the sensor.

image


usermods/QuinLED_Dig_Uno_Temp_MQTT_XML/usermod.cpp
```diff
@@ -24,14 +25,29 @@ void userConnected()
...
   if (temptimer - lastMeasure > 60000) {
     lastMeasure = temptimer;
-    
+
+//Check for successful connection
+    if (WLED_CONNECTED){
+      sensor.requestTemperatures();
+
+//Gets prefered temperature scale based on selection in definitions section
+      #ifdef WLED_ENABLE_USERMODXML
```
blazoncek, line 42: What if you have two or more usermods that are able to use this feature, both writing to e.g. usermodxml1?
I don't see locking mechanism.
IMO best way to interact with usermod is to call methods from usermod itself and not a usermod manupulating global variables.

marrobHD: I dont how to do this. Im a beginner. Maybe close pr

blazoncek: Another way could be to add/create a usermod API so usermods could exchange information.
In a similar way as addToJsonInfo() etc. is done.
The interaction between usermods was already implemented in Auto Save and Four Line Display usermods.

marrobHD: Well, maybe we better close the PR. I have absolutely no clue to add this

Aircoookie:
Hi and thank you for the PR!
I do think it is a good idea to make usermods more interfaced with external integrations, however the XML API, while not deprecated, has been superseded by the JSON API in pretty much all aspects. Convenient manipulation of segments and presets for example, is a lot easier in JSON and in my opinion it should be used for all new integrations in favor of XML, and thus I believe JSON is the best choice for usermods as well :)

PR #1941, opened 2021-05-03
Add user configurable IR handling

scottrbailey:
This adds a new IR remote type "JSON remote" and allows users to configure and customize any remote by editing ir.json instead of editing and compiling c code.  The keys in ir.json are hex encoded IR codes. 

The user configurable commands can be:
1.  HTTP API commands, 
2. JSON API commands
3. limited c functions, for a few things not available in either of the above (incBrightness, decBrightness, presetFallback)

I added a user_mod directory although this really is not a user_mod. It contains: 
1.  ir.json files for the currently supported remotes, plus a few more I added. 
2. An Excel spreadsheet with labels, codes, actions for currently supported remotes
3. A python script to convert the spreadsheet to json files.
I'm unsure about including this whole folder.  It may be better to remove it and just provide the ir.json files from the wiki, which we'd do in either case.


Aircoookie:
Hi! Thank you, this seems to be able to make using any IR remote a lot easier, without requiring code changes!
I've merged this to a temporary branch so @blazoncek and I can do some review, I hope I will merge it to master soon :)

PR #1967, opened 2021-05-18
Allow to override the max segments setting when compiling

poelzi:
I ran into segment limits quite often, this allows to increase the number with a compile flag

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -61,7 +61,9 @@
   /* How much data bytes all segments combined may allocate */
   #define MAX_SEGMENT_DATA  2048
 #else
+#ifndef MAX_NUM_SEGMENTS
   #define MAX_NUM_SEGMENTS    16
+#endif
   #define MAX_NUM_TRANSITIONS 16
   #define MAX_SEGMENT_DATA  8192
 #endif


blazoncek:
This is not a good idea without some serious memory checks or workarounds in the background.
If you need more than 12 segments on 8266 without modifying memory allocation parts of the code, you are welcome to try my experimetal fork (used for many WLED enhancements, past and future) where the limit is 18 and 24.

blazoncek:
You can see the effect of out-of-memory condition here: #1974

poelzi:
@blazoncek I changed it, so it only is allowed on esp32. I'm using an ESP32 with ~1400 rgbw leds and 32 segments with no problem.

blazoncek:
Due to recent changes in JSON parsing it will be possible to have 32 segments by default on newer builds on ESP32 and 16 segments on ESP8266.
You can close the PR since it may not be needed.

XPloadED:
> Due to recent changes in JSON parsing it will be possible to have 32 segments by default on newer builds on ESP32 and 16 segments on ESP8266.
> You can close the PR since it may not be needed.

What do you mean by newer builds? I am using the latest 0.13 b2 but I am stuck with 16 segments on my ESP32.
32 would be Perfect for my project as I need exactly that amount.

Cuacco:
Hi
I am trying to create a smile face using a LED matrix 16x16, I am tryting to create a usermod with my code, but  I have not beena ble to find a function that let me activate individual LEDs, I tried with segments (some to keep in black (off) and other to iluminate (on)) but I reached the limit of segments... any idea to explore or advice?

blazoncek:
There is a [JSON API](https://github.com/Aircoookie/WLED/wiki/JSON-API#per-segment-individual-led-control) that allows individual pixels to be set.

PR #1968, opened 2021-05-18
Add option to enforce off state

### LINKED ISSUE #1966: Turning off bus in off state causes random LED states
**Describe the bug**
#1245 causes a heavy regression on most of my uses of NodeMCU modules. Usually the first, sometimes more LEDs of a bus start to lit randomly. I always use RG174 cables to minimize interferenca and very good meanwell power supplies. The severity of this problem depends on the LED in use, SK6812 seem to be more sensitive, the WS2812 wire kind of strip is also super sensitive and starts to lit up daily.

Either we need a option to not stop the bus, if brightness is 0, or at least send a update very second or so.
The current state causes lots of problems.

**To Reproduce**
Depends on my factors, cable length, how dirty the mains are, etc. 
This kind of strip seems more sensitive: https://www.btf-lighting.com/products/usb-rgb-led-pixels-string-christmas-lights-2m-5m-10m-rf-21keys-controller-ws2812b-ic-addressable-individually-light-10-leds-m-5v

**Expected behavior**
LEDs should be off when state is off

**WLED version**
 - Board: NodeMCU
 - Version v3
 - Format both

poelzi:
fixes #1966

This PR adds a advanced option in the LED configuration option to enforce the off state of strips. This helps if long lines or dirty mains causes stray LED states.


wled00/wled.cpp
```diff
@@ -221,7 +221,7 @@ void WLED::loop()
 
     yield();
 
-    if (!offMode || strip.isOffRefreshRequred)
+    if (!offMode || strip.isOffRefreshRequred || enforceOff)
       strip.service();
```
blazoncek, line 225: This will cause strip to be updated if `enforceOff==true` is this the intended behaviour?
How does this enforce "off" state of the strip?
When strip is updated it executes effect functions (including mode_solid()).

blazoncek:
I am thinking that the same result can be achieved by selecting black color and Solid effect, regardless of the brightness.

poelzi:
The problem is that all integrations will not work as expected, which is super annoying. Instead of using lights.turn_off with some devices I have to create an automation that selects a specific preset that is different on each device to enforce a off state.

blazoncek:
Functionality implemented with the logic used for TM1814 LED strips using UI override in LED settings..

PR #1969, opened 2021-05-18
Update to handle both 24-key and 24-key old remotes

scottrbailey:
Blazoncek pointed out that once I removed the short circuit that sent all codes between the range of 0xF70000 to 0xF80000 to decodeIR24, it wasn't represented in the switch statement.  I catch that both new and old 24-key remotes shared the same value in irEnabled.

### DIFF
--- a/usermods/JSON_IR_remote/readme.md
+++ b/usermods/JSON_IR_remote/readme.md
@@ -1,0 +1,33 @@
+# JSON IR remote
+
+## Purpose 
+
+The JSON IR remote allows users to customize IR remote behavior without writing custom code and compiling. 
+It also enables using any remote that is compatible with your IR receiver. Using the JSON IR remote, you can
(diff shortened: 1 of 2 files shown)


blazoncek:
@Aircoookie I would urge you to merge this (and original PR) feature into master ASAP. It is beyond awesome.

scottrbailey:
I would like some feedback on what to do with files in usermods. Do we want to include them or just make available from wiki?

blazoncek:
I would put a link to them on a Wiki page and make them available for download.

emreakmese:
How to make an addition like nanoleaf remote, designed in pentagon shape? It doesn't matter how many surfaces it will be. Is there a way to integrate this into WLED with components like ESP8266&NRF24l01&MPU6050 or other components?

blazoncek:
@Aircoookie bumping this.
It would be good if you merged this one in to allow proper 24-key IR handling.

PR #1970, opened 2021-05-18
Use relative paths for all links.

poelzi:
Carefully using relative links for everything and using wss in case the
site was loaded over https.
Now it is possible to use a nginx or apache to redirect to a wled
node as a subdirectory.

I'm using home-assistant as a webapp which requires that all resources are loaded through https. I embed the wled through a iframe that is served by nginx and forwards the wled traffic. This requires currently to use a new hostname for each wled node.
This patch allows also to use one host and map one path to one node.

This is the config my nixos generates and it works very well.
```
server {
        listen 0.0.0.0:443 ssl http2 ;
        listen [::]:443 ssl http2 ;
        server_name wled.example.com ;
        location /.well-known/acme-challenge {
                root /var/lib/acme/acme-challenge;
                auth_basic off;
        }
        root /var/www;
        ssl_certificate /var/lib/acme/wled.example.com/fullchain.pem;
        ssl_certificate_key /var/lib/acme/wled.example.com/key.pem;
        ssl_trusted_certificate /var/lib/acme/wled.example.com/chain.pem;
        location ^~ /devtest/ {
                proxy_pass http://wled-devtest.example.com/;
                proxy_http_version 1.1;
                proxy_set_header Upgrade $http_upgrade;
                proxy_set_header Connection $connection_upgrade;
                include /nix/store/dwprz9w34r2yfi9m1b2c9vv3br8807jb-nginx-recommended-proxy-headers.conf;
        }
}
```


poelzi:
Somebody needs to test this also, I tried to click everywhere and test all functionality, but it could easily be I missed something

blazoncek:
Could you verify and rebase this PR to use most recent master?

dbixler:
I just saw this PR and I had already done my own implementation of this exact functionality, LOL.  Pretty similar to what I did, but since you probably have a lot more experience with WLED than I do, I'll drop it and I'll see about getting this change and testing it further.  So far, it looks pretty good.

blazoncek:
This PR needs to be brought up to speed with 0_14 branch to be of any use.

Aircoookie:
A very late thank you for your PR and sorry for never getting around to it. Closing in favor of #2788 , as that is current to 0.14.

Issue #1246, opened 2020-10-11
Overview of icons available in Wicons woff font?


GrumpyMeow:
Hi,

I'm trying to find more suitable icons for sliders.
I tried to decode the font and/or find it somewhere online.
Unfortunatly i failed. 

Do you maybe have something like an overview/reference/hint for me?

Aircoookie:
Hi!
To reduce file sizes, this is a custom icon pack that only contains the icons that are used in the UI (+maybe just a few icons more I planned to use at some point)
I'm attaching the latest version of that font, I do believe it has more icons though than the woff currently used in the index but I frankly don't know at this point 😅 

The font was made with the free icomoon.io service, which is great. The icons come from a collection/font called `Material-Icons-Complete-Filled`, though I don't have the source for that anymore...

Hope it will help you at least a little. If you want more icons you'll likely have to replace the base 64 woff with a different version that contains all my icons (but likely with different IDs) plus the ones that you want to add.

[mainui-v1.0.zip](https://github.com/Aircoookie/WLED/files/5361100/mainui-v1.0.zip)

<img width="1470" alt="overview" src="https://user-images.githubusercontent.com/21045690/95679241-be143d80-0bd1-11eb-84b7-b92d045eb04b.PNG">

GrumpyMeow:
Thank you for the information. I'm working on the "Sound Reactive" fork and would like to do some "polishing". All the current sliders have the same icon.

PR #1976, opened 2021-05-20
Add brightness lock

poelzi:
I have presets with and without brightness settings, depending on the individual preset.
Late at night, I often have to disable preset rotation, because the brigtness is just too intense.

The new brightness lock flag allows to lock the current brightness level when switching presets or other brightness changes.

== TODO
[ ] add icon for stats instead of U and L labels. I don't know how the font is generated
[ ] Better styling for the disable state of the brightness slider. It kind of works in firefox, but the webkit seems to ignore the disable
pseudo state for some reason. It would be nice if the complete bar would be grayed out, not only the slider button.


blazoncek:
You can use available icons from Wicons font (i.e. `<i class="icons">&#xe38f;</i>`).
The font was recently pruned but quite a few icons still remain.
Check #1246

PR #1977, opened 2021-05-20
Multi-button support.

blazoncek:
Added multiple button support, including _inverse_ buttons (active when pin is high) and configurable touch buttons with the possibility to adjust touch threshold.
Each button can be assigned 3 (or 2 for switches) actions (presets): short press, long press and double press.
Default number of button is 2 for ESP8266 and 4 for ESP32 but can be overridden by defining WLED_MAX_BUTTONS precompiler variable.


blazoncek:
@Aircoookie you may postpone merging until we figure out how to enable selection of A0 in UI for analogue button on 8266 and warn users of pins without ADC on ESP32.
But I think these are minor nuances, not problematic for general audience.

Aircoookie:
There is only a single analog pin on ESP8266, so we could just ignore the selected pin and code it to A0. If the user attempts to setup more than analog buttons, all but the first are disabled (8266 only)

Had a read through the changes and I am very surprised about the huge change set in wled. cpp (basically the entire file), even though little or nothing changed. Have you optimized formatting here?

Will test later today :) Adalight pin 3 issue is moot IMO, even if we disable it in code if utilized by LEDs or another hardware, it doesn't stop anyone sending data via Serial and messing up the LEDs. Perhaps we should rather add a warning that Adalight is not available if pin 3 is utilized.
I'm happy to include PSRAM support too, will add a define though so it can be set at compile time. Also, I think we should un-comment the Status LED code, it will not be compiled in if `STATUS_LED` is not defined, anyways.
Thanks again for the PR, multi button is awesome!

blazoncek:
Regarding wled.cpp it might have been caused by CR/LF to LF conversion, since I am switching between Windows and Mac computers.

PR #2008, opened 2021-06-02
added SN_Photoresistor usermod

DickSwart:
I added a new photoresistor/luminance sensor usermod, it is different from `photoresistor_sensor_mqtt_v1` as it is a v2 usermod and it displays the luminance in the UI and sends MQTT messages.


blazoncek:
@DickSwart why not make this usermod runtime configurable?
You can check samples I made in Temperature, Animated Staircase, PIR Sensor Switch, Auto Save and 4 Line Display usermods.

PR #2031, opened 2021-06-18
made sn_photorisistor configarable

DickSwart:


blazoncek:
Hi Dick, we made some changes in how usermod settings are processed. During this process I had to change your usermod to comply with new data structure (it should be much simpler now).
Could you revise your usermod if it is still performing well?

I would also encourage you to add the possibility to select analog input pin on ESP32 (and ignore the setting on ESP8266).

PR #2043, opened 2021-06-27
Usermod settings v2

blazoncek:
- added POST field parsing
- simpler handling in readFromConfig()

Temperature & 4 Line Display also got some enhancements.


Aircoookie:
Very awesome! I have reverted some of the style changes, as the bigger text size in inputs make the labels too small on mobile devices. Love the larger checkboxes though.

Dallas parasite power is also a great addition, didn't test it, but nice to have :)

PR #2044, opened 2021-06-29
move common HTTP constants to separate header file

jaredgisin:
This PR clean up the code by eliminating repetitive hard-coded strings literals in the HTTP code. Compile-time values (in a new file, http.h) are provided for commonly used HTTP status codes, headers and content-types. Misuse of them will result in build time errors instead of difficult to diagnose runtime failures.

### DIFF
--- a/wled00/http.h
+++ b/wled00/http.h
@@ -19,7 +19,7 @@
 #define HTTP_HDR_ACCESS_CONTROL_MAX_AGE "Access-Control-Max-Age"
 #define HTTP_HDR_CACHE_CONTROL "Cache-Control"
 #define HTTP_HDR_CONTENT_TYPE "Content-Type"
-#define HTTP_HDR_CONTENT_ENCODING "Content_Encoding"
+#define HTTP_HDR_CONTENT_ENCODING "Content-Encoding"
 #define HTTP_HDR_ETAG "Etag"
 #define HTTP_HDR_IF_NONE_MATCH "If-None-Match"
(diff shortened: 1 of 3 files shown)


jaredgisin:
Good afternoon! I'm curious on what the policy and frequency is on when merge happen. Is there any reason not to merge this or changes that I should make to get this in?

jaredgisin:
@blazoncek Other than the fact that I am now forced to rebase this due to other changes that have since gone in. Its here any reason that this can't be merged once I update it?

blazoncek:
First of all **NEVER** add string constants to a #define. This is IMO bad practice in any case.
If needed create global (or local) static string constants (possibly using flash memory and not RAM).
I used the technique in a few usermods to reduce flash pressure and reduce amount of RAM used at the same time.

Still, in the end it is not my call. ;)

jaredgisin:
I wouldn't go so far as to say NEVER do anything. There are often good reasons to defy best practices. The values are never going to changes ever during the course of runtime, so have them be inserted into the code during the preprocessor is certainly one way that can be done. Having said that, setting these up them as string constants seems like a great alternative to me. Given I have to rebase this anyhow, I'll work on changing it.

blazoncek:
Contrary to ints and longs (and floats) strings may take a lot of storage space and depending on compiler optimisations may not get de-duplicated thus consuming even more space.

Having them declared as const char arrays and using a pointer to that const char array is the most optimised way of using strings as constants.

PR #2057, opened 2021-07-04
Adding Attack and Decay Settings

Model-Man:


Aircoookie:
Hey, I think you meant to add this as an issue in https://github.com/atuline/WLED , not as a pull request here :)

PR #2061, opened 2021-07-05
Usermod Settings polishing/documentation

embedded-creations:
Major Changes
- Add getValueFromJsonKey() wrapper function to check if value is present
- Add getValueFromJsonKey(element, destination, defaultvalue) - working, but not used as of now, can be removed if desired
- Update Usermod Settings html "number" field to use step="any", and make wider to make maximum values fully visible
  - step="any" allows viewing/submitting full 7/8-digit float values, and the arrow buttons step by 1 now, instead of .00001 (which wasn't good for integers or floats)
- Update usermod_v2_example.h with more complete example and documentation for Usermod Settings
- Update/rename usermode_rotary_brightness_color, to be used as an example of more robust parsing Usermod Settings values

LMK if I should have squashed these commits

TODO:
- [x] readFromConfig() has three options for how to load values from the config JSON, we need to pick one
  - Number 1 was your preference before you saw the three examples.  Number 2 is my preference after seeing the three examples
- [ ] html hasn't been generated/compressed yet


Aircoookie:
Thank you!

Thinking about it, I think we should go with option number 2 (your prefered) as the default for the example usermod. 1 is fine if no complete validation is necessary, but I like about your solution that the variable doesn't have to be repeated, which could be error-prone.
`getValueFromJsonKey()` thus is similar to the `CJSON` macro I'm using in `cfg.cpp`, only that that is only syntactic sugar for the `|` operator and does not explicitely check for presence of the value.

The optional check in number 1 and 3 (`top["a"].isNull() || top["b"].isNull() || top["c"].isNull()`) is horrible. Number 2 syntax is a lot nicer.

I do have some small suggestions:
- I believe the three-argument `getValueFromJsonKey()` to set a default value if not present is a great idea. The example usermod should perhaps show this capability for half of the values.
- The `Serial.println()` in  `getValueFromJsonKey()` in case of a missing value should be placed in flash `Serial.println(F("element.isNull()"));`
- the function name `getValueFromJsonKey()` is a bit long to type/remember. Although less concise, `getValueFromKey()`, `getJsonValue()`, `getKeyValue()`, `getValueFromJson()` or `getKey()` might be good alternatives.

embedded-creations:
Thanks for the feedback, I made your suggested changes.  I removed the printf which was just meant for my debugging

embedded-creations:
The HTML still needs to be generated/compressed!

Issue #2056, opened 2021-07-04
Multiple pin LED outputs are not persistent across reboot if they exceed total LED count


nolanhergert:
**Describe the bug**
I am able to get multiple LED strips (and respective segments) set up and working. I made sure to save to preset 16. When I AC cycle, the segment data comes back up, but only the first of the LED outputs on the LED Preferences page come back up. All other settings on that page persist fine.

**To Reproduce**
Go to LED Preferences. Add a 2nd LED strip, hit save. Observe the 2nd LED strip is lit up. AC cycle. Observe 1st but not 2nd LED strip is lit up and only the first LED strip is shown on the LED Preferences page.

**Expected behavior**
Expect the 2nd LED strip to light up and be on the LED preferences page.

**WLED version**
 - Board: ESP32 w/ custom breakout and ESP32 dev board
 - Version v0.13.0-b0 "Toki"
 - Format: Binary

**Additional context**
Everything is working except the persistence part! Thanks for your help!

nolanhergert:
Worked around by ensuring that Total LED count (on top of LED preferences page) is set to a value greater than the sum of the LED strips. 

A few options:
  * Update "Total LED count" value automatically to the sum of the LED strip lengths before submission
  * Make error saying the user needs to update this "Total LED count" value to be what is supported by their power supply/microcontroller. However, I would argue this should be handled by "maximum current" field and not LED count.

nolanhergert:
Created a pull request to facilitate discussion on solution #1 above: https://github.com/Aircoookie/WLED/pull/2065.

It is working for me when I add this line manually to html_settings.h, regenerate with npm, and re-upload to ESP32 dev board. No clue if it messes up other things though!

blazoncek:
This is something @Aircoookie explicitly does not want (ATM). There is a (legitimate) reason why it is currently possible to have discrepancy between actual amount of LEDs (sum of pixels on strips) and a total count in LED settings page.
It may be changed in the future when current reason has a workaround.

nolanhergert:
While I am ok with there being a discrepancy (I'm curious what the reason is, just for my understanding), part of the problem is that the code accepts a discrepancy initially (displays on the 2nd LED strip despite it being over Total LED count) but then rejects the discrepancy upon rebooting. It <understandably> looks like the checks are not consistently applied.

Aircoookie:
Hey! Thank you for the PR as well.
There are two reasons a discrepancy is possible:
[...]

PR #2065, opened 2021-07-06
Auto-update total LED count from LED output lengths

nolanhergert:


nolanhergert:
Not the right implementation according to #2056

PR #2066, opened 2021-07-06
Bug 2064, 2063 - PinManager usage

### LINKED ISSUE #2063: Bus manager classes can accidentally free pins it didn't successfully allocate
**Describe the bug**

Bugs in bus manager classes cause de-allocation of pins already allocated to other resources.

**To Reproduce**

Compile with some pins in use.  For example:
`-D IR_PIN=0  -D BTNPIN=1 -D RLYPIN=2`

Edit exported JSON to add LEDs, such that the `BusConfig` structure matches the following array of three busses:

```C++
BusConfig bc[] {
    // type, count, start, colorOrder, reversed, skip, pins[5]
    { 41, 5, 0, 0, false, 0, {  3,  4,  5,  6,  7 } }, // 5-channel analog
    { 50, 5, 0, 0, false, 0, {  8,  9, -1, -1, -1 } }, // WS2801
    { 51, 5, 0, 0, false, 0, { 10, 9, -1, -1, -1 } }, // APA102
    { 45, 5, 0, 0, false, 0, {  3,  4,  5,  6, -1 } }, // 4-channel analog
};
```

**Expected behavior**

* 5-channel analog succeeds, allocates pins 3..7.
* WS2801 succeeds, allocates pins 8 & 5.
* IR, button, relay, WS2801, and 5-channel analog retain their allocated channels.
* APA102 fails, because at least pin 9 already allocated to WS2801.
* 4-channel analog fails, because at least pin 3 already allocated to 5-channel analog.

**Actual behavior**

* For APA102, `BusDigital` constructor [fails to allocate pin 9 at line 114](https://github.com/Aircoookie/WLED/blob/8c44147a4596112a7f5d5720ae809949399e9d44/wled00/bus_manager.h#L113-L115).  Therefore, `cleanup()` is called, which [de-allocates pin 9 at line 198](https://github.com/Aircoookie/WLED/blob/8c44147a4596112a7f5d5720ae809949399e9d44/wled00/bus_manager.h#L198) ... **_even though pin 9 was already allocated to WS2801_**.

* For 5-channel analog, `BusPwm` constructor [fails to allocate pin 3 at line 233](https://github.com/Aircoookie/WLED/blob/8c44147a4596112a7f5d5720ae809949399e9d44/wled00/bus_manager.h#L232-L233).  It therefore calls `deallocatePins()`, which  [frees pin 3 at line 324](https://github.com/Aircoookie/WLED/blob/8c44147a4596112a7f5d5720ae809949399e9d44/wled00/bus_manager.h#L324) ... **_even though pin 3 was already allocated to 5-channel analog_**.

**WLED version**

 - Board: D1 Mini clone, any board
 - Version: All versions since commit 9518c5f2e4440f93384c5d706c28e99e7f12eb84
 - Format: code review

**Additional context**

I expect to have a fix soon, will submit PR then.

Thank you for your help!

### LINKED ISSUE #2064: Ethernet support does not allocate pins
**Describe the bug**

Ethernet configuration does not check if the pins are already allocated, and does not allocate pins to prevent other use.

**To Reproduce**

Use ethernet-enabled binary.
Send configuration that sets relay to pin 12.
Click power button in UI

**Expected behavior**

LEDs turn off, but not the board/ethernet.

**Actual behavior**

See [thread](https://github.com/Aircoookie/WLED/issues/1027#issuecomment-723481641) ... powers off because ethernet [did not reserve](https://github.com/Aircoookie/WLED/blob/8c44147a4596112a7f5d5720ae809949399e9d44/wled00/wled.cpp#L438-L448) the pins it needs.

**WLED version**
 - Board: ESP32-POE
 - Version: 0.12.0
 - Format: binary

**Additional context** 

Also would occur on other ethernet boards. Fix may be easy.  I will see about adding a PR, but will need help testing it.

Thank you for your help!

henrygab:
### CURRENTLY NOT BUILT OR TESTED

However, ready for review and discussion of proper way to fix.

Fixes #2063 - Bus Manager classes `free()`ing pins it shouldn't

The fix is to only store the pin number in the private class variable **_after_** allocation succeeds.

Fixes #2064 - Ethernet support should allocate pins using `PinManager` class

The fix is to attempt to allocate the pins, and track the ones that were successfully allocated, so can safely roll-back if all pins were not allocated.  Only when all are successfully allocated, then initialize ethernet.


wled00/bus_manager.h
```diff
@@ -107,13 +107,15 @@ class BusDigital : public Bus {
   public:
   BusDigital(BusConfig &bc, uint8_t nr) : Bus(bc.type, bc.start) {
     if (!IS_DIGITAL(bc.type) || !bc.count) return;
-    _pins[0] = bc.pins[0];
-    if (!pinManager.allocatePin(_pins[0])) return;
+    uint8_t pin = bc.pins[0];
```
blazoncek, line 110: Unnecessary. You can `allocatePin(bc.pins[0])` and if successful assign `_pins[0] = bc.pins[0];`

wled00/bus_manager.h
```diff
@@ -229,10 +231,11 @@ class BusPwm : public Bus {
     #endif
 
     for (uint8_t i = 0; i < numPins; i++) {
-      _pins[i] = bc.pins[i];
-      if (!pinManager.allocatePin(_pins[i])) {
+      uint8_t currentPin = bc.pins[i];
```
blazoncek, line 234: Same as above.

wled00/wled.cpp
```diff
@@ -437,14 +437,64 @@ void WLED::initConnection()
...
-      (int)     es.eth_mdio, 
-      (eth_phy_type_t)   es.eth_type,
-      (eth_clock_mode_t) es.eth_clk_mode
-    );
+    // Use PinManager to ensure pins are available for
+    // ethernet AND to prevent other uses of these pins.
+    bool s = true;
+    byte pinsAllocated[4] { 255, 255, 255, 255 };
+
+    if (s && allocatePin((byte)es.eth_address)) {
```
blazoncek, line 445: eth_address does not constitute a pin.

wled00/wled.cpp
```diff
@@ -437,14 +437,64 @@ void WLED::initConnection()
...
+        s = allocatePin(16);
+        break;
+      ETH_CLOCK_GPIO17_OUT:
+        s = allocatePin(16);
+        break;
+      default:
+        s = false;
+        break;
+    }
+
```
blazoncek, line 482: You can use logic `if (s) {s &= pinManager.allocatePin(x)); ...}` for a shorter notation.

blazoncek (commented):
You will need to pass that information to LED settings too. To prevent user assigning Ethernet pint to a LED channel, button relay etc.
The issue with usermods remain, though...

wled00/wled.cpp
henrygab: Good catch.  I was just going through the LAN8720, and realized this today.  Fixed with f740d48.

wled00/bus_manager.h
henrygab: True.  My style comes from programming in adverse environments, where buffers can (and are) changed beneath you, causing Time-of-Check/Time-of-Use bugs.

Doesn't apply here, so I've updated the code per your comments in bef9711.

wled00/bus_manager.h
henrygab: Same as above.

wled00/wled.cpp
henrygab: I think I would need to short-circuit the evaluation, so it would end up being:

```C++
if (s && (s = pinManager.allocatePin(x))) {
    pinsAllocated[y] = x;
}
```

I feel assignment statements within `if` conditions may be considered an arcane coding style.  I prefer to spend a couple more lines of code, if it makes the code easier to grok for later maintainers, since it compiles out the same. 

However, to smooth the review, I've made this change in 497c605.

blazoncek:
Thinking about the code a bit more, Ethernet pins should be allocated _prior_ to any other pins since adding an Ethernet board and disabling it by using its pins for other purposes is counter-productive.
It should be other HW stuff that will need to take into account that pins have been allocated.
You can see a sample of reservations in my fork (xml.cpp), unfortunatelly this still doesn't work for usermods settings since pin reservations are read from cfg.json directly. The solution would be to store pins used by Ethernet board in cfg.json file together with other eth information.
This piece of code should take care for usermods:
```
  #ifdef WLED_USE_ETHERNET
  JsonObject ethernet = doc.createNestedObject("eth");
  ethernet["type"] = ethernetType;
  JsonArray pins = ethernet.createNestedArray("pin");
  if (ethernetBoards[ethernetType].eth_power>0)    pins.add(ethernetBoards[ethernetType].eth_power);
  if (ethernetBoards[ethernetType].eth_mdc>0)      pins.add(ethernetBoards[ethernetType].eth_mdc);
  if (ethernetBoards[ethernetType].eth_mdio>0)     pins.add(ethernetBoards[ethernetType].eth_mdio);
  switch (ethernetBoards[ethernetType].eth_clk_mode) {
    case ETH_CLOCK_GPIO0_IN:
    case ETH_CLOCK_GPIO0_OUT:
      pins.add(0);
      break;
    case ETH_CLOCK_GPIO16_OUT:
      pins.add(16);
      break;
    case ETH_CLOCK_GPIO17_OUT:
      pins.add(17);
      break;
  }
  #endif
```

henrygab:
> You will need to pass that information to LED settings too. To prevent user assigning Ethernet pint to a LED channel, button relay etc.
> The issue with usermods remain, though...

<HR/>
<details><summary>I thought the settings code allocated pins for items?</summary><P/>

When I was tracing how things get applied, I saw that the configuration setting code generally attempted to allocate pins:

#### set.cpp handling of POST settings

* Old values for the relay pin, IR pin, and buttons get de-allocated [in set.cpp lines 80-84](https://github.com/Aircoookie/WLED/blob/14b7ec2a80b8697d99cde2fe08ca41ed361e1413/wled00/set.cpp#L80-L84).
* Pins for LEDs get [put into a `BusConfig` structure](https://github.com/Aircoookie/WLED/blob/14b7ec2a80b8697d99cde2fe08ca41ed361e1413/wled00/set.cpp#L121) without calling `allocatePin()`, but the `BusDigital` and `BusPwm` classes will later allocate the pins internally before they use them.
* IR pin is allocated at [line 130](https://github.com/Aircoookie/WLED/blob/14b7ec2a80b8697d99cde2fe08ca41ed361e1413/wled00/set.cpp#L).
* Relay pin is allocated at [line 138](https://github.com/Aircoookie/WLED/blob/14b7ec2a80b8697d99cde2fe08ca41ed361e1413/wled00/set.cpp#L138).
* Button pins are allocated at [line 149](https://github.com/Aircoookie/WLED/blob/14b7ec2a80b8697d99cde2fe08ca41ed361e1413/wled00/set.cpp#L149).

#### cfg.cpp handling of configuration file's settings

* Pins for LEDs get [put into a `BusConfig` structure](https://github.com/Aircoookie/WLED/blob/14b7ec2a80b8697d99cde2fe08ca41ed361e1413/wled00/cfg.cpp#L118) without calling `allocatePin()`, but the `BusDigital` and `BusPwm` classes will later allocate the pins internally before they use them.
* Pins for buttons get allocated at [line 134](https://github.com/Aircoookie/WLED/blob/14b7ec2a80b8697d99cde2fe08ca41ed361e1413/wled00/cfg.cpp#L134).
* Pin for IR is allocated at [line 170](https://github.com/Aircoookie/WLED/blob/14b7ec2a80b8697d99cde2fe08ca41ed361e1413/wled00/cfg.cpp#L170).
* Pin for relay is allocated at [line 181](https://github.com/Aircoookie/WLED/blob/14b7ec2a80b8697d99cde2fe08ca41ed361e1413/wled00/cfg.cpp#L181).

</details>
<HR/>

Open questions:
* When (if ever) should ethernet pins be de-allocated?
* In cfg.cpp, button 0 defaults [are alluded to](https://github.com/Aircoookie/WLED/blob/14b7ec2a80b8697d99cde2fe08ca41ed361e1413/wled00/cfg.cpp#L155), but is the default pin ever allocated?

When you refer to the issue with usermods, are you referring to the fact that they need updating to use the pinManager to allocate/deallocate pins?  (It appears that they should have access to the pinManager object...).

blazoncek:
>     * When (if ever) should ethernet pins be de-allocated?
> 
> When you refer to the issue with usermods, are you referring to the fact that they need updating to use the pinManager to allocate/deallocate pins? (It appears that they should have access to the pinManager object...).

Pin allocations happen in cfg.cpp and usermods at their discretion.
At each change.
Yes.

Everything that is runtime configurable has to have allocation/deallocation logic.
All of the code is already implemented in my fork, dev branch (except for the ethernet change deallocation which will require a bit more thought).

wled00/cfg.cpp
```diff
@@ -153,14 +153,21 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
     }
   } else {
     // new install/missing configuration (button 0 has defaults)
-    if (fromFS)
-      for (uint8_t s=1; s<WLED_MAX_BUTTONS; s++) {
+    if (fromFS) {
+      // relies upon only being called once with fromFS == true, which is currently true.
+      uint8_t s = 0;
+      int8_t pin = btnPin[0]; // initialized to #define value BTNPIN, or zero if not defined(!)
```
blazoncek, line 159: No need for new variable (unnecessary stack & CPU cycle use).
Otherwise another good catch!

blazoncek (approved):

wled00/cfg.cpp
henrygab: Sorry, habits from years of defensive coding, where user-supplied buffers were potentially maliciously modified between time-of-check, and time-of-use.  Removed new variable in 42993e8.

henrygab:
I'm wondering if I should be looking at (and submit PRs to) your fork, instead. 😃 

The ethernet change de-allocation is trickier, as you wouldn't want to de-allocate the pins when the config doesn't change.  I'm not even sure the config should be allowed to change ... I haven't dug deep or run tests, but at first look, the ethernet library does not appear to support repeated calls to `begin()` ... potentially leaking memory (e.g., [this `esp_eth_phy_t` structure](https://github.com/espressif/arduino-esp32/blob/c37557c711ba9cd93dd4f37cef0db226e03f7912/libraries/WiFi/src/ETH.cpp#L271)).  Again, this is just a first look, but ... perhaps ethernet pin settings should only change with reboot?

blazoncek:
Indeed.
The code forces reconnect when wifi settings change. Mabe a reboot would be in order instead?
@Aircoookie thoughts?

henrygab:
> The code forces reconnect when wifi settings change. Mabe a reboot would be in order instead?

I am starting to feel strongly that having ethernet initialization occur so often is not a good thing, so will open a new bug on that topic.

On the plus side, I don't think ethernet will break with the present code, even though there is no code to de-allocate ethernet pins.  The first configured ethernet type will continue to work, until the user changes to another ethernet type.

Aircoookie:
Tested and working well!

Aircoookie:
@henrygab I agree that ethernet initialization should only occur once. However, this must not necessarily be at boot, since if Ethernet type was set to `None` before, it is fine to init.
Changing Ethernet modes is very unlikely, it would only occur if a user accidentally selected the wrong mode. I do not think that it is essential to handle such an edge case, the behavior you currently implemented seems sensible :)

blazoncek:
Turns out that Ethernet uses (and requires) a few more GPIOs than those specified in ethernetBoards[] array.

Official documentation:
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/hw-reference/esp32/get-started-ethernet-kit-v1.1.html#ip101gri-phy-interface

henrygab:
@blazoncek ... Yikes!  Sounds like a few more pins will need to be marked as "in use" if enabling ethernet on ESP32 boards.  Luckily, that should be fairly straightforward to add.  According to the [schematic](https://raw.githubusercontent.com/ldijkman/WT32-ETH01-LAN-8720-RJ45-/main/WT32_ETH01_V2.schematic.pdf), the following ESP32 pins are connected to to the LAN8720:

Pin | Name | LAN8720 | Notes
-----|-----|-----|-----
IO21 | `RMII_EMAC_TX_EN` | TXEN | Matches your link
IO19 | `RMII_EMAC_TXD0` | TXD0 | Matches your link
IO22 | `RMII_EMAC_TXD1` | TXD1 | Matches your link
IO25 | `RMII_EMAC_RXD0` | RXD0/Mode0 | Matches your link
IO26 | `RMII_EMAC_RXD1` | RXD1/Mode1 | Matches your link
IO27 | `RMII_EMAC_CRS_DV` | CRS_DV/Mode2 | Matches your link
IO0 | `RMII_EMAC_REF_CLK` | nINT/REFCLK0 | Uses an external 50MHz oscillator
IO23 | `RMII_MDC` | MDC | Matches your link
IO18 | `RMII_MDIO` | MDIO | Matches your link
IO13 | `RMII_EMAC_RX_ER` | RXER/PHYAD0 | **_What is this?  Prevents use of JTAG... 😢_** 
IO16 | `OSC_EN` | n/a | Enables the external 50MHz oscillator

Unfortunately, I'm still having zero luck getting my WT32-ETH01 boards to connect via ethernet at the moment using anything but their stock firmware.  Hoping to have time to debug this weekend....

henrygab:
OK, I setup a PR that should handle these additional pins.  Testing is appreciated, so I can move it out of draft status?

blazoncek:
Will try later this weekend or next week.

henrygab:
@blazoncek -- FYI, I have just made two significant bug fixes.

[First](https://github.com/Aircoookie/WLED/pull/2094/commits/ffa19f6e1e9da74315f55e83598be395c711e70b), I fixed the inverted logic for whether the pin needed to have output capability.  No actual effect given the current pins used, but would have been a latent, lurking bug for future developers.

Second, I had a copy/paste error, where I didn't save the actual pins allocated.  No actual effect in testing, unless an allocation failed.  If that occurred, it would "leak" allocated pins.

henrygab:
I didn't like the complexity of the roll-back code.  It worked, but was not clean, and would be an error-prone area for other clients of pin manager.

I have an [alternate PR](https://github.com/Aircoookie/WLED/pull/2095) that exposes a multi-pin allocation function.  That function either succeeds for **_all_** pins, or it allocates **_none_** of the pins.  That function also ignores pins set to `-1`, to simplify the ability to generate constant arrays of configuration that include optional pins.

This is useful anytime a function requires more than one pin, including rotary encoders, I2C, SPI, and of course ethernet.

PR #2067, opened 2021-07-06
Esp32s2 support

blazoncek:
Updated Wemos Shield parameters.


Aircoookie:
Thank you :)

WLED running on the S2, that's very exciting!
It's far from running well though 😅 all still a bit early stuff as it seems. 97% flash usage (but with the 1MB FS partition table!), only 27k free heap and the filesystem is not mounting for me. I need to look into that some more.

FYI, not only does the Esp32S2 only have 1 I2S, it also has only 4 RMT and 8 LEDC. So we'll need to set even more limits soon.

blazoncek:
> FYI, not only does the Esp32S2 only have 1 I2S, it also has only 4 RMT and 8 LEDC. So we'll need to set even more limits soon.

Indeed this part slipped my attention when merging latest changes in bus_wrapper (NeoBusChannel).

PR #2068, opened 2021-07-07
New usermod: Support for RGB Rotary Encoder Board

ezcGman:
Hey there,

I've spent the last days creating my first usermod :) And guess what: **It's another rotary encoder usermod!!** ;) But not a usual one: This mod is for this awesome DIY LED rotary encoder board from Adam Zeloof: https://github.com/isotope-engineering/RGB-Encoder-Board:

https://user-images.githubusercontent.com/3090131/124680599-0180ab80-dec7-11eb-9065-a6d08ebe0287.mp4

So this little PCB has 20 LEDs on it and what my usermod essentially does is:
- It will always reflect the current brightness in basically percentage on the board using the LEDs :) It will of course also update if the brightness is updated from anywhere except the rotary encoder (Web UI, app, API, MQTT...).
- It features three different modes:
  - An awesome rainbow color circle (mode 3)
  - Just a single red LED going around (mode 2)
  - Or like mode 3 but just a single color (obviously I went with WLED orange!)
- Everything can be configured from the WLED Web UI Usermod settings page, incl. pins, brightness, mode, etc...

There's only one known issue which I just can't get fixed:
Sometimes there seems to be a race condition on startup of WLED, where the LED ring will not be updated. I already tried listening to "canShow" of the LED Bus and make sure to update in the loop() and not in setup()... But it's really not happening often and only on boot. And it fixes itself by just making an update, like turning the knob, change the brightness in the web UI, change the settings, ...

Of course, I'm seeking for feedback here but especially around this:

I made use of the `BusDigital` class to control the LEDs. I tried to not mess with the configured outputs / hardware in the LED Preferences, so I went for using the "last" available LED bus possible, using `WLED_MAX_BUSSES` to detect that. It's the first time I'm messing with LEDs in WLED, so I hope I didn't do sth wrong and mess with something somehow. I tested effects, solid colors, changing outputs... All seem to still work fine.
I also decided against using the BusManager, as this would make these 20 LEDs/the board show up under the hardware section, count towards the total LEDs, will be controlled by WLED/effects, users can mess around with the config there, etc. etc. So I thoight that is not the right way to to, but to simply create an instance of the DigitalBus class.

Yeah, hope this can get merge some day soon and happy for your feedback :)

Greetings,

Andy!

### DIFF
--- a/usermods/rgb-rotary-encoder/readme.md
+++ b/usermods/rgb-rotary-encoder/readme.md
@@ -18,18 +18,18 @@ ESP32:
 ```
 [env:custom_esp32dev_usermod_rgb_encoder_board]
 extends = env:esp32dev
-build_flags = ${common.build_flags_esp32} -D RGB_ROTARY_ENCODER
-lib_deps = ${env.lib_deps}
+build_flags = ${common.build_flags_esp32} -D WLED_RELEASE_NAME=ESP32 -D RGB_ROTARY_ENCODER
+lib_deps = ${esp32.lib_deps}
     lennarthennigs/ESP Rotary@^1.5.0
 ```
 
 ESP8266 / D1 Mini:
 ```
-[env:custom_d1_mini_usermod_dallas_temperature_C_multi_relay]
+[env:custom_d1_mini_usermod_rgb_encoder_board]
 extends = env:d1_mini
 build_flags = ${common.build_flags_esp8266} -D RGB_ROTARY_ENCODER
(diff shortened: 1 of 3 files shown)


ezcGman:
Just for completion: Already got some feedback on Discord that I gonna apply:
Messages start here: https://discord.com/channels/473448917040758787/718943978636050542/862261478862487553

ezcGman:
Applied some feedback I got on Discord

PR #2071, opened 2021-07-11
Use MQTT to stream data to multiple devices

DasBasti:
I built a project where it is necessary to send a "picture" to multiple distributed devices. I used an MQTT broker to do that and added a usermod to allow to send binary color values to the strip over the MQTT broker and group channel.


blazoncek:
MQTT is not very convenient for time critical operations.
Why not rely on UDP? Need 100% delivery even if that means out-of-band delivery?

DasBasti:
I choose MQTT because this is used to send a static picture to a bunch of devices all over the internet at the same time.

It is not meant to be for real time animations.

blazoncek:
Did you test this on ESP8266 with at least 1200 LEDs and 12 segments defined?

DasBasti:
no, only ESP32 with 64 LEDs

blazoncek:
> no, only ESP32 with 64 LEDs

Please do since I am currently hard pressed for time.
And observe available memory (heap).
You don't need actual 1200 LEDs attached to ESP, just define them in LED Setting page.

DasBasti:
Tested with ESP8266 with MQTT_MAX_PACKET_SIZE=1024*4 and 1000 LED: Heap free 14.8kB
ESP32 1000LED: Heap Free 177.2kB

There was no animation running. Should I test with running animation?

blazoncek:
15k is a quite low on the heap size.
For stable run there should be at least 18k free at any time.
How much do you get with 600 LEDs? And yes, create a large number of segments and run Fireworks Starburst effect on all of them.

PR #2084, opened 2021-07-19
Upload files & skinning

blazoncek:
Backup & restore configuration & presets, for those rare cases. :)
External holidays.json so extending _Easter egg_ backgrounds can be done by user.
Ability to upload:
 - `ir.json` with JSON codes for IR remote
 - `skin.css` with CSS classes to replace default/built-in ones
 - `holiday.json` with JSON array of holiday dates
 - `cfg.json` if configuration needs to be restored
 - `presets.json` for transferring presets from one instance of WLED to another
Added ability to change "Stop LED" to "LED count" for segments in UI customization.


wled00/data/index.js
```diff
@@ -233,7 +234,7 @@ function onLoad() {
     });
 	} else
 		loadBg(cfg.theme.bg.url);
-	loadSkinCSS('skinCss');
+	if (cfg.comp.css) loadSkinCSS('skinCss');
```
blazoncek, line 237: Unfortunately using a local storage option will prevent using skinned UI on new installs.
Which is the idea behind skinning to have a distinctive look on 1st boot (prepared in advance WLED installs).

Aircoookie: There is no (easy) way to pre-install a `skin.css` yet.
However, I agree that a new client device should use the file by default, if available.
For now I'll enable the option by default, though I'll be looking into a way to save an unnecessary 404 request in the (probable) case a skin.css is not present.

wled00/wled_server.cpp
```diff
@@ -15,6 +15,22 @@ bool isIp(String str) {
   return true;
 }
 
+void handleUpload(AsyncWebServerRequest *request, const String& filename, size_t index, uint8_t *data, size_t len, bool final){
+  if(!index){
+    request->_tempFile = WLED_FS.open(filename, "w");
```
henrygab, line 20: The PR suggests only five filenames to be allowed for upload.

If I am correct that any client could call this upload endpoint, then it appears that this opens the ability to upload *arbitrary* filenames (and contents) via this endpoint.

For example, is there anything to prevent uploading `/wsec.json` using this new upload interface? _[[using `curl`, not the WLED UI]]_

Should this code validate the filename matches one of a list of allowed filenames?

wled00/wled_server.cpp
```diff
@@ -15,6 +15,22 @@ bool isIp(String str) {
...
+    DEBUG_PRINT("Uploading ");
+    DEBUG_PRINTLN(filename);
+    if (filename == "/presets.json") presetsModifiedTime = toki.second();
+  }
+  if (len) {
+    request->_tempFile.write(data,len);
+  }
+  if(final){
+    request->_tempFile.close();
```
henrygab, line 29: It appears possible to destroy the configuration, if either no data is sent, or the transfer is interrupted before completed (e.g., WiFi hiccup causes disconnect at inopportune time.)

Does the WLED_FS interface allow renaming an existing file?

If so, I would like to recommend a slightly more complex, but fail-safe, method:

1. For each allowed file, pre-define an intermediate file name that is used for the upload.  (e.g., for `/presets.json`, name could be `/presets.json.uploading`)
2. Upload only goes to these intermediate filenames
3. Optionally validate the contents (function pointer for each allowed filename)
4. Delete the target file from the file system (e.g., `/presets.json`)
5. Rename the uploaded file (e.g. `/presets.json/uploading` --> `/presets.json`)

In addition, at boot, perform the following steps for each target filename in the allowed upload list:
1. If the intermediate file exists:
    A. If the target filename exists, delete the intermediate file (failed or interrupted upload)
    B. Else (target filename does not exist), rename the intermediate file to the target filename

In this way, the upload process will have many fewer failure points.

henrygab (commented):
Some questions (concerns?) on this recent change...

wled00/wled_server.cpp
```diff
@@ -15,6 +15,22 @@ bool isIp(String str) {
   return true;
 }
 
+void handleUpload(AsyncWebServerRequest *request, const String& filename, size_t index, uint8_t *data, size_t len, bool final){
+  if(!index){
+    request->_tempFile = WLED_FS.open(filename, "w");
+    DEBUG_PRINT("Uploading ");
+    DEBUG_PRINTLN(filename);
```
henrygab, line 22: Wouldn't it also allow re-writing the HTML pages that are being served?  While I don't want to sound alarmist, I don't think this should be allowed (at least for non-debug builds).

wled00/wled_server.cpp
blazoncek: Indeed it could be possible to upload _any_ file. Though internal files are never overwritten (and cannot be).
If you want to upload wsec.json you can, but this does not pose a security issue since you cannot read it.

Will think about a possible solution.

wled00/wled_server.cpp
blazoncek: Internal files are stored in flash memory and cannot be overwritten.

wled00/wled_server.cpp
blazoncek: It could be possible as it is possible to remove files currently using /edit page. Nothing new here.
Assuming you use Backup & Restore functions, normal behavior would be to first back-up configuration (and later restore it if needed).
If there is an adversary trying to sabotage your WLED (assuming you let him/her into your network) he/she can easily do it even without backup/restore functions.
If you have WLED on a publicly accessible network, ... well, he/she can wreck havoc just changing effects or overwriting/deleting presets even with password lock.

Nevertheless I agree on your comments and suggestions.

wled00/wled_server.cpp
henrygab: I hadn't realized the internal files cannot be overwritten.  That's great, as it prevents changing the wifi settings page HTML (for example) to submit wifi credentials elsewhere.  Given that, uploading of any file is not really an issue.  👍

PR #2092, opened 2021-07-23
Implementation of OPC Protocol

fnirino:
Tested implementation of [Open Pixel Control](http://openpixelcontrol.org/) protocol for setting pixel colors.

Test setup: NodeMCU v1.0 (ESP8266-12) with 24 LED matrix

**Highlights:**
- Receives 60FPS without dropping any packet with a 30 frame buffer
- Fixed framerate using same "delay" as E1.31 implementation
- Displays correct message in "Receiving live data" page
  
**Improvements I've thought of (but don't know how feasible / useful they are):** 
- Web page error toast on too many dropped packets.
- Getting a SysEx ID to set brightness, effects, palettes, etc.
- Web settings for listening channel and port
  
**Notes**
Only supports _one client at a time_ and listens on _one channel_ (besides from broadcast)
Default port is 7890 (as per opc test scripts default port)
  

Open to suggestions and feedback!

## Summary by CodeRabbit

* **New Features**
  * Added Open Pixel Control (OPC) protocol support for realtime LED control
  * OPC server listens on port 7890, accepting a single client connection
  * Supports OPC channels 0 and 1 for RGB LED data control


wled00/wled.h
```diff
@@ -37,6 +37,7 @@
 #ifndef WLED_DISABLE_WEBSOCKETS
   #define WLED_ENABLE_WEBSOCKETS
 #endif
+#define WLED_ENABLE_OPC
```
blazoncek, line 40: You could add #ifndef WLED_DISABLE_OPC to allow complation without OPC. Similar as a line above.

Aircoookie:
This still looks potentially interesting. While I don't think realtime over TCP makes enough sense to add this by default, an OPC usermod might be a great idea.

softhack007:
@coderabbitai review

netmindz:
It seems a bit of a dead protocol that didn't really take off, folk can just use ArtNet, sACN or DDP

PR #2095, opened 2021-07-24
Fix ethernet pin allocation defaults, simplify pinManager usage

henrygab:
This PR provides the following:
* significantly improved debuggability for pin allocations (in debug builds)
* Default LED pin update for esp32-eth builds (Pin 16 was used by ethernet on WT32-ETH01)
* PinManager functions to allocate multiple pins in an "all-or-nothing" form

The last greatly simplifies error handling logic for any code or user mod that needs more than one pin.  Example hardware that this would apply to includes rotary encoders, I2C, SPI, and of course ethernet.

This builds cleanly.  Ethernet was tested only on WT32-ETH01 boards.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -292,7 +292,7 @@ board = esp32-poe
 platform = espressif32@2.0
 upload_speed = 921600
 build_unflags = ${common.build_unflags}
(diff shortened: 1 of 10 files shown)


henrygab:
Closing PR for now, as blazoncek is strongly against a change of default IO.

Will instead work on changing the code to **_only_** initialize ethernet at startup ... never without a reboot ... and to do so prior to other configuration options potentially allocating pins.  May need help with the web pages, but ... we'll see.

PR #2101, opened 2021-07-28
Fix for missing off-only MQTT messages.

blazoncek:
If _off-only_ option was selected in PIR usermod settings, MQTT messages were not sent if WLED was on when PIR sensor detected motion.
This fixes it.


blazoncek:
I have tested this now for quite some time and I think it can be safely merged.

PR #2109, opened 2021-07-30
Clear localStorage when restoring presets.

blazoncek:
Delete pmtLS and wledP localStorage items on presets restore.
This enables re-loading of new presets.


Aircoookie:
Copying my earlier discord message here for reference:

I noticed this inconvenience too, so I had already implemented an ESP-side fix . If a new presets.json is uploaded, it will update the presetModifiedTime, prompting the UI to refetch the new presets.json on load. Can you confirm that on master, it still shows the old preset cache if a new file was uploaded? https://github.com/Aircoookie/WLED/blob/815940913b6d237e7f3374f44650aed7b0b53bda/wled00/wled_server.cpp#L23

blazoncek:
Actually your solution is a better fix since it allows other clients to update cache.
It may be good to update ETag too, which I did in a separate change which I'll make another PR.
I think we can close this one without merging.

blazoncek:
Unfortunately I have bad news regarding changing "pmt" upon upload.
I have an old iPad I occasionally use for testing and it looks like I used it at some point to access my test ESP. Even when reloading a few times the locally stored presets were still active and I was unable to remove them from iPad.
Now I am thinking about adding a button to the UI Settings page to force clean local storage.

blazoncek:
Tested with button in UI settings and now I can clear wledP, wledPmt and wledPalx on every device.

Aircoookie:
I'd imagine such button to be of limited use when restoring presets, while it works, it is basically just a manual workaround for the automatic cache invalidation (in the form of a new `pmt`) and would be required on each client UI that previously controlled the device. Can be done automatically with a new file upload on the device used to upload the new presets.json (as implemented in your PR here), but that doesn't help for other clients. Still, if there are edge cases like your iPad, a force clean button may not be a bad idea.

Regarding the iPad, IIRC you had played with various caching strategies (e.g. `max-age`), could it be possible that the iPad still uses a cached version of the UI HTML which somehow doesn't implement pmt validation?
You could also test:
- Look for a preset that is present in the UI on the iPad
- Delete said preset from a different device and refresh iPad UI
- Good: The preset should be gone (fresh preset.json fetched)
- Bad: The preset is still there, ipad is using local version erroneously. Attempting to load will cause error 12.

blazoncek:
I do agree that a button is a "band-aid" to a deeper problem.
However it does solve the issue a regular user may stumble upon and is at least a last resort option that will definitely solve the problem.

BTW I did clear website data and killed Safari but somehow old presets still existed until cleared.

PR #2123, opened 2021-08-07
Ethernet configuration fix, improve PinManager

henrygab:
This PR provides the following:

* Ethernet Configuration Fixes
  * Ethernet is configured first (even prior to LED pins)
  * Ethernet is only configured once
  * [Additional (non-configurable) ethernet pins](https://github.com/Aircoookie/WLED/pull/2066#issuecomment-886014394) are now allocated using PinManager
* Pin Manager improvements
  * Allocation / deallocation macros take an "ownership" tag parameter
    * Helps avoid accidentally free'ing pins that were allocated by other code
    * _Costs ~20 bytes RAM on ESP8266, ~40 bytes RAM on ESP32_
  * new functions to allocate multiple pins at once; Simplifies error handling
    * Example hardware that this would apply to includes rotary encoders, I2C, SPI, and of course ethernet.
  * additional debug output (debug builds only)
* Other minor fixes
  * Updates to platformio.ini to enable more platforms to build under TravisCI
  * Updates to some (v2) user modules to use new Pin Manager allocation routines
  * Some operator precedence fixes (showed as build warnings, verified as true errors)
  * Fixed some typos

<details><summary>Builds cleanly under TravisCI (click to expand platforms built)</summary></P>

See [successful build](https://github.com/henrygab/WLED/actions/runs/1107024943) from penultimate commit 203a75c.

* travis_esp8266
* travis_esp32
* anavi_miracle_controller
* codm-controller-0.6
* codm-controller-0.6-rev2
* d1_mini
* d1_mini_5CH_Shojo_PCB
* d1_mini_debug
* d1_mini_ota
* elekstube_ips
* esp01_1m_full
* esp07
* esp32_eth
* esp32dev
* esp32s2_saola
* esp8285_4CH_H801
* esp8285_4CH_MagicHome
* esp8285_5CH_H801
* h803wf
* heltec_wifi_kit_8
* m5atom
* nodemcuv2
* sp501e
* wemos_shield_esp32

</details>

Basic ethernet functionality was tested only on WT32-ETH01 boards.


platformio.ini
```diff
@@ -399,7 +402,7 @@ build_flags = ${common.build_flags_esp32}
   -D TEMPERATURE_PIN=23
 lib_deps = ${esp32.lib_deps}
   OneWire@~2.3.5
-  U8g2@~2.28.11
+  olikraus/U8g2 @ ^2.28.8
```
henrygab, line 405: v2.28.11 was not available via platformio.
Adding the fully-qualified library and requiring only v2.28.8 works.

platformio.ini
```diff
@@ -482,4 +485,6 @@ build_flags = ${common.build_flags_esp32} -D WLED_DISABLE_BROWNOUT_DET -D WLED_D
   -D SPI_FREQUENCY=40000000
   -D USER_SETUP_LOADED
 monitor_filters = esp32_exception_decoder
-lib_deps = ${esp32.lib_deps} TFT_eSPI
+lib_deps =
```
henrygab, line 488: Similarly, using explicit library version (across two lines) allowed this to build.

usermods/Animated_Staircase/Animated_Staircase.h
```diff
@@ -306,22 +306,26 @@ class Animated_Staircase : public Usermod {
 
   public:
     void setup() {
+      // standardize invalid pin numbers to -1
```
henrygab, line 309: PinManager APIs treat `-1` as a special case

usermods/Animated_Staircase/Animated_Staircase.h
```diff
@@ -480,10 +484,10 @@ class Animated_Staircase : public Usermod {
             (oldBottomAPin != bottomPIRorTriggerPin) ||
             (oldBottomBPin != bottomEchoPin)) {
           changed = true;
-          pinManager.deallocatePin(oldTopAPin);
-          pinManager.deallocatePin(oldTopBPin);
-          pinManager.deallocatePin(oldBottomAPin);
-          pinManager.deallocatePin(oldBottomBPin);
+          DEALLOCATE_PIN(oldTopAPin, PinOwner::UM_AnimatedStaircase);
```
henrygab, line 487: PinManager APIs treat attempts to deallocate pin `-1` as a no-op.  Simplifies client cleanup.

usermods/EleksTube_IPS/ChipSelect.h
```diff
@@ -58,12 +58,12 @@ class ChipSelect {
   void setMinutesTens()                           { setDigit(MINUTES_TENS); }
   void setHoursOnes()                             { setDigit(HOURS_ONES); }
   void setHoursTens()                             { setDigit(HOURS_TENS); }
-  bool isSecondsOnes()                            { return (digits_map&SECONDS_ONES_MAP > 0); }
-  bool isSecondsTens()                            { return (digits_map&SECONDS_TENS_MAP > 0); }
-  bool isMinutesOnes()                            { return (digits_map&MINUTES_ONES_MAP > 0); }
-  bool isMinutesTens()                            { return (digits_map&MINUTES_TENS_MAP > 0); }
-  bool isHoursOnes()                              { return (digits_map&HOURS_ONES_MAP > 0); }
-  bool isHoursTens()                              { return (digits_map&HOURS_TENS_MAP > 0); }
+  bool isSecondsOnes()                            { return ((digits_map & SECONDS_ONES_MAP) > 0); }
```
henrygab, line 61: These had operator precedence errors that were caught by compiler.  Easy fix.

wled00/cfg.cpp
```diff
@@ -18,6 +18,13 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
 
   //long vid = doc[F("vid")]; // 2010020
 
+  #ifdef WLED_USE_ETHERNET
```
henrygab, line 21: Per discussions with AirCookie and Blazoncek, initialization of ethernet should occur prior to other configuration (e.g., pins should be allocated to ethernet before being allocated to LEDs).

wled00/pin_manager.h
```diff
@@ -4,26 +4,103 @@
  * Registers pins so there is no attempt for two interfaces to use the same pin
  */
 #include <Arduino.h>
+#include "const.h" // for USERMOD_* values
+
+typedef struct PinManagerPinType {
```
henrygab, line 9: An array of this struct is used to allocate multiple pins at once.

wled00/pin_manager.h
```diff
@@ -4,26 +4,103 @@
...
+/*
+ * Allows PinManager to "lock" an allocation to a specific
+ * owner, so someone else doesn't accidentally de-allocate
+ * a pin it hasn't allocated.  Also enhances debugging.
+ * 
+ * RAM Cost:
+ *     17 bytes on ESP8266
+ *     40 bytes on ESP32
+ */
+enum struct PinOwner : uint8_t {
```
henrygab, line 23: Reduced (vs. prior PR) RAM overhead by changing to single byte.

Built-in owners have high bit set to one (0x80).
User modules use their assigned user module ID.

wled00/pin_manager.h
```diff
@@ -4,26 +4,103 @@
...
   public:
-  void deallocatePin(byte gpio);
-  bool allocatePin(byte gpio, bool output = true);
-  bool isPinAllocated(byte gpio);
+  bool _deallocatePin(byte gpio, PinOwner tag);
+  bool _allocatePin(byte gpio, bool output, PinOwner tag);
+  bool _allocateMultiplePins(const managed_pin_type * mptArray, byte arrayElementCount, PinOwner tag );
+
+  #if !defined(ESP8266) // ESP8266 compiler doesn't understand deprecated attribute
+  [[deprecated("Replaced by macro ALLOCATE_PIN(gpio, output), for improved debugging")]]
```
henrygab, line 88: By marking the old API deprecated, it provides a backwards-compatible API surface, while making clear where older code should be updated.

henrygab (commented):
These are just some comments to guide the review.  Feel free to ignore as you prefer.

wled00/wled.cpp
```diff
@@ -105,6 +105,10 @@ void WiFiEvent(WiFiEvent_t event)
       break;
     case SYSTEM_EVENT_ETH_DISCONNECTED:
       DEBUG_PRINT(F("ETH Disconnected"));
+      // BUGBUG -- Is it appropriate to mess with WiFi when ethernet was disconnected?
+      //           Even if necessary in the past, ethernet is now only configured once,
+      //           during initial setup.  Until it's initialized, this event shouldn't
+      //           occur. If it is, why need to do anything here?
```
blazoncek, line 111: It may be necessary to reconnect WiFi if wired connection is somehow lost.
Reconnecting WiFi (if necessary) is the only way to get access to ESP once wired connection is lost apart from physical access.

blazoncek:
🙇 
Nicely done!

wled00/wled.cpp
henrygab: Ah... ok, that makes sense.  I'll make one more commit to remove this comment.  Thanks!

henrygab:
Great!  Will this go into your fork, or should I leave this PR open?

blazoncek:
I have found out that mixing forks in PR is a bad thing. Messes up GH.
Let's leave it here and let @Aircoookie decide. I think it is a great addition nevertheless.
I only have one ethernet shield (and still only for testing) so have no immediate need for the PR myself.

Aircoookie:
Please leave this open, these are some great improvements :)

Will merge after a bit of testing with ethernet boards.

Quick question: What is the benefit of using macros for the allocate/deallocate functions?
Having PinOwner be an extra, optional argument with the default value of `PinOwner::None` (as you have done with `isPinAllocated()`) would be more intuitive I believe, and would be an extension to the current PinManager API rather than a complete replacement from a user(mod) perspective.

henrygab:
> Quick question: What is the benefit of using macros for the allocate/deallocate functions?

Good question.  Short answer is that I should remove them now.  
<details><summary>Click for long answer.</summary><P/>

It's a remnant of earlier PRs and checks.  Initially, I did not have the owner tag.  Instead, I had the macro expand to include `__FILE__` and `__LINE__` in debug builds.  This helped me track down why ethernet was having difficulty on a fresh WT32-ETH01 board (conflict with pin 16, which is default LEDPIN, and allocated first).  I was still considering leaving the `__FILE__` and `__LINE__` bits in the debug version of the macro expansion until very recently.  However, with the addition of the owner tag and some extra debug prints, the complexity of finding out what is conflicting became significantly easier, so I removed the debug version of the expansion recently... I just didn't fully realize I could swap back until you mentioned it.  🤦‍♂️🤷‍♂️
</summary>

> Having PinOwner be an extra, optional argument with the default value of `PinOwner::None` (as you have done with `isPinAllocated()`) would be more intuitive I believe, and would be an extension to the current PinManager API rather than a complete replacement from a user(mod) perspective.

I agree that having default value(s) can makes some APIs easier to use.  Here, providing a default value of `PinOwner::None` would be counter-productive, as it would encourage use that defeats the benefits and purpose of providing a pin owner.  I would have preferred not to allow that tag to be passed at all, except for the API compatibility concerns.

If someone really wants (or needs) to provide a tag of `PinOwner::None`, they can still do so, although I feel it should be a conscious decision (not due to skipping an optional parameter).  They can also still use the legacy APIs ... code using those still compiles, although there will be a deprecation warning.

All that said, let me go ahead and get rid of the macros.... I'll move the PR to draft status temporarily, while I do that....

henrygab:
Huh.  I couldn't find the spot to change this PR back to draft.

However, I made all the changes, [built everything](https://github.com/henrygab/WLED/runs/3287812777), and commit fa4ec23 is a the squashed result in this branch/PR.

Thanks for pushing me to cleanup those macro remnants; I agree it makes the code much cleaner.  👍

blazoncek:
Discussion about PinOwner brought this thoughts to my mind:
Would it be possible to extend current enum without changing pinmanager.h? In such a way to allow usermod creator to simply extend current enum by adding a custom one in usermod files.

henrygab:
> Would it be possible to extend current enum without changing pinmanager.h? In such a way to allow usermod creator to simply extend current enum by adding a custom one in usermod files.

Generally, no.

Scoped enumerations provide compile-time type safety, which also means that no implicit conversions to/from integer types are provided.  This helps(`*`) ensure only valid values are used as parameters, as values in structs, etc.  Of course, the performance is identical to using an unscoped enum, or using integer types, because they are each, after compilation, just an integer value. 

So, generally, the answer is "no", they must insert the new value into the scoped enum.  This is a good thing, as it makes it easier to avoid re-use of an existing value, and all values are in one place when debugging.  For a UserMod that hasn't added itself to the the list already, it can simply use `PinOwner::UM_Unspecified`....

<details><summary>* But also yes (click to expand)</summary></P>

The answer is also "yes", because C++ still allows evil constructs that violate type safety.  Here's a few examples. which I think explain the situation better than words:

```C++
// For an unspecified user mod, the expected way:
pinManager.allocatePin( pin, PinOwner::UM_Unspecified );
// C-style cast is still allowed in C++
pinManager.allocatePin( pin, ((enum PinOwner)(0x55)) );
// static_cast is evil, but it can be used here....
pinManager.allocatePin( pin, static_cast<enum PinOwner>(0x55) );
// another evil method using static_cast<>
#define TERRIBLE_WORKAROUND(tag) static_cast<enum PinOwner>(pin)
pinManager.allocatePin( pin, TERRIBLE_WORKAROUND(0x55) );
// while prototyping, while making it a one-line change to switch
// to an allocated PinOwner tag...
#define UM_XYZZY_TAG static_cast<enum PinOwner>(0x56)
pinManager.allocatePin( pin, UM_XYZZY_TAG );
```

While those might be fine options for prototyping, I generally would not recommend allowing use of `static_cast<>` in the main branch, as it removes type-safety.

</details><HR/>

In summary, I think the PR elegantly meets all the conflicting desires:
1. Type safety is provided at compile-time for API usage
2. The backwards-compatible API still exists (using compiler deprecation warning for ESP32 builds)
3. User mods can use `PinOwner::UM_Unspecified`, and avoid updating the const / enums _(although I would recommend this **not** be allowed for usermods in the usermods_list.cpp file)_
4. For prototyping, still have ability to bypass the type safety and use **_any_** value desired

blazoncek:
I was not concerned by type safety (after all it is integer as you say and I also like enums) but rather with a usermod creator changing core files.
I would rather see an API for PinOwner _class_ to allow extensions (i.e. adding custom values or just incrementing local counter of pin owners).

This could be as simple as:
```
class PinOwner {static int addOwner(...){...;return used++;}; private: static int used=32;...}
...
// in usermod
int ownerID = PinOwner::addOwner(...some UM info...);
...
pinManager.allocatePin(..., ownerID);
```

And you can still have _reserved_ values assigned manually or using #defines in core files.

henrygab:
> I was not concerned by type safety

True, that was one of my own goals... helping avoid coding errors via type safety.  Sorry if I made it seem this was a goal of yours.

> I would rather see an API for PinOwner _class_ to allow extensions ...

<details><summary>An interesting concept, with its own plusses and minuses...</summary><P/>

I've used both types of systems in the past.  (_I was a kernel-mode developer for ~10 years._)  The dynamic allocation has distinct benefits, especially when you don't know the potential values at compilation time, and need to validate the provided values existed at compilation.

Here, there's no validation that the values provided were in the list that existed at compilation.  This allows the question of which values are "allocated" to be managed separately from the rest of the code.  (adding a value doesn't actually require adding the value into the `pin_manager.h` file ... so long as there's a defined location for allocating the values used.)

I also has a slight preference for the hard-coded values, to simplify debugging a bit.  Pin owner "0x85" is always going to be a button, vs. needing old debug output (or memory overhead for dynamic owner information tracking) to figure out who owner "3" is ... especially after configuration is loaded / unloaded a few times.

</details>

While I still think this is true, to avoid doubt:

@blazoncek, are you still approving this PR w/o additional changes?

Thanks!

blazoncek:
Yes. 😄

Aircoookie:
Thank you for replacing the macros 🎉 

I also believe usermod creators editing core-files is fine (as in, adding themselves to an existing list). This is already the case for the usermod ID in `const.h` and if they desire to have an unique `PinOwner::UM_XX` identifier, that is fine, too.

Approving of this PR in its current form too :)

usermods/usermod_v2_four_line_display/usermod_v2_four_line_display.h
```diff
@@ -167,17 +167,15 @@ class FourLineDisplayUsermod : public Usermod {
...
-        return;
-      }
-      PinManagerPinType pins[2] = { { sclPin, true }, { sdaPin, true} };
-      if (!pinManager.allocateMultiplePins(pins, 2, PinOwner::UM_FourLineDisplay)) {
-        sclPin = -1;
-        sdaPin = -1;
-        type = NONE;
-        return;
+      if (type == NONE) return;
+      bool allocated = false;
```
blazoncek, line 171: Can be removed.

Aircoookie:
While testing with an ethernet board today (first of all, pin manager works great) I've noticed that Ethernet does not initialize until after a reboot (even if the type was `None`) before. The reason is that `initEthernet()` is only called in `deserializeJson()`, which is (currently) not called after saving settings. I've added an extra `WLED::instance().initEthernet();` in set.cpp, this fixed that problem and Ethernet is initialized immediately :)

henrygab:
I've done iterators before.  They are terrible ... especially if you want to implement with in C++11 *and* still be able to initialize as const data.  Iterators are perhaps excessive here.

Question: where is the definition of number of pins for ESP8266 vs. ESP32?

blazoncek:
> Question: where is the definition of number of pins for ESP8266 vs. ESP32?
> 
> P.S. - The point of iterators is to allow your code to be simpler:
> 
> ```c++
> for (auto i : pinManager)
> ```

Told you I suck at C++. 😃 

[RandomNerdTutorials](https://randomnerdtutorials.com/) have some good references for both ESP8266 and ESP32.

Aircoookie:
> Question: where is the definition of number of pins for ESP8266 vs. ESP32?

Hardcoded in `isPinOk()`. Perhaps two defines in the top or `const.h` would be more maintainable.

henrygab:
Just wanted to let you know that adding the pins to XML and JSON serialization is still on my radar.  Real life just has to take precedence.  👪 😃

blazoncek:
> Just wanted to let you know that adding the pins to XML and JSON serialization is still on my radar. Real life just has to take precedence. 👪 😃

I've implemented quick & dirty way of handling those. Ethernet arrays need to be moved out of header files (which is a bad idea anyway).

henrygab:
See PR https://github.com/Aircoookie/WLED/pull/2167 for serialization of all allocated pins.

PR #2125, opened 2021-08-09
Added usermod "battery status basic"

itCarl:
quick and dirty usermod to show the battery level in the info modal.


itCarl:
https://lazyzero.de/elektronik/esp8266/wemos_d1_mini_a0/start

PR #2134, opened 2021-08-12
Add new palettes from SR branch

scottrbailey:
Palettes added in Sound Reactive fork. 

I'm not sure why the button for Candy2 did not get the linear gradient background. This was true in SR too. I'm guessing it probably has something to do with it being the last palette. But if you see the reason for this, please let me know.


blazoncek:
Will check this one soon.

Issue #2136, opened 2021-08-14
Sync Groups


smartroad:
When clicking the Sync button it syncs all the current WLED installs to the current unit. Is there a way to set groups for syncing? I have several WLED lights in various rooms and would like to have them as seperate groups to sync to.

I was thinking in the settings there could be an option for setting a sync group number just 0-9 (although maybe 0-255 as it would have to be a byte anyway?).

Dunno if that sounds like a good idea or not, or if I have somehow missed an existing feature!

itCarl:
You could simply change the upd port of the corresponding wleds you want to have in a group.
Settings > Sync Interfaces/Setup > UDP Port

smartroad:
> You could simply change the upd port of the corresponding wleds you want to have in a group.
> Settings > Sync Interfaces/Setup > UDP Port

Would that affect any other functionality of the system? Using external programs (i.e. LEDFX) or identification of other instances from the web interface. 

Actually just made me think, given WLED can find other instances on the network, when clicking on the sync maybe it could ask which other instances to attach to. Would be a more user friendly focused implementation compared to changing group or port numbers.

Aircoookie:
Changing the UDP port does not impact other functionality :)
The secondary UDP port also runs the instance discovery as well as being the default port for the tpm2.net protocol.
The primary port can be changed freely. I agree that a smaller group number would be more intuitive, that might be a good idea.

Directly creating sync groups on the device would be awesome. A challenge would be that the device would need to remember the instances to sync to even if they become unreachable for some time and this would likely cost a not insignificant amount of memory. Nevertheless, adding a concept of "Rooms" would definitely make using multiple WLED devices together a lot more accessible.

smartroad:
Given the current sync method, where a "master" node initiates the sync, could the client nodes just remeber the Masters ID and then ping it when back online? The Master could reply with:

"Sure still grouped here are your settings"
or
"Not in my group anymore do your own thing"

I am a hobbyist coder so I am probably way over simplifying it lol but would a simple protocol like that work?

pbolduc:
[...]

PR #2150, opened 2021-08-21
Sync groups

blazoncek:
This could help solve #2136.
Not yet tested.


blazoncek:
The real question is: do we enable all groups by default? For sending? For receiving? Or both?
Personally I have found that it may be better to have only  one group (i.e. 1) enabled by default for both send and receive.

blazoncek:
Tested and working properly.
Older devices follow the master regardless of sending group (backwards compatible?).

Aircoookie:
Awesome!
Will give it a try soon too.
Also believe that in a typical use case, having send and receive only enabled for the first group would make the most sense. Otherwise, once you want a new dedicated group, you'd need to reconfigure all existing devices to no longer receive/send on that group.

Aircoookie:
Made a slight change so that old UDP packets are treated as group 1 only - that way it is easier to set up a new group without updating all existing instances.

Functionality is awesome!
One thing I noticed is that adding the options in xml.cpp causes a high dynamic string length (+260 bytes) due to `d.Sf.G1.checked=1;d.Sf.R1.checked=1;` and so on. This hasn't caused problems in my testing, still we might want to consider doing the bitshifting in JS, reducing the number of transmitted options from 16 to 2. We already do this with `FC()` and `Wd()` in time settings for the weekdays. Will implement this now :)

Aircoookie:
Done. Also added a check if no sync group is enabled, in that case we don't need to send UDP at all, since it will be rejected by all recipients anyways (except for <0.13)

PR #2157, opened 2021-08-26
Add network debug printer

visigoth:
Add a network debug print facility that redirects `DEBUG_PRINT*` to the network. It blindly sends UDP packets with the intended string to a specified host and port.

It could be smarter by splitting the string up across the expected, measured, or known MTU of the network, for example, but that's a bit too much work compared to the likeliness of it being needed.

On the host side, I used netcat to receive the log statements: `nc -p 7868 -u -l -s <network ip>`

The security settings page has been updated to specify the network debug host and port, as well as enable/disable such logging. However, I found it would not persist correctly, so my usage has been to just edit the settings inline in `wled.h`.


blazoncek:
I will pull this into my fork and give it a spin.
Have you tested it on ESP8266 maxed out on LEDs and segments?

visigoth:
> Have you tested it on ESP8266 maxed out on LEDs and segments?

no; i only have an ESP32 here.

blazoncek:
I like the idea, but currently don't have time for rigorous testing.
Please get ESP8266 and configure it with 1200 LEDs (no need to attach them) and create 16 segments, then observe free heap and compare it to the one without network debug printer.

blazoncek:
Hi, I finally manged to pull into my development fork and play with it.
I do have some suggestions:

- make host IP resolved upon instance creation and not at each print
- allocating 1024 bytes from stack in printf() is dangerous, use malloc() with safety checks
- I am not fond of the idea of storing hostname & port to config since debug build is not something ordinary people need; and debug destination does not really change that often to require
- I would prefer compile time option for hostname and port
- I would make NetDebug replace all DEBUGxx_PRINT functions for completeness sake
- I would make WLED_DEBUG_NET also define/replace WLED_DEBUG

But that would be it.

wled00/net_debug.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+void NetworkDebugPrinter::println(unsigned int n) {
+    print(n);
+    print("\n");
+}
+
+void NetworkDebugPrinter::printf(const char *fmt...) {
+    va_list args;
+    va_start(args, fmt);
+    char s[1024];
```
pbolduc, line 69: searching for `udp max packet size mtu` it seems you could increase the max message size to near 1454 without fragmenting the packets.

wled00/net_debug.cpp
```diff
@@ -0,0 +1,75 @@
+#include "wled.h"
+
+#ifdef WLED_DEBUG_NET
+
+NetworkDebugPrinter NetDebug;
+
+void NetworkDebugPrinter::print(const char *s) {
```
pbolduc, line 7: be a little more defensive, for pointer parameters ensure they are not nullptr otherwise it will crash

pbolduc (changes requested):

wled00/net_debug.cpp
```diff
@@ -0,0 +1,75 @@
...
+    print(s);
+}
+
+void NetworkDebugPrinter::println() {
+    print("\n");
+}
+
+void NetworkDebugPrinter::println(const char *s) {
+    print(s);
+    print("\n");
```
pbolduc, line 48: for each print statement, you are pushing a UPD packet, perhaps refactor so internal extra new lines are automatically  appended to the output buffer.   You can use `debugUdp.write('\n');` or `debugUdp.write(0x0a); // 0x0a is newline`

pbolduc (changes requested):

blazoncek:
I removed settings page configuration and replaced it with compile time option.
Also replaced destination of regular DEBUG_PRINT macros in case host is defined. It makes using network debug much easier to use that way.

7fcc8be73c727a2475eb3780efdc3681d3869e43

visigoth:
cool; thank you @blazoncek for commandeering and making it mergable!

PR #2170, opened 2021-09-03
Fix error 12 issues

scottrbailey:
Store error flag from readObjectFromFile in new variable to prevent error 12 when IR code is not found or ir.json does not exist.
Fix presetFallback so error is not thrown if preset does not exist.

### DIFF
--- a/wled00/ir.cpp
+++ b/wled00/ir.cpp
@@ -568,60 +568,62 @@ void decodeIRJson(uint32_t code)
   char objKey[10];
   const char* cmd;
   String cmdStr;
-  byte irError;
   DynamicJsonDocument irDoc(JSON_BUFFER_SIZE);
   JsonObject fdo;
   JsonObject jsonCmdObj;
 
   sprintf(objKey, "\"0x%X\":", code);
 
-  irError = readObjectFromFile("/ir.json", objKey, &irDoc) ? ERR_NONE : ERR_FS_PLOAD;
+  readObjectFromFile("/ir.json", objKey, &irDoc);
   fdo = irDoc.as<JsonObject>();
   lastValidCode = 0;
-  if (!irError) 
+  if (fdo.isNull()) {
+    //the received code does not exist
+    if (!WLED_FS.exists("/ir.json")) errorFlag = ERR_FS_IRLOAD; //warn if IR file itself doesn't exist
+    return;
+  }
+
+  jsonCmdObj = fdo["cmd"];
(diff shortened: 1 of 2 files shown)


wled00/ir.cpp
```diff
@@ -566,16 +568,17 @@ void decodeIRJson(uint32_t code)
...
   String cmdStr;
+  byte irError;
   DynamicJsonDocument irDoc(JSON_BUFFER_SIZE);
   JsonObject fdo;
   JsonObject jsonCmdObj;
 
   sprintf(objKey, "\"0x%X\":", code);
 
-  errorFlag = readObjectFromFile("/ir.json", objKey, &irDoc) ? ERR_NONE : ERR_FS_PLOAD;
+  irError = readObjectFromFile("/ir.json", objKey, &irDoc) ? ERR_NONE : ERR_FS_PLOAD;
```
blazoncek, line 578: I would keep errorFlag but add different error code.
I agree that receiving unknown IR codes will display error in such case so it may require another approach to differentiate between nonexistent ir.json and unknown IR code.

Aircoookie: I agree with you in that a missing or mal-formatted `ir.json` should throw an error while an unknown IR code should not. `WLED_FS.exists("/ir.json")` would be an option. If we do not want to add any extra logic though, I would err towards not setting an error as all (`irError` is a local var, thus I would not regard it as an "error" communicated to the user)

blazoncek: > I agree with you in that a missing or mal-formatted `ir.json` should throw an error while an unknown IR code should not. `WLED_FS.exists("/ir.json")` would be an option. If we do not want to add any extra logic though, I would err towards not setting an error as all (`irError` is a local var, thus I would not regard it as an "error" communicated to the user)

Ha! The best solutions are the simplest ones. Unfortunately I never looked up FS API to know it better so I implemented:

```c++
  JsonObject fdo = irDoc.createNestedObject("cmd");
  // attempt to read command from ir.json
  // this may fail for two reasons: ir.json does not exist or IR code not found
  // if the IR code is not found readObjectFromFile() will clean() irDoc JSON document
  // so we can differentiate between the two
  errorFlag = readObjectFromFile("/ir.json", objKey, &irDoc) ? ERR_NONE : ERR_FS_IRLOAD;
  if (irDoc["cmd"].isNull()) {
    errorFlag = ERR_NONE;
    return;
  }
```

PR #2175, opened 2021-09-03
Revert changes from PR1902

henrygab:
@ahmed-shehata approved in the discourse discussions.

Plan is to temporary remove this from main code.

Two follow-up PRs will re-enable this as a user module:
1. A PR to provide UMs the ability to read / in-place-edit pixel data (e.g., color correction curves, etc.)
2. Ahmed will then provide a PR with the user module that includes these capabilities


blazoncek:
This may be a good opportunity to introduce _hooks_ for usermods to interact with core WLED.
Such as:

- get button state
- get/set other usermod values (including HW objects such as I2C driver, etc)
- get IR codes
- get MQTT, UDP or other messages without explicitly subscribing 
- etc

henrygab:
Let's get this PR pulled in first?

henrygab:
> I was thinking in general terms as this may be an excellent example. 😄

In that case, yes, absolutely!

I think both you @Aircoookie have reviewed this, and the original author also agreed this sounded better as a user module.  Is there anything remaining I need to do, which is keeping this from being merged?

blazoncek:
I have no objections but since @Aircoookie merged previous PR in I'll leave it to him to handle this one as well.

Aircoookie:
No objections on my side either.

I was going to leave this in until shortly before it is available as usermod, but anyone who needs it can reset to a commit that includes it for the time being :)

PR #2184, opened 2021-09-10
FX optimisations and segment names

Aircoookie:


wled00/FX.cpp
```diff
@@ -2946,18 +2895,22 @@ uint16_t WS2812FX::mode_starburst(void) {
...
+#undef STARBURST_MAX_FRAG
+#undef STARBURST_MAX_STARS
 
 /*
  * Exploding fireworks effect
  * adapted from: http://www.anirama.com/1000leds/1d-fireworks/
  */
-
+#ifdef ESP8266
+  #define MAX_SPARKS 20 // number of fragments
```
Aircoookie, line 2906: I think this is a bit few and will makes the effect look far less brilliant in the (likely) case few segments are used.
I will think of a "smart" logic here to enable both running on many segments, and maintaining `min(seglen/2, 80)` sparks for few segments.

wled00/FX.h
```diff
@@ -55,17 +53,17 @@
 /* each segment uses 52 bytes of SRAM memory, so if you're application fails because of
   insufficient memory, decreasing MAX_NUM_SEGMENTS may help */
 #ifdef ESP8266
-  #define MAX_NUM_SEGMENTS    12
+  #define MAX_NUM_SEGMENTS    18
```
Aircoookie, line 56: Definitely need a lot of testing here. I'm very doubtful that 18 segments + names + full `/json` request will fit in a 10k JSON buffer

wled00/FX.h
```diff
@@ -55,17 +53,17 @@
...
   /* How many color transitions can run at once */
   #define MAX_NUM_TRANSITIONS  8
   /* How much data bytes all segments combined may allocate */
-  #define MAX_SEGMENT_DATA  2048
+  #define MAX_SEGMENT_DATA  4096
```
Aircoookie, line 60: Pretty sure this will run out of memory with 1600 LEDs defined. Need to test though, but I'm not sure I'm comfortable with allocating this much. Falling back to solid is better than becoming unresponsive due to out of memory.

wled00/FX.h
```diff
@@ -345,25 +344,31 @@ class WS2812FX {
...
+      uint32_t call;  // call counter
+      uint16_t aux0;  // custom var
+      uint16_t aux1;  // custom var
       byte* data = nullptr;
       bool allocateData(uint16_t len){
         if (data && _dataLen == len) return true; //already allocated
         deallocateData();
         if (WS2812FX::instance->_usedSegmentData + len > MAX_SEGMENT_DATA) return false; //not enough memory
-        data = new (std::nothrow) byte[len];
+        // if possible use SPI RAM on ESP32
```
Aircoookie, line 357: PSRAM is slower than internal RAM theoretically. I don't know whether this makes a difference in practice, even if accessed a lot within the effect function.
Still, not bad to have considering it can be enabled separately with `WLED_USE_PSRAM`.

wled00/FX.cpp
```diff
@@ -1318,29 +1262,28 @@ uint16_t WS2812FX::mode_two_dots()
   fill(SEGCOLOR(2));
```
Aircoookie, line 1262: Perfect :)

wled00/FX_fcn.cpp
```diff
@@ -235,12 +233,12 @@ void WS2812FX::setPixelColor(uint16_t i, byte r, byte g, byte b, byte w)
...
     uint32_t col = ((w << 24) | (r << 16) | (g << 8) | (b));
 
-    bool reversed = IS_REVERSE;
+    /* Set all the pixels in the group */
     uint16_t realIndex = realPixelIndex(i);
     uint16_t len = SEGMENT.length();
 
     for (uint16_t j = 0; j < SEGMENT.grouping; j++) {
-      int indexSet = realIndex + (reversed ? -j : j);
+      uint16_t indexSet = realIndex + (IS_REVERSE ? -j : j);
```
Aircoookie, line 241: Let's keep this `int`. Even though it makes no immediate logical sense, this safeguards against an effect function causing a crash on ESP32 by writing outside of segment bounds (https://github.com/Aircoookie/WLED/commit/789c00dde12634a9db415e012d06550f3d0370e5)

wled00/FX_fcn.cpp
```diff
@@ -563,14 +561,14 @@ uint32_t WS2812FX::getLastShow(void) {
...
 
-//TODO these need to be on a per-strip basis
-uint8_t WS2812FX::getColorOrder(void) {
-  return COL_ORDER_GRB;
-}
-
-void WS2812FX::setColorOrder(uint8_t co) {
-  //bus->SetColorOrder(co);
-}
+// there is no longer any need for these two
```
Aircoookie, line 564: can be fully removed.

wled00/FX_fcn.cpp
```diff
@@ -1012,18 +1030,33 @@ uint32_t WS2812FX::color_from_palette(uint16_t i, bool mapping, bool wrap, uint8
 
 
 //load custom mapping table from JSON file
-void WS2812FX::deserializeMap(void) {
-  if (!WLED_FS.exists("/ledmap.json")) return;
-  DynamicJsonDocument doc(JSON_BUFFER_SIZE);  // full sized buffer for larger maps
+void WS2812FX::deserializeMap(uint8_t n) {
+  String fileName = String(F("/ledmap"));
```
Aircoookie, line 1034: I will change this to use `sprintf_P`, which should be a little more heap friendly than String.

wled00/json.cpp
```diff
@@ -18,6 +18,29 @@ void deserializeSegment(JsonObject elem, byte it, byte presetId)
   uint16_t start = elem[F("start")] | seg.start;
   int stop = elem["stop"] | -1;
 
+    if (elem["n"]) {
+      // name field exists
+      String name = elem["n"];
+      if (name.length()) {
+        if (seg.name) delete seg.name;
+        seg.name = new char[name.length()+1];
```
Aircoookie, line 26: This is not very safe. A long string could be used to run into low memory situations deliberately, or `strcpy` could write to memory if the allocation failed.
We should
- limit length with strlcpy to 32 or at most 64 characters
- check that `seg.name` is not nullptr before copying

wled00/json.cpp
```diff
@@ -18,6 +18,29 @@ void deserializeSegment(JsonObject elem, byte it, byte presetId)
   uint16_t start = elem[F("start")] | seg.start;
   int stop = elem["stop"] | -1;
 
+    if (elem["n"]) {
+      // name field exists
+      String name = elem["n"];
```
Aircoookie, line 23: Same as above, we can easily use a char array here

wled00/json.cpp
```diff
@@ -339,6 +362,8 @@ void serializeSegment(JsonObject& root, WS2812FX::Segment& seg, byte id, bool fo
   byte segbri = seg.opacity;
   root["bri"] = (segbri) ? segbri : 255;
 
+  if (seg.name != nullptr) root["n"] = String(seg.name);
```
Aircoookie, line 365: I don't think `String()` cast is necessary.

wled00/data/index.js
```diff
@@ -579,10 +579,11 @@ function populateSegments(s)
 				<span class="checkmark schk"></span>
 			</label>
 			<div class="segname" onclick="selSegEx(${i})">
-				Segment ${i}
+				${inst.n ? inst.n : "Segment "+i}
 			</div>
 			<i class="icons e-icon flr ${expanded[i] ? "exp":""}" id="sege${i}" onclick="expand(${i})">&#xe395;</i>
 			<div class="segin ${expanded[i] ? "expanded":""}" id="seg${i}">
+				<input type="text" class="ptxt noslide" id="seg${i}t" autocomplete="off" maxlength=32 value="${inst.n?inst.n:""}" placeholder="Enter name..."/>
```
Aircoookie, line 586: This is missing in the "Add segment" div built by `makeSeg()`. This causes a JS exception in the first line of `setSeg()` upon trying to save the new segment.

wled00/data/index.js
```diff
@@ -579,10 +579,11 @@ function populateSegments(s)
 				<span class="checkmark schk"></span>
 			</label>
 			<div class="segname" onclick="selSegEx(${i})">
-				Segment ${i}
+				${inst.n ? inst.n : "Segment "+i}
 			</div>
 			<i class="icons e-icon flr ${expanded[i] ? "exp":""}" id="sege${i}" onclick="expand(${i})">&#xe395;</i>
 			<div class="segin ${expanded[i] ? "expanded":""}" id="seg${i}">
```
Aircoookie, line 585: There is a too big gap between segment top line and name field. Will decrease margin in css.

Aircoookie (commented):
This is so awesome! :)

I've made a review to remind myself of a few points I'd like to change. Will make a few commits tonight.
Especially not yet sure on the dynamic FX data buffer and segment limit increases. Could cause a lot of trouble on ESP8266. Maybe we can release v0.13.0-b3 with these new limits and dial them back for v0.13.0 only if the community encounters issues.

Thank you!

wled00/FX_fcn.cpp
Aircoookie: Just realized this was `int16_t` before, so `uint16_t` might be fine.

wled00/data/index.js
blazoncek: Yes. I do have slightly modified CSS in my fork. Sorry.

wled00/FX.h
blazoncek: 16 definitely does. And theoretically 18 does as well.

wled00/FX.h
blazoncek: 😄 
Are you sure you want to run 1600 LEDs on 8266?
I tested it with 1200 LEDs defined and 13 segments (with a few usermods). Beyond 13th segment ESP would run out of memory. But if you keep LED count to a reasonable 600 then it is still fine.

wled00/FX.h
blazoncek: Most people will not use it but I have tested it and it works.

wled00/FX_fcn.cpp
blazoncek: There should be other safeguards to prevent crossing segment bounds.

wled00/json.cpp
blazoncek: Thanks for pointing it out. Perhaps delete[] is also in order...

wled00/data/index.js
blazoncek: Uh! I tried to port my rewrite of segment handling (which is a bit different). Perhaps I missed something.

blazoncek:
> Especially not yet sure on the dynamic FX data buffer and segment limit increases. Could cause a lot of trouble on ESP8266. Maybe we can release v0.13.0-b3 with these new limits and dial them back for v0.13.0 only if the community encounters issues.

I did quite extensive calculations regarding memory use on 8266 and came up with the numbers in this PR.
The code will leave approximately 18-20k free heap (as seen in Info page) using 18 segment max and all effects will be running on all segments (which is not the case currently) using a reasonable amount of 600 LEDs.

Of course if a user selects _unreasonable_ combination of all parameters then yes, the ESP will occasionally (often?) crash.
I've done my testing primarily using 600 LEDs in total and using 30-50 pixels per segment.

wled00/FX.cpp
blazoncek: Many segments and effects with great memory requirements don't go well together. 😄

wled00/FX.h
Aircoookie: I never would 😄 way too little FPS and yeah, you run out of memory really quickly.
But yeah, I have "allowed" up to 5k for the LED buffer, which is around 1600.
Maybe making the effect data memory limit dynamic would be a good compromise here. 6-8k for LED buffer and data combined. With "only" 600 LEDs, you'll have a massive 4.2-6.2k for data, which shrinks to 1-3k if you "need to" max out the LED count. Best of both worlds :)

wled00/FX.cpp
```diff
@@ -976,29 +947,27 @@ uint16_t WS2812FX::mode_chase_flash_random(void) {
 /*
  * Alternating pixels running function.
  */
-uint16_t WS2812FX::running(uint32_t color1, uint32_t color2) {
-  uint8_t pxw = 1 + (SEGMENT.intensity >> 5);
-  uint32_t cycleTime = 35 + (255 - SEGMENT.speed);
+uint16_t WS2812FX::running(uint32_t color1, uint32_t color2, bool theatre) {
+  uint8_t width = (theatre ? 3 : 1) + (SEGMENT.intensity >> 4);  // window
```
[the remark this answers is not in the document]
blazoncek: I'm not sure I completely understand.
By removing a base function and replacing it with another I had to make minor adjustments to keep visually similar end effect.
Casual user may never notice.
Even for the professional user with time critical effects this change did not make significant deviation and I did not receive any negative feedback.

wled00/FX.cpp
blazoncek: My base use case is a megatree with 16 segments, 50 LEDs each (and a star at the top).
Having a single segment with the same number of LEDs would definitely make it look boring in such case.
But then, when you cross the 600 LEDs barrier you are already in the ESP32 territory where the number of fragments increases.
For the 8266 my actual use case is a megatree with 12 segments, 33 LEDs each (and the star) totaling slightly above 420 LEDs and my roof trim with 4 segments and approximately 600 LEDs. In both cases the effect still looks ok.

If we are to employ dynamic max fragment calculation we should only do it at the time when number of segments change.

wled00/FX.h
blazoncek: Perhaps a redefinition of what someone should expect from 8266 would be in order.
We made many changes that indeed push 8266 to the limits (if not beyond) so reducing the maximum number of LEDs would not be too bad (1200 perhaps, 2 strips of 600 each?).
I may have exaggerated with maximum segment numbers since a few users requested it, but none of them used maximum number of LEDs.

Perhaps we could make a maximum number of segments dependant on maximum number of LEDs?

wled00/FX.h
blazoncek: If you agree we can lower the max number to 16 and recommend ESP32 for greater number of segments.
IMO 12 may be too low if someone wants to use a staircase usermod.

wled00/FX.h
blazoncek: Using PSRAM I can see a drop of 5-7 FPS (using Fireworks Starburst effect which uses most memory) with 3-6 segments (600 to 300 pixels each, total of 1800 pixels).

wled00/FX.h
fishbone-git: Can these parameters not be something left to platformio.ini to define, based on the device type? some code to do a sanity check for "safe" limits would still be required, of course.  I would expect anyone wanting to push the limits in terms of LEDs and segments shouldn't be balking at the extra cost of ESP32, but I sense there are still quite a few using this to mod devices with very lightweight devices built in (I do both)

wled00/data/index.js
blazoncek: My functions for segments look like this:
```javascript
function makeSeg()
{
	var ns = 0;
	if (lowestUnused > 0) {
		var pend = parseInt(gId(`seg${lowestUnused -1}e`).value,10) + (cfg.comp.seglen?parseInt(gId(`seg${lowestUnused -1}s`).value,10):0);
		if (pend < ledCount) ns = pend;
	}
	gId('segutil').scrollIntoView({
		behavior: 'smooth',
		block: 'start',
	});
	var cn = `<div class="seg">
	<div class="segin expanded">
		<input type="text" class="ptxt noslide" id="seg${lowestUnused}t" autocomplete="off" maxlength=32 value="" placeholder="New segment ${lowestUnused}"/>
		<table class="segt">
			<tr>
				<td width="38%">Start LED</td>
				<td width="38%">${cfg.comp.seglen?"LED count":"Stop LED"}</td>
			</tr>
			<tr>
				<td><input class="noslide segn" id="seg${lowestUnused}s" type="number" min="0" max="${ledCount-1}" value="${ns}" oninput="updateLen(${lowestUnused})"></td>
				<td><input class="noslide segn" id="seg${lowestUnused}e" type="number" min="0" max="${ledCount-(cfg.comp.seglen?ns:0)}" value="${ledCount-(cfg.comp.seglen?ns:0)}" oninput="updateLen(${lowestUnused})"></td>
				<td><button class="btn btn-xs" onclick="setSeg(${lowestUnused});resetUtil();"><i class="icons bth-icon" id="segc${lowestUnused}">&#xe390;</i></button></td>
			</tr>
		</table>
		<div class="h" id="seg${lowestUnused}len">${ledCount - ns} LEDs</div>
		<div class="c"><button class="btn btn-p" onclick="resetUtil()">Cancel</button></div>
	</div>
</div>`;
	gId('segutil').innerHTML = cn;
}

function resetUtil()
{
	gId('segutil').innerHTML = '<button class="btn btn-s" onclick="makeSeg()"><i class="icons btn-icon">&#xe18a;</i>Add segment</button>';
	for (var i=0; i<expanded.length; i++) if (expanded[i]) expand(i); // collapse all expanded elements
}
```

My populateSegments() is a bit different since I exposed power button and brightness slider. I would recommend doing the same in master.

wled00/FX.cpp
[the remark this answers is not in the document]
Aircoookie: That's not at all what I'm saying, your adjustments are perfect :)

I was referring to the impact of the intensity slider, and whether a higher intensity setting should increase or decrease a section's width. I've since made up my mind though, more intensity == wider makes more sense here, both for consistency and compatibility.

wled00/FX.h
Aircoookie: Staircase is a very good usecase for this indeed. Just did some thourough testing and 16 segments seem to be a good limit. 16 named segments used 9187 bytes of JSON buffer, just barely fitting into the 9216 buffer. And there is also unpredictable buffer usage for usermod JSON. We might need to go with 10240 buffer size, I hope the extra 1K of allocation won't cause instabilities here.

What has me completely baffled though is that I'm observing a 200 b per _active_ segment higher memory usage even without segment names or effects using segment data. This is really strange, as segments do no dynamic allocation apart from name/data. Additionally, names seem to use about double the mem they should be (16 x 32-char long name uses ~1.2k, where it should only 512b).
My only explanation as to why this is happening is that the last network request is still buffered so that the serialized string length somehow matters (and a serialized segment string is 187 characters long, so that would be fitting). If that were the case, `/json`,`/json/si` and `/json/info` would all provide different values though, which does not happen.

Aircoookie: Just checked, this 200 b per seg behavior is also present in 0.13.0-b2. Thus not a new issue introduced by segment names.

blazoncek: Thanks for your thorough testing and debugging.
I used 10k JSON buffer in the past but have reduced it to 9k when we did pre-serialize `col[]`.
Regarding the dynamic JSON buffer: I was thinking about serializing HTTP requests as to reduce chance of _out of memory_ conditions. What do you think?

Aircoookie:
Personally might prefer the square style for the apply button purely because of symmetry, the round one looks also nice though.

PR #2186, opened 2021-09-11
Can-Bus intergration ESP32

badFreddy0815:
Can-Bus integration
it would be great if one could change the presets via canbus because the ESP32 has integrated canbus


Aircoookie:
Hey! I think you meant to open this as an enhancement request (issue). Pull requests are for when you made code changes you would like to contribute.

Made an issue for it: https://github.com/Aircoookie/WLED/issues/2188

PR #2201, opened 2021-09-18
Allow normal WLED Control of DMX fixtures when not getting any realtime data

jeffc:
This lets WLED control a DMX fixture (or series of fixtures) as a "normal light" without taking away the ability to pass raw DMX data over the network.

This was useful to me because I wanted to be able to run routines that controlled the RGBW/pan/tilt/effects of a fixture during shows, but have WLED simply control RGBW (and leave the more advanced features alone) when a show wasn't running.

I'm open to making this behavior explicitly toggleable if you think that's necessary (eg, in case someone wants the DMX bus to be quiet when not getting data rather than WLED reassuming control of it)


platformio.ini
```diff
@@ -300,7 +300,7 @@ board_build.partitions = ${esp32.default_partitions}
 
 [env:esp32_eth]
 board = esp32-poe
-platform = espressif32@2.0
+platform = espressif32@2.1
```
blazoncek, line 303: Is this change necessary?
If yes: Why not updating to 3.4?

jeffc: Ugh, I didn't realize the PR was pulling in all the commits I made on the branch since I created it. I assumed it was frozen at the commit I based the PR on.

I've been using the branch for local dev on a specialized use case since then, so it's no longer in any state to merge. Reverting this PR, and I'll untangle it and open a new one later

blazoncek: No problem. I do not know any of DMX stuff so was just curious about changes I understood. 😄

PR #2206, opened 2021-09-18
Avoid redundant localIP calls, each call takes 0.700 us on ESP32 240Mhz

pbolduc:
Removes redundant localIP calls. Based on benchmarks `WiFi.localIP()` takes 0.700 us on ESP32 240 MHz.  Not super expensive, but why waste 168 clock cycles if we dont need to.

### DIFF
--- a/wled00/src/dependencies/network/Network.cpp
+++ b/wled00/src/dependencies/network/Network.cpp
@@ -2,18 +2,18 @@
 
 IPAddress NetworkClass::localIP()
 {
-  IPAddress ipAddress;
+  IPAddress localIP;
 #if defined(ARDUINO_ARCH_ESP32) && defined(WLED_USE_ETHERNET)
-  ipAddress = ETH.localIP();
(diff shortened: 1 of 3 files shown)


blazoncek:
Seems reasonable. 👍

wled00/src/dependencies/network/Network.cpp
```diff
@@ -3,12 +3,12 @@
 IPAddress NetworkClass::localIP()
 {
 #if defined(ARDUINO_ARCH_ESP32) && defined(WLED_USE_ETHERNET)
-  if (ETH.localIP()[0] != 0) {
-    return ETH.localIP();
-  }
+  IPAddress ipAddress = ETH.localIP();
```
Aircoookie, line 6: This is not equivalent. If `ETH.localIP()[0] == 0`, it should try `WiFi.localIP()` next. WiFi is still supported in ETH builds.

Other than that, a very welcome optimization :)

PR #2214, opened 2021-09-19
Fixed pin reservations.

blazoncek:
Added ethernet pin reservations.
Minor tweaks in usermods.
@Aircoookie please review.


wled00/wled.cpp
```diff
@@ -8,6 +8,90 @@
 #include "soc/rtc_cntl_reg.h"
 #endif
 
+#ifdef WLED_USE_ETHERNET
```
Aircoookie, line 11: This should stay in its separate file, `wled.cpp` is already quite cluttered and should maybe itself be split at some point. We can add `wled_ethernet.cpp` if required.

wled00/xml.cpp
```diff
@@ -186,6 +187,52 @@ void sappends(char stype, const char* key, char* val)
   }
 }
 
+void extractPin(JsonObject &obj, const char *key) {
```
Aircoookie, line 190: This is a good solution until we remake settings to read `cfg.json` directly.

Aircoookie: An possibly more lightweight alternative would be adding the pins currently allocated in `PinManager`, although I'm not sure if each usermod will actually allocate all pins it has configured in `cfg.json`, although that would likely be the case

wled00/xml.cpp
```diff
@@ -264,50 +312,72 @@ void getSettingsJS(byte subPage, char* dest)
...
-              if (i++) oappend(SET_F(","));
-              oappendi(obj["pin"].as<int>());
-            }
+    if (!mods.isNull()) fillUMPins(mods);
+
+    #ifdef WLED_ENABLE_DMX
+      oappend(SET_F(",2")); // DMX hardcoded pin
+    #endif
+
+    #ifdef WLED_ENABLE_ADALIGHT
```
Aircoookie, line 331: Can be removed. Prohibiting pin 3 just because Adalight could be received is not reasonable.
Adalight receiving code will not allocate the pin. Serial data in is not something we can control/manage in WLED code (with the exception of not sending `Ada` on boot if pin 3 is allocated), it is sent by the computer "just like that".

Aircoookie: Just checked, Adalight/ Serial JSON receiving code is now disabled when the pin is allocated :) For most users, pin 3 availability is more crucial than Adalight functionality.

wled00/wled.cpp
blazoncek: Yes, there are a few functions and perhaps globals that would be petfect fit for fcn.cpp.

wled00/xml.cpp
blazoncek: Talked about that withHenry Gabrielsky.

wled00/xml.cpp
blazoncek: I opted to have Ada light compile in optionally instead of default.
Might consider the same for master (only change wled.h)

Aircoookie: Let's have it compiled in by default. The main use case is people using the instance as an ambilight with just a single bus on pin 2/16.
Instead, what do you think about a warning like this for LED/UM settings?
`⚠ Using pin 3. Adalight and Serial JSON will be disabled.`

blazoncek: This kind of approach adds whole lot of issues IMO.
One workaround could be to add checkbox somewhere in settings for enabling disabling Ada/Serial.

Aircoookie: No need for a checkbox, it is automatically disabled once pin 3 is allocated for anything else:
https://github.com/Aircoookie/WLED/blob/master/wled00/wled_serial.cpp#L24

A note in kno.wled.ge might also be sufficient to let users know that configuring pin 3 will disable serial input.

blazoncek: Ok. 😄

blazoncek:
@Aircoookie any progress on this?
IMO: Not specifying GPIO3 anywhere and not making reservation for Ada will allow user to select GPIO3 for usermod which will disable Ada/Serial use without proper notification to user. If you are OK with this, so am I. 😄

Aircoookie:
Yep, 100% fine with that 😄  I'm just now starting to make mockups for the new settings page layout (will share once I have a working layout demo!), and since that will run completely off `cfg.json`, so adding a warning then will add little overhead.
For the meantime, no additional warning is fine. I will make a kno.wled.ge page for Serial and add a note there (currently adalight and serial JSON are not documented well)

Sorry that I didn't completely get it the first time, but why did you decide to `extern` the ethernet config definitions? I've reverted that for now, but if there is a good reason I'm missing, we can of course keep the change.

Ready to merge IMO👍

blazoncek:
I moved DEFINITON of ethernet pin array to wled.cpp because I needed to include DELCARATION in cfg.cpp and/or set.cpp (no longer sure which from the top of my head).
I prefer to NOT have DEFINITIONs in .h files, just declarations. This allows for multiple includes without linker warnings/errors.
I suggest to move all such functions, globals, etc into a separate .cpp file (oappend(), ...)

Aircoookie:
Right, you needed to include it in xml.cpp.
As far as I'm aware, the only real drawback of header-only style (both declaration and definition in the same .h file) is that it needs to be recompiled each time a cpp file that `#includes` it is changed.
Still, I think I agree with you that separate declaration and definition is more solid against various linker issues. I will merge this as-is and then I'll make a new PR to create a `wled_ethernet.cpp` and maybe a `wled_util.cpp` for oappend etc. to slightly reduce the length of `wled.cpp`

PR #2217, opened 2021-09-20
Allocate segment data based on currently active segments

Aircoookie:
Allows more stars/sparks in Fireworks 1D and Starburst if fewer segments are configured.
Doubles stars/spark maximum if at most half of max segments is active.
Quadruples star/spark maximum if at most a quarter of max segments is active.


blazoncek:
I like what yodid here. 👍

PR #2244, opened 2021-10-02
Add new Usermod

Proto-molecule:
Please review this usermod for a possible addition. 
It's built on top of the original v2 Rotary Encoder and v2 Four Line display.
I took the latest version of both and changed the functionality around for a new looking user interface on the display.
Ive had some interest to add this and the creator of the original suggested it may be better to create a pull request as a new separate usermod instead of modifying the original. 
Ive added it in the simplest way that made sense to me, however it may not be the best.

### DIFF
--- a/usermods/usermod_v2_four_line_display_ALT/readme.md
+++ b/usermods/usermod_v2_four_line_display_ALT/readme.md
@@ -1,0 +1,45 @@
+# I2C 4 Line Display Usermod ALT
+
+Thank you to the authors of the original version of these usermods. It would not have been possible without them!
+"usermod_v2_four_line_display"
+"usermod_v2_rotary_encoder_ui"
+
+The core of these usermods are a copy of the originals. The main changes are done to the FourLineDisplay usermod.
+The display usermod UI has been completely changed.
+
+
+The changes made to the RotaryEncoder usermod were made to support the new UI in the display usermod. 
+Without the display it functions identical to the original.
+The original "usermod_v2_auto_save" will not work with the display just yet.
+
+Press the encoder to cycle through the options:
+    *Brightness
+    *Speed
+    *Intensity
+    *Palette
+    *Effect
+    *Main Color (only if display is used)
+    *Saturation (only if display is used)
+
+Press and hold the encoder to display Network Info
+    if AP is active then it will display AP ssid and Password
+
+Also shows if the timer is enabled
+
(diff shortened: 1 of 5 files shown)


blazoncek:
Hi,
I have taken a good look into these two usermods (since I finally got the rotary encoder) and made some minor adjustments.
The idea and implementation is fantastic (though it needed a bit of polishing).
I have yet to test them with ESP8266 but on ESP32 they work great - better than original - and provide a more useful experience.
If you are willing to check modifications they are currently in my fork (under same name) but will slowly get into master as well.
I will add a configuration option to display seconds while in clock mode (currently always displayed) and reduce the amount of data displayed at a time to reduce the CPU time spent in usermod.

Proto-molecule:
Hey Blaž,

Wow! thank you for looking at code and making improvements! I started coding earlier this year and this being my first interaction with GitHub, there is a LOT I don't know. Hopefully the code didn't look too "green". I'm honestly quite honored its part of the WLED master. 
And yeah, my main goal with the display was to make it simple enough for kids to use, and reduce CPU time so that it doesn't cause lag in the effects. Which is why I initially removed the clock seconds. You've managed to add it back in a cleaver way!
 I'll browse though the code and somehow let you know if I find anything.

blazoncek:
Hey there.
I've made some more changes. I moved the sorting code into rotary encoder usermod since it was not used elsewhere and reducing the number of needed usermods is important on 8266 since it can only have 4 (by default).
I am currently using 6 of them on my regular installs (Temperature, PIR sensor switch, 4 Line Display, Multi Relay, Animated Staircase, Auto Save). I could skip Animated Staircase though.

Proto-molecule:
Ive browsed over and created a fully request of some minor bugs.

https://github.com/blazoncek/WLED/pull/12

sashakozlov86:
Hello!  Thanks for the cool addon for WLED!!!  I wanted to clarify with you: how can I translate the inscriptions on the display "brightness", "effect speed", "intensivity", "saturation"  abbreviated names of the months of the year and so on into Cyrillic?  I suspect that I just need to rename some variables in some file or the value of the print command on the display, but I can’t figure out where exactly.  I will be very grateful to you!  Thanks a lot!

blazoncek:
Please use WLED [forum](https://wled.discourse.group) or Discord.

catalinalb:
Are here any updates? How can I use this rotary encoder and the display? I have changed the usermod in the platformIO main file, uploaded, connected the display and the encoder, but nothing...

PR #2259, opened 2021-10-08
Add strip off refresh option in LED settings.

blazoncek:
New strip initialization logic.
Minor code clen-up.


Aircoookie:
Merging so that b4 can be released :) If there are issues, let's discuss here further. If I missed something major, a hotfix to master is fine.

PR #2261, opened 2021-10-09
Basic Usermod Overlay support & Seven Segment UM

AK5nowman:
This adds overlay support to usermodsv2. It does not remove legacy overlay support from the main code base. Need to refactor the remaining overlay before that can be done.


Aircoookie:
Thank you. Approving of the extra usermod hook :)
@blazoncek feel free to merge once strings are optimized

PR #2284, opened 2021-10-20
Multirelay button support.

blazoncek:
Added button hook for usermods.

### DIFF
--- a/wled00/button.cpp
+++ b/wled00/button.cpp
@@ -10,10 +10,11 @@ static const char _mqtt_topic_button[] PROGMEM = "%s/button/%d";  // optimize fl
 
 void shortPressAction(uint8_t b)
 {
-  if (!macroButton[b])
-  {
-    toggleOnOff();
-    colorUpdated(CALL_MODE_BUTTON);
+  if (!macroButton[b]) {
+    switch (b) {
+      case 0: toggleOnOff(); colorUpdated(CALL_MODE_BUTTON); break;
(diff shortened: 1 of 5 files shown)


wled00/button.cpp
```diff
@@ -186,61 +227,43 @@ void handleButton()
...
-            applyPreset(macroDoublePress[b], CALL_MODE_BUTTON);
-  
-            // publish MQTT message
-            if (buttonPublishMqtt && WLED_MQTT_CONNECTED) {
-              char subuf[64];
-              sprintf_P(subuf, _mqtt_topic_button, mqttDeviceTopic, (int)b);
-              mqtt->publish(subuf, 0, false, "double");
-            }
-          } else buttonWaitTime[b] = millis();
-        } else shortPressAction(b);
```
Aircoookie, line 236: I appreciate the extra usermod hook and the `button.cpp()` cleanup by moving the actions into separate functions :)

Just one minor nitpick:
Rather than a default double press button action (set as toggle on button 0, cycle palettes on button IDs >0), for UX in my opinion it would be preferable to fire the short press action immediately upon release if no double press macro is defined. Otherwise, users who don't wish to use double press will still have a 350ms delay on every press until the short press action is fired.

I have changed this back just for button 0 right now (since its default double press action is just on/off, same as the short press action anyways), on buttons >0 it is debatable if the default double press action (palette cycle) or the potential for instant short press action is more desirable, so I'd be fine with leaving it as-is for now.

blazoncek: Agreed on all topics.

Issue #627, opened 2020-01-25
Play tpm2 files with sd card


harueg:
Really great software !  Thank you !

Would it be possible to read tpm2 files from a sd card.

file format :  https://www.ledstyles.de/index.php?thread/18969-tpm2-protokoll-zur-matrix-lichtsteuerung/

Existing stand alone software  : https://github.com/TheChief79/tpm2arduino

https://hackaday.io/project/5714-glediator-from-sdcard-arduino

Thanks !

Aircoookie:
Hi, that indeed seems like a very good addition. Additionally I've received a request for being able to playback .fseq files (LED data format xLights uses), so those two feature requests are very similar except for the specific protocol used!

harueg:
Yep, that would be great  both protocols should work for me.
Thank you

harueg:
Hi @Aircoookie 
Any news on this topic ?

Thanks

Harald

Aircoookie:
Hi!
No, sorry. I know it's been a while and this is definitely coming, but I'll need to add some other features first, like basic file sysyem support.

harueg:
Ah, ok.
So I will keep looking forward to it !

Thanks and good night
Harald

Legsmaniac:
Cool!  I'll look forward to this.
This is something that fits nicely in with my recent requests too.
Thank you.

timdonovanuk:
+1 to this! Would be cool to see FSEQ format support in WLED - you could then just create an effect/show in XLights and upload and have it appear as an effect

pbolduc:
Related issue #1278

harueg:
Hi @Aircoookie 

Just trying, well a programmer, to get this fseqlib modified to neopixel(bus), because fastled still doesn't support rgbw.
That's trickier than thought.  Would be great, if that implementation in WLED would work some when.
Some pictures from a recent project with "glitter" effect.
![20210109-9T4A2848](https://user-images.githubusercontent.com/60239890/111679351-81540e80-8821-11eb-8800-0dc4f4f657ee.JPG)
![20210109-9T4A3298](https://user-images.githubusercontent.com/60239890/111679360-85802c00-8821-11eb-9b0d-407ef1447e21.JPG)
![20210109-9T4A3490](https://user-images.githubusercontent.com/60239890/111679405-887b1c80-8821-11eb-909f-d270130da960.JPG)

processor83:
[...]

PR #2292, opened 2021-10-27
TPM2 Playback from LittleFS and SD

constant-flow:
## Changes
- new `esp32-eth-sd` env
- tpm2 playback from LittleFS and SD
- play recording as live-mode
- made tpm2 uploadable via `/edit` interface

## Related
- [Issue #627](https://github.com/Aircoookie/WLED/issues/627)

### DIFF
--- a/wled00/tpm2record.cpp
+++ b/wled00/tpm2record.cpp
@@ -1,0 +1,292 @@
+#include "wled.h"
+
+#ifdef WLED_USE_SD
+#define USED_STORAGE_FILESYSTEMS "SD, LittleFS"
+#include "SD_MMC.h"
+#else
+#define USED_STORAGE_FILESYSTEMS "LittleFS"
+#endif
+
+// This adds TPM2-file storing and playback capabilities to WLED.
+// 
+// What does it mean:
+//   You can now store short recorded animations on the ESP32 (in the ROM: no SD required) with a connected LED stripe.
+//
+// How to transfer the animation:
+//   WLED offers a web file manager under <IP_OF_WLED>/edit here you can upload a recorded *.TPM2 file
+//
+// How to create a recording:
+//   You can record with tools like Jinx
+//
+// How to load the animation:
+//   You can specify a preset to playback this recording with the following API command
+//   {"tpm2":{"file":"/record.tpm2"}}
+//
+//   You can specify a preset to playback this recording on a specific segment
+//   {"tpm2":{"file":"/record.tpm2", "seg":{"id":2}}
+//   {"tpm2":{"file":"/record.tpm2", "seg":2}
+//
+// How to trigger the animation:
+//   Presets can be triggered multiple interfaces e.g. via the json API, via the web interface or with a connected IR remote
+//
+// What next:
+//   - Playback from SD card is the next plan, here the length of the animation is less of a problem.
+//   - Playback and Recording of RGBW animations, as right now only RGB recordings are supported by WLED
+
+// reference spec of TPM2: https://gist.github.com/jblang/89e24e2655be6c463c56
+// - A packet contains any data of the TPM2 protocol, it
+//     starts with `TPM2_START` and ends with `TPM2_END`
+// - A frame contains the visual data (the LEDs color's) of one moment
+
+// --- CONSTANTS ---
+#define TPM2_START      0xC9
+#define TPM2_DATA_FRAME 0xDA
+#define TPM2_COMMAND    0xC0
+#define TPM2_END        0x36
+#define TPM2_RESPONSE   0xAA
+
+// infinite loop of animation
+#define RECORDING_REPEAT_LOOP -1
+
+// Default repeat count, when not specified by preset (-1=loop, 0=play once, 2=repeat two times)
+#define RECORDING_REPEAT_DEFAULT 0
+
+// --- Recording playback related ---
+File     recordingFile;
+uint8_t  colorData[4];
+uint8_t  colorChannels    = 3;
+uint32_t msFrameDelay     = 33; // time between frames
+int32_t  recordingRepeats = RECORDING_REPEAT_LOOP;
+unsigned long lastFrame   = 0;
+uint16_t playbackLedStart = 0; // first led to play animation on
+uint16_t playbackLedStop  = 0; // led after the last led to play animation on
+
+// skips until a specific byte comes up
+void skipUntil(uint8_t byteToStopAt)
+{
+  uint8_t rb = 0;
+  do { rb = recordingFile.read(); }
+  while (recordingFile.available() && rb != byteToStopAt);
+}
+
+void skipUntilNextPacket()  { skipUntil(TPM2_START); }
+
+void skipUntilEndOfPacket() { skipUntil(TPM2_END); }
+
+void getNextColorData(uint8_t data[])
+{
+  data[0] = recordingFile.read();
+  data[1] = recordingFile.read();
+  data[2] = recordingFile.read();
+  data[3] = 0; // TODO add RGBW mode to TPM2
+}
+
+uint16_t getNextPacketLength()
+{
+  if (!recordingFile.available()) { return 0; }
+  uint8_t highbyte_size = recordingFile.read();
+  uint8_t lowbyte_size = recordingFile.read();
+  uint16_t size = highbyte_size << 8 | lowbyte_size;
+  return size;
+}
+
+void processCommandData()
+{
+  DEBUG_PRINTLN("processCommandData: not implemented yet");
+  skipUntilNextPacket();
+}
+
+void processResponseData()
+{
+  DEBUG_PRINTLN("processResponseData: not implemented yet");
+  skipUntilNextPacket();
+}
+
+void processFrameData()
+{
+  uint16_t packetLength = getNextPacketLength(); // opt-TODO maybe stretch recording to available leds
+  uint16_t lastLed = min(playbackLedStop, uint16_t(playbackLedStart + packetLength));
+
+  for (uint16_t i = playbackLedStart; i < lastLed; i++)
+  {
+    getNextColorData(colorData);
+    setRealtimePixel(i, colorData[0], colorData[1], colorData[2], colorData[3]);
+  }
+
+  skipUntilEndOfPacket();
+
+  strip.show();
+  // tell ui we are playing the recording right now
+  uint8_t mode = REALTIME_MODE_TPM2RECORD;
+  realtimeLock(realtimeTimeoutMs, mode);
+
+  lastFrame = millis();
+}
+
+void processUnknownData(uint8_t data)
+{
+  DEBUG_PRINT("processUnknownData - received:");
+  DEBUG_PRINTLN(data);
+  skipUntilNextPacket();
+}
+
+void clearLastPlayback() { 
+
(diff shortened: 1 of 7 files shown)


wled00/tpm2record.cpp
```diff
@@ -0,0 +1,292 @@
...
+#define RECORDING_REPEAT_LOOP -1
+
+// Default repeat count, when not specified by preset (-1=loop, 0=play once, 2=repeat two times)
+#define RECORDING_REPEAT_DEFAULT 0
+
+// --- Recording playback related ---
+File     recordingFile;
+uint8_t  colorData[4];
+uint8_t  colorChannels    = 3;
+uint32_t msFrameDelay     = 33; // time between frames
```
guardmedia, line 58: @constant-flow - Is there a reason why we have a 33 ms delay between frames? If not, can we make this parameter driven, or part of the TPM2 file header somehow?

Use case that I'm thinking of is converting xLights export 'fseq' files (Falcon Player) into TPM2 format which typically uses 25ms, 40ms, or 100ms per frame (40 fps, 20 fps, and 10 fps respectively).

constant-flow: @guardmedia Currently it's not bound to any limitation, it's simple for the sake of having a defined state. The TPM2 spec is rather lose on that end. The most recent specs are these from 2013 [LINK (sorry in 🇩🇪 )](https://www.ledstyles.de/content/index.php?attachment/28200-tpm2-specs-v1-0-2013-ger-pdf/)

There is a command `0x0C` that can be used to communicate the master speed. Unfortunately there is no concrete example what the payload should look like to set it to a specific frame rate, e.g. 60FPS

I was in contact with one of the original authors (Pesi) of the TPM2 protocol. He is not having the time nor the need as of now to continue the active development. He is still online in the **ledstyles** forum and would be okay if someone else continues the work on the "standard". My plan was to take their specifications and add clear examples and more precise rules how to set certain parameters, basically a new spec version. Next to framerate I'd also define color channel handling, e.g is it RGB or GRB or RGBW that the file is storing. The new specs i'd base of an [english translation](https://gist.github.com/jblang/89e24e2655be6c463c56) that already exists and publish it on Github. If Pesi likes the new specs, he might even promote it and host it under his reserved `tpm2.de` domain.

So I'm totally with u, we should have the framerate and other information packed into the file. I simply haven't had the time to work on the new specs with examples ... but I got the okay from both parties to alter what is there :) So maybe the WLED community can push (revive) the TPM2 format. (naming is a bit unlucky as it's conflicting with ISO/IEC 11889:2015)

A converter or player for fseq is definitely a cool and required feature for some more advanced / custom animations or use-cases.

guardmedia: Thank you for the detailed response! Doing a quick search through the TPM2 spec, it seems that framerate is typically driven by the sending source, baud rate, or processing power. In all documentation I could find, the actual FPS was always determined by the author. For example, [Jinx!](http://www.live-leds.de/) uses a fixed framerate of 25fps or 20fps if using the **-p** switch.

From the Jinx! user manual (http://www.live-leds.de/download/190/) ...

Page 27
> _**Attention:** Jinx! will always produce 25fps. So be sure that your output device is able to handle it (e.g. serial
line with a too small baud rate), otherwise a frame skip will appear_

Page 35
> **Limit output frame rate to 20fps**
Additionally you can use the -p switch to limit the output frame rate to 20fps. This option is for debugging
purposes and not designed to use in a production state. The internal generators will work with 25fps
regardless off the -p switch.

Long story short, 33 ms delay (~30 fps) sounds like a reasonable rate. Once I have more time, I will look into implementing FSEQ directly instead of trying to re-write the TPM2 protocol.

Thank you for the great idea to add SD ROM support into WLED - looking forward to seeing this getting merged!

blazoncek:
I will do my best to get this merged. Not in 0.13 though.

wled00/tpm2record.cpp
constant-flow: Nice to see your implementation for the FSEQ format. Yeah I checked the "default" FPS but sorry 20/25 FPS is not really what i thought is usable 😂 . My recorder uses 30 FPS for now (I hope i can publish it as well).

The ideal solution is to set FPS via the headers in the recording, we all agree.

Let's see when our PR will be merged in 💍

ElToberino:
I hope it will be possible to merge this PR soon.
I like the support of ArtNet/sACN very much because it enbales easily driving large matrixes via programs like Jinx! (Without the need of configuring the matrixes via WLED.)
Adding the support of tpm2-playing from SD would increase the possibilties of WLED into a new dimension - with an "infinte" number of custom _standalone_ effects/Shows recorderd with Jinx! or similar programs. 
Especially used in combination with the timers/schedules this could be really be killer feature and would make WLED even more unique.

ElToberino:
Meanwhile I compiled V0.13-b6 with these changes.
Everything works fine - it's really a great feature. I wonder if it would make sense to make the variable "recordingRepeats" (in wled00/tpm2record.cpp) configurable via JSON. I know it's possible just to use the playlist feature but I think the confugurable value would be a simpler and straighter solution.

blazoncek:
Hi @constant-flow 

I was finally able to delve into the code, examine it and prepared a few suggestions.

First of all, idea is great and has a lot of potential - as a usermod.

If you can rewrite it as a usermod (you may check existing usermods: multi relay, temperature, PWM fan, etc) it has a greater chance of being merged as it becomes optional for users. It will also allow for greater flexibility as the usermod framework allows usermods to tap into core WLED without changing original source.
Even though this may seem difficult at first glance it is rather simple (no need to change wled.cpp or json.cpp, etc). Once you put it into basic shape I can help you fine-tune (via draft PR).

Second, I liked the idea of employing SD card.

Unfortunately it turns out that there are 2 ways to connect SD card to ESP: MMC & SPI. Enhancing (usermod) with the ability to use either would be best as it seems that higher level API for both is the same. The actual connection may be selected using usermod's configuration page.

I am sorry this has taken so long to review the code but I had other priorities unfortunately.

I hope you will agree with the idea of rewriting your enhancement as a usermod.

Best!

blazoncek:
No comment?
Shall we then just close the PR?

constant-flow:
Currently on vacation 🌴 ☺️
Let me check later if i find some time.

Thanks for your suggestions though I'm still not so sure how easy an integration as a usermod would be.

blazoncek:
No problem.
It is actually quite easy to use usermod framework.

BTW take a look at my fork, audioreactive-prototype (or 2D-support) branch for the inspiration.

constant-flow:
I had a look into the docs again and finally understood how Usermods can alter the capabilities of the API-Commands.
After a quick test I can verify that the usermod-route doesn't break a major feature (the only thing so far is when selecting a recording in the `/edit` route, I was able to peek into the animation ... maybe we can add this to usermods feature-list too ... but thats a nice to have)

Also now it relies on the `REALTIME_MODE_GENERIC` instead of a more descriptive one.

My plan now is to clean it up a bit more and abstract it from the recording format, to enable not only TPM2 but also any other format future devs want to playback like fseq, so they can easily support their format with this usermod.

```
{"playback":
   {
      "file":"/record.tpm2",
      "format":"tpm2"
   }
}
```

Concerning the two ways to access the SD. I think I could integrate that via the usermod settings (if necessary). Do you have a link to projects using the MMC and one using SPI, so I can have a look whats the general difference in code to access files on them?

blazoncek:
Hi. Checked your usermod prototype linked above.
Please utilise addToConfig() and readFromConfig() to store/load SD card type and GPIO selection as well as any other configurable option.
Do the initialization in setup() (consider also that GPIO pins or type may change during runtime - do not require reboot if possible).
There is plenty of helpful information in existing usermods provided above.
Unfortunately I do not have any information on how to implement MMC and/or SPI interface other than a few Google links.

constant-flow:
Please check update here about [new progress](https://github.com/Aircoookie/WLED/pull/2413#issuecomment-1236206323) 

Anything missing for a PR?

constant-flow:
@ElToberino thx for checking and reporting errors.

> 1.) In the usermod descrpition you wrote that default fps rate is 25. But when I played a tpm2-file without defining a fps rate, the animation (recorded with Jinx! @ 25 fps) was way too fast. Manually setting the fps rate to 25 via preset worked and the animation was played with the proper speed.

Looks like I forgot a default value from what you report. Will have a look. @harueg the TPM2 doesn't specify a framerate via the header. It's 25 FPS by convention from **jinx!** I think the [TPM2 "standard" 🇩🇪](https://www.ledstyles.de/index.php?thread/18969-tpm2-protokoll-zur-matrix-lichtsteuerung/) doesn't even define a default framerate  

> 2.) Skipping / Leaving the real time mode didn't work. In opposite to the web interface which behaved properly and left the realtime mode, the leds still continued playing the animation - flickering and overlayed by the now chosen WLED effect.

I will check on that, I hope I'll find some time to investigate. Last time the interface was still showing it being in realtime mode even though it stopped it via `exitRealtime()` but i didn't test it with concurrent animations.

constant-flow:
The more I look into this messed up PR the more I feel I should close this PR and open 2 new PRs
1.) playback-recordings (TPM2)
1.) SD card support

I hope i find the time to do it soon. Maybe then, things will be merged into main as usermod

harueg:
Here is the plan for the protocol, I don't think everything has been implemented yet.
https://www.ledstyles.de/content/index.php?attachment/28200-tpm2-specs-v1-0-2013-ger-pdf/
A frame rate isn't mentioned.
 And thank you !

ElToberino:
@constant-flow:
I took a deeper look at the code and found a solution for the two issues in usermod_playback_recordings.h:
1. change code in line 56
`#define PLAYBACK_FRAME_DELAY 40`   
     // set default fps rate:    calculation: 1000/PLAYBACK_FRAME_DELAY = fps -> 1000/40 = 25fps

2. add code in line 411:
`if (realtimeOverride == REALTIME_OVERRIDE_ALWAYS) realtimeOverride = REALTIME_OVERRIDE_ONCE;`   
     // workaround - not 100% satisfying, but necessary with current real time mode web interface
 `if (realtimeOverride == REALTIME_OVERRIDE_ONCE)  { 
    exitRealtime();
    playbackFile.close();
    file_clearLastPlayback();
  }`
     // stopping file playback and leaving real time mode on web interface request

With these modifications everything works as intendend.
I also think it would be better to start a new PR. Your tpm2 feature is really essential and should be merged asap.

@harueg:
I don't think the tpm2 protocol will ever be changed or extended. That is not required and would break many applications. It's widely used on many players and devices - most of them are programmed for a fps rate of 25 because of the Jinx! output rate.

constant-flow:
@ElToberino thx for investigating and solving this issue 🙇 . I merged the current main into my branch (v14) and added your fixes. It's still on my [branch](https://github.com/constant-flow/WLED/tree/feature/playback-recording), from what I tested the fixes work, maybe check it.

I'll try PR both mods individually soon, didn't have enough time today.

constant-flow:
So i created to PRs to close this one

[PR1 : Playback of .tpm2](https://github.com/Aircoookie/WLED/pull/2876)
[PR2 : SD card support: MMC or configurable SPI](https://github.com/Aircoookie/WLED/pull/2877)

PR #2323, opened 2021-11-08
Fix settings JS buffer too small

### LINKED ISSUE #2303: Instability with 10 busses
I noticed that there is a bug on version 0.13.0-b4 "Toki".  in practice, if you use all 10 LED outputs it is no longer possible to reach the favorite LED menu, the page no longer loads.  the ESP must be reset.  thanks a lot

Aircoookie:
The previous buffer size was 2048b. This was not enough for the LED settings page with 10 busses.
This increases it to 3096b for ESP32 (issue should not be present on ESP8266 due to the 3 bus limit).
Measured free stack remaining before the change is 4k, after the change 3k, which should be enough leeway to not cause a stack overflow.

Likely fixes #2303

### DIFF
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -265,7 +265,11 @@
 #endif
 
 // string temp buffer (now stored in stack locally)
-#define OMAX 2048
+#ifdef ESP8266
+#define SETTINGS_STACK_BUF_SIZE 2048
+#else
+#define SETTINGS_STACK_BUF_SIZE 3096 
+#endif
 
 #ifdef WLED_USE_ETHERNET
 #define E131_MAX_UNIVERSE_COUNT 20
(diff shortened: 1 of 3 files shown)


blazoncek:
Risky. Why not malloc() (or new)?

Aircoookie:
Dynamic alloc is risky in its own way 😄 Memory fragmentation, running out (arguably not on ESP32) and programmer error (not freeing something properly, I admit that happened to me quite a few times).

If there is no/little risk of stack overflow, doesn't hurt to use stack variables if they are only needed on a local scope. malloc is the correct tool if you need the memory on a larger scope, or if you need so much memory that it does not or does not for sure fit inside the stack. As far as I can judge, increasing this buf from 2k to 3k should cause no issues.
What I agree on is that I dislike this function, `String(buf)` is just blasphemy. I'll be happy to see it go with the new settings page.

PR #2334, opened 2021-11-11
Improv support

Aircoookie:
Add support for the [Improv serial](https://www.improv-wifi.com/) protocol, allowing configuration of WiFi credentials over Serial.
This is also used by ESP Web Tools v4, this will enable connecting the unit right after installation using install.wled.me without using WLED-AP.

Demo: https://install.wled.me/v2.htm (note, currently works best with an ESP8266)


Aircoookie:
Drafting as there will be a small breaking change to the improv protocol to simplify the checksum.

PR #2336, opened 2021-11-11
Single json buffer

Aircoookie:
JSON buffer is statically allocated, this should help reduce memory fragmentation and also offer a small performance benefit.
Need to check that this works if a network request comes in while something in the loop is using the buffer (IR/config/ledmap/preset cycle/serial/MQTT)

By @blazoncek


wled00/cfg.cpp
```diff
@@ -407,13 +407,12 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
...
-    bool allComplete = usermods.readFromConfig(usermods_settings);
-    if (!allComplete && fromFS) serializeConfig();
+    needsSave = usermods.readFromConfig(usermods_settings);
   }
 
-  if (fromFS) return false;
+  if (fromFS) return needsSave;
   doReboot = doc[F("rb")] | doReboot;
-  return (doc["sv"] | true);
+  return (doc["sv"] | needsSave);
```
Aircoookie, line 415: Let's keep the default value `true` here. This is for changing settings from the `/cfg/json` endpoint only

wled00/json.cpp
```diff
@@ -809,23 +809,31 @@ void serveJson(AsyncWebServerRequest* request)
     return;
   }
 
+#ifdef WLED_USE_DYNAMIC_JSON
   AsyncJsonResponse* response = new AsyncJsonResponse(JSON_BUFFER_SIZE);
-  JsonObject doc = response->getRoot();
+#else
+  while (jsonBufferLock) delay(1);
```
Aircoookie, line 815: The more I think about it, the more I think this can go horribly wrong in case JSON is requested via network while `jsonBufferLock` is already true from a loop task because the network interrupt happened just while another thing was using the buffer. As far as I know `delay()` does not work correctly in the system/interrupt context, and even if it did, it would lock up as it would not return to the main loop to finish the task.

Of course it seems to work just fine regardless (perhaps because the time the doc is locked is short), but somehow I don't have a good feeling about it.

I have two ideas how we could guarantee for everything to not intersect:
1) use `noInterrupts()` and `interrupts()` before and after every use of doc in the loop. This could have some very hard to debug side effects though if something within that code relies on interrupts, or interrupts can't get executed in time
2) Instead of locking (one loop task using doc can't intersect another anyways, as they are all run sequentially), set a dirty flag from the interrupt to let the loop know that it needs to clean the document and start over. If e.g. a Serial JSON command gets dropped because of that, that is way less critical than a potential crash. The only thing I'd be worried about is preset saving, would be bad for the doc to go while it is being written to the FS...

https://arduino.stackexchange.com/questions/70369/disable-interrupts-when-doing-critical-things is interesting

wled00/cfg.cpp
blazoncek: Ok.

wled00/cfg.cpp
```diff
@@ -407,13 +407,12 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
   DEBUG_PRINTLN(F("Starting usermod config."));
   JsonObject usermods_settings = doc["um"];
   if (!usermods_settings.isNull()) {
-    bool allComplete = usermods.readFromConfig(usermods_settings);
-    if (!allComplete && fromFS) serializeConfig();
+    needsSave = usermods.readFromConfig(usermods_settings);
```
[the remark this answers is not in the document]
blazoncek: Boolean logic gives me headaches more often than not.
😀

wled00/json.cpp
blazoncek: Indeed. I have noticed 2 cases of lockups (with no apparent requests happening but I do have MQTT enabled).
This will need serial debugging which I will do in the next few days.

blazoncek: One thing that crosses my mind is to add a time-out option to the loop as to not lock up ESP.
Next would be try/catch handling if buffer is still locked after timeout (same behaviour as out of memory condition) with more graceful outcome.

blazoncek:
Config and ledmap are safe. I can see problems on serial, MQTT or network access.

blazoncek:
I have identified potential lockup in:
- onAlexaChange() : applyPreset() attempts to lock buffer
- handle buttons : same as above
- presetFallback() in IR : same as above
- handleNightlight() : same as above
- time macros : same as above
- handlePlaylist() : same as above
- handleSet() if fileDoc not assigned
- applyMacro() but this looks like an unused function

Oddly though, ESP8266 never locked up.

blazoncek:
Got it!
```
Strip time[ms]: 0/1
IR recv: 0xFF30CF
IR JSON buffer requested.
API req: win&PL=2
Apply preset JSON buffer requested.
```
... and there was a lock-up.

blazoncek:
Recent commits are without locking up ESP. Looks promising.
I will update every locking call with a cleaner code though.

blazoncek:
A lot less reboots on ESP8266 since latest push. 😀

wled00/json.cpp
blazoncek: A few bugs found and squashed. No more lockups.

blazoncek:
Upadte: ESP8266 (and ESP32) running smooth for 2 days now. No issues whatsoever.

wled00/ir.cpp
```diff
@@ -654,9 +661,7 @@ void handleIR()
       {
         if (results.value != 0) // only print results if anything is received ( != 0 )
         {
-          Serial.print("IR recv\r\n0x");
-          Serial.println((uint32_t)results.value, HEX);
-          Serial.println();
+          DEBUG_PRINTF("IR recv: 0x%lX\n", (unsigned long)results.value);
```
Aircoookie, line 664: This should probably stay `Serial`, so normal users can identify key codes of custom remotes to make an ir.json config, without having to build a debug version.
Should check for TX pin (GPIO1) allocation though, otherwise it could mess with LEDs or other HW connected on the pin.

wled00/util.cpp
```diff
@@ -0,0 +1,33 @@
+#include "wled.h"
+#include "fcn_declare.h"
+#include "const.h"
+
+
+bool requestJSONBufferLock(uint8_t module)
+{
+  unsigned long now = millis();
+
+  while (jsonBufferLock && millis()-now < 1000) delay(1); // wait for a second for buffer lock
```
Aircoookie, line 10: I thought to myself that this can't possibly work, because `delay()` in ESP8266 network callbacks will crash and burn.

Turns out that this _does_ in fact work, because ESP8266 and 32 differ in behavior:
ESP8266 would be stuck for a second and not reply to the request in the best case and crash in the worst case if it encountered a locked buffer in a network callback. Luckily this _should_ never happen as the ESP8266 can only handle network callbacks wherever `yield()` is placed. As such, as long as we make sure we never call `yield()` while the buffer is locked, we are 100% fine. In fact, buffer lock logic is _not required at all_ on ESP8266 since there is simply no way for two interactions that both need the buffer to intersect.

On ESP32, the buffer lock _is_ needed and works as intended! Due to being dual core, network callbacks are run on core 1 while the main loop is run on core 0. If core 1 handles a network request, it will wait out the lock, as (unlike with 8266) the loop task will continue to run, thus only blocking for a few milliseconds at most.

I'm interested in how ESP32-S2 would behave here, given that it is single core. I would assume same as ESP8266, but it might also be running two tasks on the same core (and rapidly switching between them).
In both cases, it _should_ work fine.

I think this is ready for merge and public testing as part of 0.13.0-b6 :)

wled00/ir.cpp
blazoncek: Exactly, as it happened recently on some other issue posted.
It may be better to add this info somewhere else (info page i.e. last IR received).

wled00/util.cpp
blazoncek: Never thought about implications of net callbacks since I was unaware of the limitations.
Still, this code is now running bug free for a month on a ESP01, D1 mini and D1 min32.
I went so far as to install it on all of my production units. Not a single one of them has crashed since!

PR #2343, opened 2021-11-17
Battery level via analog input

Stephan1908:
Hi,

I built a mobile lamp from an old power bank.
In the case of a lamp with a rechargeable battery (3.7V), is it possible to read in the voltage of the rechargeable battery using an analog input? And then get this value to Home Assistant?

Thanks and Greetings
Stephan


Aircoookie:
Hey!
I don't think you meant to make a PR :)
This is possible with the https://github.com/Aircoookie/WLED/tree/master/usermods/battery_status_basic usermod, for HA I believe you would need to add a custom sensor, it is not part of the official integration.

Dutchy-79:
Hi, I'm also looking for monitoring voltage and just don't get how to make my own voltage measurement user mod.

I'm running on a 12v car battery and now I have a spare wemos with 2 resistors connected to A0 ground and battery voltage.

Though I want to reduce power consumption and would like to use the wled wemos A0 for the same task and prefer to also send the data to home assistant or else mqtt.

It's not just for battery use, you can also monitor your power supply with the same method.

So hopefully it will be a real built-in feature one day

PR #2360, opened 2021-11-26
Ir code formatting, and IP address, on Serial

WJCFerguson:
Even though I cut/paste my codes from the serial output, my JSON failed to work because it wasn't the right format.  It would be great to make the whole lookup less fragile but in the meantime this easy fix to print it in upper-case hex in the right format might help avoid problems.

There may be some reason why printing the IP on serial is a bad idea, but it seems like a good one to me (?).  Perhaps I should put this in two PRs, as they're very independent.

### DIFF
--- a/wled00/ir.cpp
+++ b/wled00/ir.cpp
@@ -654,9 +654,7 @@ void handleIR()
       {
         if (results.value != 0) // only print results if anything is received ( != 0 )
         {
-          Serial.print("IR recv\r\n0x");
-          Serial.println((uint32_t)results.value, HEX);
-          Serial.println();
+          Serial.printf("IR recv: 0x%X\n", (uint32_t)results.value);
         }
         decodeIR(results.value);
         irrecv->resume();
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -610,6 +610,9 @@ void WLED::initInterfaces()
 {
   DEBUG_PRINTLN(F("Init STA interfaces"));
 
+  Serial.print("IP address: ");
+  Serial.println(Network.localIP());
+
 #ifndef WLED_DISABLE_HUESYNC
   IPAddress ipAddress = Network.localIP();
   if (hueIP[0] == 0) {


wled00/wled.cpp
```diff
@@ -610,6 +610,9 @@ void WLED::initInterfaces()
 {
   DEBUG_PRINTLN(F("Init STA interfaces"));
 
+  Serial.print("IP address: ");
```
blazoncek, line 613: IP address is printed out as part of the debug output when connection is established.
No need to print it separately.

wled00/ir.cpp
```diff
@@ -654,9 +654,7 @@ void handleIR()
       {
         if (results.value != 0) // only print results if anything is received ( != 0 )
         {
-          Serial.print("IR recv\r\n0x");
-          Serial.println((uint32_t)results.value, HEX);
-          Serial.println();
+          Serial.printf("IR recv: 0x%X\n", (uint32_t)results.value);
```
blazoncek, line 657: printf uses precious RAM. (As unfortunately does original code.)
Please use F() macro to utilize flash memory.

blazoncek (changes requested):
Displaying IR codes to serial, without properly allocating serial output, is considered a bug.
Using debug build allocates GPIO1 used for serial output.
Since this code is quite legacy it would be good to bring it up to speed with recent development.

wled00/wled.cpp
WJCFerguson: How can you see this?  If I connect with `picocom -b 115200` but that gives corrupted initial output, but that is the correct baud for later Serial output.  What settings get the initial debug output?  And can they be made the same as the later serial output so you can see both?

wled00/ir.cpp
WJCFerguson: Ah, will change to `F()` macro.

Looking at the code it looks to me as long as the message is <64 bytes, `print` and `printf` use roughly the same amount of stack and no dynamic allocation.  (print allocates enough to encode a long as binary...  `8x8 + 1`.)

wled00/wled.cpp
WJCFerguson: (I'm using an ESP32 D1 Mini.  I see suggestions the initial baud depends on the crystal clock rate or something...?)

WJCFerguson:
I'm going to close this since it seems like the IR code output format was possibly already fixed?  I'm now seeing it in a compatible format and can't explain why I was getting it differently before.

I'd love to know how to get the IP address, but I haven't succeeded so far.

> Displaying IR codes to serial, without properly allocating serial output, is considered a bug.

Not sure what you mean by properly allocating serial output?  Since the IR code always outputs I assumed it was fine to just write to Serial without checking if it's connected.  Perhaps not?

> Using debug build allocates GPIO1 used for serial output.

I suggested these for the benefit of people who don't or can't build the firmware - in the case of the IR codes because I had to build firmware to find out why it wasn't working.

> Since this code is quite legacy it would be good to bring it up to speed with recent development.

ir.cpp?  Seems pretty rough.  I am merely a drive-by submitter, sorry I'm not deeply enough into it or familiar enough with the project's recent development to know how to refactor it, and nor do I have the time.

PR #2374, opened 2021-12-01
added DMXinfo to JSON information

max9901:
Hey guys, 

First of all i Love Wled! amazing software!

Right now im using it in a semi profesional dmx setup and i find me self needing some more details in the json settings overview about the dmx settings to sanity check correct settings. 

Therefore the following PR.

### DIFF
--- a/wled00/json.cpp
+++ b/wled00/json.cpp
@@ -511,6 +511,20 @@ void serializeInfo(JsonObject root)
   root[F("udpport")] = udpPort;
   root["live"] = (bool)realtimeMode;
 
+  JsonObject dmx_info = root.createNestedObject("DMX");
+  switch (DMXMode) {
+    case DMX_MODE_DISABLED     : dmx_info["mode"] = "DMX_MODE_DISABLED"; break;
+    case DMX_MODE_SINGLE_RGB   : dmx_info["mode"] = "DMX_MODE_SINGLE_RGB"; break;
+    case DMX_MODE_SINGLE_DRGB  : dmx_info["mode"] = "DMX_MODE_SINGLE_DRGB"; break;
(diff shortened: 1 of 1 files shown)


Aircoookie:
Hey!
I'm happy that you like WLED so much and thank you for the PR!
Since these values are configuration, you can already get them using `/json/cfg`:
Mode: `live.dmx.mode`
Address: `live.dmx.addr`
Port: `live.port`
Mulicast: `live.mc`

Not sure if it would make sense to add them to info as well.

max9901:
So sorry after some extra searching in the code base i found indeed this option:
`/cfg.json`

`json/cfg` is not working btw.

also this `cfg.json` is not a subset of `json` which i my opinion would be nice to be a subset

Will close this issue!

I am working at a multisegment dmx mode where use the effect dmx mode per segement. And have overlapping segments which can be disabled to have full controll, i call this mode `DMX_MODE_SEGMENT_EFFECT`

https://github.com/max9901/WLED/blob/feature/multi-dmx/wled00/e131.cpp

To make the overlapping segments working i added a `hidden` flag to the segment structure. which will hide this segment from displaying. So that WLED can deal with overlapping segments.  For this hidden flag to take effect i created an extra callmode that only works with the  `DMX_MODE_SEGMENT_EFFECT` so that normal operation is unchanged for WLED

If u find the time i would love to hear your opinion. 

I created this because I have 900 rgb leds in a single WLED system and want to be able to program it with professional DMXcontrollers. This gives me all the freedom I need without the need to send massive amounts of sACN data to control all leds individually.

Aircoookie:
Indeed it would be great to have cfg as part of the root `/json` object, unfortunately that is not easily possible as it would overflow the JSON sending buffer size.

Ah sorry, `/json/cfg` is only available in 0.13 (betas), not yet in 0.12. It simply returns `/cfg.json` though, so ok to use that, too.

PR #2393, opened 2021-12-07
Improved speed and reduced memory usage for Blends effect

guardmedia:
color_blend() function only generates 256 colors from the palette before repeating. Improved the function to use either 256 max or the segment length if shorter.


Aircoookie:
Verified this is the case, thank you for the improvement!

PR #2414, opened 2021-12-14
Add support for SR extension for effect names

blazoncek:
(without breaking compatibility)

Also adds function that usermods may use to get Effect or Palette name.
SR extensions add ability to control which sliders are displayed (some are irrelevant on certain effects) and possibility to replace slider description, if color slots are displayed and if palettes are displayed.

The control commands follow the @ sign which follows effect name. The controls are implemented entirely in UI (not part of this PR).


blazoncek:
This could allow to store default values for effect into extension as well as controlling sliders and colors.

PR #2415, opened 2021-12-14
Pin manager support for sharing multipin buses.

blazoncek:
Currently implemented I2C but can be enhanced for SPI as well in the future by adding special owner.


Aircoookie:
Looks good to me 👍

Aircoookie (approved):

Issue #2417, opened 2021-12-14
DDP receiving strip effect spontaneous effect problems (animation lagged and choppy)


blark:
### What happened?

I am using two ESP32s with 0.13.0-b6. My wifi connection is good, I am using external antennas and have good RSSI.

Strip 1 has ~550 LEDs
Strip 2 is connected via DDP and has ~200 LEDs

Effects run fine when I first select them, but then after some time (this is not consistent) they become choppy and lagged on the second DDP strip. Stopping the animation and restarting it has no effect, the 2nd controller must be rebooted -- then it immediately begins working fine again. Eventually the effect will degrade again and exhibit the same choppiness.

### To Reproduce Bug

Set up a similar setup to the one described above, and wait some time while an effect is active. 

I noticed that when I scroll through the list of effects and change them somewhat quickly (to see what they look like) it seems to speed up the onset of this behaviour. 

### Expected Behavior

Effects should work properly on both strips. 

### Install Method

Binary from WLED.me

### What version of WLED?

WLED 0.13.0-b6

### Which microcontroller/board are you seeing the problem on?

ESP32

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

blazoncek:
What is the time-frame for choppiness?
I am running this on Ethernet enabled ESP32 with 1000 client (network) pixels on 3 ESPs (ESP01, Wemos D1 mini, ESP32) and not seeing any choppiness or stutter or other issues.

guardmedia:
@blark, I had the same issue when using WiFi, and ultimately resolved it by switching to Ethernet connected instead.

Like you, I know my WiFi is good (enterprise equipment, 10g lan, isolated RF, etc..), but couldn't figure out why there were occasional glitches or freezing. Eventually I setup Wireshark, and confirmed occasional delays between packets which aligned to the display issues seen. Tracing this back to the access point, I confirmed the delay was actually coming from the esp32. No connection or packet errors - it just wasn't sending days every once and a while.

After enabling debug in WLED, I noticed errors thrown by WifiUDP code in Arduino-ESP32. Although it recovers, every time the error is printed this correlated to packet loss (omission?) from esp32. After hours of searching through GitHub, setting up debug output in IDF layers, and testing multiple esp32 boards, I eventually gave up and switches to Ethernet. It has been butter smooth since then, which leads me to believe there's something in the underlying ESP-IDF code.

[...]

PR #2425, opened 2021-12-16
Add UDP rate limiter for DDP over WiFi

guardmedia:
Proposed resolution for #2417 -- tested with 12 ms rate limit with up to 4096 pixels.

### DIFF
--- a/wled00/udp.cpp
+++ b/wled00/udp.cpp
@@ -561,6 +561,8 @@ void sendSysInfoUDP()
 // 1440 channels per packet
 #define DDP_CHANNELS_PER_PACKET 1440 // 480 leds
 
+#define DDP_MIN_PACKET_DELAY 12 // minimum time (ms) between each UDP packet
+
 //
 // Send real time UDP updates to the specified client
 //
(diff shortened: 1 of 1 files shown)


blazoncek:
What will happen if I have 10 buses with relatively small number of LEDs (i.e. 150).
I would love to have all of them synced.

According to the code this would mean about 3.75ms allowed per packet so 37.5ms for all packets (~27FPS). Am I correct?

blazoncek:
Can you check if #2444 could be a solution for this as well?

guardmedia:
> Can you check if #2444 could be a solution for this as well?

Yes, that works! Tested with 4096 pixels limited to 10 FPS and observed zero WiFiUDP errors. Even managed to push to 12 FPS with only a few errors (~1-2 every 5 minutes).

It might be good to include a tooltip to inform users about recommended breakpoints to avoid UDP buffer overflow when using DDP though. The numbers below assume only 1 bus (e.g. your example above with 10 busses with 150 LEDs yields a different number).

> 480 pixels = 42 FPS
> 960 pixels =~ 42 FPS
> 1440 pixels =~ 28 FPS
> 2880 pixels =~ 14 FPS
> 4096 pixels =~ 10 FPS

Closing this PR in favor of #2444 instead!

PR #2429, opened 2021-12-18
Usermod quinled-an-penta: Updated IOs for v1r1 release

ezcGman:
The upcoming v1r1 release has an IO change, so I updated the default config, plus added some info to the ReadMe.

Note: I have no clue why it also lists the commits which are already merged in the WLED master... My master was rebased with Aircoookie/WLED:master and this branch was rebased with my master... So it should kwow all the commits...


blazoncek:
Why don't you condense all those commits into one and then make PR?
It is much cleaner that way.

https://github.com/Aircoookie/WLED/wiki/How-to-properly-submit-a-PR

ezcGman:
> Why don't you condense all those commits into one and then make PR? It is much cleaner that way.
> 
> https://github.com/Aircoookie/WLED/wiki/How-to-properly-submit-a-PR

Can I on creating the MR? I only know you can when actually merging it, then it offers to "squash commits".

If I can somehow, I do. As said, I have no clue why it lists all the commits; these are already in, it should only be one commit...

Aircoookie:
Strange, I think this can happen if you don't pull master into your working branch before making the second PR, seen it before though.

Squash and merge will take care of it.

ezcGman:
Think I know what happened: https://github.com/Aircoookie/WLED/commits/master/usermods/quinled-an-penta
When my first commit for that UM was merged, the commits were also squashed. Git then looses the commit hashes of these squashed commits ;)

PR #2440, opened 2021-12-23
Adding support for WAN networks in LED syncing

supersimple33:
So as of present, different WLED modules will not find each other when broadcasting to all network connections. This is due to how WAN emulates LAN networks and only certain devices can be found during a search. With this PR I have reworked some functions to allow them to target specific IP addresses. This reworking allowed me to add new settings to sync settings being 10 different IPs which can be targeted across the WAN to allow WLEDs to discover each other if the IP of each device is known. If you have any problems with my code or would prefer me to rework my implementation please let me know below and I will address these changes.


blazoncek:
Please resubmit using instructions provided [here](https://github.com/Aircoookie/WLED/wiki/How-to-properly-submit-a-PR).
It is difficult to track changes otherwise.

PR #2444, opened 2021-12-25
Experimental configurable framerate support

Aircoookie:
Updated arduino core versions
Better performance on esp32 core 3.x due to IRAM_ATTR
Fixed analog busses init to full white/on


blazoncek:
Tested it, works ok. I did change `setTargetFps()` to use min(max()) as proposed and discovered that when upgrading my FPS went to 1 instead of 42.
So I added this to cfg.cpp: `strip.setTargetFps(hw_led["fps"].isNull() ? WLED_FPS : hw_led["fps"].as<byte>());`
Changing it in settings worked and brought WLED back to working order (some effects do not play well with 1 FPS).

PR #2448, opened 2021-12-26
Adding support for WAN networks in LED syncing

supersimple33:
So as of present, different WLED modules will not find each other when broadcasting to all network connections. This is due to how WAN emulates LAN networks and only certain devices can be found during a search. With this PR I have reworked some functions to allow them to target specific IP addresses. This reworking allowed me to add new settings to sync settings being 10 different IPs which can be targeted across the WAN to allow WLEDs to discover each other if the IP of each device is known. If you have any problems with my code or would prefer me to rework my implementation please let me know below and I will address these changes.

### DIFF
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -278,6 +278,14 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
   JsonObject if_nodes = interfaces["nodes"];
   CJSON(nodeListEnabled, if_nodes[F("list")]);
   CJSON(nodeBroadcastEnabled, if_nodes[F("bcast")]);
+  JsonArray if_node_list = if_nodes["cusnodes"]; // not sure if I should be using F here, couldn't understand what it does
+  for (uint8_t i = 0; i < 10; i++) { // 10 is arbitrary defined in settings
+    JsonArray if_node_ip = if_node_list[i];
+    CJSON(specialSearchNodes[i][0], if_node_ip[0]);
+    CJSON(specialSearchNodes[i][1], if_node_ip[1]);
+    CJSON(specialSearchNodes[i][2], if_node_ip[2]);
+    CJSON(specialSearchNodes[i][3], if_node_ip[3]);
+  }
 
   JsonObject if_live = interfaces["live"];
   CJSON(receiveDirect, if_live["en"]);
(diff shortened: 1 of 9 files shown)


supersimple33:
I have updated this PR to be compliant with the instructions provided. I also fixed the failed checks, it turned out I was accidentally calling a method that doesn't exist.

supersimple33:
Latest push resolved new conflicts.

blazoncek:
I am sure this makes sense for your configuration but generally I would recommend against exposing WLED devices to WAN.
If by WAN you refer to multiple VLANs where your WLEDs are, then this is legitimate proposal. Still, limiting the number to 10 may not satisfy everyone (including me). In such case I would propose enabling UDP broadcast forwarding to VLANs on your router (you may need an enterprise grade router which I assume you have in case of VLANs).

For start I dug two articles touching this area:
https://techhub.hpe.com/eginfolib/networking/docs/switches/WB/15-18/5998-8166_wb_2920_mrg/content/ch03s10.html
https://community.ui.com/questions/Multicast-Sonos-Phorus-and-Play-Fi-Broadcast-255-255-255-255-lessportgreater-Discovery-Solution/1ce890c2-5e7e-4ef2-a42a-e9c59444fd3f#answer/fae150b3-ad70-4e9a-9b8b-9f20777e46e7

supersimple33:
So I have multiple WLEDs across VLANS but I cannot forward broadcasts for separate reasons, thus what if I reformed the commit? 

My whole objective with this PR is to help other users stuck on VLANs get their lights to sync. The lights won't sync with each other across different VLANs so I think there should be an option within WLED to expose the IPs of other lights. Obviously, it wouldn't make sense to expose the lights to the entire WAN so I would add some way to add specific IPs to search for on the Instance List. I would then store said extra IPs in some form of LinkedList in order to accommodate any number of extra lights. 

I think this might even be helpful for regular LAN users as they would no longer be obligated to have all sync notifications sent to every device on the network via the broadcast IP. This would allow different WLED owners and systems to operate independently of each other which while already accomplishable in one way through light groups would grant users greater privacy and control over their system. The only thing I believe that might be bad is that the chips would be absorbing slightly more work that was previously being done by the router but more performance-driven users could avoid by simply using the traditional methods in place right now. 

Let me know if you think this would be a good idea. I am happy with my current implementation as it works well enough for my use case but if you think this would be helpful to others I would work on it.

blazoncek:
Hi @supersimple33 ,
Sorry to get back to you so late, I hope you haven't given up on this PR yet. 😄 

Could you update this PR to be compatible with 0.14? I'll see what we can do with it then. 👍

netmindz:
This should really be a usermod rather than core, trying to sync devices that are on different networks feels very much an edge case. Cluttering the UI for other users with entry for 10 ips.

blazoncek:
The idea of usermod for this is great. Unfortunately there are no hooks or callbacks for notifications in usermods.

The closest thing is recent addition of `onStateUpdate()` hook. But that will not cover node broadcasts.

Do you have an idea how to implement this?

blazoncek:
@netmindz & @supersimple33 I have checked the relevant code and the proposed move to usermod is viable.

The only change to core files would be to modify `sendSysInfoUDP()` as proposed in this PR and then write a usermod that does broadcasting to clients from within `loop()` with similar frequency as original broadcasts.

Are you willing to implement it in this way? It has a much greater chance of being merged that way.

supersimple33:
Yeah I'd agree that a user mod is a better way of implementing this since my current implementation would needlessly complicate the UX of the average user.

blazoncek:
Hi. Is there still interest to get this PR (in the form of usermod) into WLED? If yes, please update or close if there is no interest.

PR #2456, opened 2021-12-27
Added support for LPD6803

### LINKED ISSUE #1500: Is it possible 6803 IC Chip LED to operate? #851
> 
> 
> Hi,
> it could be that the LPD6803 uses either the WS2801, LPD8806, P9813 or APA102 protocols.
>...

The LPD6803 does not work using any of those types, however its possible to make WLED work with those kind of LED strips.
I understand that the low level communication is provided by NeoPixelBus. 
Based on some already supported LED types by NeoPixelBus, I created two new files for the LPD6803 type:
Lpd6803ColorFeatures.h
Lpd6803GenericMethod.h

I have a 10 meters (2 x 5m) LED string working with WLED 0.11.0 and a Wemos D1 mini (ESP8266), for almost 10 days.
It works with GRB and BRG types, even when mixed (thanks to new enable color order override for a specific range of the strip,  on v0.11.0)

Already sent the files to Makuna. In case he decides to integrate them on NeoPixelsBus, will be easy add support for the LPD6803 to WLED, by doing minimal changes on NpbWrapper.h (possibly on the BusWrapper_h file on the dev version too).

Sousanator:
You can now select LPD6803 IC chip from the LED selection menu and use it.

Resolves #1500


blazoncek:
@Aircoookie I don't see a reason why we shouldn't merge this one in.
Approved.

PR #2463, opened 2021-12-31
Allow overriding of color order by LED pixel range.

### LINKED ISSUE #2462: Manage color ordering by LED offset, rather than GPIO.
Yes. 
I have 1200 LEDs spread over 5 GPIOs. On two of the GPIOs, I have a mix of LED strings with RGB and BRG color ordering.  In both cases, I want them to apply color and effects to them as a single segment, but the colors are always wrong on one of the strings because the color order config is tied to the GPIO. 

I want to be able to set LED color order based on LED offset, rather than by GPIO, and I want to do this multiple times throughout my 1200 LEDs. This is similar to what can be done in xLights, where you can specify color ordering on a per-model (string) basis. 

- Remove the `Color Order` option from `LED outputs`
- Add a new section to `LED Settings` for specifying color ordering. 
  - Drop-down for choosing the default color order (ex: RGB)
  - Similar to segment management, allow addition of color order overrides for LED ranges (start, length, color order). 

* Moving one of the strings to a separate GPIO is not always possible. In my case, there is 30 feet between the esp32 and the first string. If I moved the second string to a separate GPIO, it'd mean I'd have to run a 40 foot cable, which would be pushing it for range, as well as wasteful.

* Per-segment color ordering (#1428) is not viable as I want the two strings on a GPIO with different color ordering to operate as a single segment, not separately.
* Additional reasons why per-segment color ordering is no good:
  * It feels like the wrong place. Segments are for effects/presets, not for lower-level color order mapping. Some presets might have 10 segments, while others will have just one.
  * Managing segments across multiple presets can be tedious.  

* Using the [`COLOR_ORDER_OVERRIDE`](https://github.com/Aircoookie/WLED/blob/17d2fb80f2c053ff923187d4ab57898068511ade/wled00/bus_wrapper.h#L386-L389) define is not viable, as I want to override the color order in multiple places.
  * Color order can only be overridden for a single range (between `COO_MIN` and `COO_MAX`).

**Additional context**

Current approach I'm using (and unhappy with) requires me to define separate segments, and manually remapping the color palette from RGB to BRG for the strings of LEDs that have the different color ordering. This is very frustrating as any changes to the palette requires me to remap colors on the individual segments, which is very tedious. 

I went so far as to implement per-segment color ordering, but quickly found that it was not in the right direction. Segment configurations do not survive power cycles, unless they've been saved to a preset. And, again, managing segments across many presets is tedious.

justfalter:
Adds a means for overriding color order for ranges of LED pixels. Default color order is still managed at the GPIO level.  This is meant to fix #2462. Up to 10 color order overrides defined.

- [x] Use `ColorOrderMap` to hold color order information.
- [x] Serialization of config to/from filesystem.
- [x] Modify `handleSettingsSet` to take a configuration
- [x] Modify `xml.cpp` to serialize settings for front-end
- [x] Modify LED settings page to allow for configuration of color order map.

### Performance: 

I figured watching the FPS would be the best way to know if my changes negatively impacted performance. Hopefully the reported FPS is accurate. 

My methodology:
- 1 segment for all 1191 LEDs
- Disabled network discovery.
- Effect: Pride 2015, Speed 255, Effect 255
- Measure FPS 10 times over 10 seconds:
```
#!/bin/bash
num_samples=10
total=0
for i in $(seq 1 ${num_samples}); do
  sleep 1
  fps=$(curl -s "${WLED_BASE_URL}/json/info"|jq '.leds.fps')
  total=$(($total + $fps))
  echo "$i: ${fps}"
done

avg=$(echo "${total}/${num_samples}" | bc -l)
echo "Average: ${avg}"
```

Results:
| Env | FPS | 
|-|-|
| master (736053e24e4933148bd28b6fb2d3720005de9c1e) | 40.4 |
| PR - 0 overrides | 40.4 |  
| PR - 1 overrides | 40.9 |  
| PR - 2 overrides |  40.8 |
| PR - 3 overrides |  39.7 |
| PR - 4 overrides |  39.2 |
| PR - 5 overrides |  40.3 |
| PR - 6 overrides |  40.9 |
| PR - 7 overrides |  40.7 |
| PR - 8 overrides |  39 |
| PR - 9 overrides |  38.9 |
| PR - 10 overrides | 39 |

It appears that performance is comparable between master and 0 overrides. 10 overrides appears to be 3.5% off of the base case.

I ran this several times and, while there was some variability, I'm not really sure why things would vary *up* and *down* like they had. Perhaps it has to do with how FPS is calculated, how the busses are updated, or maybe there are other timing/interrupts occurring.  I thought maybe it was an alignment issue with the `ColorOrderMapEntry` struct, manually padding made no difference. The compiler should take care of aligning things, anyway.  Regardless, performance seems pretty close to what is expected.

### Config

Color order overrides are managed on the LED Settings page:
image

The above configuration appears within cfg json as:
```
{
  "hw": { "com":[{"start":253,"len":90,"order":2},{"start":696,"len":90,"order":2}] }
}
```


justfalter:
The following will add a GPIO with 200 LEDs on it, defaulting to BRG color order, and overriding the color order for pixels 100-200 to be RGB.

```
curl 'http://192.168.55.58/settings/leds?' \
  -H 'Connection: close' \
  -H 'Origin: http://wled-testy.local' \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -H 'Referer: http://wled-testy.local/settings/leds?' \
  --data-raw 'MA=850&LAsel=50&LA=0&LT0=22&CO0=2&LC0=200&L00=18&L10=&L20=&L30=&L40=&BT0=0&BE0=2&BT1=-1&BE1=0&BT2=-1&BE2=0&BT3=-1&BE3=0&TT=32&IR=-1&IT=0&data=&RL=12&RM=on&BO=on&CA=128&BP=0&GC=on&BF=100&TF=on&TD=700&TL=60&TB=0&TW=1&AW=3&CB=0&PB=0&data2=&XS0=100&XC0=100&XO0=1' \
  --insecure
```

justfalter:
@Aircoookie any thoughts on this?

blazoncek:
It has been considered but is waiting for peer review. Please wait.

wled00/bus_manager.h
```diff
@@ -73,6 +73,62 @@ struct BusConfig {
...
+  }
+
+  const ColorOrderMapEntry* get(uint8_t n) const {
+    if (n > _count) {
+      return nullptr;
+    }
+    return &(_mappings[n]);
+  }
+
+  inline uint8_t getPixelColorOrder(uint16_t pix, uint8_t defaultColorOrder) const {
```
blazoncek, line 116: I am fond of inline functions but I don't see the benefit here.

wled00/const.h
```diff
@@ -39,6 +39,8 @@
   #endif
 #endif
 
+#define WLED_MAX_COLOR_ORDER_MAPPINGS 10
```
blazoncek, line 42: I would add different default value (5) for ESP8266 to reduce memory pressure and CPU cycles.

wled00/xml.cpp
```diff
@@ -421,6 +421,17 @@ void getSettingsJS(byte subPage, char* dest)
       oappend(SET_F("mA\";"));
     }
 
+    oappend(SET_F("resetCOM();"));
```
blazoncek, line 424: Add a value for maximum number of maps.

wled00/data/settings_leds.htm
```diff
@@ -347,6 +347,50 @@
...
+
+    function remCOM() {
+      var entries = d.getElementsByClassName("com_entry");
+      var i = entries.length;
+      if (i === 0) return;
+      entries[i-1].remove();
+      btnCOM(i-1);
+    }
+
+    function resetCOM() {
```
blazoncek, line 382: Use this function to get maximum number of maps from ESP (see below).

wled00/data/settings_leds.htm
```diff
@@ -347,6 +347,50 @@
...
+      for (let e of d.getElementsByClassName("com_entry")) {
+        e.remove();
+      }
+      btnCOM(0);
+    }
+
+    function btnCOM(i) {
+      gId("com_add").style.display = (i<10) ? "inline":"none";
```
blazoncek, line 390: Use variable maximum number of maps.

wled00/set.cpp
```diff
@@ -136,6 +136,18 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
       doInitBusses = true;
     }
 
+    ColorOrderMap com = {};
+    for (uint8_t s = 0; s < WLED_MAX_COLOR_ORDER_MAPPINGS; s++) {
```
blazoncek, line 140: What if there are entries missing? It would be better to break instead of adding 0s.

blazoncek (approved):
Code reviewed (but not yet tested) and gets a go from my side (with minor adjustments suggested).

wled00/bus_manager.h
justfalter: My thinking was that, given that `getPixelColorOrder()`  is called each time a pixel's color is set, inlining `getPixelColorOrder()` would improve performance by eliminating the overhead of having to call and return from the function.

wled00/set.cpp
justfalter: Fair point, thank you!

wled00/const.h
justfalter: Done.

blazoncek (approved):
I hope to have time to pull this into my fork and test this weekend.
Otherwise it looks cool and up to current standards in WLED.

blazoncek:
Works wonderfully!
@Aircoookie I recommend we merge this.

PR #2465, opened 2022-01-02
Add New effects and functions

Proto-molecule:
Hey id like to share some additions and changes I’ve been working on. Please review them and add them if you like. I did my best do add everything correctly however I’m sure I made some mistakes.

- New firework style effect. It utilized the following two new functions.

- New color addition functions that preserves the sums ratio even if the color value goes above 255. 
	e.g. (R255 G255 B0)+(R0 G255 B0)=(R255 G255 B0), and 
	(R255 G128 B0)+(R255 G128 B0)=(R255 G128 B0)
	There may have already been something to do this but I could fid it.

- New anti aliasing pixel drawing function. It can draw floating point positions of a pixel between two LEDs.

- Changed the bouncing balls effect a little to use these two new functions. To really see them in action, slow the effect all the way down and turn on three colors. The balls will track smoothly from pixel to pixel and the colors will blend together. 

- New criss-cross bouncing ball effect (its not the same as mirroring the segment).

- New Random Float number generator
       I needed random float numbers for the new effect however the way its added may be incorrect. It seems to work but if
       you have any suggestion on a better way than please.


Aircoookie:
I really like the antialiasing functionality and the new effects!

Maybe I will slighly rewrite it to rely less on floating point operations for performance reasons, but this will be merged soon :)

blazoncek:
Totally agree.

Proto-molecule:
Great!

blazoncek:
Please do not force push commits once PR is open.

Proto-molecule:
My bad, accidentally undid the commit on my fork and it closed it here.

Proto-molecule:
eff, I'm sorry again. I should probably set up a dummy repo to run my git testing.

bole5:
@Aircoookie @blazoncek 
It would be great if this PR could be merged. More effects could later be changed to use the new antialiasing rendering, which is IMO looks much nicer. Not sure how much you can reduce usage of floating points since antialiasing depends on it.

wled00/FX_fcn.cpp
```diff
@@ -860,6 +860,73 @@ void WS2812FX::fade_out(uint8_t rate) {
...
+     return RGBW32(r * ratio, g * ratio, b * ratio, w * ratio);
+   }
+  }
+
+/*
+*  draw moving pixles with anti aliasing
+*  Inspired by Daves Garage Ep. 9 
+*  https://www.youtube.com/watch?v=RF7GekbNmjU&list=PLF2KJ6Gy3cZ7ynsp8s4tnqEFmY15CKhmH&index=7
+*/
+void WS2812FX::drawAntiAliasing(float fPos, float pixelSize, uint32_t color){
```
blazoncek, line 896: I would recommend overloading setPixelColor() instead, using float for index and only do 1 pixel AA.

wled00/FX_fcn.cpp
```diff
@@ -860,6 +860,73 @@ void WS2812FX::fade_out(uint8_t rate) {
...
+      max = g;
+   } else if (b>w){
+      max = b;
+   }else{
+      max = w;
+   }
+   if(max < 256){
+     return RGBW32(r, g, b, w);
+   }else{
+     float ratio = 255.0f / max;
```
blazoncek, line 886: Could do without using float.

blazoncek (commented):
The idea (for AA and color adding function) presented is nice but I would recommend using different approach and not bundle it with new effects (even though they may require it).

wled00/FX.cpp
```diff
@@ -2639,9 +2658,11 @@ uint16_t WS2812FX::mode_bouncing_balls(void) {
   
   bool hasCol2 = SEGCOLOR(2);
   fill(hasCol2 ? BLACK : SEGCOLOR(1));
+  //fill(hasCol2 ? BLACK : SEGCOLOR(1));
+  // for (int i = 0; i < SEGLEN; i++) {if (random(10) > 5)blendPixelColor(i, BLACK, 100); }  // perhaps add a comet tail option in the future?
   
   for (uint8_t i = 0; i < numBalls; i++) {
-    float timeSinceLastBounce = (time - balls[i].lastBounceTime)/((255-SEGMENT.speed)*8/256 +1);
+    float timeSinceLastBounce = (time - balls[i].lastBounceTime)/((255-SEGMENT.speed)*15/256 +1);
```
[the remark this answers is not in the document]
Proto-molecule: I think it was just personal preference. If I remember (months ago) I found the speed of the old effect too fast.
Not really an important change. 
Also, maybe because my original post suggested slowing the effect all the way down to demo the Anti Aliasing.
Who knows

wled00/FX.cpp
```diff
@@ -4207,5 +4232,153 @@ uint16_t WS2812FX::mode_aurora(void) {
...
+  float          speedVariable           = SEGLEN >> 2;                         // Try to keep the explosions prepositional to SEGLEN
+  float          particleIgnition        = 250.0f;                              // How long to "flash"
+  float          fadeTimeVariable        = 1000.0f;                             // Fade out time
+  float          fragSizeVariable        = .5f;                                 // Particle size grows with explosion size
+  float          randval = (0.002f / numStars) * ((SEGMENT.speed >> 2) + 1);    // SEGLEN affects numStars affects "time in loop" affects probability of new firework spawn.
+  float          intenseVal = ((SEGMENT.intensity/255.0f) * 3) + 1.1f;
+
+
+
+  if((millis() - SEGENV.aux0) > FRAMETIME)                                      // frame rate needed to keep spawns and tail fades consistant
```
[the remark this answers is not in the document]
Proto-molecule: ok, I didn't realize that at the time. Thank you!

wled00/FX.cpp
```diff
@@ -4207,5 +4232,153 @@ uint16_t WS2812FX::mode_aurora(void) {
     setPixelColor(i, mixedRgb[0], mixedRgb[1], mixedRgb[2]);
   }
   
+  return FRAMETIME;
+}
+
+/*
+* Fireworks Woah effect
```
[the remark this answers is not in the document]
Proto-molecule: Pretty much. But does it really look the same?  If you look at the notes for the "Fireworks Starburst" code, they credit the idea to a reddit post. The creator of the video in that post is Dave Plumber. The same person I linked for the "Anti Aliasing" code.
AA is the cornerstone of this effect. Take away AA then it begins to look a lot like Starburst (minus the unsymmetrical particle distribution). Also something is off about Starburst's random spawns.

wled00/FX_fcn.cpp
Proto-molecule: haven't looked that much into the setPixelColor() function. If there's a way to not use floats without giving up functionality, then I'm open to ideas. The new effects DO depend on this function, but I'm sure they can be rewritten to use a different function for AA.

wled00/FX_fcn.cpp
Proto-molecule: Im not sure how.

Proto-molecule:
sorry for the late response. I have not been active with WLED at all. Put It on the back burner.

blazoncek:
There were plenty of changes in WLED in preparation for 0.14 including 2D support.
Together with that I rewrote the whole effect logic system which makes this PR obsolete/unusable.
To help future enhancements like your PR I borrowed the AA logic somewhat and overloaded setPixelColor.
All available in 0_14 branch ATM.

blazoncek:
@Proto-molecule can you revisit this PR and recreate changes using API from 0.14?
It incorporates floating point/anti-aliased `setPixelColor()` and 'addPixelColor()`.
Then please use gaps in effects IDs for new effects.

I would also recommend closing this PR unmerged and opening a new one.

Proto-molecule:
Ok yes, I'll kind of have to since I’m use this fireworks effect in my personal builds and I’d like to be on the latest WLED. However there is a slight problem. The original anti-aliasing function was able to control the pixel length using a floating point number (not just a length of 1). This feature was arguably the most important part to this fireworks effect. The exploding pixels would start out large and shrink as they fade away over time to a size of even less than 1. 

In addition, think there are other scenarios when controlling the size of the pixel in this way would be advantageous. For example the bouncing-ball effect looks a little funny when the ball is a single LED on a 500led strip. Grouping the LEDs in the segment will fix this but then also cause AA to bleed into multiple LEDs. Another candidate could also be the meteor effect. The “meteorSize” could be passed onto the AA function. 

The way the AA function is written in 0.14, I don’t see much use for it aside from a pixel with a length of just “1” moving around the strip.

With all this being said, I do understand the reason for minimizing floating point calculations. I’ll do some thinking and try to come up with something.

blazoncek:
Then perhaps change terminology as AA is normally dealing with pixel and subpixel smoothing.
In such case an overloaded function with pixel width could be used (if AA==true then use new parameter pixel width).

I can see an issue in such case that since segment length/size is determined in pixels you may not get *exact* virtual segment length/size. You may need to accommodate that in effect itself.

Proto-molecule:
ok I'm closing and may resubmit an updated version.

PR #2469, opened 2022-01-03
Fix for the Animated_Staircase usermod animation

IgorMarkovic:
As of v13.0-b3 I see an animation problem when the usermode is switching on/off the segments. Since that version a crossfade effect was added when segments are switched on/off. The Animated_Staircase usermod however always uses segment 1 when turning on/off a segment. The result is that segment 1 will be animated multiple times when the stairs light is turned on/off.


KlodZed:
Thank you very much.

PR #2473, opened 2022-01-08
Create v2 usermod_word_clock_matrix.h

bwente:
Tried using the old usermod on the new build, found out a lot has changed since then. My best attempt to update it. Still needs some help, but it is working. I would like to preconfigure some of the default settings. I am also having an issue with  Error 12: Preset Not Found


Aircoookie:
Hey and thank you for the PR. Would you consider adding a readme file so users know exactly which word clock design your mod supports?

PR #2477, opened 2022-01-09
Improvements to usermod "MQTT switch"

herm:
1) Automatically load this user mod when it is configured.
2) Use pin manager to track allocated pins.

### DIFF
--- a/usermods/mqtt_switch_v2/usermod_mqtt_switch.h
+++ b/usermods/mqtt_switch_v2/usermod_mqtt_switch.h
@@ -32,6 +32,8 @@ static const bool switchDefaults[NUM_SWITCH_PINS] = { MQTTSWITCHDEFAULTS};
 #undef ON
 #undef OFF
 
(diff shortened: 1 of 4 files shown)


blazoncek:
Please while you are at it, update the usermod to use Usermod configuration pages to allow user to change GPIO pins during runtime.

On a side note, there is MQTT support for buttons built into core WLED and for controlling relays/switches there is a Multi Relay usermod both of which far surpass this usermod in every regard.

herm:
I had a look at the multi relay usermod. It probably makes sense to merge these two usermods. Currently multi relay lacks the following two functions
1) Specify default value after power on
2) Home assistant autodiscovery

Especially the second one is important for me. I will try to port it from my usermod.

blazoncek:
You just need to modify onMqttConnect() and onMqttMessage() in Multi relay to make it compliant with HA auto discovery.
A minor change.

As for default values: that would be a bit more involved to allow user to set them in UI which is preferred way since 0.12.

PR #2478, opened 2022-01-09
Add define to set data pins to HIGH when relay is off

c3n:
I'm using an IRF3708 MOSFET instead of a relay to disconnect the LEDs when WLED is off. Since it only disconnects GND and not +5V a LOW on the data pin creates a small leakage current which I didn't want. On ESP8266 the LED gets turned off so if the pin is shared with data there is no problem.
This PR adds a compile-time define (that one could set in the .ini file) that switches all configured digital data pins on ESP32 to HIGH using rmt_set_idle_level whlie the relay is off.
Tested with 2 SK6812 Strips connected to an esp32

### DIFF
--- a/wled00/button.cpp
+++ b/wled00/button.cpp
@@ -279,6 +279,29 @@ void handleButton()
   if (analog) lastRead = millis();
 }
 
+void esp32RMTInvertIdle()
+{
+  bool idle_out;
+  for (uint8_t u = 0; u < busses.getNumBusses(); u++)
+  {
+    Bus *bus = busses.getBus(u);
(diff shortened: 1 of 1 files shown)


blazoncek:
What happens if first bus is not RMT driven? (AKA APA102 LEDs which use SPI)

softhack007:
Hi,

> Since it only disconnects GND and not +5V 

Is there a specific reason why you disconnect GND instead if disconnecting 5V?

c3n:
Yes, I used a MOSFET to disconnect the LED strip instead of a relay and don't need to worry about 3.3V/5V

PR #2492, opened 2022-01-17
Usermod rotary brightness, color and color temperature

dkneisz:
add usermod to control brigness and color/color tempearature using a rotary encoder

- pins for rotary encoder (CLKPin and DTPin) can be defined in usermod settings
- change amount for RGB (Hue) and color temperature  can be defined in usermod settings
- button needs to be configured in LED preferences and set in usermot settings
- single press switches between brightness and color/color temperature mode
- double press switches betweeen RGB and white

### DIFF
--- a/usermods/usermod_rotary_bct/README.md
+++ b/usermods/usermod_rotary_bct/README.md
@@ -1,0 +1,27 @@
+# Rotary Encoder (Brightness, Color and Color temperature)
+
+Based on "usermod_rotary_brightness_color"
+
+V2 usermod that allows changing brightness and color/color temperature using a rotary encoder, 
+change between modes by pressing a button, change between RGB and white by double-pressing a button (many encoders have one included)
+
+
+## Installation
+
+define `USERMOD_ROTARY_ENCODER_BCT` e.g.
+
+`#define USERMOD_ROTARY_ENCODER_BCT` in my_config.h
+
+or add `-D USERMOD_ROTARY_ENCODER_BCT` to `build_flags` in platformio_override.ini
+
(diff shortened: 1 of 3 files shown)


blazoncek:
You can use core function `void colorKtoRGB(uint16_t kelvin, byte* rgb)` for conversion from K to RGB.
For other conversion functions take a look into colors.cpp

IMO it would be better to add CCT functionality to existing and very good ALT rotary encoder (I am also currently working on a few enhancements there).

dkneisz:
> You can use core function `void colorKtoRGB(uint16_t kelvin, byte* rgb)` for conversion from K to RGB. For other conversion functions take a look into colors.cpp
Thanks, I will look into that. This will be better than including an additional library.

> IMO it would be better to add CCT functionality to existing and very good ALT rotary encoder (I am also currently working on a few enhancements there).
I'm not sure which usermod you exactly mean, but I didn't want to change an existing usermod and decided to use an existing one as a kind of template and create a new one. That way someone can use the original if they only need the functionality provided there, if they don't need/use my enhancements. I know that there are multiple usermods for rotary encoders, and they all do different things. I chose one that (almost) fits my needs and extended it. There are some that can do more, but I don't needed it, so I didn't choose to use them. Maybe combining all functionality into one usermod and make every functionality deactivateable (is that even a word?) would be the "way to go". But then again, I don't want to change a mod that some other user made, because it may have been the intention of that user just to have this functionality. However if this functionality would be included into WLED (so not a usermod anymore), that would be a different topic.

dkneisz:
I changed the code to use `colorKtoRGB` instead so I don't have to include an external library. I also removed that part mentioning that in the original comment.

PR #2497, opened 2022-01-19
Add Home Assisant MQTT autodiscovery for usermod multi_relay.

herm:
See title.
Refs #2477.

### DIFF
--- a/usermods/multi_relay/readme.md
+++ b/usermods/multi_relay/readme.md
@@ -36,7 +36,7 @@ wled/deviceMAC/relay/0 on|off
 
 1. Register the usermod by adding `#include "../usermods/multi_relay/usermod_multi_relay.h"` at the top and `usermods.add(new MultiRelay());` at the bottom of `usermods_list.cpp`.
 or
-2. Use `#define USERMOD_MULTI_RELAY` in wled.h or `-D USERMOD_MULTI_RELAY`in your platformio.ini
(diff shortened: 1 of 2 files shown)


blazoncek:
Haha. 😁 Will check it out.
Unfortunately I do not use HA so no way of testing that part.

blazoncek:
Please check the modifications and report back.

herm:
Looks good to me. 
Tested and works fine.

blazoncek:
Relay is a remote controlled switch, whereas switch may mean plain physical (rocker) switch.
I intend to keep term relay since it is also used in MQTT message broadcasts with other devices.

BTW I added ability for periodic broadcasts in case broker goes down for some reason.

Will be merging shortly.

herm:
Periodic broadcast should not be required.
If the broker goes down we lose the connection and on reconnect (in `onMqttConnect`) the current state is published. The same happens when WLED loses the WIFI connection or reboot. If some other client loses the connection to the broker the broker has stored the retained message and will deliver it to the client. It's one of the nice things about MQTT that the application doesn't has to care about this kind of connection problems.

blazoncek:
Not everyone is using Home Assistant (including me) and retained messages are something that should not hold operational states since they may not reflect current state.
Retained messages also pollute broker and a user needs to manually clean them if device is no longer available.
I checked several IoT devices I own and all of them periodically broadcast current state they are in.

PR #2511, opened 2022-01-25
Set default relay pin to 11 for ESP32C3

lsiudut:
Pin 12 is by default SPIHD. If set during boot may effectively prevent
system from booting completely, what happened on my custom ESP32C3 based
board with extenal SPI flash.

I've seen multiple successful reports of booting ESP32C3 boards
without this change. My guess would be that those are using chips
with integrated flash which may be behaving differently. It doesn't
change the fact that according to datasheet using of pin 12-17 should
be avoided.

### DIFF
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -223,10 +223,12 @@ WLED_GLOBAL int8_t btnPin[WLED_MAX_BUTTONS] _INIT({0});
 #else
 WLED_GLOBAL int8_t btnPin[WLED_MAX_BUTTONS] _INIT({BTNPIN});
 #endif
-#ifndef RLYPIN
-WLED_GLOBAL int8_t rlyPin _INIT(12);
-#else
+#ifdef RLYPIN
 WLED_GLOBAL int8_t rlyPin _INIT(RLYPIN);
+#elif CONFIG_IDF_TARGET_ESP32C3
+WLED_GLOBAL int8_t rlyPin _INIT(11);
+#else
+WLED_GLOBAL int8_t rlyPin _INIT(12);
 #endif
 //Relay mode (1 = active high, 0 = active low, flipped in cfg.json)
 #ifndef RLYMDE


blazoncek:
You can create a build environment and define `-D RLYPIN=11` for such purpose.

lsiudut:
Yes, but it would still cause problems for anyone who builds the project in other than blessed ways. This approach seems to be more explicit.

Frankly it seems that WLED could use a bigger rework related to release of S2/3 and C3/6 models, plenty of assumptions in place are legacy esp32-centric.

That said, if doing thing by build environment is the way you prefer to carry it forward then I don't mind closing this PR.

lsiudut:
To provide more context, I've been building the project in Arduino IDE. The resulting image worked for the first time, subsequent boots were failing though. Took few cycles to pinpoint the culprit.

I also had to hack the code a bit to enable gpios >35 pins for ESP32-S2.

I don't know if you have plans to support those platform longer term, if so then I'm more than happy to drop some code from time to time. I intend to follow the authors thought, as it clearly works so far.

blazoncek:
this is a rabbit hole that can go very deep. and we've been there a few times already. 😃 
rather than stuffing everything into the code we could default every GPIO to -1.
the truth is there are several controller boards developed with certain _default_ values in mind (historical).
there are no controller boards for C3 (or S2) available specifically for WLED that I am aware of so they may implement anything in the future. such controllers (their developers) could add build environment to PIO and submit PR.
C3 and S2 are on the _supported in the future_ list (even though not yet there) and they will get their platformio.ini build environment entries (which can include overrides for almost any GPIO).

and yes, ArduinoIDE sucks in this regard.

> Frankly it seems that WLED could use a bigger rework related to release of S2/3 and C3/6 models, plenty of assumptions in place are legacy esp32-centric.

You are welcome to do a cleanup.

lsiudut:
Thanks for shedding some light on the state of things. Makes sense.

> You are welcome to do a cleanup.

I was expecting this kind of response, but I believe we both know perfectly that one-random-contributor big cleanups are rarely the way to go. Let me assure you that it wasn't my intention to complain, I merely shared observation as I play with S2 and C3 a lot these days. Could probably made it more elegant.

I'm looking into using those uC to control certain parts of lights in my home, as a result I'll likely make an attempt of landing few features, like rs486/modbus or CANBUS support. As it come for reworks, no promises, still happy to help out if it comes to testing the new platforms though.

blazoncek:
I was a random one-time contributor once. 😉 
There was no offense meant. I just don't own any C3 or S2 or have the time to tackle those so any muscle is welcome, being on-time or a dedicated.

PR #2516, opened 2022-01-28
Binary Websockets for Peek

Aircoookie:
This switches the protocol used for the Peek feature via WebSockets from JSON to binary format.
This reduces transmitted data to a third (3 byte per LED instead of 9) and as such should improve stability.
Maximum transmitted LEDs was also increased from 180 to 256.
`/json/live` was not documented to be an official endpoint of the JSON API and is now disabled by default, however can be re-enabled using `-D WLED_ENABLE_JSONLIVE`


wled00/my_config_sample.h
```diff
@@ -13,8 +13,8 @@
  * DO NOT make changes to the "my_config_sample.h" file directly! Your changes will not be applied.
  */
 
-// force the compiler to show a warning to confirm that this file is included
-#warning **** my_config.h: Settings from this file are honored ****
+// uncomment to force the compiler to show a warning to confirm that this file is included
+//#warning **** my_config.h: Settings from this file are honored ****
```
blazoncek, line 17: Finally! 😉

PR #2517, opened 2022-01-29
Extend Serial functions. Baudrate changeable during runtime and fast LED data retrieval 

ChuckMash:
Adds some serial connection functionality

Can now change baud rate during runtime to be faster
Retrieve LED strip data as JSON blob or preferably Bytes (fast!)

[Test results for retrieving LED strip data with various baud rates](https://gist.github.com/ChuckMash/99f90434ff61df9470f18468d9cbe017)
Results shown from ESP8266, but note that FPS is even higher when tested on ESP32

[Example python serial client with runtime baud rate change and LED strip data retrieval](https://gist.github.com/ChuckMash/54aa01ef7260015b69afecc42be99916)


wled00/wled_serial.cpp
```diff
@@ -19,6 +19,15 @@ enum class AdaState {
   TPM2_Header_CountLo,
 };
 
+void update_baud_rate(int rate){
+  if (!pinManager.isPinAllocated(1)){
```
blazoncek, line 23: Please add proper pin allocation detection.
isPinAllocated() supports PinOwner class and utilize that for detection (in this case it is allowed to use GPIO1 when PinOwner is DebugOut).

wled00/wled_serial.cpp
```diff
@@ -46,6 +57,40 @@ void handleSerial()
...
+        } else if ( next == 0xB1 ){ update_baud_rate( 230400 );
+        } else if ( next == 0xB2 ){ update_baud_rate( 460800 );
+        } else if ( next == 0xB3 ){ update_baud_rate( 500000 );
+        } else if ( next == 0xB4 ){ update_baud_rate( 576000 );
+        } else if ( next == 0xB5 ){ update_baud_rate( 921600 );
+        } else if ( next == 0xB6 ){ update_baud_rate( 1000000 );
+        } else if ( next == 0xB7 ){ update_baud_rate( 1500000 );
+        
+        } else if (next == 'l'){ // LED Data return in JSON blob. Slow, but easy to use on the other end.
+          if (!pinManager.isPinAllocated(1)){
```
blazoncek, line 71: Same as previous.

wled00/wled_serial.cpp
```diff
@@ -46,6 +57,40 @@ void handleSerial()
...
+            uint16_t used = strip.getLengthTotal();
+            Serial.print("[");
+            for (uint16_t i=0; i<used; i+=1){
+              Serial.print(strip.getPixelColor(i));
+              if (i != used-1) {Serial.print(",");}}
+            Serial.println("]");}
+            
+        } else if (next == 'L'){ // LED Data returned as bytes. Faster, and slightly less easy to use on the other end.
+          if (!pinManager.isPinAllocated(1)){
```
blazoncek, line 80: Same as previous.

blazoncek (changes requested):

Aircoookie:
I like the functionality of dynamically changing baudrate at runtime. I will consider implementing setting a default boot baudrate as a new setting as well. (and documenting the fact that staying at 115200 would still be required for debugging and using Improv)

Agreeing with pinOwner change request, will also do a quick cleanup before merge, so please let me know once ready from your side @ChuckMash :)

Aircoookie:
One more note: For the binary protocol, I'd prefer we go with a standard protocol, e.g. [tpm2](https://gist.github.com/jblang/89e24e2655be6c463c56).
Receiving of tpm2 is also supported by WLED. This just means that the packet starts with 0xC9 0xDA, two bytes with the packet length (LED count *3) and ends with 0x36 after the data in RGB format.
Would you agree with this?

wled00/wled_serial.cpp
ChuckMash: I will go back and take a look.

I used isPinAllocated(1) here because it is used elseware already in wled_serial for returning state and info JSON.
https://github.com/Aircoookie/WLED/blob/master/wled00/wled_serial.cpp#L63-L64

ChuckMash:
> One more note: For the binary protocol, I'd prefer we go with a standard protocol, e.g. [tpm2](https://gist.github.com/jblang/89e24e2655be6c463c56). Receiving of tpm2 is also supported by WLED. This just means that the packet starts with 0xC9 0xDA, two bytes with the packet length (LED count *3) and ends with 0x36 after the data in RGB format. Would you agree with this?

Agreed. If WLED already receives tmp2, then this is also what it should send for binary data.

Unless WLED can already receive 4 byte RGBW with tmp2 in addition to 3 byte RGB, I might need to change how my PR handles RGBW though.

Aircoookie:
Seems to work well for me, merging :)
Looking forward to a new PR whenever you have time!

PR #2523, opened 2022-02-02
Old style multicolor christmas lights

fishbone-git:
One for later on.  This replicates old style multicolor christmas strings with 5 colours, either static (speed=0) or in a chase.  This is meant optimally for strings, but strips can look good with some spacing set.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -780,6 +780,29 @@ uint16_t WS2812FX::mode_chase_random(void) {
   return chase(SEGCOLOR(1), (SEGCOLOR(2)) ? SEGCOLOR(2) : SEGCOLOR(0), SEGCOLOR(0), false);
 }
 
+/*
+ * Special chase which sets up like an old school multicolor light string.
+ */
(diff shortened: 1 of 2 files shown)


blazoncek:
You can replace Haloween (53) IMO.

Aircoookie:
I believe this is very similar to the Colorful effect with max intensity and the C9 New palette. Probably not worth a standalone effect to add purple, the 5th color. Might add it as a new palette though :)

blazoncek:
FYI It is possible to create a custom palette via JSON upload in 0.14.

fishbone-git:
good tips. Thanks! 14 looks great so far almost looks like effects via usermods could be not far over the horizon (enumeration of ID looks like the big hurdle)

blazoncek:
> looks like effects via usermods could be not far over the horizon (enumeration of ID looks like the big hurdle)

That was the intention of the rewrite. It may be possible to make a list of reserved IDs. But usermod effects are up to users (and a possible usermod setting).

PR #2527, opened 2022-02-06
Add usermod only sends mqtt messages on button and pir

tmjd:
I wanted to have buttons and PIR sensors generate MQTT messages but not attempt to change the state of the LEDs. That is so I can use Home Assistant to make any changes to WLED. So I've made this rather simple usermod to do just that.

Let me know if there is anything else I need to do for this.

### DIFF
--- a/usermods/buttons_only_send_mqtt/readme.md
+++ b/usermods/buttons_only_send_mqtt/readme.md
@@ -1,0 +1,17 @@
+# Usermod only send mqtt message on button and PIR
+
+This user mod will only send mqtt messages on button presses (short, long, double) and
+PIR motion. It will not attempt to turn on or off the LEDs or apply any presets. 
+To send double button presses it is necessary to set a non-zero value for the `double`
+action for the button on the Time & Macros under the Button actions section.
+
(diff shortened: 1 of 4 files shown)


blazoncek:
You can achieve the same effect by creating empty preset (i.e. `{"v":true}`) and assign it to button or PIR button.
For PIR, PIR_sensor_switch usermod allows greater flexibility.

Aircoookie:
Or just assign an ID of a non existing preset for the button action, which means you don't have a faux preset in the UI. Make sure to use a high ID like 240 so you don't accidentally make an actual preset with that ID in the future.

tmjd:
The reason I didn't like using a non existing preset is that if I'm on the WLED UI page it shows an error that the preset was not found. I like creating an empty preset like @blazoncek suggested. I hadn't previously looked into making a preset be an API macro which IIUC is what I think making it empty with `{"v":true}` is doing and causing the Macro to be a no-op.  

I'm sure what I've stated above is probably known to you both but hopefully this closed PR will be something someone in the future finds helpful.

Thanks for the great project.

PR #2528, opened 2022-02-06
esp32cam

bitbrain-za:
Adds esp32-cam as another option. Tested without any problems.

https://www.banggood.com/Geekcreit-ESP32-CAM-WiFi-+-bluetooth-Camera-Module-Development-Board-ESP32-With-Camera-Module-OV2640-p-1767566.html

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -12,7 +12,7 @@
 ; default_envs = travis_esp8266, travis_esp32
 
 # Release binaries
-default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, esp32dev, esp32_eth, esp32s2_saola, esp32c3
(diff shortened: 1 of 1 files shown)


Aircoookie:
Thank you for the PR.
Does the default `esp32dev` environment/binary not work on esp32cam?

bitbrain-za:
> Thank you for the PR. Does the default `esp32dev` environment/binary not work on esp32cam?

It almost definitely does... No hurt feeling if this gets rejected, I just like it explicit.

Aircoookie:
I'll close this for now, since I'd like for the platformio.ini file to stay as short as possible. I will go through other environments too and remove those that are no longer needed.

PR #2530, opened 2022-02-07
bugs, json remote repeat, cmd &R=

Proto-molecule:
Just some bug I found.

the repeat command isn't working quite right when using "irEnabled == 8" (custom json ir remote).
it would instead repeat the codes of the built in remotes.

api color commands are broken, (&R=, &G=, &B= etc.). 

I think these patches fix the issues.


blazoncek:
Looks like you were spying on me. 😉 
Just yesterday I discovered another flaw in IR code and updated it. While I did notice inconsistency in repeatable commands I didn't mess with it. Yet.

Your approcah is correct and I approve the PR. Still, IMO we could replace those ifs with a switch codeblock and just have ifs for IR_2x.

Proto-molecule:
That’s a funny coincidence

Currently messing with an IR project. These are just my  personal fix’s  I used without changing too much.
I figured a PR would be more helpful rather than opening an issue.

Switch codeblock 👍. I had a feeling you’d work your magic once you saw all the ifs 😆

PR #2531, opened 2022-02-07
Change Default Relay Pin to -1

bole5:
By default, relay pin is asigned to GPIO12. This may bring unwanted side-effects for WLED users. 
A few examples:
- debugging is not possible as JTAG uses GPIO12 for TDI
- using this pin for relay prevents booting of WLED on ESP32C3
- regardless of if you have a relay or not, this PIN will be initialised and therefore not usable until reassigned in settings

This PR changes default relay pin assignment to -1 resulting in better future stability, compatibility and allowing debuging. Of course it is possible to define RLYPIN when compiling own firmware, but given that PIN12 already caused problems in two use cases, it might be prefered to let users that actually use relay set their preferred pin explicitly.

Disclaimer: I am aware that default assigment of pins has been discussed before and this PR has no intention to re-start the discussion about general principles.

### DIFF
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -225,7 +225,7 @@ WLED_GLOBAL int8_t btnPin[WLED_MAX_BUTTONS] _INIT({0});
 WLED_GLOBAL int8_t btnPin[WLED_MAX_BUTTONS] _INIT({BTNPIN});
 #endif
 #ifndef RLYPIN
-WLED_GLOBAL int8_t rlyPin _INIT(12);
+WLED_GLOBAL int8_t rlyPin _INIT(-1);
 #else
 WLED_GLOBAL int8_t rlyPin _INIT(RLYPIN);
 #endif


bole5:
Hi @Aircoookie,
Thanks for merging this PR. Just noticed that we need to modify following info pages to reflect that user should select relay pin and default state (high/low) in the UI:
 - https://kno.wled.ge/features/relay-control/
 - https://github.com/Aircoookie/WLED/wiki/Control-a-relay-with-WLED

I did not find a way to submit a PR for Wiki pages.

Aircoookie:
Thank you, updated!

kno.wled.ge is built from this repository: https://github.com/Aircoookie/WLED-Docs
If you want to submit a PR for a change to a single page, you can easily do so by using the little pencil symbol in the top right of every page.

The GitHub wiki is deprecated and will no longer be updated, but instead link to kno.wled.ge :)

fishbone-git:
@Aircookie, the github wiki has been formally deprecated for over 3 months.  The header mentions "gradually dismantled" but what is the prerequisite for simply wiping it and replacing with only a redirect to kno.wled.ge? Is it a case of proofreading and checking that all content has been copied?  If so, need help?

Issue #2500, opened 2022-01-20
Expose per-segment color space info to JSON API


Aircoookie:
**Is your feature request related to a problem? Please describe.**
UI and external integrations cannot determine the supported colors spaces/models of configured busses on a per-segment level.

**Describe the solution you'd like**
A JSON array, as lightweight as possible while still being concise, that makes available the following information:
- Whether a bus (partly) driven by that segment supports RGB color
- Whether a bus in the range of that segment supports a white channel
- Whether a bus in the range of that segment supports color temperature (CCT)

In case a bus supports CCT, it usually also supports at least the white channels of the 3 main colors, which control the brightness of the light.

An initial draft proposal for this functionality stores this info in an byte `lc` (light capability) per segment, e.g.

Bit | Capability
|---|---|
0 | Segment supports RGB color
1 | Segment supports white channel
2 | Segment supports color temperature
3-7 | Reserved (expect any value)

Therefore:

`lc` value | Capabilities
|---|---|
0 | None. Indicates a segment that does not have a bus within its range, e.g. because it is not active.
1 | Supports RGB
2 | Supports white channel only
3 | Supports RGBW
4 | Supports CCT only, no white channel (unused)
5 | Supports CCT + RGB, no white channel (unused)
6 | Supports CCT (including white channel) 
7 | Supports CCT (including white channel) + RGB

Note that CCT is controllable per-segment, while RGB color and white channel have 3 color slots each per segment.

```json
{
  "info": {
    "seg": [
      {"lc": 3},
      {"lc": 1},
      {"lc": 4},
      {"lc": 0}
    ]
  }
}
```
Above the proposed JSON structure. `{"info":{"leds":{"lc":[3,1,4,0]}}}` would be more lightweight, but wouldn't make it possible to add more per-segment info in the future, and also doesn't make it as clear to an API user that `lc` is a per-segment value.

**Describe alternatives you've considered**
Always display the color wheel, even though it has no effect on the bus configuration (e.g. segment that controls only CCT busses).
This does not allow for a user interface that can dynamically show only the controls that actually apply to that segment.

**Additional context**
[...]

PR #2533, opened 2022-02-08
Add per-segment light capability info

Aircoookie:
Implements proposal discussed in #2500 
- ~~Adds `info.seg` array for per-segment info. Contains all active segments in ascending ID order. If a segment between two active segments is not active, it is still included (this is so that the ID does not have to be explicitely given)~~ 
- Adds `info.leds.seglc` array, giving the capabilities of the busses included in each segment in ascending ID order up to last active segment (0 for non-active segment)
- Adds `info.leds.lc`, giving the OR'd capabilities of all currently active segments
- Fixes `info.leds.rgbw`, now is true only if a bus that supports _both_ RGB _and_ White channel is configured
- Deprecates `info.leds.rgbw`

TODO
- Add dynamic show/hide of controls based on segment capabilities to the UI (in another PR)
- Update documentation
- Specify effect of the "Dual" auto white mode (slider value 0 = Auto-calculate white channel from RGB using the "Brighter" mode, 1-255 manual mode only) on the reported light capability:
Options:
a) always report RGB as part of the capabilities as the RGB color has an effect if the set white value is 0
b) report RGB as part of the capabilities only if any of the 3 white channels is currently set to 0
c) never report RGB as part of the capabilities if white or CCT hardware only (which most accurately reflects the hardware, and is true for values 1-255, but is incorrect due to the fact that the brightness of the white channels is defined by the set RGB color if the white value is set to 0)
d) enhance "Dual" mode by adding a dedicated control to switch between manual and auto white mode to the UI instead of mapping white 0 to auto-white.

Personal preference is b) with the perspective of implementing d) later (Reasoning: "Dual" auto white mode is not the most logical/intuitive, why would RGB color have an effect on light state if white is set to 0, but not otherwise)
The approach currently implemented by this PR is c).


blazoncek:
Controlling sliders and other UI elements is implemented in two branches (handle-names - just support for SoundReactive extensions to be removed from mode names - and slider-control for full implementation).

blazoncek:
Interesting things will happen when a segment spans different buses.
This is technically possible and I have found a use case where it might be handy. Still...

If we change "Reset segments" behaviour to create per-bus segments instead of a single one spanning all buses, we may be on a right track. It will be a users intentional change to get a single segment over multiple buses.
This will cause some frustration for those that relied on multiple pins to drive seemingly one long strip, but I think they can adjust. Or we can check if all buses feature exactly the same configuration and then allow "Reset segments" to create single segment.

blazoncek:
By _interesting things_ I was referring to the fact that it will behave almost as is currently.
No differentiation for CCT.

blazoncek:
What also crossed my mind is the complexity of detecting if a segment features different buses (if you can remember I had this implemented in my first CCT implementation).
Why not just have bus info and keep segments the way they were?
It is true that in such case receiver may not know what a segment supports but in reality it should not be concerned with such thing since WLED handles it gracefully.
Some people may have their HA poll WLED every few seconds.

Aircoookie:
Hmm, I think I'll go with a simple `info.leds.seglc` array after all instead of a new `info.seg` array.
Option C has just too much overhead, and in option B (currently implemented) the array structure is different (inactive segments between active ones are included since their IDs are missing) from that of `state.seg`, which is stupid from an API perspective.

Option A:
```
{"seglc":[1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1]}
```
Extra JSON buffer size 278, transmitted string len: 43

Option B:
```
{"seg":[{"lc":1},{"lc":1},{"lc":1},{"lc":1},{"lc":1},{"lc":1},{"lc":1},{"lc":1},{"lc":1},{"lc":1},{"lc":1},{"lc":1},{"lc":1},{"lc":1},{"lc":1},{"lc":1}]}
```
Extra JSON buffer size 528, transmitted string len: 153

Option C:
```
{"seg":[{"id":0,"lc":1},{"id":1,"lc":1},{"id":2,"lc":1},{"id":3,"lc":1},{"id":4,"lc":1},{"id":5,"lc":1},{"id":6,"lc":1},{"id":7,"lc":1},{"id":8,"lc":1},{"id":9,"lc":1},{"id":10,"lc":1},{"id":11,"lc":1},{"id":12,"lc":1},{"id":13,"lc":1},{"id":14,"lc":1},{"id":15,"lc":1}]}
```
Extra JSON buffer size 784, transmitted string len: 271

Aircoookie:
> What also crossed my mind is the complexity of detecting if a segment features different buses (if you can remember I had this implemented in my first CCT implementation).

I was worried about this as well :D Profiling shows that this functionality requires 163 microseconds on 16 segment ESP8266, so essentially neglegible

PR #2534, opened 2022-02-08
Float and better 3rd party library compatibility

henrygab:
These 

#### Float

As you know, neither the ESP8266 nor ESP32 have hardware acceleration for doubles (only single-precision, aka float).  This changes a couple locations where a value was declared as a `double` (which is the default for number literals with decimal point), to explicitly be declared as a `float` (by using `f` suffix).

#### Compatibility

The header `wled_math.h` defined a number of constants using `#define` using exceptionally short symbol names (`A`, `B`, `C`, etc.).  This caused some integration headaches when 3rd party libraries also polluted the global namespace with short variables (with different meaning, of course!), resulting in conflicts for those symbols' definitions.

Changing a library's exposed symbols can break builds.  Therefore, it's easier to change wled's implementation.  Plus, it's just bad form to define such short symbols in the global namespace ... it's just asking for trouble!

The solution chosen is to declare the function using these variables as a template, with no definition.  Then, provide a single implementation, using `float` for parameter and return type.  This allows declaring the values as `static const` within the function, rather than using globally defined symbols.

The GCC compiler generates **_identical assembly_** (verified at -O0 for ESP GCC, and at -O2 on x86-64).  From caller's perspective, nothing has changed ... the function works identically also.

Obligatory link to godbolt.org, showing equivalence.

https://godbolt.org/z/nz3x7qzc6


blazoncek:
Maybe I am already too old... Much simpler would be to replace #define A ... with #define _CONST_A_ ...
But I think @Aircoookie already mentioned once that he intends to implement math in separate .cpp and .h files (avoiding global #define)

😄

Aircoookie:
I have to admit that I really don't understand the use of template here, though I believe limiting the scope of these constants to `atan_t` is a great idea as long as the compiler correctly recognizes these vars as const and keeps them out of RAM.

There is in fact even a redefinition warning within WLED itself, which this PR should fix:
```
In file included from wled00\ntp.cpp:3:
wled00\wled_math.h:85: warning: "B" redefined
   85 | #define B -0.287434475393028
      |
In file included from wled00\ntp.cpp:2:
wled00\wled.h:671: note: this is the location of the previous definition
  671 | #define B(c) (byte(c))
      |
```

henrygab:
> I have to admit that I really don't understand the use of template here

Hmmm... I might have been able to declare the function as `inline`?  I suppose with LTO, the linker will remove the duplicates as well?

Aircoookie:
Or just use less arbitrary / short macro names :)

henrygab:
> as long as the compiler correctly recognizes these vars as const and keeps them out of RAM

good news: the godbolt link shows that the functions compile to identical assembly.

blazoncek:
This does not link if math functions are moved to separate .cpp file.

PR #2541, opened 2022-02-11
Added ledmap selection to presets.

### LINKED ISSUE #2275: Allow easy selection of ledmap in preset
Having multiple mapping files, or configs which can be combined with a Preset, by having another "Combobox " in the Preset settings, where you can select a mapping-config

**Additional context**
For Example, mapping-config1 is combined with Preset 1 Let's say it's from top to bottom
mapping-config2 is combined with Preset 2, and is just from the middle to the edges.

**Why?**
Would allow to for example safe "Pictures" for a LED Matrix by changing the mapping, but it would be too much work to update the mapping config everytime, and it wouldn't be possible to change it by a Time related Preset, or for example in a playlist

Thank you for maintaining this wonderful Project <3

blazoncek:
Can close #2275.
Screenshot 2022-02-11 at 14 20 55


Sturm2002:
Hi,
what exactly is the syntax for the ledmap.json ?
could you maybe give an quick example ?

Sturm2002:
Files :

```
ledmap1.json: 
    {"map":[10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0]}
ledmap2.json:
    {"map":[5, 4, 3, 2, 1, 0, 10, 9, 8, 7, 6]}
```
also tried :
- without whitespaces
- "maps" instead of "map"


When creating a new Preset, and selecting a ledmap, the ledmap won't be selected, when you then try to select it again, it still doesnt get applied, and the extended window of the preset won't open.

https://user-images.githubusercontent.com/50213395/153616215-71e019ce-2289-4ed5-84d7-43debf33eebf.mp4

blazoncek:
My tests were all successful. There is a bug (in my fork) after saving a preset that UI is not updated but preset still works.
I use map without spaces: `{"map":[5,4,3,2,1,0,11,10,9,8,7,6,17,16,15,14,13,12]}`.

blazoncek:
This has been superseded by https://github.com/Aircoookie/WLED/pull/2737
Closing.

PR #2547, opened 2022-02-16
Sync segment bounds

Aircoookie:
- Added extra option that toggles receiving start + stop + offset.
- Fixed `effectCurrent` not applied to any segments by colorUpdated (bug since led.cpp simplification in https://github.com/Aircoookie/WLED/pull/2427/commits/7eb029dcb607d1f2d87ee6625665dee579eedd16)
- Major state changed logic rework/simplification
- Changed behavior of main segment, instead of usually segment 0 it is now usually the first selected active segment
- FIxed segment repeat turning off segments
- Reverted ESP8266 platform to 2.7.4, as there are occasional WebSockets/malloc related crashes with 3.0.2 (not exclusive to this PR, but also on master)

This will require a lot of testing since there are quite a few seldomly used edge cases in the state change stuff, like nightlight, UDP, HTTP API, ... I'm positive it fixes more than it breaks though!


blazoncek:
No need to revert to 2.7.4 for ESP8266.
There is a workaround that works quite well without severe side effects by changing this in ws.cpp.
```
void handleWs()
{
  if (millis() - wsLastLiveTime > WS_LIVE_INTERVAL)
  {
    #ifdef ESP8266
    ws.cleanupClients(2);
    #else
    ws.cleanupClients();
    #endif
    bool success = true;
    ...
```
...and the second...
```
void sendDataWs(AsyncWebSocketClient * client)
{
...
    size_t len = measureJson(doc);
    size_t heap1 = ESP.getFreeHeap();
    buffer = ws.makeBuffer(len); // will not allocate correct memory sometimes
    size_t heap2 = ESP.getFreeHeap();
    if (!buffer || heap1-heap2<len) {
      releaseJSONBufferLock();
      ws.cleanupClients(0); // disconnect all clients to release memory
      return; //out of memory
    }
...
```
I have this implementation in my fork since December and works quite well. I did add a WS reconnect logic to index.js in case the client gets disconnected.
I'll make a PR if you agree on the change or just include changes into this PR.

Aircoookie:
Thank you for adding the error handling + reconnect!

Will make one small change:
With automatic reconnect, I am worried that if someone uses >2 devices/tabs for whatever reason, they will constantly kick each other off WS due to the 2 connection limit and attempt to reconnect, which will cause a lot of unneeded TCP traffic and extra load on the esp.
I think it makes more sense if it only reconnects once a request was made via HTTP and a JSON response received successfully, that way the UI instances actively controlling the device will have priority over any idle ones.

wled00/data/index.js
```diff
@@ -974,9 +974,15 @@ function makeWS() {
...
-	ws.onclose = function(event) {
-    	d.getElementById('connind').style.backgroundColor = "#831";
-  	}
+	ws.onclose = (e)=>{
+		d.getElementById('connind').style.backgroundColor = "#831";
+		ws = null;
+		if (lastinfo.ws > -1) setTimeout(makeWS,500); //retry WS connection
+	}
+	ws.onopen = (e)=>{
+		ws.send("{'v':true}");
```
Aircoookie, line 989: @blazoncek One tip for your fork, this line is actually unneeded :)
WLED WS will always send `si` automatically upon initial connection
https://github.com/Aircoookie/WLED/blob/master/wled00/ws.cpp#L18

blazoncek:
You can increase to 3 WS connections and it may still be stable on most occasions since if RAM is depleted the workaround in `sendDataWs()` will keep memory in check.
More than 3 and ask yourself if ESP8266 is designed for that. ;)

PR #2559, opened 2022-02-24
Add locate button for easy lat/lon auto-fill

Aircoookie:
Adds button to Time settings that opens a new tab to https://locate.wled.me
This then passes the coordinates back to the settings page using `window.opener.postMessage()`.
Makes it a lot more convenient to use sunrise/set as no manual coordinate lookup and entry is needed.


blazoncek:
This is cool. 👍

PR #2561, opened 2022-02-25
IR rewrite.

blazoncek:
- added CCT (WW/CW) support
- support for applying change to main segment or all selected segments


Aircoookie:
Looks good! 👍 
`setValuesFromFirstSelectedSeg()` is indeed not required in `set.cpp` as that is the first thing done in `stateUpdated()`.

PR #2565, opened 2022-03-03
Settings PIN code protection.

blazoncek:
Protect settings pages with PIN code.


wled00/wled_server.cpp
```diff
@@ -443,16 +462,21 @@ void serveSettings(AsyncWebServerRequest* request, bool post)
...
+    if (subPage != 252) strcat_P(s, PSTR(" settings saved."));
+    else {
+      server.removeHandler(editHandler);
+      createEditHandler(correctPIN);
+    }
     if (!s2[0]) strcpy_P(s2, PSTR("Redirecting..."));
 
-    if (!doReboot) serveMessage(request, 200, s, s2, (subPage == 1 || subPage == 6) ? 129 : 1);
-    if (subPage == 6) doReboot = true;
+    serveMessage(request, 200, s, s2, (subPage == 1 || (subPage == 6 && doReboot)) ? 129 : (correctPIN ? 1 : 10));
```
Aircoookie, line 478: Putting a longer redirect time here unfortunately won't defeat a script (or browser back button).
We need to lock the set.cpp code from accepting new PINs for a few seconds (same for OTA lock now that reboot on Save is removed)

wled00/data/settings_sec.htm
```diff
@@ -46,6 +33,12 @@
 			fO.value = '';
 			return false;
 	    }
+		function checkNum(o) {
```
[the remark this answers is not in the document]
blazoncek: Agreed that complexity is prefered but since we do not have HTTPS it makes no sense (well, less sense) to make pin ultra complex. 4 digit pin should provide adequate protection until we add HTTPS.

wled00/wled_server.cpp
blazoncek: The actual brute force protection is in line 448.

blazoncek:
Any progress on this?

Aircoookie:
Tested, looks good mostly :) Can merge after 0.13.0 release (how about tomorrow for 0.13.0?)

Two minor things I've noted:
You only get the `/settings/pin` page when going to a specific settings page (e. g. LED settings), but entering the correct pin redirects you back to the main settings menu. This is weird. I'd recommend we prompt for the PIN already for the main settings menu.

xml.cpp serves the cleartext pin to GetV(). Should do something similar to WiFi password instead (e.g. replace with asterisks)

blazoncek:
Agreed on 0.13.

I also agree that returning to main setting page is annoying but couldn't figure out how to get the proper destination.
Also serving plaintext PIN looks awful. Unfortunately no good idea came to my mind how to prevent removing existing PIN if field is left blank.

blazoncek:
Is superseded by #2610

blazoncek:
This has been superseded by https://github.com/Aircoookie/WLED/pull/2737
Closing.

PR #2567, opened 2022-03-04
UI: show color controls based on segment light capabilities

Aircoookie:
Implements UI part of #2500 
E.g. the color wheel is hidden if no selected segment supports RGB color.


blazoncek:
You could have based this on _slider-control_ branch.

Aircoookie:
Indeed, but they only really overlap in that they can hide some UI controls, the reason for the hiding and thus JS code is different (effect based vs. segment based)
I still plan to include the majority of the _slider-control_ branch changes at some point :)

PR #2578, opened 2022-03-13
Add option to set module name at runtime, and add extra examples in platformio_overrides.ini

SpeakingOfBrad:
This modifies wled.h to allow for defining a name at runtime that will be used for the module name.

Also adds some examples to platformio_overrides.ini for some handy defines that can be used to set up your light at runtime.
This also has the added benefit that when you use the "Factory reset" in Security & Updates, WLED stays properly configured for the hardware it's attached to.

### DIFF
--- a/platformio_override.ini.sample
+++ b/platformio_override.ini.sample
@@ -44,3 +44,6 @@ build_flags = ${common.build_flags_esp8266}
 ;   for the Magic Home LED Controller use PWM pins 5,12,13,15
 ;   for the H801 controller use PINs 15,13,12,14 (W2 = 04)
 ;   for the BW-LT11 controller use PINs 12,4,14,5
+;   
+; set the name of the module - make sure there is a quote-backslash-quote before the name and a backslash-quote-quote after the name
+;   -D SERVERNAME="\"WLED\""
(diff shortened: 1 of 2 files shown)


SpeakingOfBrad:
Ideally I'd also like there to be a way to set the color order of the led strip at runtime, but the only way I've been able to do so is to modify the end of line 90 in FX_fcn.cpp, and I'm not sure how to squeeze a define in there that could be overridden.

blazoncek:
I think this may be better suited into[ WLED docs](https://github.com/aircoookie/WLED-docs) with description how `build_flags` are used and a few _specific_ examples.
The [compile](https://kno.wled.ge/advanced/compile-wled/) WLED section is quite obsolete.

Even though `SERVERNAME` is a welcome addition it may not have great value since it would require separate build environment for each _module_. It may allow you to rename **default** device name to something other than WLED though.

BTW next time please follow [this](https://github.com/Aircoookie/WLED/wiki/How-to-properly-submit-a-PR) guide.

Aircoookie:
Agreed that Docs should be updated, will do so later. In particular I have missed the fact that there are TWO compilation pages:
https://kno.wled.ge/basics/compiling-wled/
https://kno.wled.ge/advanced/compile-wled/
the latter of which is almost entirely obsolete (save for the screenshots) due to its focus on hardcoded analog builds which are no longer required or even supported since 0.12.
Still I welcome your changes to the PIO example override and will merge this.

PR #2579, opened 2022-03-15
Update IRremoteESP8266 to 2.8.2

misery:
This fixes some C3 issues


Aircoookie:
Alright, it passed CI the second try. Merging :)

blazoncek:
My PIO would not update to 2.8.2 ...
Says library missing.

Aircoookie:
Try again now, I believe it might be a caching problem since 2.8.2 was only released yesterday

PR #2585, opened 2022-03-17
Fix cross-origin error in File Mode

WoodyLetsCode:
Hi,
when you use the UI in File Mode the Peek-Button does not work because of this error:
grafik

To fix this, I simply changed line **48** in `liveviewws.htm` from 
```
var ws = top.window.ws;
``` 
to 
```
var ws;
```
.

I'm a bit confused why you did implement it like this. As far as I know, `top.window.ws` is the same as `ws`. Or am I wrong here?

Anyway, I tested File Mode and now I don't get this error anymore and the Peek-Button now works again:
grafik

Also in Online Mode Peek still works.


Aircoookie:
Hey!
Since the liveview is loaded in an iframe, `top.window.ws` will attempt to use the WebSocket connection of the parent page (UI), which puts less strain on the ESP. Your change will open a second WS connection instead.
I've pushed an alternate solution, a try/catch block that will only open a second connection in File mode.

WoodyLetsCode:
Hi!
Thanks for your explanation.
I think your implementation is a good solution here.

PR #2595, opened 2022-03-23
Add usermode to control Wiz lights

eibanez:
Hello! I just wrote and tested this usermod, which enables WLED to control Wiz lights. This is done by replicating the UDP API. I included a small README in the pull request, which I'm copying below for your reading pleasure ;)

Happy to answer any questions!

# Controlling Wiz lights

This usermod allows the control of [WiZ](https://www.wizconnected.com/en/consumer/) lights that are in the same network as the WLED controller.

The mod takes the colors from the first few pixels and sends them to the lights.

## Configuration

First, enter how often the data will be sent to the lights (in ms).

Then enter the IPs for the lights to be controlled, in order. There is currently a limit of 10 devices that can be controled, but that number
can be easily changed by updating _MAX_WIZ_LIGHTS_.

### DIFF
--- a/wled00/usermods_list.cpp
+++ b/wled00/usermods_list.cpp
@@ -217,6 +217,6 @@ void registerUsermods()
   #endif
 
   #ifdef USERMOD_WIZLIGHTS
-  usermods.add(new QuinLEDAnPentaUsermod());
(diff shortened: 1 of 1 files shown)


eibanez:
Thanks!

PR #2607, opened 2022-03-30
Next universe overflow and Art-Net DMX start address

4lloyd:
This PR fixes two things:
1. LEDs overflowing to the next universe could cause some strange fast blinking behaviour on some LEDs near the overflow point. This was caused by differences in the calculations between the first universe and the universes after that. Because of this some LED were controlled by multiple universes at once.
2. When using Art-Net the DMX channel was offset by +1 and the last channel of a LED at the end of the first universe could therefore not always be used. The cause of this is that the DMX data in an Art-Net packet starts at 0 and in E1.31 at 1.

Both problems are only visible when using another DMX start address then the first one.

This change is a breaking change because from now on the DMX start address for Art-Net starts at 1 instead of 0. This should be the expected behaviour when using Art-Net (or any DMX device) as far as I know. This makes the working of the DMX start address in Art-Net the same as in E1.31. Nothing has changed for when the E1.31 protocol is used. Old configurations using Art-Net and DMX start address 0 still work the same.

I have tested the code in multiple ways, but I don’t have a LED strip long enough to test multiple universe overflows.

### DIFF
--- a/wled00/e131.cpp
+++ b/wled00/e131.cpp
@@ -177,7 +177,7 @@ void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol){
         const uint16_t dmxChannelsPerLed = is4Chan ? 4 : 3;
         const uint16_t ledsPerUniverse = is4Chan ? MAX_4_CH_LEDS_PER_UNIVERSE : MAX_3_CH_LEDS_PER_UNIVERSE;
         if (realtimeOverride) return;
-        uint16_t previousLeds, dmxOffset;
+        uint16_t previousLeds, dmxOffset, ledsTotal;
         if (previousUniverses == 0) {
           if (dmxChannels-DMXAddress < 1) return;
           dmxOffset = DMXAddress;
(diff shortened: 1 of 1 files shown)


pk88888:
Nice! i can try to test it during next few days

pk88888:
Ive just checked it on 5 universes and it works! Great, now we can use more than one universe via artnet. Thank you!

blazoncek:
Unfortunately this PR broke main segment sync by reverting `if (realtimeOverride && !(realtimeMode && useMainSegmentOnly)) return;`

Aircoookie:
@blazoncek check https://github.com/Aircoookie/WLED/commit/958cd35e21ad796a9d56879811fa466b7fdba09d, I've simplified these checks again so that override also works in main segment mode :)

Aircoookie:
@blazoncek check https://github.com/Aircoookie/WLED/commit/958cd35e21ad796a9d56879811fa466b7fdba09d, I've simplified these checks again so that override also works in main segment mode :)

blazoncek:
I don't seem to get it how this behaves but I will trust you since I have no way of testing Art-Net or E1.31. 😄 
You can undo last commit (in e131.cpp) in "blazoncek-merge-working".

lordjaxom:
I was just doing my first experiments with DMX on a WLED controller controlling a RGBW LED matrix and was running into exactly this issue. Last LED of first universe was blinking and the second universe was one LED short.

Compiled main branch from scratch and now it works, thank you :+1:

PR #2608, opened 2022-03-31
V2 usermod for wordclock with 11x10 LEDs and 4 minute dots

Hasenpups:
- implementation of V2 usermod to handle wordclock with 11x10 pixels and 4 additional dots for the minutes.
- the usermod has two parameters available in the GUI to enable/disable the mod and to display a "es ist" on the clock.


blazoncek:
There is no need to close PR, add commits and open new PR.
You can add commits to original PR without issues (though we'd prefer a single commit with all relevant changes so it is easier to track what's changed).

Hasenpups:
> There is no need to close PR, add commits and open new PR. You can add commits to original PR without issues (though we'd prefer a single commit with all relevant changes so it is easier to track what's changed).

Okay, thank you. I will keep an eye on this for the next time. Was my first PR :)

PR #2610, opened 2022-03-31
Full blazoncek fork merge

Aircoookie:
Introduces all features and fixes of @blazoncek dev branch.
PR for feedback and adjustments.


blazoncek (commented):
Most of the uncommented changes are cosmetic or bugfixes (including all usermods).
applyPreset()/handlePresets() may need special treatment as may refreshLightCapabilities().

wled00/json.cpp
```diff
@@ -478,9 +463,11 @@ void serializeState(JsonObject root, bool forPreset, bool includeBri, bool segme
 
   root[F("mainseg")] = strip.getMainSegmentId();
 
+  bool selectedSegmentsOnly = root[F("sc")] | false;
```
blazoncek, line 466: New feature to allow saving only selected segments to preset.

wled00/json.cpp
```diff
@@ -823,6 +795,78 @@ void serializeNodes(JsonObject root)
   }
 }
 
+void serializeModeData(JsonArray fxdata)
```
blazoncek, line 798: Handlers for SR extensions to mode/effect data.

wled00/presets.cpp
```diff
@@ -4,101 +4,139 @@
...
-bool applyPreset(byte index, byte callMode)
+#ifdef ARDUINO_ARCH_ESP32
+static char *tmpRAMbuffer = nullptr;
+#endif
+
+static volatile byte presetToApply = 0;
+static volatile byte callModeToApply = 0;
+static volatile bool checkPlaylist = false;
+
+bool applyPreset(byte index, byte callMode, bool fromJson)
```
blazoncek, line 15: The biggest change: async preset loading.

wled00/wled.cpp
```diff
@@ -181,14 +99,24 @@ void WLED::loop()
     yield();
     #endif
 
+    handlePresets();
```
blazoncek, line 102: Async preset loading entry point.

wled00/wled.cpp
```diff
@@ -212,6 +140,13 @@ void WLED::loop()
     yield();
   }
 
+  // 15min PIN time-out
+  if (strlen(settingsPIN)>0 && millis() - lastEditTime > 900000) {
```
blazoncek, line 144: This could be made customizable at some point.

blazoncek:
@Aircoookie any progress on this?

A few main points to test:
- auto-white global override/per pin setting
- settings PIN code/protection
- slider control (effect defaults, slider visibility & name, palette hiding & defaults)
- ESP8266 web stress test
- HA integration (CCT & sensor info)
- UI across different devices (I only have Apple, FF & Safari)

Also, it may be good to add Play and Pause glyphs to Wicons font.

Aircoookie:
Thank you for the latest additions! :)
I have to be honest that I will likely not find the time/energy to get to this in the next 14 days. New study course is absolutely mental, I am thinking about canceling it since I have 0 time left for WLED :(
Would you make another branch with just the new UI (the 3 files index.js/css/htm only, you do not have to include the ESP side code to make it actually work). That might make it a little easier, I'm basically trying to get the diff between this branch and main down piece by piece until I can merge it completely 👍

Aircoookie:
Reason I can't just copy UI files myself is that I would get attribution for all your changes which is not fair imo. And I did not find an easy option to merge only some files (without creating a huge change set by reverting changes to all other files)

wled00/wled_server.cpp
```diff
@@ -112,13 +112,11 @@ void initServer()
     serveSettings(request);
   });
   
-  server.on("/settings.js", HTTP_GET, [](AsyncWebServerRequest *request){
-    serveSettingsJS(request);
-  });
+  // "settings.js" request also handled by serveSettings()
```
blazoncek, line 115: This will not work in every situation. Tested.

Aircoookie: Thanks for letting me know, will check it out :)

tools/cdata.js
```diff
@@ -267,7 +267,7 @@ writeChunks(
         str
           .replace(
             /function GetV().*\<\/script\>/gms,
-            "</script><script src=\"settings/settings.js?p=0\"></script>"
+            "</script><script src=\"/settings/s.js?p=0\"></script>"
```
blazoncek, line 270: FYI: including / at the beginning may break reverse proxy configurations.

wled00/wled_server.cpp
```diff
@@ -534,24 +546,19 @@ void serveSettings(AsyncWebServerRequest* request, bool post)
     }
 
     if (subPage == 252) {
-      server.removeHandler(editHandler);
-      createEditHandler(correctPIN);
+      createEditHandler(correctPIN && !otaLock);
```
blazoncek, line 549: Why not OTA lock?

wled00/wled_server.cpp
```diff
@@ -228,70 +235,67 @@ void initServer()
...
   server.on("/rangetouch.js", HTTP_GET, [](AsyncWebServerRequest *request){
     AsyncWebServerResponse *response = request->beginResponse_P(200, "application/javascript", rangetouchJs, rangetouchJs_length);
-    response->addHeader(F("Content-Encoding"),"gzip");
+    response->addHeader(FPSTR(s_content_enc),"gzip");
     setStaticContentCacheHeaders(response);
     request->send(response);
   });
   
-  createEditHandler(correctPIN);
+  createEditHandler(correctPIN && !otaLock);
```
blazoncek, line 258: OTA lock should have no influence on ability to visit /edit page.
Security setting cannot be changed using /edit.

Aircoookie: `/edit` page is able to set (not retrieve sec) arbitrary settings, but more critically, able to alter the pages displayed to the browser, which is a security risk. It can even be seen as a form of OTA update :)
PIN is an adequate protection, but if the user has OTA lock active before and after the update, they would still expect `/edit` to be locked. I will change the label of OTA to make clear that it also locks edit though.

tools/cdata.js
Aircoookie: Right! Will make sure they do not start with `/` although that requires the workaround with both .js/.css handlers as the main settings page will access `/settings.js` while sub pages access `/settings/settings.js`. Adding redundant handlers is the lesser problem compared to proxies not working though!

wled00/wled_server.cpp
blazoncek: Sorry but I fail to understand how a user can update if OTA is locked? If you mean serially then all measures are irrelevant since with physical access anything is possible.
If OTA is locked it will prevent any kind of updates (not true ATM due to bug) to firmware even if PIN is entered and that should be the main purpose of it IMO (which also holds for WiFi settings).

Aircoookie: They can't update the firmware, but they _can_ arbitrarily change the `index.htm` served to clients.

Regarding ArduinoOTA: I think it should stay separate from OTA lock - most common use case for OTA lock is allowing access to port 80 only. This allows local updates via ArduinoOTA without having to unlock OTA each time. If ArduinoOTA is disabled via the checkbox and OTA locked, it can't be re-enabled without the correct OTA password. Should add a note though that ArduinoOTA will remain active unless disabled manually.

blazoncek: I still think that locking OTA should be consistent for both options: web update and Arduino OTA. Otherwise an adversary can update firmware even if a user thinks he/she locked OTA updates and considers WLED secure.

> `/edit` page is able to set (not retrieve sec) arbitrary settings, but more critically, able to alter the pages displayed to the browser

Locking settings (and `/edit`) using PIN code should be sufficient for most cases where WLED can be put to public use (with the warning about non encrypted transfer). Removing fake `index.htm` can be done without too much effort in case when adversary intercepted PIN and uploaded fake `index.htm`. Combined with OTA lock it can prevent most malicious attacks (except carefully planned ones) even with plain HTTP. I would not expect `/edit` to be locked when OTA is disabled/locked. It has become a feature that was suggested as a help in case of many problem solving options. It will also prevent uploading _ledmaps_ and such which should not be a condition of OTA lock.

PIN in itself will also prevent updating OTA and will lend a grace period of 15 minutes (when entered) in which time you can do OTA (Arduino or web) if you need to update.

blazoncek:
I have an update on async preset loading.

WLED UI knows content of a preset so instead of sending `"ps":X` it can send complete JSON API for a preset (including HTTP API if that is the case) and alongside just notify core that preset has been applied (i.e. `"pt":X`).
In such case a call to `applyPreset()` is not needed and correct state can be returned immediately to caller (ie UI).
For other preset loading events a normal async loading does not constitute a problem since the caller (button, usermod, ...) does not care.

If an external source (i.e. not WLED UI) wants to change a preset by using preset ID I cannot see a reason why would it _require_ a returned state to reflect exact preset content immediately since it can subscribe to websockets or periodically request a state to get the correct state at any given time (since WLED state may change from another source and in such case it would maintain incorrect state internally after another source changed WLED state).

So I would propose to change WLED UI to send complete preset API to WLED and pass a preset ID alongside API ("pt" from above). This may simplify preset loading and issues tied to it.

blazoncek:
This has been superseded by #2737 
Closing.

PR #2613, opened 2022-04-02
ESPDmx replaced with SparkFunDMX

berkut0:
Hi.

Recently I developed [my own controller "LightBrain"](https://github.com/BrightestBrains/LightBrain) with RS485/DMX support.

I tried to complile firmware with DMXEsp code originally used by WLED, but the firmware could not make the DMX work. I tried to isolate the problem and compiled code purely from this library which was supposed to send simple messages, but my DMX device didn't understand it. Then I wrote a simple code in micropython and the messages started to come to the device. The code was crooked and the messages only reached the device after two sends, but it showed some problems with the ESPDmx code.

Then I ran isolated tests with the SparkFunDMX library and everything worked flawlessly. Moreover, the library already has a read from port functionality, which can be used later. 

https://user-images.githubusercontent.com/8849896/161372533-3f97cb14-9e91-4e9e-a686-3b0accf480cf.mp4


berkut0:
Oops, my bad

Didn't check compatibility with esp8266, only ESP32

Aircoookie:
We could either stick with ESPDMX for ESP8266 and use SparkFunDMX only on the ESP32, or we could switch to the LX libraries as proposed in #817 which would also allow for configurable pins

berkut0:
LXESP32DMX great and full of capabilities library: DMX / RDM (!) / Multitasking

But there are my thoughts:
1. More restrictive BSD license than MIT.
2. Huge code base with differences in intitialization and usage. ESPDmx and SparkFunDMX in usage pretty similar and changes are insignificant.
3. SparkFunDMX tested and proven

Aircoookie:
Agreed! Then let's go with a combination of SparkFunDMX and ESPDmx :)

berkut0:
I'm not very experienced in all this git stuff.
Can I make another commit and try another pull request?
Should I use #ifdef?

Also, does WLED currently use LEDPIN in const.h? Since the web interface has the ability to select a function pins, it looks obsolete, but I haven't been very investigative. And what you think, would it be hard to implement rx/tx/rt pin selection in web interface?

berkut0:
@blazoncek I will stick to the instructions, thank you!

Aircoookie:
Closing in favor of #2652

Issue #417, opened 2019-12-05
add ArtNet ArtPoll support 


daavery:
it would be nice if WLED supported ArtNet in addition to E1.31.

artnet can be either very simple - ArtNetDMX packet only or complex with support for RDM and remote programming of node features

I'm looking at various artnet code bases there a couple for esp but none are async :
https://github.com/rstephan/ArtnetWifi
https://github.com/natcl/Artnet
https://github.com/mtongnz/ESP8266_ArtNetNode_v2
https://github.com/hideakitai/ArtNet

there is a  more complete codebase for reference:
https://github.com/OpenLightingProject/libartnet

the spec for it all is:
https://art-net.org.uk/resources/art-net-specification/

thanks

Aircoookie:
Thank you for finding me a few links. Looking at https://github.com/rstephan/ArtnetWifi, it seems to communicate via UDP, just like E1.31 and be quite straightforward. If we are lucky, I should be able to add some nice async Artnet support!

abyssdj:
What is the progress on this so far? Great job on the software, it is much better than alternatives I have been using!

T-Arens:
Wanted to ask the same, is there any code for Artnet yet? Because otherwise I might give it a try. :-) I just used ESP8266_ArtNetNode_v2 on an ESP-01 to get my [RGBW LED Moving Head](https://www.lixada.com/p-l0761us.html) connected via WiFi, and that works beautifully with the free PC software [DMXControl](https://www.dmxcontrol.org). It has a nice editor for creating custom devices, too. Now I'd like to integrate my WLED based sideboard illumination, and here I am, thinking about Artnet. :-)

pille:
initially i wanted to have that, too. i'm now hooked on E1.31, which was elementary supported and improved that. since there's currently a lot going on (refactoring, open DMX pull requests) i would postpone this a little bit. the idea is to implement ArtNet eventually alongside the E1.31 implementation, as they are just envelopes for the same letter.

Aircoookie:
@T-Arens decided I want to work on this today. If everything works out, we might have artnet support in a few hours!

Aircoookie:
@pille yep, I'm just extending the `ESPAsyncE131` to deal with the different ArtNet packet structure. Will just call `handleE131Packet()` and work exactly the same as E1.31 as far as WLED is concerned.

abyssdj:
[...]

PR #2615, opened 2022-04-02
Add ArtPoll support

4lloyd:
This PR adds support for ArtPoll and ArtPollReply. This solves #417.

At the moment the OEM and ESTA manufacturer code are set to zero. Someone of the WLED team may request these codes, so the device will show up as WLED in lighting consoles. This is not a must for the code to work, but more of a "nice to have".

These codes can be requested via:
- OEM code: https://art-net.org.uk/join-the-club/oem-code-application/
- ESTA manufacturer code: https://tsp.esta.org/tsp/working_groups/CP/mfctrIDs.php

The changes were tested on the esp32dev and esp8266_2m platform.

Parts of the code are inspired by the code of: https://github.com/hideakitai/ArtNet. That library also has a MIT license.

### DIFF
--- a/wled00/e131.cpp
+++ b/wled00/e131.cpp
@@ -243,7 +243,8 @@ void handleArtnetPollReply(IPAddress ipAddress) {
         const uint16_t dmxChannelsPerLed = is4Chan ? 4 : 3;
         const uint16_t ledsPerUniverse = is4Chan ? MAX_4_CH_LEDS_PER_UNIVERSE : MAX_3_CH_LEDS_PER_UNIVERSE;
         const uint16_t dimmerOffset = (DMXMode == DMX_MODE_MULTIPLE_DRGB) ? 1 : 0;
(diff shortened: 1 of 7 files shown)


Aircoookie:
Thank you very much! I will likely merge this after #2610

erhan-:
Hello everyone, seems like there are some good news. https://github.com/Aircoookie/WLED/pull/2737 is merged. Will these changes be also merged now as told in https://github.com/Aircoookie/WLED/issues/417#issuecomment-1166325693 ?

Same for https://github.com/Aircoookie/WLED/pull/2619

Thanks for your contribution and help.

Aircoookie:
Hey, I just pushed a small change, by allocating ArtPollReply on the stack when needed, we can save precious global heap RAM.
Since I lack host software that supports ArtPoll, please test if it still works as intended if you have time.

4lloyd:
> Hey, I just pushed a small change, by allocating ArtPollReply on the stack when needed, we can save precious global heap RAM. Since I lack host software that supports ArtPoll, please test if it still works as intended if you have time.

I tried the change and Art-Net is still working 👍

PR #2619, opened 2022-04-04
Some Art-Net improvements

4lloyd:
This PR adds some improvements to the Art-Net processing:
1. Better checks if the received universes and LED data are in use. Now, WLED should only listen and switch to real-time mode when the universe is actually in use.
2. Corrected calculations when using legacy DMX start address 0.
3. Changed HMTL to prevent DMX start address 0.
4. Checking brightness against actual brightness. Making brightness changing the same for all modes with dimmers (except for effect mode). This also fixes an issue that the brightness would not change to the Art-Net value when changing between: Art-Net -> web interface (manual brightness change) -> Art-Net.
6. Added some mitigations to prevent possible integer underflow.
7. Added some small changes to make the code better readable.


4lloyd:
Hi! I did some extra tests but couldn't find anything wrong with the R and G channels.

In #2607 the first DMX start address was changed from 0 to 1. Maybe it has to do something with that?

If you can tell me your test configuration, I can try to replicate the problem.

pk88888:
I will check it again tomorrow and let you know

4lloyd:
@pk88888 Did you already have some time to check the R and G channel behaviour?

pk88888:
> @pk88888 Did you already have some time to check the R and G channel behaviour?

Sorry for such a delay ... I have already checked what it is about and this problem with the channel order does not result from your changes, but from the general order of channels in WLED.

If I have the GRB order selected in the WLED settings, my luminaire shines well and each color is displayed correctly.

At the moment when the driver gets data from Artnet or sACN, I have to set the RGB order in the madmapper or resolume so that the luminaire shines properly.

Aircoookie:
Merged in #2794 :) The long delay was due to me being uncertain whether prohibiting start address 0 would be a good change, however line 120 of `e131.cpp` takes care of differences between E1.31 and Art-Net just fine.

PR #2622, opened 2022-04-06
Extend Serial Functions. Adds realtime streaming to fast LED data retrieval 

ChuckMash:
Extends the serial TPM2 functionality a bit.

The bytes 'O' and 'o' now can start and stop the streaming of LED data being returned over the serial connection without specifically having to send 'L' byte to request it.

Any byte other than 'O' halts the streaming data so that other serial functionality can be used without specifically sending the stop streaming byte.

In the event of high frequency polling the LED data with 'L', this can help save resources and complexity.
Should also dissuade "over polling" of the serial connection.

### DIFF
--- a/wled00/wled_serial.cpp
+++ b/wled00/wled_serial.cpp
@@ -20,6 +20,7 @@ enum class AdaState {
 };
 
 uint16_t currentBaud = 1152; //default baudrate 115200 (divided by 100)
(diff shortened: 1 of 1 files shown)


ChuckMash:
I spoke too soon, need to investigate some related things before the pull request.
Closing for now.

PR #2638, opened 2022-04-18
WiZ Lights usermod - Adding more options and features

ChuckMash:
Adds the following configuration options to the WiZ Lights usermod

- Send Delay (ms)
    - Adds an optional millisecond delay after sending UDP packet to each WiZ light. 
    - Can help smooth out effects when using a larger number of WiZ lights
- Use Enhanced White
    - Adds the ability to use the WiZ lights onboard white LEDs instead of sending maximum RGB values
    - Tunable with warm and cool LEDs as supported by WiZ bulbs
    - Note: Only sent when max RGB value is set, need to have automatic brightness limiter disabled
    - ToDo: Have better logic for white value mixing to better take advantage of the lights capabilities
- Always Force Update
    - Can be enabled to always send update message to light, even when color matches what was previously sent.
- Force update every x minutes
    - Configuration option to allow adjusting the default force update timeout of 5 minutes.
    - Setting to 0 has the same impact as enabling Always Force Update

Other:
Increased default  MAX lights to 15
Assorted code changes and cleanup
Changed UI variable names
WiZ light count in UI now starts at 1 instead of 0

-----

ATTN: @eibanez for final approval to changes to their usermod.

### DIFF
--- a/usermods/wizlights/readme.md
+++ b/usermods/wizlights/readme.md
@@ -6,8 +6,30 @@ The mod takes the colors from the first few pixels and sends them to the lights.
 
 ## Configuration
 
-First, enter how often the data will be sent to the lights (in ms).
-
-Then enter the IPs for the lights to be controlled, in order. There is currently a limit of 10 devices that can be controled, but that number
+- Interval (ms)
+    - How frequently to update the WiZ lights, in milliseconds.
+    - Setting too low may causse ESP to become unresponsive.
+- Send Delay (ms)
+    - An optional millisecond delay after updating each WiZ light. 
+    - Can help smooth out effects when using a larger number of WiZ lights
+- Use Enhanced White
+    - Enables using the WiZ lights onboard white LEDs instead of sending maximum RGB values.
+    - Tunable with warm and cool LEDs as supported by WiZ bulbs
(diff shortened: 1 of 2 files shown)


eibanez:
@ChuckMash thanks for the changes, they look good! Thanks for the clean up and additional features.

Could you please add the following to the bottom of the README file? I meant to do it in my original submission but I lost the change in a merge.

```
## Related project

If you use these lights and python, make sure to check out the [pywizlight](https://github.com/sbidy/pywizlight) project. I learned how to
format the messages to control the lights from that project.
```

ChuckMash:
> @ChuckMash thanks for the changes, they look good! Thanks for the clean up and additional features.

It was my pleasure, thanks to you for getting the ball rolling on this usermod.




> Could you please add the following to the bottom of the README file? I meant to do it in my original submission but I lost the change in a merge.
> 
> ```
> ## Related project
> 
> If you use these lights and python, make sure to check out the [pywizlight](https://github.com/sbidy/pywizlight) project. I learned how to
> format the messages to control the lights from that project.
> ```

I added this, and I also remembered to add the addition configuration options to the README as well as updated the default bulb count mentioned.

We should be good to proceed now.

eibanez:
🚀

PR #2645, opened 2022-04-27
Add Multiway Switch

tomvleeuwen:
This pull requests adds a switch type that is independent from its position.
With this new "multiway" switch, flipping the switch will always toggle the On/Off state of WLED, regardless of previous actions via the webinterface. This is how a traditional multiway switch behaves as well.
I think it is more intuitive since you will only flip the switch if you want the On/Off state to change.


blazoncek:
The same can be achieved by creating a toggle preset (`T=2`) and assigning it to ordinary switch for both transitions.

tomvleeuwen:
Yes that works as well (although I had to use `{"on":"t"}` as preset since I had no clue where to enter `T=2`).

Should I close this PR?

blazoncek:
When creating preset you just enter T=2 after unchecking "Use current state".

PR #2648, opened 2022-04-30
Added Stairway_wipe to usermod list

lost-hope:
I added the Usermod Stairway_Wipe_Basic to the usermod_list as `STAIRWAY_WIPE_BASIC`
And I also Added a Web UI filed to be able to set the wipeState without recompiling the code.


Aircoookie:
Thank you for the PR. What is the use case for adding `wipeState` (a state variable that changes during the wipe) to settings? Shouldn't it always start at 0?

lost-hope:
You are right, I misunderstood the use of wipeState. Removed it and also changed the wipestate back to 0 as initial. (probably changed that while testing)

lost-hope:
@Aircoookie anything missing ?

PR #2657, opened 2022-05-05
Enable ESP watchdog by default

poelzi:
Use the ESP watchdog to detect hanging ESP and reset the firmware.
Can be disable by defining WLED_WATCHDOG_TIMOUT 0
Default timeout is 3 seconds on ESP32 and something on ESP8266 (currently ignored)

This will at least help with cases of some locked up controllers, but does not fix the underlaying problem.

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -276,6 +276,15 @@ void WLED::loop()
   loops++;
 #endif        // WLED_DEBUG
   toki.resetTick();
+
+#if WLED_WATCHDOG_TIMEOUT > 0
+  // we finished our mainloop, reset the watchdog timer
+  #ifdef ARDUINO_ARCH_ESP32
+    esp_task_wdt_reset();
+  #else
+    ESP.wdtFeed();
+  #endif
+#endif
 }
 
 void WLED::setup()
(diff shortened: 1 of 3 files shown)


poelzi:
I tested this by adding a endless loop, but have now experienced the watchdog work in RL. Animation stopped, some seconds later the node started with default settings.

blazoncek:
Rather than blindly enable WDT it woud be better to determine **where** the original WDT starvation occurs.
Normally all non customized code should feed the WDT on regular basis.

Still, having a flexibility to set WDT manually may be a good idea for those rare exceptions.

poelzi:
Of course, deadlocks and other problems should be fixed, but having a dead controller is ofter even more problematic. The timeout is so large, that custom code does not need to feed it, but finishing the loop is enough.

poelzi:
Esphome for example enables the watchdog.
https://github.com/esphome/esphome/blob/71fe2f7ed334c84d209fd641ec99bd090e1b79da/esphome/components/esp32/core.cpp#L41

Pretty common on Microcontrollers.

poelzi:
@blazoncek Interesting. I was able to fix this by disableing the watchdog when OTA is initiated.

blazoncek:
It works now. Thank you.
After a bit of additional testing I'll give it a go.

blazoncek:
I am not sure if my ESP is experiencing brownouts (since I have detector disabled) but sometimes it freezes and WDT is not resetting it.
Any ideas?

blazoncek:
Approved from my side.

blazoncek:
After using this on both ESP8266 and ESP32 without issues (and benefiting from WDT lockup prevention) I think it is a worthy addition.

PR #2666, opened 2022-05-10
✨ feat(usermod): add `NonVolatile Segments` v2 usermod

Stoom:
When using a board like a Quinled Dig Quad with multiple channels along with the `Make a segment for each output` option, the segments all get reset on each boot due to the initialization sequence.  This is a usermod workaround that restores the segment basic settings and a few additional build defs to prevent flashing the strips on boot.

The following are the build defs that I am using to create a better experience when used in a multizone environment:
```
  -D DEFAULT_BRIGHTNESS=255
  -D DEFAULT_SEGMENT_OPACITY=1

; User mods
  -D USEMOD_NV_SEGMENTS
  -D USERMOD_NV_SEGMENTS_AUTOSAVE_SECONDS=21600 ; Default to saving every quarter of a day
  -D USERMOD_NV_SEGMENTS_MAX_AUTOSAVE_SECONDS=86400 ; Max save timeframe of once per day
```

The settings looks something like this:
image
image

### DIFF
--- a/usermods/usermod_v2_nonvolatile_segments/readme.md
+++ b/usermods/usermod_v2_nonvolatile_segments/readme.md
@@ -1,0 +1,31 @@
+# NonVolatile Segments
+
+v2 usermod to automatically save and restore segment configures on reboot
+
+* brightness
+* colors
+* palette
+* CCT
+* name
+* power state
+
+If you have multiple channels and use the `create a segment per channel` featuer
+then on every reboot you'll loose any modifications you made to the segments.
+This is a workaround to the way the WLED initializes that will restore these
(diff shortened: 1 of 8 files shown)


Aircoookie:
Thank you for the usermod!

A way simpler way to achieve this thought is to create a preset including your segment configuration and setting that as boot preset.

Stoom:
Does that save the name of the segment though? I think for a majority yes you're right.

blazoncek:
Presets (with checked "Use current state") store *everything* including segment names.
Another option to have *current* state saved is Auto save usermod which saves state after a (configurable amount of) few seconds.
Closing as this can be achieved in other ways.

PR #2667, opened 2022-05-10
Extend Serial Functions. Adds Continuous Serial Streaming to fast LED data retrieval

ChuckMash:
Extends the serial TPM2 functionality a bit.

The bytes 'O' and 'o' now can start and stop the continuous streaming of LED data (as TPM2) being returned over the serial connection without specifically having to repeatedly send 'L' byte to request it.

Any received byte other than 'O' halts the streaming data so that other serial functionality can be used without specifically sending the stop streaming byte.

Continuous Serial Streaming only sends LED data when the strip has been updated.
Generally matching with "Frames / second" 

In the event of high-frequency polling of LED data with 'L', this eliminates the need to poll at a high rate,  providing new LED data updates as it becomes available, saving time and resources.


ChuckMash:
Since stalebot has reminded me, any thoughts on this PR?
If undesirable, I will close it out

PR #2685, opened 2022-06-05
add option to mix colors when clock pointers overlap

lordjaxom:
This pull request adds a checkbox "Mix colors" in the analog clock overlay settings. When checked, the colors of the clock pointers will be mixed when overlapping. I.e. if minute and second show at the same pixel, it will be yellow instead of green.


platformio.ini
```diff
@@ -12,7 +12,7 @@
 ; default_envs = travis_esp8266, travis_esp32
 
 # Release binaries
-default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, esp32dev, esp32_eth, esp32s2_saola, esp32c3
+; default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, esp32dev, esp32_eth, esp32s2_saola, esp32c3
```
blazoncek, line 15: remove updates to platforio.ini from your PR.

use platformio_override.ini for your custom build options

blazoncek (changes requested):

wled00/overlay.cpp
```diff
@@ -42,8 +42,8 @@ void _overlayAnalogClock()
     }
   }
   if (!analogClockSecondsTrail) strip.setPixelColor(secondPixel, 0xFF0000);
-  strip.setPixelColor(minutePixel, 0x00FF00);
-  strip.setPixelColor(hourPixel, 0x0000FF);
+  strip.setPixelColor(minutePixel, analogClockMixColors?(strip.getPixelColor(minutePixel)|0x00FF00):0x00FF00);
```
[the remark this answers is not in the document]
blazoncek: What about other colors? Mixing primaries may be straightforward, others not so.
Your intention is legitimate, though.

platformio.ini
lordjaxom: fixed.

wled00/overlay.cpp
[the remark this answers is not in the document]
blazoncek: IMO this is better than ORing.

wled00/data/settings_time.htm
```diff
@@ -190,6 +190,7 @@ <h3>Clock</h3>
 			12h LED: <input name="OM" type="number" min="0" max="255" required><br>
 			Show 5min marks: <input type="checkbox" name="O5"><br>
 			Seconds (as trail): <input type="checkbox" name="OS"><br>
+			Add pointer colors: <input type="checkbox" name="OA"><br>
```
blazoncek, line 193: Even though technically it is adding colors, the better term would be "mix".

blazoncek (approved):
Apart from a different term in settings everything looks good to me.
Unfortunately I have never tested analog clock functionality so cannot judge how mixing/adding colors will compare to existing way pointers are displayed but since the behaviour is optional there is no reason not to merge.

As a side note: there is an open PR that adds color addition function that preserves color balance after addition (scales 16 bit result back to 8 bit). May be worth exploring as an upgrade.

wled00/data/settings_time.htm
lordjaxom: I need the feature just so it is visible that minute and second (and possibly hour) are shown at the same pixel. Usually the segment is solid black before the overlay. But mixing the pointers into the effect's colors opens up a range of new possibilities too :-)

I found the function AddPreserveRatio in the PR, looks promising. 

Maybe after it's merged I'll make the option a dropdown to choose from a bunch of merge options?

blazoncek: That is good enough for me. 👍 
Even though that particular PR may not get merged I will ask the author to make a new one featuring just additive function.

blazoncek:
@lordjaxom could you update this PR to be compatible with 0_14 branch (and change destination to 0_14) as it is ready for public beta testing.
So this PR could go into beta testing as well.

lordjaxom:
Actually I'd rather like to drop this PR in favor of 2736 Analog Clock Usermod.

PR #2686, opened 2022-06-07
Allow multiple LEDs to be grouped when controlling via Artnet or sACN

### LINKED ISSUE #2325: E1.31 (DMX) group pixels
I am in the process of making a stage lighting installation with pixel tape and WLED. So far it's working really well, but controlling all pixels (60 leds/m) is eating up a lot of channels and universes. I don't need to control each led separately, in my case about 10-20 "groups" of 3 or 6 leds per meter would be enough. This would drastically reduce the amount of pixels to output and thus make it possible to run faster animations and maybe a cheaper license for the outputting software. Is this something that is possible?

HazeCore:
This PR adds the possibility to control more LEDs on a strip with less channels to better suit the needs of bigger productions in which WLED could come really handy as a cost-efficient LED controller. (Thanks for that btw ;) )

It adds capabilities to merge LEDs into groups of equal length which allows the DMX source to control more LEDs with fewer channels while sacrificing resolution.

The size of a group can be adjusted with an input field in the "Sync Interfaces" Dialog, named "DMX group size", which allows to set how many LEDs in a row on a stripe get controlled by the same DMX channel. The default value is 1, which act exactly like the original behaviour.
The value gets saved together with the other settings. 
The maximum group size is 1000.

I've tested the build:
- with an ESP32 and an ESP8266
- for start Addresses within a single Universe and more than one Universe
- with the sACN Protocol

It would close #2325.


blazoncek:
The functionality is already implemented in WLED by using _main segment only_ live data syncing (6079effae3dcb86230d235a2759b4813026b0855).
It uses regular segment grouping and spacing.

elf5000:
> This PR adds the possibility to control more LEDs on a strip with less channels to better suit the needs of bigger productions in which WLED could come really handy as a cost-efficient LED controller. (Thanks for that btw ;) )
> 
> It adds capabilities to merge LEDs into groups of equal length which allows the DMX source to control more LEDs with fewer channels while sacrificing resolution.
> 
> The size of a group can be adjusted with an input field in the "Sync Interfaces" Dialog, named "DMX group size", which allows to set how many LEDs in a row on a stripe get controlled by the same DMX channel. The default value is 1, which act exactly like the original behaviour. The value gets saved together with the other settings. The maximum group size is 1000.
> 
> I've tested the build:
> 
>     * with an ESP32 and an ESP8266
> 
>     * for start Addresses within a single Universe and more than one Universe
> 
>     * with the sACN Protocol
> 
> 
> It would close #2325.

Hi! Please share bin. file esp32, want to test this mode) In the new versions, they lost this very necessary function..

elf5000:
> For the Modifications made, i had forked the WLED version. You can still find the fork in my Profile... It's not very up-to date though and you probably have to recompile it by yourself. I will have a look if i can fild a binary File in my _very messy_ folder structure :D

Thanks a bunch!

ivanpianobcn:
Hello! I know this is quite old, but I'm trying to compile the source, and I'm facing some issues (I don't know much about compiling...)
Could any one compile the file and send/upload please?
I need this feature...

Thanks a lot

DedeHai:
this is not a support forum, use other channels please.

PR #2687, opened 2022-06-10
add "lighthouse2" effect

Mar-Pfa:
creates moving dot that can move e.g. around a circle also with support of multiple rows and smoooth transition
lighthouse (1)

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -388,6 +388,86 @@ uint16_t WS2812FX::mode_rainbow_cycle(void) {
   return FRAMETIME;
 }
 
+/*
+* moves a dot in primary color from left to right
+* dot width and speed is variable
+* dot fade in/out smooth
+* act as multi line display with a width of 36 columns (maybe later variable!?)
+*
+* SEGMENT.intensity = width of the dot
+* SEGMENT.speed = motion speed
+* ........................xXXx..->....
+* ........................xXXx..->....
+* ........................xXXx..->....
+*/
+
+uint16_t WS2812FX::mode_lighthouse2()
+{    
+  int rowLength = 36;
+
+  uint16_t counter = (((now * SEGMENT.speed)>>6)  + 1) % (rowLength << 8);
+
+  int fade = counter & 0xFF;
+  counter = counter >> 8;
+
+  double f = fade * 3.14159 / 512;
+  double f1 = sin(f) * 256 / 1;
+  double f2 = cos(f) * 256 / 1;
+  int fade1 = 255 - (int) f1;
+  int fade2 = 255 - (int) f2;
+  if (fade1 < 0)
+    fade1 = 0;
+  if (fade2 < 0)
+    fade2=0;
+
+  int posIn = counter % rowLength;
+  int posOut = (posIn + (SEGMENT.intensity>>4) + 1) % rowLength;
+
+  uint32_t color = SEGCOLOR(0);     
+  int w1 = W(color);
+  int r1 = R(color);
+  int g1 = G(color);
+  int b1 = B(color);
+
+  int r1New, g1New, b1New;  
+
+  for(uint16_t i = 0; i < SEGLEN; i++) {
+    int column = i % rowLength;    
+    
+    if (column == posIn)
+    {
+      // fade in             
+      r1New = (r1 * (fade1)) >> 8;
+      g1New = (g1 * (fade1)) >> 8;
+      b1New = (b1 * (fade1)) >> 8;
+    } else if (column == posOut)
+    {
+      // fade out            
+      r1New = (r1 * fade2) >> 8;
+      g1New = (g1 * fade2) >> 8;
+      b1New = (b1 * fade2) >> 8;
+    }  else if (
+           (posIn < posOut && column>posIn && column < posOut) 
+        || (posIn > posOut && ((column<posOut) || (column>posIn)))
+      )
+    {
+      // always on       
+      r1New = r1;
+      g1New = g1;
+      b1New = b1;
+    } else {
+      // black      
+      r1New = 0;
+      g1New = 0;
+      b1New = 0;
+    }
+    
+    setPixelColor(i, r1New, g1New, b1New, w1);    
+  }
+  return FRAMETIME;
+}
+
+
 
 /*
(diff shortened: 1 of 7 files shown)


wled00/FX.cpp
```diff
@@ -388,6 +388,86 @@ uint16_t WS2812FX::mode_rainbow_cycle(void) {
...
+* SEGMENT.intensity = width of the dot
+* SEGMENT.speed = motion speed
+* ........................xXXx..->....
+* ........................xXXx..->....
+* ........................xXXx..->....
+*/
+
+uint16_t WS2812FX::mode_lighthouse2()
+{    
+  int rowLength = 36;
```
blazoncek, line 406: Why 36? Not everyone wants a row length of 36.

wled00/FX.cpp
```diff
@@ -388,6 +388,86 @@ uint16_t WS2812FX::mode_rainbow_cycle(void) {
...
+
+  uint16_t counter = (((now * SEGMENT.speed)>>6)  + 1) % (rowLength << 8);
+
+  int fade = counter & 0xFF;
+  counter = counter >> 8;
+
+  double f = fade * 3.14159 / 512;
```
blazoncek, line 413: Use float. It is faster and has enought resolution.

wled00/FX.cpp
```diff
@@ -388,6 +388,86 @@ uint16_t WS2812FX::mode_rainbow_cycle(void) {
...
+  double f1 = sin(f) * 256 / 1;
```
blazoncek, line 414: Use WLED math functions. They are accurate & fast enough but use less memory.

wled00/FX.cpp
```diff
@@ -388,6 +388,86 @@ uint16_t WS2812FX::mode_rainbow_cycle(void) {
...
+
+  int r1New, g1New, b1New;  
+
+  for(uint16_t i = 0; i < SEGLEN; i++) {
+    int column = i % rowLength;    
+    
+    if (column == posIn)
+    {
+      // fade in             
+      r1New = (r1 * (fade1)) >> 8;
```
blazoncek, line 440: Use color_blend().

wled00/FX.cpp
```diff
@@ -388,6 +388,86 @@ uint16_t WS2812FX::mode_rainbow_cycle(void) {
...
+           (posIn < posOut && column>posIn && column < posOut) 
+        || (posIn > posOut && ((column<posOut) || (column>posIn)))
+      )
+    {
+      // always on       
+      r1New = r1;
+      g1New = g1;
+      b1New = b1;
+    } else {
+      // black
```
blazoncek, line 459: Recommended is to use SEGCOLOR(1) instead.

blazoncek (changes requested):
IMO this FX reqiures special set-up to be really effective.
Please consider rewriting it using comments and think about how to make it length independent.

Mar-Pfa:
Thanks for the Feedback! I improved and simplified the code a lot. Only thing i was not able to bring in was the wled_math.h. I wanted to use sin_t / cos_t - when adding wled_math.h to to fx.cpp i get build errors because of dublication. without adding the library i get build errors as the functions are not know :-(
Any ideas on that?

wled00/FX.cpp
Mar-Pfa: changed it to be configurable with parameter 2, now the dotwidth is fixed for now (but will be variable again with next version)

blazoncek: Taking 3rd parameter will be available in 2D version of WLED sometime after summer.

blazoncek:
You may need to add
```
#include "wled.h"
#include "fcn_declare.h"
```
to FX.cpp

Mar-Pfa:
:-( doesn't work...  wled.h is already in, adding fcn-declare.h does not help. the build problem is:

.platformio/packages/toolchain-xtensa/bin/../lib/gcc/xtensa-lx106-elf/4.8.2/../../../../xtensa-lx106-elf/bin/ld.exe: .pio\build\d1_mini_ota\src\ntp.cpp.o: in function `cos_t(float)':
ntp.cpp:(.text._Z5cos_tf+0xb4): multiple definition of `cos_t(float)'; .pio\build\d1_mini_ota\src\FX.cpp.o:FX.cpp:(.text._Z5cos_tf+0xb4): first defined here
.platformio/packages/toolchain-xtensa/bin/../lib/gcc/xtensa-lx106-elf/4.8.2/../../../../xtensa-lx106-elf/bin/ld.exe: .pio\build\d1_mini_ota\src\ntp.cpp.o: in function `sin_t(float)':
...

don't know how to fix this

At least -> with using just the plain "sin(x)" method it works so maybe we can still merge my pull-request!? getting this wled_math.h to work in FX.cpp might be more effort!?

blazoncek:
> don't know how to fix this

Oh, yes, it requires wled_math be a .cpp instead of .h. A known issue in current main branch.
You may need to wait until #2610 gets merged.
sin() and cos() will not be merged.

Mar-Pfa:
@[blazoncek](https://github.com/blazoncek) finally found an alternative :-) I was not aware of the sin16 / cos16 - so fixed the code, now it works and without sin() / cos() - hopefully it's fine now!?

blazoncek (approved):

blazoncek:
You may want to test-drive 2D implementation from my fork which will unfortunately make this PR obsolete if you set up your lights as 2D matrix.
Otherwise this is ok.

blazoncek:
I can. 😄 But will wait for @Aircoookie to peer review and approve (may take a while though).
Adding new effects is a delicate matter since it is very annoying (for users) if it gets removed later on due to duplicates.

Aircoookie:
Thank you for the great PR (and lamp!) and sorry for taking so long to review it.
You couldn't realistically have anticipated 2D support when you made this PR, so I apologize.
In 0.14, I personally think configuring the light as a matrix and using e.g. the standard lighthouse effect with "Bar" 1D->2D mapping would suffice.
Still I really like the smooth fade in and out, I think it could serve as a nice base for a combined 1D and 2D effect - nice smooth fade on 1D and a circular "fresnel lens" type of thing moving across the matrix in 2D, so I will leave this open for now :)

PR #2692, opened 2022-06-20
add TM1829 support

h3ndrik:
Provide TM1829 support.
Similar to TM1814. This is an adressable RGB 12V led strip. This type is already supported by NeoPixelBus.


blazoncek:
Unfortunately I have no way of testing this but it looks clean and to the point.
Approved.

blazoncek:
Question regarding off-refresh.
TM1814 requires so called "off-refresh" which is sending data to the chip also in WLED off state (sending black color similar to static/solid mode). Does TM1829 also require such signal to prevent the chip to enter demo mode?

blazoncek:
Looks like this is going to make it into WLED via #2737

Aircoookie:
Now found its way into main :)

PR #2723, opened 2022-07-19
BME280 Usermod Enhancements

albarlow:
This is my first pull request so my apologies if I've not done anything right.  I think I did something wrong the first time I tried to submit, I think I've got it write this time.

**Changes**
I added a Usermod interface for key settings. I used a PinArray for the SDA/SCL pins, but you can't name these individually.

I have also made a display to show the temperature/humidity values in the web interface's Info screen.

I had to change the definition of those items in order to allow these new functions to work. I have not noticed any negative side effects to this change.

At the moment, I've not figured out how to make Celsius/Farenheit toggleable due to the way the #define setup works.

Finally, I have added a routine to publish MQTT Discovery Topics for Home Assistant (toggleable in the Usermod screen).

**Testing Done**
I've been testing this on two devices, one ESP32 and one ESP8266, for a few months and haven't noticed any problems.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2067,9 +2067,9 @@
       "integrity": "sha512-a6sumDlzyHVJWb8+YofY4TW112G6p2FCPEAFk+59gIYHv3XHRhm9ltVQ9kli4hNWeQBwSpe8cRN25x0ROunMOw=="
     },
     "terser": {
-      "version": "4.8.0",
-      "resolved": "https://registry.npmjs.org/terser/-/terser-4.8.0.tgz",
-      "integrity": "sha512-EAPipTNeWsb/3wLPeup1tVPaXfIaU68xMnVdPafIL1TV05OhASArYyIfFvnvJCNrR2NIOvDVNNTFRa+Re2MWyw==",
+      "version": "4.8.1",
+      "resolved": "https://registry.npmjs.org/terser/-/terser-4.8.1.tgz",
+      "integrity": "sha512-4GnLC0x667eJG0ewJTa6z/yXrbLGv80D9Ru6HIpCQmO+Q4PfEtBFi0ObSckqwL6VyQv/7ENJieXHo2ANmdQwgw==",
       "requires": {
         "commander": "^2.20.0",
         "source-map": "~0.6.1",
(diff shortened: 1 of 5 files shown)


blazoncek:
Thank you for the updated usermod.

I do have one request some WLED users asked me. Can you modify this usermod to have similar API as does Temperature usermod so that other usermods could benefit from it?
By API I am thinking about public methods, JSON & MQTT properties. Please review Temperature usermod for values and structure.

albarlow:
I'm not sure whether I've done exactly what you're after regarding the API - I've added public variables to the BME280 usermod based on those in the Temperature usermod (TempC, TempF, Dew Point, Heat Index and Pressure).  Only complication is that this usermod utilises different function calls depending on whether user defines celsius or not whereas Temperature reads in Celsius and converts to F for reporting purposes.  I don't want to risk breaking any functionality for other users by rewriting to copy this.  I have handled this for the temperature, but the Dew Point and Heat Index are relative to the temperature so the user would need to know what the BME280 usermod was set to.

I've also addressed some areas where I'd previously assumed Celsius for reporting purposes as my test case is using Farenheit that I noticed when I was addressing the Celsius/Farenheit matter for API.

I didn't see any MQTT discovery in Temperature usermod and I think I'm doing similar with the actual reported measurements.

Regarding JSON Reporting; again, I think I've approximately matched the format, I've pasted an image of the temperatures being reported.

image

....Further update, I thought it would be a good idea to round the values using the user defined settings (previously rounding was only applied to MQTT output).

blazoncek:
Excellent. The only remaining thing is compile time options. Remove those.
Modern usermod should not contain any compile time option as anything can/should be configured at runtime.

albarlow:
Okay, so I've figured out how to Make Celsius/Farenheit switchable from the Usermod interface...quite please with that - I've been able to toggle back and forth, after a reboot.

There are now no compile-time-only variables, anything the user needs to change can be done through the Usermod interface.

I've revamped the Readme file so it is, I think, much clearer to read and presents the usermod as a product, not as a rework of a previous usermod.  Let me know if you think I should clarify anything on it.

This probably represents a breaking change for people already using this usermod as I don't know what happens to any of those values previously defined at compile time when the next version gets pushed out.  Does this need to be noted on the next version change log?

blazoncek:
It would be nice to include description of changes in usermod.

albarlow:
Added changelog

blazoncek:
This is now almost perfect. 😄 
The only thing amiss is pins string for configuration and missing pin allocation using PinManager.
The format for configuration item is "...-pin" not "pin-...." or just use "pin". Unfortunately this is not documented anywhere.
The correct name insures other parts of WLED know those pins are in use.

You may want to take a look into 4 Line Display ALT usermod since it also features I2C interface.

albarlow:
Okay, got the pin manager implemented - when I looked at it 6 months ago, I couldn't figure it out but now I was able to make a bit more sense of it...I'm glad that's working.  I'm also glad that it now doesn't have to reboot the device to change pins.

Originally I had named the "pins-sda-scl" as "pins" but thought it would be unclear for user, but I borrowed the idea of having a text box underneath with the pin sequence.

You might want to check I did the const.h and pin_manager.h bit correctly - I think I followed the format correctly but I don't want to break anything if I misunderstood the numbering arrangement.

I've wrapped any strings I could with F() macro, I am from more of a .NET background and I am more used to everything being Strings rather than char arrays...getting there gradually.

blazoncek:
Perfect!

PR #2724, opened 2022-07-19
Update PIR Switch Usermod

albarlow:
I have added Home Assistant Discovery option (toggleable in Usermod screen).

I also created an Occupancy timer that is only used for external reporting.

Future enhancement could be it include the occupancy timer in the JSON reporting display and being able to turn the occupancy sensor on/off.

I was testing this for around 6 months and it was working well.  I've updated it today to replace changes made to the upstream version since.

I amended the MQTT publish procedure but have, so far as I can see, preserved the original functionality.


blazoncek:
Occupancy is not a function of PIR (motion) sensor.
You can configure your HA to mark room as occupied if there is a motion in it within your desired time.
This will not get merged as is.

albarlow:
Just so I'm clear for future...is the objection to the occupancy sensor the fact that it would prospectively be cluttering up an existing Usermod or is it more of a logical arguement that the motion sensor doesn't sense occupancy, it only senses motion?

My rationale for wanting the device to make the decision on occupancy rather than Home Assistant was that, at least when I last worked on the HA side of writing an automation, I don't think you could mark zones as occupied, but it sounds like that may be something that's since been updated, so I'll take a look at that.

I can certainly make a reduced version that just does the HA Discovery to expose the motion state.  That should be a fairly minimal change.  I'll leave this PR open, if that's okay and try and get it done in the next day or so.

albarlow:
Okay...I've reverted my original commit and submitted a new one with only the HA Discovery elements.  Does that look more suitable?

blazoncek:
@albarlow looks like my recent update messed commit history (as 0.14 has been merged to main).
Could you close this PR and open a new one and just include my recent update to `PIR_sensor_switch.h`? This way GH will honour your contribution. If you do not mind that then I can just commit updated PIR myself.

blazoncek:
I am not really sure that retaining HA discovery message is good idea.

albarlow:
Hi, I'm happy for you to go ahead and merge it under your authority - I've had a bit of a look through and with the main branch having moved forward a version I don't want to mess anything up redoing it if you're in a position to just click ok and merge it.

I think the rationale for retaining the discovery message would be that then WLED doesn't need to periodically resend the discovery message (which, given it doesn't need to update after creation, saves network bandwidth) and I'm not _entirely_ sure what happens in Home Assistant if the message isn't retained.  I used the same logic in my proposed modification as I did in the other usermod updates I put in PRs for.

I'm happy to be guided by you on the matter of MQTT retention, as you probably have more experience than me...

If you mean that you're not sure about including the Discovery Message logic at all, I of course defer to your authority on the matter, my use case for this is that I use home assistant, and the difference for a user who maybe knows how to breadboard together a D1 Mini, PIR and Lightstrip is that with the discovery message, the MQTT device appears in home assistant automatically ready for the user to work with.  Without the discovery message, the user has to figure out the YAML to go and edit config files to create a manual sensor, which isn't part of a device and therefore not grouped with other relevant sensors.

Seeing as the Discovery message can be toggled off, for non-home assistant users, it should have no impact on functionality for those who just want the sensor to use internally?

What would have been the ideal solution for my aim in putting the disovery message in would have been to have usermod sensor data be able to appear directly as sensors in the WLED home assistant integration, but I think I can make a good guess why that isn't an option.

blazoncek:
I was merely curious as why retaining a HA discovery message.
Retained messages are permanently stored on MQTT broker and if not used cautiously they may incur additional network traffic as they are always sent to a new subscriber.
They may also be difficult to remove for inexperienced users and cause broker database increase needlessly.

It will allow HA to recognise already set-up WLED devices without them needing to broadcast discovery messages periodically, though.

albarlow:
Sorry - I probably got a bit overzealous in my previous reply...not being sure about whether the retention was the MQTT retention or the retention of the code I was covering my bases :) 

With the device discover messages all being within a homeassistant parent, that would hopefully make it unlikely that users would be subscribing to that key without it being home assistant or having some idea of its workings.

blazoncek:
Closing in favour of #2853

PR #2725, opened 2022-07-19
BH1750 upgrades

albarlow:
Moved the definitions into the main usermods_list.cpp instead of having a section to copy across.

Added Home Assistant Discovery topic for light sensor.  This is toggleable from the usermod menu.


blazoncek:
Can you do the same things you did in BME280 usermod, please?

albarlow:
Just working on this...there are a number of `'D` values that can be defined at compile time, but they can be set in the Usermod menu.  Given that you expressed a preference to remove compile-time variables, should I take those out of the Readme?  Or is the fact that they are set in `platformio.ini`, not in the code itself enough to make it okay?

albarlow:
For the meantime, I have left the information about compile-time options in the readme.  I've implemented the pin manager, F() for strings, added API call for reading illuminance.

blazoncek:
Compile time options are still welcome (to set a default values) but all of them should be able to be overridden using configuration.
Will have a look later of the changes.

blazoncek:
Looks like there is an error in recent update to the usermod.
#2844

PR #2740, opened 2022-08-06
init DS1307RTC only if needed

Matoran:
DS1307RTC is only required by USERMOD_RTC and USERMOD_ELEKSTUBE_IPS.
If we don't use it, we do not want to execute `Wire.begin();` which configure gpio port 21 et 22 as I2C ports. This created a regression as it was possible to use gpio 21 and 22 for relays.


blazoncek:
Hopefully solved in #2737

PR #2746, opened 2022-08-11
Usermod: Ping Pong Clock

schmoizz:
I have created a Usermod to use WLED on a Ping Pong LED Clock as shown here: [Instructables](https://www.instructables.com/Ping-Pong-Ball-LED-Clock/)

- Color of the clock can be configured in the Usermod
- The Clock can be turned on and off in the Settings


blazoncek:
Do not force push commits. It breaks traceability.

PR #2752, opened 2022-08-16
Add ADS1115 usermod

dima-zhemkov:
Add ADS1115 usermod


blazoncek:
This will conflict with upcoming 0.14 version which features Audioreactive usermod with ID 31.
Please re-base your usermod to 0_14 branch or update ID to 32 or greater. Consider other PRs for usermods as well when deciding on ID value.

PR #2760, opened 2022-08-21
Update python dependencies

Aircoookie:


platformio.ini
```diff
@@ -324,18 +325,30 @@ board = esp32dev
 platform = ${esp32.platform}
 platform_packages = ${esp32.platform_packages}
 build_unflags = ${common.build_unflags}
-build_flags = ${common.build_flags_esp32} -D WLED_RELEASE_NAME=ESP32 #-D WLED_DISABLE_BLYNK #-D WLED_DISABLE_BROWNOUT_DET
+build_flags = ${common.build_flags_esp32} -D WLED_RELEASE_NAME=ESP32 -D WLED_WATCHDOG_TIMEOUT=0 #-D WLED_DISABLE_BLYNK #-D WLED_DISABLE_BROWNOUT_DET
```
blazoncek, line 328: We may even remove WDT as a default or provide special binary with WDT enabled for those interested.

andyshinn: I agree. Until this is proven to be a good default maybe it should default ot off.

PR #2787, opened 2022-09-17
Feature/nine additional alexa devices for presets

siggel:
This feature branch adds 9 additional alexa devices to the configured one which are named like the first (up to) 9 presets, such that one can give intuitive commands like "Alexa, switch on Fireworks", "Alexa switch on Rainbow" etc. It has been confirmed working for D1mini by several users in https://wled.discourse.group/t/call-different-presets-from-alexa/306 and https://wled.discourse.group/t/german-sonnen-aufgang-untergang-present-wechsel/5945

However, it misses an on/off switch in the Alexa configuration, maybe someone experienced may want to add this such that the user can decide whether to use the one Alexa device configured there or 9 additional Alexa devices.


Aircoookie (changes requested):
I like this addition in general, however please add in a number option in Sync settings like `Also add devices for the first [ ] presets` (with n 0-9), since I don't believe all users would want all their WLED devices to add up to 10 devices to their Alexa system, it should be a little more fine-tunable. 

Edit: Just re-read you had that idea yourself :) I can add in the option too, no problem! Will merge afterwards :)

siggel:
Hi @Aircoookie 

> Edit: Just re-read you had that idea yourself :) I can add in the option too, no problem! Will merge afterwards :)

That would be great. I had a quick look and saw that it looks easier than expected (not that eeprom reading layout anymore) but still I would not be sure where the parameter would fit better: seems that alexaInvocationName is located below "id" whereas all other alexa stuff is located below "va". So maybe you code it yourself like you prefer.

If you add some html line like 
`Also add devices for the first <input name="AP" type="number" min="0" max="9" required> presets` 
and map it to a parameter like `alexaNumberOfDevicesFromPresets` (default 0) you should change line 25 in alexa.cpp to
`for (byte presetIndex=1; presetIndex<=alexaNumberOfDevicesFromPresets; ++presetIndex)`. That loop will create that number of additional devices (or less if less presets exist).

Aircoookie (approved):

Aircoookie:
@siggel all done now! I also changed the device callback logic a bit to simplify it (go by IDs instead of comparing names), so if you have time it would be appreciated if you can confirm it still works as intended for you too.

siggel:
@Aircoookie 

Test failed:
When I select "0" Alexa does not even find the original alexaInvocationName device. When I select e.g. "2" it works like a charm, i.e. Alexa finds the original AlexaInvocationName device and first 2 (out of 3 available) preset name devices.
When deleting the devices from Alexa and changing back to "0" Alexa will again not find any devices anymore.

I'll take a deeper look into the code.

siggel:
Found it: in case of 0 preset devices you return too early missing to start the server in the end (for the alexaInvocationName device). I'll fix it, test it and if it works push it.

siggel:
Tests with https://github.com/Aircoookie/WLED/pull/2787/commits/73661e9446d1e2810fe9cf72e9e7952bcd977f1e succeeded:
- 0 (out of 3 available) preset devices result in alexaInvocationName device 
- 1 (out of 3 available) preset devices result in alexaInvocationName and first preset device
- 9 (out of 3 available) preset devices result in alexaInvocationName and 3 preset devices

Aircoookie:
Thank you for testing and fixing!

SANGERA2:
Brilliant work guys. Thanks very much for all your hard work. This will be a great feature used by loads of people!

PR #2803, opened 2022-09-30
added new compile #defines

ingDIY:
added #define to easily enable/disable Alexa, Blynk, HueSync, IR into platformio_override.ini using these:
-D WLED_ENABLE_ALEXA
-D WLED_ENABLE_BLYNK
-D WLED_ENABLE_HUESYNC
-D WLED_ENABLE_INFRARED

added warning to inform that the DEFAULT value should not be lower than 250 mA

### DIFF
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -285,7 +285,7 @@
 #ifdef ESP8266
 #define SETTINGS_STACK_BUF_SIZE 2048
 #else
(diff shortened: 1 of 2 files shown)


blazoncek:
Thank you for this PR.

Can you rebase it for 0_14 branch, please.

Other than that ABL setting of 0 is supposed to disable ABL so no warning should be displayed in such case.

PR #2804, opened 2022-10-01
added new #defines settings for usermods

ingDIY:
added the possibility to change the GPIO function. some PCB has external pullups, some use the internal ones.
it can be configured defining for example
-D USERMOD_ROTARY_ENCODER_GPIO=INPUT
in platformio_override.ini

added the possibility to define the defaults settings for auto-save usermod.
Now "preset number", "autosave on boot flag" and "autosave after sec" can be defined (for example in platformio_override.ini)

PS: this is a duplication of PR #2776. I'm really sorry about it, I needed to delete my atuline/WLED fork to use Aircoookie/WLED fork and the PR #2776 was automatically closed.

PR #2776, opened 2022-09-06
added some useful default settings #define

ingDIY:
added the possibility to customize some defaults using #define in platformio_override.ini:
- GPIO function for rotary encoder (needed on my custom PCB).
- change the defaults settings for auto-save usermod.

added a warning for ABL_MILLIAMPS_DEFAULT defined as less than 250 mA (this should help users that build WLED using this setting).

PR #2809, opened 2022-10-02
defines improvement

ingDIY:
Here they are the PR #2776, #2803 and #2804 rebased for 0_14 branch, I hope that now they are OK!

### DIFF
--- a/usermods/usermod_v2_rotary_encoder_ui_ALT/usermod_v2_rotary_encoder_ui_ALT.h
+++ b/usermods/usermod_v2_rotary_encoder_ui_ALT/usermod_v2_rotary_encoder_ui_ALT.h
@@ -278,9 +278,13 @@ public:
       return;
     }
 
-    pinMode(pinA, INPUT_PULLUP);
-    pinMode(pinB, INPUT_PULLUP);
-    pinMode(pinC, INPUT_PULLUP);
+    #ifndef USERMOD_ROTARY_ENCODER_GPIO
+      #define USERMOD_ROTARY_ENCODER_GPIO INPUT_PULLUP
+    #endif
+    pinMode(pinA, USERMOD_ROTARY_ENCODER_GPIO);
+    pinMode(pinB, USERMOD_ROTARY_ENCODER_GPIO);
+    pinMode(pinC, USERMOD_ROTARY_ENCODER_GPIO);
+
     loopTime = millis();
 
     currentCCT = (approximateKelvinFromRGB(RGBW32(col[0], col[1], col[2], col[3])) - 1900) >> 5;
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -314,11 +314,12 @@
     #define E131_MAX_UNIVERSE_COUNT 12
   #endif
 #endif
-
+#define ABL_MILLIAMPS_DEFAULT 200
 #ifndef ABL_MILLIAMPS_DEFAULT
-  #define ABL_MILLIAMPS_DEFAULT 850  // auto lower brightness to stay close to milliampere limit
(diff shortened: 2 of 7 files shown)


wled00/wled.h
```diff
@@ -24,10 +24,18 @@
 //#define WLED_DISABLE_OTA         // saves 14kb
 
 // You can choose some of these features to disable:
-//#define WLED_DISABLE_ALEXA       // saves 11kb
-//#define WLED_DISABLE_BLYNK       // saves 6kb
-//#define WLED_DISABLE_HUESYNC     // saves 4kb
-//#define WLED_DISABLE_INFRARED    // saves 12kb
+#ifndef WLED_ENABLE_ALEXA
```
blazoncek, line 27: These changes are unnecessary.
Features like Alexa, Blynk, Hue, etc. should be compiled by default and only excluded in customised build environment by defining DISABLE defines.

usermods/usermod_v2_rotary_encoder_ui_ALT/usermod_v2_rotary_encoder_ui_ALT.h
```diff
@@ -278,9 +278,13 @@ class RotaryEncoderUIUsermod : public Usermod {
       return;
     }
 
-    pinMode(pinA, INPUT_PULLUP);
-    pinMode(pinB, INPUT_PULLUP);
-    pinMode(pinC, INPUT_PULLUP);
+    #ifndef USERMOD_ROTARY_ENCODER_GPIO
```
blazoncek, line 281: Please add a comment in readme.md

wled00/const.h
```diff
@@ -319,6 +319,7 @@
   #define ABL_MILLIAMPS_DEFAULT 850  // auto lower brightness to stay close to milliampere limit
 #else
   #if ABL_MILLIAMPS_DEFAULT < 250  // make sure value is at least 250
```
blazoncek, line 321: Please add an option to have ABL=0 which should disable ABL and not produce warning message.

blazoncek (changes requested):
Thanks, looks good.
There are just a few issues that need fixing before it is merged.

wled00/wled.h
ingDIY: ok, I removed the code (I did it because in SR branch all of these features are disabled)

usermods/usermod_v2_rotary_encoder_ui_ALT/usermod_v2_rotary_encoder_ui_ALT.h
ingDIY: done!

wled00/const.h
ingDIY: I added a check, I'm not sure that I've understood how the ABL is disabled if `ABL_MILLIAMPS_DEFAULT = 0`
by the way I added a `#if` for the check.

I'm also not sure about the 250mA lower limit value, because in "FX-fcn.cpp" at line 1063 there is a check with a value of 150:
`if (ablMilliampsMax < 150 || actualMilliampsPerLed == 0) { //0 mA per LED and too low numbers turn off calculation`

ingDIY:
dear @blazoncek I commited all the changes, If there is something wrong I'm here :)

wled00/const.h
blazoncek: If you do not understand something then you test all possible variations and observe the behaviour. 😄 

If the limiter is set to <150 it will not use it, effectively lighting LEDs as much as intended. It has nothing to do with ABL_MILLIAMPS_DEFAULT.

blazoncek (approved):

PR #2812, opened 2022-10-03
defines improvement

ingDIY:
removed a #define used for debugging


ingDIY:
I think that something is wrong with merge, I just changed one line😰😭

PR #2818, opened 2022-10-07
please add "Pxie Dust" effect 

karizmas:
https://user-images.githubusercontent.com/23518059/194646725-9535c15e-2671-43dc-91ef-4bc2938f0000.mp4

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -251,7 +251,7 @@ class WS2812FX {
   
   // segment parameters
   public:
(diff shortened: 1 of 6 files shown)


blazoncek:
What is the purpose of this PR?
It uses an old (and obsolete) code and I can't see the effect code.

Aircoookie:
This effect looks cool. If you can point me to the code source and it uses a free license, I would be open to porting it. Please make an issue with type enhancement instead of a PR next time though.

Issue #2112, opened 2021-08-01
UDP sync unreliable


ThatBigPrint:
**Describe the bug**
First issue,

when I have multiple wled modules one or two of them don't sync properly you then need to resubmit the change then it applies this is intermittent.

second issue,
 
the wled app finds the uniquely named modules but after you click the tick it only populates a few but if you manually type in IP address it populates as it should ?

**To Reproduce**
first issue you just need multiple items on one sync id for instance I have 11 

second issue again multiple items on same network I have 15 in this instance 

**Expected behavior**
all wled to sync reliably

app to be able to find nodes every time and populate 

**WLED version**
 - Board: [e.g. ESP32 dev]
 - Version [e.g. 0.13.0, beta] although other nodes I'm having the same issue with are on stable old releases 
 - Format [e.g. Binary, self-compiled] both 

**Additional context**
Anything else you'd like to say about the problem?

Thank you for your help!

huggy-d1:
How many wifi access points do you have?  Are all WLED installs connect to the same wifi AP?  The sync protocol does not route as far as I know since it's a simple UDP broadcast.  WLED simply sends out to whoever is capable of listening, and is listening.  The networking configuration determines whether they are capable of listening.

ThatBigPrint:
they are all on the same network they only produce ap if they can't find the known SSID ? so far they are all on the network and not complaining, the issue lies in reliable sync and pulling into the app on both platforms... they initially worked but since I deleted the list and re searched only one or two come back and its always the same one or two

ThatBigPrint:
a bit more info on this... the Host light seems to drop out for whatever reason, and then comes back around it happens on all of the elements whichever is the host I'm even considering making a hidden dummy hoast? anyway use this link to see a video of the issues https://youtu.be/eC7rsYlEwow

thanks
chris

blazoncek:
WLED Sync uses UDP. That is in itself enough said about reliability on WiFi.
You can slightly improve the situation by telling WLED to transmit sync message twice. Use Settings page.

keck9939:
The sync feature relies on mDNS and UDP broadcasting. Depending on your networking equipment and topology, these two features may or may not be enabled/blocked. I found with the default settings on my network, that enumerating the WLEDS on the network was unreliable and that nodes would periodically drop off the network. After enabling mDNS and making sure UDP broadcasts would get through, everything started to work reliably.

blazoncek:
Mitigation added in #2830

PR #2830, opened 2022-10-14
Add number of UDP retries

Squall-DA:
Add sync option for number of UDP messages to improve WLED sync reliability. Value can be configured from 1 - 255 messages. Anywhere between 5 and 10 messages seems to create a reliable, well synced system on my network. This takes the place of the notify twice option. 

This should help mitigate Issue #2112


blazoncek:
Hi, thakns for the PR. I do have a few questions and general comment, though.

Have you tried/tested this with 60FPS and 32 segments (with segment options syncing enabled, possibly with 4096 LEDs defined)? How much does it add to FPS drop (if any) or other performance hit?

A side note: UDP is inherently unreliable, especially with WiFi. Its intention is low latency. Adding multiple retransmissions will only add to the latency and may not solve bad WiFi coverage/signal strength issues. I do agree that multiple retransmissions will increase the chance of transmitting a packet successfully but it may still not be 100% successful.

If you want 100% successful delivery utilise websockets (you will need to implement client on WLED side yourself). Be mindful that websockets consume precious RAM and WLED currently only supports up to 8 clients on ESP32 (3 on ESP8266).

Please re-base your PR for 0_14 branch to prevent any conflicts with upcoming version.

Squall-DA:
@blazoncek I went through and rebased my branch and pull request and performed those tests you suggested. I do not see performance hit between 1 UDP message or 200 UDP sync messages. I would expect the performance hit to be minimal as the sync messages are only set when changing color and effect related settings and not a constant sync (at least to my understanding of the code). The only minor issue I noticed is that separate WLED devices may not be perfectly synced. I believe for most use cases of WLED sync not perfectly synced is still better than not displaying the correct color/effect (at least it is in mine).

wled00/set.cpp
```diff
@@ -247,7 +247,10 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
     notifyAlexa = request->hasArg(F("SA"));
     notifyHue = request->hasArg(F("SH"));
     notifyMacro = request->hasArg(F("SM"));
-    notifyTwice = request->hasArg(F("S2"));
+
+    t = request->arg(F("UR")).toInt();
+    if (t>0) ubUdpNumRetries = t;
```
blazoncek, line 252: I would limit retries to 10 or some other low number otherwise a user may enter 999999 retries or more.

wled00/wled.h
```diff
@@ -354,7 +354,7 @@ WLED_GLOBAL bool notifyButton _INIT(false);                       // send if upd
 WLED_GLOBAL bool notifyAlexa  _INIT(false);                       // send notification if updated via Alexa
 WLED_GLOBAL bool notifyMacro  _INIT(false);                       // send notification for macro
 WLED_GLOBAL bool notifyHue    _INIT(true);                        // send notification if Hue light changes
-WLED_GLOBAL bool notifyTwice  _INIT(false);                       // notifications use UDP: enable if devices don't sync reliably
+WLED_GLOBAL uint8_t ubUdpNumRetries _INIT(1);                     // Number of UDP messages to send. Increase to increase reliability
```
blazoncek, line 357: Why did you choose default of 1 retry? Current behaviour is without any retries.

wled00/wled_eeprom.cpp
```diff
@@ -102,7 +102,7 @@ void loadSettingsFromEEPROM()
   busses.add(bc);
 
   notifyButton = EEPROM.read(230);
-  notifyTwice = EEPROM.read(231);
+  if (EEPROM.read(231)) ubUdpNumRetries = 2;
```
blazoncek, line 105: Different than default. Intentional and why?

wled00/data/settings_sync.htm
```diff
@@ -127,7 +127,7 @@ <h3>WLED Broadcast</h3>
 Send Alexa notifications: <input type="checkbox" name="SA"><br>
 Send Philips Hue change notifications: <input type="checkbox" name="SH"><br>
 Send Macro notifications: <input type="checkbox" name="SM"><br>
-Send notifications twice: <input type="checkbox" name="S2"><br>
+Number of UDP Messages: <input name="UR" type="number" min="1" max="255" class="d5" required><br><br>
```
blazoncek, line 130: I would change the description to something like "Retransmissions:" or "UDP packet retransmissions:"

blazoncek (changes requested):
Other than what I commented I agree with the  PR.
@Aircoookie I propose to merge it into 0_14 after the changes are made.

wled00/wled.h
Squall-DA: This was an accidental carry over from a previous thought process. Good catch this should be zero.

wled00/wled_eeprom.cpp
Squall-DA: Same issue as this comment https://github.com/Aircoookie/WLED/pull/2830#discussion_r998020179. The intention here is to carry over the previous settings to this new system but it should set retries to one.

wled00/data/settings_sync.htm
```diff
@@ -127,7 +127,7 @@ <h3>WLED Broadcast</h3>
 Send Alexa notifications: <input type="checkbox" name="SA"><br>
 Send Philips Hue change notifications: <input type="checkbox" name="SH"><br>
 Send Macro notifications: <input type="checkbox" name="SM"><br>
-Number of UDP Messages: <input name="UR" type="number" min="1" max="255" class="d5" required><br><br>
+UDP packet retransmissions: <input name="UR" type="number" min="1" max="30" class="d5" required><br><br>
```
blazoncek, line 130: why minimum of 1? perhaps I do not want retransmissions. 😄

wled00/set.cpp
```diff
@@ -249,7 +249,7 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
     notifyMacro = request->hasArg(F("SM"));
 
     t = request->arg(F("UR")).toInt();
-    if (t>0) ubUdpNumRetries = t;
+    if ((t>0) && (t<30)) ubUdpNumRetries = t;
```
blazoncek, line 252: why minimum of 1? perhaps I do not want retransmissions. 😄

blazoncek (approved):
Please change to t>=0 and it is approved.

wled00/data/settings_sync.htm
Squall-DA: Had my blinders on.....

Squall-DA:
@blazoncek / @Aircoookie Do I need to do anything to resolve the conflicting autogenerated file?

blazoncek:
pull updated 0_14 branch or remove generated .h file

PR #2831, opened 2022-10-14
fix: filename uploads

ahadcove:
File name uploads were expecting files to be named starting with a slash. But when uploading using form data you can't have a file name with / in it. So this pr will just add a slash to the file being uploaded if it doesn't include it


blazoncek:
The code was intentional. It was not intended to upload files from just about anything.

Aircoookie (approved):
Hmm, not prepending the `/` is hardly a security measure. If we want to stop unauthorized uploads, PIN is the way to go.
IMO `/edit` (and maybe file uploads in general) should still be blocked by OTA lock, what do you think @blazoncek ?

blazoncek:
It is not a security issue but sort of prevention of accidental overwrites.
Or uploading just about anything and filling up FS.
I think implementation could be better.

As far as OTA lock goes IMO it is best to have it do only two things: actual OTA lock (including Arduino upload) and prevetion of WiFi changes.
PIN should be used to secure everything else.

blazoncek:
I know. I'm quirky.
🤷‍♂️

blazoncek:
The code you updated breaks correct config restore on 0.14.
@ahadcove please implement fix properly for all cases.

ahadcove:
> The code you updated breaks correct config restore on 0.14.
> 
> @ahadcove please implement fix properly for all cases.

Ahh okay. Looking into it now

ahadcove:
> The code you updated breaks correct config restore on 0.14. @ahadcove please implement fix properly for all cases.

Code has been updated to check for both /cfg.json and cfg.json
https://github.com/Aircoookie/WLED/pull/2850

PR #2834, opened 2022-10-16
Async preset saving

Aircoookie:


wled00/const.h
```diff
@@ -25,10 +25,18 @@
   #ifdef ESP8266
     #define WLED_MAX_BUSSES 3
   #else
-    #if defined(CONFIG_IDF_TARGET_ESP32S2) || defined(CONFIG_IDF_TARGET_ESP32C3)
-      #define WLED_MAX_BUSSES 5
+    #if defined(CONFIG_IDF_TARGET_ESP32C3)    // 2 RMT, only has 1 I2S but NPB does not support it ATM
```
Aircoookie, line 28: Not sure if I like this, as it would be no problem to add more busses if they are clocked, analog, or network.
IMO just the changes in bus_wrapper (attempting to make more busses than channels available fails silently) are sufficient for the time being.

Aircoookie (commented):
Otherwise, all approved!

wled00/const.h
blazoncek: I have considered non-RMT (BB) buses but then decided to not include it as it seems that BB version of Esp32 method does not compile.
I've seen references to use Esp826h BB methods but haven't had the time to check it.

blazoncek: In such case it would make sense to allow more buses.
Otherwise thanks for the perspective, I didn't think of it.

Aircoookie: Probably not worth looking into BB, on 8266 it leads to random infrequent resets, and I would assume the same for single-core 32 variants. I think if your use case requires 3+ digital busses, best to go with OG ESP32.

blazoncek: Just checked, C3 also only features only 6 LEDC channels allowing only 2 RGB analog strips.
Would it make sense to still limit max buses to 4?

blazoncek: Increased number of buses and added comments. Do you agree?

Issue #2847, opened 2022-10-24
Restoring cfg.json does nothing


blazoncek:
### What happened?

Restoring configuration (`cfg.json`) using Backup & restore feature of Security Settings page does not restore configuration.

### To Reproduce Bug

Backup configuration on one device and try to restore it on another device.
After the upload, "File uploaded." message appears on screen (instead of "Configuration restore successful. Rebooting."). Clicking Save & Reboot reverts to previous configuration.

### Expected Behavior

Configuration restored.

### Install Method

Self-Compiled

### What version of WLED?

0.13.3

### Which microcontroller/board are you seeing the problem on?

ESP8266, ESP32

### Relevant log/trace output

```shell
The problem lies in line 55 of `wled_server.cpp` (`if (filename == "/cfg.json")`) where filename is incorrectly compared.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

softhack007:
Just also had a look at the code. I think the problem was introduced with this commit:
https://github.com/Aircoookie/WLED/pull/2831 "fix: filename uploads"

Instead of comparing against "/cfg.json" we should also check for "cfg.json". Also it might be better to use 
```c++
if (filename.equalsIgnoreCase("/cfg.json") || filename.equalsIgnoreCase("cfg.json")) {
```

also putting the `doReboot =true` after `request->send(...)` could help, to ensure that the reboot does not happen before thereply was sent.

blazoncek:
You changed handling of original request (by prepending /). Handle that in subsequent callbacks.
filename has to be in lowercase when saving. adjust for that as well.

softhack007:
Im not sure if LittleFS is case-sensitive ... so it might be sufficient to use `filename.equals()`
The WLED web UI only allows for two specific filenames when uploading - cfg.json and presets.json.

Everything else is "extra sugar". To be 100% bullet-proof, you would also need to:
* convert files to lowecase
* strip leading and trailing blanks
* make sure filesname size is below the allowed limit (32? chars)
* reject file upload to forbidden locations
* etc etc
* and maybe present file errors to the user -> `strerror(errno)`.

blazoncek:
@softhack007 yes to all points. /upload was originally created just for the purpose of config an presets upload and then used for ir.json, skin.css, etc.
All of the logic was done in the browser in JS as to reduce ESP strain. Prepending / broke things.

PR #2850, opened 2022-10-24
fix: restoring cfg.json #2847

ahadcove:
This is in response to https://github.com/Aircoookie/WLED/issues/2847

This issue will resolve not being able restore cfg.json using Backup & restore feature within the Security Settings page

### DIFF
--- a/wled00/wled_server.cpp
+++ b/wled00/wled_server.cpp
@@ -39,20 +39,20 @@ void handleUpload(AsyncWebServerRequest *request, const String& filename, size_t
   if (!index) {
     String finalname = filename;
     if (finalname.charAt(0) != '/') {
-      finalname = "/" + finalname; // prepend slash if missing
+      finalname = '/' + finalname; // prepend slash if missing
     }
 
     request->_tempFile = WLED_FS.open(finalname, "w");
-    DEBUG_PRINT("Uploading ");
+    DEBUG_PRINT(F("Uploading "));
     DEBUG_PRINTLN(finalname);
-    if (finalname.equals("/presets.json") >= 0) presetsModifiedTime = toki.second();
+    if (finalname.equals(F("/presets.json")) >= 0) presetsModifiedTime = toki.second();
   }
   if (len) {
     request->_tempFile.write(data,len);
   }
   if (final) {
     request->_tempFile.close();
-    if (filename.indexOf("cfg.json") >= 0) { // check for filename with or without slash
+    if (filename.indexOf(F("cfg.json")) >= 0) { // check for filename with or without slash
       doReboot = true;
       request->send(200, "text/plain", F("Configuration restore successful.\nRebooting..."));
     } else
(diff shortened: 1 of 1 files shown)


blazoncek:
String::operator== should be overloaded. Not sure how exactly. 😁

ahadcove:
I switched to index of based on blazoncek recommendation.
Any issue that you all see with this?

softhack007:
I think a simple test would tell us: Just download cfg and presets, then do a factory reset and restore presets (first), then restore cfg and check if everything is back as expected. OK, only Wifi credentials will be lost because they are in another hidden file.

Aircoookie:
Thank you for the fix, changed it to equals for presets.json since `finalname` will always have the leading slash and it uses less memory this way (not both `/presets.json` and `presets.json` in RAM). And I don't want to over-use F() strings either in file/callback operations.

blazoncek:
@Aircoookie oops!

wled00/wled_server.cpp
```diff
@@ -39,20 +39,20 @@ void handleUpload(AsyncWebServerRequest *request, const String& filename, size_t
...
-      finalname = "/" + finalname; // prepend slash if missing
+      finalname = '/' + finalname; // prepend slash if missing
     }
 
     request->_tempFile = WLED_FS.open(finalname, "w");
-    DEBUG_PRINT("Uploading ");
+    DEBUG_PRINT(F("Uploading "));
     DEBUG_PRINTLN(finalname);
-    if (finalname.equals("/presets.json") >= 0) presetsModifiedTime = toki.second();
+    if (finalname.equals(F("/presets.json")) >= 0) presetsModifiedTime = toki.second();
```
Aircoookie, line 48: Needs no `F()` as also present as ram string in presets.cpp

wled00/wled_server.cpp
```diff
@@ -39,20 +39,20 @@ void handleUpload(AsyncWebServerRequest *request, const String& filename, size_t
...
-      finalname = "/" + finalname; // prepend slash if missing
+      finalname = '/' + finalname; // prepend slash if missing
     }
 
     request->_tempFile = WLED_FS.open(finalname, "w");
-    DEBUG_PRINT("Uploading ");
+    DEBUG_PRINT(F("Uploading "));
     DEBUG_PRINTLN(finalname);
-    if (finalname == "/presets.json") presetsModifiedTime = toki.second();
+    if (finalname.equals(F("/presets.json")) >= 0) presetsModifiedTime = toki.second();
```
softhack007, line 48: I think that string.equals() returns bool, so should not be tested for ">=0"

blazoncek:
🔥 
😁

wled00/wled_server.cpp
Aircoookie: 100% correct 😅

Aircoookie (approved):

ahadcove:
> Thank you for the fix, changed it to equals for presets.json since `finalname` will always have the leading slash and it uses less memory this way (not both `/presets.json` and `presets.json` in RAM). And I don't want to over-use F() strings either in file/callback operations.

Great. Thanks for fixing!

blazoncek (approved):
Looks fine to me too.

PR #2853, opened 2022-10-24
Add HA discovery option to PIR sermod

blazoncek:
Replaces #2724 where I accidentally pushed changes from main branch messing up commit history.

### DIFF
--- a/usermods/PIR_sensor_switch/usermod_PIR_sensor_switch.h
+++ b/usermods/PIR_sensor_switch/usermod_PIR_sensor_switch.h
@@ -164,32 +164,37 @@ private:
   }
 
   // Create an MQTT Binary Sensor for Home Assistant Discovery purposes, this includes a pointer to the topic that is published to in the Loop.
-  void _createMqttBinarySensor(const String &name, const String &topic, const String &deviceClass)
+  void publishHomeAssistantAutodiscovery()
   {
-    StaticJsonDocument<600> doc;
-    
-    doc[F("name")] = String(serverDescription) + " " + name;
-    doc[F("state_topic")] = topic;
-    doc[F("payload_on")] = "on";
-    doc[F("payload_off")] = "off";
-    doc[F("unique_id")] = String(mqttClientID) + name;
-    if (deviceClass != "")
-      doc[F("device_class")] = deviceClass;
-    doc[F("expire_after")] = 1800;
+    if (WLED_MQTT_CONNECTED) {
+      StaticJsonDocument<600> doc;
+      char uid[24], json_str[1024], buf[128];
 
-    JsonObject device = doc.createNestedObject(F("device")); // attach the sensor to the same device
-    device[F("name")] = serverDescription;
-    device[F("identifiers")] = String(F("wled-sensor-")) + mqttClientID;
-    device[F("manufacturer")] = "WLED";
-    device[F("model")] = F("FOSS");
-    device[F("sw_version")] = versionString;
-    
-    String temp;
-    serializeJson(doc, temp);
-    DEBUG_PRINTLN(topic);
-    DEBUG_PRINTLN(temp);
+      sprintf_P(buf, PSTR("%s Motion"), serverDescription); //max length: 33 + 7 = 40
+      doc[F("name")] = buf;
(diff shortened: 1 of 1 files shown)


blazoncek:
@albarlow can you verify this still works as intended as I do not own HA?

albarlow:
The code is looking right but I'm not getting the message publishing.  I deleted out the existing message and then rebuilt the firmware from the pir-ha-discovery channel and it's not published the message.  I'll take a further look tomorrow in case its something at my end.

blazoncek:
I moved the initial broadcast into connected() as to not need retained message.
If the message is retained the broadcast can go into setup() instead.

blazoncek:
@albarlow I have it correctly publish (when MQTT broker connection is established) but comparing the topic to other HA autodiscovery topics (in Multi Relay and Temperature usermods) this seems off as it publishes on `wled/#` global topic instead of `homeassistant/#`

https://www.home-assistant.io/docs/mqtt/discovery/

albarlow:
I have tried again this evening and I can confirm that for me this is also working now, my existing entity in home assistant was able to detect this as the same device and has linked it the brightness sensor usermod output so they all appear in the same device (pictured below with BH1750 usermod output).

I like the use of linking it to onMQTTConnect - that feels better than having it on setup.  Is that something I can port to the other two usermods I've been working on as I don't recall seeing that as something to harness previously(?)

image

blazoncek:
Yes, you can. I've added (but not documented) the feature when implementing Multi Relay usermod.

blazoncek:
I've also had this verified independently by @srg74 and he also confirmed it working. So I'll merge this PR and close the other.

Sangoku:
Sorry, but this seems to break HA integrations. 
I have a fresh installation on esp01s 
platform ini overide: 

[platformio]
default_envs = WLED_motion

[env:WLED_motion]
board = esp01_1m
upload_speed = 74880
upload_port = COM7
platform = ${common.platform_wled_default}
platform_packages = ${common.platform_packages}
board_build.ldscript = ${common.ldscript_1m128k}
build_flags = ${common.build_flags_esp8266}
; *********************************************************************
; *** Use custom settings from file my_config.h
  ;  -DWLED_USE_MY_CONFIG
; *********************************************************************das
;
;
; *** To use the below defines/overrides, copy and paste each onto it's own line just below build_flags in the section above.
; 
; disable specific features
;  -D WLED_DISABLE_OTA
 -D WLED_DISABLE_ALEXA
 -D WLED_DISABLE_BLYNK
 -D WLED_DISABLE_HUESYNC
 -D WLED_DISABLE_INFRARED
;  -D WLED_DISABLE_WEBSOCKETS
; PIN defines - uncomment and change, if needed:
  -D LEDPIN=2
;   -D BTNPIN=0
;   -D TOUCHPIN=T0
;   -D IRPIN=4
;   -D RLYPIN=12
;   -D RLYMDE=1
; digital LED strip types - uncomment only one ! - this will disable WS281x / SK681x support
;   -D USE_APA102
  ; -D USE_WS2801
;   -D USE_LPD8806
; PIN defines for 2 wire LEDs
   ; -D CLKPIN=0
   ; -D DATAPIN=2
; to drive analog LED strips (aka 5050) hardware configuration is no longer necessary
; configure the settings in the UI as follows (hard):
;   for the Magic Home LED Controller use PWM pins 5,12,13,15
;   for the H801 controller use PINs 15,13,12,14 (W2 = 04)
;   for the BW-LT11 controller use PINs 12,4,14,5
;   
; set the name of the module - make sure there is a quote-backslash-quote before the name and a backslash-quote-quote after the name
  -D SERVERNAME="\"WLED-Bathrom\""

; enable PIR usermod
  -D USERMOD_PIRSWITCH
  -D PIR_SENSOR_PIN=0

; wifi sid:
  -D CLIENT_SSID="\"111\""
  -D CLIENT_PASS="\"2222!!!\""

Something is broken...  im cant get the entry to show up. 

MQTT does not publish the motion topic to..  no matter what i try.

blazoncek:
Please use a pre-compiled binaries from @srg74 and report back.

Sangoku:
> Please use a pre-compiled binaries from @srg74 and report back.

as far i see he has only esp32 this is a esp01s. Or am i missing it?

PR #2856, opened 2022-10-29
Effect ID's back to 0.13 numbering

ewoudwijma:
Needed for sync between versions and allow 0.13 presets to work in 0.14

This is a high prio PR as we need to avoid people saving incompatible 0.14 presets, so the sooner the better.


Aircoookie:
Agreed on SR-0.13 IDs, they should make the switch from SR-0.13 to vanilla 0.14 a lot easier.

However, I see no issue with assigning the IDs of removed effects (Police, Candy cane...) to new effects.

ewoudwijma:
No problem with that indeed! As long as they are used for new effects and not sr effects which already had an id (as is the current situation). So they are gaps for now but can be filled when new effects are created.
(Gaps work fine)

ewoudwijma:
Background: effects where renumbered as the plan was earlier to add them conditional, but that plan luckily changed so the reason for renumbering also is gone

blazoncek:
I disagree as several SR effects were useless or not implemented and were removed while transitioning (there are a few more I consider duplicates). And having effect sync between SR effects and AC is (almost) pointless since both SR or AudioReactive feature audio sync which is better option IMO as non audio devices will look odd without actual audio playing audio effects. Adding a version info to UDP sync is easier and far less disruptive.

Rearranging effects in this way will make a few gaps, needlessly consuming precious RAM on ESP8266 without added benefit.

0.14 code has conditional 2D support (2D effects were grouped for that purpose) and SR fork is still maintained. Keeping effect IDs synchronised between forks is bound to fail at some point unless SR embraces upstream. And the main idea behind making AR usermod was to make audio development independent of main WLED code which is not the case with SR.

I recommend to add mapping to SR or make a script/app to replace/update preset IDs in `presets.json`. Something similar has been done by @scottrbailey if memory serves me well.

ewoudwijma:
This change will make sr dev team and sr userbase happy, and is the right thing to do. the disadvantage of having about 10 gaps is minimal, also because new effects may be created which can consume these slots.

Aircoookie:
Thought for a bit about this, and I am a little conflicted, as I like the current organization more (that is, all 2D effects have consecutive IDs, at least until more 2D FX are added) and an extra use of 100b for the 12 extra slots is also not insignificant.
Still, I believe having compatibility is overall worth the compromise here. Would you also adjust the order `addEffect()` is called in the bottom of FX.cpp?

blazoncek:
There was a question somewhere why `"fx":"Blink"` doesn't work. And I almost laughed, now I am considering if I was wrong.
Legacy can be a curse as it can also be a boon.

When a decision has to be made about future I always remember what Apple did with floppy disks. Ahead of time and many objected, now nobody cares and we live floppy free.

Presets can be adjusted with just a few clicks, once and for all. And I bet that 99.9% of users have less than 20 presets.

There is another option that I implemented at the beginning. Inserting effects with `addEffect()`. It was removed once @ewoudwijma implemented sound simulation as all effects form SR were suddenly displaying something even without audio. We can still bring that back but that would require to move all sound effects to usermod (for carity). Awkward IMO.

Third option could be to implement (optional) effect sorting in the `beginStrip()` depending on the usermod settings page. I.e. "Force SR effect order".

What do you think?

softhack007:
Well let me try to tell the story from a user's perspective.

Some time ago, there were 2D and soundreactive effects in SRWLED only. Then, with our joint effort to bring "upstream" and "soundreactive" together, upstream now also has these features. So people now may choses between several versions that have 2D and sound-controlled effects, and these efects have very similar names across all forks. 

Switching between SR and "MoonMod" fork works easily, however people will get into trouble when they go from SRWLED to "upstream WLED" 0.14, because effects IDs are different, and their previous (precious) presets will look weird.


I am expecting a load of problem reports and support requests if the situation stays like this. 
IMHO the PR from ewowi is a good approach, because none of us needs write on-the-fly conversion features to ensure compatibility. Maybe around 100 **bytes** of RAM for some gaps (that could be filled later) are a price we can easily pay, as it will allow us to concentrate on coding instead of tedious discussions with users.

I know it was initially designed differently, and actually when Blaz first mentioned the change of effect IDs, I did not understand consequences. Ewowi is right with saying that we should do something now, so that people do build up presets which are incompatible between forks.

blazoncek:
After the discussion with @Aircoookie I have no more objections regarding numbering but I would still request reduction of effects that can be achieved by mirroring/reversing segments (Gravimeter & Gravcenter, etc) or are sub par in quality.

ewoudwijma:
Thanks guys for the discussions on this topic, I understand the different viewpoints and really do appreciate the work of blaz on optimizing the code and minimizing the memory impact and I am super impressed of the results of this. This change is a small trade off for user support and also the amount of questions we can expect as Softhack says. 
I support Blaz his suggestion of trying to remove duplicates. Only thing is this should be the result of a discussion between users. For me the 'Harry Baas' effects is a point of reference, see https://github.com/MoonModules/WLED/wiki/Merge-history#october-20-2022. I spend a lot of time together with Harry to realize this and is really beautiful so the effects they are using should not be removed. Also note these effects are not running as it should on 0.14 so there is still work to be done to make them work. I have that on my list and will send PR requests to upstream in the future.

blazoncek:
Of all the single note effects I tested the only ones that would make the cut from my side are:

- Gravimeter (other Grav... can be recreated using this)
- Juggles
- Matripix
- Midnoise (but has visual issues)
- Noisefire (low quality)
- Noisemeter or Pixels (they look almost identical)
- Pixelwave
- Plasmoid (? not really sure)
- PuddlePeak
- Puddles
- RipplePeak

Of the double note ones:

- Blurz
- DJ Light
- Freqmap
- Freqmatrix
- Freqpixels
- Noisemove
- Rocktaves
- Waterfall

I have no objections to 2D effects.
We need to come to a final decision at some point (and not that far in the future) so please find some time to resolve this.

ewoudwijma:
Hi @blazoncek, could you move this to another issue as this is a a separate discussion.

@Aircoookie , can you merge this PR as we came to a conclusion.

Aircoookie:
Yes, I will however add a comment to the unimplemented effects (and maybe change the initialization order in FX.cpp to be continuous again)

ewoudwijma:
Fair enough! Now I realize what you mean by order ;-), of course makes sense !

Aircoookie (approved):

wled00/FX.cpp
```diff
@@ -7500,20 +7464,76 @@ void WS2812FX::setupEffectData() {
...
   addEffect(FX_MODE_BLURZ, &mode_blurz, _data_FX_MODE_BLURZ);
+
+  addEffect(FX_MODE_FLOWSTRIPE, &mode_FlowStripe, _data_FX_MODE_FLOWSTRIPE);
+
+  addEffect(FX_MODE_WAVESINS, &mode_wavesins, _data_FX_MODE_WAVESINS);
   addEffect(FX_MODE_ROCKTAVES, &mode_rocktaves, _data_FX_MODE_ROCKTAVES);
+
+  // --- 2D  effects ---
+#ifndef WLED_DISABLE_2D
+  addEffect(FX_MODE_2DSPACESHIPS, &mode_2Dspaceships, _data_FX_MODE_2DSPACESHIPS);
```
blazoncek, line 7495: Why weren't these moved to the appropriate place above as others were?
The place here suggests that their ID would be higher than `FX_MODE_ROCKTAVES` for example.

The same goes for other 2D effects. They should be placed in their corresponding position since if we ever change `addEffect()` function to ignore ID (which would be best to once and for all remove the necessity to update `#define`s each time an effect is added) their IDs are going to change. Don't forget that adding effect can (and will) happen in usermods as well without changing `#define`s in FX.h.

ewoudwijma: Because we then ended up with a lot of #ifndef WLED_DISABLE_2D I grouped all the 2D's and non 2D's together.

Best is if in some way the effect ID is not also used as index in the _mode array, but that has a performance drawback (probably)

PR #2861, opened 2022-11-03
Automatically set PC Mode if unset

Randomblock1:
I added a short JavaScript snippet that enables PC mode when the browser's User Agent isn't mobile and PC mode is unset. So you can override it and it will respect that.

[Mozilla says](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent#mobile_tablet_or_desktop), "we recommend looking for the string Mobi anywhere in the User Agent to detect a mobile device." So that's what I did.


blazoncek:
Looks like this is working as intended. Approved.

PR #2865, opened 2022-11-04
Sync Interfaces settings: hide sections for disabled features

softhack007:
This is a fix from SRWLED 0.13.3. It hides sync settings for features that were disabled at compile time.

- Alexa
- MQTT
- Hue Sync
- Blynk

Background: SR WLED is by default build with some of these features disabled. This is not clear to users, because the setting pages were still there.

This patch puts the related html parts between `<div>` elements that can be hidden with a simple javascript call.

Result looks like this
unknown


blazoncek:
You can optimise that to use CSS classes.

CSS
```css
.hide { display: "none"; }
```
JS
```js
function toggle(el) { gId(el).classList.toggle("hide"); gId('No'+el).classList.toggle("hide"); }
```
HTML
```html
<div id="Blynk"> ... </div>
<div id="NoBlynk" class="hide"> ... </div>
```
xml.cpp
```c++
#ifdef WLED_DISABLE_BLYNK
oappend("toggle('Blynk');");
#endif
```

softhack007:
Thanks, good idea. I'll try that in the next few days 👍 and update PR accordingly.

blazoncek:
IMO ready to rock. 😄

blazoncek (approved):

blazoncek:
> Diff contains other recent commits for some reason, looks good to me too though :)

Yes, pulled from @softhack007 repo into WLED and pushed back. Sorry.

softhack007:
> > Diff contains other recent commits for some reason, looks good to me too though :)
> 
> Yes, pulled from @softhack007 repo into WLED and pushed back. Sorry.


I was already wondering what github is doing ... nevermind.
Thanks @blazoncek and @Aircoookie, I'm currently very busy so your help to finalize my PR is very welcome 😄

PR #2870, opened 2022-11-06
UDP Network debugging

Aircoookie:
Merges functionality of #2157 as a compile time option (thank you for streamlining it by removing settings @blazoncek !)


blazoncek:
I think it is ok to merge this. 😃

PR #2876, opened 2022-11-09
playback of .tpm2

constant-flow:
Enable playback of animation recordings via presets (tpm2 right now, extensible to other formats).
Animations can be stored in flash `/edit` or on SD card (see other upcoming PR)

## Summary by CodeRabbit

* **New Features**
  * Playback Recordings usermod enables playback of previously recorded LED animation sequences stored on LittleFS or optional SD card storage
  * Configure playback behavior including repeat options and frame rate overrides via preset API commands with flexible segment targeting

* **Documentation**
  * Added comprehensive setup and configuration guide for the Playback Recordings usermod


ElToberino:
@constant-flow:
Thanks for your work. Everything is running as intendend!

I have a question/suggestion concerning the 2D mode coming up with V0.14:
If I define a matrix (let's say WxH -> 30x5 eg.) in the settings, the other realtime modes (tpm2.net, artnet etc.) seem to handle the segment handling in another way: Entering the realtime mode, the whole matrix plays the data as it is shown and sent out in Jinx!.
Running your usermod in WLED 2D-mode, the data is only played in the first row of the matrix (30 Leds) and so the animation is not played correctly. Redefining the WLED settings to 1D (150 Leds), everything works fine.

So my question: 
Would it make sense to make your usermod handling the tpm2 data as the other realtime modes do?
I read your intention was to make the file playable on a certain segment, which is also a nice feature.
Perhaps it would be possible to make these two modes optional with via the json setting?

blazoncek:
It uses `setRealtimePixel()` so everything should be ok.
`setRealtimePixel()` is not aware of 2D setup or segments (beyond main segment).

ElToberino:
I just can tell you what I noticed:
With a 2D-Setup (30x5) the file is only played in the first row. Changing the setup to 1D with 150 Leds, everything works as it should. All other realtime modes don't show this behaviour and work fine with 2D-Setup.

martenbiehl:
Is the 2d setup something that prohibits this user mod from being merged? I would really like to use it. Especially the segments feature is important for my use case.

ElToberino:
I've just tested your mod with my matrix and faced an issue:
My matrix has a size of 39x22 Leds; these 858 Leds are driven on pin 16 and pin 4 of my esp32. (Each pin drives 429 Leds).
Playing the tpm2 files (from flash and from SD) I noticed that the animation is played much slower than it should do. The files are recorded with 25 fps. Playing with the default setting of 25 fps they are played far too slow. Increasing the fps rate via Json seems to help, but I can never reach the proper speed. I tried many different fps rates, but the maximum playback speed I could reach was about 68% of the intended speed.

So I wonder if this a limitation of the ESP32 or WLED or if there is a code issue. (My tpm2 records are fine; I also habe tested it with another tpm2 player where everything is running properly.)

constant-flow:
So far the implementation is rather naïve and doesn't do any kind of double/triple buffering. Loading data for so many LEDs is something I didn't test and cannot test (no hardware at hand). All data is read just-in-time from the file. Did you load the data from SD or from flash, I bet SD is slower than the Flash as well. I think if required the performance can be boosted, but at some point there will be definitely a limit by the hardware itself to playback that fast, latest when the animation doesn't fit into memory or when it's too many LEDs to get the info for.

constant-flow:
@ElToberino concerning the 2D setup. I located the origin of the issue. My code relies on the `stop` attribute of a segment, which in fact is `stopX`. So I try to find out if there is a clever way (maybe already implemented way) to support 2D and segments at the same time.

ElToberino:
@constant-flow: Thanks for investigating the 2D setup!
Concerning the slow playback speed I also think this is caused by the large number of 858 leds though - in principle - the ESP32 should be able to handle this number easily without any problems. This issue also occurs when reading the file directly from ESP's flash memory, so it is not sd card related. As you said, it really could be caused by missing buffering but, unfortunately, I don't have sufficient skills yet to implement and test a buffering.
Thanks for your work and your feedback!

blazoncek:
Try adjusting WLED FPS in LED settings (advanced part). Start with 25FPS.

ElToberino:
@blazoncek:
Thanks, but that's what [I've already done.](https://github.com/Aircoookie/WLED/pull/2876#issuecomment-1319273438)

constant-flow:
@ElToberino you may want to try the [development branch of this PR](https://github.com/constant-flow/wled/tree/feature/tpm2-playback) I just commited a change that hopefully fixes the 2D Segments issue. More than welcome to get feedback if it works for you. I will start to look into buffering and timing today.

ElToberino:
@constant-flow:
Thank you - that sounds great. I'll test the 2D-fix this evening and will report.

ElToberino:
@constant-flow:
I've just tested your changed code - it really works fine with 2D setup now. Thanks a lot for your work!

ElToberino:
Hi, @constant-flow!
I don't know if you want/need this feedback, but I tried the current version (development branch - I know it's marked as WIP) of your usermode with the current 0.14 beta but didn't get it running. 

The code of this draft in here also works fine with the current version (of course showing the issue of far too slow speed with my 858 leds.)

Best regards!

ElToberino:
It's not stale.
I still hope it's going on in here...

constant-flow:
@ElToberino do you remember what you meant with your [post](https://github.com/Aircoookie/WLED/pull/2876#issuecomment-1405106048).

From what i read it is (referencing this commit graph), you mean commit `0df38e67` on `pr/tpm2-playback` was working and you tested `7ab6cc37 (added buffering (WIP, not tested))` on `feature/tpm2-playback` and it didn't work?

Can you recall what specifically didn't work? Not compiling, No playback, Wrong speed, Wrong display, or simply still too slow?

image

Looking at the graph, it seems like I branched off to far off from the working state. Merging the buffering part into the functional version creates way to many conflicts. So probably a rewrite is required. Hooray 😞

constant-flow:
@blazoncek @Aircoookie @ElToberino 

OK, I was able to manually merge the changes of my buffering code (7ab6cc37) and the code of this PR (0df38e67).
You can find it in my fork [pr/tpm2-playback-dev](https://github.com/constant-flow/WLED/tree/pr/tpm2-playback-dev).

I literally didn't do any changes other to the code other than the changes from the branches.
✅ Playing back TPM2 from flash memory does work for me.

# 🫵 How to test (please do 🙇 )
## Build and Deploy
Select and build the `[env:esp32dev_playback]` build env and adjust your upload mode. 

You have three option (you probably know 🤓 but I add it for completeness)
1. **OTA:** Adjust the IP if you use OTA
1. **No-OTA:** uncomment the `upload_protocol` & `upload_port` lines in this env and flash via USB.
1. **DIY:** Add the buildflag `-D USERMOD_PLAYBACK_RECORDINGS` to your env. (You don't need to switch to `env:esp32dev_playback` here)

## Store playback file
So please test this first you need to upload a TPM2 (e.g. `record.tpm2`) file via the edit route, for me `http://192.168.178.42/edit`

... after uploading it should look similar to this:

image

## Create preset
Then setup a preset, according to your filename:

> Example preset:

```
{"playback":{"file":"/record.tpm2","fps":10}}
```

Hit **Save** 

image

## Play recording
Click the preset, to play it

# Please reply
if this works for you, only then, i'd check how much work FSEQ would be to implement.
I guess I'd then also create a new PR.

blazoncek:
> you need to upload a TPM2 (e.g. `record.tpm2`)

@constant-flow can you provide a sample TPM2 file? I have none.

constant-flow:
Here is a [simple file]-(https://discord.com/channels/473448917040758787/900781399244681296/900783478625411183) living on the discord.

Please try with a segment defined, as it relies on the segments

blazoncek:
@constant-flow I've checked your fork, tpm2 branch and although the code is clean and well written I would recommend different approach.

Please exercise a new possibility in 0.14 to create custom effects and use that to:
- display recording (it will automatically use segment)
- use segment name for recording file (user can easily switch recordings, no special fiddling with presets)
- use `SEGENV.call==0` as a trigger to load recording

I would also like to see in the code:
- add enable/disable config item

I do not know anything about TPM2 but was wondering if it does not have FPS field? Does it?

blazoncek:
Also tested the usermod and more often than not I get a crash. Sometimes it will play *something*. :)
[playback_crash.txt](https://github.com/Aircoookie/WLED/files/12208930/playback_crash.txt)

ElToberino:
Hi, @constant-flow,
I've just seen you're back - this is really nice. 
I'll try to find some time this week to test you're fork and I will report after that.

ElToberino:
Having compiled and tested your code I have to report some problems unfortunately:
With my 768 leds matrix (96x8) I always got a crash when calling the playback-preset. Reducing the number of leds to 512 (64x8) I faced the same malfunction.
So I reduced again to only 256 leds (32x8) - and the playback worked as intended.

Additional setup information: Tested with two different files recorded with Jinx! (size: 508 kb and 86 kb). WLED configured as a 1-D-strip, matrix patching was done in Jinx!).

As it was before in your first PR (before buffering was added) there seem to be problems with a larger number of leds, though 768 leds should be no problem for an esp32 (with tpm2.net via Wifi everything works fine.)

softhack007:
@coderabbitai please review.

Additionally we need a few summary lists:
* noteworthy good practices in this PR
* list of related PR (open or closed) that also tried to add fseq playback or SDCard support
* major changes to make this PR compatible with the current code in `main`

PR #2883, opened 2022-11-15
2d mapping

ctjet:


blazoncek:
Please remove unnecessary changes in platformio.ini and sample file. Use platformio_override.ini instead.
There is also no need for "copy" file.

I do have a few questions:

* How do you intend to simplify user entry if his/her matrix only consists of fixed size panels? Manually editing JSON data seems tedious.
* ArduinioJson arrays consume enormous amounts of RAM. How do you handle (let's say) 1024 pixel data reading/storing to/from config?

My suggestion would be to *add* your mapping to the existing code instead of replacing it.

PR #2885, opened 2022-11-15
Fix ledmap dropdown in presets

ewoudwijma:
The ledmap drop down in presets is currently not working, it should show the ledmap included in the api and if dropdown is updated, the api should also be updated. And if set to non should be removed.
Also when set to none, the customMappingTable should be cleared.

Removed also the 0 as non functional dropdown option, and shows ledmapx.json instead of only x.

Note: json files should be named ledmap1.json until ledmap9.json. Ledmap0.json will not be read but is used as "no ledmap". 

This is fixed in this pull request

### DIFF
--- a/wled00/presets.cpp
+++ b/wled00/presets.cpp
@@ -13,7 +13,7 @@ static volatile byte callModeToApply = 0;
 static volatile byte presetToSave = 0;
 static char quickLoad[3];
 static char saveName[33];
-static bool includeBri = true, segBounds = true, selectedOnly = false, playlistSave = false;;
+static bool includeBri = true, segBounds = true, selectedOnly = false, playlistSave = false;
 
 static const char *getName(bool persist = true) {
   return persist ? "/presets.json" : "/tmp.json";
@@ -208,6 +208,7 @@ void savePreset(byte index, const char* pname, JsonObject sObj)
       playlistSave = true;
     }
   }
+  loadLedmap = sObj["ledmap"];
 }
 
 void deletePreset(byte index) {
--- a/wled00/util.cpp
+++ b/wled00/util.cpp
@@ -504,7 +504,7 @@ um_data_t* simulateSound(uint8_t simulationId)
 
 
 void enumerateLedmaps() {
-  ledMaps = 1;
+  ledMaps = 0;
   for (size_t i=1; i<10; i++) {
     char fileName[16];
     sprintf_P(fileName, PSTR("/ledmap%d.json"), i);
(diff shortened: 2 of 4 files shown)


wled00/presets.cpp
```diff
@@ -208,6 +208,7 @@ void savePreset(byte index, const char* pname, JsonObject sObj)
       playlistSave = true;
     }
   }
+  loadLedmap = sObj["ledmap"];
```
blazoncek, line 211: IMO this is taken care of in line 373 of json.cpp.

wled00/util.cpp
```diff
@@ -504,7 +504,7 @@ um_data_t* simulateSound(uint8_t simulationId)
 
 
 void enumerateLedmaps() {
-  ledMaps = 1;
+  ledMaps = 0;
```
blazoncek, line 507: As stated above, ledmap 0 always *exists*. If there is no file "ledmap.json" it means no ledmap is used, use default (ever increasing) LED order (AKA clear current ledmap).

blazoncek (changes requested):
I have not recently tested ledmap loading so I cannot comment if some other change in the code broke functionality. Will do, once I get the time.

The changes proposed work against original desire of ledamp 0 clearing current ledmap (as `ledmap.json` is used as a default ledmap (loaded at boot) if present). This is due to legacy reasons (there is no `ledmap0.json`).

I cannot approve current implementation as I see it flawed.

blazoncek:
I have tested ledmap loading (from presets) in current main (0.14, build 2211130) and it works as expected/designed.
No changes are necessary.

blazoncek:
> Next from the comments you made in the code which I will follow up I still think the ui is not intuitive, so I will update this pr, up to you to accept or reject then.
> 
> Don't you think it's a bug that the drop-down is not updated in the api?

No need to do UI changes as I am adding clarification to that and also a possibility to change ledmap on its own without saving it in preset 1st.
No, I do not think it is a bug. There is no need for that information in UI.
Screenshot 2022-11-16 at 19 26 02
Screenshot 2022-11-16 at 19 26 15

MoonModules:
Cool, let me know when it is ready.
Will you also update https://kno.wled.ge/advanced/mapping/ to describe the dropdown as I still do not understand why you can update that dropdown if nothing happens, hopefully that is part of your fix

MoonModules:
> I have tested ledmap loading (from presets) in current main (0.14, build 2211130) and it works as expected/designed.
> No changes are necessary.

Haha, no changes necessary but you copied half of my proposed changes ;-)

blazoncek:
Just to make you happy. 😜

MoonModules:
Using ledmapsX.json in the dropdown instead of only X ;-)

PR #2889, opened 2022-11-19
feat(wifi): add compile time configurtation

jkoelker:
Add `WLED_AP_SSID` and `WLED_AP_PASS` defines to allow configuring the SoftAP SSID and Password at compile time. Default to existing values.

Add `WLED_AP_SSID`_UNIQUE flag to append the device portion of the mac address to `WLED_AP_SSID`.

Example use of all flags (note the quoting to preserve "stringification"):

```
build_flags =
    -D WLED_AP_SSID='"MyAP"'
    -D WLED_AP_PASS='"MyPassword"'
    -D WLED_AP_SSID_UNIQUE
```

### DIFF
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -9,6 +9,7 @@
 
 //Defaults
 #define DEFAULT_CLIENT_SSID "Your_Network"
+#define DEFAULT_AP_SSID     "WLED-AP"
 #define DEFAULT_AP_PASS     "wled1234"
 #define DEFAULT_OTA_PASS    "wledota"
 
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -364,7 +364,13 @@ void WLED::setup()
 #endif
   updateFSInfo();
 
-  strcpy_P(apSSID, PSTR("WLED-AP"));  // otherwise it is empty on first boot until config is saved
+  // generate module IDs must be done before AP setup
+  escapedMac = WiFi.macAddress();
+  escapedMac.replace(":", "");
+  escapedMac.toLowerCase();
+
+  WLED_SET_AP_SSID(); // otherwise it is empty on first boot until config is saved
+
   DEBUG_PRINTLN(F("Reading config"));
   deserializeConfigFromFS();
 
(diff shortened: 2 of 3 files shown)


blazoncek:
As this may be a nice touch for some, it will cause troubleshooting for most users a pain if used.
Not to mention that any docs/FAQs will not reflect user's setup.

Just a thought.

jkoelker:
By default everything is as is, and any user that is using the binary releases would be using the default values. If someone chooses to compile themselves and set the values.

I have just under a thousand devices deployed and have been maintaining a private fork to support them, and this has made it insanely more easy to identify which devices are not connected, as well as to bootstrap more than one device at a time. I figured I should start cleaning up and contributing back upstream.

If this is accepted I planned on updating the advanced documentation, I could put a big warning there to prevent people from using it if they don't understand it.

blazoncek:
It is not an issue (usually) for people that compile themselves, but regular people will install just about any binary they get from internet, not knowing what it brings and then complain (or rant) in support channels.

I do welcome the idea and the implementation is slick but I would require acompanying WLED-docs PR as well prior to merging this, just to be on the safe side.

If I remember it correctly there was a discussion on Discord about unique AP SSIDs but the thing never took hold.

Aircoookie:
LGTM. I don't really expect end users to use this, and if a mass deployer like @jkoelker uses it, they will have accompanying docs for their device that matches what they have set.
IMO we shouldn't offer support at all to users with binaries not by Serg or us, since there is always the possibility source is modified in OSS.

Aircoookie:
What I would still add is a compile-time check that throws `#error` if `WLED_AP_PASS` is changed, but `WLED_AP_SSID` remains the default.
I do see an issue with the AP being named `WLED-AP` but having a nonstandard password. Beyond that, if the SSID is also changed to reflect that it is not a default build, i don't care.

blazoncek:
There may be an issue in identifying "default"/official builds.
Still, it is always possible to instruct users to flash official firmware when debugging/identifying issues. The problem may be with users having devices in hard to reach locations.

Nevertheless I am in favour of merging this when docs are updated to reflect a possibility of different AP SSID and/or password.

jkoelker:
Thank you all for the discussion and review!

The `#error` is a great idea. I added that to check that if either the password or if the unique flag was set that the SSID also should be set.

I have also changed the unique flag to follow the format of the other two defines `WLED_AP_SSID_UNIQUE` (it was `WLED_UNIQUE_AP_SSID`).

The docs update is at https://github.com/Aircoookie/WLED-Docs/pull/91.

Aircoookie:
Thank you for the changes and the very detailed docs!
Custom SSID with default password as well as unique SSID starting with `WLED-AP` are ok in my opinion, so I removed those two checks. Just default SSID with custom password is not good.

PR #2891, opened 2022-11-19
Refactored DMX effect mode + new segment controls

mxklb:
# Abstract

This enables WLED to control segments via DMX with `Effect` modes ++ introduces some additional refactoring.

1. Added missing third color RGBW channels.
2. Added missing `effect option` channel (_mirror_ and _reverse_).
3. Existing `Effect` mode is clearly divided into 15|18 DMX channel modes (+ whites).
4. New `Effect Segment` modes are introduced to control segments by 15|18 DMX channels.
5. Hidden "feature" `Presets` is no longer hidden & less restrictive according to its DMX start address. 

# For documentation

- Existing `Effect` mode now controls 15 channels (including third color RGB, without whites):

    `[bri,effectCurrent,effectSpeed,effectIntensity,effectPalette,effectOption,R,G,B,R2,G2,B2,R3,G3,B3]`

- A new `Effect + White` mode controls 18 channels (same as above + including white channels `[..,W,W2,W3]`).
- `Effect Segment` is the same as `Effect` with 15 channels per segment; expect `channel 1 = "Segment Dimmer"`.
- `Effect Segment + White` is the same as `Effect Segment` incl whites, so it uses 18 DMX channels per segment.
- `Preset` mode triggers presets by sending preset ids and controls brightness = 2 DMX channel `[bri,pre]`.

The `effectOption` channel is divided into 4 macro parts to control _mirror_ and _reverse_ states as follows:

| value      | mirror   | reverse | |
|--------------|-----------|------------|--|
| 0..63 | false      | false        | **none** |
| 64..127 | false      | true        | **reverse** |
| 128..191 | true      | false        | **mirror** |
| 192..255 | true      | true        | **mirror & reverse** |

Segment modes introduce an additional `DMX segment spacing`. If spacing is `s = 0` subsequent DMX addresses for all segments are created. When `s > 0` a gap of `s` DMX addresses between segments is used. This feature is good to place segments (= DMX fixtures) inside DMX universe more efficient.

    Segment DMX Start Address (i) = DMXAddress + i * (dmxEffectChannels + DMXSegmentSpacing)

Where `DMXAddress` = start address as configured in UI, `dmxEffectChannels` = 15|18 depending on selected segment mode, `DMXSegmentSpacing` = address gap as configured in UI and `i` is the index id of each segment as existing.

As existing `Effect` mode, these modes are not realtime, and only support 1 universe as configured in UI.

Note: 1 DMX universe = max 512 DMX addresses. So max number of segments depends on  `DMXAddress`, selected segment mode and configured `DMX segment spacing`. To calculate how many segments can be controlled via DMX:

    Max segments = floor[ (512 - DMXAddress) / (dmxEffectChannels + DMXSegmentSpacing) ]

# PR notes

I tested everything expect `Preset`. Everything shall work as written here. **Please test and provide some feedback**! 

According to #2325 there will be some more DMX changes to come in future. I decided to provide this PR as is, because it does not interfere with the changes made for the other DMX modes mentioned. Maybe we merge this first, because it's finished from my point of view. We may provide further DMX changes in another PR. Alternate I could pull changes from #2325 into here to provide a more feature rich DMX advancement in a whole, if I find time to .. TBD

IMO separating this from the other DMX changes discussed in #2325 does no harm, it shall be merged anyhow.


wled00/e131.cpp
```diff
@@ -162,38 +162,71 @@ void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol){
...
+          else
+            dataOffset = DMXAddress;
+          // Modify address for Art-Net data
+          if (protocol == P_ARTNET && dataOffset > 0)
+            dataOffset--;
+          // Skip out of universe addresses
+          if (dataOffset > dmxChannels - dmxEffectChannels)
+            return;
+
+          if (e131_data[dataOffset+1] < strip.getModeCount())
```
blazoncek, line 196: The approach here is wrong. `col[]` and `colSec[]` are deprecated (and may be removed in the future).
Similar goes for `effectXxxxxx` variables. Their purpose is just informational.

Please use the following snippet:
```c++
Segment &segment = strip.getSegment(id);
segment.setMode(e131_data[...]);
segment.speed = e131_data[...];
...
```

wled00/e131.cpp
```diff
@@ -162,38 +162,71 @@ void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol){
...
+          colSec[0]       = e131_data[dataOffset+ 8];
+          colSec[1]       = e131_data[dataOffset+ 9];
+          colSec[2]       = e131_data[dataOffset+10];
+          if (dmxEffectChannels == 13) {
+            col[3]        = e131_data[dataOffset+11]; // white
+            colSec[3]     = e131_data[dataOffset+12];
+          }
+          if (isSegmentMode) {
+            opacity = e131_data[dataOffset+ 0];
+            segmentUpdated |= applyValuesToSegment(seg);
```
blazoncek, line 213: This function is unnecessary if you follow advice from above.

wled00/led.cpp
```diff
@@ -46,6 +46,22 @@ void applyValuesToSelectedSegs()
   }
 }
 
+// applies global legacy values to segment i (opacity, col, colSec, effectCurrent...)
```
blazoncek, line 49: Unnecessary and just adds confusion.

blazoncek (changes requested):
Applying segment values should be done in a up-to-date fashion, not using globals like `col[]`, etc.

mxklb:
@blazoncek thank you for the review. I pushed changes that shall resolve your comments as suggested (hopefully).

mxklb (commented):
done

mxklb:
Question: There is also a 3rd color for segments. Maybe 3rd RGBW would also be good to add to these `Effect` control modes, what do you think? I already missed it while using by some effects .. This would need additional 3|4 DMX channels per mode, so 11+3=14 or 13+4=17 (per segment). I also was thinking about enabling `Reverse Direction` and `Mirror Effect` within another +1 DMX channel (often DMX lights have such kind of mode switch channels). I think this would make this refactoring even more complete. If you think this would add value to these modes & users, let me know ..

wled00/e131.cpp
```diff
@@ -194,35 +194,36 @@ void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol){
...
-          col[0]          = e131_data[dataOffset+ 5];
-          col[1]          = e131_data[dataOffset+ 6];
-          col[2]          = e131_data[dataOffset+ 7];
-          colSec[0]       = e131_data[dataOffset+ 8];
-          colSec[1]       = e131_data[dataOffset+ 9];
-          colSec[2]       = e131_data[dataOffset+10];
+            if (e131_data[dataOffset+1] != seg.mode)  {strip.setMode(id,  e131_data[dataOffset+1]); stateChanged = true;}
+          if (e131_data[dataOffset+2]   != seg.speed)     {seg.speed     = e131_data[dataOffset+2]; stateChanged = true;}      
+          if (e131_data[dataOffset+3]   != seg.intensity) {seg.intensity = e131_data[dataOffset+3]; stateChanged = true;}
+          if (e131_data[dataOffset+4]   != seg.palette)   {seg.palette   = e131_data[dataOffset+4]; stateChanged = true;}
```
blazoncek, line 200: You can use `seg.setPalette()` to enable palette transitions.

wled00/e131.cpp
```diff
@@ -194,35 +194,36 @@ void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol){
...
+              strip.setBrightness(bri, true);
+            }
           }
         }
-        if (segmentUpdated) {
-          transitionDelayTemp = 0;
-          stateUpdated(CALL_MODE_NOTIFICATION);
-          return;
+        if (stateChanged) {
+          transitionDelayTemp = 0;                      // act fast
```
blazoncek, line 224: Not sure if this will help since transitions are started when changing segments.
It may also not be desired as it is inconsistent with regular behaviour.

blazoncek (approved):
Just additional info.
Otherwise looks ok, though I cannot test since I lack any DMX set-up.

wled00/e131.cpp
mxklb: There are also some other segment methods like `seg.setMode(), seg.setOpacity(), ..`. Due to transition triggers, to me it's unclear what is the better approach here. But I'll test what feels better/smoother when sending DMX. I'll leave a comment on these tests afterwards to decide ..

wled00/e131.cpp
mxklb: I'll test if `transitionDelayTemp = 0` has any impact here. It's in there because it was there in past. As you said, this could be obsolete, I'll find out.. Or maybe this needs to be placed before every `seg.setXXX()` method? Nevertheless, to call `stateUpdated` here I think is better than applying single channel changes immediately by `seg.setXXX()` methods (for every segment), triggering transitions. Maybe it's better to explicitly avoid transitions if possible. Starting transitions of DMX segment changes could impact performance of remote control, especially for larger amount of segments.

blazoncek: If you do not want transitions use `strip.setTransition(0);`. It will affect all segments, all kinds of transitions.

mxklb:
- The code now uses setter methods as @blazoncek mentioned (except effect _speed_ & _intensity_ - I couldn't find).
- Added missing effect options DMX channel (_mirror_ & _reverse_) 
- Added missing third color RGBW channels.

I tested most of it and it's working as proposed. Finally I don't know if there still should be a 'Legacy Effect Mode', to handle down compatibility using old `Effect` implementation with '11'|13 DMX channels? Maybe this may gain some value for DMX setups based on old `Effect` mode. **What do you think**, shall these old mode still exist to not break some user setups when released? Anyway users should be informed to migrate to refactored modes in release notes. It maybe a good idea to remove the old 'Legacy Effect Mode' within a future release upon .. TBD

It would be nice if someone finds time to test these new modes implemented here and provide some feedback.

blazoncek:
Good job. There are no setters for speed & intensity (and new custom parameters).
Please post request for testing in Discord beta-testing channel.

mxklb:
Beta test request: https://wled.discourse.group/t/dmx-effect-mode-refactoring-beta-testing/6498

lorandnagy:
Hi there!

Is this pull where also segment RGB should be implemented? Did not fint it after upload. Because of rekordbox I can test just that out. No effect possibiliti for me.

mxklb:
No, there is no realtime RGB segment control in this PR. Only `Effect` mode refactoring as in description.

wled00/e131.cpp
```diff
@@ -190,7 +190,7 @@ void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol){
           if (protocol == P_ARTNET && dataOffset > 0)
             dataOffset--;
           // Skip out of universe addresses
-          if (dataOffset > dmxChannels - dmxEffectChannels)
+          if (dataOffset > dmxChannels - dmxEffectChannels + 1)
```
blazoncek, line 193: why not just use `>=` instead of addition? more efficient code is always better.

mxklb: `dmxChannels - dmxEffectChannels = X`, so `dataOffset > X  != dataOffset > X+1`. 

So `dmxOffset >= X != dataOffset > X+1`. Lets test X=10, dmxOffset=10.

-  10 >= 10 = true
- 10 > 11 = false

There must be +1 channel. This case happens when WLED Effect Segment Fixture is placed at the last fixture in DMX universe only! The last segment gets skipped due to `dataOffset > dmxChannels - dmxEffectChannels`. This check was wrong, to control the last WLED Segment Fixture correctly, this check needs +1. Otherwise all segments are controllable expect last one. This is what this fix is for.

mxklb: +1 should not have a noticable impact on perfomance, but fixes a serious issue with DMX control here. I think this should not be a problem, do you? Of caues there are different ways to implement this, i'll stick to this because to me this fix does what it's supposed to do, with a minimum amount of code changes.

blazoncek: A single +1 is not a problem, >100,000 repetitions may become one. 😄 
Nothing more than that.

softhack007:
Hi, nice to see you are very busy in the DMX part of WLED:-)

I have a side-question, slightly off-topic but I need a DMX expert so I thought I'd ask here:

 the [SparkFun DMX output driver](https://github.com/Aircoookie/WLED/blob/main/wled00/src/dependencies/dmx/SparkFunDMX.cpp) does some weird things on non-existing pins, including installation of a timer interrupt that is not needed but may cause trouble. When building WLED for new MCUs like -S3, these parts of the  driver causes a watchdog reset, see issue #2928 .

I'm thinking of removing the DMX receive code from the driver, as currently it's unneeded code that seems to cause strange side-effects.This will only affect DMX receive from physical connections; DMX output and DMX over network will not change.

Can you confirm that the DMX receiver code from the sparkfun driver is not needed, and will not be needed for the changes you are working on?

mxklb:
> Can you confirm that the DMX receiver code from the sparkfun driver is not needed, and will not be needed for the changes you are working on?

As you see in diff to upstream, changes only act on e131/artnet package data. Changes to DMX receiver code shall not impact any changes made in this PR.

mxklb:
Beta tested successful: https://wled.discourse.group/t/dmx-effect-mode-refactoring-beta-testing/6498/29?u=mxklb

blazoncek:
I agree with the merge. @Aircoookie @softhack007 will you do the honours?

mxklb:
Is there anything missing or I can do?

softhack007:
> Is there anything missing or I can do?

Sorry, this PR got a bit out-of-sight for me over xmas/new year. I'll cross-read the code - and in case nothing critical pops up, i'll agree to merge.

blazoncek:
I lack any DMX HW so can't test this but code looks ok to me.
No objections, approved.

softhack007 (approved):
Also an  OK from my side, however I also don't have DMX hardware to test with.

mxklb:
It's tested by myself && in https://wled.discourse.group/t/dmx-effect-mode-refactoring-beta-testing

As far as we haven't missed something, it's working exactly as described above. IMO it's save to merge ..

mxklb:
https://kno.wled.ge/interfaces/e1.31-dmx > 0.14.0-b1 has to be updated, check docs above to describe the new modes ..

blazoncek:
@mxklb please make a PR with updated content to [WLED-docs](https://github.com/aircoookie/WLED-docs)

mxklb:
check out https://github.com/Aircoookie/WLED-Docs/pull/106

PR #2892, opened 2022-11-20
2d Mapping with Matrix Gaps

ctjet:
This enables WLED in 2D mode to support gaps and offsets of panels in the matrix. Panels may now be spaced and ordered however the user sees fit.

Screenshot 2022-11-19 175952

### Changes

wled00/cfg.cpp
- No longer store matrix dimensions with panel units, instead, store the total count of panels
- Store an X and Y offset in pixels from top left of matrix for each panel. (uint8_t)

wled00/FX_2Dfcn.cpp
- No longer calculate matrix serpentine, bottom, right, and vertical. This functionality is now in 2D settings webpage. Instead use already set panel offsets and just calculate pixel offsets per panel relative to that.

wled00/data/settings_2D.htm
- Add "matrix generator" field. Use this field to populate the panels array and new "Number of panels" field with the correct values for a new matrix.
- Add offsets for each panel from the top left of the matrix. These can be manually tweaked to allow for gaps and weird spacings.

Other supporting changes were made for above functionality.

### Notes
I am by no means a Javascript or Web developer. All of the JS and HTML, as well as my layout of the 2D settings page, should be checked and potentially tweaked.

This should not be merged until it is decided how to handle custom maps with gaps. This functionality currently can balloon the size of the Custom Map array as every unused pixel takes up 2 bytes. This is a problem especially for ESP8266.

I am not sure if this breaks the config when upgrading to this version due to the config changes. Must be tested.


blazoncek:
Looks slick.

blazoncek:
Each iteration looks better. 👏 
Adding unequal panel support may be your next step and then it will be almost perfect.

ctjet:
Added per panel w/h, still looking into a fix for a few bugs

wled00/FX.h
```diff
@@ -809,22 +806,24 @@ class WS2812FX {  // 96 bytes
...
-      panelW,
       matrixWidth,
       matrixHeight;
 
     typedef struct panel_bitfield_t {
-      bool bottomStart : 1; // starts at bottom?
-      bool rightStart  : 1; // starts on right?
-      bool vertical    : 1; // is vertical?
-      bool serpentine  : 1; // is serpentine?
+      unsigned int xOffset  : 8; // x offset relative to the top left of matrix in LEDs
```
blazoncek, line 816: You may want to move these four variables out of bitfield to allow lower CPU overhead.

blazoncek:
@ctjet can you, please, pull recent commits and resolve conflicts?

ctjet:
@blazoncek should be updated, do I need to do anything special for 0.14?

blazoncek:
> do I need to do anything special for 0.14?

No. It just needs to be in line with recent main.

wled00/FX.h
```diff
@@ -654,10 +654,9 @@ class WS2812FX {  // 96 bytes
       timebase(0),
       isMatrix(false),
 #ifndef WLED_DISABLE_2D
-      hPanels(1),
-      vPanels(1),
-      panelH(8),
-      panelW(8),
+      panels(1),
+      matrixWidth(DEFAULT_LED_COUNT),
```
blazoncek, line 658: This set-up is not 2D, please change it that both dimensions are >1.

wled00/FX.h
```diff
@@ -808,20 +807,24 @@ class WS2812FX {  // 96 bytes
...
     #define WLED_MAX_PANELS 64
     uint8_t
-      hPanels,
-      vPanels;
+      panels;
 
     uint16_t
-      panelH,
-      panelW;
+      matrixWidth,
```
blazoncek, line 813: These two are no longer part of `strip`.

wled00/FX.h
blazoncek: I am still waiting for these 4 to get removed from bitfield.

wled00/FX_2Dfcn.cpp
```diff
@@ -43,8 +43,17 @@ void WS2812FX::setUpMatrix() {
 
   // isMatrix is set in cfg.cpp or set.cpp
   if (isMatrix) {
-    Segment::maxWidth  = hPanels * panelW;
-    Segment::maxHeight = vPanels * panelH;
+    // calculate width dynamically because it will have gaps
+    matrixWidth = 0;
```
blazoncek, line 47: You will need to revert to `Segment::maxWidth` and `Segment::maxHeight`.

Do similar for the code below.

blazoncek (changes requested):
There were some changes upstream that require you to update your code. It may not work properly otherwise.

I am also wondering why did you include `html_simple.h` as its source didn't change.

blazoncek:
Maybe it is not ready after all.
@ctjet how do you plan to arrange panels? Always by entering x and y offsets?
That is awkward IMO.

blazoncek:
Got it working now.
I'm still not sure entering offsets is a good UX.
It might be better to add graphical UI at some point, but that is beyond my skills.

Aircoookie (approved):

blazoncek (approved):

wled00/FX.h
blazoncek: Did it myself.

blazoncek:
@softhack007 are you ok to merge this?

softhack007 (approved):
@blazoncek just did a quick code review, however 2D is not really my field ...
I think its good to merge 👍

Issue #1964, opened 2021-05-17
Can I add the function of scanning nearby WIFI to select connection?


zx804262156:
Can  add the function of scanning nearby WIFI to select connection?
Because there are too many WIFI name symbols, it is difficult to fill in。

Aircoookie:
Hi!
That functionality is not currently supported, but if you'd like to add it yourself, I'd likely merge your pull request :)

huggy-d1:
If you use a PC to set up WLED the first time, you can easily type the symbols.  Of course, in these smart device times, not everyone has a PC available with wifi (so it can connect to the WLED AP and assign the SSID / password).

Another option which I find useful is to create a little "note" on my smart device with the wifi details.  Then I can copy/paste it into WLED fields pretty easily and without worrying I have not (temporarily) bricked WLED - until it automatically reverts back to AP mode.

sansillusion:
It would be great to see an easy wifi selection, would make it easier for
less tech savy people I make lamps for.

Le lun. 17 mai 2021 9 h 13 a.m., huggy-d1 ***@***.***> a
écrit :

bad-jesus:
@laserguru 
Any progress on this? Would be interested to see how far you got if not.

laserguru:
Got sidetracked and never got to coding it, but I have made a sketch of the concept I had so far for the design of the interface, to share what little progress I got made.

Besides the main network selector, I think it's very important to have a simple and clear way two get back to the network selection mode (AP mode). My idea was to use the same GPIO line that needs to be pulled for flashing, because there's a user accessible button there most of the time.
There was a firmware, not sure which one (I think ESPeasy or Tasmota), that used the built in LED to indicate when you could press the button without triggering another boot mode. I think that can be implemented here too. That LED shares the pin with UART1 TX for DMX (UART0 is the one used for the bootloader), but if that gets initialized a little bit later, the LED can be flashed a couple of times to indicate you can press the button to get into AP mode. After that the startup would continue as normal.

blazoncek:
Default button (tagged button 0, defaults to GPIO0 but can be changed) already has the "start AP mode" built in. Just long press it for 5s-9s.

laserguru:
I just noticed that in button.cpp! I'm also going back the revision history and realized that was implemented **_ages_** ago! As in literally a couple of years. I had never noticed the feature! And it seems many users neither, judging by some tutorials I've come across that suggest switching off your router to invoke AP mode.

[...]

PR #2895, opened 2022-11-20
feat(json): add wifi scanning

jkoelker:
Adds wifi scanning to the JSON api at `/json/networks`. The initial request will trigger a scan, subsequent requests will scan or return the results depending on the state of the `WiFiScan`.

Add a `Scan` button next to the client ssid input, on click, scan for networks, and change the input to a select with the found ssids.

Fixes: #1964


jkoelker:
Scan button:
Screenshot from 2022-11-20 20-50-42

Feedback when scanning (button is disabled):
Screenshot from 2022-11-20 20-50-57

Input is swapped out to a `select`, options are de-duplicated by SSID (for multiple BSSIDs sharing the same SSID) and sorted by RSSI:
Screenshot from 2022-11-20 20-51-03

wled00/data/settings_wifi.htm
```diff
@@ -6,18 +6,100 @@
 	<meta content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" name="viewport"/>
 	<title>WiFi Settings</title>
 	<script>
-		var d=document;
-		var loc = false, locip;
+		let d = document;
```
blazoncek, line 9: `var` vs. `let` : I tend to use `var` for globals like `d` and use `let` within functions, so I recommend sticking with `var` here.
Truth is, there were no guidelines for this so there is plenty of mixed cases all over the place.

wled00/data/settings_wifi.htm
```diff
@@ -6,18 +6,100 @@
...
+		let loc = false, locip;
+		let scanLoops = 0;
+
+		function Local() {
```
blazoncek, line 13: No added benefit of using function here except wasting a few bytes of flash memory. Better to leave the code where it was in `S()`.

wled00/data/settings_wifi.htm
```diff
@@ -6,18 +6,100 @@
...
+				locip = localStorage.getItem("locIp");
+				if (!locip) {
+					locip = prompt("File Mode. Please enter WLED IP!");
+					localStorage.setItem("locIp", locip);
+				}
+			}
+		}
 		function H(){window.open("https://kno.wled.ge/features/settings/#wifi-settings");}
 		function B(){window.open("/settings","_self");}
+		function N() {
```
blazoncek, line 25: Check (and fix) the indentation tabs (looks like you have a few spaces there).

wled00/data/settings_wifi.htm
```diff
@@ -6,18 +6,100 @@
...
+			const url = (loc?`http://${locip}`:"") + "/json/networks";
```
blazoncek, line 26: If it must be `/json` then just shorten it to `/json/net`. I would prefer it elsewhere if possible, though.
See comments below.

wled00/data/settings_wifi.htm
```diff
@@ -6,18 +6,100 @@
...
+
+			const button = d.getElementById("scan");
```
blazoncek, line 28: It may be a good idea to make `function gId(e) { return d.getElementById(e); }` at the beginning to save some flash memory and then use `gId("scan");`.

wled00/json.cpp
```diff
@@ -872,6 +880,35 @@ void serializePalettes(JsonObject root, AsyncWebServerRequest* request)
   }
 }
 
+void serializeNetworks(JsonObject root)
+{
+  JsonArray networks = root.createNestedArray("networks");
```
blazoncek, line 885: Use `F("string")` to store strings in flash memory instead or precious RAM.

wled00/json.cpp
```diff
@@ -922,12 +959,13 @@ void serveJson(AsyncWebServerRequest* request)
...
-  else if (url.indexOf("nodes") > 0) subJson = 4;
-  else if (url.indexOf("palx")  > 0) subJson = 5;
-  else if (url.indexOf("fxda")  > 0) subJson = 6;
+  if      (url.indexOf("state") > 0) subJson = JSON_PATH_STATE;
+  else if (url.indexOf("info")  > 0) subJson = JSON_PATH_INFO;
+  else if (url.indexOf("si")    > 0) subJson = JSON_PATH_STATE_INFO;
+  else if (url.indexOf("nodes") > 0) subJson = JSON_PATH_NODES;
+  else if (url.indexOf("palx")  > 0) subJson = JSON_PATH_PALETTES;
+  else if (url.indexOf("fxda")  > 0) subJson = JSON_PATH_FXDATA;
+  else if (url.indexOf("networks") > 0) subJson = JSON_PATH_NETWORKS;
```
blazoncek, line 968: Use shorter string as it is stored in RAM in this case. See above.

blazoncek (changes requested):
I like the idea and the implementation (hacking input field as I did in Usermod settings).

Still I would love to see fixes implemented prior to merging.

wled00/data/settings_wifi.htm
jkoelker: Ooops, this was left over from my first attempt where i had to call it in two places, moving back ;).

wled00/data/settings_wifi.htm
jkoelker: Totally, `var` will leak out into the global `window` but yea, I'll switch all these back to `var` to remain consistent here in the codebase.

wled00/data/settings_wifi.htm
jkoelker: Good idea, I also did the same for `function cE(e) { return d.createElement(e); }` since it was used a few times as well.

wled00/json.cpp
jkoelker: I always forget coming from esp32 development were `F()`  is basically `#define F(x) x` and all constant strings are in flash anyway. I need to remember esp8266 support ;).

wled00/data/settings_wifi.htm
jkoelker: I think putting it within the `/json` url prefix makes the most sense from an api point of view, but with memory taken into account, would moving it to `/net` make the more sense and dispatching to its own handler?

For now I'll just make it the shorter `/json/net`, but open to moving it all together.

wled00/json.cpp
blazoncek: F() is flash string helper, which is a class method. On the other hand PSTR() is just a moniker on ESP32 that does nothing.

blazoncek (approved):
Looks fine now.

blazoncek:
@Aircoookie do we merge this for 0.14 (as it doesn't break any other feature) or wait for 0.14.1?

Aircoookie:
Like this a lot!

PR #2899, opened 2022-11-22
Don't use internal pulldown on button inputs

DerSpatz:
After stumbling upon issues with a PCB design, I found out that the issue was caused by the use of internal pulllup resistors on the input pins that is not mentioned in the documentation. Another suggestion for my fix for the documentation was to mention that on ESP32 (but not ESP8266), an internal pulldown (instead of pullup) is used when the input type is set to "inverted button".

This behaviour is confusing because of several reasons:

1.  It is different between ESP32 and ESP8266, and many boards can be used with both MCUs on pin compatible modules (Wemos D1 etc.).
2. Normally, a button is connected between signal and ground for several reasons: It is easier to route the PCB as a GND layer should be all over the board, whereas the power trace must be routed to the button if it is connected between VDD and signal. GND is usually less noisy than VDD. Debouncing capacitor should be connected between signal and GND. Less EMI as the return path for the signal is GND instead of VDD.
3. As most buttons are connected between signal and GND, with a pullup to VDD on most boards, setting an internal pullldown resistor might cause additional issues, as then the external pullup and the internal pulldown act as a voltage divider, which can then cause the signal to be under the threshold voltage for high level. As the internal pull resistor value is not very precise ("between 10k and 100k", https://esp32.com/viewtopic.php?t=5111), this adds another layer of confusing randomness to the problem, as even though the button might work on a certain pin on a certain board, it might not work on a different pin or a different board.

The "inverted button" has a different use case: wiring is the same as with a normal button (pullup to VDD, switch and debounce capacitor connected to GND), but a normal switch is normally open and a inverted switch is normally closed.
On a normally open switch, the signal is usually pulled high, but is connected to GND when the switch is activated.
On a normally closed switch, the signal is usually connected to GND, but is pulled high when the switch is activated.

I also checked https://kno.wled.ge/basics/compatible-hardware/ and none of the boards in the list uses buttons that are connected between signal and VDD. So changing the behaviour as suggested should not cause any issues with existing designs.

Here's the link to the PR for the fix and additional information in the documentation:https://github.com/Aircoookie/WLED-Docs/pull/93

Best solution would be to allow the user to set the input type (pullup, pulldown, none), but this is beyond my coding capabilities. I'm am hardware guy.

### DIFF
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -200,11 +200,7 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
       int8_t pin = btn["pin"][0] | -1;
       if (pin > -1 && pinManager.allocatePin(pin, false, PinOwner::Button)) {
         btnPin[s] = pin;
-        #ifdef ESP32
-        pinMode(btnPin[s], buttonType[s]==BTN_TYPE_PUSH_ACT_HIGH ? INPUT_PULLDOWN : INPUT_PULLUP);
-        #else
         pinMode(btnPin[s], INPUT_PULLUP);
-        #endif
       } else {
         btnPin[s] = -1;
       }


blazoncek:
I understand your reasoning but the pull up/down code was added (on request) as a mechanism for users that only attach buttons to the ESP.

If you have added pull up/down resistors to your PCB design and do not want the behavior currently implemented, you can always fork and compile yourself. We can help by adding compile time override instead of removing existing code as a temporary solution.
You must understand that a lot of hobbyists just buy an ESP, a few wires, PSU and a LED strip. And it works for them (sometimes).
Adding a button (or switch) is then just a matter of attaching it to available GPIO pins.

I do not approve your solution but I promise I will think of a way to accommodate designs employed in your PCB.

DerSpatz:
The current version of the code does not support normally closed buttons, as normally closed buttons would be triggered all the time *except* when they're pushed. 

With my suggested change (keep the input as active high, but use an internal pullup) normally closed buttons would be supported *without* changing any of the current functionality: you can still use only wires to connect buttons, but now you can both types of available switches. Only change is that you have to connect the button between GPIO and GND, but this is the most common way to wire them anyways, and all boards I have seen that have button headers connect them from GPIO to GND.

I am not suggesting disabling the internal pullups (even though this option can be useful), but only to get rid of the pulldown for the reasons mentioned above. You *add* functionality while at the same time reducing confusion.

Here's a simple simulation of different button setups to make it more clear: https://tinyurl.com/2hjbblrs
ipd=internal pulldown, ipu=internal pullup, epu=external pullup, nc=normally closed, no=normally open
1&2 are the current state with normally open buttons, which work as intended
3&4 are the current state with normally closed buttons, which trigger all the time (not OK)
5 is my PR, which only triggers the normally closed button when it is pressed
6 is the worst case situation with external pullup and internal pulldown

softhack007:
I agree with the assessment from @blazoncek; actually I can remember that INPUT_PULLDOWN was added due to request from another user. Simply removing this breaks existing functionality so not exactly a good idea. 

In fact a proper solution means to allow all three modes (pullup, none, pulldown) in combination with two button interpretations (inverted and non-inverted) = 6 possible combinations. Not sure if users will be more happy with that.

So some more analysis seems necessary. Another way could be to have a new compile time option, like `-D WLED_DONT_PULL_MY_BUTTONS_DOWN`.

blazoncek:
Considered normally closed button connected to +3.3V?
With a pull down (external or internal).

This is where inverted button type comes in. Not as you described it.

Internal pull up/down ensures the GPIO is not floating if external resistor is lacking.
As I said I will add an option to disable internal pull resistors in the future to accommodate your PCB design.

PR #2905, opened 2022-11-25
Fixes platformio.ini for the ESP32 S3

JPZV:


softhack007:
Hi, thanks for the update.

A few requests from my side:
1/ I would prefer to stay with espressif32@5.1.1, simply because it is a well-tested version especially with respect to I2S (audioreactive). 
2/ Can you explain a bit more about compilation errors you have? I never experienced any.
3/ It seems that the previously used platform_packages version has been removed. Instead of fixing another version, it should be sufficient to use `platform_packages =` (empty line) as this means "use the default package".

Thanks again for the contribution and research you did!

platformio.ini
```diff
@@ -401,8 +401,8 @@ lib_deps = ${esp32c3.lib_deps}
 [env:esp32s3dev_8MB]
 ;; ESP32-S3-DevKitC-1 development board, with 8MB FLASH, no PSRAM
 board = esp32-s3-devkitc-1
-platform = espressif32@5.1.1
-platform_packages = platformio/framework-arduinoespressif32@3.20004.220825
+platform = espressif32@5.2.0
```
softhack007, line 404: Se my comment. Can we stay with espressif32@5.1.1? What are the compile errors you experienced?

platformio.ini
```diff
@@ -401,8 +401,8 @@ lib_deps = ${esp32c3.lib_deps}
...
+platform_packages = platformio/framework-arduinoespressif32@3.20005.220925
```
softhack007, line 405: `platform_packages = ` (empty) should work.

softhack007 (changes requested):
see comments above

JPZV:
Hi, regardling about espressif32@5.1.1, you can check the linked issues (https://github.com/platformio/platform-espressif32/issues/912 and https://github.com/platformio/platform-espressif32/pull/914), but in short, there's an issue while compiling for ESP32-S3 saying that the sdkconfig.h doesn't exists. I'd recommend you to see both issues for more details ;) but it seems that it's only regardling to the ESP32-S3. If you want, I'll leave my override.ini down here so you can test it

[platformio_override.txt](https://github.com/Aircoookie/WLED/files/10094582/platformio_override.txt)

(please remember to change its extension to .ini)

For the second point, I only got two compiling error, the first one was about the sdkconfig.h, and the second one about the missing package which was removed. While for the third point, I didn't know that it was able to use the default package, so I don't think it's a problem. After changing both lines, the build was completely straight forward

softhack007:
Really I am very careful with updating espressif releases, from bad experience. Our chinese friends seem to be very innovative sometimes, however not good with _not_ breaking exisiting functionality.

After a bit of research, it seems there are two "main variants" of ESP32-S3:
* One with "classic" qio flash and no PSRAM. I've actually compiles for it earlier today, it got no "sdkconfig.h not found" errors.
* One with octal flash and PSRAM. These seem to need `memory_type=qspi_opi` to work

My suggestion is to create two platformio.ini entries for S3: 
* one for "S3 with qio", and 
* a new one for supporting your board that seems to need a newer espressif release. 

The "S3 with qio" should only get the `platform_packages =` change, while your new entry could use `platform = espressif32@5.2.0`. I will also double-check that the exisiting `[env:esp32s3dev_8MB]` still works for my board.

@blazoncek what do you think?

JPZV:
> however not good with _not_ breaking existing functionality.

Yeah, you're right there hahaha. I lost most of my hair because of them.



> * One with octal flash and PSRAM. These seem to need `memory_type=qspi_opi` to work.

Seems that you're right. My board (the [TinyS3](https://esp32s3.com/tinys3.html)) actually has 8MB extra QSPI PSRAM, so maybe you're right, they overlook that scenario and maybe updating to 5.2.0 may produce compiling errors for non-PSRAM S3 boards.

Sadly, I cannot test the Sound reaction feature as I only have one TinyS3 (and they're quite expensive for me actually) and I'll not use that feature on my project :(

softhack007:
> Sadly, I cannot test the AudioReactive feature as I only have one TinyS3 (and they're quite expensive for me actually) and I'll not use that feature on my project :(

No problem - if you update the platformio.ini entries, then I can verify (over the weekend) that audioreactive usermod still compiles for both TinyS3 and previous S3 environments.

JPZV:
Sorry for the delay. I added the new environment for the PSRAM version, though, I'm not sure if that's how it's suppose to be added to the platformio.ini as this is the first project where I use it.

If there's anything else that I can do, please let me know

blazoncek:
As much as I'd like to test (or review) I lack S3. 😄 
As for platform update, I agree with the notion that updating it too soon may cause unnecessary headaches.
I was testing 5.1.1 for about 6 weeks before deciding to recommend it (for S2 & C3). I still do not recommend it for classic ESP32.

softhack007 (approved):

softhack007:
Hi again, I still have problems to compile with your new environment "out of the box".
- Error: Unknown board ID 'esp32-tinys3'

If I use `board = esp32s3box` instead, compilation starts but aborts with `'esp32_adc2gpio' was not declared in this scope`

Any Ideas?

```
wled00/src/dependencies/blynk/Blynk/BlynkApiArduino.h:93:19: note: in expansion of macro 'BLYNK_DECODE_PIN'
     uint8_t pin = BLYNK_DECODE_PIN(it);
                   ^~~~~~~~~~~~~~~~
In file included from C:/Users/user/.platformio/packages/framework-arduinoespressif32/cores/esp32/esp32-hal-gpio.h:29,
                 from C:/Users/user/.platformio/packages/framework-arduinoespressif32/cores/esp32/esp32-hal.h:79,
                 from C:/Users/user/.platformio/packages/framework-arduinoespressif32/cores/esp32/Arduino.h:36,
                 from wled00/wled.h:63,
                 from wled00/FX_fcn.cpp:26:
wled00/src/dependencies/blynk/Blynk/BlynkApiArduino.h: In member function 'void BlynkApi<Proto>::processCmd(const void*, size_t)':
C:/Users/user/.platformio/packages/framework-arduinoespressif32/variants/esp32s3box/pins_arduino.h:13:48: error: 'esp32_adc2gpio' was not declared in this scope
 #define analogInputToDigitalPin(p)  (((p)<20)?(esp32_adc2gpio[(p)]):-1)
```

JPZV:
So sorry, I misspelled it, it should be _um_tinys3_, but, for some reason, it also fails with the same error (`'esp32_adc2gpio' was not declared in this scope`)

Funny enough, it can be compiled with _esp32-s3-devkitc-1_ (just like the non-PSRAM version) and it works on my board, but not sure if the `esp32_adc2gpio` could make some troubles with WLED (at least I can control sk6812 RGBW leds with it)

Here's my build log

```
Processing esp32s3dev_8MB_PSRAM (board: esp32-s3-devkitc-1; platform: espressif32@5.2.0; framework: arduino)
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Verbose mode can be enabled via `-v, --verbose` option
*** use existing my_config.h ***
CONFIGURATION: https://docs.platformio.org/page/boards/espressif32/esp32-s3-devkitc-1.html
PLATFORM: Espressif 32 (5.2.0) > Espressif ESP32-S3-DevKitC-1-N8 (8 MB QD, No PSRAM)
HARDWARE: ESP32S3 240MHz, 320KB RAM, 8MB Flash
DEBUG: Current (esp-builtin) On-board (esp-builtin) External (cmsis-dap, esp-bridge, esp-prog, iot-bus-jtag, jlink, minimodule, olimex-arm-usb-ocd, olimex-arm-usb-ocd-h, olimex-arm-usb-tiny-h, olimex-jtag-tiny, tumpa)        
PACKAGES:
 - framework-arduinoespressif32 @ 3.20005.220925 (2.0.5)
 - tool-esptoolpy @ 1.40201.0 (4.2.1)
 - toolchain-riscv32-esp @ 8.4.0+2021r2-patch3
 - toolchain-xtensa-esp32s3 @ 8.4.0+2021r2-patch3
Converting wled00.ino
LDF: Library Dependency Finder -> https://bit.ly/configure-pio-ldf
LDF Modes: Finder ~ chain, Compatibility ~ strict
Found 38 compatible libraries
Scanning dependencies...
Dependency Graph
|-- FastLED @ 3.5.0
|   |-- SPI @ 2.0.0
|-- IRremoteESP8266 @ 2.8.2
|-- ESPAsyncWebServerAircoookie @ 2.0.7+sha.b9349dc
|   |-- AsyncTCP @ 1.2.0+sha.38b8475
|   |-- FS @ 2.0.0
|   |-- WiFi @ 2.0.0
|   |-- LittleFS @ 2.0.0
|   |   |-- FS @ 2.0.0
|-- NeoPixelBus @ 2.7.0+sha.b32f719
|   |-- SPI @ 2.0.0
|   |-- I2S @ 1.0
|-- AsyncTCP @ 1.2.0+sha.38b8475
|-- Wire @ 2.0.0
|-- SD @ 2.0.0
|   |-- FS @ 2.0.0
|   |-- SPI @ 2.0.0
|-- SD_MMC @ 2.0.0
|   |-- FS @ 2.0.0
|-- SPI @ 2.0.0
|-- EEPROM @ 2.0.0
|-- ArduinoOTA @ 2.0.0
|   |-- Update @ 2.0.0
|   |-- WiFi @ 2.0.0
|   |-- ESPmDNS @ 2.0.0
|   |   |-- WiFi @ 2.0.0
|-- DNSServer @ 2.0.0
|   |-- WiFi @ 2.0.0
|-- ESPmDNS @ 2.0.0
|   |-- WiFi @ 2.0.0
|-- Ethernet @ 2.0.0
|   |-- WiFi @ 2.0.0
|-- LittleFS @ 2.0.0
|   |-- FS @ 2.0.0
|-- WiFi @ 2.0.0
|-- ESP32 Async UDP @ 2.0.0
|-- WebServer @ 2.0.0
|   |-- WiFi @ 2.0.0
|   |-- FS @ 2.0.0
Building in release mode
Retrieved (...)
Retrieving maximum program size .pio\build\esp32s3dev_8MB_PSRAM\firmware.elf
Checking size .pio\build\esp32s3dev_8MB_PSRAM\firmware.elf
RAM:   [==        ]  22.3% (used 73068 bytes from 327680 bytes)
Flash: [======    ]  62.1% (used 1301301 bytes from 2097152 bytes)
Retrieved `.pio\build\esp32s3dev_8MB_PSRAM\firmware.bin' from cache
================================================================================================= [SUCCESS] Took 12.09 seconds =================================================================================================

Environment           Status    Duration
--------------------  --------  ------------
esp32s3dev_8MB_PSRAM  SUCCESS   00:00:12.094
```

softhack007:
`'esp32_adc2gpio' was not declared in this scope` seems to be a problem with latest arduino-esp32.
Maybe you should create a bug report for that project, as many -S3 targets are affected.

As a quick workaround:
* edit  C:/Users/user/.platformio/packages/framework-arduinoespressif32/variants/um_tinys3/pins_arduino.h
* comment out this line `#define analogInputToDigitalPin(p)  (((p)<20)?(esp32_adc2gpio[(p)]):-1)`
* replace with `#define analogInputToDigitalPin(p)  (((p)<20)?(analogChannelToDigitalPin(p)):-1)`

save, recompile, it works (for me).

JPZV:
Thanks for the Workaround! Can confirm it compiles now and it works on my TinyS3!

 I'll let them know right away!

PR #2908, opened 2022-11-26
FX data optimization

Aircoookie:
- Fixed some inaccurate effect metadata.
- Reduced fxdata size by 400b by using single char flags


ChrisGebhardt:
There seems to be a conflict with the usage of FX colors vs. palette colors in the UI logic. After this change I am unable to use the UI to select 3 colors for the Color Gradient palette, and in fact this palette option sometimes disappears. (along with related Colors Only palette) Although, perhaps this will be irrelevant if a custom gradient designer materializes. :)

blazoncek:
> ... sometimes disappears. (along with related Colors Only palette)

This is intentional and depends on the effect used.
It may be erroneous (a bug) but it is still intentional.

ChrisGebhardt:
I think the best current compromise is that effects with no colors specified in the FX data should default to the generic 1,2,3 color selector. (ie. effects that simply pull colors from the current palette, such as Palette and TwinkleFox)  This way allows for the common use case of defining a custom Color Gradient palette for these effects. It doesn't help if the effect draws on top of a BG color, such as TwinkleUp. In this case, we'd need the 1,2,3 selector for a custom palette plus extra effect specific colors. And SEGMENTs have had NUM_COLORS 3 colors forever, so I doubt that'll change..

blazoncek:
IMO each effect should have correct fxdata. Defaults may get skewed in the future.
@ChrisGebhardt You are welcome to go through each and every effect and fix fxdata. Then make a [PR](https://github.com/Aircoookie/WLED/wiki/How-to-properly-submit-a-PR).

Aircoookie:
Thank you for noticing this issue! Indeed the special palettes (*Colors 1&2) rely on color selections, this is a detail I had thought of, but then forgot.

I think the implementation most in spirit with FX data (show controls only when they do something) would be to always show color selectors if a color-based palette is selected, and hiding them otherwise (if the FX doesn't use colors, but only palette)

blazoncek:
> Thank you for noticing this issue! Indeed the special palettes (*Colors 1&2) rely on color selections, this is a detail I had thought of, but then forgot.

This has been thought of and implemented. If the effect implements (uses) colors 1, 2 & 3 all palettes are shown. If it doesn't use color 3 not all "* Color..." palettes are shown.
Some effects still have incorrectly set fxdata so the palettes are missing most likely.

Aircoookie:
@blazoncek thank you, though this results in FX having to show color controls if they want to utilize `*` palettes, even though the color controls don't do anything if any standard palette is selected.
Disabling these palettes doesn't really make any sense, since any effect supporting palettes can take advantage of `*` palettes.
I'd propose always having `*` palettes active and showing color controls either if the effect calls for it OR `*` palette is selected. Will hopefully implement tonight and finally release b1 tomorrow or Friday.

blazoncek:
The idea of hiding palettes & colors comes from SR fork where the implementation allowed all three colors to show if not defined. It also showed all or none of the palettes. This has been adjusted but most likely not for all effects properly. I did go through all of them (logically) and tried to add relevant data. May have missed a few effects as some are not straightforward.

This is one reason I added palette icon to effect name in UI, to visually help finding problematic effects (the ones that actually use any kind of palette). Let's have a chat on Discord to further clarify.

ChrisGebhardt:
I agree with the minimalization of FX data by not having to specify 1,2,3 (or something like FG,BG,3) to enable use of all color palettes. Indeed, this has nothing to do with the effect itself so would violate the spirit of FX data. I also like the UI design of hiding unrelated controls. Perhaps a further tweak would be to allow some overlap of functionality but clearly label this in the UI. So if an effect uses "FG" and "BG" but then the user selects "Color Gradient" (assuming made visible), the labels would change to "1 / FG", "2 / BG", "3" to indicate their dual purpose.  Then again, the best solution is probably to eliminate these types of ad-hoc 2/3-color palettes once a user-friendly custom palette designer is available, leaving segment colors solely for effects.

PR #2912, opened 2022-11-28
PWM outputs usermod

ardiloot:
Add generic PWM outputs usermod for controlling servo motors, LED brightness, etc. More info in readme file

### DIFF
--- a/usermods/pwm_outputs/readme.md
+++ b/usermods/pwm_outputs/readme.md
@@ -1,0 +1,21 @@
+# PWM outputs
+
+v2 Usermod to add generic PWM outputs to WLED. Usermode could be used to control servo motors, LED brightness or any other device controlled by PWM signal.
+
+## Installation
+
+Add the compile-time option `-D USERMOD_PWM_OUTPUTS` to your `platformio.ini` (or `platformio_override.ini`). By default upt to 3 PWM outputs could be configured, to increase that limit add build argument `-D USERMOD_PWM_OUTPUT_PINS=10` (replace 10 by desired amount).
+
+Currently only ESP32 is supported.
+
+## Configuration
+
+By default PWM outputs are disabled, navigate to Usermods settings and configure desired PWM pins and frequencies.
+
+## Usage
+
+If PWM output is configured, it starts to publish its duty cycle value (0-1) both to state JSON and to info JSON (visible in UI info panel). To set PWM duty cycle, use JSON api (over HTTP or over Serial)
+
+```json
+{"pwm_0": 0.0, "pwm_1": 0.1, ...}
+```
--- a/usermods/pwm_outputs/readme.md
+++ b/usermods/pwm_outputs/readme.md
@@ -17,5 +17,11 @@ By default PWM outputs are disabled, navigate to Usermods settings and configure
 If PWM output is configured, it starts to publish its duty cycle value (0-1) both to state JSON and to info JSON (visible in UI info panel). To set PWM duty cycle, use JSON api (over HTTP or over Serial)
 
 ```json
-{"pwm_0": 0.0, "pwm_1": 0.1, ...}
+{
+    "pwm": {
+        "0": {"duty": 0.1},
+        "1": {"duty": 0.2},
+        ...
+    }
+}
 ```
--- a/usermods/pwm_outputs/usermod_pwm_outputs.h
+++ b/usermods/pwm_outputs/usermod_pwm_outputs.h
@@ -13,15 +13,21 @@
 class PwmOutput {
   public:
 
-    PwmOutput() {
-      DEBUG_PRINTLN("pwm_output[-1]: setup disabled output");
-    }
-
-    PwmOutput(int8_t pin, uint32_t freq) : pin_(pin), freq_(freq) {
+    void open(int8_t pin, uint32_t freq) {
       DEBUG_PRINTF("pwm_output[%d]: setup to freq %d\n", pin_, freq_);
-      if (pin_ < 0 || !pinManager.allocatePin(pin_, true, PinOwner::UM_PWM_OUTPUTS)) {
-        return;
+
+      if (enabled_) {
+        if (pin == pin && freq == freq_) {
+          return;  // PWM output is already open
+        } else {
+          close();  // Config has changed, close and reopen
+        }
       }
+
+      pin_ = pin;
+      freq_ = freq;
+      if (pin_ < 0 || !pinManager.allocatePin(pin_, true, PinOwner::UM_PWM_OUTPUTS))
+        return;
       
       channel_ = pinManager.allocateLedc(1);
       if (channel_ == 255) {
@@ -44,34 +50,21 @@ class PwmOutput {
       if (channel_ != 255)
         pinManager.deallocateLedc(channel_, 1);
       channel_ = 255;
+      duty_ = 0.0f;
       enabled_ = false;
-      duty_ = -1.0f;
     }
-  
+
     void setDuty(const float duty) {
       DEBUG_PRINTF("pwm_output[%d]: set duty %f\n", pin_, duty);
-      if (!enabled_) {
+      if (!enabled_)
         return;
-      }
       duty_ = min(1.0f, max(0.0f, duty));
-      uint32_t value = static_cast<uint32_t>((1 << bit_depth_) * duty_);
+      const uint32_t value = static_cast<uint32_t>((1 << bit_depth_) * duty_);
       ledcWrite(channel_, value);
     }
 
     bool isEnabled() const {
       return enabled_;
-    }
-
-    int8_t getPin() const {
-      return pin_;
-    }
-
-    uint32_t getFreq() const {
-      return freq_;
-    }
-
-    float getDuty() const {
-      return duty_;
     }
 
     void addToJsonState(JsonObject& pwmState) const {
@@ -84,13 +77,41 @@ class PwmOutput {
       getJsonValue(pwmState[F("duty")], duty_);
     }
 
+    void addToJsonInfo(JsonObject& user) const {
+      if (!enabled_)
+        return;
+      JsonArray data = user.createNestedArray("PWM pin " + String(pin_));
+      data.add(1e2f * duty_);
+      data.add(F("%"));
+    }
+
+    void addToConfig(JsonObject& pwmConfig) const {
+      pwmConfig[F("pin")] = pin_;
+      pwmConfig[F("freq")] = freq_;
+    }
+
+    bool readFromConfig(JsonObject& pwmConfig) {
+      if (pwmConfig.isNull())
+        return false;
+        
+      bool configComplete = false;
+      int8_t newPin = pin_;
+      uint32_t newFreq = freq_;
+      configComplete &= getJsonValue(pwmConfig[F("pin")], newPin);  
+      configComplete &= getJsonValue(pwmConfig[F("freq")], newFreq);  
+      
+      open(newPin, newFreq);
+
+      return configComplete;
+    }
+
   private:
     int8_t pin_ {-1};
     uint32_t freq_ {50};
-    uint8_t bit_depth_ = 12;
+    static const uint8_t bit_depth_ {12};
     uint8_t channel_ {255};
-    bool enabled_ {false};
     float duty_ {0.0f};
+    bool enabled_ {false};
 };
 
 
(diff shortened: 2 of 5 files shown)


usermods/pwm_outputs/readme.md
```diff
@@ -0,0 +1,21 @@
...
+## Configuration
+
+By default PWM outputs are disabled, navigate to Usermods settings and configure desired PWM pins and frequencies.
+
+## Usage
+
+If PWM output is configured, it starts to publish its duty cycle value (0-1) both to state JSON and to info JSON (visible in UI info panel). To set PWM duty cycle, use JSON api (over HTTP or over Serial)
+
+```json
+{"pwm_0": 0.0, "pwm_1": 0.1, ...}
```
blazoncek, line 20: Perhaps converting this to an array of objects would be better. I.e.: `{"PwmOutputs":[{"id":0,"duty":0.1},{...},...]}`
This would allow some *ugly* constructs below.

usermods/pwm_outputs/usermod_pwm_outputs.h
```diff
@@ -0,0 +1,169 @@
+#pragma once
+#include "wled.h"
+
```
blazoncek, line 3: Please add #error directive (enclosed in #ifdef) to inform users that usermod is only available for ESP32.

```c++
#ifndef ESP32
  #error This usermod does not support the ESP8266.
#endif
```

usermods/pwm_outputs/usermod_pwm_outputs.h
```diff
@@ -0,0 +1,169 @@
...
+        return;
+      pinManager.deallocatePin(pin_, PinOwner::UM_PWM_OUTPUTS);
+      if (channel_ != 255)
+        pinManager.deallocateLedc(channel_, 1);
+      channel_ = 255;
+      enabled_ = false;
+      duty_ = -1.0f;
+    }
+  
+    void setDuty(const float duty) {
```
blazoncek, line 47: Perhaps adding an overloaded function that will take `uint8_t` as a parameter would be beneficial so you can pass either 0.0-1.0 range or 0-255 (or 16-bit if you want).

usermods/pwm_outputs/usermod_pwm_outputs.h
```diff
@@ -0,0 +1,169 @@
...
+
+    void loop() {
+    }
+
+    void addToJsonState(JsonObject& root) {
+      for (int i = 0; i < USERMOD_PWM_OUTPUT_PINS; i++) {
+        const PwmOutput& pwm = pwms_[i];
+        if (!pwm.isEnabled())
+          continue;
+        root["pwm_" + String(i)] = pwm.getDuty();
```
blazoncek, line 100: Avoid `String` if possible and enclose strings in `F()` macro helper.
This *ugly* construct can be avoided if used with nicer JSON format (see above).

usermods/pwm_outputs/usermod_pwm_outputs.h
```diff
@@ -0,0 +1,169 @@
...
+    void addToJsonInfo(JsonObject& root) {
+      JsonObject user = root["u"];
+      if (user.isNull())
+        user = root.createNestedObject("u");
+
+      for (int i = 0; i < USERMOD_PWM_OUTPUT_PINS; i++) {
+        const PwmOutput& pwm = pwms_[i];
+        if (!pwm.isEnabled())
+          continue;
+        JsonArray data = user.createNestedArray("PWM pin " + String(pwm.getPin()));
```
blazoncek, line 125: Why not simply `sprintf()`?

usermods/pwm_outputs/usermod_pwm_outputs.h
```diff
@@ -0,0 +1,169 @@
...
+        data.add(1e2f * pwm.getDuty());
+        data.add(F("%"));
+      }
+    }
+
+    void addToConfig(JsonObject& root) {
+      JsonObject top = root.createNestedObject(USERMOD_NAME);
+      for (int i = 0; i < USERMOD_PWM_OUTPUT_PINS; i++) {
+        const PwmOutput& pwm = pwms_[i];
+        top["pin_" + String(i)] = pwm.getPin();
```
blazoncek, line 135: Similar to above, having nicer JSON will automatically group PWM channel parameters together in usermod settings.

usermods/pwm_outputs/usermod_pwm_outputs.h
```diff
@@ -0,0 +1,169 @@
...
+    uint32_t getFreq() const {
+      return freq_;
+    }
+
+    float getDuty() const {
+      return duty_;
+    }
+
+  private:
+    int8_t pin_ {-1};
```
blazoncek, line 74: Why not use initialiser in default constructor as you did in parametric construcor? It is cleaner.
Be also aware that parametric constructor doesn't initialise all properties after the change.

```c++
PwmOutput() : pin(-1), freq_(50), ... {}
```

usermods/pwm_outputs/usermod_pwm_outputs.h
```diff
@@ -0,0 +1,169 @@
...
+
+        int8_t newPin = pwm.getPin();
+        uint32_t newFreq = pwm.getFreq();
+        configComplete &= getJsonValue(top["pin_" + String(i)], newPin);  
+        configComplete &= getJsonValue(top["freq_" + String(i)], newFreq);  
+
+        // Recreate output if config has changed
+        if (newPin != pwm.getPin() || newFreq != pwm.getFreq()) {
+          pwm.close();
+          pwm = PwmOutput(newPin, newFreq);
```
blazoncek, line 154: This invokes default copy constructor. Right?
IMO not the best choice...

blazoncek (commented):
I am nitpicking as I found your solution quite amusing. Please do not be offended, the implementation is otherwise ok.

usermods/pwm_outputs/readme.md
ardiloot: Thank you for you comments! Separated PWM states to separate object:
```json
{"pwm": {"0": {"duty": 0.5123}, "1": {"duty": 0.1}}}
``` 
Used `pwm` instead of `PwmOutputs` to match current style. Also, used `id` as keys, as it is much easier to get current duty cycle value for specific output (e.g `state['pwm']['0']['duty']`)

usermods/pwm_outputs/usermod_pwm_outputs.h
ardiloot: Good point, made the change

usermods/pwm_outputs/usermod_pwm_outputs.h
ardiloot: Added this `uint8_t` overload, but ended up removing it again, as it was not used. It seemed, that supporting both 0.0-1.0 and 0-255 ranges would not bring much of benefit, but will complicate code and other logic (e.g. what to save in the state json and how to use different ranges when setting duty cycles). Also, for my application, 0-255 range is not suitable because of need for higher resolution.

usermods/pwm_outputs/usermod_pwm_outputs.h
ardiloot: good point, moved all strings to flash. Also, removed usage of `String` by replacing them by C-style strings. Not very happy about mixing C-code into C++, but don't know any other better alternative.

usermods/pwm_outputs/usermod_pwm_outputs.h
ardiloot: See previous comment

usermods/pwm_outputs/usermod_pwm_outputs.h
ardiloot: Fixed as suggested

usermods/pwm_outputs/usermod_pwm_outputs.h
ardiloot: With new logic there is no need for constructor at all, defaults stayed at variable declaration line.

usermods/pwm_outputs/usermod_pwm_outputs.h
ardiloot: The idea was to be very clear, that if pin or frequency is changed, then `PwmOutput` is reset to default state (i.e there cannot be any leftover state in some internal variables). Using assignment operator was one way to to that, the other way of dynamically allocating and deallocating new `PwmOutput` instance seemed worse option.

Made a change, so changing pin/frequency values will not require reconstructing `PwmOutput` instance (just calling `open(pin, freq)` method will do).

ardiloot:
Thank you for your thorough review! Let me know if you have any more ideas

usermods/pwm_outputs/usermod_pwm_outputs.h
blazoncek: Do not think only about your needs. Think as well what others may need as you are writing a public usermod.
I mentioned `uint8_t` as an example, you can use `uint16_t` with 16-bit resolution and scale it down to available 12-bit.
Someone may tap into your usermod in the future (look into my own PWM_fan usermod how I tapped into Temperature usermod).

Using floats is something I would recommend against if it can be done using integers. Just a personal opinion.

usermods/pwm_outputs/readme.md
blazoncek: I am not sure this will work with usermod settings implementation. Test it thoroughly, please.

usermods/pwm_outputs/usermod_pwm_outputs.h
blazoncek: Some other classes implement `begin()` which does the same.

usermods/pwm_outputs/usermod_pwm_outputs.h
ardiloot: Just to clarify, are you suggesting to support multiple ranges (0.0-1.0, 0-255, 0-65535) or just switch to (0-65535)? Switching to 16 bit range instead of float range is easy (and can do if you think it is better), however I am not sure how to support multiple ranges (should it be a configuration per output?). Only supporting 0-255 range is too limiting imo (for me and probably for others), as ESP32 PWM has resolution up to 16bits (future hw might have even more).

In general I am thinking to implement dynamic bit depth (number of steps are automatically selected based on the requested frequency), like in esphome: https://esphome.io/components/output/ledc.html#recommended-frequencies

usermods/pwm_outputs/readme.md
ardiloot: Yes, was not sure either, but tested it and everything works as expected. As a bonus, also pin conflict detection (red background at settings page) works.

usermods/pwm_outputs/usermod_pwm_outputs.h
blazoncek: I am only suggesting to implement integer function, if you need 16bit, so be it.

As for PWM resolution: depends on the PWM frequency. You are using a very low frequency which allows you to utilize full resolution. But if I want much higher frequency, resolution would decrease. (get the chart from Espressif)

ardiloot: I think I get it now (long wires I guess), added `void setDuty(const uint8_t duty)` and `void setDuty(const uint16_t duty)` overloads (in case some future dependent usermod will need them).

blazoncek: Good.
Though my experience with overloading integers like this can sometimes confuse compiler.
To test this add `setDuty(10)` somewhere in the code. You should see compiler complaining about ambiguous overload.
Perhaps a singe 16bit would be enough with info in readme that setDuty() supports float in range [0.0-1.0] and integer in range [0-65535].

ardiloot: good point, removed ambiguous overload

dosipod:
I only though of doing a shutter with this just now after i tested with a poor man robot arm ( i am sure not the intended use case ). I have to be honest this works in a reliable way that can easily compete with with something like Tasmota . In term of easy of use I give this 10 out of 10 , see the demo done with random number generator . We have also tested saving the api to presets and that works also.Shutters and blinds should be very simple to do as is .

I am not sure if there is a way to  control the speed and why the  use of float and not  integer for the position but otherwise excellent functionality that is not know by many 


image
https://user-images.githubusercontent.com/87589371/233036267-8f3b7e12-d930-4eb4-a52c-e4273b81c988.mov

PR #2925, opened 2022-12-05
added ble & wifi switch

johne:


usermods/Ble2Json_v2/usermod_v2_ble2json.h
```diff
@@ -0,0 +1,381 @@
+#pragma once
+
+#include "wled.h"
+// #include "ble.h"
```
blazoncek, line 4: If the file is not needed (is included below) remove it (and references to it) from PR.

usermods/Ble2Json_v2/WLED_ESP32_4MB_1MB_FS.csv
```diff
@@ -0,0 +1,4 @@
+# Name,   Type, SubType, Offset,  Size, Flags
```
blazoncek, line 1: Why this file?

usermods/Ble2Json_v2/usermod_v2_ble2json.h
```diff
@@ -0,0 +1,381 @@
...
+
+  public: 
+    MyCallbackHandler(Ble2JsonCallbacks *_userMod) {
+      userMod = _userMod;
+    }
+
+	// void onRead(BLECharacteristic *pCharacteristic) {
+	// }
+
+	void onWrite(BLECharacteristic* pCharacteristic) {
```
blazoncek, line 129: Indentation.

usermods/Ble2Json_v2/usermod_v2_ble2json.h
```diff
@@ -0,0 +1,381 @@
...
+     * 2. Try to avoid using the delay() function. NEVER use delays longer than 10 milliseconds.
+     *    Instead, use a timer check as shown here.
+     */
+    void loop() {
+      if (millis() - lastTime > 3000) {
+        // Serial.println("I'm alive!");
+        lastTime = millis();
+
+        if (configDirty) {
+          serializeConfig();
```
blazoncek, line 223: Where do you get actual configuration from?

softhack007:
Is `platformio_override.ini` really needed? The _override.ini is usually for custom build environments, so should not be a requirement to overwrite an existing override file. Also you run into a risk that WLED build targets evolve, and some day your override is outdated andwill not work any more.

I think it would be better to describe in the readme.md what are the changes needed to use this usermod, like
* additions to `build_flags`
* additions to `lib_deps`
If you are looking for an example, check the (exhaustive) readme of audioreactive usermod https://github.com/Aircoookie/WLED/blob/main/usermods/audioreactive/readme.md


If you still think that an override is needed, you could use the `extends` keyword:
```ini
[env:esp32dev_ble]
extends = env:esp32dev_qio80
build_flags = ${env:esp32dev_qio80.build_flags}
    -D your_special_flags

lib_deps = ${env:esp32dev_qio80.lib_deps}
      espressif/stuff_you_need @ version

[env:esp32S3_ble]
extends = env:esp32s3dev_8MB
build_flags = ${env:esp32s3dev_8MB.build_flags}
    -D your_special_flags

lib_deps = ${env:esp32s3dev_8MB.lib_deps}
      espressif/stuff_you_need_more @ OtherVersion

```

usermods/Ble2Json_v2/WLED_ESP32_4MB_1MB_FS.csv
softhack007: if really needed, then it would be better to add a new file into `tools/`, and reference it with `board_build.partitions = tools/WLED_ESP32_4MB_for_BLE.csv`
--> see my other comment about platformio_override.init.

Also, if you really need a different partitions table, then please also provide one for 8MB (typical flash size on -S3) and for 16MB (the future of ESP32).

usermods/Ble2Json_v2/ble.h
```diff
@@ -0,0 +1,100 @@
...
+
+    void onDisconnect(BLEServer* pServer) {
+      Serial.println(" - ServerCallback - onDisconnect");
+    }
+};
+
+class SecurityCallback : public BLESecurityCallbacks {
+
+  uint32_t onPassKeyRequest(){
+    return 000000;
```
softhack007, line 31: previously you use `#define PASSKEY 999999`. Is it correct to return 0 here? Is there any special reason to write `000000` instead of 0? or did you intend to return a string, or maybe a hex number?

softhack007:
> I am wondering if this is just a prototype or a fully functional usermods as I do not understand where do you handle JSON input (which is normally done using HTTP requests and is handled by AsyncWebServer). If so please change it to draft so we can discuss.
> If I am missing the point and this _is a functional_ PR then please add regular JSON processing as well as configuration.

Same question from my side. It seems that this usermod allows to create a BLE connection, can it also do other things, or maybe can it be configured to do things when a JSON file is provided, similar to IR remote maybe?

wled00/wled.h
```diff
@@ -769,6 +769,9 @@ WLED_GLOBAL volatile uint8_t jsonBufferLock _INIT(0);
 #define W(c) (byte((c) >> 24))
 
 class WLED {
+private:
```
softhack007, line 772: Does it make sense to have these methods also without BLE usermod? 
If not, please put them between `#ifdef USERMOD_BLE_2_JSON` ... `#endif`

wled00/wled.cpp
```diff
@@ -603,6 +603,7 @@ void WLED::initConnection()
   ws.onEvent(wsEvent);
   #endif
 
+  if (wifiDisabled) return; // can still use ethernet
```
softhack007, line 606: Please put this between `#ifdef USERMOD_BLE_2_JSON` ... `#endif` --> see my comment on wled.h.

wled00/wled.cpp
```diff
@@ -864,3 +865,12 @@ void WLED::handleStatusLED()
...
 }
+
+void WLED::disableWiFi() {
+  wifiDisabled = true;
+  WiFi.disconnect(true);
+}
+
+void WLED::enableWiFi() {
+  wifiDisabled = false;
+}
```
softhack007, line 876: Please put them between `#ifdef USERMOD_BLE_2_JSON` ... `#endif` --> see my comment on wled.h.

softhack007 (changes requested):

softhack007:
btw, great work, thanks 👍 
The fact I'm making so many requests means I really like what you did, and I'm curious to see how this will evolve :-)

wled00/wled.h
ChuckMash, line 772: I'm working on 3 usermods that require this sort of thing, one of which is this exact feature set, bluetooth json api, by means of a different method. This PR beat me to the submission.
So I will say that locking this behind a single usermod define would be troublesome for other usermods.

wled00/wled.cpp
ChuckMash: > Please put this between `#ifdef USERMOD_BLE_2_JSON` ... `#endif` --> see my comment on wled.h.

I'm working on 3 usermods that require this sort of thing, one of which is this exact feature set, bluetooth json api, by means of a different method.
This PR beat me to the submission. So I will say that locking this behind a single usermod define would be troublesome for other usermods.

ChuckMash: > I'm not sure if disabling Wifi should be part of the "standard WLED", as it could (possibly) render the device unusable (no more access to web UI, no AP could be opened), and a factory reset would be needed to "revive" WLED? Or is there an existing "kill switch" already to shut down BLE and start Wifi again?
> 
> @blazoncek what would you recommend? Maybe create a new flag like `WLED_SUPPORT_WIFI_OFF` ?

My current implementation has serial overrides when the option is compiled in, and the usermod that uses these features should have additional overrides. Getting locked out of wifi should never be an option.
I was going to put in a bluetooth serial JSON api usermod in today, but then I saw this.

blazoncek: Even though BLE could substitute WiFi connection (in theory) none of the underlying libraries are up to it AFAIK.
So unless the full push/pull API for BLE is implemented somewhere in addition to wled_server.cpp for WiFi, disabling WiFi should not be done in usermod but should better be a standard WLED setting.

Though, if properly implemented I have no objections if it is done within usermod.

blazoncek:
@johne please provide more information about your PR.
And change it to draft as it seems you are still working on it.

softhack007:
Converted to draft

johne:
I am so sorry.. I did not mean for this to become a pr yet... I guess just pushing to my branch automatically created it... that said, it is pretty close to done now... the only part of the json api I haven't implemented yet is palette details... the front end degrades very nicely without that data though... my plan is to implement that and be done with it... in the app I wrote: https://github.com/johne/WledAppV2 I just hide the "config" tab... but there is a way in the app to switch back to wifi. meaning you just switch back and then you can get to the config tab... the ble doesn't support the syncing through udp, which is why, for my home, I want to find a > 4mb esp32 with ethernet... so I can have ota _and_ syncing... I don't need it for my trailer, however, cause when i'm home I can just switch to wifi to configure, but access all other front end features through ble... I will address the comments above inline.

usermods/Ble2Json_v2/ble.h
```diff
@@ -0,0 +1,100 @@
...
+#define CHARACTERISTIC_UUID "beb5483e-36e1-4688-b7f5-ea07361b26a8"
+#define PASSKEY 999999
+
+/////////////////////
+//BLE Secure Server//
+/////////////////////
+
+class ServerCallback: public BLEServerCallbacks {
+    void onConnect(BLEServer* pServer) {
+      Serial.println(" - ServerCallback - onConnect");
```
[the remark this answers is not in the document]
johne: I will make the changes on the strings... I already created a set of BLE_DEBUG so I can turn off and on without the noise of the other debug prints.

usermods/Ble2Json_v2/usermod_v2_ble2json.h
```diff
@@ -0,0 +1,381 @@
...
+    //Private class members. You can declare variables and functions only accessible to your usermod here
+    unsigned long lastTime = 0;
+
+    bool bleOnFlag = false;
+    uint32_t blePairingPin = PASSKEY;
+
+    bool bleInitted = false;
+    bool configDirty = false;
+
+    void checkBleInit(bool fromSetup) {
```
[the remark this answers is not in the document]
johne: I think this has changed... again, sorry that I put the PR up prematurely

usermods/Ble2Json_v2/WLED_ESP32_4MB_1MB_FS.csv
johne: the program fits in the provided 8mb partition scheme... but I will move that file to tool... I just wanted to keep my stuff as isolated as possible.

usermods/Ble2Json_v2/usermod_v2_ble2json.h
johne: I think i've addressed all the formatting issues... I installed prettier and have auto-format on... then, when I edit an existing wled file, I just edit in `vi` so as not to invoke the auto-formatting

usermods/Ble2Json_v2/usermod_v2_ble2json.h
johne: the changes happens in response to seeing a `"bleToggle":true` in the state... it is a transient state flag that toggles between ble and wifi mode.

usermods/Ble2Json_v2/ble.h
johne: there are two passkeys... one from the server and one from the client... we don't care about the server's passkey, only the one from the client... so returning 0 is fine for the server (esp32 device).... I formatted it as all zeros because I copied this code from somewhere... I can change it to just one zero.  the "real" default passkey is all 9's... but that is another thing one can change in the config (though, prolly should be added to the "secure" config... in the future)

usermods/Ble2Json_v2/readme.md
```diff
@@ -0,0 +1,10 @@
+# Usermods API v2 ble 2 json usermod
+
+this usermod exposes all the json 
+
+## Installation
```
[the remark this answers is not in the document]
johne: the readme (and all the other requested changes) will be my next effort... likely this weekend

wled00/wled.h
johne: well... it can if there are other usermods that want to be able to disable wifi during the "setup" phase of booting. I certainly don't mind putting them in an `#ifdef`, though

wled00/wled.cpp
johne: i'm not sure where this leaves this part of the code... if you are suggesting there should be a standard wled config for shutting off wifi, I would argue against that as that flag (without this mod or one like it) would render the device un-reachable. with this mod (and other potential mods), one can then turn back on wifi through a ble call to set the state like this:

```
{"bleToggle":true}
```

I think maybe we do as @ChuckMash suggests and use `WLED_SUPPORT_WIFI_OFF` ... then we can state in the readme to make sure to define that flag in the build. keep in mind, they way i've implemented it here, you can only turn wifi off at `setup` time... so the mod has to be functioning at boot to set wifi off... if you install another version of wled without the mod, the wifi will go back to normal (which is what we'd want)

wled00/wled.cpp
johne: as per the convo above, i'm not sure what the community wants here... there seems to be conflicting opinions above

johne:
> > I am wondering if this is just a prototype or a fully functional usermods as I do not understand where do you handle JSON input (which is normally done using HTTP requests and is handled by AsyncWebServer). If so please change it to draft so we can discuss.
> > If I am missing the point and this _is a functional_ PR then please add regular JSON processing as well as configuration.
> 
> Same question from my side. It seems that this usermod allows to create a BLE connection, can it also do other things, or maybe can it be configured to do things when a JSON file is provided, similar to IR remote maybe?

yeah... this is due to the fact that I accidentally opened this PR too early... now, if you look at it, you will see the (almost) full implementation of the json api... the only thing it isn't supporting yet (but will soon) is the palette details... I even have a front end that supports all the features (but config) the current website does (in fact, the react-native app I wrote downloads the matching release from github to use the _exact_ same interface)

usermods/Ble2Json_v2/usermod_v2_ble2json.h
johne: thx... I think I cleaned up all commented out code, but I will double check

johne:
here are some screenshots of the app working in ble mode... as I was gathering the shots, I realized the "info" tab doesn't work... so i'll have to figure out why (i'm guessing I just missed handling one of the json api endpoints in the app)... other than that, every screens works as normal to control the lights.... 
IMG_8513
IMG_8512
IMG_8511
IMG_8510
IMG_8509

johne:
the app also works in wifi mode as well (very similar to the current wled app)

johne:
also, it supports the "websocket" type mode of getting updates to state based on changes from other sources (another connected ble device, a playlist that makes timed changes, etc)

blazoncek:
Please remove `platformio_override.ini` from the PR. It is intentionally not included in main (or any other) branch.

bbusse:
I consider disabling wifi and/or ble a core functionality and would like to see both in WLED rather than a usermod. 

We use buttons to control our device. At times we want to disable wifi and/or ble, also with buttons. Since we have buttons to steer the device, we don't need any other remote means for controlling WLED but sometimes we want to use the web ui or mobile app. So we use a button to switch it on again

So saying this would render WLED unusable is certainly not right. WLED works perfectly fine without a wireless connection and valid use cases exist

Further on the plus side it does reduce both attack surface and power consumption While reduction of power consumption is always a worthwhile goal also consider that some devices run on battery

Funkelfetisch:
this seems so close to completion, can I help?

Moustachauve:
Have you considered using NimBLE for BLE library? it seems more efficient

PR #2926, opened 2022-12-05
Bluetooth Serial JSON Control and WiFi Deferment

ChuckMash:
Due to a very similar PR (#2925) also today, I will open this a draft to encourage discussion between the 2 similar implementations.

* Allows access to JSON API over Bluetooth Serial Connection.
* __Normal WiFi functionality is disabled while Bluetooth is in use. OTA disabled due to size constraints__
* Bluetooth device shows up as whatever is set in "Server description", "WLED" by default.
* Normal WiFi functionality can be accessed for general WLED setup under the following conditions
  * Normal WiFi functionality until "Apply preset at boot" is set.
  * If active preset changes by any means during first 5 seconds, normal WiFi functionality will resume. (simple button press would suffice)
  * If 'e' is sent over the bluetooth serial connection, normal WiFi functionality will resume.
* [Serial Bluetooth Terminal](https://play.google.com/store/apps/details?id=de.kai_morich.serial_bluetooth_terminal) Android app for testing
* [Reference: Python to ESP32 bluetooth serial](https://medium.com/@18218004/devlog-6-bluetooth-and-esp32-ba076a8e207d)
* [Example Python Script](https://gist.github.com/ChuckMash/7752b77ea1bf204a7e8f0acac6e27641) for WLED control over Bluetooth Serial

### DIFF
--- a/tools/WLED_ESP32_NO_OTA.csv
+++ b/tools/WLED_ESP32_NO_OTA.csv
@@ -1,0 +1,5 @@
+# Name,   Type, SubType, Offset,  Size, Flags
+nvs,      data, nvs,     0x9000,  0x5000,
+otadata,  data, ota,     0xe000,  0x2000,
+app0,     app,  ota_0,   0x10000, 0x200000,
+spiffs,   data, spiffs,  0x210000,0x1F0000,
--- a/usermods/bluetooth_serial/bluetooth_serial.h
+++ b/usermods/bluetooth_serial/bluetooth_serial.h
@@ -1,6 +1,7 @@
 #pragma once
 #include "wled.h"
 #include "BluetoothSerial.h"
+#include "esp_bt_device.h"
 
 BluetoothSerial SerialBT;
 
@@ -17,10 +18,21 @@ class BluetoothSerialUsermod : public Usermod {
       SerialBT.begin(serverDescription);
 
       Serial.println("Bluetooth Serial Usermod setup complete");
+
+      // The name the bluetooth device will appear as, same as set in server description
       Serial.print("Bluetooth Name: ");
       Serial.println(serverDescription);
+
+      // The local bluetooth device address
       Serial.print("Bluetooth Address: ");
-      Serial.println("todo");
+      const uint8_t* point = esp_bt_dev_get_address();
+      for (int i = 0; i < 6; i++) {
+        char str[3];
+        sprintf(str, "%02X", (int)point[i]);
+        Serial.print(str);
+        if (i < 5) Serial.print(":");
+      }
+      Serial.println();
     }
 
 
--- a/usermods/bluetooth_serial/bluetooth_serial.h
+++ b/usermods/bluetooth_serial/bluetooth_serial.h
@@ -9,6 +9,7 @@ class BluetoothSerialUsermod : public Usermod {
  
   public:
 
+
     void setup() {
 
       // Tell WLED to defer all WiFi Connectivity
--- a/usermods/bluetooth_serial/bluetooth_serial.h
+++ b/usermods/bluetooth_serial/bluetooth_serial.h
@@ -17,6 +17,10 @@ class BluetoothSerialUsermod : public Usermod {
       SerialBT.begin(serverDescription);
 
       Serial.println("Bluetooth Serial Usermod setup complete");
+      Serial.print("Bluetooth Name: ");
+      Serial.println(serverDescription);
+      Serial.print("Bluetooth Address: ");
+      Serial.println("todo");
     }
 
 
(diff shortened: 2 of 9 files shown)


tools/WLED_ESP32_NO_OTA.csv
```diff
@@ -0,0 +1,5 @@
+# Name,   Type, SubType, Offset,  Size, Flags
```
blazoncek, line 1: Is the code size >1.5MB?

usermods/bluetooth_serial/bluetooth_serial.h
```diff
@@ -0,0 +1,90 @@
...
+
+    void setup() {
+
+      // Tell WLED to defer all WiFi Connectivity
+      deferConnections = true;
+      
+      // Begin Bluetooth device with same name as server description
+      SerialBT.begin(serverDescription);
+
+      Serial.println("Bluetooth Serial Usermod setup complete");
```
blazoncek, line 21: Use DBUG_PRINTLN() and use flash strings F()

usermods/bluetooth_serial/bluetooth_serial.h
```diff
@@ -0,0 +1,90 @@
...
+      Serial.println();
+    }
+
+
+    // Cleans up and lets WLED run WiFi normally.
+    // Once called, this usermod is disabled until next boot
+    void stop(){
+      SerialBT.flush();
+      SerialBT.end();
+      deferConnections=false;
```
blazoncek, line 45: Perhaps also add `forceReconnect = true;`.

usermods/bluetooth_serial/bluetooth_serial.h
```diff
@@ -0,0 +1,90 @@
...
+    void loop() {
+
+      // If Wifi Deferment has ended, do nothing
+      if(!deferConnections){
+        return;
+      }
+
+      // If not set up or if preset changes in first 5* seconds of boot, disable Bluetooth and end WiFi deferment
+      unsigned long now = millis();
+      if(now > 1000 && now < 6000 && (currentPreset != bootPreset || currentPreset == 0 && bootPreset == 0)){
```
blazoncek, line 59: Please add parentheses to clear ambiguous condition.

blazoncek (changes requested):
I am wondering how do you pair? If it is needed.
Otherwise minor issues but I would like them fixed.

tools/WLED_ESP32_NO_OTA.csv
ChuckMash: > Is the code size >1.5MB?

Yes, 1.8MB+ with the compile options in the usermod readme

ChuckMash:
> I am wondering how do you pair? If it is needed.

For example if you would want to use this to access the JSON API from an Android device you would find the device under available devices with the WLED server description name, the pairing is automatic from there.

tools/WLED_ESP32_NO_OTA.csv
blazoncek: No OTA will be a hard sell.

blazoncek:
> For example if you would want to use this to access the JSON API from an Android device you would find the device under available devices with the WLED server description name, the pairing is automatic from there.

So something like mouse or keyboard? I would still like to see some pairing option with PIN if possible.

ChuckMash:
>  I would still like to see some pairing option with PIN if possible.

Agreed, I've been looking into it

usermods/bluetooth_serial/bluetooth_serial.h
ChuckMash: Is your thinking here to replace all serial out with DEBUG equivalent or just this message?
Getting things like the bluetooth device name and hardware address can be useful, but I can see how they may belong as a debug message instead.

blazoncek: Avoid using Serial.print() for printing debug messages. Use it only if you intend output to Serial interface.
Debug can go to networked recipient.

ChuckMash:
That should deal with the bulk of the talking points so far, with exception of the pairing process which I am still looking into.

wled00/wled.cpp
```diff
@@ -716,6 +716,10 @@ void WLED::initInterfaces()
 
 void WLED::handleConnection()
 {
```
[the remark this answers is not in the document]
johne: tried it on a wt32-eth01 device... it didn't work... I set up 2 characteristics per json endpoint and it gets through about 1/2 of them before rebooting... i'm guessin the eth stack takes up some memory which doesn't leave enough for my other characteristics... I could route all calls through 1 set of characteristics and just include something in the json to id what endpoint it is in response to... that might help save a bit on memory (if that is the problem)... I might put in some debug prints to show the memory growing as I allocate characteristics to confirm that memory is the problem

ChuckMash:
Stalebot reminded me, I don't have the time at the moment to finish this up.
Back when I was working on it there was upstream issues with bluetooth serial and being able to set a pairing pin.
Some time later I may re-open this draft and continue.

PR #2933, opened 2022-12-08
Add overlay effects /Glitter and Sparkle/

drhideg:
- add glitter and sparkles overlay effects
- rearrange glitter and solid glitter in glitter_base
- some whitespace removed

It can be disabled with `#define WLED_DISABLE_OVERLAY`.
This overlay effects has no backround palette or color, only the FX itself.
I am using it with overlapped segments where background come from segment 0 and effect from segment 1. Like Colorwaves+Glitter or Breathe+Sparkle.


blazoncek:
Thanks, unfortunately whitespace cleanup make PR hard to review.

I have a recommendation: Use new possibility of 3 additional options that are available for every effect. You can then use your `glitter_base()` on top of every effect without need to recompile.

I can elaborate if you do not understand.
Screenshot 2022-12-08 at 17 15 06

drhideg:
If I think correctly I need to add the c1|c2|c3 (which is not used) for every single effect from [JSON API](https://kno.wled.ge/interfaces/json-api/#effect-parameters) and modify like this:
```diff
-static const char _data_FX_MODE_COLORWAVES[] PROGMEM = "Colorwaves@!,Hue;!;!";
+static const char _data_FX_MODE_COLORWAVES[] PROGMEM = "Colorwaves@!,Hue,,,,Enable Glitter;!;!";
```
and add something like this to the end of the effects:
```diff
+if (SEGMENT.check1) {
+  glitter_base(0, SEGCOLOR(0));
+}
```
I think i will loose the ability to change the glittering pixels color and to set the effect speed and intensity.

If you thought about something like I did above :)

blazoncek:
> If you thought about something like I did above :)

Yes, except use 2nd or 3rd checkmark (perhaps star) as 1st is usually meant for color/palette.

drhideg:
Reworked as as we were talking about for Breathe, Colorwaves, Colortwinkles and Aurora.

Aircoookie:
Like the idea of overlay effects!
I think I do prefer the original overlaying segments approach though...

Benefits of using checkmarks:
- No additional segments needed
- Easier to setup

Benefits of using extra segment:
- Can adjust speed and intensity of glitter separately
- Possible to cover only part of the effect with glitter
- No (large!) increase in FX data size
- Does not occupy a checkmark slot
- Possible to add further overlay effects in the future
- No need to add code to _every_ effect that should support the overlay

blazoncek:
> Benefits of using extra segment:

There are drawbacks as well. 😉 
Complex to set up for beginners and people not involved.

drhideg:
I vote for the first one. It covers what I miss and more customizable.
I would like to arrange it at the end of the effect list, if that's not a problem.

Aircoookie:
Another idea would be not adding a new dedicated effect, but having the regular Glitter effect check the previous segment and generate just glitter without the palette background if there is a (partial) overlap between the previous segment and the Glitter segment.

drhideg:
That would be good, but there is a lot of question and the implementation is above my knoWLEDge.
2D and ledmaps,  2 segments with even and odd leds with spacing, 2 non-overlapping segment and a 3rd is overlap with 1st.

Probably the best I can do, is Glitter and Sparkles got an FX only checkox where no palette/color used and let the user decide does it suit her/him.

blazoncek:
Adding glitter overlay using `checkmark 2` (layer icon) and `custom 3` slider can be added to almost any effect with minimal updates of those effects. Assuming those don't already use checkmark 2 and custom 3.

The only change in any of the effects would be:
```c++
...
if (SEGMENT.check1) glitter_overlay(); // just before return FRAMTIME;
...
static const char _data_FX_... = "Effect@!,!,...,...,Glitter intensity,...,Glitter,...;..."
```
And glitter_overlay() would simply be:
```c++
void glitter_overlay() {
  if (SEGMENT.custom3 > random8(31)) {
    if (SEGMENT.is2D()) {
      SEGMENT.setPixelColorXY(random16(SEGMENT.virtualWidth()),random16(SEGMENT.virtualHeight()), ULTRAWHITE);
    } else {
      SEGMENT.setPixelColor(random16(SEGLEN), ULTRAWHITE);
    }
  }
}
```
The drawback is that stacking different overlays is not possible with this approach.

blazoncek:
There is no point in recent commits by adding 2D to glitter_base() the idea was to have independent overlay function that can serve 1D and 2D effect functions.
I cannot approve this change.

drhideg:
Of course.

PR #2942, opened 2022-12-11
New Usermod: SHT temperature & humidity sensors

ezcGman:
Hey there,

I don't think we have enough temperature sensors, do we? ;) WLED already has support for the Dallas temp sensors (e.g. used on the QuinLED Dig-Uno and Dig-Quad), DHT temp sensors and now I also add support for various SHT i2c sensors, which are used on the new QuinLED Dig-Octa, as also on the An-Penta.

Some facts:
* Supports four different SHT sensors
* Uses external lib: https://github.com/RobTillaart/SHT85
* Of course it uses the new global i2c pins
* Also publishes the values via MQTT, *and*
* Also add MQTT device discovery for home assistant!

Please let me know what you think and where I can improve :)

Thanks for considering this mod and greetings,

Andy!


usermods/sht/usermod_sht.h
```diff
@@ -0,0 +1,312 @@
...
+      PinManagerPinType pins[2] = { { i2c_sda, true }, { i2c_scl, true } };
+      if (!pinManager.allocateMultiplePins(pins, 2, PinOwner::HW_I2C)) {
+        DEBUG_PRINTF("[%s] SHT pin allocation failed!\n", _name);
+        shtInitDone = false;
+        cleanupShtTempHumiditySensor();
+        cleanup();
+        return;
+      }
+
+      #ifdef USERMOD_SHT_TYPE_SHT31
```
blazoncek, line 38: Can you make this runtime configurable?
It will require de-init on parameter change but I can see you already implemented de-init.
Example of usermod config drop-down list can be seen in 4LD ALT (`appendConfigData()`).
You may as well query I2C to detect sensor type if this is available.

usermods/sht/usermod_sht.h
```diff
@@ -0,0 +1,312 @@
...
+      }
+      // Check if mod has been en-/disabled
+      else if (enabled != oldEnabled) {
+        enabled ? setup() : cleanup();
+        DEBUG_PRINTF("[%s] Usermod has been en-/disabled\n", _name);
+      }
+      // Config has been changed, so adopt to changes
+      else if (enabled) {
+        if (oldHaMqttDiscovery != haMqttDiscovery && haMqttDiscovery) {
+          publishHomeAssistantAutodiscovery();
```
blazoncek, line 198: At this stage (during boot invocation) WiFi may not yet be connected.

usermods/sht/usermod_sht.h
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    void addToJsonInfo(JsonObject& root)
+    {
+      if (!enabled && !isShtReady()) {
+        return;
+      }
+
+      JsonObject user = root["u"];
+      if (user.isNull()) user = root.createNestedObject("u");
```
blazoncek, line 214: For some strings use F() helper to store them in flash memory (not needed for "u" but others may benefit).

usermods/sht/usermod_sht.h
```diff
@@ -0,0 +1,312 @@
...
+
+    void appendDeviceToMqttDiscoveryMessage(JsonDocument& root) {
+      JsonObject device = root.createNestedObject("dev");
+      device["ids"] = escapedMac.c_str();
+      device["name"] = serverDescription;
+      device["sw"] = versionString;
+      device["mdl"] = ESP.getChipModel();
+      device["mf"] = "espressif";
+    }
+
```
blazoncek, line 297: Please add `getTemperatureC()` and `getTemperatureF()` public data exchange functions so that temperature data may be used in other usermods (check Temperature (Dallas) usermod).
You may as well add humidity functions.

usermods/sht/usermod_sht.h
```diff
@@ -0,0 +1,312 @@
...
+      jsonTemp.add(shtCurrentTemp);
+      jsonTemp.add(" °C");
+    }
+
+    void publishTemperatureAndHumidityViaMqtt() {
+      if (!WLED_MQTT_CONNECTED) return;
+      char buf[128];
+
+      sprintf_P(buf, PSTR("%s/temperature"), mqttDeviceTopic);
```
blazoncek, line 245: Make sure buffers are adequate bu using `snprintf()` or equivalent bounds check..

usermods/sht/usermod_sht.h
```diff
@@ -0,0 +1,312 @@
...
+      */
+    uint16_t getId()
+    {
+      return USERMOD_ID_SHT;
+    }
+};
+
+// strings to reduce flash memory usage (used more than twice)
+// Config settings
+const char ShtUsermod::_name[]    PROGMEM = "SHT Temperature & Humidity Sensor";
```
blazoncek, line 310: That is a **very** long name that gets stored in `cfg.json`. Please use shorter string.

usermods/sht/usermod_sht.h
```diff
@@ -0,0 +1,312 @@
...
+const char ShtUsermod::_enabled[] PROGMEM = "Enabled";
+const char ShtUsermod::_haMqttDiscovery[] PROGMEM = "Add-To-Home-Assistant-MQTT-Discovery";
```
blazoncek, line 312: Same as above.
There is no need to capitalize first letters as it is done automatically.
I will add dash (-) to space handling there as well.

blazoncek (changes requested):
Thank you for excellent contribution.
I do have a few requests for change as it may benefit the usability.

As for global I2C there were a few complaints that sharing I2C may cause issues. Please test that with 4LD ALT usermod (or any other that uses I2C). Perhaps other usermods may need a rewrite to make them work (a separate PR).

Otherwise the code is clean and well structured.

ezcGman:
Thanks Blaz!! I'll try checking out tonight with other i2c mods. I think I even have a four line display laying around here, so can actually test! Will let you know!

usermods/sht/usermod_sht.h
ezcGman: Hrrmm, buuuut: `publishHomeAssistantAutodiscovery` checks for `WLED_MQTT_CONNECTED`, which implies network availability, right? :)

I mostly tend to have methods checking their requirements themselves, so I can just call them and they check themselves if they can safely execute.

usermods/sht/usermod_sht.h
ezcGman: Ah yeah, sure. I actually had that on my list and forgot, incl. Fahrenheit support! Will add!

blazoncek:
You may as well follow the guideline I posted on Discord #dev-general. Any function defined within class definition is automatically `inline`.
So try to move function definitions out of class definition.

ezcGman:
@blazoncek Except for testing the mod with other mods using the global i2c pins (like 4LD), all feedback should have been applied. Will try to test the mod with the 4LD mod the next days and report back!

But please, if you have some time already, check if your feedback has been properly addressed / solved :)

THX!

blazoncek (approved):
Excellent!

ezcGman:
OK, so I have been testing a bit with also compiling with the 4LD ALT mod. Sadly, I don't have an i2c 4LD, but an SPI one... So couldn't really test having two i2c devices running simultaneously :/

However, I added a bit of debugging and see what the `i2cAllocCount` does when en-/disabling both mods. And I really got weird results: Most of the time, the counter kept increasing with every en- or disable; super weird. Tested it happend with both mods together and also with both compiled exclusively. Something at the core must cause this. But I think this deserves its own issue and tracking; don't think this should block the mod really :/

I also saw that the `i2cAllocCount` is only in-/decremented if you use the allocateMultiple / deallocateMultiple methods, but not the single de-/allocate methods. Is that intentional?

However, I made a few changes to how I de-/allocate pins.

Hope this concludes that mod and is good to go :) THX!!

blazoncek:
> I also saw that the `i2cAllocCount` is only in-/decremented if you use the allocateMultiple / deallocateMultiple methods, but not the single de-/allocate methods. Is that intentional?

Yes, that is intentional and mandatory.

usermods/sht/usermod_sht.h
```diff
@@ -0,0 +1,491 @@
...
+ *
+ * @see Usermod::setup()
+ * @see UsermodManager::setup()
+ *
+ * @return void
+ */
+void ShtUsermod::setup()
+{
+  if (enabled) {
+    PinManagerPinType pins[2] = { { i2c_sda, true }, { i2c_scl, true } };
```
softhack007, line 247: I think you need to check explicitly for (i2c_sda >= 0) && (i2c_scl >= 0)
Because pinmanager accepts -1 and will return  true.
@blazoncek please confirm if I understand that correctly.

usermods/sht/usermod_sht.h
```diff
@@ -0,0 +1,491 @@
+#pragma once
+
```
softhack007, line 2: Better to add #include <Wire.h> here, so the SHT85 driver uses the "correct" Wire driver.

usermods/sht/usermod_sht.h
ezcGman: Oh yeah, just checked, good catch! Will fix that tonight!

usermods/sht/usermod_sht.h
ezcGman: Just curious: In which cases would it not / how can situation arise that it would use the wrong one?

softhack007: Well it would be possible that the sht85 includes its "preferred" i2c Wire implementation, if it does not find the global one. But this is just a mesure of precaution.

softhack007: Update: seems not strictly needed. usermod_list.cpp includes wled. h, which already include Wire.h

https://github.com/Aircoookie/WLED/blob/b94e0ef797b0c93fc97f9c9b194ec6b2e41a1999/wled00/wled.h#L89

So you don't need to do that again.

usermods/sht/usermod_sht.h
blazoncek: > Because pinmanager accepts -1 and will return true.

Yes, correct.

blazoncek:
@softhack007 if everything is ok with you, I am ok merge it.

usermods/sht/usermod_sht.h
softhack007: Just waiting for this check to be added, otherwise looks very good 👍

softhack007:
@blazoncek Just waiting for the pin=-1 check to be added, otherwise looks very good 👍

usermods/sht/usermod_sht.h
ezcGman: Now checking if they're smaller than zero additionally to the `allocateMultiplePins()` check

ezcGman:
@blazoncek @softhack007 Should be done :)

softhack007 (approved):
Thanks, I approve.

PR #2946, opened 2022-12-12
Use "pd" JSON API call for direct preset apply

Aircoookie:
Simplifies preset application code by using different commands for the two preset application methods:
- `"pd"` sets preset with contents included in the same API request
- `"ps"` now always async loads preset from filesystem

Fixes two minor bugs:
- `"ps"` (async apply preset) not working if other state changes (not part of the preset contents) are also present in the same API command
- no WS broadcast on direct preset change unless brightness changes (since `stateChanged` is reset to false to prevent currentPreset being reset to 0 by `stateUpdated()`)

Slight optimizations in `presets.cpp` - key removal is no longer always needed with async saving.


blazoncek:
I briefly reviewed the code and have to say I'll need to take a second look to understand all changes.
The missing WS broadcast should (perhaps) be handled in `notify()` as the e629c90a71587d08991a3912bb4f75d95faad1f3 added UDP broadcast that was also missing (correctly identified root cause).
I agree that simplifying code is the way to go but am not sure if async "ps" will not break integrations.

blazoncek (approved):
All seems in order to me. 😄

blazoncek:
Will you update WLED-docs as well? And/or mark "pd" internal?

PR #2951, opened 2022-12-14
support the ESP32-C3-12F module from AI-Thinker and associated boards

andyshinn:
An addition to support the Ai-Thinker ESP32-C3-12F module which has 2MB flash and needs a special board and partition layout. Board is not officially supported upstream yet (https://github.com/platformio/platform-espressif32/issues/806) so a custom board was added.

In Discord it sounded like we didn't want this to be a primary supported board, yet. But opening this PR early so others can see it if they want to pull it in to compile.

### DIFF
--- a/boards/esp32-c3-12f.json
+++ b/boards/esp32-c3-12f.json
@@ -1,0 +1,36 @@
+{
+  "build": {
+    "arduino":{
+      "ldscript": "esp32c3_out.ld"
+    },
+    "core": "esp32",
+    "f_cpu": "160000000L",
+    "f_flash": "80000000L",
+    "flash_mode": "dio",
+    "extra_flags": [
+      "-DARDUINO_USB_MODE=1"
+    ],
+    "mcu": "esp32c3",
+    "variant": "esp32c3"
+  },
+  "connectivity": [
+    "wifi"
+  ],
+  "debug": {
+    "openocd_target": "esp32c3.cfg"
+  },
+  "frameworks": [
+    "arduino",
+    "espidf"
+  ],
+  "name": "Ai-Thinker ESP32-C3-12F",
+  "upload": {
+    "flash_size": "2MB",
+    "maximum_ram_size": 327680,
+    "maximum_size": 2097152,
+    "require_upload_port": true,
+    "speed": 460800
+  },
+  "url": "https://www.electrodragon.com/product/esp-led-strip-board/",
+  "vendor": "Ai-Thinker"
+}
--- a/platformio.ini
+++ b/platformio.ini
@@ -195,7 +195,7 @@ build_flags =
 ; restrict to minimal mime-types
   -DMIMETYPE_MINIMAL
 
-lib_deps = 
+lib_deps =
   ${env.lib_deps}
   #https://github.com/lorol/LITTLEFS.git
   ESPAsyncTCP @ 1.2.2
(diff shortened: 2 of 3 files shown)


platformio.ini
```diff
@@ -641,3 +641,9 @@ lib_deps =
   ${esp32.lib_deps}
   TFT_eSPI @ ^2.3.70
 board_build.partitions = ${esp32.default_partitions}
+
+[env:esp32-c3-12f]
+extends = env:esp32c3
+board = esp32-c3-12f
+platform = espressif32@5.2.0
```
softhack007, line 648: Is this strictly needed? Will it also work with `platform = espressif32@5.1.1` ?
I'm asking because espressif32@5.1.1 has been tested also with other new boards, and there is currently more confidence in the 5.1.1.

boards/esp32-c3-12f.json
```diff
@@ -0,0 +1,36 @@
...
+  "debug": {
+    "openocd_target": "esp32c3.cfg"
+  },
+  "frameworks": [
+    "arduino",
+    "espidf"
+  ],
+  "name": "Ai-Thinker ESP32-C3-12F",
+  "upload": {
+    "flash_size": "2MB",
```
softhack007, line 28: Am i correct that the "flash_size" and "name" lines are the only difference to `esp32-c3-devkitm-1`

maybe to make the whole thing more generic, lets rename it to something like esp32-c3-devkitm_2M?
@blazoncek what do you think?

tools/WLED_ESP32_2MB_C3.csv
```diff
@@ -0,0 +1,5 @@
+# Name,   Type, SubType, Offset,  Size, Flags
+nvs,      data, nvs,     0x9000,    20K,
+otadata,  data, ota,     0xe000,    8K,
+app0,     app,  ota_0,   0x10000,   1536K,
+spiffs,   data, spiffs,  0x190000,  384K,
```
softhack007, line 5: just a small file renaming here - to be more consistent with other partition tables - 
please name this file `WLED_ESP32-C3_2MB.csv`

softhack007 (changes requested):
see comments to individual files.

platformio.ini
blazoncek: Yes, that is true. C3 is working for me with 5.1.1 using build environment present in WLED.

boards/esp32-c3-12f.json
blazoncek: Flash size may be different on many boards and add to that that some may feature PSRAM as well.
It will be inevitable to add as many build environments as there are possible boards *or* provide instructions how users can create their own environment in PIO override.

softhack007: I think the 2M partition table may be usefull for others too, because currently we don't have a WLED partions.csv for mcu's with 2M flash only. @andyshinn is OTA still possible with 2M?

blazoncek: > @andyshinn is OTA still possible with 2M?

Not for C3, S2, S3 or regular ESP32

Plaenkler:
What is the status of this PR? Requested changes are pending since 2022?

blazoncek:
A newer more thorough and generic PR is open for 8266, C3, S2 & S3. #3144

Issue #2931, opened 2022-12-07
Usermod settings SCL and SDA not showing platformio values


ewoudwijma:
### What happened?

Usermod settings do not show values for build flags HW_PIN_SCL and HW_PIN_SDA on new install. Show -1 instead:

<img width="483" alt="image" src="https://user-images.githubusercontent.com/1737159/206289478-518f3e58-0635-453d-8c4c-a55bafefb35e.png">

### To Reproduce Bug

erase flash: pio run --target erase

Set in platformio environment xyz
  -D HW_PIN_SCL=22 -D HW_PIN_SDA=21

build xyz

go to settings / usermod 

### Expected Behavior

<img width="358" alt="image" src="https://user-images.githubusercontent.com/1737159/206290689-531faa03-7e18-4c08-84d4-4ed57e08b922.png">

### Install Method

Self-Compiled

### What version of WLED?

0.14.0 latest

### Which microcontroller/board are you seeing the problem on?

ESP32

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ewoudwijma:
Same applies for 

<img width="352" alt="image" src="https://user-images.githubusercontent.com/1737159/206291719-3b16c07b-b1fb-4131-92b4-b75f188a68ec.png">

(HW_PIN_DATASPI, HW_PIN_MISOSPI, HW_PIN_CLOCKSPI)

blazoncek:
They do, as a placeholder (remove -1 or any other input by deleting digits). This is intentional as to not allocate GPIO during boot.
HW_PIN_... macros are just informational which GPIO is reserved (if used) for hardware accelerated IO.

Not a bug but intentional behaviour.

ewoudwijma:
I saw the placeholder functionality.

I would say setting hw_pin_sda/scl (or not setting) determines if gpio should be allocated or not.
Might be confusing for end users as build flags normally are shown as default in the ui.

We are currently reviewing the sda/scl functionality. After review will let you know our advice.

blazoncek:
> I would say setting hw_pin_sda/scl (or not setting) determines if gpio should be allocated or not. Might be confusing for end users as build flags normally are shown as default in the ui.

Well, it is not intended in the way you think. Confusing or not these setting are meant for advanced users who know how are they compiling and for what.
If you want to pre-allocate I2C and/or SPI GPIO pins, do so by introducing new compile time options. These may have nothing in common with HW_PIN_...
 
[...]

PR #2955, opened 2022-12-15
Add build flags for global i2c & SPI pins

ezcGman:
Hey there,

it's me again! We also don't have enough build flags, do we? :) So please let me add some more!

This PR is to make the new global i2c and SPI pins configurable via build flags. Why would I want that? Because for the QuinLED boards, we compile a build that already has everything set (like LED outputs, relay pin, IR pin, etc.) using build flags. And now in combination with my new SHT usermod (https://github.com/Aircoookie/WLED/pull/2942), we can also provide a build with that mod included, but we currently can't set the i2c pins; users would need to do that themselves still.

With this PR, we could pre-configure everything on the boards so boards are even more Plug-n-Play:
```
[env:custom_i2c_pin_test]
board = esp32dev
platform = ${esp32.platform}
build_flags = ${common.build_flags_esp32}
  -D I2CSDAPIN=32
  -D I2CSCLPIN=13
  -D SPIMOSIPIN=0
  -D SPIMISOPIN=1
  -D SPISCLKPIN=2
```

We only need the i2c pins, but I thought it would be stupid to not do the same for the SPI pins :)

Thanks for considering this PR and greetings,

Andy!

### DIFF
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -392,12 +392,22 @@
 
 #define INTERFACE_UPDATE_COOLDOWN 2000 //time in ms to wait between websockets, alexa, and MQTT updates
 
+// HW_PIN_SCL & HW_PIN_SDA are used for information in usermods settings page and usermods themselves
+// which GPIO pins are actually used in a hardwarea layout (controller board)
+#if defined(I2CSCLPIN) && !defined(HW_PIN_SCL)
+  #define HW_PIN_SCL I2CSCLPIN
+#endif
+#if defined(I2CSDAPIN) && !defined(HW_PIN_SDA)
+  #define HW_PIN_SDA I2CSDAPIN
+#endif
+// you cannot change HW I2C pins on 8266
 #if defined(ESP8266) && defined(HW_PIN_SCL)
   #undef HW_PIN_SCL
 #endif
 #if defined(ESP8266) && defined(HW_PIN_SDA)
   #undef HW_PIN_SDA
 #endif
+// defaults for 1st I2C on ESP32 (Wire global)
 #ifndef HW_PIN_SCL
   #define HW_PIN_SCL SCL
 #endif
@@ -405,6 +415,18 @@
   #define HW_PIN_SDA SDA
 #endif
 
+// HW_PIN_SCLKSPI & HW_PIN_MOSISPI & HW_PIN_MISOSPI are used for information in usermods settings page and usermods themselves
+// which GPIO pins are actually used in a hardwarea layout (controller board)
+#if defined(SPISCLKPIN) && !defined(HW_PIN_CLOCKSPI)
+  #define HW_PIN_CLOCKSPI SPISCLKPIN
+#endif
+#if defined(SPIMOSIPIN) && !defined(HW_PIN_MOSISPI)
+  #define HW_PIN_MOSISPI SPIMOSIPIN
+#endif
+#if defined(SPIMISOPIN) && !defined(HW_PIN_MISOSPI)
+  #define HW_PIN_MISOSPI SPIMISOPIN
+#endif
+// you cannot change HW SPI pins on 8266
 #if defined(ESP8266) && defined(HW_PIN_CLOCKSPI)
   #undef HW_PIN_CLOCKSPI
 #endif
@@ -414,10 +436,7 @@
 #if defined(ESP8266) && defined(HW_PIN_MISOSPI)
   #undef HW_PIN_MISOSPI
 #endif
-#if defined(ESP8266) && defined(HW_PIN_CSSPI)
-  #undef HW_PIN_CSSPI
-#endif
-// defaults for VSPI
+// defaults for VSPI on ESP32 (SPI global, SPI.cpp) as HSPI is used by WLED (bus_wrapper.h)
 #ifndef HW_PIN_CLOCKSPI
   #define HW_PIN_CLOCKSPI SCK
 #endif
@@ -426,9 +445,6 @@
 #endif
 #ifndef HW_PIN_MISOSPI
   #define HW_PIN_MISOSPI MISO
-#endif
-#ifndef HW_PIN_CSSPI
-  #define HW_PIN_CSSPI SS
 #endif
 
 #endif
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -676,32 +676,32 @@ WLED_GLOBAL uint16_t ledMaps _INIT(0); // bitfield representation of available l
 // Usermod manager
 WLED_GLOBAL UsermodManager usermods _INIT(UsermodManager());
 
-// global I2C SDA pin [HW_PIN_SDA] (used for usermods)
+// global I2C SDA pin (used for usermods)
 #ifndef I2CSDAPIN
 WLED_GLOBAL int8_t i2c_sda  _INIT(-1);
 #else
 WLED_GLOBAL int8_t i2c_sda  _INIT(I2CSDAPIN);
 #endif
-// global I2C SCL pin [HW_PIN_SCL] (used for usermods)
+// global I2C SCL pin (used for usermods)
 #ifndef I2CSCLPIN
 WLED_GLOBAL int8_t i2c_scl  _INIT(-1);
 #else
 WLED_GLOBAL int8_t i2c_scl  _INIT(I2CSCLPIN);
 #endif
 
-// global SPI DATA/MOSI pin [HW_PIN_DATASPI] (used for usermods)
+// global SPI DATA/MOSI pin (used for usermods)
 #ifndef SPIMOSIPIN
 WLED_GLOBAL int8_t spi_mosi  _INIT(-1);
 #else
 WLED_GLOBAL int8_t spi_mosi  _INIT(SPIMOSIPIN);
 #endif
-// global SPI DATA/MISO pin [HW_PIN_MISOSPI] (used for usermods)
+// global SPI DATA/MISO pin (used for usermods)
 #ifndef SPIMISOPIN
 WLED_GLOBAL int8_t spi_miso  _INIT(-1);
 #else
 WLED_GLOBAL int8_t spi_miso  _INIT(SPIMISOPIN);
 #endif
-// global SPI CLOCK/SCLK pin [HW_PIN_CLOCKSPI] (used for usermods)
+// global SPI CLOCK/SCLK pin (used for usermods)
 #ifndef SPISCLKPIN
 WLED_GLOBAL int8_t spi_sclk  _INIT(-1);
 #else
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -130,7 +130,7 @@
   #include "src/dependencies/dmx/ESPDMX.h"
  #else //ESP32
   #include "src/dependencies/dmx/SparkFunDMX.h"
- #endif  
+ #endif
 #endif
 
 #include "src/dependencies/e131/ESPAsyncE131.h"
@@ -393,8 +393,8 @@ WLED_GLOBAL bool arlsForceMaxBri _INIT(false);                    // enable to f
  #ifdef ESP8266
   WLED_GLOBAL DMXESPSerial dmx;
  #else //ESP32
-  WLED_GLOBAL SparkFunDMX dmx; 
- #endif 
+  WLED_GLOBAL SparkFunDMX dmx;
+ #endif
 WLED_GLOBAL uint16_t e131ProxyUniverse _INIT(0);                  // output this E1.31 (sACN) / ArtNet universe via MAX485 (0 = disabled)
 #endif
 WLED_GLOBAL uint16_t e131Universe _INIT(1);                       // settings for E1.31 (sACN) protocol (only DMX_MODE_MULTIPLE_* can span over consequtive universes)
@@ -578,7 +578,7 @@ WLED_GLOBAL int16_t currentPlaylist _INIT(-1);
 //still used for "PL=~" HTTP API command
 WLED_GLOBAL byte presetCycCurr _INIT(0);
 WLED_GLOBAL byte presetCycMin _INIT(1);
-WLED_GLOBAL byte presetCycMax _INIT(5); 
+WLED_GLOBAL byte presetCycMax _INIT(5);
 
 // realtime
 WLED_GLOBAL byte realtimeMode _INIT(REALTIME_MODE_INACTIVE);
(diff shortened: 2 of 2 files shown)


softhack007:
Could it be that this already exists?

https://github.com/Aircoookie/WLED/blob/8619e8fc0bfbf97b0834384aa683e66bb1d7337a/wled00/const.h#L401-L406

https://github.com/Aircoookie/WLED/blob/8619e8fc0bfbf97b0834384aa683e66bb1d7337a/wled00/const.h#L420-L432

https://github.com/Aircoookie/WLED/blob/8619e8fc0bfbf97b0834384aa683e66bb1d7337a/wled00/wled.h#L679-L683

ezcGman:
No, these are just helpers that always have the HW default pin per board. Each board can have different HW pins and also different const defined in the Arduino lib. Hence WLED uses these HW_* to access the default HW pin for SPI/i2c, regardless of the board it's compiled for.

Those are not even considered for these new global / usermod i2c/SPI pins, not sure why though... Maybe a better change would be instead of new pins (what I did) initialize these usermod i2c/SPI pins with these HW_* const, instead of -1 🤔 

Let's see what maintainers say!

ezcGman:
Actually no... I don't want to set the default HW pin; I want to set the usermod i2c pin specifically... Still, maybe a change would be;

* Add new build flags / consts as I did for specifically setting the usermod i2c/SPI pins
* Still, when these new consts are absent, initialize these with the HW_* consts, instead of -1

blazoncek:
@softhack007 this actually aligns with what @ewoudwijma wanted to do with #2931.
I do agree that this approach is acceptable but **will not be added** to built in platformio.ini (unless @Aircoookie approves and creates such environment). This is due to the fact that people share binaries and if they lay hands on such build which will prevent them to use I2C or SPI pins we will get many support requests or unnecessary issues opened.

softhack007:
> This is due to the fact that people share binaries and if they lay hands on such build which will prevent them to use I2C or SPI pins we will get many support requests or unnecessary issues opened.

I fully agree. Still i'm not sure if a second `-D I2CSCLPIN=42` is necessary, when the same information can passed to WLED already by `-D HW_PIN_SCL=42`. 

There is only two I2C hardware units available, and only one can be used when doing a `Wire.Begin(..,..);`.  So my suggestion would be to remove new preprocessor defines like `I2CSCLPIN` and utilize the existing ones.

softhack007:
> Those are not even considered for these new global / usermod i2c/SPI pins, not sure why though... Maybe a better change would be instead of new pins (what I did) initialize these usermod i2c/SPI pins with these HW_* const, instead of -1 🤔

I agree - this is problematic when you want to make a build that runs "out of the box" and utilizes all features of a specific board instantly. Another proposal yould be to have _one_ new define like `-D I2C_ENABLED_AT_STARTUP`. If this is set, initialize i2c_sda )and other) with whatever is in HW_PIN_SCL.

softhack007:
> Actually no... I don't want to set the default HW pin; I want to set the usermod i2c pin specifically.

This is very much the same thing. On ESP32, you can choose any GPIO as I2C pins, at the same time you only have _one_ Wire and _one_ Wire1 globally. So when _one_ usermod chooses 40/41 for SCL/SDA, then _all_ other usermods have to follow. Only exception is 4LD, which also offers "software i2C".

blazoncek:
When you put it that way it makes sense. 👍

blazoncek:
You are correct, this area is still very much WIP.

1. yes, these are used to *override* default values of GPIOs assigned to HW accelerated interface (as some implenetations of controller boards use differnet GPIO layout than what is usually recommended/default by Espressif). they do not allocate (reserve, use) those GPIO in any way. their intent is informational.
2. see 1.
3. not necessarily, since I2C or SPI can be shared by many devices WLED may use them in the future
4. see 1., setting globals to -1 was intentional to prevent unnecessary GPIO allocation

You are correct with your thinking.

ezcGman:
Thanks for the answers, Blaz!

Sooo, elaborating on your reply earlier:
> I do agree that this approach is acceptable

Did you mean this / my current approach here? Or my suggestion to also initialize them with the HW_* macros? Or were you referencing the other issue you linked? 🙈 I got confused, sorry :D

> but will not be added to built in platformio.ini (unless @Aircoookie approves and creates such environment).

With "added to built in platformio.ini" you mean you won't provide a build config in platform.io using them? That's totally fine; We provide our own builds / pre-flash controllers with these. It's just that we can build such a build (haha) and have factory pre-flash it to the controllers.

> This is due to the fact that people share binaries and if they lay hands on such build which will prevent them to use I2C or SPI pins we will get many support requests or unnecessary issues opened.

Totally makes sense, yeah.

Thanks again!

blazoncek:
> Did you mean this / my current approach here?

Yes, introducing new set of #defines for setting `i2c_sda`, `i2c_scl`, ... is ok IMO. They may work in tandem with `HW_PIN_...` even though they may introduce some ambiguity. We will need to add description of what `HW_PIN_...` constants are used for.

> With "added to built in platformio.ini" you mean you won't provide a build config in platform.io using them?

Yes, exactly that. I would encourage you (or @intermittech) and @ewoudwijma (or @srg74) to provide `platformio_override.ini` with approriate build environments for users that want to compile source themselves for those particular boards you support/sell. Or just binaries for those that do not compile.

ewoudwijma:
@softhack007 and I worked together on the same topic, analysis can be found in https://github.com/Aircoookie/WLED/issues/2931

What we did at least partly is overlapping with your ideas

Very beta coding can be found here:
https://github.com/MoonModules/WLED/tree/I2CSPI-refactor

@ezcGman , if you want you can give it a try with this bin:
[WLEDMM_0.14.0.10_esp32_4MB_max.bin.zip](https://github.com/Aircoookie/WLED/files/10245113/WLEDMM_0.14.0.10_esp32_4MB_max.bin.zip)

To see what we did best is to look at the commit diffs 
https://github.com/MoonModules/WLED/compare/mdev...MoonModules:WLED:I2CSPI-refactor

ezcGman:
@ewoudwijma I think what you guys do there is on a whole other level / does more than I wanted to achieve with this tiny PR. So whatever you do there might (if my PR gets merged) overwrite what I did here ;)

I was actually just looking for a way to set these pins via build flags, similar to how WLED does it for all the other pins (Relay, IR, buttons, LED outs...) :)

So I'm still hoping to get this change merged, if it's acceptable / doesn't cause issues and I bet what you are doing there might overwrite what I did here, as you spent way more thought into this than I did ;)

softhack007:
One point is still not clear for me. 
I don't see the point of having to two configurable options for the same thing, like I2CSCLPIN and HW_PIN_SCL.

The problem is, it will not work properly when I2CSCLPIN and HW_PIN_SCL have different values. I have explained before, so I'm not gonna repeat myself about that.

All you want can be achieved with setting HW_PIN_SCL, and then adding _one_ new flag like INIT_I2C_AT_STARTUP that can be used to request initialization of global i2c_sda with HW_PIN_SCL, instead of initializing with -1 (and other i2c and spi pins alike).

The problem is, it will not work properly when I2CSCLPIN and HW_PIN_SCL have different values.

blazoncek:
> One point is still not clear for me. I don't see the point of having to two configurable options for the same thing, like I2CSCLPIN and HW_PIN_SCL.

These *are not the same*. One assigns global `i2c_sda` & `i2c_scl` the other just defines which GPIO should be used as HW accelerated pin. This due to possible pre-allocation in current implementation.

> The problem is, it will not work properly when I2CSCLPIN and HW_PIN_SCL have different values. I have explained before, so I'm not gonna repeat myself about that.

This very much depends on the implementation of software I2C driver. I am not saying we *want* to use it as such, but someone might.

> All you want can be achieved with setting HW_PIN_SCL, and then adding _one_ new flag like INIT_I2C_AT_STARTUP that can be used to request initialization of global i2c_sda with HW_PIN_SCL, instead of initializing with -1 (and other i2c and spi pins alike).

I do not like this idea as it is counter intuitive compared to initialisation of other GPIOs like relay, button, etc.

I would rather dump HW_PIN_SDA & HW_PIN_SCL as these are informational (and always were as such). More like helpers for usermod to decide if it uses SW or HW driver.

ezcGman:
> I don't see the point of having to two configurable options for the same thing, like I2CSCLPIN and HW_PIN_SCL.

Because they're not the same thing, see my detail questions and Blaz' answers above:

> they do not allocate (reserve, use) those GPIO in any way. their intent is informational.

> The problem is, it will not work properly when I2CSCLPIN and HW_PIN_SCL have different values. I have explained before, so I'm not gonna repeat myself about that.
Thanks for not providing clarity on what you mean, if it seems to be unclear to people. I don't see an issue why this wouldn't work. `HW_PIN_SCL` is not used anywhere at the moment and even if it would: At least on ESP32 you can have two i2c busses working simultaneously. One mod (or WLED core) could use the bus that is behind `HW_PIN_SCL`, a mod can use the ones behind the `i2c_scl` WLED GLOBAL.

ewoudwijma:
> I do not like this idea as it is counter intuitive compared to initialisation of other GPIOs like relay, button, etc.
> 
> I would rather dump HW_PIN_SDA & HW_PIN_SCL as these are informational (and always were as such). More like helpers for usermod to decide if it uses SW or HW driver.

This is what is confusing for me, build flags are normally initializing values, also for relay, button etc. Why shouldn't hw_pin values work similar. I don't see a point in making them informational only. 
Besides that, we are coders so we want values to DO something. Informational only is more for politicians or lawyers 😜

blazoncek:
I suggest that we go back to the start and re-define everything and build from there.
Let's determine what we have:

1. usermods may want to share I2C bus
2. there are 2 global instances of HW I2C bus (Wire and Wire1) both share-able.
3. WLED itself does not use I2C and would want as many GPIO pins available for LEDs
4. some controller board vendors may want to provide binaries pre-compiled for their boards (with as many pre-defined pins as possible)
5. those controller boards may have different GPIO layout than the default recommended by Espressif
6. users may accidentally delete pre-compiled configuration (stored in cfg.json) and would not want to erase flash to recover default settings (they want visual cue)
7. we want usermods to be "compatible" with each other regarding sharing of I2C
8. users may want to utilise 2 I2C buses for different devices
9. there may be usermods that can utilise I2C or SPI bus depending on the selected device type (e.g. 4 Line Display)
10. there may be composite devices that use I2C or SPI pins alongside other pins (e.g. digital microphone)

From this we may get:

1. WLED introduces global I2C pin variables (for one I2C bus only) defaulting to -1 (accommodates 1., 2., 3.)
2. we define pin override macros (#defines) to allow setting of non-default I2C pin values (accommodates 4., 5.)
3. we define constants of controller board HW wired I2C pins for UI information (accommodates 5., 6.)
4. we need to re-write/update **all** usermods to use new global I2C pins instead of the ones provided in usermod's settings (accommodates 7., 9.)
5. we leave second I2C bus (Wire1) for custom implementation by users (acommodates 8., 10.)

All points, except 2. & (partially) 4., have been implemented in current version and this PR adds support for 2. as I see it.

I propose to merge this PR and after that we only need to work on 4. (audio_source.h may need update to use Wire1 for the special microphone).

ewoudwijma:
> I propose to merge this PR and after that we only need to work on 4. (audio_source.h may need update to use Wire1 for the special microphone).

I am afraid this ignores some of the things I and @softhack007 are currently working on.

This introduces new build flags while we have already HW_PIN * flags (and FLD_PIN flags for FLD). I think we need to use the HW_PIN flags to implement what Is done in this PR (or rename them to I2CSDAPIN etc) but not have both.

This as been worked on (and is WIP) as described in https://github.com/Aircoookie/WLED/issues/2931.

So I think we wait on advice of @softhack007 on this matter

softhack007:
Hi @blazoncek, @ezcGman and @ewoudwijma,

First I'd like to apologize, as my words seemed to cause frustration and I had the feeling of not being listened to at all.

In addition to what was already said, let me summarize what is important from my point of view.
Current there is ongoing work to improve handling of I2C busses (and possibly SPI, too) which will be relevant for a future release of WLED, like 0.14.1. From this perspective, it is important to make good decision already now, because everything that will be put into users hands will become "the legacy" in 14.1 and we'll have to live with that once users made use of it. personally I am mainly working on the "low level" part, while I leave to @ewoudwijma and @blazoncek to find a good UI integration that is consistent with the overall concepts in WLED. 

There are some aspects where I fully agree with what was said before:
* There should be a way to pre-define I2C pins at compile time, similar to IRPin of BtnPin. So I absolutely agree with the Intention of this PR, to have such a possibility.

* The global variables `i2c_sda` and `i2c_scl` should reflect what is defined in usermod global preferences. Values "-1" should mean "don't use I2C".
* It is true that when building boards / devices / shields, you want to have the possibility to say "these are the I2C pins you should use".
* The whole discussion about "hardware I2C PINS" is relevant for ESP32 only, because on ESP8266 there is nothing you can change, pins are fixed, this is already considered in WLED 0.14.
* personally, I would also prefer to dump HW_PIN_SDA & HW_PIN_SCL, as these have no real meaning and can only help for coloring things in the UI (see below).

The ESP32 hardware is a bit tricky, because in reality there is no "Hardware (accelerated) I2C PIN" as you would have on other arduino environment. It works completely the other way. There are two I2C units on the chip, and you have "Wire" and "Wire1" respectively. Using Wire.begin without parameter is not a good idea, and WLED 0.14 already has a way to prevent this. And it works basically, when you only have one I2C device you want to use. For more than one device, we need architecture changes for 0.14.1.

When a usermod calls Wire.begin() without parameters, this means that - silently - the values from pins_arduino.h will be used as GPIO, no matter what is in HW_PIN_SDA etc - not really good from my POV. The other important point is, once that i2C pins were set during the very first Wire.begin(pin_sda, pin_scl), these pins become "the hardware pins". A reboot is needed to set other gpio.

This brings me to the problem i see currently, when allowing users to independently define I2CSCLPIN and HW_PIN_SCL:
* It is inconsistent to IRPin or Btnpin - we would not want IRPin + BtnPin, and at the same time have HW_IR_PIN HW_BUTTON_PIN. Because it creates confusion and it does not make much sense.

* As the first Wire.begin(...) determines what are the "hardware PINs", effectively when I2CSDAPIN is defined, then HW_PIN_SDA becomes meaningless - as a consequence of how the ESP32 works. 

* Currently I don't see that we will need these HW_I2C_PINS even if we allow software based I2C in all usermods. Most users will not treat that correctly, and we end up with a very inconsistent situation in pinManager - like 6 I2C_HW allocations, but the ones really used by the i2C units are still different from that.

I hope this helps to understand my concerns. To say it again - I am fully for providing a way for users to define "preferred I2C pins" of their special board/shield. I am just saying that having both I2CSCLPIN and HW_PIN_SCL that can be overriden by build_flags is not making much sense technically, it could lead to confusion, and might might our life harder once we make improvements for 0.14.1.

What do you think?

blazoncek:
Both of you are still missing the point of this PR. From my perspective it is clear and approved. Please re-read my post above.
There is no need to talk of intricacies of I2C initialisation for this PR, it does not deal with that.

Still if you want: This PR, combined with existing code in WLED (cfg.cpp, lines 294-322; set.cpp, lines 507-560) satisfies all needs for initialising I2C and SPI buses/interfaces. Once the variables `i2c_sda` and `i2c_scl` are set the interface is initialised and hardware set up. There is no need to do that anywhere else! It is consistent and global, applied at boot.

What *is* needed is to modify existing usermods to heed these settings and notify them if pins change from set.cpp (may need a reboot, I understand this). Nothing more!

If you and @ewoudwijma still cannot wrap your heads around HW_PIN_SDA or HW_PIN_SCL take another look at const.h, lines 395-433, where their definition is set. They are nothing more than **constants** with references to platform dependant or user specified values. Their purpose is to let any code identify if `i2c_sda` and `i2c_scl` is set to "user's hardware" or not. Nothing else. A visual cue for them is also available in usermods settings page if you delete characters in I2C & SPI input fields.

For clearer picture consider this:
```c++
if (i2c_sda == HW_PIN_SDA) {
  // we have SDA pin that corresponds to what is used in hardware/controller board
  ...
} else {
  // warn user that setup is not according to his/her hardware
  ...
}
```

@softhack007 you are comparing apples to oranges when you talk about IRPIN, etc. and HW_PIN_SDA/HW_PIN_SCL. IRPIN corresponds to I2CSCLPIN/I2CSDAPIN **from this PR**. HW_PIN_SDA/HW_PIN_SCL have to do with controller board design (like Dig-Octa or Wemos Shield) or hardware wiring so that any code may be aware of that. HW_PIN_... have nothing to do either with I2C interface or variable initialisation. They never did.

If you still think this is confusing I may remove all references to HW_PIN_... and replace visual cue in xml.cpp, lines 700-704, with I2CSDAPIN, etc. The problem with this approach is that you will not be able to have "help for the user" when he/she later adds I2C sensor but has a binary that does not define `i2c_...` variables using I2CSDAPIN (which will be stock with WLED.) So build environments like `[env:wemos_shield_esp32]` or (currently inexistent) `[env:dig_octa]` etc. will not be able to help users setting appropriate pins for their hardware.

softhack007:
Hi @blazoncek I you still think I'm talking gibberish and don't know what I'm saying, then maybe we have a bigger problem
and let's discuss privately on discord.

It seems my opinion is to "brainy" or something, so I'll shut up for now and will stop making your life miserable.
Go on as you like. I'l concentrate on MoonModules, as I don't want to get anybody into a situation where they fear of losing their faces.

If you want, contact me on discord. Finito.

softhack007:
Just in case anybody here was concerned about yesterday's disussion - all good. @blazoncek and myself agreed on a compromise that allows to merge this PR without possible problems later, and comments will be added to explain the purpose of new build flags. So approved from my side 👍

ezcGman:
Thanks guys! And I feel a little bad for kicking off this heated discussion here <3

softhack007:
> And I feel a little bad for kicking off this heated discussion here <3

Don't worry. From my perspective this discussion was due any way, so you just gave a spark, but another other "spark" could have caused the same. What's important is that we we able to find a good way forward.

MoonModules:
I looked at latest dev:

Is it correct HW_PIN_SCL should now only providing info ?

If so, I see in code it is used to provide values and allocate pins with it. Is this correct?

image

and should I2CSCLPIN be added to platform io entries where now HW_PIN_SCL is used?

Same question for HW_PIN_CLOCKSPI and a bunch of other HW_PIN values

blazoncek:
Yes, this is all correct.

ewoudwijma:
If all my statements are correct then it contradicts with the note you added. There is a lot written above.
I am confused still. Can you maybe summarize the relevant texts above to get a better understanding?

blazoncek:
Defining `I2CSDAPIN` will assign `i2c_sda` (and consequentially allocate GPIO using PinManager). It will also define `HW_PIN_SDA` if not defined explicitly.
Defining `HW_PIN_SDA` will only inform code what is hardwired SDA GPIO. If left undefined platform specific constants are used instead of user specified value. `i2c_sda` will remain unassigned (-1) and no allocation is done.

All code should use `i2c_sda` whenever dealing with GPIO pins, and use `HW_PIN_SDA` in comparisons to detect if user entered HW accelerated/assisted GPIO. No user code should ever use `I2CSDAPIN` anywhere (the only exception is wled.h).

Some usermods will require updates.

PR #2959, opened 2022-12-16
maximum segments reached and Select all bugfix

blazoncek:
Fixes "Maximum segments reached" text and corrects behaviour of Select all checkbox


Aircoookie:
Confirmed working as should (including unselect)!

Aircoookie (approved):

wled00/data/index.js
```diff
@@ -1694,15 +1689,15 @@ function makeSeg()
...
 
-function resetUtil()
+function resetUtil(off=false)
 {
-//	gId('segutil').innerHTML = '<button class="btn btn-s" onclick="makeSeg()"><i class="icons btn-icon">&#xe18a;</i>segment</button>';
-	gId('segutil').innerHTML = '<div class="seg btn btn-s" style="border-radius:24px;padding:0;">'
+	gId('segutil').innerHTML = `<div class="seg btn btn-s ${off?'off':''}" style="border-radius:24px;padding:0;">`
 	+ '<label class="check schkl"><input type="checkbox" id="selall" onchange="selSegAll(this)"><span class="checkmark"></span></label>'
-	+ '<div class="segname" onclick="makeSeg()"><i class="icons btn-icon">&#xe18a;</i>Add segment</div></div>';
+	+ `<div class="segname" ${off?'':'onclick="makeSeg()"'}><i class="icons btn-icon">&#xe18a;</i>Add segment</div></div>`;
```
softhack007, line 1696: Is is correct to use a "back-tick" here? Just curious, as I'm not so much experienced with JS.

softhack007 (approved):
Nothing from my side, just curious question about using "back-tick" instead of single quotes.

wled00/data/index.js
blazoncek: Back-tick (as I learned during my WLED tenure) is a special string marker. Anything within back-ticks is paresed in a similar way as PHP parses strtings. `${...}` is the parsing trick. Within it, regular JS can be embedded.

PR #2963, opened 2022-12-19
SHT Usermod: Fixed MQTT discovery using correct unit; Added getters and isEnabled() check

ezcGman:
Hey there,

I'm sorry to open another PR so close after the big one got merged, but sadly a bug sneaked in :( Fixing it with these two commits!

Thanks and greetings,

Andy!

### DIFF
--- a/usermods/sht/usermod_sht.h
+++ b/usermods/sht/usermod_sht.h
@@ -56,9 +56,13 @@ class ShtUsermod : public Usermod
     bool readFromConfig(JsonObject &root);
     void addToJsonInfo(JsonObject& root);
 
-    float getTemperatureC();
-    float getTemperatureF();
-    float getHumidity();
+    bool isEnabled() { return enabled; }
+
+    float getTemperature();
+    float getTemperatureC() { return shtCurrentTempC; }
+    float getTemperatureF() { return shtCurrentTempF; }
+    float getHumidity() { return shtCurrentHumidity; }
+    const char* getUnitString();
 
     uint16_t getId() { return USERMOD_ID_SHT; }
 };
@@ -162,9 +166,9 @@ void ShtUsermod::publishTemperatureAndHumidityViaMqtt() {
   char buf[128];
 
   snprintf_P(buf, 127, PSTR("%s/temperature"), mqttDeviceTopic);
-  mqtt->publish(buf, 0, false, String((unitOfTemp ? getTemperatureF() : getTemperatureC())).c_str());
+  mqtt->publish(buf, 0, false, String(getTemperature()).c_str());
   snprintf_P(buf, 127, PSTR("%s/humidity"), mqttDeviceTopic);
-  mqtt->publish(buf, 0, false, String(shtCurrentHumidity).c_str());
+  mqtt->publish(buf, 0, false, String(getHumidity()).c_str());
 }
 
 /**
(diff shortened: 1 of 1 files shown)


SK360:
Tested the changes and temperature reports to HA properly now when set to F. Thanks for fixing!

usermods/sht/usermod_sht.h
```diff
@@ -191,7 +191,7 @@ void ShtUsermod::publishHomeAssistantAutodiscovery() {
   json[F("stat_cla")] = F("measurement");
   snprintf_P(buf, 127, PSTR("%s-temperature"), escapedMac.c_str());
   json[F("uniq_id")] = buf;
-  json[F("unit_of_meas")] = F("°C");
+  json[F("unit_of_meas")] = F((unitOfTemp ? "°F" : "°C"));
```
blazoncek, line 194: I am not sure this is the correct way of using F().
To be on the safe side use `unitOfTemp ? F("°F") : F("°C")`

And there are some other similar statements `unitOfTemp ? ... : ...` that I find suboptimal regarding code size.

blazoncek (changes requested):
Please, while you're at it, do some code optimization.
There are some "not so good" code constructs throughout.

usermods/sht/usermod_sht.h
ezcGman: What would be different? In both cases it passes either "°C" or "°F". the ternary operator is solved before calling F()

blazoncek: Please check the definition of F() wrapper. It is not a regular function.

softhack007: I agre with @blazoncek ... F() is used to make sure that strings are read directly from flash memory. It's actually a compile-time thing, so probably will not have the desired effect if the parameter of F() is not a string constant but a condition. Also 8266 has a different Implementation of F() it might work on esp32 but maybe not work on ESP8266.

Better to change as Blaz suggested.

ezcGman: Yeah, think it makes sense. I was under the assumption that the ternary operator is resolved before being passed on. But I missed that F() is actually a preprocessor macro, hence the ternary operator is most likely not resolved at the time, as this would be done at runtime, if I remember / understand correctly. Thanks for pointing it out, addressed it!

ezcGman:
Apologies for adding a few more changes to this PR. Usually it's bad practice to add different changes to a PR, but hence this is all around "my mod", I thought it's ok if I spare you from opening more PRs and just throw them in here.

I basically added a getter for the current temperature based on what's configured for the mod (F or C), a getter for humidity and a check if the mod is enabled.

I'm currently re-doing an older mod I did which also had SHT logic, which I'm now dropping in favor of this mode and avoiding duplicate code. Hence I added those getters and the enabled checker.

Hope it's OK that I added those changes here. Thanks!

ezcGman:
Just realized I missed using flashstring helper in `::getUnitString`. fixing

blazoncek:
If you are using short strings and they appear several times in your source then it is ok to use classic strings without helper macros or PROGMEM keyword.
This being due to the fact that classic string is stored in precious RAM eating away heap and there is a penalty using string from flash. All strings stored in RAM are optimised by compiler so that only one instance is stored where flash strings will appear as many times as they are entered in the code.

ezcGman:
Sooooo, summarizing:
* Use `PROGMEM` / `F()` for strings where possible / as much as possible to save RAM
* But if you do and you have repeating strings, use `PROGMEM` and store them as class member variables  as I do for the config strings?

blazoncek:
Short string (up to 10-12 bytes), used many times (more than 2 across all source files) -> use regular string.
Otherwise use PROGMEM / F() / PSTR()

ezcGman:
Ok, think I'm getting it slowly :)

OK, but it generally doesn't hurt to put more into PROGMEM? Or does it? Because you said "there is a penalty using string from flash.": I guess you mean a performance penalty reading them? So actually don't put as much as you can? ;)

ezcGman:
I think then there's nothing left to do for me. I do see it saying that there is one pending change requested, but that conversation is marked as resolved. So not sure if there is something left to do about your comment to that review:
> There are some "not so good" code constructs throughout.

I could remove `F()` around strings that are "up to 10-12 bytes", if putting them into PROGMEM is actually an issue? Totally fine doing this, I'm just not 100% sure if that's requested, that's all :)

blazoncek:
Will have another look on weekend and try to optimise if there is any need.

ezcGman:
Sure, no rush! Would love to know where/how to optimize to do it right :) Excited! Thx!

blazoncek:
@ezcGman please test it after my slight modification. If it still works it is good to merge. 😄

blazoncek (approved):

ezcGman:
Thanks blaz! Confirmed everything still working. Also compared the two Fahrenheit values (what the lib returns (which calculates it from the raw read sensor value) plus now calculating it from the Celsius values) and they're exactly the same:
```
[SHT-Sensor] Fahrenheit calculated: 81.667198
[SHT-Sensor] Fahrenheit read from lib: 81.667198
```

So good to merge, thanks!

PR #2964, opened 2022-12-21
Add HTTP authentication

drhideg:
I know that without SSL it's not the real thing, but at least it's some kind of protection for the device.
Digest auth is work on both platform.
Basic auth on esp8266 is not work at the moment because the way `base64_encode_expected_len` calculated in esp8266 core.
Please consider apply suggested fix in the used fork of ESPAsyncWebServer based on this [issue](https://github.com/me-no-dev/ESPAsyncWebServer/issues/575)

-TODO: modify the official [WLED-App](https://github.com/Aircoookie/WLED-App)


blazoncek:
Related to https://github.com/Aircoookie/WLED/issues/308 which was marked "wontfix".
Please follow Wiki on how to submit PR.

drhideg:
Now I read https://github.com/Aircoookie/WLED/issues/308 and https://github.com/Aircoookie/WLED/issues/520. There were valid use cases and no major disadvantage except type in the password, but that can be handled by the browser.

Sorry for the incorrect procedure with the PR

softhack007:
Maybe a stupid question - what is the use case for authentication, considering that we already have password protection for certain settings?

I would never put WLED into a public network, as it will not stand any hacker attack for longer than one minute. Html is just one api - we also have json, mqtt, and other interfaces that are insecure so could used to either "knock out" wled, or to mis configure it.

softhack007:
Oh, and please don't do this "delete trailing spaces" thing in your PR. It makes it very hard to review your changes, because very often it's just a deleted blank at the end of the line but the code actually remains the same.

drhideg:
Sorry for whitspeaces, now i turned off removing them is VSCode.
Mine is just the total control of the device, others in the household can use the button attached on the device.
I am think it like a different user levels.
Dont need to put WLED in public, sometimes network is out of your control (dormitory, rented room).

drhideg:
If you think it as a parental lock to not every user of the network can control a device it is a basic and accepted solution. Not the same when device is open for all or have to make an effort to control. It is not forced to use, only an option.

PR #2965, opened 2022-12-21
Add usermod specific help

ewoudwijma:
Shows a ? button on top of every usermod where this is implemented. 
After pressing the button it will open the relevant page in kno.wled.ge.
To do: create these pages in kno.wled.ge

As discussed with @aircoookie


ewoudwijma:
Yeah, good point, on the other hand you could save on the short help texts which are now stored on esp.
What problems you will encounter if you increase size from 3K to 4K?

ewoudwijma:
That’s a good idea 👍 !

why oappend uses stack anyway?
Isn’t malloc a better idea?
Or is it 8266 related?

ewoudwijma:
Could we minimise stack buffer by shortening the names of the functions and also shortening the names of the fields:

addDropdown -> aD
addOption->aO
addInfo->aI

addInfo('AudioReactive:dynamics:limiter'... -> ai('ar:dyn:limiter'... etc

that would save substantially on buffer size

blazoncek:
> why oappend uses stack anyway? Isn’t malloc a better idea?

Legacy. Yes, but adds hassle with free.

> Could we minimise stack buffer by shortening the names of the functions and also shortening the names

Yes, but not content of the string. Except if you modify usermod strings.

ewoudwijma:
Okay found the solution building on [Usermod level up](https://mm.kno.wled.ge/moonmodules/merge-history/#usermod-level-up) :

Instead of building the pages for all usermods in one go only build it for one usermod at the time. By this we can use oappend to the max to provide as much semantics and help as needed (eg also [Pin dropdowns](https://mm.kno.wled.ge/moonmodules/merge-history/#drop-downs-for-pin-variables) .

will submit this to WLED MM next week.

tell me when you are ready to PR this to WLED AC 👍

blazoncek:
If you split usermod settings pages then 3k buffer is of lesser issue.
Please make a PR for split usermod pages and include changes proposed here.
I will ask you to *test* new settings layout with all current v2 usermods.

ewoudwijma:
👍 

What do you want to do with the remaining content in menu option 'Usermod'. Know oney the globals will stay there, nothing more:

image

In MoonModules we left it there for upstream compatibility but I think it is better to move these values to Led preferences as there are other pins. Maybe even better would be to replace the Usermod menu option by a new menu option called Pins or something and move all the (non usermod) pin settings there (buttons, relays, ir, ...)

What do you think we should do?

Aircoookie:
oappend uses stack since the stack is always allocated anyways, allocating 4k+ heap during a network request might exacerbate low memory problems.
Not a fan of split usermod pages (current settings system is to be replaced by a single page application soon anyways), I would propose adding another helper function `addHB()` so that `oappend` does not need to contain button HTML or the `https://kno.wled.ge/usermods/` link root. That way, e.g. the super short `oappend(SET_F("addHB('PIRsensorSwitch');"));` would suffice:

```js
function addHB(um)
{
  addInfo(um + ':help',0,
    '<button onclick=\"location.href=&quot;https://kno.wled.ge/usermods/' + um + '&quot;\" type=\"button\">?</button>')
}
```

(untested)

ewoudwijma:
> would propose adding another helper function `addHB()` so that `oappend` does not need to contain button HTML or the `https://kno.wled.ge/usermods/` link root. That way, e.g. the super short `oappend(SET_F("addHB('PIRsensorSwitch');"));` would suffice:

Okay, will change the PR to do that. 

In the meantime I added the readme contents of some of the usermods into https://mm.kno.wled.ge/usermods/4LineDisplay/:

Screenshot 2023-01-05 at 17 15 47

@Aircoookie , is this the way forward you want? I think we should split the content. Developer stuff should remain in the readme files and user content in kno.wled.ge. What do you think?

If you are okay with this I can PR the mm.kno.wled.ge usermod stuff to kno.wled.ge...

Aircoookie:
Thank you!

Would you give an example of what you would consider "developer stuff"? I think the most straightforward approach would be if the kno.wled.ge page replaced the readme altogether (to reduce the risk of information duplication and possible misalignment). The drawback of this is that usermod updates would often need to be combined with a KB page update.

MoonModules:
Developer stuff: I meant more install instructions and change log. I agree we can move everything except change log.

The drawback applies to all code changes, usermods or non usermods so I think that is more a discipline/priority thing.

Aircoookie:
Fully agreed!

I don't think the readme is displayed/referenced anywhere but when browsing code in GitHub itself.

MoonModules:
So what are the next steps? 

A proposal:

Fase 1 - knowledge repo
1. Move all the readme.md contents (except change log) to kno.wled.ge/usermods  (# = ##, ## = ### etc)
2. Add the usermods in the knowledge menu (mkdocs.yml), as there are a lot, most used on top, less used lower

Fase 2 - wled repo
1. Cleanup usermods: do we want to remove some? My proposal: make 4ld_ALT and rotary_ALT the default ones and remove the originals as the _ALT versions are way better
2. add addHB function and calls (in appendConfigData) in every usermod (see https://github.com/MoonModules/WLED/commit/2187aa5728cc05787dcce4fd90abda45450c58d7)
3. Remove the readme.md files in the repo and replace by change_log.txt keeping the change logs, add a link in this file to the knowledge page of that usermod

Shoot!!

blazoncek:
More important question is: Are we going to provide binaries with usermods enabled? If so, which ones and in which configuration/combination?

On all of my devices I have Multi relay and PIR sensor as those may be most useful. Next in line would be Auto Save (for beginners) and Temperature as most controller boards now feature Dallas sensor. For more fancier builds ALT 4 Line Display and Rotary encoder usermods are great. And quite a few users implemented (or requested) Animated Staircase usermod.
Then there are more specialty usermods: PWM Fan, Battery, BH & BM & DHT sensors, etc.

I am not mentioning Audioreactive which should have its own build on every release IMO.

blazoncek:
@Aircoookie @ewoudwijma @MoonModules  @softhack007 what is the status of this PR?
I do not see any updates as discussed above. Do we abandon this PR in favour of a new one or bring it up to speed with most recent code?

To summarize what needs to be done:
- add `addHB()` function to `settings_um.htm`
- add `addInfo("addHB('usermod');")` to every usermod
- create (at least simple) WLED-docs help page for each usermod (may only need install/compile instructions and description of fields from readme.md)
- (optionally) clean-up usermods

ewoudwijma:
If you want us to do, happy to do so.
Will just pull the usermods info we already made in mm.kno.WLED.ge to up

netmindz:
Do you still have time to look into this @ewoudwijma ?

PR #2966, opened 2022-12-21
Hex custom palettes and smooth random palette change

blazoncek:
Ability to enter custom palettes in hex strings.
Smooth transitions for random palette changes.


blazoncek:
Sample hex palette:
`{"palette":[0,"FF0000",62,"FF0000",63,"FFA000",126,"FFA000",127,"00FF00",191,"00FF00",192,"0000FF",255,"0000FF"]}`

wled00/FX_fcn.cpp
```diff
@@ -230,16 +231,28 @@ CRGBPalette16 &Segment::loadPalette(CRGBPalette16 &targetPalette, uint8_t pal) {
   switch (pal) {
     case 0: //default palette. Exceptions for specific effects above
       targetPalette = PartyColors_p; break;
-    case 1: //periodically replace palette with a random one. Doesn't work with multiple FastLED segments
-      if (millis() - _lastPaletteChange > 5000 /*+ ((uint32_t)(255-intensity))*100*/) {
+    case 1: {//periodically replace palette with a random one. Transition palette change in 500ms
+      uint32_t timeSinceLastChange = millis() - _lastPaletteChange;
```
softhack007, line 235: millis() returns unsigned log.does this fit into int32? Maybe better use unsigned long, to avoid overflows when computing timesincelastchange. Also cross-check if t_lastpalettechange is ulong.

wled00/FX_fcn.cpp
```diff
@@ -230,16 +231,28 @@ CRGBPalette16 &Segment::loadPalette(CRGBPalette16 &targetPalette, uint8_t pal) {
...
                         CHSV(random8(), random8(160, 255), random8(128, 255)));
         _lastPaletteChange = millis();
+        timeSinceLastChange = 0;
       }
-      targetPalette = randomPalette; break;
+      if (timeSinceLastChange <= 500) {
+        targetPalette = prevRandomPalette;
+        // there needs to be 255 palette blends (48) for full blend but that is too resource intensive
+        // so 128 is a compromise
+        size_t noOfBlends = ((128U * timeSinceLastChange) / 500U);
```
softhack007, line 250: This means a full palette blend will happen in 500ms =1/2 sec, right?
Then 128 steps seems sufficient, as it means you can be accurate up to 4ms which is a lot more than what humans will ever notice :-)

wled00/FX_fcn.cpp
```diff
@@ -1511,15 +1521,28 @@ void WS2812FX::loadCustomPalettes()
...
+          if (pal[0].is<int>() && pal[1].is<const char *>()) {
+            // we have an array of index & hex strings
+            size_t palSize = MIN(pal.size(), 36);
+            palSize -= palSize % 2; // make sure size is multiple of 2
+            for (size_t i=0, j=0; i<palSize && pal[i].as<int>()<256; i+=2, j+=4) {
+              uint8_t rgbw[] = {0,0,0,0};
+              tcp[ j ] = (uint8_t) pal[ i ].as<int>(); // index
+              colorFromHexString(rgbw, pal[i+1].as<const char *>()); // will catch non-string entires
+              for (size_t c=0; c<3; c++) tcp[j+1+c] = rgbw[c]; // only use RGB component
+              DEBUG_PRINTF("%d(%d) : %d %d %d\n", i, int(tcp[j]), int(tcp[j+1]), int(tcp[j+2]), int(tcp[j+3]));
```
softhack007, line 1534: Maybe comment out these debug_print once it works ... because people will not understand what that means when running a debug build.

softhack007 (commented):
Hi, looks good. I am not en expert on JS, so concentrate on the c++ part.
Maybe check that you time keeping variables are all "unsigned long", because you do arithmetic on time so when the result of millis() is not kept completely that could lead to problems. However I'm not 100% if "int32" corresponds to int or long on ESP32.

wled00/FX_fcn.cpp
```diff
@@ -1511,15 +1521,28 @@ void WS2812FX::loadCustomPalettes()
...
-            DEBUG_PRINTF("%d(%d) : %d %d %d\n", i, int(tcp[i]), int(tcp[i+1]), int(tcp[i+2]), int(tcp[i+3]));
+        if (!pal.isNull() && pal.size()>4) { // not an empty palette (at least 2 entries)
+          if (pal[0].is<int>() && pal[1].is<const char *>()) {
+            // we have an array of index & hex strings
+            size_t palSize = MIN(pal.size(), 36);
+            palSize -= palSize % 2; // make sure size is multiple of 2
+            for (size_t i=0, j=0; i<palSize && pal[i].as<int>()<256; i+=2, j+=4) {
+              uint8_t rgbw[] = {0,0,0,0};
+              tcp[ j ] = (uint8_t) pal[ i ].as<int>(); // index
+              colorFromHexString(rgbw, pal[i+1].as<const char *>()); // will catch non-string entires
```
softhack007, line 1532: Just for my understanding- does this also work with both rgb (FF0C6A), and rgbw (80A899BA)?

wled00/FX_fcn.cpp
```diff
@@ -230,16 +231,28 @@ CRGBPalette16 &Segment::loadPalette(CRGBPalette16 &targetPalette, uint8_t pal) {
...
         randomPalette = CRGBPalette16(
                         CHSV(random8(), random8(160, 255), random8(128, 255)),
                         CHSV(random8(), random8(160, 255), random8(128, 255)),
                         CHSV(random8(), random8(160, 255), random8(128, 255)),
                         CHSV(random8(), random8(160, 255), random8(128, 255)));
         _lastPaletteChange = millis();
+        timeSinceLastChange = 0;
       }
-      targetPalette = randomPalette; break;
+      if (timeSinceLastChange <= 500) {
```
softhack007, line 246: timesincelastchange > 0 && timesincelastchange < 500

wled00/FX_fcn.cpp
blazoncek: Difference should always be less than uint32_t.

wled00/FX_fcn.cpp
blazoncek: Yes.
Actually code is executed every 150-250us (micro seconds). For each segment that has random palette.

wled00/FX_fcn.cpp
blazoncek: Yes, but W is ignored as CRGBPalette16 only takes RGB.

wled00/FX_fcn.cpp
blazoncek: unsigned int is always > 0 
😄

softhack007: Not exactly ... unsigned is always >= 0, so the corner case of timesincelastchange == 0 could be possible.

ChrisGebhardt: Would there be a case for basing the transition time and speed on strip.getTransition() ?  (Say 5x for palette during time, 1x for transition time, with a minimum of 5s, 500ms respectively)

blazoncek: > Would there be a case for basing the transition time and speed on strip.getTransition() ? (Say 5x for palette duration time, 1x for transition time, with a minimum of 5s, 500ms respectively)

No IMO as effect transition can be 0. We do want *some* transition for random palette change.

blazoncek: > Not exactly ... unsigned is always >= 0, so the corner case of timesincelastchange == 0 could be possible.

You got me. 😁 Still, if timesincelastchange==0 there is not palette blending happening. `for` loop takes care of that.
You do not need to do extra comparison.

Code optimisation. 😉

softhack007: Ok 👍 point for you. Then not need to change that.

blazoncek:
I'd love to see someone test this on ESP8266 as my setup may be biased. 😀

blazoncek:
I think I nailed down most if not all bugs. Do I merge?

PR #2970, opened 2022-12-23
Fix presets.json corruption

### LINKED ISSUE #2434: sorry there was an issue loading your presets
### What happened?

lost all my presets

on the preset tab i get this message
"sorry there was an issue loading your presets".

### To Reproduce Bug

dont know

### Expected Behavior

keep my preset

### Install Method

Binary from WLED.me

### What version of WLED?

v.0.13.0-b4 build 2110110

ESP32

### Relevant log/trace output

_No response_

### Anything else?

_No response_

### Code of Conduct

calumr:
Handle whitespace before any opening '{'. 

Fixes #2434

### DIFF
--- a/wled00/file.cpp
+++ b/wled00/file.cpp
@@ -137,6 +137,7 @@ bool bufferedFindObjectEnd() {
 
   if (!f || !f.size()) return false;
 
+  bool foundNonSpace = false;
   uint16_t objDepth = 0; //num of '{' minus num of '}'. return once 0
   //size_t start = f.position();
   byte buf[FS_BUFSIZE];
@@ -146,9 +147,10 @@ bool bufferedFindObjectEnd() {
     uint16_t count = 0;
     
     while (count < bufsize) {
+      if (buf[count] != ' ') foundNonSpace = true;
       if (buf[count] == '{') objDepth++;
       if (buf[count] == '}') objDepth--;
-      if (objDepth == 0) {
+      if (objDepth == 0 && foundNonSpace) {
         f.seek((f.position() - bufsize) + count +1);
         DEBUGFS_PRINTF("} at pos %d, took %d ms", f.position(), millis() - s);
         return true;


blazoncek:
This will *not* fix corruption. The corruption can also occur within preset itself.
It just detects non space characters between preset objects. Which is good but I do not understand the concept you are aiming at.

calumr:
Sorry, you're right and I should have worded this better. It won't fix corruption, it will just stop it from occurring (for the case I have found). 

When I was testing this locally `bufferedFindObjectEnd` was not returning the end of the object, so oldLen was always 1 in `writeObjectToFile`. With this fix `oldLen` is now correct.

calumr:
I explained it a bit more in the issue:

It's not handling the case where there is a space before the object begins:

"9": { "on": true, "bri": 120, ...

It gets called with f.position() pointing at the first <space> before the opening '{'. objDepth is set to 0 initially aso the while (count < bufsize) loop only gets run once.

Aircoookie (approved):
One thing I do not understand is how a space would be inserted between key and object. Regardless, this sanity check seems reasonable to do.

softhack007:
@blazoncek, @Aircoookie: should we merge this change, or not necessary? I don't know enough about file/json handling to decide what to do with this PR...

blazoncek:
IDK. As we use ArduinoJson to create a JSON object in the file there is no reason for blanks to be inserted (check `writeObjectToFileUsingId()`and `writeObjectToFile()`).

This lets me believe that either a) @calumr pasted wrong JSON or b) ArduinoJson on his system/build behaves differently.
By differently I mean it add spaces between keys and values which is very odd.

The proposed change is very subtle and should be rigorously tested before merging. I do not have time for such testing ATM.

blazoncek:
I've played a bit with this PR and was unable to reproduce insertion of spaces between preset ID and preset content by WLED alone.

But I did discover when such spaces may appear in _presets.json_, though. If you want to **manually** edit presets then minified version WLED stores is cumbersome and unwieldy so you are tempted to use some JSON prettify-er. If you then forget to minify presets and upload such *pretty* JSON to WLED it will function correctly but will fail upon modifying or saving new presets.

As such I deem this PR irrelevant as it only adds complexity in determining functionality of the function while not solving the problem entirely (spaces or tabs may be present elsewhere too). IMO a note in KB to minimize JSON after it has been manually edited should be sufficient as in-memory handling of _presets.json_ to fix minimization after the upload to WLED is infeasible.

PR #2975, opened 2022-12-26
Update Usermod Battery

itCarl:
## 📝 Changelog
- added "auto-off" feature
- added "low-power-indication" feature
- added "calibration/offset" field to configuration page
- added getter and setter, so that user usermods could interact with this one
- update readme (added new options, made it markdownlint compliant)
- added usermod/project logo
- usermod rename to just "battery" instead of "battery_status_basic"

## 🤞 Other
If i missed something, did a mistake or if you want me to change something please let me know.

Unfortunaly i have not manged to implement the "remaining battery time" feature calculation. Due to the frequent changes in the estimated current, it is difficult to establish a stable basis for calculation. In order to be able to implement such a function sensibly, one would have to oritate on already existing systems, e.g. smartphones. Because these already calculate reasonable values. Minemeinung dazu ists man nie zu 100% die genaue laufzeit "errechnen" kann. You can only make assumptions, because you can never look into the future.
An idea for the calculation would be that one takes the estimated power consumption, which is already determined by wled, for a certain time recorded and thus an average value is calculated. With this average value you can then continue.
Also notice that not all battery types have the same discharge curve. Batteries don't discharge linearly.


usermods/Battery/battery_defaults.h
```diff
@@ -0,0 +1,67 @@
+// pin defaults
+// for the esp32 it is best to use the ADC1: GPIO32 - GPIO39
+// https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/adc.html
+#ifndef USERMOD_BATTERY_MEASUREMENT_PIN
+  #ifdef ARDUINO_ARCH_ESP32
+    #define USERMOD_BATTERY_MEASUREMENT_PIN 32
```
softhack007, line 6: I'd suggest you use gpio35 by default, for two reasons:
* gpio 32 is also used as default pin in audioreactive
* on many "lolin32" boards, gpio35 is connected to battery voltage divider circuit already

itCarl: Thanks for the suggestion 👍.
I should probably order 2-3 esp32. I don't have much experience with 32s yet.
The only project I have used them on was a LoRa project.

usermods/Battery/usermod_v2_Battery.h
```diff
@@ -0,0 +1,783 @@
...
+    unsigned long nextReadTime = 0;
+    unsigned long lastReadTime = 0;
+    // battery min. voltage
+    float minBatteryVoltage = USERMOD_BATTERY_MIN_VOLTAGE;
+    // battery max. voltage
+    float maxBatteryVoltage = USERMOD_BATTERY_MAX_VOLTAGE;
+    // all battery cells summed up
+    unsigned int totalBatteryCapacity = USERMOD_BATTERY_TOTAL_CAPACITY;
+    // raw analog reading 
+    float rawValue = 0.0;
```
blazoncek, line 29: Use 0.0f.

usermods/Battery/usermod_v2_Battery.h
```diff
@@ -0,0 +1,783 @@
...
+    {
+      if (!lowPowerIndicatorEnabled) return;
+      if (lowPowerIndicationDone && lowPowerIndicatorReactivationThreshold <= batteryLevel) lowPowerIndicationDone = false;
+      if (lowPowerIndicatorThreshold <= batteryLevel) return;
+      if (lowPowerIndicationDone) return;
+      if (lowPowerActivationTime <= 1) {
+        lowPowerActivationTime = millis();
+        lastPreset = currentPreset;
+        applyPreset(lowPowerIndicatorPreset);
+        colorUpdated(CALL_MODE_DIRECT_CHANGE);
```
blazoncek, line 108: `applyPreset()` is asynchronous in 0.14 and above and does not need `colorUpdated()`.

usermods/Battery/usermod_v2_Battery.h
```diff
@@ -0,0 +1,783 @@
...
+     */
+    void setup() 
+    {
+      #ifdef ARDUINO_ARCH_ESP32
+        DEBUG_PRINTLN(F("Allocating battery pin..."));
+        if (batteryPin >= 0 && pinManager.allocatePin(batteryPin, false)) 
+        {
+          DEBUG_PRINTLN(F("Battery pin allocation succeeded."));
+        } else {
+          if (batteryPin >= 0) DEBUG_PRINTLN(F("Battery pin allocation failed."));
```
blazoncek, line 135: No need for `if` IMO

usermods/Battery/usermod_v2_Battery.h
```diff
@@ -0,0 +1,783 @@
...
+      // Auto off -- Master power off
+      if (autoOffEnabled && (autoOffThreshold >= batteryLevel))
+        turnOff();
+
+      // SmartHome stuff
+      // still don't know much about MQTT and/or HA
+      if (WLED_MQTT_CONNECTED) {
+        char subuf[64];
+        strcpy(subuf, mqttDeviceTopic);
+        strcat_P(subuf, PSTR("/voltage"));
```
blazoncek, line 206: `snprintf_P()` can do a better job.

usermods/Battery/usermod_v2_Battery.h
```diff
@@ -0,0 +1,783 @@
...
+      lp[FPSTR(_preset)] = lowPowerIndicatorPreset; // dropdown trickery (String)lowPowerIndicatorPreset; 
+      lp[FPSTR(_threshold)] = lowPowerIndicatorThreshold;
+      lp[FPSTR(_duration)] = lowPowerIndicatorDuration;
+
+      DEBUG_PRINTLN(F("Battery config saved."));
+    }
+
+    void appendConfigData()
+    {
+      oappend(SET_F("addInfo('Battery:min-battery-voltage', 1, 'v');"));
```
blazoncek, line 373: Try to use shorter names for config variables as `oappend()` buffer size is only 3kB and may be too small if more than 1 usermod is enabled.

blazoncek (commented):
No major issues found.

usermods/Battery/usermod_v2_Battery.h
```diff
@@ -0,0 +1,783 @@
...
+
+    // custom map function
+    // https://forum.arduino.cc/t/floating-point-using-map-function/348113/2
+    double mapf(double x, double in_min, double in_max, double out_min, double out_max) 
+    {
+      return (x - in_min) * (out_max - out_min) / (in_max - in_min) + out_min;
+    }
+
+    float truncate(float val, int8_t dec) 
+    {
```
softhack007, line 76: How often will you run this function?
Because it seems like a very inefficient implementation. For example pow() is really really slow, and I don't understand why to use round and then again check delta == 5.

Ok if you only need this function a few times per second. Otherwise I can give you a more efficient code without pow(), so just ask me.


How often will you run this function?
Because it seems like a very inefficient implementation. For example pow() is really really slow, and I don't understand why to use round and then again check delta == 5.

Ok if you only need this function a few times per second. Otherwise I can give you a more efficient code without pow(), so just ask me.

usermods/Battery/usermod_v2_Battery.h
```diff
@@ -0,0 +1,783 @@
...
+
+    /*
+     * setup() is called once at boot. WiFi is not yet connected at this point.
+     * You can use it to initialize variables, sensors or similar.
+     */
+    void setup() 
+    {
+      #ifdef ARDUINO_ARCH_ESP32
+        DEBUG_PRINTLN(F("Allocating battery pin..."));
+        if (batteryPin >= 0 && pinManager.allocatePin(batteryPin, false))
```
softhack007, line 131: On esp32 you need an addition condition, to check that the pin is actually supporting analogread(). 
` && (digitalPinToAnalogChannel(batterypin) >= 0)`

softhack007 (changes requested):
Thanks for your PR,

I'd like to see a few minor improvements - see previous comments
- check if your own truncate() function can be replaced by something more efficient
-use gpio35 as default on esp32
- add check that batteryPin is supporting ananlogread()

usermods/Battery/usermod_v2_Battery.h
softhack007, line 131: Please add pinowner to allocatePin() call, so the the pin is now registered for "none". 

`allocatePin(batteryPin, false, PinOwner::UM_Unspecified);`

usermods/Battery/usermod_v2_Battery.h
itCarl: 👍

usermods/Battery/usermod_v2_Battery.h
itCarl: 👍

usermods/Battery/usermod_v2_Battery.h
itCarl: 👍

usermods/Battery/usermod_v2_Battery.h
itCarl: Done. 👍
(see next commit)

usermods/Battery/usermod_v2_Battery.h
itCarl: Done. 👍
(see next commit)

usermods/Battery/usermod_v2_Battery.h
itCarl: Good point. I just wondered myself. May have been an old relict from a test i did or something.
(fixed it, see next commit)

usermods/Battery/usermod_v2_Battery.h
itCarl: Done. 👍
(see next commit)

itCarl: Done. 👍
I also added "UM_Battery" to the list in pin_manager.h.
(see next commit)

softhack007 (approved):
Thanks, changes look good 👍 I approve.

blazoncek:
👍

PR #2981, opened 2022-12-27
Big Clive's "Supercomputer" Effect

not-surt:
Simple 1D effect trying to replicate Big Clive's "supercomputer" flashing LED toy. (eg. https://youtu.be/H1k7mxhWQ5Q?t=747) 
Meant to look like the flashing lights on computers in old movies and television programmes.
LEDs flash at fixed frequency but due to variation go out of sync and appear increasingly random over time.

This was just a fun exercise so I won't be offended if you don't think this is useful enough or distinct enough to warrant a merge.

Tested on ESP8266 and ESP32 with  a 128 LED matrix.
Not really sure the best way to get an spatially independent random value per LED so I just experimented until I got a passable result.

The _Speed_ parameter specifies the cycle length.
The _Fade_ parameter specifies the proportion of the cycle length used to transition between the palette extremes. A low value is good for reproducing the effect of single colour flashing LEDs and max value good is for reproducing the effect of RGB cycling LEDs.
The _Drift_ parameter reproduces the desynchronisation by randomly adjusting the cycle length per LED.
The _Offset_ parameter adds a random offset per LED in case the start up synchronisation is undesirable.


not-surt:
Oops, just noticed a bug.

blazoncek:
Can you not do that?!

not-surt:
@blazoncek I thought I saw a bug and figured it would be better to close than risk having buggy code merged.
Turned out it was just WLED fading in the LEDs rather than a bug in my timing.
But like I said, I'm not fussed, happy enough to keep this to my self.

blazoncek:
Mark it draft in such case.

PR #2982, opened 2022-12-28
Fix Hex values in comments in pin_manager.h

itCarl:
I just noticed and it bothered me.
That's why i opened this PR


softhack007:
Looks ok for me. @blazoncek any concerns about this?

PR #2985, opened 2022-12-28
Add WLED_DISABLE_AP to completely disable the AP feature

FreezyExp:


blazoncek:
This defies the purpose of WLED.
Use your own fork for that.

FreezyExp:
Being able to just remove the unneeded functionality seems like it would not harm the main purpose of driving RGB installations.

blazoncek:
Port WLED to Arduino if you want WiFi-less variant.
It should be *fairly easy*.

FreezyExp:
The WIFI is still there and still works, I just don't need it to become it's own access point, hence the desire to be able to compile that out.

Aircoookie:
This would always require hard-coding custom WiFi credentials at compile time (or use Improv, but I would certainly not put a bin without AP support on the installer), I don't think that is practical in 99% of cases.

FreezyExp:
p.s. I calculated the space savings after removing other AP related strings & code:

#define WLED_DISABLE_AP // saves 2.57kb

PR #2986, opened 2022-12-28
Wled disable more

FreezyExp:
I shaved off another 20 KB or so by removing IR support and by adding more WLED_DISABLED wrappings around various files.
It gets kind of messy, but it removes unused structs, includes, lopping off entire blobs.

It would be nice if WLED_DISABLE was not paired with WLED_ENABLE so inconsistently, a more thorough cleanup is required.

The fcn_declare.h file was a great starting point for breaking and then fixing all related calls.
Perhaps this file could be changed into a module wrapper, which then includes the files needed for a certain functionality.

Those files could include a header per-plugin

#include "wled.h"

#ifndef WLED_DISABLE_X
// includes needed for this module
// variables
// structs 
// function & macro declarations in the order of:
// functions used from wled
// functions used within the module
// functions used by other modules
#endif

this would fascilitate a thorough cleanup of all the #ifdef / #ifndef wrappers
by replacing those with empty macros instead of empty functions

And make it a lot easier to track all the various structs / pointers / counters / variables needed for all these modules.
As those would all be moved into the module and would no longer exist in various other files.

Aside from this commit though, I currently have no time to go through the steps of this major refactoring of the code to clean up the project.

### DIFF
--- a/wled00/button.cpp
+++ b/wled00/button.cpp
@@ -25,11 +25,13 @@ void shortPressAction(uint8_t b)
   }
 
   // publish MQTT message
+  #ifndef WLED_DISABLE_MQTT
   if (buttonPublishMqtt && WLED_MQTT_CONNECTED) {
     char subuf[64];
     sprintf_P(subuf, _mqtt_topic_button, mqttDeviceTopic, (int)b);
     mqtt->publish(subuf, 0, false, "short");
   }
+  #endif
 }
 
 void longPressAction(uint8_t b)
@@ -44,11 +46,13 @@ void longPressAction(uint8_t b)
   }
 
   // publish MQTT message
+  #ifndef WLED_DISABLE_MQTT
   if (buttonPublishMqtt && WLED_MQTT_CONNECTED) {
     char subuf[64];
     sprintf_P(subuf, _mqtt_topic_button, mqttDeviceTopic, (int)b);
     mqtt->publish(subuf, 0, false, "long");
   }
+  #endif
 }
 
 void doublePressAction(uint8_t b)
@@ -63,11 +67,13 @@ void doublePressAction(uint8_t b)
   }
 
   // publish MQTT message
+  #ifndef WLED_DISABLE_MQTT
   if (buttonPublishMqtt && WLED_MQTT_CONNECTED) {
     char subuf[64];
     sprintf_P(subuf, _mqtt_topic_button, mqttDeviceTopic, (int)b);
     mqtt->publish(subuf, 0, false, "double");
   }
+  #endif
 }
 
 bool isButtonPressed(uint8_t i)
@@ -120,12 +126,14 @@ void handleSwitch(uint8_t b)
     }
 
     // publish MQTT message
+    #ifndef WLED_DISABLE_MQTT
     if (buttonPublishMqtt && WLED_MQTT_CONNECTED) {
       char subuf[64];
       if (buttonType[b] == BTN_TYPE_PIR_SENSOR) sprintf_P(subuf, PSTR("%s/motion/%d"), mqttDeviceTopic, (int)b);
       else sprintf_P(subuf, _mqtt_topic_button, mqttDeviceTopic, (int)b);
       mqtt->publish(subuf, 0, false, !buttonPressedBefore[b] ? "off" : "on");
     }
+    #endif
 
     buttonLongPressed[b] = buttonPressedBefore[b]; //save the last "long term" switch state
   }
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -52,6 +52,7 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
     CJSON(staticSubnet[i], nw_ins_0_sn[i]);
   }
 
+#ifndef WLED_DISABLE_AP
   JsonObject ap = doc["ap"];
   getStringFromJson(apSSID, ap[F("ssid")], 33);
   getStringFromJson(apPass, ap["psk"] , 65); //normally not present due to security
@@ -71,6 +72,7 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
     apIP[i] = ap_ip;
   }
   */
+#endif
 
   noWifiSleep = doc[F("wifi")][F("sleep")] | !noWifiSleep; // inverted
   noWifiSleep = !noWifiSleep;
@@ -642,6 +644,7 @@ void serializeConfig() {
     nw_ins_0_sn.add(staticSubnet[i]);
   }
 
+#ifndef WLED_DISABLE_AP
   JsonObject ap = doc.createNestedObject("ap");
   ap[F("ssid")] = apSSID;
   ap[F("pskl")] = strlen(apPass);
@@ -653,7 +656,8 @@ void serializeConfig() {
   ap_ip.add(4);
   ap_ip.add(3);
   ap_ip.add(2);
-  ap_ip.add(1);
+  ap_ip.add(1);  
+#endif
 
   JsonObject wifi = doc.createNestedObject("wifi");
   wifi[F("sleep")] = !noWifiSleep;
@@ -1006,8 +1010,10 @@ bool deserializeConfigSec() {
   JsonObject nw_ins_0 = doc["nw"]["ins"][0];
   getStringFromJson(clientPass, nw_ins_0["psk"], 65);
 
+#ifndef WLED_DISABLE_AP
   JsonObject ap = doc["ap"];
   getStringFromJson(apPass, ap["psk"] , 65);
+#endif  
 
   JsonObject interfaces = doc["if"];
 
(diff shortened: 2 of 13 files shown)


wled00/button.cpp
```diff
@@ -25,11 +25,13 @@ void shortPressAction(uint8_t b)
   }
 
   // publish MQTT message
+  #ifndef WLED_DISABLE_MQTT
   if (buttonPublishMqtt && WLED_MQTT_CONNECTED) {
     char subuf[64];
     sprintf_P(subuf, _mqtt_topic_button, mqttDeviceTopic, (int)b);
     mqtt->publish(subuf, 0, false, "short");
   }
+  #endif
```
FreezyExp, line 34: this pattern is repeated a lot in this file, consider writing a helper functions.

wled00/const.h
```diff
@@ -8,10 +8,12 @@
 #define GRADIENT_PALETTE_COUNT 58
 
 //Defaults
+#ifndef USE_MY_CONFIG
 #define DEFAULT_CLIENT_SSID "Your_Network"
 #define DEFAULT_AP_SSID     "WLED-AP"
 #define DEFAULT_AP_PASS     "wled1234"
 #define DEFAULT_OTA_PASS    "wledota"
+#endif
```
FreezyExp, line 16: this should then also be added to the my_config_sample
this removed warnings from the compiler when replacing these from my_config

wled00/fcn_declare.h
```diff
@@ -9,10 +9,12 @@
  */
 
 //alexa.cpp
+#ifndef WLED_DISABLE_ALEXA
```
blazoncek, line 12: All these are irrelevant as they are mere declarations.

wled00/wled.cpp
```diff
@@ -42,7 +42,9 @@ void WLED::loop()
   #endif
 
   handleTime();
+#ifndef WLED_DISABLE_IR
```
blazoncek, line 45: Please use same indentation across the file.

wled00/wled.cpp
```diff
@@ -484,6 +490,9 @@ void WLED::beginStrip()
 
 void WLED::initAP(bool resetAP)
 {
+#ifdef WLED_DISABLE_AP
```
blazoncek, line 493: I will not approve this.

wled00/wled.h
```diff
@@ -641,7 +645,9 @@ WLED_GLOBAL byte optionType;
 
 WLED_GLOBAL bool doSerializeConfig _INIT(false);        // flag to initiate saving of config
 WLED_GLOBAL bool doReboot          _INIT(false);        // flag to initiate reboot from async handlers
+#ifndef WLED_DISABLE_MQTT
```
blazoncek, line 648: While you're at it, group MQTT related vars together.

blazoncek (changes requested):
I appreciate your effort in trying to minimize code size but ...

Do not sneak disabling AP into otherwise benign PR.

wled00/const.h
softhack007: I agree with @blazoncek - removing preprocessor #define will save exactly zero bytes by itself. But it makes the code harder to read.

wled00/fcn_declare.h
```diff
@@ -165,8 +173,10 @@ bool parseLx(int lxValue, byte* rgbw);
 void parseLxJson(int lxValue, byte segId, bool secondary);
 
 //mqtt.cpp
+#ifndef WLED_DISABLE_MQTT
```
softhack007, line 176: For all changes you propose in fcn_declare.h, the same is true: removing prototype declarations saves exactly zero bytes.
Even with prototypes, there will be linker errors when the function implementation is missing. So I see no good reason to remove these prototypes.

wled00/const.h
FreezyExp: Perhaps a CUSTOM_AP_SSID, CUSTOM_AP_PASS, etc, but that would make for even clunkier code.
Wrapping this in the WLED_DISABLE_AP I proposed would also work, as these are not needed when the AP function of the WIFI is disabled. Wifi still works, it just required compiling in the wifi-password. And since this is a custom build only thing, that should work just fine.

wled00/fcn_declare.h
FreezyExp: Removing the prototypes by itself indeed does very little, but it gives compile errors that can then be removed by marking the non-functional code with the appropriate wrappings. 

As a second step I went into the declaration sites and ensured they and their headers were also removed.

Overall those #ifndef / #ifdef locations could be made more readable if they were properly abstracted away.
But that is not the state of the project and went beyond what I needed to achieve.

All I know is that at the end of this PR the build was small enough to allow an OTA update using a workaround I found in one of the other PRs. While still allowing for websockets so that Home Assistant works.

blazoncek:
I will add this here since I closed the other PR.

People get binaries from everywhere, without knowing who or how somebody compiled them. Hence a missing AP will make troubleshooting issues that much more difficult as holding a button for >6s or factory resetting are two of the most common steps to get AP working.

Disabling AP during compile time **will not** make it into official code. Period.

As I said in the closed PR: Use your own fork if you need AP-less WLED. If you do it properly you will not have much issues when pulling from upstream.

softhack007 (changes requested):
Hi, thanks for the contribution.
Saving some flash is always desirable, as we are always fighting to fit everything into our devices. Most changes look OK to me, except for some modifications of header files, and introduction WLED_DISABLE_AP.

I've tried to build your version, and compared to "main" in AC.

**env:esp32dev_qio80**
Flash saved: 1240 bytes plus another 2164 bytes with WLED_DISABLE_AP = 0.2% of flash capacity

**env:esp01_1m_full**
Flash saved: 884 bytes, plus another 1724 bytes with WLED_DISABLE_AP = 0.2% of flash capacity

🤔  Maybe I did something wrong, because I could not reproduce your savings. Please advise.

<br/>

> The fcn_declare.h file was a great starting point for breaking and then fixing all related calls. Perhaps this file could be changed into a module wrapper, which then includes the files needed for a certain functionality [...].


Honestly I'm not a fan of this idea (but you may convince me). The reason is, almost all our "modules" are single .cpp files already. It will create more chaos if all "data and functions used within the module" would be added to global header files. 

Also adding "wrapper functions" to completely de-activate a feature might have unwanted effects in some cases, as it either creates another function call (as a "real" wrapper), or may lead to strange side-effects like calling (void)(0) for deactivated functions (as a macro).

In contrast to languages like java, C/C++ has no clear notion of "modules". I think that const.h / wled.h already are a good solution. If we go further, that would mean to introduce C++ _namespaces_, and we are really far away from that...

<br/>

I have left some comments to your modifications, please consider them.
Cheers, and thanks for your effort to make WLED better 👍

wled00/cfg.cpp
```diff
@@ -52,6 +52,7 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
     CJSON(staticSubnet[i], nw_ins_0_sn[i]);
   }
 
+#ifndef WLED_DISABLE_AP
```
softhack007, line 55: @duncan we already explained that WLED_DISABLE_AP will not be accepted in the main WLED repo. Please keep WLED_DISABLE_AP parts out of this PR, and maintain them in your own fork.

softhack007 (changes requested):

blazoncek:
If you really want to disable AP and have this PR approved do it in such way as to only modify default state which is available in `apBehavior` variable.

I would recommend only changing line 293 in wled.h to read:
```c++
#ifdef WLED_DISABLE_AP
WLED_GLOBAL byte apBehavior _INIT(AP_BEHAVIOR_BUTTON_ONLY);       // access point opens only by pressing button 0 for >6s
#else
WLED_GLOBAL byte apBehavior _INIT(AP_BEHAVIOR_BOOT_NO_CONN);       // access point opens when no connection after boot by default
#endif
```

FreezyExp:
> 

Thank you for the explanation that DISABLE_AP would cause factory reset debug issues, now the decision makes sense to me.
When I have time I will look into splitting that de-feature out.

FreezyExp:
> If you really want to disable AP and have this PR approved do it in such way as to only modify default state which is available in `apBehavior` variable.
> 
> I would recommend only changing line 293 in wled.h to read:
> 
> ```c++
> #ifdef WLED_DISABLE_AP
> WLED_GLOBAL byte apBehavior _INIT(AP_BEHAVIOR_BUTTON_ONLY);       // access point opens only by pressing button 0 for >6s
> #else
> WLED_GLOBAL byte apBehavior _INIT(AP_BEHAVIOR_BOOT_NO_CONN);       // access point opens when no connection after boot by default
> #endif
> ```

My entire reason was to find things I could disable to shave off enough to be able to do OTA again on 1M ESP devices.
As removing them from their built in locations is a delicate chore as I had to solder additional wires directly to the chiplet.

blazoncek:
> My entire reason was to find things I could disable to shave off enough to be able to do OTA again on 1M ESP devices. As removing them from their built in locations is a delicate chore as I had to solder additional wires directly to the chiplet.

You cannot.
I posted a procedure in issues that will allow 1M devices to be updated OTA. It requires 2 step process with *minimal* uploader.
Use that.

blazoncek:
Closing in favor of #3022 which shaved a substantial amount of flash bytes depending on build options.
Also disabling compilation of AP functionality is not an option.

PR #2988, opened 2022-12-29
Fix for displaying 1bpp bmp files (usermod EleksTube IPS)

dvdavide:
When working with a 1 bit per pixel bmp file, this forces the color palette to zero and something other than zero (tried 255 and it worked for me). Before this fix, the monochrome bmp would display fine in the file manager but would display all black on the ips screen.

### DIFF
--- a/usermods/EleksTube_IPS/TFTs.h
+++ b/usermods/EleksTube_IPS/TFTs.h
@@ -141,7 +141,7 @@ private:
     h = read32(bmpFS); // height
     (void) read16(bmpFS); // color planes (must be 1)
     bitDepth = read16(bmpFS);
-    
+
     if (read32(bmpFS) != 0 || (bitDepth != 24 && bitDepth != 1 && bitDepth != 4 && bitDepth != 8)) {
       Serial.println(F("BMP format not recognized."));
       bmpFS.close();
--- a/usermods/EleksTube_IPS/TFTs.h
+++ b/usermods/EleksTube_IPS/TFTs.h
@@ -133,38 +133,34 @@ private:
       return false;
     }
 
-    read32(bmpFS); // filesize in bytes
-    read32(bmpFS); // reserved
+    (void) read32(bmpFS); // filesize in bytes
+    (void) read32(bmpFS); // reserved
     seekOffset = read32(bmpFS); // start of bitmap
     headerSize = read32(bmpFS); // header size
     w = read32(bmpFS); // width
     h = read32(bmpFS); // height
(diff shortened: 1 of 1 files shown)


softhack007:
@blazoncek could you also take a look at the "legacy" code of this usermod? Especially the line below is a bit unusual - is is ok that usermods call strip.service() directly?

https://github.com/dvdavide/WLED/blob/dc2cfee75754b0ec6e071355ed51905606fa8227/usermods/EleksTube_IPS/TFTs.h#L78
>     if (!realtimeMode || realtimeOverride || (realtimeMode && useMainSegmentOnly)) strip.service();

Same concerns for the usermod loop(), which directly fiddles around with segment data structure, I'm wondering if that's the correct way to handle segments...

usermods/EleksTube_IPS/TFTs.h
```diff
@@ -141,14 +141,18 @@ class TFTs : public TFT_eSPI {
...
       bmpFS.close();
       return false;
+    } else {
+      if (bitDepth == 1) monochromatic = true;
     }
 
     uint32_t palette[256];
+    uint32_t palette0 = 255;
     if (bitDepth <= 8) // 1,4,8 bit bitmap: read color palette
     {
```
softhack007, line 157: Coding style : use `(void) read32(bmpFS)` if you want to ignore return value of your function. This will avoid some compiler warnings especially for new MCUs. Also it makes the code clearer, as you explicitly say "yes I want to ignore the function return value this time".

blazoncek:
Unfortunately I never used EleksTube so know nothing of it.
As far as I am concerned every usermod is an optional thing so, apart from excessive RAM or flash use, I am relatively ok with whatever contributors provide.
Calling `strip.service()` is a bit unusual though.

Perhaps @Aircoookie knows more.

Aircoookie:
`strip.service()` is called periodically, because reading the BMP and drawing it to the screen is slow and blocks for too long, thus causing low framerates in effects.

Not a fan of hardcoding 255, since this will discard the color value saved in the bitmap palette. Will need to give it a try later. @dvdavide would you provide an example 1bpp bitmap for testing? It might be that your image editor handles BMP palettes differently from paint.NET.

dvdavide:
You can find an example of the fonts I use at https://github.com/dvdavide/Elekstube-fonts
I converted them in ms paint (windows 11), and as i said they show up fine in the wled file manager preview, but the tft display shows black, and not stuck on last picture as when a file is missing/unreadable.

PR #2993, opened 2022-12-31
fixes typo in Analog_Clock.h

spdustin:
Retrieving the config in `readFromConfig()` results in defaults being set to both `hourMarksEnabled` (`false`) and `hourMarkColor` (`#0000FF`) due to differences in capitalization compared to how they're saved in `addToConfig()`

### DIFF
--- a/usermods/Analog_Clock/Analog_Clock.h
+++ b/usermods/Analog_Clock/Analog_Clock.h
@@ -56,7 +56,7 @@ private:
 
     // runtime
     bool     initDone         = false;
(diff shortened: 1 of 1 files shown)


blazoncek:
Can you, while you're at it, add flash string helper `F()`?

spdustin:
Apologies, that commit has a few whitespace deletes as well. But I think I wrapped all the strings, and it passed my local unit and on-device tests.

spdustin:
It also sets a more rational default of `#161616` for hour marks (if enabled).

PR #2994, opened 2023-01-01
added leading zero to #HHMM #DDMM #MMDD

relax81:
Added the leading zero to make it look more consistent and centered on the X-Axis of the matrix.


blazoncek:
I do not want leading zeros on date and hour.

Aircoookie:
It depends on the use case what style is preferable. Perhaps these should be accessible using separate strings, like `#0DMM`.

softhack007:
Why not use strftime, and let users set their preferred format in WLED settings? This would cover all "preferred formats" which are very different around the word...

https://cplusplus.com/reference/ctime/strftime/

blazoncek:
Too geeky. 😉

softhack007:
Where are the good old times 😛 when time format strings were well-known, time was in GMT, and everybody knew how to read unix times 😁

relax81:
Looks like it's not as easy as I thought :) My thought was that it would make sense to have always 4 digits when it's called HHMM or DDMM, but people seem to be different preferences. . Should I do Aircookies approach then with #0DMM or will it be too many options?

blazoncek:
For centering display (if there is such thing; perhaps I did implement it, I do not recall now) it would be enough to change `%d.%d` into `%2d.%d` (for both day/month). This will add a leading blank to first number.
The same can be made for hour mark.

Aircoookie:
Another option would be the format I originally made for Cronixie: https://github.com/Aircoookie/WLED/blob/main/usermods/Cronixie/usermod_cronixie.h#L52

In any case, the format would ideally be consistent across all date/time display functionality in WLED. I might make a new format that can be used here too.

YvesLev:
Hello!
I use #HHMM and it is working but it would be way better if the text would be centered for hours between 0 and 9.
(8:45 vs 14:13).
On my 16 X 32 Matrix, the Hour:Minute on first line (Segment 0) is not centered compared to the Temperature on second line (Segment 1)...
It does not look neat from 0:00 to 9:59 :-) .

blazoncek:
Temperature? There is no option to display temperature on Scrolling text effect/mode.

YvesLev:
> Temperature? There is no option to display temperature on Scrolling text effect/mode.

I'm using a rest command in Home Assistant to send the temperature.
But that's not the point.
The point is that any text of less than 5 characters (to avoid scrolling) you put in a segment name (Matrix 32w X 8h) gets centered but this is not true for the #HHMM format.

YvesLev:
Personaly, for the Hour, I would have prefered no leading space, no leading zero and just get the whole string (H:MM) centered no matter if there are 4 or 5 digits( 1:24, 12:45, the "1:45" should be 4 chars centered.  Not "01:45", Not " 1:45"). 
Maybe the format tag should be #HMM rather than #HHMM?

Yves

blazoncek:
I hope I have earned a beer. 😉

YvesLev:
Problem solved!
That was quick!

Big thanks!

IMG_6267

YvesLev:
However, in the PEEK view (PEEK button in the top menu bar in the wled UI), the #HHMM does not seem centered...

Not a deal breaker though...  just letting you know...

PR #3005, opened 2023-01-06
whitespace cleanup

blazoncek:
OCD kicked in.
It may wreck havoc on other PRs so I do not expect this to be merged but I had to try.
What do you say @Aircoookie ?


Aircoookie (approved):
Indeed likely to cause a merge conflict here or there, which should be easy to fix though.

spdustin:
This is a very welcome PR, will make my own future PRs a lot easier to manage. No more cherry picking!

PR #3010, opened 2023-01-09
Api based text overlay usermod

vasiliucatalinmihai:
I know thare are some ppl that was searching for this. Is just a simple overlay that renders a text on the matrix based on some data on the json api

### DIFF
--- a/usermods/usermod_v2_text_overlay/readme.md
+++ b/usermods/usermod_v2_text_overlay/readme.md
@@ -1,0 +1,53 @@
+# Text overlay
+* works only on matrix
+
+This user mod adds an text on top of whatever the matrix is rendering (effects, solid colors ...)
+
+I built this UserMod for sending from home assitant the text, notification like :)
+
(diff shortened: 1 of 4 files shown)


blazoncek:
Thank you for the effort but unfortunately this may already be obsolete (except for perhaps longer text, but that is to be determined) as just tonight we added overlay functionality to ScrollingText effect.
You can set up as many as you like (up to 32) segments with text overlay using overlapping segments.

Aircoookie:
Closing this for now in favor of the native Scrolling Text effect. If this covers a use case not possible with the effect, please comment and I will reopen and consider merging.

blazoncek:
I have considered adding the ability to use a text file stored on FS.
Unfortunately reading text from FS (i.e. flash memory) and at the same time reading character data from PROGMEM (flash) yielded a lot of LED flickering.
Perhaps reading a file into RAM would be better so may consider that in the future.

PR #3012, opened 2023-01-10
Fixes tooltip interfering with pointer events

spdustin:
This commit adds `pointer-events: none;` to `.slider .tooltiptext, .option .tooltiptext` selectors in `index.css`, and includes compiled C code for the web UI.

The effect slider tooltip has a transition that causes it to catch pointer events, and it won't go away until the mouse is entirely out of its bounding box. As a result, if you're adjusting any slider below the first one, the tooltip gets in the way of you modifying the one above it.

By adding `pointer-events: none;` to the tooltip, the tooltip is no longer aware of the mouse/pointer, and closes itself immediately after your mouse leaves the bounds of the element that called it in the first place.


spdustin:
Apologies: this commit also fixes extra whitespace in `index.css` that @blazoncek missed during his recent whitespace-killing spree. 😉

blazoncek:
You are a life saver!

PR #3015, opened 2023-01-11
Usermod seven segment display reloaded. Add the ability to configure the previous zero hour setting. Fix for using led mapping

dvino:
Usermod seven segment display reloaded

Add the ability to configure the previous zero hour setting.
To work correctly with LED mapping, the way the usermod mask is applied has been changed.


usermods/seven_segment_display_reloaded/usermod_seven_segment_reloaded.h
```diff
@@ -246,7 +246,7 @@ class UsermodSSDR : public Usermod {
   void _setMaskToLeds() {
     for(int i = 0; i <= umSSDRLength; i++) {
       if ((!umSSDRInverted && !umSSDRMask[i]) || (umSSDRInverted && umSSDRMask[i])) {
-        strip.setPixelColor(i, 0x000000);
+        busses.setPixelColor(i, 0x000000);
```
blazoncek, line 249: Use of `busses` is incorrect and may break in the future. This variant does not respect ledmaps.
Use either `SEGMENT.setPixelColor()` or `strip.setPixelColor()`.

You may even employ `handleOverlayDraw()` if needed.

blazoncek (changes requested):
Use of internal `busses.setPixelColor()` is inappropriate.

usermods/seven_segment_display_reloaded/usermod_seven_segment_reloaded.h
dvino: I used busses.setPixelColor() instead of strip.setPixelColor() precisely to eliminate the influence of ledmask on the usermod mask. Usermod mask umSSDRMask should not depend on the selected ledmask. It is necessary to extinguish certain LEDs in the physical sequence, not in the one that is motivated by the ledmask. This means that if usermod has extinguished physical LEDs 1 and 2, then no matter which ledmask is selected these LEDs will not light up.

dvino: Since the variables customMappingTable and customMappingSize are private, I cannot use them to calculate the physical LED position.

blazoncek: Ledmaps (not ledmasks) are user created files and as such a user is responsible for their content. If a user decides to remap SSD LEDs his/her choice should be observed regardless of your opinion as otherwise behaviour becomes inconsistent.

I will not approve the use of `busses.setPixelColor()` in a mainstream WLED.

dvino: You don't quite understand why I do this. This usermod assumes that the position of the LEDs is not quite suitable for creating effects. I use ledmap for this purpose. But there can be several ledmap files. For example ledmap1.json for direction effects from top down, ledmap2.json for effects fromleft to right of the clock matrix.
In the usermod settings you enter the physical location of the desired display segments. But when applying ledmap and using `strip.setPixelColor` to create the ledmask, the position of the LEDs must also be recalculated. 

There are several ways to solve this situation.
- Using `busses.setPixelColor()` to apply the ledmask to specific LEDs, bypassing the ledmap.
- Using `strip.setPixelColor` and recalculating ledmask according to ledmap using values of variables [customMappingTable](https://github.com/Aircoookie/WLED/blob/main/wled00/FX.h#L887) and [customMappingSize](https://github.com/Aircoookie/WLED/blob/main/wled00/FX.h#L888). But these variables are in the private area of the  class `WS2812FX`. To do that need have to make them public.
- Also as in the previous example it is necessary to recalculate ledmask, but for this during initialization from the preset `ledmapX.json` will have to read this file repeatedly at the usermod. I think this is unreasonable.
- In usermod settings to write necessary masks for each ledmapX file and load already needed one. But usermod settings becomes unreasonably cumbersome. At the time when all this can be implemented in usermode code.

Do you have a different algorithm idea?

> If a user decides to remap SSD LEDs his/her choice should be observed regardless of your opinion as otherwise behaviour becomes inconsistent.

Just the opposite. The current implementation is misleading and does not allow multiple ledmap files. When loading a new ledmap file you need to recalculate the values of [these variables](https://github.com/Aircoookie/WLED/blob/main/usermods/seven_segment_display_reloaded/usermod_seven_segment_reloaded.h#L17-L23) manually according to the ledmap and enter new values in the settings. Initially the code was created without the possibility of using multiple ledmap files

blazoncek: Rethink your approach.

DJprasenjit:
in esp8266 this not working. please tell me fix

blazoncek:
@DJprasenjit Use Discord or WLED [forum](https://wled.discourse.group) for help and support questions.

blazoncek:
@dvino as there are no updates to this PR I am closing it within week without merging.
If you still want to pursue it further please update as I requested or close if you do not want to.

PR #3022, opened 2023-01-12
Disable MQTT more.

blazoncek:
Disable Alexa more.

Alternative to #2986


softhack007:
Good idea 👍 especially when considering that #2986 saved only 1240  bytes, when not removing AP mode.

blazoncek:
@Aircoookie your take? Useful?

Aircoookie (approved):
Looks very good 👍

PR #3052, opened 2023-01-24
sACN/E1.31 Port Priority

mxklb:
As of https://github.com/Aircoookie/WLED/issues/768 I implemented some kind of minimal E1.31 basic port priority handling some days ago.

It's not implemented as the [standard](https://support.etcconnect.com/ETC/Networking/General/Difference_between_sACN_per-address_and_per-port_priority) says, but may be useful in situations where multiple senders can't be avoided:

    When UI setting for E1.31 Port Priority != 0
        
        Senders with priority != configured E1.31 Port Priority will be skipped

With this minimal code change it's possible to define WLED instances to act on specific sender priorities only.

Note: Using multiple E1.31 senders is not recommended. Furthermore, multiple senders will reduce performance and it's strictly advised to avoid it. Anyway if avoidance is not possible, this would at least handle crossfire more better than now.


mxklb:
To implement the standard including Highest/Latest Takes Precedence (HTP/[LTP](https://wiki.openlighting.org/index.php/LTP)) there must be a more general refactoring. It needs a wrapper class around the access of `e131_data`, combined with per device incoming package tracking. Every line with `e131_data[x]` access would need a change. IMO this may be implemented later on ..

I think this PR, handling multiple sender packages by priority, could solve some related issues.

Help wanted: Would be nice if someone could test this PR :face_in_clouds: 
- using multiple E1.31 senders with different priorities,
- and provide feedback or suggestion for improvement

mxklb:
Please use discourse for discussion & feedback https://wled.discourse.group/t/e1-31-port-priority-beta-testing/7270

RichardTea:
This PR is about 90% of the way to fully complying with the standard.
Pretty much all that's needed is to delete the UI to set a priority value - you don't need it.

The basic priority algorithm is as follows:

Store the current priority value (1 byte) and the UUID (16 bytes) for each universe.

When a packet is received:

1. Check the universe value. If the universe is not used, ignore the packet.
2. Check the Start Code value. (slot 0 in the payload). If nonzero, ignore the packet. (Very important)
3. Check the priority value. If it is smaller than the current universe priority value, ignore the packet.
4. If it is equal to the current priority value and the current universe UUID is not null, compare the UUID for the universe. If the UUID is not equal, ignore the packet.
5. If it is greater than the current priority value, store the priority and UUID as current for the universe.
6. Restart the per-universe timeout timer and accept the DMX levels into the system

When the timeout timer expires, reset the priority and UUID to zero, and start the data-lost behaviour.

The above is compliant but not great as doesn't merge multiple same-priority sources and it will briefly flicker to data-lost when the 'highest priority' source goes away, but it's close enough for many purposes. If there's a short fade from E1.31 to internal programs then that flicker wouldn't be visible in practice.

Professional products use a more complex (and memory-hungry) method to avoid the 'step-back' flicker, but given that WLED currently flickers like crazy with two or more active sources I don't think it's a problem.

mxklb:
> Pretty much all that's needed is to delete the UI to set a priority value - you don't need it.

I am not sure about that. Actually with UI priority == 0, priority handling is completely disabled, it behaves as before. This means no priority handling at all = no performance loss due to its implementation. When deleting UI priority value, priority handling would be enabled all the time. Due to the fact that using multiple senders is more an edge case than default WLED E1.31 usage, IMO it shall be better to have it switchable on demand somehow (could also be discussed to only use compiler flag, simple UI switch or anything else ..). 

++ Additional: UI priority values > 0 can explicitly be set to skip lower priority senders, which IMO is also nice to have.

More a matter of documentation and what WLED likes to have. To [achieve last 20%](https://en.wikipedia.org/wiki/Pareto_principle) would be really complicated :smile: 

I don't know what's best here, I think **maintainers shall decide**:

- Always On (no UI, not switchable)
- UI value incl. skip lower priorities (as is)
- Compile option or UI switch (default = on|off ???).

blazoncek:
@mxklb IMO your approach is best. Allow wider set of choices without sacrificing performance and at the same time allow user to make a choice.

mxklb:
From my point of view this is finished. Ready for testing and to be merged ..

wled00/wled.h
```diff
@@ -403,6 +403,8 @@ WLED_GLOBAL uint16_t e131ProxyUniverse _INIT(0);                  // output this
 #endif
 WLED_GLOBAL uint16_t e131Universe _INIT(1);                       // settings for E1.31 (sACN) protocol (only DMX_MODE_MULTIPLE_* can span over consequtive universes)
 WLED_GLOBAL uint16_t e131Port _INIT(5568);                        // DMX in port. E1.31 default is 5568, Art-Net is 6454
+WLED_GLOBAL byte e131Priority _INIT(0);                           // E1.31 port priority (if != 0 priority handling is active)
```
Aircoookie, line 406: We could probably default this to 1 to be more standard compliant in the 99% of cases where the user doesn't care about priorities. If I understand correctly, in order to be fully compliant we'd need to do separate priority handling for each universe, which I would prefer not to expend memory on, given that priority handling is only required in the first place if the user doesn't follow best practices (one sender at a time)

Aircoookie (approved):

wled00/wled.h
mxklb: 0 = priority handling OFF = as actually is, without priority handling. So default 0 should be good in 99%.

> If I understand correctly, in order to be fully compliant we'd need to do separate priority handling for each universe

The only thing missing is to handle HTP = [Highest Takes Precedence](https://wiki.openlighting.org/index.php/HTP) for same priority sources. This means, merging each address value of same priority sources by HTP algorithm. This would need a refactoring as pointed out [here](https://github.com/Aircoookie/WLED/pull/3052#issuecomment-1407229536). If standard compliant is really wanted, this shall be covered by another additional PR .. 

WLED "best practice": Nevertheless it's a common way for E1.31 gear to take over / switch controllers by priority.

With this PR users are able to switch E1.31 senders without flickering, when switching to higher priority senders. When switching to lower priority senders a gap of `timeout=3` seconds is implemented, as higher priority senders disappear. As pointed out [here](https://github.com/Aircoookie/WLED/pull/3052#issuecomment-1407713063), this 'step back' is no big issue. It could be discussed if default `timeout=3` is to long, smallest possible would be `1`. Actually I don't know if `1s` is enough, so I leave it at `3s` for now. This highest priority tracking timeout may be changed in future, if E1.31 priority users frequently report it is to long .. TBD

mxklb: > We could probably default this to 1 to be more standard compliant in the 99% of cases

Of cause we could also init default `e131Priority` with 1, so priority handling is always active. This is also good but may impact performance due to `highPriority` handling some what. I think performance drop due to handling as implemented should not be noticeable. Finally this is an opt in|out decision, as you like ..

mxklb:
@RichardTea I can cherry pick your commit into here if you like. Looks correct, but I can't test this.

mxklb:
@RichardTea thank you for your improvement. I cherry-picked your changes into here.

Finally I found some time to test it a little. At least with one E1.31 controller.

I can confirm by personal tests:

- with or without priority handling with one single E1.31 sender everything works as expected/before
- E1.31 packages with priority < UI config are successfully skipped, only priority >= gets processed
- @RichardTea changes looking good, doesn't harm actual behavior ..

Still missing tests. Due to lack of E1.31 hard/software features:

- @RichardTea changes: 1. Preview data & 2. Non-zero start code
- Priority device 'take-over' using higher priority packages (while still receiving lower priority packages)
- Tracking & timeout handling, when highest priority sender stops sending ('step-back' as described above)

So far, everything I could test is working as expected. Let's see if someone can test missing parts ..

RichardTea:
Easiest test tool is: https://sacnview.org/

There's prebuilt installers for Windows, macOS, Linux (deb) and Linux (appimage)

Open Transmit windows (wifi-like icon)

There is an option labelled "blind-mode data" to send Preview data.
It should be possible to have two Transmit windows active on the same universe at once, one doing "blind mode" and the other 'live', and WLED should be ignoring the blind-mode levels. Previously it would flicker between the two.

To test the startcode, set a single transmit window to Priority Mode: Per Source
This will make it send both zero start code level data and 0xDD start code per-channel priority data. Per-channel priority should be ignored by any receiver that doesn't support that feature.

Note that every sACNView Transmit window currently has a different CID, so it's not currently possible to use it to test one source sending multiple universes at the same time. There's already an open issue in the sACNView project for that.

mxklb:
Thanks for the tool, I already had an older sacnview appImage, but didn't remember to use that :)

I tested multiple senders to take over by priority and can confirm that it is working as expected.
Also the timeout 'step-back' is working  as expected, without noticeable flicker.
Additional I didn't recognize any performance loss due to WLED priority handling.

> There is an option labelled "blind-mode data" to send Preview data.

Sending "blind-mode" preview data it is now correctly ignored, confirmed.

> To test the startcode, ..

WLED receives data, nevertheless if priority mode 'Per Source' or 'Per Channel' is selected.

- Maybe there is still something missing, I couldn't figure out .. or I don't know how to ..

This PR is now mostly finished and should be a good sACN improvement

wled00/wled.h
mxklb: As of my latest test comment
> Additional I didn't recognize any performance loss due to WLED priority handling.

Defaulting to priority handling always ON and eventually removing UI config (as @RichardTea mentioned) is still an option. Doing so, none of the different use cases / users have to worry about at all. But at the cost of the ability to always skip packages lower UI config. And a the little performance loss due to priority handling .. TBD

@Aircoookie @blazoncek Please decide & resolve how to finally implement this.

blazoncek:
As I lack any DMX or other "live" experience I will trust that you know what are you doing and have much greater knowledge that I do.
So, just get a few fellow "live" users to test and if the feedback is positive we'll merge these changes in.

mxklb:
I already promoted it to find testers: https://wled.discourse.group/t/e1-31-port-priority-beta-testing/7270

Sadly they are the minority of WLED users and finding 'live' testers/debuggers is even worse .. let's wait.

So far big thanks to @RichardTea maybe he can additional test startcode and overall behaviour ..

RichardTea:
> > To test the startcode, ..
> 
> WLED receives data, nevertheless if priority mode 'Per Source' or 'Per Channel' is selected.
> 
> * Maybe there is still something missing, I couldn't figure out .. or I don't know how to ..

The old, incorrect behaviour was that WLED would occasionally briefly flash the strings to the Per-Channel-Priority values before returning to the proper RGB(W) colours.

The bug is most visible with static (unchanging) levels of all-at-zero and per-channel-priority at the default of all-at-100. It then flickers some or all of the strings to a little under 50% for a random period up to ~800-900ms before returning back to the proper static "off" level.

With the patch above, that flash should never happen.

swiss6th:
> The bug is most visible with static (unchanging) levels of all-at-zero and per-channel-priority at the default of all-at-100. It then flickers some or all of the strings to a little under 50% for a random period up to ~800-900ms before returning back to the proper static "off" level.

I believe I observed this behavior last year; couldn't figure out the issue. I assumed it was a hardware flaw in my design, but this describes it perfectly.

> With the patch above, that flash should never happen.

This is great! Thank you, @mxklb and @RichardTea, for all your work on this PR. 😄

bobh66:
To add a data point - I just finished a production where I used this to override the local architectural system (ETC Paradigm) which was sending out priority 100 packets and the WLED receivers would toggle between 0 from the arch system and the values that my ETC Element was sending.  I changed the priority on the Element and WLED to 101 and it worked great.

Thanks for the feature - I really appreciate it!

PR #3060, opened 2023-01-29
Dual mode 2D + 1D with auto segment creation.

blazoncek:
Added support for combining matrix and a strip to a single WLED instance.
Matrix has to be wired first, followed by strip.
Both matrix and strip support multiple segments. Some segment validation included.


Aircoookie:
Seems to work well!

Issue #2983, opened 2022-12-28
WLED devices expose an impossible (or undocumented) light capability


frenck:
### What happened?

As per documentation, these are the light capabilities of WLED segments:

![image](https://user-images.githubusercontent.com/195327/209787918-8c13a186-b1c0-49f3-b4c4-f8f27779afaf.png)

Source: https://kno.wled.ge/interfaces/json-api/#light-capabilities

However, that is untrue or a bug. I've got several ways to get `lc` values returned that exceed the documented bits, suggesting there is an additional bit in use that is currently marked as reserved/unused.

### To Reproduce Bug

Multiple reproductions:

- Configure a WLED device with a single segment "PWM CCT"; the `lc` value communicated is `14`, which exceeds the current documented set.

- Config a WLED device with a single segment "PWM RGB + CCT"; the `lc` value communicated is `15`, which exceeds the current documented set.

### Expected Behavior

To either not exceed the documented values or documentation that correctly provide information on the light capabilities.

### Install Method

Binary from WLED.me

### What version of WLED?

0.14.0-b1

### Which microcontroller/board are you seeing the problem on?

ESP8266

### Anything else?

I've currently assumed there is another bit involved and adjusted my library yesterday accordingly:

See: <https://github.com/frenck/python-wled/pull/876>

That said: I don't like to assume any of this. Hence the creation of this issue.

Ran into this while working on: <https://github.com/home-assistant/core/pull/75248>

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ezcGman:
For some context: The line in question is this where the fourth/undocumented bit also gets set: https://github.com/Aircoookie/WLED/blob/main/wled00/FX_fcn.cpp#L754

blazoncek:
The docs say "reserved (expect any value)".
Use only the lower 3 bits.

Let's say (for now) that the upper nibble controls UI elements only.

frenck:
> Use only the lower 3 bits.

That turned out to be not valid. For example, an RGB+CCT with auto white, will provide the white bit, but since it is auto, that isn't applicable (which seems to be handled by the undocumented bit?) In the above case, the primary color is a tuple of 3 (R,G,B) and not 4 (R,G,B,W).

So it seems to have significance, and the WLED UI seems to care about it as well.

> Let's say (for now) that the upper nibble controls UI elements only.

It also changes the data send over the line.

blazoncek:
The relevant code:
```c++
[...]

PR #3073, opened 2023-02-02
Support white addressable LED strips

Aircoookie:
Adds support for white-only addressable LED strips with 3 white LEDs per WS2811, like these:
https://de.aliexpress.com/item/1005004771896342.html
https://de.aliexpress.com/item/1005004092153344.html

Still draft, as code style can be improved and I also would like to add support for addressable CCT:
https://de.aliexpress.com/item/1005005004675542.html


blazoncek:
I am in the process of adding UCS890x LEDs (16bit).
I have 8903 running but 8904 is giving me headaches as apparently something doesn't do bounds check and I see heap corruption.

Would you care to take a look? I'll publish UCS890x branch.

blazoncek:
UCS890x has been solved by NeoPixelBrightnessBus patch.
See https://github.com/Makuna/NeoPixelBus/issues/638

MichaelBMF:
Hi i'm new to WLED and thx for that great project!

This PR will add support for IC controlled CCT strips. But as far as i understand, this COB CCT addressable strip will still be not supported: 
https://www.btf-lighting.com/collections/fcob-led-strip/products/fcob-cct-ic-led-light-strip-addressable-576-leds-2700k-to-6500k-10mm-dc24v-flexible-high-density-cob-ra90. 

Perhaps on the weekend i have time for some coding. The part with new types and capabilities i  think i already understand. There is only need for an WS28xx type that only needs two channel for cw and ww i think.

Aircoookie:
@MichaelBMF seems like these use 1 WS2811 for each zone, which is easier code wise than the 2 WS2811 for 3 zones arrangement. Will add support for these too (in a separate PR)

blazoncek:
It seems there is a bug regarding RGBW strips.
```
{
	"seglc": [
		1,
		3,
		6
	]
}
```
is reported for 3 segments. Segment 1 contains WS2812 & SK6812 strips (spanning 2 GPIO), segment 2 only has SK6812 and segment 3 only has PWM White CCT. Segment 1 is a 2D segment (20x20). Bus 1 has 396 LEDs, bus 2 has 25 LEDs (SK6812, none for AW), bus 3 is PWM CCT (none for AW).

Aircoookie:
> It seems there is a bug regarding RGBW strips.

Good catch! The second (RGB + white, lc 3) and third segment (white + cct, lc 6) here are as expected. However the first incorrectly reports RGB only.
This is not an issue about different strip types, but about `Segment::refreshLightCapabilities()` not taking into account 2D, it just checks start/stop (or X coordinates in 2D), so it will consiger segment 0 to be only the first 20 LEDs.
Mixing strip types within the same matrix is probably a niche use case, still capabilities should properly take 2D into account of course.

blazoncek:
One more note: `Segment::refreshLightCapabilities()` does not take into account any mapping that has been done.

blazoncek:
Yet another note: 2D segments may have "gaps" (see led-gap branch/PR) which may mask unused pixels (-1).

blazoncek:
I have a fix that seems to work correctly for 1D, 2D and mix of 2D and 1D segments.
Ledmaps included.
Will be available in led-gaps branch.

blazoncek:
This also resolves #2983

snenyl:
I got white-only addressable LED strips with 3 white LEDs per WS2811 how do I enable this so the mapping is correct?

PR #3085, opened 2023-02-08
allow alternative northern style ("viertel vor ..." instead of "dreiviertel ..") in wordclock v2

f-zappa:


Aircoookie:
I live in NRW and if asked, I would have probably deemed "Es ist dreiviertel vier" to be grammatically incorrect :D Thank you for the extra option.

mbruehl-ger:
There is a mixup at "viertel vor"  (288)    

 if (nord) {
              setMinutes(9); <- 13
            } 
              else {
              setMinutes(12);    <- 9 
              
              
              
 Also - it seems like the meander wiring is obsolete as the new 2dMatrix settings seem to work for the regular settings....

Hasenpups:
The current implementation is wrong now. I will create a pull request tomorrow to fix it. 
Wife is angry now because the display is wrong 🙁 after an update

PR #3090, opened 2023-02-11
Add Panel Visualization

ewoudwijma:
Copied from MoonModules repo


blazoncek:
You should've used separate branch to allow me to do modifications prior to merging.

ewoudwijma:
Okay, will do that 1 sec

blazoncek:
If you use a branch other than main, I can pull locally that branch, make modifications and push back to your repo.
I can do that with main too, but it then screws up my local repo.

PR #3091, opened 2023-02-12
Add Ucs890x support and swaps NeoPixelBrightnessBus with NeoPixelBusLg

blazoncek:
This will require extensive testing on ESP8266 as NeoPixelBus v2.7.1+ will no longer compile correctly with current PlatformIO packages (framework & toolchain) for ESP8266. I recommend using latest from PlatformIO repository unless there are other reasons not to.

Even though UCS890x support was added to NPB v2.6.9 it contained a bug preventing 64bit RGBW UCS8904 from working properly. The fix was included since NPB v2.7.3.

NeoPixelBusLg **will return unscaled** RGB values in GetPixelColor() causing some effects to potentially misbehave.

Even though NeoPixelBusLg contains Gamma correction it is currently fixed at gamma 2 (fast one) which is different than gamma used by WLED, unless we use *on the fly* gamma correction which does it using math (slow) instead of LUT.


Makuna:
Just a note/history on a statement above:
"NeoPixelBusLg will return unscaled RGB values in GetPixelColor() causing some effects to potentially misbehave."
Due to quantization, often GetPixelColor on NeoBrightnessBus would not return the same color you just called SetPixelColor, this was how the math worked and caused lots of hidden issues (and support from me when people don't understand).  If you truly need something like GetPixelColor, then you should be using an external buffer that doesn't modify the values with either brightness nor gamma and then render using a shader with brightness and gamma as part of its math.  I almost removed GetPixelColor from NeoPixelBusLg.

blazoncek:
@Makuna thanks for taking your time. We implemented double buffering a while ago (user selectable). It is implemented either global or per segment basis (segment being part of a strip).
I made the above comment after quickly skimming through NeoPixelBusLg code. We already determined that the statement is false and NPBLg will return same scaled-then-descaled color information as NeoPixelBrightnessBus did (with potential quantization loss).

wled00/bus_wrapper.h
```diff
@@ -612,6 +767,22 @@ class PolyBus {
...
+      case I_32_I1_UCS_3: (static_cast<B_32_I1_UCS_3*>(busPtr))->SetPixelColor(pix, Rgb48Color(col.R,col.G,col.B)); break;
+      #endif
+//      case I_32_BB_UCS_3: (static_cast<B_32_BB_UCS_3*>(busPtr))->SetPixelColor(pix, Rgb48Color(col.R,col.G,col.B)); break;
+      case I_32_RN_UCS_4: (static_cast<B_32_RN_UCS_4*>(busPtr))->SetPixelColor(pix, Rgbw64Color(col.R,col.G,col.B,col.W)); break;
+      #ifndef WLED_NO_I2S0_PIXELBUS
+      case I_32_I0_UCS_4: (static_cast<B_32_I0_UCS_4*>(busPtr))->SetPixelColor(pix, Rgbw64Color(col.R,col.G,col.B,col.W)); break;
+      #endif
+      #ifndef WLED_NO_I2S1_PIXELBUS
+      case I_32_I1_UCS_4: (static_cast<B_32_I1_UCS_4*>(busPtr))->SetPixelColor(pix, Rgbw64Color(col.R,col.G,col.B,col.W)); break;
+      #endif
```
climoge, line 784: Hi. This code block doesn't work because `Rgbw64Color` has 16 bits parameters but `col` has 8 bits members, so the LED strip is always black.

If you want to use the dynamic conversion from 8 bits to 16 bits, `Rgbw64Color` has a constructor with a `RgbwColor`, so the working code is :

```cpp
case I_32_RN_UCS_4: (static_cast<B_32_RN_UCS_4*>(busPtr))->SetPixelColor(pix, Rgbw64Color(col)); break;
#ifndef WLED_NO_I2S0_PIXELBUS
case I_32_I0_UCS_4: (static_cast<B_32_I0_UCS_4*>(busPtr))->SetPixelColor(pix, Rgbw64Color(col)); break;
#endif
#ifndef WLED_NO_I2S1_PIXELBUS
case I_32_I1_UCS_4: (static_cast<B_32_I1_UCS_4*>(busPtr))->SetPixelColor(pix, Rgbw64Color(col)); break;
#endif
```

blazoncek: It does work on my system.

climoge: @blazoncek I don't know what is your system, but I can assure you that my LED strip is always black without my modification.
Thanks Makuna, I'm a newbie in WLED and NeoPixelBus but I tested extensively your library, enough to understand the problem of the 8bits -> 16bits conversion.

blazoncek:
@differentunic @climoge can you test UCS8904 (and possibly UCS8903) if the LEDs perform properly?
The branch has been updated to latest main code but I did not change Rgbw64Color behaviour as @climoge suggested.

I will need to dig in my garage to find UCS8904 LEDs @differentunic sent.

climoge:
Sorry, i'm away from home until April 19. I'll check your merge next week.

Differentunic:
Tested with the factory setting, and I am unable to get any of the LEDs to light up with either the USC8904 (RGBW) or the UCS8903 LED type selected. When any other LED type is selected they turn on, and respond to changes.

blazoncek:
@climoge @Differentunic I've now replaced explicit color conversion into NPB implicit conversions.
It should be cleaner and work correctly.

@Makuna perhaps adding a few more color conversion constructors would be nice (instead of having `Rgb48Color(RgbColor(col))` where `col` is `RgbwColor` object.)

Differentunic:
>  I've now replaced explicit color conversion into NPB implicit conversions. It should be cleaner and work correctly.

Seems to have fixed it. UCS8904 (RGBW) Is showing the correct colors. The last thing is to add the white slider on the colors tab.
I also don't currently have any UCS8903 leds, but i assume they behave correctly, as they are just UCS8904 without a white channel.

Thanks a bunch @blazoncek!

image
image

Makuna:
> @climoge @Differentunic I've now replaced explicit color conversion into NPB implicit conversions. It should be cleaner and work correctly.
> 
> @Makuna perhaps adding a few more color conversion constructors would be nice (instead of having `Rgb48Color(RgbColor(col))` where `col` is `RgbwColor` object.)

Yes, that was missed, adding shortly.

blazoncek:
@climoge can you confirm the recent update is working so I can finally merge (or make adjustments)?

PR #3105, opened 2023-02-24
Homekit integration

imcmib:


PDBeal:
HK_SETUP_ID, HK_PAIRING_CODE, HK_CATEGORY, and HK_FLAG are all undefined with no example configuration to know where or what datatype  or value to define these parameters

PR #3106, opened 2023-02-25
Usermod: Klipper percentage 

lost-hope:
I made a usermod, that adds an overlay that corresponds with the printing percentage reported by the Klipper API.


blazoncek:
I do not use Klipper or know how it works but the code seems straightforward and does not break existing stuff.
Apart from some minor cosmetic issues I agree with the merge.

lost-hope:
> Apart from some minor cosmetic issues I agree with the merge.

could you point them out, so i can get better in the future?

blazoncek:
It is just a few tabs vs. spaces issue.

Issue #2984, opened 2022-12-28
"White" channel only segments confusingly provide 2 or 3 sliders 


frenck:
**Is your feature request related to a problem? Please describe.**

Currently, when configuring a device with a single segment with a single white channel, creates a confusing UI to control and a confusing API to work with.

In the end, the case described above is basically a dimmable LED strip (single color, assumingly just "white").

The point is, WLED will give you 3 sliders for controlling this LED strip, providing the same functionality, but are fully decoupled:

- The master slider for the brightness.
- The brightness slider for the segment.
- The white value slider for the segment.

All of these do the same thing in this case: Controlling the brightness; which makes no sense and is (in my opinion) highly confusing.

**Describe the solution you'd like**

The master brightness makes sense in a way, as in, that is part of how WLED works in general (although, still confusing in a single-segment setup).

I would suggest removing the slider for the white values in the case as described above and only providing a brightness slider for the segment (omitting the white value).

**Describe alternatives you've considered**

I've not, really. For external control, I've considered sending the same values for brightness and white value (so external control is at least logical).

**Additional context**

![CleanShot 2022-12-28 at 10 25 24](https://user-images.githubusercontent.com/195327/209789843-56d78cfb-9200-4ed4-a220-3a473fd9753c.png)

Ran into while working on: <https://github.com/home-assistant/core/pull/75248>

Thank you for your ideas for making WLED better!

Aircoookie:
I agree that having 3 sliders for apparently the same purpose in typical use cases (1 segment, solid effect) is confusing.

However, keep in mind that even "White only" segments can take advantage of segment color slots (consider the Blink effect blinking between maximum brightness and a dimmer background color value)

ezcGman:
So this is solved now? :)

Trying to get everything closed and solved to finally get this added to HA :)
https://github.com/home-assistant/core/pull/75248

To finally have CCT support for WLED in HA :)

Thanks!

frenck:
This is not blocking the referenced PR in Home Assistant Core.

blazoncek:
The core of this issue is not "fixed" and probably will never be.
But.
I have added UI logic to hide one of the sliders (segment opacity/brightness) if only one segment is used. So the number of possible options to change brightness is reduced from 3 to 2 sliders (one being global brightness the other being "color").

ezcGman:
Yeah, I'm with you. I like the improvement for this now! Thx!

PR #3107, opened 2023-02-25
Tweaks & bugfixes.

blazoncek:
One pixel segment handling.
- added 0D FX metadata (1 pixel effects)
- ignore color for non-RGB & non-White segment (on/off)

Bugfix
- proper auto segment creation
- hide palettes for non RGB segments
- some tweaks for #2984
- force Solid for some FX (causing crash) on 1 pixel segment

UI Optimisations
- slider tooltips
- tiny bit smaller tooltips
- hide segment power if only one segment
- gap between sliders


wled00/json.cpp
```diff
@@ -125,36 +125,43 @@ void deserializeSegment(JsonObject elem, byte it, byte presetId)
...
-
-      if (!colValid) continue;
+        if (!colValid) continue;
 
-      seg.setColor(i, RGBW32(rgbw[0],rgbw[1],rgbw[2],rgbw[3]));
-      if (seg.mode == FX_MODE_STATIC) strip.trigger(); //instant refresh
+        seg.setColor(i, RGBW32(rgbw[0],rgbw[1],rgbw[2],rgbw[3]));
+        if (seg.mode == FX_MODE_STATIC) strip.trigger(); //instant refresh
+      }
+    } else {
```
Aircoookie, line 161: IMO treating any non-black color as white on the FX/bus level and hiding color controls in UI is enough in this case, as an api user setting for example #FFAA0000 as the color and it resulting in #FFFFFFFF could be considered a bug. Additionally this might cause problems when trying to change colors on segments that are not currently active (as they won't intersect with any bus and thus have no RGB/white capability). Same case for palettes.

Other than that, approving all code changes, really like the 0D filter implementation! I will still need to do a little bit more 8266 testing as it seems to bootloop...

blazoncek: It should only set color to 0xFFFFFFFF for segment with no capabilities i.e. On/Off segment (the one containing On/Off buses only) and leave other segments intact. On/Off segments as such should be considered an exception IMO. If you insist on keeping color as set by API and dealing with it on bus level you may achieve an awkward situation when setting black color (and getting feedback that it is black). Which I am not very fond of.

To explain a bit:
- set primary color to 0 (black), secondary to white (0xFFFFFFFF) with the intention of inverting the output
- use blink effect to alternate between primary and secondary colors (should work as expected)
- switch to solid effect (two possible outcomes)
  a) the output is off as only primary color is used (but this is expected as we want to invert output)
  b) the output is on (if 0xFFFFFFFF is forced on bus level, not expected)
- now hit on/off segment power button
  a) the output will remain off (it is not inverted as intended)
  b) the output will be off
- same observation can be seen for master power

If you do not specify array and/or segment ID then only selected segments are chosen (ignoring inactive ones?). If selected segment has RGB or White the color is processed normally.

Regarding palettes: the code in Segment::color_from_palette() already ignores palettes if segment does not have RGB, even if palette is chosen, it only uses selected color.
Regarding segments that are not active (stop==0): those are often (if not always) recreated using presets which will most likely include all other segment options. If you delete a segment, apply modifications to all segments (including inactive, though I do not know how) and then re-create a segment, it will only require modifying color and/or palette (a minor nuisance).

Still, I will check other parts of the code if this can be brought in-line to make behaviour consistent.

Aircoookie: Perhaps the easiest would be to completely ignore the API color value, and just have Primary always be ON and Secondary OFF. That way, setting primary to black would have no effect at all. 

Thank you for the changes! I think we can also remove the palette check on the API level because as you said it is already ignored anyways on the segment level.

A stable 8266 0.14 version seems to be our biggest obstacle currently - even in b1 I already got a lot of reboots when interacting with the UI and now it seems to have reached a state of full unusability - either a reboot on loading one of the UI resources (seems to get caught up on either palette or FX data most of the time) or flat out complete bootlooping. It's not a problem with a specific piece of code but seems to just be general resource exhaustion (perhaps too much dynamic memory allocation/fragmentation?)
Maybe we'll be forced to change strategy and release b2 for ESP32 only - I'd hate to do that but of course we can't let 8266 compatibility make progress come to a standstill. I will expect that getting rid of most of set.cpp/xml.cpp will improve stability, so want to get started on that right once the next release is made.

blazoncek: > Perhaps the easiest would be to completely ignore the API color value, and just have Primary always be ON and Secondary OFF. That way, setting primary to black would have no effect at all.

I looked at this possibility but could not come up with a practical solution (without too many `if`s). This being due to no distinction between primary and secondary color below segment/FX level. (Segment::setPixelColor() or anything below it is not the correct place - technically or logically.)

I am intrigued by instability of your ESP8266. I have 1 in production use (on my roof borderline) with PIR, Temperature and Multi relay usermods and it still works normally though free heap plunges to a worrisome 12k after some time. It has 480 pixels attached. On a test unit with same usermods plus 4LD usermod the heap gets a bit lower still (256 pixels) but still works flawlessly. It does require reboot prior to OTA as it seems that 10k free heap is not enough to do OTA (safety code reconnects WiFi in such case). I changed that to determine low memory condition not at 8k but 6k if that will make it any better.
I even test on ESP01 an no reboots or bootloops detected there as well. I do not do a rigorous testing there though.

blazoncek: An observation regarding ESP8266:
If you disable websockets (yes, painful) then it has a few kB more of free heap.
In this case I can see heap corruption within AsyncWebServer as a response from `{"ps":2,...}` shows buffer overrun condition and instead of state response shows state + some grabbled output (including HTTP response).
When using websockets there is often a WS error, though I cannot pinpoint what it is, with occasional message "WS buffer allocation failed.".

blazoncek: Played a bit more with 8266.
It seems that AsyncWebServer is not releasing memory as quickly as it could (and as thoroughly as it should). So unless you have <7 segments (with above mentioned usermods which produce ESP with about 28k of free heap at boot) you will encounter out of memory condition soon.

I have tried to update platform to espressif8255@4.1.0 (to use newer compiler and packages) with mixed outcome. Basically everything works and I can see memory manager panicks in OOM cases. It does not help with low memory condition though. I also had to remove IRAM_ATTR from some functions (2D) as existing code no longer fit into IRAM.

I have checked my old notes and free heap at boot used to be around 32k so most recent 0.14.0-b2 reduces free heap by 4k. I did not yet try to compile with options disabled as I want to get existing setup to show increase in free heap.

Aircoookie:
Using the latest commit and `espressif8266@4.1.0`, stability seems improved to about the same degree as main (still frequent resets for me, but the UI loads reliably again and no bootloops even with all features enabled), so I can finally approve :)

Aircoookie (approved):

blazoncek:
Were you able to determine reset cause? What I observed when memory got very low weren't resets, but constant WiFi disconnects which made WLED device unreachable until cold boot (heap was not released!).

When I disabled websockets, the reliability improved enormously as (boot) free heap increased and there were no memory leaks (or hogs) from AsyncWebServer.
The bottom line of my tests is that if ESP8266 starts with a free heap below 30k (the first heap printout) with websockets enabled it will struggle with operation long term if you intend to have more than 8 (or so) segments.

Issue #2998, opened 2023-01-04
Usermod "Battery" features


itCarl:
First of all this isn't a real Bug neither a FR. sorry. I just want to get in touch with people that want to improve the Usermod.

I have seen that recently many good improvements have been added to the usermod. In fact, I am aware of the "bugs"/missing features.
That's why I started to write a software to simulate a battery (battery packs will be supported later).

The Software is still far from finished but a start has been made. The main goal, however, is to use the software to better understand and analyze batteries in order to significantly improve the usermod.
I think the software becomes very interesting especially when you have varying currents and you want to look at the runtime when a certain wled effect is activated. In my head I also have such features as "SOH, State Of Health", "Battery types" or "Smart battery calibration". Knowing these parameters from a battery for just a usermod might be a bit too much though.

Some Screenshots
---
![batteryDashboard01](https://user-images.githubusercontent.com/45098772/210637730-22d8e2c8-80c7-4db9-a5fd-40fb88d30d7b.PNG)
![batteryDashboard02](https://user-images.githubusercontent.com/45098772/210637734-59e0e409-a42a-40c0-9c9e-5a9c176a9152.PNG)

This Software is still kinda basic. But can fully calculate parameters for static loads.

ℹ️ Sources
- [**S**tate **O**f **C**harge (SOC) calculation, Video by Stafl Systems](https://www.youtube.com/watch?v=rOwcxFErcvQ)
- [SOC estimation methods, Video by Owl WiS](https://www.youtube.com/watch?v=7DTzShuFN6M)
- [**S**tate **O**f **H**ealth (SOH), Video by Stafl Systems](https://www.youtube.com/watch?v=imjbyeS1caI)
- [SOC Wikipedia](https://en.wikipedia.org/wiki/State_of_charge)

PS: I would have really liked to put this in the Repo dicussion section. But there isn't a dicussion section.

softhack007:
Hi @itCarl, I just added some initial support for 
LiPo batteries, as I have a device with built-in LiPo connector/charger.

I found two interesting things:
* as ESP32 cannot measure above ~2.9V, these devices all use a small "voltage devider" circuit (basically two resistors of the same value). So measured voltage must be multiplied by 2.
* LiPo cells seems to have a different  voltage / capacity curve that is non-linear at its "edges",  see https://blog.ampow.com/amp/lipo-voltage-chart/

I'm happy with my results now, and I think it will also work on other esp32 with build-in LiPo connector.

[...]

PR #3116, opened 2023-03-08
Changing voltage calculation to a weighted running average

Erwin-Repolust:
This helps improve precision in the readout and makes it more stable.
The calibration will now not be multiplied by 2 and therefore reflects the change in volts.
Removed check for "invalid voltages" as that just makes it hard for people to check whats going wrong in their setup when they only get "invalid voltage" in the web gui.

### DIFF
--- a/usermods/Battery/battery_defaults.h
+++ b/usermods/Battery/battery_defaults.h
@@ -26,6 +26,15 @@
   #endif
 #endif
 
+//the default ratio for the voltage divider
+#ifndef USERMOD_BATTERY_VOLTAGE_MULTIPLIER
+  #ifdef ARDUINO_ARCH_ESP32
+    #define USERMOD_BATTERY_VOLTAGE_MULTIPLIER 2.0f
+  #else //ESP8266 boards
+    #define USERMOD_BATTERY_VOLTAGE_MULTIPLIER 4.2f
+  #endif
+#endif
+
 #ifndef USERMOD_BATTERY_MAX_VOLTAGE
   #define USERMOD_BATTERY_MAX_VOLTAGE 4.2f
 #endif
--- a/usermods/Battery/usermod_v2_Battery.h
+++ b/usermods/Battery/usermod_v2_Battery.h
@@ -114,6 +114,17 @@ class UsermodBattery : public Usermod
       }      
     }
 
+    float readVoltage()
+    {
+      #ifdef ARDUINO_ARCH_ESP32
+        // use calibrated millivolts analogread on esp32 (150 mV ~ 2450 mV default attentuation) and divide by 1000 to get from milivolts to volts and multiply by voltage multiplier and apply calibration value
+        return (analogReadMilliVolts(batteryPin) / 1000.0f) * voltageMultiplier  + calibration;
+      #else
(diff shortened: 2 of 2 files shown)


blazoncek:
@itCarl as you were the last (I think) to update battery usermod, can you verify the code?

itCarl:
> @itCarl as you were the last (I think) to update battery usermod, can you verify the code?

Actually i wasn't the last one. Frank aka @softhack007 is the one you should ask.
In my opinion adding smoothing to the voltage makes a lot of sense as @softhack007 and i already discussed here -> #2998.

However, I would suggest that a smoothing calculation should not only be there when working with an esp32 but also in the general voltage calculation formula.

Erwin-Repolust:
> > @itCarl as you were the last (I think) to update battery usermod, can you verify the code?
> 
> Actually i wasn't the last one. Frank aka @softhack007 is the one you should ask. In my opinion adding smoothing to the voltage makes a lot of sense as @softhack007 and i already discussed here -> #2998.
> 
> However, I would suggest that a smoothing calculation should not only be there when working with an esp32 but also in the general voltage calculation formula.

Added code and smoothing for the esp8266. The existing code for it was not working properly anyway I believe.
Also there is a new variable now that lets u adjust for different voltage dividers in case u wanna measure something over ~5v.
Changed the filter code to the one mentioned in #2998.
I have not tested the bit for the esp8266 as I don't have access to the hardware but it should work.

3v is used cause of this graph:
image

itCarl:
Hey 👋,
i just compiled your code Erwin and something is not quite right with the esp8266 code.
I mainly use D1mini's for testing and development and those have a voltage divider built-in.
Wemos D1mini v3.1.0
[Datasheet](https://www.wemos.cc/en/latest/d1/d1_mini_3.1.0.html) -- [Schematic](https://www.wemos.cc/en/latest/_static/files/sch_d1_mini_v3.0.0.pdf)

Wemos D1mini v4.0.0
[Datasheet](https://www.wemos.cc/en/latest/d1/d1_mini.html) -- [Schematic](https://www.wemos.cc/en/latest/_static/files/sch_d1_mini_v4.0.0.pdf)

R1 220k and R2 100k. To get a 'nice' max. of 1v from a 4.2v source we can add a 100k to A0. So that R1 becames 220k + 100k = 320k. When we now apply the voltage divider formula = Vout = (Vin * R2) / (R2 + R1). The result will be something like (4.2v (in) * 100k) / (100k + 320k) = 1v (out). So this is why the current formula multiplies just by maxBatteryVoltage.
Thats why we don't really need the resistor ratio in the voltage calculation formula for the esp8266 inside the usermod. Unless someone is actually using the bare esp8266 chip.
Anyway.
When using the new formula the voltageMultiplyer would be around 1.4 (=4.2v/3.0v) for the resistor setup im currently using.

If i messed something up or understood something wrong. please let me know.

Erwin-Repolust:
Ahh thanks for the explanation and for testing, as I said I have no access to the esp8266 right now and from what I could read online it all depends on the specific board you own. There might be no voltage divider at all ~0-1v or one that gives roughly ~0-3.3v and then there is your case where u add a resistor to get ~0-4.2v.
So we just set defaults for both platforms then.
I exposed the new voltage multiplier in the gui. It does help people who use different voltage divider setups in the end.

Aircoookie (approved):

PR #3118, opened 2023-03-08
Improve indent

Mattstir:
Improve indent, so its more aligned and correctly indented according to logic groups.

No functional change.

### DIFF
--- a/wled00/data/404.htm
+++ b/wled00/data/404.htm
@@ -1,47 +1,47 @@
 <!DOCTYPE html>
 <html>
-    <head>
-        <meta charset="utf-8">
-        <meta content='width=device-width' name='viewport'>
-        <meta name="theme-color" content="#222222">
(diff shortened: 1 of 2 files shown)


Mattstir:
As this is my first PR for WLED wanted to start with something small ;) 
Do I also need to push the updated built `html_*.h` files? I'm asking because they changed a lot but IMO shouldn't just because of white space changes...
thx ;)

Mattstir:
My guess as there is no feedback: There is no interesst in this simple improvement?

Aircoookie:
Thank you very much for the contribution! :)
Lack of feedback is just due to limited time spent on WLED on my part in the recent past, it does not mean there is no interest, just that I hadn't taken a look yet.

Mattstir:
Thanks for the feedback and the merge ;) I will consider more contributions in the future 🙌

PR #3121, opened 2023-03-10
Art-Net transmit support for network LEDs

troyhacks:
Like DDP, this allows WLED to address network systems using the Art-Net protocol.

Universe starts at zero, because that's the first universe in Art-Net.

Works with RGB. It's coded to also work with RGBW, but I couldn't find a great place to enable it without mucking with things I don't understand.

### DIFF
--- a/wled00/bus_manager.cpp
+++ b/wled00/bus_manager.cpp
@@ -383,24 +383,20 @@ uint8_t BusOnOff::getPins(uint8_t* pinArray) {
 
 BusNetwork::BusNetwork(BusConfig &bc) : Bus(bc.type, bc.start, bc.autoWhite) {
   _valid = false;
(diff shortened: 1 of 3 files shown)


blazoncek:
Looks clean to me.
The only drawback I can see is memory footprint for ESP8266. Could you move static constants into PROGMEM?
It does incur additional step of reading from PROGMEM but will reduce heap usage on ESP8266 which is already short on RAM.

dosipod:
@troyhacks  This seems to no longer work ( I recall it worked there was chang somewhere to "fix it" ) 
At the moment you can not use art-net transmit from wled to wled on AC or MM , I can still see some packets on xlights capture but seems busted  . Tried multiple versions  upto 15 but the same case . DDP is okay 
image

Issue #3132, opened 2023-03-15
DMX_MODE_PRESET lags out the ESP32


strikeout:
### What happened?

Im using an external hardware DMX Controller to send ArtNet (or e1.31) packets to WLED to select various presets.
The hardware controller sends its values aprox 2 times per second. This  completely overwhelms the ESP32 - especially when the UI is open. You can see the same preset being selected again and again, lagging out the controller - until the WS connection drops completly.

### To Reproduce Bug

Use DMX_MODE_PRESET, send repeated message with the same channel/value, see the system lag to death.

### Expected Behavior

I expect DMX messages with the same channel/value pair to be dropped and not be re-applied.

### Install Method

Binary from WLED.me

### What version of WLED?

v14.0.0-b1

### Which microcontroller/board are you seeing the problem on?

ESP32

### Anything else?

Would a tiny buffer to save the last DMX message for comparison to current received message help? 
Then do a simple `if (lastDmx == currentDmx) return;`

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

blazoncek:
Unfortunately there is no logic (or state history) within WLED to support such behaviour.
Please update settings on your DMX equipment as MCU is already on the verge of capacity handling all possible uses.

If, on the other hand, you know how to solve the issue then please make a PR.

blazoncek:
Check JSON API state & info responses. "ps" and "pl".
One is for playlist the other for preset.

strikeout:
Is my assumption that the global `currentPreset` is populated and usable in the following form correct?

```
case DMX_MODE_PRESET:       // 2 channel: [Dimmer,Preset]
      ...
      if (currentPreset != e131_data[dataOffset+1]) {
        applyPreset(e131_data[dataOffset+1], CALL_MODE_NOTIFICATION);
      }
      ...
```

mxklb:
Thanks for testing. Preset mode is new and wasn't tested that much :smile: 

``` 
    case DMX_MODE_PRESET:       // 2 channel: [Dimmer,Preset]
      if (uni != e131Universe || availDMXLen < 2) return;
      if (currentPreset() != e131_data[dataOffset+1]) {
        applyPreset(e131_data[dataOffset+1], CALL_MODE_NOTIFICATION);
      }
      if (bri != e131_data[dataOffset]) {
        bri = e131_data[dataOffset];
        strip.setBrightness(bri, true);
      }
      return;
      break;
```
[...]

PR #3134, opened 2023-03-16
Fixes DMX_MODE_PRESET preset and brightness selection via DMX controller

strikeout:
No more overloading of the ESP when using analogue controllers sending many packets (often the same) per second to WLED. Packets selecting the _same_ preset are discarded now. 

Additional fixes:
- brightness selection slider in UI no longer flickers when preset value != dmx value
- makes playlists work together with presets as well now, allowing for a smooth selection between the two 

/ #3132

### DIFF
--- a/wled00/e131.cpp
+++ b/wled00/e131.cpp
@@ -174,10 +174,20 @@ void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol){
 
     case DMX_MODE_PRESET:       // 2 channel: [Dimmer,Preset]
       if (uni != e131Universe || availDMXLen < 2) return;
-      applyPreset(e131_data[dataOffset+1], CALL_MODE_NOTIFICATION);
-      if (bri != e131_data[dataOffset]) {
+
+      // only apply preset if value changed 
+      if (currentPreset != e131_data[dataOffset+1] && e131_data[dataOffset+1] != 0 && 
+          // only apply preset if not in playlist, or playlist changed
+          (currentPlaylist < 0 || currentPlaylist != e131_data[dataOffset+1])) { 
+        presetCycCurr = e131_data[dataOffset+1];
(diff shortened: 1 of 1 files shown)


wled00/e131.cpp
```diff
@@ -174,10 +174,20 @@ void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol){
...
+        presetCycCurr = e131_data[dataOffset+1];
+        unloadPlaylist(); // applying a preset unloads the playlist
+        applyPreset(e131_data[dataOffset+1], CALL_MODE_NOTIFICATION);
+      }
+      // only change brightness if value changed
+      if (bri != e131_data[dataOffset]) {                                        
         bri = e131_data[dataOffset];
-        strip.setBrightness(bri, true);
+        strip.setBrightness(scaledBri(bri), false);
+        stateUpdated(CALL_MODE_WS_SEND);
```
blazoncek, line 190: Is this strictly necessary?
As `applyPreset()` is asynchronous in 0.14 and beyond it will trigger "CALL_MODE_NOTIFICATION" when the preset is actually applied (which is *after* `strip.setBrighntess()` call).

blazoncek (approved):
Not sure about added `stateUpdated()` call, but otherwise approved.

wled00/e131.cpp
strikeout: Without the call, the brightness slider in the UI is _never_ updated. 
With this line tho, the UI slider updates async (without lagging the ESP to death), and then the slider represents the actual brightness state as per DMX values. 
The only thing I'm not sure of, is the `strip.setBrightness` direct _true/false_ bool - I set it to false to adhere to the refresh cycles instead of writing directly to the busses - so there might be a few ms delay but that seemed ok in my tests, and it felt more stable.

mxklb:
Looking good :+1:

PR #3142, opened 2023-03-19
Set Dancing Shadows default palette to Party

xxv:
As a default, the Party palette looks much better than a solid color.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -4085,7 +4085,7 @@ uint16_t mode_dancing_shadows(void)
       spotlights[i].type = random8(SPOT_TYPES_COUNT);
     }
 
-    uint32_t color = SEGMENT.color_from_palette(spotlights[i].colorIdx, false, false, 0);
+    uint32_t color = SEGMENT.color_from_palette(spotlights[i].colorIdx, false, false, 255);
     int start = spotlights[i].position;
 
     if (spotlights[i].width <= 1) {


blazoncek:
Hi. This may be true to some users and not to others.
As this changes existing behaviour without proper justification it may not get merged into WLED.

xxv:
The only justification I can offer is that I'm the original Dancing Shadows pattern author and I made a mistake when I first submitted it. It should have been Party by default.

blazoncek:
Would you care then to look at the other default values (located in the effect name string)?
This was the change introduced in 0.14.

xxv:
You mean the `_data_FX` values? They look good, though if you're looking to set some defaults there, I think the "Pixel" mode for "expand 1D Fx" in the 2D mode looks better than the others, e.g. `m12=0`  Are you suggesting to use `pal=11` instead of changing this value?

The only other thing for "Dancing Shadows" that I'd change is the text "# of shadows" to "# of spotlights" (the number of shadows cast would depend on the number of objects you have, not light sources), but it's not a big deal to me.

If you'd like I can make those changes too while this is being updated.

blazoncek:
Yes, please do so. All of it, as I was unable to figure out how this effect was supposed to look. 😁

xxv:
It looks particularly good when in two configurations (party/rainbow colors set on both of them): 

1) used with long linear strips (I use 20m long addressable fairy lights, spaced 10/m) density set to high, speed medium/low. it looks like things moving along the strands.
2) used with very dense strips (144/m, 60/m) as a light source with objects set in front of them, shining onto a wall/ceiling. This configuration was the original idea, as it casts moving shadows that are both analog and digital at the same time. 

I'll make the updates :smile:

blazoncek:
@Aircoookie shall we merge this without waiting for @xxv to add metadata?

xxv:
Y'all should go ahead. I've been busy with some life stuff and unable to get to a machine to add/test my changes (as small as they are). I'll submit a follow-up PR with the other changes when I next get a chance.

PR #3144, opened 2023-03-21
ESP8266 core 4.1.0, ESP32 core 5.2.0 (S2,S3,C3)

blazoncek:
Also updates NeoPixelBus to 2.7.3 (adding UCS890x support)


softhack007:
Hi, 
`platform = espressif32@5.2.0` seems to be a good choice for the new MCUs, works well on the devices I have.
maybe also add `platform_packages = ` for the new MCUs, so the 8266 values from [common] are not used on esp32. 

As i said in my other comment, `-Dregister= ` is absolutely dangerous, because it also affects inline assembler (macro = textual replacement) and it definitely breaks arduino core on ESP32-C3.


Will do a more detailed review later.

softhack007:
PS: You're right, these massive warnings about "register" from the fastled lib are absolutely annoying. I've move the `-Dregister=` override into the 8266 section - please feel free to un-comment it, in case you get working binaries on 8266 (could not check that tonight).

platformio.ini
```diff
@@ -113,7 +113,7 @@ build_flags =
   -D DECODE_SONY=true
   -D DECODE_SAMSUNG=true
   -D DECODE_LG=true
-  ;-Dregister= # remove warnings in C++17 due to use of deprecated register keyword by the FastLED library
+  ;-Dregister= # remove warnings in C++17 due to use of deprecated register keyword by the FastLED library ;; warning: this breaks framework code on ESP32-C3 and ESP32-S2
```
blazoncek, line 116: Enabling this did not break S2 or C3 compiles for me.

platformio.ini
```diff
@@ -247,10 +248,11 @@ lib_deps =
 [esp32c3]
 ;; generic definitions for all ESP32-C3 boards
 platform = espressif32@5.2.0
+platform_packages = 
 build_flags = -g
   -DARDUINO_ARCH_ESP32
   -DARDUINO_ARCH_ESP32C3
-  -DCONFIG_IDF_TARGET_ESP32C3
+  -DCONFIG_IDF_TARGET_ESP32C3=1
```
blazoncek, line 276: Also never had issues with this.
But there is a difference if you add a space between `-D` and `CONFIG_...`.

platformio.ini
```diff
@@ -511,10 +517,11 @@ build_flags = ${common.build_flags_esp8266} -D LEDPIN=12 -D IRPIN=-1 -D RLYPIN=2
 lib_deps = ${esp8266.lib_deps}
 
 [env:lolin_s2_mini]
-platform = {esp32s2.platform}
+platform = ${esp32s2.platform}
+platform_packages = ${esp32s2.platform_packages}
 board = lolin_s2_mini
 board_build.partitions = tools/WLED_ESP32_4MB_1MB_FS.csv
-build_unflags = ${common.build_unflags}
+build_unflags = ${common.build_unflags} -DARDUINO_USB_CDC_ON_BOOT=1
```
blazoncek, line 561: This wrecks havoc on my Lolin S2 mini board.
Has to be 0 for it to work.

softhack007: It's an "unflag". It mean that `-DARDUINO_USB_CDC_ON_BOOT=1` will be delete from build_flags. The lolin board.json in framework 5.2.0 included this flag, so this line removes it again.

platformio.ini
softhack007: True, it's a very subtle difference. I just found out recently that the official ESP-IDF way is  `-DCONFIG_IDF_TARGET_ESP32C3=1`

softhack007:
Any idea why the PlatformIO CI build on github still fails for nodemcuv2 ? It's working on my local PC.

CI build error msg:
```
Compiling .pio/build/nodemcuv2/libcde/FastLED/bitswap.cpp.o 
xtensa-lx106-elf-g++: error: unrecognized command line option '-std=gnu++17' 
 [.pio/build/nodemcuv2/libcde/FastLED/FastLED.cpp.o] Error 1 
xtensa-lx106-elf-g++: error: unrecognized command line option '-std=gnu++17' 
[.pio/build/nodemcuv2/libcde/FastLED/bitswap.cpp.o] Error 1
```

blazoncek:
@Aircoookie @softhack007 do we merge these changes?
FYI: My 8266, S2 and C3 are working without issues.

I do not know why GH actions are failing. It compiles without issues on my mac.

blazoncek:
@Aircoookie @softhack007 do we merge these changes?
FYI: My 8266, S2 and C3 are working without issues.

I do not know why GH actions are failing. It compiles without issues on my mac.

Aircoookie:
Seems like the GH actions machine uses a different xtensa toolchain as default than our local build environments. Explicitely setting the latest seems to take care of it. Approving from my side now👍

@softhack007 I would agree with your proposed changes too. If you'd like, you can add them and merge.

Aircoookie (approved):

PR #3145, opened 2023-03-22
Use constant for mDNS name to allow overriding from my_config.h

j-mutter:
I've found this useful when deploying test builds to various ESP modules to make it easier to find/load the ui on my network, so I figured others might as well.

### DIFF
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -12,6 +12,7 @@
 #define DEFAULT_AP_SSID     "WLED-AP"
 #define DEFAULT_AP_PASS     "wled1234"
 #define DEFAULT_OTA_PASS    "wledota"
+#define DEFAULT_MDNS_NAME   "x"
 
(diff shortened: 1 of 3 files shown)


blazoncek:
Hi. Thanks for the contribution but I have a question I'd like to clarify.

If I add WLED_MDNS to build flags (or my_config.h) and then flash 10 devices with the resulting binary all of them will have the same mDNS name. How will that benefit me?

blazoncek:
@j-mutter if this is still relevant please respond or the PR will be closed.

Aircoookie:
This seems useful in case you use pre-configured WiFi credentials and DHCP, to have a known mDNS address to reach the device for setup. Will go ahead and merge. Maybe this should print a warning about non-unique default mDNS name if using the binary for multiple units, but I think it's fine for now.

PR #3153, opened 2023-03-30
iPad/tablet with 1024 pixel UI PC mode.

blazoncek:
+ Optimizations.


blazoncek:
@Aircoookie @softhack007 any thoughts on this?
It does make iPad (or any tablet reporting 1024 pixels in landscape orientation) more usable without sacrificing compatibility for mobile or computer.

Aircoookie (approved):
No objections, this is a great improvement! I like that it keeps the more "spacious" original PC mode if at least 1250px width is available.

blazoncek:
@Aircoookie you may also want to look into `seg-groups` branch which adds segment grouping for quicker selection (you will need at least 3 segments to see the difference).
If you like the idea I can incorporate it into this PR or create a new one.

PR #3157, opened 2023-04-03
Add Freqpalette effect

megabug:
This sound-reactive effect uses the average value from user-selected FFT buckets to select a single color from a palette.

I couldn't find any existing SR effect that could do this - the volume-reactive ones only consider the entire signal and not a specific frequency range.

I use this on a set of shelves, where each shelf has a different FFT range that it lights up in response to. This creates a cool effect where each responds to different parts of incoming audio i.e., kick, snares, synths etc.

Bad video demo: (warning, flashing lights)

https://user-images.githubusercontent.com/15011974/229482894-710e4f1e-8718-42ce-b299-66e6bfdeef38.mp4

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -7110,6 +7110,42 @@ uint16_t mode_waterfall(void) {                   // Waterfall. By: Andrew Tulin
 static const char _data_FX_MODE_WATERFALL[] PROGMEM = "Waterfall@!,Adjust color,Select bin,Volume (min);!,!;!;1f;c2=0,m12=2,si=0"; // Circles, Beatsin
 
 
+/*
+ * Uses the average value from user-selected FFT buckets to select a single
+ * color from a palette.
+ */
+uint16_t mode_freqpalette() {
+  um_data_t *um_data;
+  if (!usermods.getUMData(&um_data, USERMOD_ID_AUDIOREACTIVE)) {
+    // add support for no audio
+    um_data = simulateSound(SEGMENT.soundSim);
+  }
+
+  // Get minimum and maximum FFT bands to consider
+  uint8_t minBand = constrain(SEGMENT.custom1 / 16, 0, 15);
+  uint8_t maxBand = constrain(SEGMENT.custom2 / 16, 0, 15);
+  if (maxBand < minBand) {
+    uint8_t temp = minBand;
+    minBand = maxBand;
+    maxBand = temp;
+  }
+
+  // Average the bands' values, scaled by user-specified sensitivity
+  uint32_t value = 0;
+  uint8_t *fftResult = (uint8_t*)um_data->u_data[2];
+  for (uint8_t band = minBand; band <= maxBand; ++band) {
+    value += fftResult[band] * SEGMENT.intensity;
+  }
+  value /= (maxBand - minBand + 1) * (256 / 8);
+  value = constrain(value, 0, 255);
+
+  SEGMENT.fill(SEGMENT.color_from_palette(value, false, false, 0));
+
+  return FRAMETIME;
+}
+static const char _data_FX_MODE_FREQPALETTE[] PROGMEM = "Freqpalette@,Sensitivity,First FFT Band,Last FFT Band;;!;1f;ix=128,c1=0,c2=255,m12=0,si=0,pal=4";
+
+
 #ifndef WLED_DISABLE_2D
 /////////////////////////
 //     ** 2D GEQ       //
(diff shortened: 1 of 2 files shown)


megabug:
> You should put epilepsy warning on above video. 😄

Good point; done.

This actually brings me to a question I had: Does the "Color Gradient" palette linearly blend between color 1 & 2?

I used a palette lookup instead of a simple `color_blend` in this new effect to keep it flexible. However, if I use the effect with the "Color Gradient" palette, it seems like it is biased towards the background color, leading to more "flashy" lights. On the other hand, if I use a `color_blend` with `SEGCOLOR(1)` and `SEGCOLOR(0)`, it seems more linear and calmer.

wled00/FX.cpp
```diff
@@ -7110,6 +7110,42 @@ uint16_t mode_waterfall(void) {                   // Waterfall. By: Andrew Tulin
...
+    maxBand = temp;
+  }
+
+  // Average the bands' values, scaled by user-specified sensitivity
+  uint32_t value = 0;
+  uint8_t *fftResult = (uint8_t*)um_data->u_data[2];
+  for (uint8_t band = minBand; band <= maxBand; ++band) {
+    value += fftResult[band] * SEGMENT.intensity;
+  }
+  value /= (maxBand - minBand + 1) * (256 / 8);
```
blazoncek, line 7139: (256 / 8) == 32
Perhaps the compiler does the same, but you never know.

wled00/FX.cpp
```diff
@@ -7110,6 +7110,42 @@ uint16_t mode_waterfall(void) {                   // Waterfall. By: Andrew Tulin
...
+  value = constrain(value, 0, 255);
```
blazoncek, line 7140: max(maxBand)==15
min(minBand)==0
max(fftResult[i]) == 255
max(SEGMENT.intensity)==255
max(sum(fftResult[i]*SEGMENT.intensity))==1048560
max(value)==2048

There are 1792 possible values of "value" that get attenuated (and clipped) by this. Intentional?
I am not fond of "guesswork" logic.

wled00/FX.cpp
```diff
@@ -7110,6 +7110,42 @@ uint16_t mode_waterfall(void) {                   // Waterfall. By: Andrew Tulin
...
+
+  SEGMENT.fill(SEGMENT.color_from_palette(value, false, false, 0));
```
blazoncek, line 7142: Fill entire segment with a single color from palette?

wled00/FX.h
```diff
@@ -316,8 +316,9 @@
 #define FX_MODE_WAVESINS               184
 #define FX_MODE_ROCKTAVES              185
 #define FX_MODE_2DAKEMI                186
+#define FX_MODE_FREQPALETTE            187
```
blazoncek, line 319: Instead of increasing total count, please reuse gaps from removed effects.

blazoncek (changes requested):
I am not sure I like the effect or the logic inside it. But that is personal taste.

I would require reuse of effect ID gaps and a few comments in the code prior to consideration of the possibility to merge. Final word will come from @Aircoookie as he will decide to include it or not.

wled00/FX.cpp
```diff
@@ -7110,6 +7110,42 @@ uint16_t mode_waterfall(void) {                   // Waterfall. By: Andrew Tulin
...
+  if (maxBand < minBand) {
+    uint8_t temp = minBand;
+    minBand = maxBand;
+    maxBand = temp;
+  }
+
+  // Average the bands' values, scaled by user-specified sensitivity
+  uint32_t value = 0;
+  uint8_t *fftResult = (uint8_t*)um_data->u_data[2];
+  for (uint8_t band = minBand; band <= maxBand; ++band) {
```
softhack007, line 7136: fftResult[] has max 16 values that need to be averaged. As the uint32_t code seems to have issues with overflows, I would propose to simply use float for the calculation (it won't be a performance problem to do that for max 16 values).

Also the RMS average  (root means squared) would be much more appropriate in this case. You can find an example in the audioreactive usermod code:
https://github.com/MoonModules/WLED/blob/794e3a86711c6923cc83ccc4fe3a2a85d7cacec1/usermods/audioreactive/audio_reactive.h#L326-L333

wled00/FX.cpp
softhack007, line 7142: Would be nice to have two options here (based on checkbox)
A) fill with a single color
B) fill with palette color up to the value of volume (similar to the gravCentric effect), like a VU meter.

I think this will add even more "creative potential" to your effect. You can implement case B just as left-to-right VU meter. Centering, reverse direction, and mirroring can be done with segment options so no need to care about that in the effect itself..

softhack007 (changes requested):
-please fix computation of average volume, and consider calculating RMS average instead of linear average.
- se if it's possible to add "VU meter" as an option, controlled by checkbox  `SEGENV.check1`

wled00/FX.cpp
```diff
@@ -7110,6 +7110,42 @@ uint16_t mode_waterfall(void) {                   // Waterfall. By: Andrew Tulin
...
+    value += fftResult[band] * SEGMENT.intensity;
```
softhack007, line 7137: You could move `* SEGMENT.intensity` out of the loop and multiply the resulting average volume by some factor derived from SEGMENT.intensity. from math point of view that equivalent, however more efficient when looking at the number of math operations.

blazoncek:
@megabug no interest to fix issues?

megabug:
> @megabug no interest to fix issues?

No, I agree with the comments, just need time to get around to them :)

softhack007:
Closing, because there was no progress on this PR since 10 months. 
A similar effect can be achieved by creating a small (min 2pixel high) matrix, and assigning each matrix column to one "shelve" of LEDs.

PR #3161, opened 2023-04-05
Wordclock - Issue with "Norddeutsch"

Hasenpups:
- fix issue that "viertel vor" and "viertel nach" were not correctly according to the settings
- fix issue that "dreiviertel" was not working correctly anymore
- harmonize wording of parameters

### DIFF
--- a/usermods/usermod_v2_word_clock/usermod_v2_word_clock.h
+++ b/usermods/usermod_v2_word_clock/usermod_v2_word_clock.h
@@ -40,39 +40,39 @@ class WordClockUsermod : public Usermod
     // Normal wiring
     const int maskMinutes[14][maskSizeMinutes] = 
     {
-      {107, 108, 109,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1}, // :00
-      {  7,   8,   9,  10,  40,  41,  42,  43,  -1,  -1,  -1,  -1}, // :05 fünf nach
-      { 11,  12,  13,  14,  40,  41,  42,  43,  -1,  -1,  -1,  -1}, // :10 zehn nach
-      { 26,  27,  28,  29,  30,  31,  32,  -1,  -1,  -1,  -1,  -1}, // :15 viertel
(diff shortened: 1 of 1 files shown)


usermods/usermod_v2_word_clock/usermod_v2_word_clock.h
```diff
@@ -423,11 +424,11 @@ class WordClockUsermod : public Usermod
     void addToConfig(JsonObject& root)
     {
       JsonObject top = root.createNestedObject("WordClockUsermod");
-      top["active"] = usermodActive;
-      top["displayItIs"] = displayItIs;
-      top["ledOffset"] = ledOffset;
-      top["Meander wiring?"] = meander;
-      top["Norddeutsch"] = nord;
+      top["Active"] = usermodActive;
```
blazoncek, line 427: Do not change config parameters as they will reset settings when users update.
Please return to original values.

blazoncek (changes requested):
Config parameter names should not be changed.
If needed add `addInfo()` metadata.

usermods/usermod_v2_word_clock/usermod_v2_word_clock.h
Hasenpups: Okay I will change it back :)

blazoncek:
One more recommendation would be to enclose JSON property strings in `F()` macro to reduce RAM usage.

blazoncek (approved):

Hasenpups:
> One more recommendation would be to enclose JSON property strings in `F()` macro to reduce RAM usage.

Done :)

Hasenpups:
> Config parameter names should not be changed. If needed add `addInfo()` metadata.

Done. 
Pull request should be finished for this bug now.

blazoncek (approved):
Although German and Slovenian language share some grammar I am not, by any means, German speaker so I do not know if the semantics of the language are correct. I assume you are German and you do that much better. 😀 

Other than that the code seems in order to me and uses recent features to make user experience better. I did not test it as I do not have Word Clock set up.

PR #3162, opened 2023-04-06
Add LAN8720 reset and new ethernet board

wled-install:
- Add LAN8720 reset according to datasheet
- Add new Ethernet controller (ABC! WLED V43 + Ethernet Shield)


blazoncek:
You may want to remove .h files from unchanged .htm files.
Otherwise looks OK.

I am wondering about Eth reset as I am seeing quite often my Ethernet WLED device becoming unresponsive over ethernet while still working correctly using buttons. May this be somehow related?

wled-install:
Ok thank you for the hint, I've removed .h files that are not changed.
Regarding ethernet getting unresponsive: I donth think my PR will help. If my understanding of WLED implementation is correct, then it calls initEthernet() just once at startup. Even if LAN IC get stuck, there is no mechanism implemented that recognise that and reinit it. For such reinit HW reset in my PR would be required.

blazoncek:
Thanks for the response.
Would you care and try to implement such thing?

wled-install:
I dont actually believe that the issue you have is becouse LAN8720 IC get stuck. I have another application with this IC, ESP32 and esp-idf V4, no problems. Maybe it is rather esp-idf V3 related (or possible, poor bord design/layout, which one do you have?). I'am not sure how to recognise (from the application layer) that the ethernet IC got problems and must be reinitialized.

blazoncek:
I am using @conradopoole shield.
But I do not know what is the culprit as it happen after several days of uptime only. Current remedy I have now is weekly reboot of the device.

wled-install:
Does it always happen after about one week uptime and never just after 1-4 days? Do you have shielded Ethernet cable/connector (outer shield) and quite long cable? If yes, maybe try unshielded one, since it looks like the board does not have correct outer shield connection (because of space I believe). It also looks like (what can be seen in the picture since no layout data available) there are also some other issues with layout design (because he wanted to fit everything into small space with only 2 layers: position of VDD/VDDA decoupling capacitors, differential line routing, GND plane). Even I don’t think they are highly problematic, you never know exactely. It would be beneficial to understand if it is rather HW or SW issue. You can try one of OLIMEX boards (they look extremely well designed) or wait some week until I have my shield, I can send you one with controller for free if you want to test. I believe meanwhile WLED can be also compiled using Arduino-esp32 V2.0.x too (that is esp-idf V4 based). It also could be on option to test. As soon as I get final version of my Ethernet shield/controller, I would start long-term run. If I come to the conclusion there are some SW issues, I can check what can be done in SW.

blazoncek:
It is always after a prolonged uptime of several days. Regarding reboot, if I do it regularly (on schedule) a soft reset is enough but that must be before it "locks up". If I do not set up schedule and device "locks up" it may occasionally be enough to unplug Ethernet but sometimes it will require power cycle.
The cable is short, just 2m, unshielded, plugged directly to switch.

I've had this shield plugged to a Dig-Uno with D1 mini32 clone and it was quite regularly loosing Ethernet (it was still broadcasting UDP messages, but web interface was unreachable). I've connected it to @srg74 universal shield couple of months ago (with integrated ESP32) and it works much better but still occasionally "locks up".

It is a test unit so it gets reboot every now and then.

wled-install:
Strange, I mean if UDP still works, it cannot be an issue of LAN8720 (as it is pure physical interafce, all the logic is in ESP32).

blazoncek:
Why close?

blazoncek:
Ah. Next time create a new branch and use that to make PR.
https://github.com/Aircoookie/WLED/wiki/How-to-properly-submit-a-PR

Issue #3117, opened 2023-03-08
Create and edit custom palettes via API


Wegi:
**Is your feature request related to a problem? Please describe.**
Not a problem, just an enhanced experience.

**Describe the solution you'd like**
I would like to use the JSON or HTTP API to upload or edit a custom palette. Currently, the /edit interface needs to be used, which does not work as nicely, because the user has to do some actions.

**Additional context**
The goal is to create and edit the palettes via API, so an enduser does not have to do anything manually, like uploading files.

Thanks for the very good work on WLED.

blazoncek:
Please see #246 and contribute if you can.
Closed as a duplicate.

Issue #246, opened 2019-10-07
Set custom palette via HTTP/MQTT


jwingefeld:
Hi,

i think this qualifies as feature request. I think it would be really convenient to be able to set custom palettes via http (maybe as json?). This could be one dedicated "slot" with a limited number of colors so RAM usage would stay predictable.

Possible use cases would be:
-  a color palette editor in the web interface
- pywal (generates color palettes for your Terminal from the current wallpaper) integration. So you'd load up a picture of a sunrise and BOOM, matching color palette on your desk.
- integration into $GENERIC_RGB_GAMING_TOOL to get colors matching your 1337-gaming-battlestation

Aircoookie:
Hi,

thank you for the feature request! I've already planned such a feature via a JSON API. Current timeline for implementation is likely December.

NietGiftig:
@Aircoookie 
Until yesterday I used my own proggy for my floorlamp leds, using node-red as home automation.
Now I have resoldered this floorlamp to the default WLED data-pin and use WLED.
Verry (very!!!) satisfied.
But this request is a very good addition to include some of my old palettes with an easy method.
I can store in, and send any posible palettes from node-red, so i hope not a huge impact to the progmem in WLED.
A limited count of free useable palette memory would be sufficient 
Any timeline possible ?

blazoncek:
It may not be JSON/HTTP or MQTT compatible but I have implemented (up to 10) custom palettes that can be stored on embedded filesystem and selected at runtime in the same way as built-in palettes.
It will still need thorough testing though as the option is part of the larger change I created (2D support and audio reactive as a usermod).
Commit: https://github.com/blazoncek/WLED/commit/df534d30bf5087888f0df644ffb8f7c5225eb263

Orace:
I'm currently working on a [palette editor UI](https://codepen.io/Oracetonick/pen/qBYdwaN).
I plan to integrate it with wled.
Feedback are welcome (I'm more of a C# guy than vanilla JS).

Aircoookie:
@Orace that is super cool and exactly what I had in mind of implementing!
A "Remove color stop" button and maybe a row of recently used colors are the only things I would consider still missing,

Orace:
@blazoncek, I plan to integrate it in WLED current UI as a _standard_ palette:

- [ ] integration
- [ ] switch to edit mode with a collapse button
- [ ] use the current color selection UI for _color stops_ (a.k.a. thumbs) 
[...]

PR #3164, opened 2023-04-08
Custom Palette Editor

werkstrom:
Ref #246 and #3117


werkstrom (commented):
Ref #3117 and #246

blazoncek:
Seems to work well but there are glitches.
a) When you click for color selector it remains open even after color selection and may cause user frustration.
b) When you click "Delete" and then drag slider, trashcan remains at original position.
c) Color slider uses whole page width (which, on 3000px display is awkward).

Proposals:
a) Add a note to close color selector manually or, if possible, close it once color is selected.
b) Remove trashcan on slider move and un-select deletion.
c) As there are only 18 possible entries and there can only be 256 positions it may make sense to limit width to 512px. Or thereabouts.

Enhancements:
1) if you can add loading of palette from FS that would be great.
2) when loading a palette, allow the use of built-in palettes (as a starting point)
3) maybe add a "Reset" button (for palettes loaded)

I will add logic to the web server to force custom palette preview update when palette is uploaded.

werkstrom:
a. Agreed, but unfortunatly that is controlled by the browser. And also it would be hard to identify _when_ you are "done" since there are many parameters to input. I tend to pull the sliders around a bit before being happy with the color. Anyway, unless we build a custom color selector... we're stuck with this behaviour ;)
b. Done
c. Aha, I thought we were planning on putting it into some other container in the WLED UI, so it is made to adjust to it's parent.  Added an outer DIV and set width. Done

Enhancments:
Can WLED provide that info? Couldn't find an API. I can only find _which_ palettes exist, not their composition.

blazoncek:
If you embed custom palette creator into WLED UI then everything is at your disposal from UI environment.
You can get `localStorage.getItem("wledPalx")` array with palette previews from which to create a starting point for custom palettes. You can also delete this item from local storage and force a reload by calling `loadPalettesData()`.

If the container with 300px width is too narrow you can re-purpose overlay containers to temporarily display palette creator.

Until there is also a remove custom palette (albeit only the last) button and backend logic for removing, this is (very useful but) still POC to me.

I hope to have some time this weekend to add a few commits to this PR.

blazoncek:
Played a little more with this:
- does not work on mobile (color picker)
- UX is bad on mobile (small selectors, text)
- should be centered on the page
- it may be best to leave it as a separate page and not embed it into WLED UI (but I am open to good implementation)
- adding `localStorage.removeItem("wledPalx");` on successful upload will force reload of palette previews upon return to WLED UI
- it should be named "Custom palette editor" and allow editing custom palettes.

I am going to change this to draft as it is not ready for prime time.

werkstrom:
The local storage appears not to have the actual palette but instead some approximated, fixed spacing gradient that also has a lower resolution that the palette file (its last index is 240 with the same color as the 255 index in the file. This means the local storage cannot be used for getting the palette data.
image
image

I have also concluded that index 255 in local storage equals palette 0, 254 = palette 1 and so on. However. Since we cannot use the info already on the client, I've resorted to loading the JSON files from the device directly. That takes time when loading the palettes. Optimal solution would be the data in local storage being correct and in same format as presetx.json. No biggie though but if I got this wrong, please correct me and I'll get the data from the right place.

Focusing on functional, leaving potential mobile adjustment to later.

blazoncek:
The idea behind palette previews is for them to load as quickly as possible and not strain ESP.
They should be enough for creating new custom palettes. Consider them as a rough templates from which to populate new palette.
For editing existing custom palette the solution to load actual JSON file seems in order.

blazoncek:
You may need to resort to IRO picker for mobile (and then use it for PC as well).

werkstrom:
Limited testing here... But it appears this issue only applies to the custom paletts in that array. The permanent palettes appear to straight off represent the palettes (they do not all have 16 positions in their gradient arrays and they are spread unevenly).
Example:
Element of the C9 palette in UI:
image
Array of item
```"48":[[0,184,4,0],[60,184,4,0],[65,144,44,2],[125,144,44,2],[130,4,96,2],[190,4,96,2],[195,7,7,88],[255,7,7,88]]```

As for now, for the custom palette editor, I read the files as they are correct, That most defiantly puts unnecessary load on the device, but currently the only way (i know of) to get the correct palette data. Would be awesome if the correct array was present for the custom palettes as well to lessen the load on the device and make the UX better. Also the preview array and gradient for the custom palettes on the main page has unnecessary data and does not reflect the palette. Here's an example where the main UI preview misses the end green color. 
In the main UI:
image
In CPal:
image

**Comment by blazoncek** [comment]
Palette previews are created in *json.cpp* in function `setPaletteColors()`. As there are two ways how to represent palette in WLED there are two ways to create previews. One is lossy or imprecise, hence lost details.
If you look for imperfections you will find more, same as like looking at pixels in JPEG image.

**Comment by blazoncek** [comment]
Nice!
The only issue is that custom palettes can be created using hex strings as well, like:
```json
{
	"palette": [
		0,	"FF0000",
		62,	"FF0000",
		63,	"FFA000",
		126,	"FFA000",
		127,	"00FF00",
		191,	"00FF00",
		192,	"0000FF",
		255,	"0000FF"
	]
}
```

werkstrom:
No biggie. CPal uses HEX do draw, should be fixable. Just not tonight. ;)

blazoncek:
This is now ready to merge IMO.

werkstrom:
You can now use the static palettes as templates for your custom palettes.

chrome_apdx0KxC8K

blazoncek:
IMO this is now perfect.
Not yet thoroughly tested but first impression is fantastic.

PR #3171, opened 2023-04-11
Add support for segment sets (groups of segments)

blazoncek:
Up to 4 segment groups can be defined.
Visible only when more than 2 segments are created and has no other functionality other than allowing quick segment selection in UI.

NOTE: Reduces possible audio simulations from 8 to 2 (keeping BeatSin and WeWillRockYou) to accommodate segment groups without increasing `Segment` RAM footprint.


blazoncek:
@Aircoookie would you agree with adding segment groups?

blazoncek:
As sound (if used from microphone) affects all segments equally I would rather see a global sound simulation option instead of per-segment one.

With such approach the implementation with 1 bit could mean 0 use simple beatsin (per segment) and 1 use global.

And honestly stuffing more simulation options could be done in usermod.

blazoncek:
I agree on name change so instead of segment groups we can use segment sets.

blazoncek:
@Aircoookie any chance to give this another try?
I have plenty of bugfixes queued. Waiting for this PR to be merged or rejected.

Aircoookie (approved):
Ready to be merged! (finally, sorry it took so long)
Ideally the `g-icon` would also show in expanded segment view (though the only place I see that would be available is above the delete button), though this does not have to be added immediately.

PR #3173, opened 2023-04-14
Allow SPI clock speed selection.

blazoncek:
Adds support for changing SPI clock speed during runtime.


Aircoookie (commented):
Approving all contents, but I still need to check why CI no longer works. I would have assumend too that merging in main would help, I may need to mess with the actions file.

blazoncek:
> Approving all contents, but I still need to check why CI no longer works.

So, should we merge or wait still?

PR #3182, opened 2023-04-23
Apply correct iOS scroll to all tabcontent

marcoscv-work:
Right now the touch scroll is only applying to `#Effects`. with this change all scrolls works properly, video of the current state:

https://user-images.githubusercontent.com/7963804/233815400-c85a0863-c901-4fa9-97c1-06fefce45466.mp4

### DIFF
--- a/wled00/data/index.css
+++ b/wled00/data/index.css
@@ -318,9 +318,6 @@ button {
 	height: 100%;
 	overscroll-behavior: none;
 	padding: 0 4px;
-}
-
-#Effects {
 	-webkit-overflow-scrolling: touch;
 }
 


blazoncek:
Interesting, I see momentum based scrolling everywhere (on all 4 tabs) when using a browser.
Please try browser (Safari, Firefox,Opera, ...). You can also try alternate WLED app by @Moustachauve which is more modern and recently updated.

It is an issue of the iOS app which has not been updated in a while.
`-webkit-overflow-scrolling` is a non-standard CSS extension which is no longer supported by any browser.

Aircoookie:
The `-webkit-overflow-scrolling` property was only added to the effects tab as that was originally the only tab that could be higher than the screen height. Approving of this change, though as @blazoncek mentioned it only has an effect on iOS 5-12 and thus we might be able to remove it entirely in a few years.

PR #3184, opened 2023-04-23
Soap, new 2D effect

blazoncek:


Aircoookie:
This effect is absolutely beatiful!

wled00/FX.cpp
```diff
@@ -7418,7 +7418,8 @@ uint16_t mode_2Dsoap() {
         int32_t joffset = *scale32_y * (j - rows / 2);
         uint8_t data = inoise16(*noise32_x + ioffset, *noise32_y + joffset, *noise32_z) >> 8;
         noise3d[i*cols + j] = scale8(noise3d[i*cols + j], SEGMENT.intensity) + scale8(data, 255 - SEGMENT.intensity);
-        SEGMENT.setPixelColorXY(i, j, CHSV(~noise3d[i*cols + j]*3, 255, 255)); //ColorFromPalette(SEGPALETTE,~noise3d[i*cols + j]*3));
+        SEGMENT.setPixelColorXY(i, j, ColorFromPalette(SEGPALETTE,~noise3d[i*cols + j]*3));
+        SEGMENT.color_wheel(1);
```
blazoncek, line 7422: Is this intentional?

wled00/FX_fcn.cpp
```diff
@@ -228,6 +228,7 @@ CRGBPalette16 &Segment::loadPalette(CRGBPalette16 &targetPalette, uint8_t pal) {
     case FX_MODE_GLITTER    : pal = 11; break; // rainbow colors
     case FX_MODE_SUNRISE    : pal = 35; break; // heat palette
     case FX_MODE_RAILWAY    : pal =  3; break; // prim + sec
+    case FX_MODE_2DSOAP     : pal = 11; break; // rainbow colors
```
blazoncek, line 231: It may also be good to select palette using metadata.

wled00/FX.cpp
Aircoookie: No 😅

wled00/FX_fcn.cpp
Aircoookie: TBH I prefer using coded defaults over metadata for palettes (metadata is VERY useful for speed, intensity and custom sliders as they have no concept of a "default value").
For palettes, I'd personally reserve setting a specific palette via metadata only where the effect absolutely only looks good with that one palette, and even in that case we could consider setting `pal=0` in metadata and hardcoding the default palette. The main reason for that is that the user may be confused by palettes changing "by themselves", potentionally even without them noticing if they change FX while not in PC mode (and I also like to keep fxdata as small as possible)

Removed `ledsbuff` and the second inner loop respectively, they seem like a remnant of porting the FX which no longer required in WLED :)

PR #3193, opened 2023-05-02
Fix errors in DDP implementation

coral:
I think these values are wrong given the spec.

`0x0A` in binary is 0b00001010

**Spec is:**
> Bits: C R TTT SSS
> C is 0 for standard types or 1 for Customer defined
> R is reserved and should be 0.
> TTT is data type 
> 	000 = undefined
> 	001 = RGB
> 	010 = HSL
> 	011 = RGBW
> 	100 = grayscale
> SSS is size in bits per pixel element (like just R or G or B data)
> 	0=undefined, 1=1, 2=4, 3=8, 4=16, 5=24, 6=32

So in this case, `0x0A` maps to 010 which means 2, 4 bits per pixel
but really what the constant wants to represent is 24 bits, which should be 110 so `0x0D` and 32 bits should be `0x1E`

### DIFF
--- a/wled00/e131.cpp
+++ b/wled00/e131.cpp
@@ -25,7 +25,7 @@ void handleDDPPacket(e131_packet_t* p) {
     }
   }
 
-  uint8_t ddpChannelsPerLed = (p->dataType == DDP_TYPE_RGBW32) ? 4 : 3; // data type 0x1A is RGBW (type 3, 8 bit/channel)
+  uint8_t ddpChannelsPerLed = (p->dataType & 0b00111000 == 0b011) ? 4 : 3; // data type 0x1B (formerly 0x1A) is RGBW (type 3, 8 bit/channel)
 
   uint32_t start =  htonl(p->channelOffset) / ddpChannelsPerLed;
   start += DMXAddress / ddpChannelsPerLed;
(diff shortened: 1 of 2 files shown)


blazoncek:
If this is true then it should be fixed upstream. Please contact @forkineye.

coral:
@blazoncek This code is not in the upstream but was added inside the WLED repo ( https://github.com/forkineye/ESPAsyncE131/blob/7ff3c945e6750308db6619814b7e75f585d60aab/ESPAsyncE131.h#L53 there is no reference to DDP in the upstream repo)

If i use `git blame` the DDP support was added in WLED here https://github.com/Aircoookie/WLED/blame/cd82a343928622d3a19c696e1e028b987e29c871/wled00/src/dependencies/e131/ESPAsyncE131.h#L56

blazoncek:
In such case let's see what @Aircoookie has to say.

blazoncek:
Looks like the specs were updated *after* the change to WLED was made.
I do remember @Aircoookie talking to DDP authors about extending/defining RGBW support.

coral:
> Looks like the specs were updated _after_ the change to WLED was made. I do remember @Aircoookie talking to DDP authors about extending/defining RGBW support.

Just to state the obvious, this is the spec http://www.3waylabs.com/ddp/ I've been using.

You are right, using waybackmachine https://web.archive.org/web/20200223131548/http://www.3waylabs.com/ddp/ the spec uses a different alignment which tracks. The older spec has a different alignment but even then I'm not sure that the code is valid, the older spec with this aligns to 8 bits per pixel so maybe there was confusion if the spec referred to pixel or channel?

Aircoookie:
Hi! I'm 99% certain that `bits per pixel element (like just R or G or B data)` in the spec refers to bits per channel, not bits per pixel. Still, great thing you noticed the spec update, we should update the byte to `0x0B` for 8 bit per channel RGB and `0x1B` for 8 bit per channel RGBW respectively.

PR #3211, opened 2023-05-21
Changes for allowing Alexa to change light color to White when auto-calculating from RGB

Sebbb:
Hello,

this addresses https://github.com/Aircoookie/WLED/issues/3209, I am sure this code can be improved as it only checks the "Global override for Auto-calculate white" status instead of the one from each configured LED output, on the other hand, I would not know what to do if there are differently configured LEDs.

But it works for me and is definitely an improvement, because the light doesn't simply go off when Alexa sends a CT value.

PS: The indentation is already broken in the original file, it uses mixed spaces and tabs, it probably would make sense to fix that, for the sake of clarity of this patch I didn't do it.

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -771,6 +771,7 @@ class WS2812FX {  // 96 bytes
       getActiveSegmentsNum(void),
       getFirstSelectedSegId(void),
       getLastActiveSegmentId(void),
+      getActiveSegsLightCapabilities(bool selectedOnly = false),
       setPixelSegment(uint8_t n);
 
     inline uint8_t getBrightness(void) { return _brightness; }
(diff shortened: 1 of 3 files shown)


blazoncek:
If you enabled edit when submitting PR and created separate branch then collaborators can edit your branch directly, improving code if necessary.

Sebbb:
@blazoncek I did, I think.. 😅

Aircoookie:
I've updated this to use the segment light capabilities instead of relying on the global override setting.
@Sebbb if you have time, please go ahead and check if it still works as intended, if so this is ready to merge!

Sebbb:
@Aircoookie I just tried it, works fine :)

PR #3216, opened 2023-05-23
ES8388 Support

netmindz:
Think I've pulled all the right commits from MM to pull in the new ES8388 driver.

I will require testing against the other parts of the audioreactive usermod before merge


blazoncek:
Thanks.
I was talking to @softhack007 the other day to modify I2C microphones to use `Wire1` instead of `Wire` to avoid conflicts with other I2C peripherals attached to WLED. Could you update the audio source to utilise that?

netmindz:
The pin numbers vary between different boards, I would need to check if there are multiple I2C bus or just single bus but different pins on different boards. I suspect there is only one.

Could you expand a little @blazoncek on what conflict you are thinking of? Perhaps open a fresh issue by stating if this is a confirmed issue or just trying to prevent potential issues, what the change would be, what testing is needed etc?

I see no reference to Wire1, so I presume you are talking about creating a new one. If this is a general change across all the AudioSources, then that should really be done in it's own PR

blazoncek:
ESP32 has two I2C HW assisted modules/interfaces. First (also default) can usually be addressed using `Wire` global variable (and this is the one most usermods use and WLED allocates as global; usually as pins 21 & 22) while the second is often addressed like `Wire1` global variable.
See: https://github.com/espressif/arduino-esp32/blob/master/libraries/Wire/src/Wire.h

`Wire1` is unused in WLED and existing usermods.

I am not familiar with complex microphones using I2S & I2C interfaces and how they implement bus sharing but if there is a slight chance they (or the drivers you are implementing in `audio_source.h`) cannot share I2C bus with other I2C devices I would prefer for audio to use its own I2C interface.
ESP32 I2C implementation (muxer) allows for easy pin switching so there should be no issues with existing hardware implementations.

BTW I noticed some usermods to implement setting I2C bus clock speed so that is also one of the considerations to take into account.

I recommend reading ESP32 datasheet for more information.
https://www.espressif.com/sites/default/files/documentation/esp32_datasheet_en.pdf

netmindz:
Thanks for the info, but as the I2C is only used for the initial initialisation of the chip, it's unlikely you are going to see a clash with other devices

blazoncek:
Then use global I2C as defined in WLED and remove local references.

netmindz:
See https://github.com/MoonModules/WLED/pull/48/files @blazoncek  - Is this what you mean?

blazoncek:
Yes, I have made similar change on my local dev branch. So IMO this is acceptable.
I hope @softhack007 also agrees.

netmindz:
Now we need to find someone who actually uses ES2743

netmindz:
Until we can find testers to confirm the change to the existing driver, we should just stick with what has been tested currently for ES8388 and can cleanup theoretical i2c improvements as an when they can be tested fully

usermods/audioreactive/audio_source.h
```diff
@@ -462,6 +462,157 @@ class ES7243 : public I2SSource {
...
+
+    void deinitialize() {
+      // Release SDA and SCL pins of the I2C interface
+      PinManagerPinType es8388Pins[2] = { { pin_ES8388_SDA, true }, { pin_ES8388_SCL, true } };
+      pinManager.deallocateMultiplePins(es8388Pins, 2, PinOwner::HW_I2C);
+      I2SSource::deinitialize();
+    }
+
+  private:
+    int8_t pin_ES8388_SDA;
```
blazoncek, line 613: As mentioned in another comment, this will go away in beta 3 and will rely on global I2C pins (`Wire` global).

In case of I2C bus conflicts the code will need to change to use `Wire1` global (second I2C bus).

blazoncek (approved):
My comments are purely for code aesthetic.

blazoncek:
@netmindz this will not compile under plain WLED AC. Looks like there are remnants of MM fork in this PR.
I will try to bring it up to speed but please test it with WLED AC after I am done.

blazoncek:
Unfortunately permissions you set on this PR prevent me from updating it.

netmindz:
The CI builds are running green so not sure why you are seeing compile errors. Also "allow edits" is also enabled on this PR

blazoncek:
CI builds do not compile usermods.
Have you even tried compiling this PR under WLED? Not from MM fork.

```
In file included from wled00/../usermods/audioreactive/audio_reactive.h:49:0,
                 from wled00/usermods_list.cpp:153:
wled00/../usermods/audioreactive/audio_source.h: In member function 'void ES8388Source::_es8388I2cBegin()':
wled00/../usermods/audioreactive/audio_source.h:475:77: error: 'ERRORSR_PRINTLN' was not declared in this scope
         ERRORSR_PRINTLN(F("AR: ES8388 failed to initialize I2C bus driver."));
                                                                             ^
wled00/../usermods/audioreactive/audio_source.h: In constructor 'ES8388Source::ES8388Source(int, int, float, bool)':
wled00/../usermods/audioreactive/audio_source.h:568:62: error: no matching function for call to 'I2SSource::I2SSource(int&, int&, float&, bool&)'
       I2SSource(sampleRate, blockSize, sampleScale, i2sMaster) {
                                                              ^
wled00/../usermods/audioreactive/audio_source.h:170:5: note: candidate: I2SSource::I2SSource(int, int, float)
     I2SSource(SRate_t sampleRate, int blockSize, float sampleScale = 1.0f) :
     ^
wled00/../usermods/audioreactive/audio_source.h:170:5: note:   candidate expects 3 arguments, 4 provided
wled00/../usermods/audioreactive/audio_source.h:168:7: note: candidate: constexpr I2SSource::I2SSource(const I2SSource&)
 class I2SSource : public AudioSource {
       ^
wled00/../usermods/audioreactive/audio_source.h:168:7: note:   candidate expects 1 argument, 4 provided
wled00/../usermods/audioreactive/audio_source.h:168:7: note: candidate: constexpr I2SSource::I2SSource(I2SSource&&)
wled00/../usermods/audioreactive/audio_source.h:168:7: note:   candidate expects 1 argument, 4 provided
wled00/../usermods/audioreactive/audio_source.h: In member function 'virtual void ES8388Source::initialize(int8_t, int8_t, int8_t, int8_t, int8_t, int8_t)':
wled00/../usermods/audioreactive/audio_source.h:580:89: error: 'ERRORSR_PRINTF' was not declared in this scope
         ERRORSR_PRINTF("\nAR: invalid ES8388 I2C pins: SDA=%d, SCL=%d\n", sdaPin, sclPin); 
                                                                                         ^
wled00/../usermods/audioreactive/audio_source.h:585:85: error: 'ERRORSR_PRINTF' was not declared in this scope
         ERRORSR_PRINTF("\nAR: invalid I2S pin: SCK=%d, MCLK=%d\n", i2sckPin, mclkPin); 
                                                                                     ^
wled00/../usermods/audioreactive/audio_source.h:593:100: error: 'ERRORSR_PRINTF' was not declared in this scope
         ERRORSR_PRINTF("\nAR: Failed to allocate ES8388 I2C pins: SDA=%d, SCL=%d\n", sdaPin, sclPin); 
```

As for edits, looks like the cause may be you forked MM, not AC and this is preventing me from pushing to the PR (as I usually do when updating PRs).
Screenshot 2023-06-16 at 22 10 45

I do have fixes locally.

netmindz:
I had compiled but missed the fact the usermod is only active in some of the targets not all.

For simplicity, I've just swapped the loggers to debug rather than error rather than messing about with the macros and removed the i2sMaster param as think it should work without. Need  to do some actual hw testing to confirm that change

netmindz:
I've been away for a few weeks, but should be able to test in next few days

netmindz:
Updated now to reflect the removal of I2C from the AudioSource setup and tested with LyraT including passthru

troyhacks:
Thanks @netmindz and @blazoncek for getting this done. I much appreciate your work on my little contribution. 😀

Issue #3196, opened 2023-05-05
Always turns on in bright orange


ElToberino:
### What happened?

At ESP32 startup /reset all leds light up in bright orange for a short moment before the "Apply Preset xx at boot" and the brigthness limiter take place. This is a problem for my setup as I enabled the brightness limiter due to a limited power supply. The initial orange draws to much current and causes a "reset loop" of the psu.
This problem was also described in the [discourse forum](https://wled.discourse.group/t/always-turns-on-orange/3307).

### To Reproduce Bug

Start up ESP32 with an "Apply Preset xx at boot" setting other than orange.

### Expected Behavior

No orange flash at startup - first light up of the leds with the "Apply Preset xx at boot" setting.

### Install Method

Self-Compiled

### What version of WLED?

WLED 0.14.0-b2 (build 2304280)

### Which microcontroller/board are you seeing the problem on?

ESP32

### Anything else?

Environment: 
ESP32 board (similar DEVKIT V1).
1536 leds attached (768 on GPIO 16, 768 on GPIO 4).

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

blazoncek:
I have the fix ready, but is not yet published.
If you are willing to test it, please join Discord and tag me in beta testing channel.

Posiek:
I have similar problem recently. It look like WLED reset and after startup it light up in bright orange. Yesterday LEDs was turn of and at 4:20 AM turned on its own. Today it turn off randomely and then after 20 minutes turn on again and light up with orange. I thought that something is connected in not proper way?

blazoncek:
Please use Discord where I posted binaries with the fix included.

moose4lord:
Sorry, but it's not clear to me if this has been fixed in the latest code. I downloaded and compiled the current main branch from GitHub, but I still see the bright orange flash when the strip is first turned on. It also flashes when I save anything on the config menus. #3220 seems to imply the fix has been merged with main, but I don't see it.

blazoncek:
@moose4lord you may have other issues then.

moose4lord:
It seems very similar to the issue @ElToberino raised. I have an ESP32 with a  strip of 64 WS2812s. In LED Preferences I have the "Turn LEDs on after power up/reset" checkbox checked, but the default brightness set to 16, so I don't scorch my eyeballs when it turns on. If I reset the ESP32, it briefly flashes orange (as if brightness=128), then resets to the correct intensity (brightness=16).

[...]

Issue #2263, opened 2021-10-10
MQTT retain setting


Aircoookie:
**Is your feature request related to a problem? Please describe.**
WLED MQTT messages are mostly retained, this could lead to broker storage issues and/or annoying residual messages.

**Describe the solution you'd like**
A checkbox for setting whether publishes from WLED should retain or not.

**Describe alternatives you've considered**
Editing mqtt.cpp

**Additional context**
By Karl_S via Discord #learning

blazoncek:
I would assert that retained messages are a bad habit, especially since there is no user friendly way to remove obsolete retained messages.
There are some justifications for them (i.e. quicker last known state retrieval, etc) but potential problems out-weight them.
Perhaps adding periodic _state_ messages (1/min) could be used instead.
It has to be noted that many (and/or large) MQTT messages may yield dropped frames.

blazoncek:
Ordinary people don't use MQTT Explorer.
They should have no need to as well.

blazoncek:
There are 3 topics that are retained: `/g`, `/c` and `/status`.
As status is also LWT it should not be non-retained.

My proposal is to remove retain options from `/g` and `/c` topics altogether.

Issue #1984, opened 2021-05-22
Show EffectID and PaletteID on MouseOver or when clicked on the GUI 


jkwim:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
When the Web GUI is used, if the EffectID and PaletteID can be shown with MouseOver or when clicked, that would help for people like me who try to send MQTT commands from another device using JSON API.

Thank you for your ideas for making WLED better!

blazoncek:
Effects (with the visualization) are listed in [KB](https://kno.wled.ge/features/effects/).
Similarly palettes are [listed](https://kno.wled.ge/features/palettes/) too.

Adding "title" to buttons to display IDs may seem good to technical users but may be confusing for others.

I propose to close this issue without implementing.

blazoncek:
Fixed in #3220

PR #3220, opened 2023-05-28
Feature implementation

blazoncek:
- #2263
  - #1984
  - #3196
Better PSRAM handling
platformio.ini update
On/Off bus handling
changelog update


Aircoookie (approved):

wled00/wled.cpp
```diff
@@ -321,25 +320,33 @@ void WLED::setup()
...
+#if defined(ARDUINO_ARCH_ESP32) && defined(BOARD_HAS_PSRAM)
+  #if defined(CONFIG_IDF_TARGET_ESP32S3)
+  // S3: reserve GPIO 33-37 for "octal" PSRAM
+  managed_pin_type pins[] = { {33, true}, {34, true}, {35, true}, {36, true}, {37, true} };
+  pinManager.allocateMultiplePins(pins, sizeof(pins)/sizeof(managed_pin_type), PinOwner::SPI_RAM);
+  #elif defined(CONFIG_IDF_TARGET_ESP32S2)
+  // S2: reserve GPIO 26-32 for PSRAM
+  managed_pin_type pins[] = { {26, true}, {27, true}, {28, true}, {29, true}, {30, true}, {31, true}, {32, true} };
+  pinManager.allocateMultiplePins(pins, sizeof(pins)/sizeof(managed_pin_type), PinOwner::SPI_RAM);
+  #elif defined(CONFIG_IDF_TARGET_ESP32C3)
```
softhack007, line 332: Not sure if the pins for S2/C3 are compatible with pinMager.isPinOK rules.  Did you check that SPI_RAM pins actually get allocated?

allocateMultiplePins() does 
``` C++
if (!isPinOk(gpio, mptArray[i].isOutput)) 
   shouldFail = true;
[...]
if (shouldFail) return false;
```
isPinOk checks:
``` C++
 #if defined(CONFIG_IDF_TARGET_ESP32C3)
    // strapping pins: 2, 8, & 9
    if (gpio > 11 && gpio < 18) return false;     // 11-17 SPI FLASH
    if (gpio > 17 && gpio < 20) return false;     // 18-19 USB-JTAG
  [...]
  #elif defined(CONFIG_IDF_TARGET_ESP32S2)
    // strapping pins: 0, 45 & 46
    if (gpio > 21 && gpio < 33) return false;     // 22 to 32: not connected + SPI FLASH
    // JTAG: GPIO39-42 are usually used for inline debugging
    // GPIO46 is input only and pulled down
#endif
```

blazoncek: Good catch!
Though the code may fail to allocate pins in current implementation, it may be future proof if we decide to allow `isPinOK()` to return `true` for SPI pins (as is the case for octal SPI on S3). There are no side effects when the allocation fails in case of `isPinOK()` returning `false`.
If allocation fails due to `isPinOK()` any and every subsequent attempt to allocate such pin will also fail. Regardless of the owner.

PR #3222, opened 2023-05-29
Add ANIMartRIX usermod

netmindz:
New usermod that makes the ANIMartRIX effects available to WLED

### DIFF
--- a/usermods/usermod_v2_animartrix/usermod_v2_animartrix.h
+++ b/usermods/usermod_v2_animartrix/usermod_v2_animartrix.h
@@ -1,0 +1,447 @@
+#pragma once
+
+#include "wled.h"
+#include <ANIMartRIX.h>
+
+#warning WLEDMM usermod: CC BY-NC 3.0 licensed effects by Stefan Petrick, include this usermod only if you accept the terms!
+//========================================================================================================================
+
+
+static const char _data_FX_mode_Module_Experiment10[] PROGMEM = "💡Module_Experiment10 ☾@;;1";
+static const char _data_FX_mode_Module_Experiment9[] PROGMEM = "💡Module_Experiment9 ☾@;;1";
+static const char _data_FX_mode_Module_Experiment8[] PROGMEM = "💡Module_Experiment8 ☾@;;1";
+static const char _data_FX_mode_Module_Experiment7[] PROGMEM = "💡Module_Experiment7 ☾@;;1";
+static const char _data_FX_mode_Module_Experiment6[] PROGMEM = "💡Module_Experiment6 ☾@;;1";
+static const char _data_FX_mode_Module_Experiment5[] PROGMEM = "💡Module_Experiment5 ☾@;;1";
+static const char _data_FX_mode_Module_Experiment4[] PROGMEM = "💡Module_Experiment4 ☾@;;1";
+static const char _data_FX_mode_Zoom2[] PROGMEM = "💡Zoom2 ☾@;;1";
+static const char _data_FX_mode_Module_Experiment3[] PROGMEM = "💡Module_Experiment3 ☾@;;1";
+static const char _data_FX_mode_Module_Experiment2[] PROGMEM = "💡Module_Experiment2 ☾@;;1";
+static const char _data_FX_mode_Module_Experiment1[] PROGMEM = "💡Module_Experiment1 ☾@;;1";
+static const char _data_FX_mode_Parametric_Water[] PROGMEM = "💡Parametric_Water ☾@;;1";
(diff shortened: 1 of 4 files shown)


blazoncek:
Where do you get `<ANIMartRIX.h>` from?
It is not included with source files.

I've seen in your repo that you include your custom `lib_deps`, perhaps adding that info to readme.md would be good.

netmindz:
I wasn't sure what the normal practice was AC and deps of usermods? I didn't just want to slam in a dep into all builds. I don't know of a way of adding lib_deps based on the USERMOD_ANIMARTRIX flag

netmindz:
Readme updated

blazoncek:
My recommendation would be to include all dependencies in usermod folder if all of the files are mature (if all files are under your control).
If, on the other hand, you consider Animartrix.h to be WIP then using `lib_deps` is a viable choice. Though I would avoid using commit ID.

netmindz:
The refactor to make it an OO library is a work in progress, so no version release tags not finalised API, so to ensure consistent builds, best practice to lock to a given revision, not just to a WIP branch

usermods/usermod_v2_animartrix/usermod_v2_animartrix.h
```diff
@@ -7,58 +7,58 @@
...
-static const char _data_FX_mode_Lava1[] PROGMEM = "💡Lava1 ☾@;;1";
-static const char _data_FX_mode_Caleido3[] PROGMEM = "💡Caleido3 ☾@;;1";
-static const char _data_FX_mode_Caleido2[] PROGMEM = "💡Caleido2 ☾@;;1";
-static const char _data_FX_mode_Caleido1[] PROGMEM = "💡Caleido1 ☾@;;1";
-static const char _data_FX_mode_Distance_Experiment[] PROGMEM = "💡Distance_Experiment ☾@;;1";
-static const char _data_FX_mode_Center_Field[] PROGMEM = "💡Center_Field ☾@;;1";
-static const char _data_FX_mode_Waves[] PROGMEM = "💡Waves ☾@;;1";
-static const char _data_FX_mode_Chasing_Spirals[] PROGMEM = "💡Chasing_Spirals ☾@;;1";
-static const char _data_FX_mode_Rotating_Blob[] PROGMEM = "💡Rotating_Blob ☾@;;1";
+static const char _data_FX_mode_Module_Experiment10[] PROGMEM = "💡Module_Experiment10 ☾@;;1;2";
```
blazoncek, line 10: Beware that this will also select "* Random Cycle" palette as a selected palette. (...;1;...)
Use either ! or empty.

netmindz: This different in AC? In MM it shows currently as this effect does not use a palette

netmindz: Why not just test my code rather than make guesses about runtime?

blazoncek: `...;1;...` is not technically correct in any case. It may behave as you intend ATM due to wrong check/bug in `if` statement.
If you do not want palettes to be used (and shown in UI) do not assign any character into this field i.e. `...;;...`
This will prevent unwanted behaviour in the future if the `if` statement changes.

I am sorry if this frustrates you and if you do not want to modify the code I am ok with that too.

netmindz: updated

netmindz:
The commit Id is listed so that we get repeatable builds @blazoncek

netmindz:
> > The commit Id is listed so that we get repeatable builds @blazoncek
> 
> I'm sorry, but I do not understand.
> 
> One more thing. Can you re-base this for 0_15 branch?

When defining any library dependencies in platformio it's always best to point at a specific release tag, rather than head of master/main as otherwise suddenly your own builds can start failing if some breaking change is pushed or it works for some folk, but not others as they have pulled different versions.

As I haven't created a release as my branch is actually a PR back to the main repo, I'm depending on a specific commit to provide the same level of build stability/repeatability

As for 0_15, yeah I can create a fresh PR based off that if you like.

Is there anything else that needs to be addressed or are you otherwise happy with this code?

blazoncek:
Thank you. If you consider it *finished* then I'm ok with it.
You can just pull 0_15 into your branch, push and then update target of this PR. No need to reopen new PR.

netmindz:
Specifically 0_15 or just pull in updates on dev?

blazoncek:
0_15 from Aircoookie repo. It is most recent with updates from a few seconds ago.

netmindz:
New PR created for fresh branch

Issue #3189, opened 2023-04-28
Feature request: Possibility to change preset on button press, not release


Sebbb:
**Is your feature request related to a problem? Please describe.**
I'm using WLED to control my bathroom light and to switch it on/off with a momentary switch. Right now, the light only toggles when I release the momentary switch (configured as "Pushbutton" in WLED), not when pressing it.

**Describe the solution you'd like**
I would like the LEDs to switch on/off when pressing the button down.

**Describe alternatives you've considered**
There is no good alternative except changing the source code (or accepting the current behavior) :)
I understand why the current implementation only triggers an action when releasing, and obviously this behavior should not break.
My idea is, when the same preset is configured for all actions (short, long, double), it makes no sense to wait for the release, and the action could trigger instantly. Another possibility could be an option to change the behavior.

**Additional context**
I already implemented it for me, but maybe it's useful for others as well.
See commit https://github.com/Sebbb/WLED/commit/d5e940a572b0edadea38f402d0ce7e3bd340bf5f
I didn't create a PR as I don't know if my idea how to solve it is good. My code probably isn't ;)

blazoncek:
Unfortunately in such case you do not know if double press is about to happen. Or if it is going to be long press.

If you've implemented it, please share your code.

Sebbb:
Hello,

that's the thing: If you configure the same preset for short/long/double press, it makes no difference and there is no need to wait. That's how I implemented it :)

Sebastian

blazoncek:
Make a PR and let's see if @Aircoookie approves.
IMO it is a viable compromise but will need to be documented in KB as well. Please prepare an update for WLED-docs too.

Sebbb:
As said, I don't think my code is very nice, so before putting time into documenting it, let's see if the overall idea is good (it's not very intuitive, a dedicated setting would be easier to understand), and if somebody can implement it in a less hackish way

blazoncek:
As you wish. I, personally, would love to see more contributions like this.
IMO it is not *hackish* that much if it is documented well.

If we'd add a checkmark/option to trigger on the "rise edge" as opposed to the "fall edge" then we would need to inform users that in such case long press and double press are no longer working. A worse situation IMO.

Sebbb:
Implemented with https://github.com/Aircoookie/WLED/pull/3226 - thanks :)

PR #3226, opened 2023-05-31
Buttons: Trigger on button press (instead of release) if all configured presets are the same

Sebbb:
Hello,

this addresses #3189. It works well for me, but I'm not sure if it doesn't maybe need additional debouncing code with some switches.

### DIFF
--- a/wled00/button.cpp
+++ b/wled00/button.cpp
@@ -257,6 +257,14 @@ void handleButton()
     //momentary button logic
     if (isButtonPressed(b)) { //pressed
 
+      // no need to wait for release or another press
+      if(macroButton[b] && macroButton[b] == macroLongPress[b] && macroButton[b] == macroDoublePress[b]) {
+        if(!buttonPressedBefore[b])
+          shortPressAction(b);
+        buttonPressedBefore[b] = true;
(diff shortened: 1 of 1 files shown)


blazoncek:
Thank you.
Would you be so kind and also add the info (PR) about same presets causing such behaviour to [WLED docs](https://github.com/aircoookie/WLED-docs)?

Sebbb:
Hi,

I never experienced issues with that, for me it works perfectly reliable. Maybe because that button check routine is only executed every few milliseconds (?) to check the button state and does not trigger an interrupt.

However, I'm definitely not a good programmer and probably wouldn't be able to add that, if it's needed :(

Sebastian

Aircoookie:
Added a slightly hacky debounce logic, should work fine though. 
On release it only reset `buttonPressedBefore` after 50 ms of uninterrupted released, and thus block further presses until the button has been released properly.

softhack007:
> Added a slightly hacky debounce logic,

Thanks @Aircoookie, I think this will do the trick 👍

Sebbb:
@Aircoookie I just compiled and tried your changes, works well :)

PR #3237, opened 2023-06-07
Support for ESP-NOW Wireless Remote Control

ctsims:
This changeset adds support for remote control of a WLED instance through the [ESP-NOW](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_now.html) protocol. 

Since it's a direct change to the central code rather than a usermod I'm adding as much detail as I can about why I think it's got broad positive value and the steps taken to try and fit in cleanly, but please let me know if there's anything else that would be helpful. 

## Motivation
ESP-NOW has a lot of desirable properties for external control. As a peer-to-peer protocol the remote doesn't need to be on a WIFI network at all, it doesn't require any additional hardware (as the protocol is supported on ESP8266 and ESP32 MCU's natively), and as noted in this [community enhancement request](https://github.com/Aircoookie/WLED/issues/2753) its speed and lack of protocol negotiation makes it very effective for low-power IoT buttons. 

## Design Approach
My implementation is easy to use with a quick home-grown button (will share a sketch when I get a chance) but is also built to support the default features of the 9 button "Wiz Connected" branded "WizMote" product because as it turns out these remotes are just ESP-NOW broadcasters with good antennas and well-engineered power latching (two AA batteries last for years). That gives WLED users access to a $10-$15 commercial option that is broadly available internationally and works out of the box. 

As an alternative I considered processing incoming esp-now payloads as JSON inputs rather than fixed button presses. That would be significantly more flexible, but the protocol has a very sharp limit to the payload capacity (250 bytes) so the added complexity didn't seem justified. It would be fairly straightforward to add new processing alternatives in the future.

## Implementation Approach

- Removable at compile-time by flag (`WLED_DISABLE_REMOTE`) 
- Disabled at runtime by default to prevent extra power draw and any hypothetical impact on concurrent WIFI usage, with a settings UI option to enable
- Tethers to a specific hardware sender with a user-facing scan to identify the appropriate MAC
image

- Uses the first 4 Presets for the scene selection buttons, following the same implementation pattern as the IR Remote feature

## Testing

- I've tested that ESP-NOW works concurrently with the WIFI on the ESP's that I have access to (D1 Mini clones and Feather Huzzah boards), and doesn't interfere with the other WIFI (There's a lot of misinformation that ESP-NOW is mutually exclusive with other wireless functions)
- I've tested the core remote features (including scene selection and the night dimmer) work out of the box
- Tested to confirm that precompile directives remove the feature, and that there's no accidental Serial / Debug output bleed

## User story 

Expected user journey for the feature
- Navigate to Settings
- Activate Remote feature
- Press a button on local wifi remote
- See the captured remote hardware address on settings page and copy/paste into settings to tether and allow its use
- Use remote to control on/off and brightness
- Add presets as needed to save light state for use with remote


wled00/presets.cpp
```diff
@@ -120,6 +120,15 @@ bool applyPreset(byte index, byte callMode)
   return true;
 }
 
+// apply preset or fallback to a effect and palette if it doesn't exist
+void presetFallback(uint8_t presetID, uint8_t effectID, uint8_t paletteID)
+{
+  applyPreset(presetID, CALL_MODE_BUTTON_PRESET);
```
blazoncek, line 126: IR is simulating button presses, moving function here is not simulating buttons. The CALL_MODE_BUTTON_PRESET is not appropriate in such case and should be adjusted when calling `presetFallback()`.

wled00/xml.cpp
```diff
@@ -328,13 +328,23 @@ void getSettingsJS(byte subPage, char* dest)
...
+    #endif
+
     #ifdef WLED_USE_ETHERNET
     sappend('v',SET_F("ETH"),ethernetType);
     #else
     //hide ethernet setting if not compiled in
     oappend(SET_F("document.getElementById('ethd').style.display='none';"));
     #endif
 
+    #ifdef WLED_DISABLE_REMOTE
```
blazoncek, line 343: Why not use #else from lines above (334)?

wled00/remote_codes.h
```diff
@@ -0,0 +1,17 @@
+
```
blazoncek, line 1: I would prefer this to be in the .cpp file as it will most likely only be used there.

wled00/remote.cpp
```diff
@@ -0,0 +1,174 @@
...
+  uint8_t byte8 = 1;    // Unknown, but always 0x01
+  uint8_t byte9 = 100;  // Unnkown, but always 0x64
+
+  uint8_t byte10;  // Unknown, maybe checksum
+  uint8_t byte11;  // Unknown, maybe checksum
+  uint8_t byte12;  // Unknown, maybe checksum
+  uint8_t byte13;  // Unknown, maybe checksum
+} message_structure;
+
+int esp_now_state = ESP_NOW_STATE_UNINIT;
```
blazoncek, line 25: You can make all internal globals `static`.

wled00/presets.cpp
ctsims: Makes total sense. Will also update to match the call signature from the other method more closely to make it easier to understand by inspection

wled00/xml.cpp
ctsims: Just oversight on my part, thanks for catching

blazoncek:
I did a read-up on ESP-NOW. Interesting and possibly very useful if implemented properly and fully.

wled00/presets.cpp
blazoncek: Looks like I was not clear enough. Sorry for that.

My thought was to just duplicate (or reuse) existing function from ir.cpp *or* make a universal function that may be usable in plethora of other situations (future-proof) by adding a call mode parameter instead of hard-coded CALL_MODE_BUTTON_PRESET.

ctsims:
Thanks for the prompt review and helpful feedback. I've updated the code in response.

I've provided answers to your questions inline below, but let me know if the purpose or fit isn't clear or doesn't seem aligned. 

> - who/what is sending ESP-NOW commands/is master?

The common use case here would be a second non-WLED ESP. 

[Here's a sketch](https://gist.github.com/ctsims/2e292a570222cf17922a6625bf9cfebc) for a two-button ESP remote which would be compatible with any WLED receiver in range (200ft or so) once the WLED light is set in the HTML settings page to accept signals from it. 

This sketch doesn't really clearly demonstrate it, but the most important value add is that since ESP-NOW doesn't require
actually getting on the wifi network you can latch the button-press to reset and fully power down the entire chip between presses. It takes an incredibly small window to spin up from power-off and send an ESP-NOW message (compared to getting on network and sending UDP),  and it's way less subject to packet loss from network saturation like UDP. 

This gels with the approach of lower power ESP projects like [Picoclick](https://makermoekoe.gitbook.io/picoclick-c3/) which are super useful with ESP-NOW but not super useful if the board needs to make a full connection to blast UDP.  This setup is how the Wizmote I linked above, and have tested with locally, can last for years on batteries while also working ~instantaneously. 

The other nice value add which I can only speak to theoretically is the ability for an external ESP remote to talk to WLED lights which are in AP mode without needing finnicky firmware configuration. From the remote's perspective there's no difference to a receiver being connected to another network or being in AP mode other than the channel (the sketch just broadcasts the message packets across all 13). I'm honestly not sure if that's a use case WLED is particularly interested in, though. 

**FYI** - One thing that's worth noting about the external remote use case is that WIFI sleep does affect the ESP's ability to receive ESP-NOW messages, so if you are testing it is good to set the WLED light to disable sleep. It's possible to work around that limitation without putting WLED in no-sleep mode by having the remote do more heavy lifting (since the messages are sequenced and idempotent there's little risk from redundancy), but I didn't want to confuse the sketch by including more detail than necessary. 

> - is there an option for one device to be master to all others?

Yes, there are a few options for how this works. ESP-NOW can direct packets to specific receivers, or it can just do a wide broadcast. 

Multiple WLED lights could be set to listen to signals from the Esp remote sketch I linked above. 

- can you daisy-chain multiple devices/make a mesh?

Yes, it's definitely possible for ESP-NOW to support that (the ESP can be in receiver / sender mode simultaneously and this is a supported use case from espressif), but it would be tricky nuanced work to incorporate it into the current sync protocols used by WLED because the packet payload size is 10-100x smaller than UDP's.

wled00/presets.cpp
ctsims: If I'm understanding you correctly I think you were clear and my change was to take the latter approach based on your feedback. 

I added a universal version of the [preset with fallback](https://github.com/Aircoookie/WLED/pull/3237/files#diff-2659ccd7699e79b87d515cd0b1676e2a25c7dd1959fafa352620805b2b4e8ed0R124) function that accepts a `callMode`, I just changed the signature a bit so that the new method name (`applyPresetWithFallback`) and argument order would more closely match the existing universal `applyPreset` method compared to my initial code.

Apologies if I'm still misinterpreting, though, and happy to do another pass.

blazoncek:
Thank you very much for enlightening me! ❤️ 

I know I am asking a lot, but would you care to implement master (broadcast) as well?

For use with regular (or instead of) UDP syncing. A single UDP sync packet is about 40+ bytes long (but depends on number of segments) so the payload (with only a few segments set up) would fit into a single ESP-NOW packet which is acceptable.
You can reuse `notify()` function by adding an option in Sync settings to enable ESP-NOW broadcasts. With handling similar to `handleNotifications()` on the receiving side.

It would be acceptable to skip ESP-NOW broadcasts if number of segments is such that packet breaks the 250 bytes barrier.

wled00/presets.cpp
blazoncek: Indeed. My bad.
You can remove the 3 lines in ir.cpp that are no longer needed in `presetFalback()` or entirely remove `presetFalback()` replacing other code as necessary.

blazoncek (approved):

softhack007:
Did you test on esp32 platform?
I see that esp32 build are failing with `wled00/wled.h:111:22: fatal error: espnow.h: No such file or directory`.
In case that ESP-NOW requires arduino.esp32 2.0.x, if can only be enabled for our experimental "V4" environment.

What about esp32-s3/-s2/-c3? Will it work with them, too?

wled00/wled.h
```diff
@@ -105,6 +107,9 @@
 #endif
 #include <WiFiUdp.h>
 #include <DNSServer.h>
+#ifndef WLED_DISABLE_REMOTE
+  #include <espnow.h>
+#endif
```
softhack007, line 112: Seems that esp32 needs a different header

https://github.com/espressif/arduino-esp32/blob/1.0.6/libraries/ESP32/examples/ESPNow/Basic/Master/Master.ino

```c++
#include <esp_now.h>
```
You could include the correct header for each platform this way: 
```c++
#ifdef ARDUINO_ARCH_ESP32
  #include <esp_now.h>
#else
  #include <espnow.h>
#endif

```

softhack007 (commented):
We should make sure that the code does not break any of our standard build for esp32.
Also can you confirm that testing with esp32 showed the code works?

ctsims:
@softhack007 - Thanks for the review and feedback. I had missed the esp32 path. ESP-NOW is definitely supported across all ESP32 device series, but in addition to the library header they use a newer version of the interface code. The change is fairly minor for the receiver, and I've implemented the new signatures and made sure that the ESP32 builds work when the remote is enabled. 

**Question:** I'll be honest that I don't have an ESP32 on hand to test with, although this is a great excuse for me to finally get one. I've heard that ESP32 support is experimental currently - should I anticipate that an arbitrary ESP32-C3 will work with `main` or is it best to test with specific boards?

It will take me a bit of time to get setup to test. For now I've configured the ESP32 build mode to disable the remote flag at compilation to make sure it doesn't interfere with anything until I've had a chance to test further, but let me know if there are other or better mitigations to that risk.

ctsims:
@blazoncek I'll check out the sync system and see how it looks. I'm assuming since there's a UDP broadcast that I wouldn't really need to handle any of the sequencing / preventing redundant propagation, etc, from scratch and it's just making the new channel?

blazoncek:
@ctsims there is no need to care about redundant calls as there are none. UDP sync only publishes (`notify()`) when there is a change in WLED state. What you can do is hook into `notify()`.
As for ESP32 variants: vanilla, C3, S2 and S3 are supported and have appropriate entries in platformio.ini. C3, S2 and S3 may require additional tweaking of build parameters if you have an exotic development board as they are still in experimental state.
If you want mature platform choose plain ESP32.

wled00/ir.cpp
```diff
@@ -70,10 +70,9 @@ void decBrightness()
   }
 }
 
-// apply preset or fallback to a effect and palette if it doesn't exist
 void presetFallback(uint8_t presetID, uint8_t effectID, uint8_t paletteID)
```
Aircoookie, line 73: As this is now just a duplicate of `applyPresetWithFallback()`, we should remove it

Aircoookie:
My Wizmote arrived today and I can confirm this to work perfectly on both ESP8266 and ESP32 platforms!
WiFi performance seems unaffected too, except for a 2-3 sec extra delay when saving WiFi settings, will check later today if I can work out why that happens. Other than that, this is ready to merge! Since it is separate functionality, I'd welcome WLED sync over ESP-NOW as a separate PR :)

softhack007 (approved):
Ok for me, as @Aircoookie confirmed it does work for esp32.

Actually I find the build flag too generic: `-D WLED_DISABLE_REMOTE`. A "remote" could be anything, but in this case the remote seems to be just one usecase.

The PR is actually about adding support for ESP- NOW, so how about naming the flag `-D WLED_DISABLE_ESPNOW` ??

wled00/fcn_declare.h
```diff
@@ -200,7 +200,7 @@ void serializePlaylist(JsonObject obj);
 void initPresetsFile();
 void handlePresets();
 bool applyPreset(byte index, byte callMode = CALL_MODE_DIRECT_CHANGE);
-void applyPresetWithFallback(uint8_t presetID, uint8_t effectID, uint8_t paletteID, uint8_t callMode);
+void applyPresetWithFallback(uint8_t presetID, uint8_t callMode, uint8_t effectID, uint8_t paletteID);
```
blazoncek, line 203: I would add `effectID=0` and `paletteID=0` as a default values;

blazoncek:
> The PR is actually about adding support for ESP- NOW, so how about naming the flag `-D WLED_DISABLE_ESPNOW` ??

Totally agree on this.

ctsims:
@Aircoookie Sorry for the delay getting back, haven't had a chance to jump back in here. Thanks for the feedback and the review! Will take a look at the esp-now syncing when I get a chance. Have really been struggling with UDP lately on my network, so it seems like a great fit. 

One quick question - What's the best way to document the [remote sender sketch](https://gist.github.com/ctsims/2e292a570222cf17922a6625bf9cfebc) or something like it for the long term? Commercial hardware compatibility seems to be a strength of the WLED project, but would love for folks to have a clear pattern for DIY open source access. Is cleaning up something into another codebase and providing a link for the docs helpful, or is that too transient / non-canonical for where you're trying to keep the reference materials?

Aircoookie:
@ctsims even just a link to that gist added to kno.wled.ge would be fine. (perhaps we should add a new ESP-NOW page to the interfaces section)

dosipod:
@ctsims  Hey , I do not have remote but used the ESP-NOW-remote-transmitter example here  https://www.instructables.com/ESP-NOW-Remote-Control/ .
I can see it is connected correctly to wled so that is very nice  idea . As you can see we are trying to use a joystick as we have both the transmitter and receiver  sketches  and we actually already built a wled robot ( using PWM servos usermod)  , you can guess where is that going   . We are only really do Arduino so our C++ is very basic  with trial and error  and I want to know where to add the receiver   code in wled and please do not say add usermod as we only know how to mess directly with the code and as this is basically for learning we do not mind crashes or issues ..etc  
image

dosipod:
And one more question , which lib to use 
https://github.com/espressif/esp-idf/blob/master/components/esp_wifi/include/esp_now.h
https://github.com/esp8266/Arduino/blob/master/tools/sdk/include/espnow.h

dosipod:
When i tested with 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF  with t https://gist.github.com/ctsims/2e292a570222cf17922a6625bf9cfebc ,  i was able to control all the units around me with only one unit having remote enabled , that could be controlled with sync groups i guess  but that is really very practical use case

blazoncek:
@ctsims I have a follow-up on this PR.

I am trying to implement ESP-NOW syncing for WLED devices in a similar fashion as current UDP sync.
I have discovered a very useful ESP-NOW library [QuickESPNow](https://github.com/gmag11/QuickESPNow) and have successfully implemented multipacket syncing with it but unfortunately it clashes with your implementation (overwrites callback function).

Would you care to look into QuickESPNow and the possibility to implement remote handling using it?

BTW I have moved ESP-NOW initialisation into `initInterfaces()` as it is most appropriate there.

ctsims:
@blazoncek Sure! Will take a look, thanks for the heads up

ctsims:
@blazoncek Confirmed that the existing functionality works just fine with the QuickESPNow functionality in [this commit](https://github.com/ctsims/WLED/commit/119ce17e856e11c8a97a408c0956d83c19a88fc6). 

What's the best way to work through getting things in one place? Want me to PR into your branch and test with everything in place?

blazoncek:
That is great to hear! I was afraid it would be more difficult to solve.
I'll try to polish ESP-NOW syncing and then use (or extend) callback function to handle remote and sync packets. I'll let you know when I'm close to finish so we can join both things.

BTW there was an issue opened regarding "Moon" button on the remote. Can you take a look into it?

ctsims:
Yeah, I definitely ran cross some tricky edges on versioning and interfaces with ESP-NOW, so glad it turned out to be drop-in.

I think the only tricky question would be how to do dispatch between the two callback methods and determine that an incoming message should trigger each individual callback. We could have each implementing functionality receive each packet and make its own decision about whether that packet is for it, or have a central function that knows specifically where to direct each incoming payload.

The default way I'd multiplex is just the length of the ESP-NOW packet payload, since the WizMote functionality doesn't have control of its own payload interface. Just make sure that the sync function (or any other WLED inbound packets) have a consistently different length than the remote packets, and use the payload length to determine which sub-function should get the ball. 

If helpful I'd be happy to get that all worked out once you've got the other side of things done, so feel free to just implement your side as if the remote isn't there, if most useful, and I can jump in on getting both sides plugged together.

(And confirming that I'll respond on the query for the sleep button, thanks for mentioning it.)

lanshicai:
Hello, can I use another ESP32 to send a simple message switching to WLED through ESP-NOW, for example:{"seg":[{"fx":76,"pal":1}]} .

blazoncek:
No. JSON API over ESP-NOW is not supported.

insane2subro:
I made a small sketch for ESP32 that can handle MQTT-ESPNOW and works with WLED to be controlled via HomeAssistant.
[](https://github.com/insane2subro/ESPNOW2MQTT)

Issue #3236, opened 2023-06-07
suspicious code in e131.cpp (detect RGB / RGBW)


softhack007:
This line is a bit suspicious, and I doubt it will work as intended (from https://github.com/Aircoookie/WLED/pull/3193)

https://github.com/Aircoookie/WLED/blob/999bec19f1bf37b344177a9aeffc752c5f4ae472/wled00/e131.cpp#L28

Compiler warning: 
````
wled00/e131.cpp:29:57: warning: suggest parentheses around comparison in operand of '&' [-Wparentheses]
   uint8_t ddpChannelsPerLed = (p->dataType & 0b00111000 == 0b011) ? 4 : 3; // data type 0x1B (formerly 0x1A) is RGBW (type 3, 8 bit/channel)
                                              ~~~~~~~~~~~^~~~~~~~
````
If it is intended as `((p->dataType & 0b00111000) == 0b011)`, then the decision cannot become true. Any value & 0b00111000 will not have the lowest 3 bits sets, as they are masked out by "&".

Actually I'm not sure how it was indented to work, maybe the original author could take a look?

blazoncek:
It has been fixed in `beta-3` branch.

Aircoookie:
Embarrassing 😅 Well, mistakes happen, thanks for catching it @softhack007 and fixing it promptly @blazoncek :)

Issue #1177, opened 2020-09-12
HTTPS support for HTTP request API


SpicerSolutions:
Is HTTPS likely to be support on the HTTP request endpoint?

I appreciate this might be difficult for people with expiring SSL certificates.

Aircoookie:
Hi!
This is not likely/almost impossible to implement on ESP8266 due to insufficient memory.
On the ESP32 it is a different story, HTTPS support is easily possible. Still we would need to figure out how to add a certificate that is ideally both not self-signed (so that it is trusted by browsers), but that could not be circumvented by the private key becoming public due to the open source of WLED. This would likely require users to generate/install their own certificates for true security, but it's definitely something I want to look into!

pbolduc:
@SpicerSolutions  You can hear `Unexpected Maker` talk about the issues with https on this video: https://youtu.be/A4Hzz84HuNY?t=733

Fonta:
Could use a selg-signed certificate by default. In the config you could then replace it with your own. That's how most devices work.

blazoncek:
This can be supported using reverse proxy which will handle HTTPS (and WSS) with recent modifications in 0.14.0-b3+.

PR #3238, opened 2023-06-08
Beta 3

### LINKED ISSUE #3069: Settings page not working on reverse proxy
### What happened?

I'm building a standalone server with Raspberry Pi that exposes the WLED interface over wifi, with a QuinLED board connected to the Pi over ethernet (to stream LED data realtime). 

When setting up a reverse proxy with nginx on the Pi I can forward all traffic to WLED and it shows the interface in the browser. But when navigating to the `/settings` page, it never succeeds in getting back a request from WLED and the connection times out after a minute or two. 

I've inspected the HTTP headers and it seems the `/settings` page is returning some different HTTP headers than the main `/` page. Specifically, it's also using gzip encoding, but it never returns the `content-encoding: gzip` and `content-length: xx` headers. It does return a `Transfer-Encoding: chunked` header.

### To Reproduce Bug

Create an nginx proxy to forward all traffic to the WLED server:

`
server {
    listen 3007;
    location / {
        proxy_pass http://{{WLED IP}};
    }
}
`

And then do a request to the proxy server `/settings` page. It never resolves.

### Expected Behavior

I expect the settings page to also work through a proxy server.

### Install Method

Binary from WLED.me

### What version of WLED?

WLED 0.13.1

ESP32

### Relevant log/trace output

_No response_

### Anything else?

I'm running WLED pre-installed on a QuinLED Dig Uno with an Ethernet hat.

### Code of Conduct

### LINKED ISSUE #3204: JSON APIs assigned to buttons  , e.g  {"ps":"1~4~"}
### What happened?

We normally used {"ps":"1~4~"}  in a preset and that will work in ui but  not when assigned to buttons it does not work    only   win&PL=~&P1=1&P2=4 works from both ui and when assigned to buttons.

I am not sure that is limited to the above api as assigning other api to buttons like  {"on":"t"} will  work . I will try to look for more if that is only limited to this 

### To Reproduce Bug

Add  {"ps":"1~4~"}  to preset and assign to button action  ( clicking on preset will work but button action will not work ) 
Add  win&PL=~&P1=1&P2=4  to preset and assign to button action  ( clicking on preset will from ui and button action  ) 

### Expected Behavior

 {"ps":"1~4~"}    and win&PL=~&P1=1&P2=4  is expected to behave the same 

### Install Method

Self-Compiled

### What version of WLED?

WLED 0.14.0-b2 (build 2305090)

ESP32

### Relevant log/trace output

```shell
More  more info on the same  are here  https://wled.discourse.group/t/physical-button/8289
and here https://discord.com/channels/473448917040758787/473448917543944193/1108240349115191386
```
The format is messed up in the post because of the use of json  so when you see {"ps":"1~4~"} it is like the picture 
image

### Anything else?

_No response_

### Code of Conduct

blazoncek:
Adds the following:
-   Bumped version to 0.14-b3 (beta 3)
-   Reverse proxy support (replaces #2788 and fixes #3069 and #1177)
-   PCF8754 support for Rotary encoder (requires wiring INT pin to ESP GPIO)
-   Rely on global I2C pins for all usermods (BREAKING CHANGE; removed individual I2C initialisations and pin set up)
-   adds better UX selecting GPIO pins in settings
-   various fixes and enhancements (fixes #3204 #3236)


Aircoookie (approved):
Ready to merge :)

`NeoGammaWLEDMethod` might be nice to have as an option, but right now I don't see any benefit to calling it from NPB. The current approach of pre-calculating gamma is optimal performance wise and has no drawbacks in my opinion.

PR #3253, opened 2023-06-19
Added NetCube Systems Universal Light Controller

Craft4Cube:
Added board definition for NetCube Systems Austria - Universal Light Controller and added Usermod for Battery Management System for above board.


usermods/ULC_Battery_Management/usermod_ulc_batterymanagement.h
```diff
@@ -0,0 +1,279 @@
...
+
+#ifndef USERMOD_BATTERY_MAX_VOLTAGE
+  #define USERMOD_BATTERY_MAX_VOLTAGE 4.2f
+#endif
+
+#ifndef USERMOD_IP5306_SDA
+  #define USERMOD_IP5306_SDA 32
+#endif
+
+#ifndef USERMOD_IP5306_SCL
```
blazoncek, line 40: I2C bus is shared among all usermods in 0.14 since #3238.
Please update code to use global I2C pins.

usermods/ULC_Battery_Management/usermod_ulc_batterymanagement.h
```diff
@@ -0,0 +1,279 @@
...
+#endif
+
+class UsermodULCBatteryManagement : public Usermod {
+
+  private:
+    IP5306 *ip5306 = nullptr;
+
+    bool initDone = false;
+    bool initializing = true;
+    // GPIO pin used for battery measurment (with a default compile-time fallback)
```
blazoncek, line 59: I2C bus is shared among all usermods in 0.14 since #3238.
Please update code to use global I2C pins. There is no need to store I2C pins for usermod unless it clashes with other I2C devices.
In such case use `Wire1` or secondary I2C bus.

usermods/ULC_Battery_Management/usermod_ulc_batterymanagement.h
```diff
@@ -0,0 +1,279 @@
...
+      }
+      x = y / pow(10, dec);
+      return x;
+    }
+
+  public:
+
+    void setup() {
+      DEBUG_PRINTLN(F("Allocating ip5306 pins..."));
+      PinManagerPinType pins[2] = { { ip5306_sda, true }, { ip5306_scl, true } };
```
blazoncek, line 109: When using global I2C there is no need to allocate pins.

usermods/ULC_Battery_Management/usermod_ulc_batterymanagement.h
```diff
@@ -0,0 +1,279 @@
...
+
+    // strings to reduce flash memory usage (used more than twice)
+    static const char _name[];
+    static const char _readInterval[];
+    static const char _maxCurrentBattery[];
+    static const char _maxCurrentExt[];
+
+    // custom map function
+    // https://forum.arduino.cc/t/floating-point-using-map-function/348113/2
+    double mapf(double x, double in_min, double in_max, double out_min, double out_max)
```
blazoncek, line 87: These functions are inlined by compiler. Please move them out of class declaration to reduce code size.

blazoncek (changes requested):
Thanks for contributing.
I would require you to update usermod to use global I2C pins instead of local.

There was also a discussion opened by @softhack007 regarding adding new board/controller build environments to platformio.ini. It is already crammed and unmaintainable as is.

platformio.ini
```diff
@@ -779,3 +780,15 @@ lib_deps =
...
 board_build.partitions = ${esp32.default_partitions}
+
+
+# ------------------------------------------------------------------------------
+# NetCube Systems Universal Light Controller
+# ------------------------------------------------------------------------------
+[env:esp32_ulc]
+board = esp32-poe
+build_unflags = ${common.build_unflags}
+build_flags = ${common.build_flags_esp32} -D WLED_RELEASE_NAME=ESP32_ULC -D RLYPIN=-1 -D WLED_USE_ETHERNET -D WLED_ETH_DEFAULT=6 -D LEDPIN=14 -D DEFAULT_LED_COUNT=37 -D APL_MILLIAMPS_DEFAULT=65000 -D BTNPIN=-1 -D WLED_DISABLE_BLYNK -D IRPIN=-1 -D USERMOD_ULC_BATTERYMANAGEMENT
```
softhack007, line 791: > -D APL_MILLIAMPS_DEFAULT=65000
why do you need this? It could be dangerous to set 65 Ampere as current limit.

usermods/ULC_Battery_Management/usermod_ulc_batterymanagement.h
softhack007: @blazoncek I think we should add the map functions for float and double somewhere in a generic header.

It seems like we are starting to collect copies of this in several usermods already. Especially the variants of "mapf" which have a division-by-zero problem.

usermods/ULC_Battery_Management/usermod_ulc_batterymanagement.h
```diff
@@ -0,0 +1,279 @@
...
+    static const char _maxCurrentExt[];
+
+    // custom map function
+    // https://forum.arduino.cc/t/floating-point-using-map-function/348113/2
+    double mapf(double x, double in_min, double in_max, double out_min, double out_max) 
+    {
+      return (x - in_min) * (out_max - out_min) / (in_max - in_min) + out_min;
+    }
+
+    float truncate(float val, byte dec)
```
softhack007, line 92: this method to "truncate" a float to a number of decimals is _horribly_ inefficient due to the use of `pow()`. Please try to replace it.

usermods/ULC_Battery_Management/usermod_ulc_batterymanagement.h
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+  public:
+
+    void setup() {
+      DEBUG_PRINTLN(F("Allocating ip5306 pins..."));
+      PinManagerPinType pins[2] = { { ip5306_sda, true }, { ip5306_scl, true } };
+      if (pinManager.allocateMultiplePins(pins, 2, PinOwner::HW_I2C))
+      {
+        ip5306 = new IP5306(ip5306_sda, ip5306_scl);
```
softhack007, line 112: For I2C bus sharing, its very important that all usermods rely on the global `Wire` object.

Please check if there is a variant of the IP5306 constructor that takes a reference to `Wire`, instead of getting SDA and SCL pins as parameters.

usermods/ULC_Battery_Management/IP5306_I2C.h
```diff
@@ -0,0 +1,584 @@
+/*@brief  library to interact with IP5306 over I2C communication
```
softhack007, line 1: Instead of using a copy of the driver library, could you replace this file with a reference to library on github?

Something like 
```
lib_deps = .... 
    github_user/IP5306 @ 0.1.23
```

this would be a better solution, because we would avoid having a full copy of the driver - that might become outdated in a year or two - as part of WLED.

Craft4Cube: The library is modified to work with the IP5306 used here. There is no library on github for it as of now.

platformio.ini
Craft4Cube: This was left in from testing by accident. It is now set to 2.5A as default for the board.

usermods/ULC_Battery_Management/usermod_ulc_batterymanagement.h
Craft4Cube: I have moved the mapf function outside the main class

usermods/ULC_Battery_Management/usermod_ulc_batterymanagement.h
Craft4Cube: I seem to have forgotten to remove this function as I do not need it anymore, anyways. I have removed it now

softhack007:
> platformio.ini. It is already crammed and unmaintainable as is.

Fully agreed. I think its better to put a prepared platformio_override.ini into the usermod folder. Just don't add directly to platformio.ini.

usermods/ULC_Battery_Management/usermod_ulc_batterymanagement.h
softhack007: You can take a look at the BME280 usermod as an example.
https://github.com/Aircoookie/WLED/blob/c04c73bbd7f448312c368e33773e510d4e8324cf/usermods/BME280_v2/usermod_bme280.h#L187-L195

- remove I2C pins from the usermod settings
- use global i2c_sda and i2c_sda if you need them in your code - read only!
- Leave out Wire.begin() as `Wire` is already initialized for you by WLED
- check for `i2c_sda < 0 || i2c_scl < 0` and disable the usermod in case the global pins are undefined (-1)


@blazoncek do you know a simpler example that shows the new concept?

usermods/ULC_Battery_Management/usermods_list.cpp
softhack007, line 1: is this a commit accident? You should not need `usermods_list.cpp` inside your usermod folder.

Craft4Cube: Seems to be by accident. I have removed it.

usermods/ULC_Battery_Management/usermod_ulc_batterymanagement.h
Craft4Cube: I have switched to the new i2c object now. However i have the issue that i cannot set default i2c pins for the board in the platformio.ini. If i use HW_PIN_SDA it does not work at all (and the pin in usermods is "unused") if i use I2CSDAPIN it shows in the usermods but only works after saving the settings again.

All testing was done with a factory reset of wled

blazoncek: Checked the code and it looks like not all instances of I2C bus were correctly updated. My mistake.
Actually there is no need to allocate any pins as they are allocated by WLED core at boot.

A usermod should only check if `i2c_sda` and `i2c_scl` are not equal to -1. In such case I2C interface is initialised and `Wire` global started.

usermods/ULC_Battery_Management/usermod_ulc_batterymanagement.h
blazoncek: @softhack007 agreed. Best added to `util.cpp` (and `fcn_declare.h`) and then removed elsewhere.

usermods/ULC_Battery_Management/usermod_ulc_batterymanagement.h
Craft4Cube: I think we are talking about seperate things. I need to set the i2c pins for my board in the target configuration. But it does not use the values I give in the target configuration

blazoncek: > I have switched to the new i2c object now. However i have the issue that i cannot set default i2c pins for the board in the platformio.ini. If i use HW_PIN_SDA it does not work at all (and the pin in usermods is "unused") if i use I2CSDAPIN it shows in the usermods but only works after saving the settings again.

If you want to allocate and assign (as opposed to only define) I2C pins on fresh install of WLED, then use `-D I2CSDAPIN=X` and `-D I2CSCLPIN=Y` if you just need to define them then use `-D HW_PIN_SDA=X` and `-D HW_PIN_SCL=Y`.
Defining pins is only informational to the user (placeholder text in UI). If you use I2CSDAPIN and I2CSCLPIN then those will overwrite HW_PIN_SDA & HW_PIN_SCL.

This is the relevant code executed at boot which will load compiled in defaults (I2CxxxPIN).
Screenshot 2023-06-20 at 22 29 30

Craft4Cube: Okay, yes using I2CSDA/SCLPIN works. But I still need to go to the Settings page and save once for it to actually work

blazoncek: I have hard time believing that. 😀

Craft4Cube: Well then, how do I make you believe, or do you have any other Idea why it does this?

Craft4Cube:
I think everything is fixed now. Anything left to do?

blazoncek (approved):

blazoncek:
Looks Ok from my side.

Issue #3260, opened 2023-06-22
Playlist cycling via presets is broken in 0.14.3


alackmann:
### What happened?

In my 0.13.x installation of WLED, you can create a preset with the API command similar to `P1=20&P2=22&PL=~` that when triggered will cycle through playlists 20,21,22 and then back to 20. These can be triggered either from the WebUI by pressing the preset OR from an external button macro (among others).

In all the 0.14.x versions I've tried, this is no longer working. Instead it repeatedly triggers the first playlist in the list. So in the example above, the playlist in preset 20.

### To Reproduce Bug

* Create 3 new playlists via the UI. In this example using IDs 20, 21 & 22 for the presets. Put different presets within each playlist.
* Create a preset (can be any ID, but let's call it 3 in this example) that has the API command `P1=20&P2=22&PL=~` as the preset content
* Using the web interface, click on the new preset 3
* Playlist 20 will play
* Click playlist 3 again
* Playlist 20 will restart again

### Expected Behavior

* On the second click, playlist 21 should play. 
* On the third click, playlist 22 should play
* On the fourth, the cycle should restart and playlist 20 should play

### Install Method

Self-Compiled

### What version of WLED?

WLED 0.14.0-b3 (build 2306210)

### Which microcontroller/board are you seeing the problem on?

ESP32

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

blazoncek:
Please try JSON API instead of HTTP API.
I.e. `{"ps":"20~22~"}`

alackmann:
Neither were working @blazoncek. My PR fixes both.

blazoncek:
It breaks other (legacy) things.

PR #3262, opened 2023-06-22
Removed current cycle assignment that clears current value

alackmann:
`presetCycCurr` appears to store the current location in the preset cycle when using presets that cycle through playlists. Do this assignment here currently sets this value to the currently playing preset when it should have the last selection playlist.

Solves #3260

### DIFF
--- a/wled00/presets.cpp
+++ b/wled00/presets.cpp
@@ -180,7 +180,7 @@ void handlePresets()
       fdo.remove("ps"); // remove load request for presets to prevent recursive crash (if not called by button and contains preset cycling string "1~5~")
     deserializeState(fdo, CALL_MODE_NO_NOTIFY, tmpPreset); // may change presetToApply by calling applyPreset()
   }
-  if (!errorFlag && tmpPreset < 255 && changePreset) presetCycCurr = currentPreset = tmpPreset;
+  if (!errorFlag && tmpPreset < 255 && changePreset) currentPreset = tmpPreset;
 
   #if defined(ARDUINO_ARCH_ESP32)
   //Aircoookie recommended not to delete buffer


alackmann:
Also confirmed that the same preset / command using the JSON API works as expected. eg. `{"ps":"20~22~"}`

blazoncek:
I've checked the code and, although not very clear, it appears that `presetCycCurr` needs to be set together with `currentPreset` in `handlePreset()` for HTTP API **preset** cycling to function properly (legacy).
So I am closing this PR without merging.

alackmann:
@blazoncek what is an example of *preset* (vs playlist) cycling? Is there a test scenario for both I could check? Certainly **playlist** cycling is broken in this latest release. I was confused tho - as checking git history `presetCycCurr` has been set to `currentPreset` for sometime. So something changed around it to break the cycling behaviour.

blazoncek:
As explained elsewhere, although a playlist *is technically a preset* it is specially crafted preset that has completely different workflow.
I also noted elsewhere that since 0.14 presets are handle asynchronously to prevent file system corruption.

blazoncek:
BTW *playlist* cycling was never envisioned nor tested. Similarly as nesting playlists (which is forbidden and not possible form UI).

alackmann:
So are you saying the current code works as long as the presets are NOT playlists? I guess I'm confused at 0.13.x works fine with playlist presets. 0.14.x seems to be a regression. The example and test I've done isn't nesting of playlists at all.

alackmann:
@blazoncek when you have a moment, can you explain the legacy behaviour this PR breaks? I've tested standard preset cycling using both HTTP and JSON api presets and they continue to work correctly. If you can outline what I've broken I can look into fixing both behaviours.

blazoncek:
TL:DR; `presetCycCurr` is a legacy parameter needed for proper preset cycling using HTTP API. Playlist is a preset that calls other presets internally and does not update `presetCycCurr`. So this is done in `handlePreset()`.
I did some tracing and if it is possible I will provide *correct* fix.

Use Git blame to trace the original commit.

blazoncek:
Looks like the mitigation was already implemented elsewhere so the PR is valid. Reopened.

Issue #3250, opened 2023-06-14
Peek fails to load if websockets connection can't be established


blazoncek:
### What happened?

If WLED is compiled with Websockets the Peek LED output preview fails to load if websocket connection can't be established for some reason (proxy, firewall, etc). It should gracefully fall-back to non Websocket implementation instead.

### To Reproduce Bug

Add a firewall or a reverse proxy which will block websockets between WLED and client.

### Expected Behavior

Peek should work with or without Websockes, regardless of the way how WLED was compiled (with or without Websockets).

### Install Method

Self-Compiled

### What version of WLED?

0.14.0-b3 2206130

### Which microcontroller/board are you seeing the problem on?

ESP8266, ESP32, ESP32-S2, ESP32-C3

### Relevant log/trace output

```shell
N/A
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

netmindz:
So if you compile without there is an alternative implementation already? Long polling?
If you make the appropriate non-ws call to a WS enabled build does it work? I.e is it just the front-end code that would need to change?

blazoncek:
Both will need an update.
And yes, if you disable websockets, peek works without.

netmindz:
Not sure how to best detect the failure, I've seen poorly written firewalls allow the websocket to be negotiated, but then block any actual messages, so protocol switch is not always enough to prove success

blazoncek:
IMO fairly easy. When WS is not established or times out fallback to HTTP.
Easy in GUI more work in backend.

netmindz:
My point is that sometimes it can report as established, but isn't really, so would require timeout for an actual message, not just the http and protocol switch

blazoncek:
Partially solved by https://github.com/Aircoookie/WLED/commit/f015227fc8aa34cd511d8b1c9d91cd0a5c5c9067
Still missing 2D peek.

blazoncek:
I have asked @ewoudwijma to look into implementing 2D peek using HTTP JSON requests.

blazoncek:
Not going to pursue this any further.

PR #3267, opened 2023-06-23
Missing WS connection fallback for liveview

blazoncek:
- fix for #3250

Remove (conditional WLED_ENABLE_LEGACY) legacy URI

Replace /sliders with /?sliders (easier reverse proxy handling)


Aircoookie (approved):
The fallback did not work as `/json/live` was not compiled in by default. Also merged the two liveview pages, which saves 0.5kB and thus fully offsets the added binary size of enabling `/json/live`. Now ready to merge IMO!

blazoncek:
There's one more thing I noticed while testing this while I was at work: WS (from browser) keeps trying to connect. Perhaps we could add a counter to stop retries after a few attempts and only repeat 1 try after each HTTP request.
What do you think?

wled00/data/index.js
```diff
@@ -1697,7 +1698,7 @@ function toggleLiveview()
 	}
 
 	gId(lvID).style.display = (isLv) ? "block":"none";
-	gId(lvID).src = (isLv) ? getURL("/" + lvID + ((wsOn) ? "?ws":"")):"about:blank";
+	gId(lvID).src = (isLv) ? getURL("/" + lvID):"about:blank";
```
Aircoookie: Why did you remove this? liveview uses the `?ws` parameter to determine if it can use WS, without it it will always fallback to json.

Aircoookie:
WS retry looks good. Also added an attempt reset when the WS connection is opened, since preset loading only happens on page load or after presets were modified.

wled00/data/index.js
blazoncek: I am sorry. I missed the "?" in the string.

PR #3270, opened 2023-06-25
Add WireGuard VPN usermod

acvigue:
Adds a WireGuard usermod for ESP32 that allows WLED devices to join a remote network in situations where wireless isolation prevents local control.


acvigue:
Is there any chance someone could take a quick look at this?

Thank you all!

usermods/wireguard/wireguard.h
```diff
@@ -0,0 +1,102 @@
...
+            } else {
+                uiDomString = "netif down :(";
+            }
+        }
+        if(is_enabled) infoArr.add(uiDomString);
+    }
+
+    void addToConfig(JsonObject& root) {
+        JsonObject top = root.createNestedObject(F("WireGuard VPN"));
+        top[F("Endpoint Address")] = endpoint_address;
```
blazoncek, line 56: Please avoid key names with spaces and make them short if possible.
You can use utility function `addInfo()` to make UI better.

usermods/wireguard/wireguard.h
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        return configComplete;
+    }
+
+    uint16_t getId() { return USERMOD_ID_WIREGUARD; }
+
+   private:
+    WireGuard wg;
+    String preshared_key;  // [Interface] PrivateKey
```
blazoncek, line 93: Why `String`? You are not doing much processing on strings so it may be more efficient to use `char[]` or `char*`.

usermods/wireguard/wireguard.h
```diff
@@ -0,0 +1,102 @@
...
+        if (!is_enabled || !getLocalTime(&timeinfo, 0)) {
+            return;
+        }
+
+        if (wg.is_initialized() || !WLED_CONNECTED) {
+            return;
+        }
+
+        if (preshared_key.length() < 5) {
+            wg.begin(local_ip, private_key.c_str(), endpoint_address.c_str(), public_key.c_str(), endpoint_port, NULL);
```
blazoncek, line 28: I am not sure if this belongs to `loop()`. Perhaps a better place may be `connected()`?
`loop()` is executed at full possible rate which may be up to 5000-6000 times per second.

blazoncek (changes requested):
I have no way of testing it so I trust it works for you.
There are some things I'd like you to address before merging, though.

blazoncek:
One more thing: Please add your contact info into readme.md for user reference in case someone needs help with usermod.

acvigue:
Can you look at the changes I made?

blazoncek:
Using char[xx] eats a lot of stack, which is limited on MCU.
It may interfere with WLED behaviour (which itself uses a lot of stack space on some occasions) so it would be better to use `new` or `malloc()` for all the strings you want to store in `setup()` or `readFromConfig()`.

Otherwise I see you updated as requested and see no objection to this usermod.

acvigue:
fixed

blazoncek:
Are you sure this works?
`setup()` is called *after* `readFromConfig()`. So `malloc()` happens *after* all `strncpy()` which is not good.

I may have unintentionally mislead you when I suggested `setup()`. Constructor would be a better place, and then `free()` in destructor. Sorry about that.

acvigue:
>  So malloc() happens after all strncpy() which is not good.

Oops, ok. I'll take a look at fixing it tomorrow.

Aircoookie:
Hmm, correct me if I'm wrong, but since these keys are private members, they are created on the heap anyways as the usermod constructor is called in `usermods_list.cpp`. Stack would only be used if they are local variables in a given function.
Sorry that I'm only taking a look at it now, but to me this seems fine as implemented in 7296e87d2f82504e498dc4ed6204669f0799a1dd ...
Either way as this is an ESP32-only mod, memory usage is far less of a concern than with mods intended for ESP8266 as well.

acvigue:
> Sorry that I'm only taking a look at it now, but to me this seems fine as implemented in https://github.com/Aircoookie/WLED/commit/7296e87d2f82504e498dc4ed6204669f0799a1dd ...
Either way as this is an ESP32-only mod, memory usage is far less of a concern than with mods intended for ESP8266 as well.

All good, thanks for the input. I went ahead and reverted my branch back to that commit and then merged in all the changes from `main` so it should be good to go.

blazoncek:
> they are created on the heap anyways

You are correct and I was wrong. _Mea culpa._

Aircoookie:
All good :) I'd say this is ready to merge!

PR #3274, opened 2023-07-01
optimization for ABL hotfix

softhack007:
This is a (temporary) optimization that should reduce slow-downs, and improve color accuracy.
The change will probably be superseded with https://github.com/Aircoookie/WLED/pull/3272, maybe the idea of "immediate = true" could be kept.

This PR adds an optional parameter `immediate` to setBrightness(), so that ApplyPostAdjustments() is only called when `immediate=true`. Only ABL will use immediate=true, to ensure electrical safety of equipment.This allows some optimizations of performance, as ApplyPostAdjustments() is time consuming and loses color accuracy.

* busses.setBrightness(bri) --> applied to all future pixels - fast, lossless
* busses.setBrightness(bri, true) --> applied directly to all previously set pixels  - slower, lossy

@blazoncek, @Aircoookie: I did not test this change thoroughly yet, so would appreciate if you take a look.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1230,7 +1230,7 @@ void WS2812FX::estimateCurrentAndLimitBri() {
     uint16_t scaleI = scale * 255;
     uint8_t scaleB = (scaleI > 255) ? 255 : scaleI;
     uint8_t newBri = scale8(_brightness, scaleB);
-    busses.setBrightness(newBri); //to keep brightness uniform, sets virtual busses too
+    busses.setBrightness(newBri, (scaleB < 255)); //to keep brightness uniform, sets virtual busses too - softhack007: apply reductions immediately
     currentMilliamps = (powerSum0 * newBri) / puPerMilliamp;
   } else {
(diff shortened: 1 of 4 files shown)


wled00/FX_fcn.cpp
```diff
@@ -1230,7 +1230,7 @@ void WS2812FX::estimateCurrentAndLimitBri() {
     uint16_t scaleI = scale * 255;
     uint8_t scaleB = (scaleI > 255) ? 255 : scaleI;
     uint8_t newBri = scale8(_brightness, scaleB);
-    busses.setBrightness(newBri); //to keep brightness uniform, sets virtual busses too
+    busses.setBrightness(newBri, (scaleB < 255)); //to keep brightness uniform, sets virtual busses too - softhack007: apply reductions immediately
```
softhack007, line 1233: in case that flickering comes back, the condition `(scaleB < 255)` should be replaced with `true`

blazoncek (approved):
Looks fine to me.
Though me and @Aircoookie are currently working on `alt-buffer` branch so you may want to look there as well. This fix could come in handy there too.

PR #3280, opened 2023-07-09
Bus-level global buffering

### LINKED ISSUE #3264: B3 Auto Brightness Limiter issue ESP32

Aircoookie:
Global buffer implementation in BusDigital

- Memory efficient as only as many color channels as needed are allocated (3 bytes for RGB and 4 bytes for RGBW)
- about 5% lower FPS than WS2812FX global buffer

Closes #3272 
Resolves #3264

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -289,7 +289,6 @@ uint16_t mode_dynamic(void) {
   if (!SEGENV.allocateData(SEGLEN)) return mode_static(); //allocation failed
 
   if(SEGENV.call == 0) {
-    //SEGMENT.setUpLeds();  //lossless getPixelColor()
     //SEGMENT.fill(BLACK);
     for (int i = 0; i < SEGLEN; i++) SEGENV.data[i] = random8();
   }
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1219,6 +1219,9 @@ uint8_t WS2812FX::estimateCurrentAndLimitBri() {
 }
 
 void WS2812FX::show(void) {
+  static unsigned long sumMicros = 0, sumCurrent = 0;
+  static size_t calls = 0;
+  unsigned long microsStart = micros();
 
   // avoid race condition, caputre _callback value
   show_callback callback = _callback;
@@ -1226,6 +1229,7 @@ void WS2812FX::show(void) {
 
   uint8_t busBrightness = estimateCurrentAndLimitBri();
   busses.setBrightness(busBrightness);
+  sumCurrent += micros() - microsStart;
 
   // some buses send asynchronously and this method will return before
   // all of the data has been sent.
@@ -1237,6 +1241,13 @@ void WS2812FX::show(void) {
   if (diff > 0) fpsCurr = 1000 / diff;
   _cumulativeFps = (3 * _cumulativeFps + fpsCurr) >> 2;
   _lastShow = now;
+
+  sumMicros += micros() - microsStart;
+  if (++calls == 100) {
+    DEBUG_PRINTF("show calls: %d micros: %lu avg: %lu (current: %lu avg: %lu)\n", calls, sumMicros, sumMicros/calls, sumCurrent, sumCurrent/calls);
+    sumMicros = sumCurrent = 0;
+    calls = 0;
+  }
 }
 
 /**
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1263,7 +1263,8 @@ void WS2812FX::show(void) {
   uint8_t busBrightness = estimateCurrentAndLimitBri();
 
   if (_globalLedBuffer) { // copy data from buffer to bus
-    for (uint16_t i = 0; i < _length; i++) busses.setPixelColor(i, _globalLedBuffer[i]);
+    //for (uint16_t i = 0; i < _length; i++) busses.setPixelColor(i, _globalLedBuffer[i]);
+    busses.setColorsFromBuffer(_globalLedBuffer);
   } else {
     // if brightness changed since last show, must set everything again to update to new luminance
     if (_renderBrightness != busBrightness) {
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1209,7 +1209,13 @@ uint8_t WS2812FX::estimateCurrentAndLimitBri() {
     uint16_t len = bus->getLength();
     uint32_t busPowerSum = 0;
     for (uint_fast16_t i = 0; i < len; i++) { //sum up the usage of each LED
-      uint32_t c = bus->getPixelColor(i);
+      uint32_t c = 0;
+      if (_globalLedBuffer)
+      {
+        c = _globalLedBuffer[bus->getStart() + i];
+      } else {
+        c = bus->getPixelColor(i);
+      }
       byte r = R(c), g = G(c), b = B(c), w = W(c);
 
       if(useWackyWS2815PowerModel) { //ignore white component on WS2815 power calculation
@@ -1253,6 +1259,7 @@ void WS2812FX::show(void) {
   show_callback callback = _callback;
   if (callback) callback();
 
+  Bus::setRestoreBri(_renderBrightness);
   uint8_t busBrightness = estimateCurrentAndLimitBri();
 
   if (_globalLedBuffer) { // copy data from buffer to bus
@@ -1260,7 +1267,8 @@ void WS2812FX::show(void) {
   } else {
     // if brightness changed since last show, must set everything again to update to new luminance
     if (_renderBrightness != busBrightness) {
-      for (uint16_t i = 0; i < _length; i++) busses.setPixelColor(i, busses.getPixelColor(i)); // LOSSY and slow!
+      for (uint16_t i = 0; i < _length; i++) busses.setPixelColor(i, busses.getPixelColor(i)); // LOSSLESS due to trick (but still slow!)
+      Bus::setRestoreBri(busBrightness);
     }
   }
 
@@ -1343,6 +1351,7 @@ void WS2812FX::setBrightness(uint8_t b, bool direct) {
     // would be dangerous if applied immediately (could exceed ABL), but will not output until the next show()
     busses.setBrightness(b);
     _renderBrightness = b;
+    Bus::setRestoreBri(b);
   } else {
     unsigned long t = millis();
     if (_segments[0].next_time > t + 22 && t - _lastShow > MIN_SHOW_DELAY) show(); //apply brightness change immediately if no refresh soon
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1219,9 +1219,11 @@ uint8_t WS2812FX::estimateCurrentAndLimitBri() {
 }
 
 void WS2812FX::show(void) {
+  #ifdef WLED_DEBUG
   static unsigned long sumMicros = 0, sumCurrent = 0;
   static size_t calls = 0;
   unsigned long microsStart = micros();
+  #endif
 
   // avoid race condition, caputre _callback value
   show_callback callback = _callback;
@@ -1229,7 +1231,9 @@ void WS2812FX::show(void) {
 
   uint8_t busBrightness = estimateCurrentAndLimitBri();
   busses.setBrightness(busBrightness);
+  #ifdef WLED_DEBUG
   sumCurrent += micros() - microsStart;
+  #endif
 
   // some buses send asynchronously and this method will return before
   // all of the data has been sent.
@@ -1242,12 +1246,14 @@ void WS2812FX::show(void) {
   _cumulativeFps = (3 * _cumulativeFps + fpsCurr) >> 2;
   _lastShow = now;
 
+  #ifdef WLED_DEBUG
   sumMicros += micros() - microsStart;
   if (++calls == 100) {
     DEBUG_PRINTF("show calls: %d micros: %lu avg: %lu (current: %lu avg: %lu)\n", calls, sumMicros, sumMicros/calls, sumCurrent, sumCurrent/calls);
     sumMicros = sumCurrent = 0;
     calls = 0;
   }
+  #endif
 }
 
 /**
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1102,8 +1102,12 @@ void WS2812FX::service() {
     // last condition ensures all solid segments are updated at the same time
     if(nowUp > seg.next_time || _triggered || (doShow && seg.mode == FX_MODE_STATIC))
     {
-      if (seg.grouping == 0) seg.grouping = 1; //sanity check
-      doShow = true;
+      if (seg.grouping == 0) seg.grouping = 1; // sanity check
+      if (!doShow) {
+        busses.setBrightness(_brightness); // bus luminance must be set before FX using setPixelColor()
+        _renderBrightness = _brightness; // save in case brightness gets changed while FX is calculated
+        doShow = true;
+      }
       uint16_t delay = FRAMETIME;
 
       if (!seg.freeze) { //only run effect function if not frozen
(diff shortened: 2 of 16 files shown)


blazoncek:
@Aircoookie that will not work if I select a preset which changes multiple segments. Each segment should have it's own queue. Or am I mistaken?

Aircoookie:
It should work, given that all segments but the _one_ currently rendering FX are safe to update immediately. Even if updating segments from JSON somehow takes longer than the effect function, it is OK since the queued changes are applied immediately after FX function completes, making the queue variables available again. And if the effect function takes longer than JSON code (probable case) , the change can be applied immediately all to segments not currently running their FX function.

wled00/FX_fcn.cpp
```diff
@@ -174,12 +177,15 @@ void Segment::deallocateData() {
...
-  if (reset) {
-    deallocateData();
-    next_time = 0; step = 0; call = 0; aux0 = 0; aux1 = 0;
-    if (_queuedChanges) setUp(_qStart, _qStop, grouping, spacing, offset, _qStartY, _qStopY, true);
-    reset = false; // setOption(SEG_OPTION_RESET, false);
+  if (!reset) return;
+  
+  deallocateData();
+  next_time = 0; step = 0; call = 0; aux0 = 0; aux1 = 0;
+  if (_queuedChangesSegId == strip.getCurrSegmentId()) { // apply queued changes
```
blazoncek, line 184: I know this looks ok, but it is not.
It mixes global `strip` into independent Segment class. IMO this is unacceptable (separation of functionality). I put a lot of effort to make Segment independent of WS2812FX and not mix code as it becomes unmanageable otherwise in the long term..
If you need to maintain ID to determine which segment is being worked on then add ID variable to Segment class.

For a better solution see another comment below.

wled00/FX_fcn.cpp
```diff
@@ -1087,7 +1098,7 @@ void WS2812FX::service() {
     // reset the segment runtime data if needed
     seg.resetIfRequired();
 
-    if (!seg.isActive()) continue;
+    if (!seg.isActive()) { _segment_index++; continue; }
```
blazoncek, line 1101: Instead of duplicating code why not just:
```c++
if (isActive()) {
...
}
_segment_index++;
```

wled00/FX_fcn.cpp
```diff
@@ -1115,6 +1126,7 @@ void WS2812FX::service() {
 
       seg.next_time = nowUp + delay;
     }
+    seg.resetIfRequired(); // another reset chance, mainly to apply new segment bounds if queued
```
blazoncek, line 1129: unnecessary

wled00/json.cpp
```diff
@@ -112,8 +112,8 @@ bool deserializeSegment(JsonObject elem, byte it, byte presetId)
   if (stop > start && of > len -1) of = len -1;
 
   // update segment (delete if necessary)
-  // we must not change segment dimensions during drawing of effects as that may produce undesired behaviour (crash)
-  seg.setUp(start, stop, grp, spc, of, startY, stopY, !strip.isServicing());
+  // we must not change segment dimensions during drawing of effects in that segment as concurrent access may cause a crash
+  seg.setUp(start, stop, grp, spc, of, startY, stopY, id);
```
blazoncek, line 116: you can check here which segment is active (being drawn to) as `id == strip.getCurrSegmentId()` to force ID detection outside of Segment class

blazoncek (changes requested):
I do not like mixing `strip` into Segment class.

Aircoookie:
Alright, maybe it is a good idea to decouple queuing from segment completely, am on it now.

wled00/json.cpp
Aircoookie: I do not want to have this logic in json.cpp as queuing should be considered internal WS2812FX and Segment functionality, it has nothing to do with JSON per se. This check would also need to be duplicated in `set.cpp` for the legacy HTTP API.

wled00/FX_fcn.cpp
Aircoookie: Necessary, but only for applying queued changes immediately after the FX function is done. Will be changed, as I will also move applying queued changes out of `resetIfRequired()`

wled00/FX_fcn.cpp
softhack007: Hi @Aircoookie, unfortunately with your last commit "Static queued segment bounds" my crashes are coming back.

Not sure which code lines are to blame, as the crash dump does not give me any meaningful information. Usually it means we have memory corruption somewhere...

wled00/json.cpp
blazoncek: I'm not sure which is uglier: 
`seg.setUp(start, stop, grp, spc, of, startY, stopY, id == strip.getCurrSegmentId());`
or
`strip.setSegment(id, start, stop, grp, spc, of, startY, stopY);` (with added queueing in setSegment)

 😉 

And BTW `set.cpp` only ever deals with one segment at a time, but you are correct and I missed that. Still, the original code forced queued segment change which was a desired thing.

softhack007:
> And if the effect function takes longer than JSON code

Actually my observation on esp32 is the opposite, when multiple segments are updated recursively in json.cpp (due to UI click to change preset).
* effects usually need 2-20ms for drawing
* deserializeState() for a complex presets (5-8 segments) can take up to 70ms

Aircoookie:
Hmm, I'm noticing ABL is completely f*cked again with buffering disabled (flickering galore, visually looks like the issue in b3).
Even with buffering enabled, there is a major artifact during on/off transitions, it starts to dim down, jumps back up to a higher brightness at about 50% of the transition, then fully dims down. I suspect it's due to integer division, will take a look soon.

wled00/FX_fcn.cpp
Aircoookie: That's unfortunate... please retest with the latest commit, although I wouldn't expect it to fix the crashes. What do I have to do to reproduce the issue? Just switching between presets with different multi-segment setups?

softhack007: Yes, to create a crash basicially I'm switching between several presets that have multiple segments. The presets segment layout should differ, this help to "enforce" the problem.

I can test again later this evening - currently busy@job.

blazoncek:
> Hmm, I'm noticing ABL is completely f*cked again with buffering disabled (flickering galore, visually looks like the issue in b3).

I am not sure ABL is at issue in case of double buffering. I've been testing a set-up with S2 and 36 LEDs in a ring a bit with ABL set up to 1000mA and with ABL turned off.
When the brightness was below 1A I was able to see (and record) brightness variations with or without ABL. The brightness variation was miniscule but still noticeable. When the brightness was above 1A there was no difference in variations (they were still visible and at the same level as before) except that with ABL the total brightness was lower as expected.
I am not sure if this was due to poor PSU (though I am using branded USB PSU) or something in the NeoPixelBus or the way we draw pixels.

When I switched to un-buffered output, the story was totally different. The ABL kicked in even when total brightness was way below pre-set value and it was pulsing. The problem must be in modified `getPixelColor()` which uses `restoreColorLossy()`.

blazoncek:
@Aircoookie the problem with ABL is in restoration of color in `getPixelColor()`.

We will be unable to restore pixels properly (or consistently) if using same approach for buffered or un-buffered display. When using buffered display we paint actual (scaled by brightness) NPB pixels in `BusDigital::show()`, at the same time `getPixelColor()` will return unmodified values from buffer.
When using un-buffered approach, if we use same principle to paint NPB pixels using `SetLuminance(255)` and then in the `BusDigital::show()` apply post adjustments after changing `PolyBus::setBrightness(bri)`, then our `getPixelColor()` may return either scaled-down pixel value or unmodified depending when and how it was used. If it is used on a pixel that was set by `setPixelColor()`in the same frame (before `show()` is called) it will return un-scaled value (the same as was set by `setPixelColor()` but if it is used on an unmodified pixel in next frame it will return scaled-down value.

Aircoookie:
Yep, that is it. I think in the unbuffered case we'll want to return to setting the user-intended brightness to NPBLg before drawing the effect function and only doing postAdjustments or setPC(getPC()) if the ABL limit is exceeded. I'll implement it tomorrow morning and see if it improves things!

blazoncek:
I already have the fix and am testing it ATM.

Aircoookie:
Just tested your changes, they do indeed fix the issue with flickering due to ABL not applying every other frame. Setting `maintainBufferConsistency = false` if buffer is enabled also seems to improve FPS by ~5%, nice find!

I still see "brightness jumping artifacts" as soon as ABL kicks in though. I've found the root cause and came to the conclusion that we must not use `applyPostAdjustments` at all, as it will take the already scaled color (if user set brightness is < 255 or we are in a transition) and scale it to the new limited brightness again, which scales it down too much. Therefore we must take care to use _restore_ when repainting pixels when ABL limits the brightness.

My latest commit effectively re-introduces the behavior of `NeoPixelBrightnessBus` by repainting (including restore) on every brightness change. Without it, I see no possibility to have correctly behaving ABL in unbuffered mode. Furthermore, getPixelColor() would not return correctly restored values, as the bus brightness is reset to `_brightness` after show() but the buffer contents are still at `newBri`.
The only drawback is that the bus needs to be repainted twice every show() _only if ABL is actually limiting_ (for limiting and then for restoring to intended brightness), but the way I implemented it skips all mapping/color order steps at least, so it should be quite fast.

I'm now very happy with the functionality and performance (on par with old `globalbuffer`) of this branch.

wled00/FX_fcn.cpp
```diff
@@ -1082,6 +1082,11 @@ void WS2812FX::service() {
     // last condition ensures all solid segments are updated at the same time
     if (seg.isActive() && (nowUp > seg.next_time || _triggered || (doShow && seg.mode == FX_MODE_STATIC)))
     {
+      if (!doShow) {
+        // returning bus brightness to its original value is done here, as to not interfere with asyncronous show()
+        // TODO if it is safe, prefer to restore brightness in show()
+        busses.setBrightness(_brightness, true); // "repaint" all pixels if brightness has changed
```
blazoncek, line 1088: Perhaps it would be ok, to restore brightness at the start of `service()`, not in a for loop.

wled00/FX_fcn.cpp
```diff
@@ -1242,8 +1247,9 @@ void WS2812FX::show(void) {
   // See https://github.com/Makuna/NeoPixelBus/wiki/ESP32-NeoMethods#neoesp32rmt-methods
   busses.show();
 
-  // return bus brightness to original value
-  if (newBri < _brightness) busses.setBrightness(_brightness);
+  // returning bus brightness to its original value is done in the next frame, as to not interfere with asyncronous show()
+  // TODO if it is safe, prefer to restore brightness here
+  //if (newBri < _brightness) busses.setBrightness(_brightness, true);
```
blazoncek, line 1252: RMT uses dual buffers (as @Makuna explained to me [here](https://github.com/Aircoookie/WLED/issues/3264#issuecomment-1638412940)) so it *is* safe to immediately update brightness.
It may not be so with I2S (which we rarely use TBH).

wled00/FX_fcn.cpp
Aircoookie: It only does it in the first iteration, and this makes sure it is restored right before a segment actually wants to draw.
If it is indeed safe to call gPC/sPC immediately after show(), that is of course preferable as it is easier to maintain.

blazoncek (approved):

Aircoookie:
Performance figures (ESP32, 4x512 SK6812 busses, 1x 2048 LED segment) 😄 

Buffer, no ABL: 24.6 FPS
Buffer, ABL on and limiting: 24.2

Unbuffered, no ABL: 25.9 
Unbuffered, ABL on, not limiting: 24.3
Unbuffered, ABL on, limiting: 22.0

blazoncek:
@softhack007 you have more experience: Can we update platform from 5.2.0 to 5.3.0 for ESP32-S2 and -C3?
I am already using 5.3.0 locally but would rather see a more experienced person change that. 😄

softhack007:
> @softhack007 you have more experience: Can we update platform from 5.2.0 to 5.3.0 for ESP32-S2 and -C3?

Hi @blazoncek I've just checked, platform 5.3.0 includes a new version arduino-esp32 2.0.6, and ESP-IDF 4.4.3.
Both look ok basically. Just one "but": new bootloader files. It looks like they updated the bootloaders, so the new install2.wled.me would also need updated bootloaders. @Aircoookie you think that's ok?

To keep changes manageable, I would suggest we start a new branch "core-updates" where fastled 3.6.0, platform 5.3.0, and (maybe) a newer irremote8266 could be tested before we update main. I can do that before the weekend.

blazoncek:
> Just one "but": new bootloader files.

That's probably why I'm seeing stable running on both C3 and S2 and some people are not.
I would also agree to update platformio.ini in a separate PR.

Aircoookie:
Let's just update it either here or in main. I will then update the bootloader binaries on the web installer for beta 4

softhack007:
> Let's just update it either here or in main

OK, then I'll change it here first.

wled00/FX.cpp
```diff
@@ -1212,19 +1212,21 @@ uint16_t mode_fireworks() {
 
   bool valid1 = (SEGENV.aux0 < width*height);
   bool valid2 = (SEGENV.aux1 < width*height);
+  uint8_t x = SEGENV.aux0%width, y = SEGENV.aux0/width; // 2D coordinates stored in upper and lower byte
```
softhack007, line 1215: It might be better to use `uint16_t` here, to also support "extreme" matrix sizes like 300 x 3.
Also uint8_t is very slow, as each computation includes an explicit " & 0xFF" to enforce correct overflows.

Aircoookie:
All seems good to me, merging tomorrow.

wled00/FX.cpp
blazoncek: Matrices with dimensions greater than 255x255 are not supported ATM.
They would require change to Segment and Panel classes.

I do agree that forcibly constraining a counter may not be the best idea. Perhaps we should change every unsigned counter/variable to use `size_t` instead, which should use optimised word size (usually `unsigned int`).

softhack007: > I do agree that forcibly constraining a counter may not be the best idea. Perhaps we should change every unsigned counter/variable to use `size_t` instead, which should use optimised word size (usually `unsigned int`).

Hi, thanks for the clarification.
This "loop/counter" variable thing is kind of personal preference, as long as its efficient. My point was just that `uint8_t` is not efficient (w.r.t speed) on esp32, as esp32 emulates 8bit with 16bit registers, and you might even have penalties for misaligned addresses of an 8bit value.

(slightly off-topic)
I must admit I'm not a friend of `size_t` when used as a generic type, because the standards define it as "type of the result of the sizeof operator". Its a special purpose type, like `time_t` is the type of unix epoch time.

I would suggest this for loops and counters:
* use `int` or `unsigned` when the results will fit into 16 bit. This type is guaranteed to be a "native" register size on any architecture (esp32, 8266, or -C3). Additionally, `int` and `unsigned` are always stored with native alignment.
* use `fast` types when more bytes are needed, and the compiler will pick the most efficient implementation for us.

basic types: https://en.cppreference.com/w/cpp/language/types
fixed-width types: https://en.cppreference.com/w/cpp/types/integer

blazoncek: I agree with every point.
I did start to use `int` (mostly in `for` loops) but was not thorough enough.

PR #3291, opened 2023-07-13
Add ability to toggle devices from nodes view.

blazoncek:


wled00/data/index.js
```diff
@@ -2571,6 +2578,24 @@ function setBalance(b)
 	requestJson(obj);
 }
 
+function rmtTgl(ip,i) {
+	event.preventDefault();
+	event.stopPropagation();
+	fetch(`http://${ip}/win&T=2`, {method: 'get'})
```
Aircoookie, line 2584: We could consider `T=2&NN` so that really only that one instance is toggled (no sync). On the other hand, it might be more intuitive to have this power button do exactly what the native one in the device UI itself does. Which would you prefer?

wled00/NodeStruct.h
```diff
@@ -23,7 +23,13 @@ struct NodeStruct
   String    nodeName;
   IPAddress ip;
   uint8_t   age;
-  uint8_t   nodeType;
+  union {
```
Aircoookie, line 26: IMO it is not a huge deal "wasting" space here, I wouldn't even add alternative type IDs starting from 1 at all, just adds overhead, as we should keep the 82/32-35 IDs anyways for compatibility.
We still have an extra byte for future enhancements without increasing the size of `NodeStruct` as 24 bit is sufficient for all build numbers until the year 2167 😄

And don't underestimate how many ESP32 variants Espressif can still throw on the market (H6? C6? C2? Hard to keep track...). Maybe we even want to distinguish different flash layouts or something with this in the future, so 7 bits seems reasonable for type.

Aircoookie (approved):
Really like this!

wled00/data/index.js
blazoncek: IMO sync should be respected if enabled.

PR #3293, opened 2023-07-17
Move CTT calculation from BusPwm to Bus class.

Robert-github-com:
This enable it to other bus types.
Need for adding RGBWW LEDs on digital bus.

### DIFF
--- a/wled00/bus_manager.h
+++ b/wled00/bus_manager.h
@@ -8,9 +8,7 @@
 #include "const.h"
 
 //colors.cpp
-uint32_t colorBalanceFromKelvin(uint16_t kelvin, uint32_t rgb);
 uint16_t approximateKelvinFromRGB(uint32_t rgb);
-void colorRGBtoRGBW(byte* rgb);
 
 #define GET_BIT(var,bit)    (((var)>>(bit))&0x01)
 #define SET_BIT(var,bit)    ((var)|=(uint16_t)(0x0001<<(bit)))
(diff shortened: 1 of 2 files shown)


wled00/bus_manager.h
```diff
@@ -150,6 +155,32 @@ class Bus {
         if (_cctBlend > WLED_MAX_CCT_BLEND) _cctBlend = WLED_MAX_CCT_BLEND;
       #endif
     }
+    static void calculateCCT(uint32_t c, uint8_t &ww, uint8_t &cw) {
```
blazoncek, line 158: This can be a private or protected, does not need to be public method.

wled00/bus_manager.h
```diff
@@ -7,6 +7,11 @@
 
 #include "const.h"
 
+//colors.cpp
+uint32_t colorBalanceFromKelvin(uint16_t kelvin, uint32_t rgb);
```
blazoncek, line 11: Why did you move all three into .h? There is no need to do that, only move a needed one.

blazoncek (commented):
Apart from minor suggestions all seems fine.

wled00/bus_manager.h
Robert-github-com: Done

wled00/bus_manager.h
Robert-github-com: Moved back

deece:
@blazoncek With the merge of #3810, this can now be closed

PR #3298, opened 2023-07-21
Support for Fw1906

Robert-github-com:
Add support for RgbwwColor and GRBCW strips that use FW1906 chips.

This change requires:
https://github.com/Aircoookie/WLED/pull/3293
and
https://github.com/Makuna/NeoPixelBus/commit/978b62f84137266bc2a47ea49759c1b15e2c8f5c
The fix in NeoPixelBus is not in any release yet, and I don't know the way to force platformio to fetch specific commit, so I temporary changed lib_deps to my branch, if someone wants to test. This needs to be changed before merging.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -177,7 +177,7 @@ lib_compat_mode = strict
 lib_deps =
     fastled/FastLED @ 3.5.0
     IRremoteESP8266 @ 2.8.2
-    https://github.com/Robert-github-com/NeoPixelBus.git
+    https://github.com/Makuna/NeoPixelBus.git#978b62f84137266bc2a47ea49759c1b15e2c8f5c
     https://github.com/Aircoookie/ESPAsyncWebServer.git @ ~2.0.7
   #For use of the TTGO T-Display ESP32 Module with integrated TFT display uncomment the following line
     #TFT_eSPI
--- a/platformio.ini
+++ b/platformio.ini
@@ -177,7 +177,7 @@ lib_compat_mode = strict
 lib_deps =
     fastled/FastLED @ 3.5.0
     IRremoteESP8266 @ 2.8.2
-    makuna/NeoPixelBus @ 2.7.5
+    https://github.com/Robert-github-com/NeoPixelBus.git
     https://github.com/Aircoookie/ESPAsyncWebServer.git @ ~2.0.7
   #For use of the TTGO T-Display ESP32 Module with integrated TFT display uncomment the following line
     #TFT_eSPI
--- a/wled00/bus_manager.cpp
+++ b/wled00/bus_manager.cpp
@@ -147,7 +147,7 @@ void BusDigital::setStatusPixel(uint32_t c) {
 
 void IRAM_ATTR BusDigital::setPixelColor(uint16_t pix, uint32_t c) {
   if (_type == TYPE_FW1906 || _type == TYPE_SK6812_RGBW || _type == TYPE_TM1814 || _type == TYPE_WS2812_1CH_X3) c = autoWhiteCalc(c);
-  if (_type != TYPE_FW1906 && _cct >= 1900) c = colorBalanceFromKelvin(_cct, c); //color correction from CCT
+  if (_cct >= 1900) c = colorBalanceFromKelvin(_cct, c); //color correction from CCT
   if (reversed) pix = _len - pix -1;
   else pix += _skip;
   uint8_t co = _colorOrderMap.getPixelColorOrder(pix+_start, _colorOrder);
--- a/wled00/bus_manager.cpp
+++ b/wled00/bus_manager.cpp
@@ -9,10 +9,6 @@
 #include "bus_wrapper.h"
 #include "bus_manager.h"
 
-//colors.cpp
-uint32_t colorBalanceFromKelvin(uint16_t kelvin, uint32_t rgb);
-uint16_t approximateKelvinFromRGB(uint32_t rgb);
-void colorRGBtoRGBW(byte* rgb);
 
 //udp.cpp
 uint8_t realtimeBroadcast(uint8_t type, IPAddress client, uint16_t length, byte *buffer, uint8_t bri=255, bool isRGBW=false);
@@ -253,29 +249,10 @@ void BusPwm::setPixelColor(uint16_t pix, uint32_t c) {
   uint8_t g = G(c);
   uint8_t b = B(c);
   uint8_t w = W(c);
-  uint8_t cct = 0; //0 - full warm white, 255 - full cold white
-  if (_cct > -1) {
-    if (_cct >= 1900)    cct = (_cct - 1900) >> 5;
-    else if (_cct < 256) cct = _cct;
-  } else {
-    cct = (approximateKelvinFromRGB(c) - 1900) >> 5;
-  }
-
+  
   uint8_t ww, cw;
-  #ifdef WLED_USE_IC_CCT
-  ww = w;
-  cw = cct;
-  #else
-  //0 - linear (CCT 127 = 50% warm, 50% cold), 127 - additive CCT blending (CCT 127 = 100% warm, 100% cold)
-  if (cct       < _cctBlend) ww = 255;
-  else ww = ((255-cct) * 255) / (255 - _cctBlend);
-
-  if ((255-cct) < _cctBlend) cw = 255;
-  else                       cw = (cct * 255) / (255 - _cctBlend);
-
-  ww = (w * ww) / 255; //brightness scaling
-  cw = (w * cw) / 255;
-  #endif
+  
+  calculateCCT(c, ww, cw);
 
   switch (_type) {
     case TYPE_ANALOG_1CH: //one channel (white), relies on auto white calculation
--- a/wled00/bus_manager.cpp
+++ b/wled00/bus_manager.cpp
@@ -140,7 +140,7 @@ void BusDigital::setBrightness(uint8_t b, bool immediate) {
 //TODO only show if no new show due in the next 50ms
 void BusDigital::setStatusPixel(uint32_t c) {
   if (_skip && canShow()) {
-    PolyBus::setPixelColor(_busPtr, _iType, 0, R(c), G(c), B(c), W(c), 0, _colorOrderMap.getPixelColorOrder(_start, _colorOrder));
+    PolyBus::setPixelColor(_busPtr, _iType, 0, c, _colorOrderMap.getPixelColorOrder(_start, _colorOrder));
     PolyBus::show(_busPtr, _iType);
   }
 }
@@ -161,12 +161,7 @@ void IRAM_ATTR BusDigital::setPixelColor(uint16_t pix, uint32_t c) {
       case 2: c = RGBW32(R(cOld), G(cOld), W(c)   , 0); break;
     }
   }
-  
-  uint8_t ww = W(c), cw = 0;
-  if (_type == TYPE_FW1906)
-    calculateCCT(c, ww, cw);
-
-  PolyBus::setPixelColor(_busPtr, _iType, pix, R(c), G(c), B(c), ww, cw, co);
+  PolyBus::setPixelColor(_busPtr, _iType, pix, c, co);
 }
 
 uint32_t BusDigital::getPixelColor(uint16_t pix) {
--- a/wled00/bus_manager.cpp
+++ b/wled00/bus_manager.cpp
@@ -148,6 +148,7 @@ void BusDigital::setStatusPixel(uint32_t c) {
 void IRAM_ATTR BusDigital::setPixelColor(uint16_t pix, uint32_t c) {
   if (_type == TYPE_FW1906 || _type == TYPE_SK6812_RGBW || _type == TYPE_TM1814 || _type == TYPE_WS2812_1CH_X3) c = autoWhiteCalc(c);
   if (_cct >= 1900) c = colorBalanceFromKelvin(_cct, c); //color correction from CCT
+  if (_type == TYPE_FW1906) calculateCCT(c, PolyBus::cctWW, PolyBus::cctCW); // FW1906 ignores W component in c
   if (reversed) pix = _len - pix -1;
   else pix += _skip;
   uint8_t co = _colorOrderMap.getPixelColorOrder(pix+_start, _colorOrder);
--- a/wled00/bus_manager.cpp
+++ b/wled00/bus_manager.cpp
@@ -140,14 +140,14 @@ void BusDigital::setBrightness(uint8_t b, bool immediate) {
 //TODO only show if no new show due in the next 50ms
 void BusDigital::setStatusPixel(uint32_t c) {
   if (_skip && canShow()) {
-    PolyBus::setPixelColor(_busPtr, _iType, 0, c, _colorOrderMap.getPixelColorOrder(_start, _colorOrder));
+    PolyBus::setPixelColor(_busPtr, _iType, 0, R(c), G(c), B(c), W(c), 0, _colorOrderMap.getPixelColorOrder(_start, _colorOrder));
     PolyBus::show(_busPtr, _iType);
   }
 }
 
 void IRAM_ATTR BusDigital::setPixelColor(uint16_t pix, uint32_t c) {
-  if (_type == TYPE_SK6812_RGBW || _type == TYPE_TM1814 || _type == TYPE_WS2812_1CH_X3) c = autoWhiteCalc(c);
-  if (_cct >= 1900) c = colorBalanceFromKelvin(_cct, c); //color correction from CCT
+  if (_type == TYPE_FW1906 || _type == TYPE_SK6812_RGBW || _type == TYPE_TM1814 || _type == TYPE_WS2812_1CH_X3) c = autoWhiteCalc(c);
+  if (_type != TYPE_FW1906 && _cct >= 1900) c = colorBalanceFromKelvin(_cct, c); //color correction from CCT
   if (reversed) pix = _len - pix -1;
   else pix += _skip;
   uint8_t co = _colorOrderMap.getPixelColorOrder(pix+_start, _colorOrder);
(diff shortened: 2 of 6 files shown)


wled00/bus_manager.cpp
```diff
@@ -162,7 +161,12 @@ void IRAM_ATTR BusDigital::setPixelColor(uint16_t pix, uint32_t c) {
       case 2: c = RGBW32(R(cOld), G(cOld), W(c)   , 0); break;
     }
   }
-  PolyBus::setPixelColor(_busPtr, _iType, pix, c, co);
+  
+  uint8_t ww = W(c), cw = 0;
+  if (_type == TYPE_FW1906)
+    calculateCCT(c, ww, cw);
+
+  PolyBus::setPixelColor(_busPtr, _iType, pix, R(c), G(c), B(c), ww, cw, co);
```
softhack007, line 169: Hi, 
setPixelColor() is on the critical path for performance.

Is it strictly necessary to extend the polyBus interface to explicitly require color components (r g,b,w,)  instead keeping them in uint32_t and just adding the new byte for cw?

If I understand your code correctly, we would require - for all kinds of busses - to split RGBW into  `R(c), G(c), B(c), W(c), 0`, and later re-assemble  them into uint32_t? This adds around 20 extra operations just to go back and forth between color formats. It might seem like a "small price", but if applied to a few thousand LEDs at 100 fps, it might become a noticeable slowdown.

Robert-github-com: Hi,
The split is not required. I can change the interface. But in PolyBus::setPixelColor the uint32 is split for 4 uint8 local variables anyway, and this operation is removed. I don't want to guess witch version is faster.

softhack007: Thanks for the clarification :-) so before changing, @blazoncek what do you think, which option would be more efficient?

blazoncek:
You can use something similar to:
`https://github.com/makuna/neopixelbus#978b62f`
in `lib_deps` to force specific commit to be used for a library.

Unfortunately I do not like using unreleased libraries so we'll probably wait until new version of NPB is released as I mentioned elsewhere that last NPB update caused us hedaches we are still recovering from.

wled00/bus_manager.cpp
```diff
@@ -141,14 +140,14 @@ void BusDigital::setBrightness(uint8_t b, bool immediate) {
...
+    PolyBus::setPixelColor(_busPtr, _iType, 0, R(c), G(c), B(c), W(c), 0, _colorOrderMap.getPixelColorOrder(_start, _colorOrder));
     PolyBus::show(_busPtr, _iType);
   }
 }
 
 void IRAM_ATTR BusDigital::setPixelColor(uint16_t pix, uint32_t c) {
-  if (_type == TYPE_SK6812_RGBW || _type == TYPE_TM1814 || _type == TYPE_WS2812_1CH_X3) c = autoWhiteCalc(c);
-  if (_cct >= 1900) c = colorBalanceFromKelvin(_cct, c); //color correction from CCT
+  if (_type == TYPE_FW1906 || _type == TYPE_SK6812_RGBW || _type == TYPE_TM1814 || _type == TYPE_WS2812_1CH_X3) c = autoWhiteCalc(c);
+  if (_type != TYPE_FW1906 && _cct >= 1900) c = colorBalanceFromKelvin(_cct, c); //color correction from CCT
```
blazoncek, line 150: That's inapproptiate as user has an option to use color temperature calculated from RGB values.

wled00/bus_manager.cpp
blazoncek: I would leave API as is and add a static variable to hold CCT information for PolyBus.

wled00/bus_manager.h
```diff
@@ -129,14 +132,14 @@ class Bus {
     }
     virtual bool hasWhite() { return Bus::hasWhite(_type); }
     static  bool hasWhite(uint8_t type) {
-      if ((type >= TYPE_WS2812_1CH && type <= TYPE_WS2812_WWA) || type == TYPE_SK6812_RGBW || type == TYPE_TM1814) return true; // digital types with white channel
+      if ((type >= TYPE_FW1906 && type <= TYPE_WS2812_WWA) || type == TYPE_SK6812_RGBW || type == TYPE_TM1814) return true; // digital types with white channel
```
blazoncek, line 135: Why did you remove 1 channel WS2812?

wled00/const.h
```diff
@@ -223,6 +223,7 @@
 #define TYPE_NONE                 0            //light is not configured
 #define TYPE_RESERVED             1            //unused. Might indicate a "virtual" light
 //Digital types (data pin only) (16-31)
+#define TYPE_FW1906              17            //RGB + CW + WW + unused channel (6 channels per IC)
```
blazoncek, line 226: I would use 28 instead of 17.

blazoncek (changes requested):
Instead of adding a 5th value in `setPixelColor()` please extend `PolyBus` class to use pre-set CCT (perhaps as additional variable). Then extend `BusManager` to set this value from CCT methods it provides.

Robert-github-com:
I was trying this # notation but it seems not working in lib_deps. Its understandable that this change needs to wait for new release of NeoPixelBus. I push this for review, and for other people to test.

wled00/bus_manager.h
Robert-github-com: I do not remove 1 channel WS2812 I added FW1906. :)

blazoncek:
It does whe properly formatted. I'm away from computer so can't provide exact format ATM.

wled00/bus_manager.h
blazoncek: Sorry, I had in mind updated type to 30.

wled00/bus_manager.cpp
Robert-github-com: It is also not used for TYPE_ANALOG_5CH.

wled00/const.h
Robert-github-com: Done

wled00/bus_manager.cpp
blazoncek: Perhaps a bug?

Robert-github-com:
> Instead of adding a 5th value in `setPixelColor()` please extend `PolyBus` class to use pre-set CCT (perhaps as additional variable). Then extend `BusManager` to set this value from CCT methods it provides.

I've done it in a little different way. Instead of adding new static variables I call a static function from class Bus in PolyBus.

blazoncek:
I hate the idea of mixing Bus into PolyBus so I removed the dependency.
I am not satisfied with the solution as it breaks regular behaviour of setPixelColor but I have yet to come up with something better. 😄 

There is also an issue you did not address and that is reverse CCT for retrieving pixel information from `getPixelColor()`. Current implementation is lacking and needs to be addressed.

Robert-github-com:
> 
I assumed that it is not an issue, no reverse CCT is used for other pixel formats in digital and PWM buses.

blazoncek:
> I assumed that it is not an issue, no reverse CCT is used for other pixel formats in digital and PWM buses.

Other CCT enabled bus (`BusPwm`) only ever deals with 1 pixel so quite less important (as no 1 pixel effects rely on previous pixel value) than `BusDigital`. But of course it is a flaw and you can address it if you wish at the same time.

blazoncek:
@Robert-github-com I have re-checked original source regarding CCT handling (`BusPwm`) and came to conclusion that best approach would be to move your `calculateCCT()` out of `Bus` class as it has nothing to do with buses as such and place it into `colors.cpp` as a standalone function.
Then it would be Ok for me to be included in `PolyBus` as you originally intended.

Nevertheless, `PolyBus` currently has no concept of CCT and API only allows for `uint32_t` for color object so there is still something missing for linking bus CCT value and `PolyBus` way of rendering it in the "correct" way. (Don't forget `BusPwm` handles CCT internally ATM.)

@Aircoookie and @softhack007 what would be your idea for adding CCT handling to `PolyBus`? Do we keep it external in a similar way as I updated it or do we extend `PolyBus` API to include either CCT info or WW & CW color bytes? Or something completely different?

Aircoookie:
@blazoncek this is something I already thought a lot about when originally adding CCT support.
Right now, there is just one CCT value per segment, so individual-pixel CCT control is not possible on the segment level. Also, there is no way to e.g. use a warm CT as primary and a cold as secondary color. This, as well as per-pixel CCT control is possible _right now_, but only with "CCT from RGB" mode, which was made with CCT-only devices in mind. For addressable RGBCCT like the FW1906, having CCT as part of the segment color objects becomes something we should seriously consider looking into.

I'd say adding support for CCT to `PolyBus` specifically should be as simple as adding another `uint16_t` (or `uint32_t` for performance?) parameter to `PolyBus::setPixelColor()`. I would pass the WW and CW bytes and do the conversion from CCT on the `BusDigital` level, though it shouldn't technically matter where it's done.
This will already give us per-segment CCT functionality, and we can look into adding true addressable CCT down the line. One option might be a custom color object used throughout all layers from Segment to BusDigital, or even use `uint64_t` down to PolyBus level. I'd assume this to have quite the performance impact though, so it'd be something we just have to test.

Also thought about cramming everything into the `uint32_t` and dealing with just 4 bits for white brightness and color temperature each, but excluded that possibility as it would be a terrible user experience, especially during transitions.

wled00/bus_manager.cpp
Robert-github-com: Removed in last update.

avwuff:
I'm trying to compile this right now and give it a try as I have a similar light strip!  Wish me luck!

avwuff:
Unfortunately the chips in my strips appear to be something other than the FW1906, despite having a similar LED configuration.  The chips themselves are entirely unlabled.

Sorry Robert, I wish I could have helped test this:
image

platformio.ini
deece, line 1: The required support is in 2.7.6, so we should probably target that:
https://github.com/Makuna/NeoPixelBus/releases/tag/2.7.6

deece:
Hi folks, I've squashed all the cleanup to create a minimal set of patches, and rebased on top of the current main. @Robert-github-com , you can grab it here: https://github.com/deece/WLED/tree/fw1906

Robert-github-com:
> Hi folks, I've squashed all the cleanup to create a minimal set of patches, and rebased on top of the current main. @Robert-github-com , you can grab it here: https://github.com/deece/WLED/tree/fw1906

Great! Seems to work fine.

valkjsaaa:
Screenshot 2023-10-20 at 21 21 23
It worked for me. One small problem is that I can't seem to be able to change it to a mode where only the white LEDs are used with a blend of cold and warm color temperatures.

As shown in the screenshot, when I change the first slider to the left (turning off the color LED altogether), the second slide automatically slides to the rightmost (cold white), If I drag the second slider (white point) anywhere towards the left, the first slide immediately goes to 100%, and the color LEDs turn on (lower CRI) and make the color of objects in the room look weird. 

Thanks!

valkjsaaa:
It appears that because I implemented enabled Calculate CCT from RGB WLED can integrate with home assistant. Not sure if this is the expected behavior 🤔

oakey81:
@deece Hey man I compiled your FW1906 fork and uploaded to my ESP32. I cant see the option for FW1906 in the LED config.

blazoncek:
> It appears that because I implemented enabled Calculate CCT from RGB WLED can integrate with home assistant. Not sure if this is the expected behavior 🤔

That is a HA issue. It will not allow WLED with CCT.

deece:
> @deece Hey man I compiled your FW1906 fork and uploaded to my ESP32. I cant see the option for FW1906 in the LED config.

Make sure you're actually flashing the right binary, clear the cache, cookies, etc. I was tearing my hair out over this, and it turned out the binary I was flashing was not the one I just built.

softhack007:
> cant see the option for FW1906 in the LED config.

If you modify UI files in wled00/data, make sure to run `npm install; npm run build` in the console, and then rebuild the firmware. If you skip the "npm" step, modified UI files will not make it into the firmware.

https://kno.wled.ge/advanced/custom-features/#changing-web-ui

oakey81:
> > cant see the option for FW1906 in the LED config.
> 
> If you modify UI files in wled00/data, make sure to run `npm install; npm run build` in the console, and then rebuild the firmware. If you skip the "npm" step, modified UI files will not make it into the firmware.
> 
> https://kno.wled.ge/advanced/custom-features/#changing-web-ui

Hey man thanks for the reply. Im not much of a coder but can manage to compile wled in vscode. I am not quite sure on how i do this at all. Any extra help would be muchly appreciated. Thanks

wled00/const.h
```diff
@@ -232,6 +232,7 @@
 #define TYPE_WS2811_400KHZ       24            //half-speed WS2812 protocol, used by very old WS2811 units
 #define TYPE_TM1829              25
 #define TYPE_UCS8903             26
+#define TYPE_FW1906              28            //RGB + CW + WW + unused channel (6 channels per IC)
```
shnhrrsn, line 235: I think the ordering here is an incorrect assumption that impacts the flexibility of this PR>

FW1906 is just a dual RGB (or, more accurately, BGR) IC — there aren't specific CW/WW pins.

It's up to the manufacture how to connect their LEDs, and unfortunately, they take liberties here.

The FW1906 strip I have, for instance, orders its channels as WW, null, G R B, CW. This ordering, while strange and annoying, actually does make sense — it lines up nicely with the physical ordering of the LEDs on the strip and optimized for the traces vs how humans want to think about it.

I'd think before this PR is super viable to merge, we need the ability to fully customize the channel orders like we can in RGB.

shnhrrsn (commented):
Really appreciate the effort here! I was able to work off of this PR and modify to fit my use case, saved me a lot of time!

wled00/const.h
oakey81: Hey there shnhrrsn. I ordered 200 FW1906 ICs and have designed a driver board utilizing this IC. My only problem is getting the firmware working on my ESP32. I have done the whole npm run but i always get an error. What steps did you do to get this working or do you have a firmware build you could share so i can test my boards. Cheers mate

softhack007: @oakey81 This thread is about line 235 in const.h, and color ordering of FW1906 chips.
please don't drift off-topic.

softhack007:
@oakey81 first you need to install [NodeJS 11.0+](https://nodejs.org/en/download/) globally. Then run `npm install; npm run build` in the VSCode console panel.

https://kno.wled.ge/advanced/custom-features/#changing-web-ui

wled00/const.h
oakey81: Yeah mate i get that but I was just asking how he actually got the firmware compiled properly and working. As he stated he got it working. Anyway this whole thread jumps form one issue to the next if you scroll back a fair way.

softhack007: we call that "code review". In contrast to "support request" ;-)

oakey81:
> @oakey81 first you need to install [NodeJS 11.0+](https://nodejs.org/en/download/) globally. Then run `npm install; npm run build` in the VSCode console panel.
> 
> https://kno.wled.ge/advanced/custom-features/#changing-web-ui
> 
> image

Thanks mate yeah it was the 'npm i' that i missed. sorted now and my setup works perfectly with all colours the right way round. I think some of the strips are not FW1906 IC and thats why the colour orders arent matching up. I actually ordered 200 x FW1906 ICs and have designed my own 5ch driver and RGBCW/WW downlight.

https://github.com/Aircoookie/WLED/assets/83114787/8db4026a-518c-4437-965c-3849a4e7cec5

blazoncek:
@oakey81 please update and rebase for 0_15 branch if you think it is viable solution worth to be included and make a PR.

oakey81:
@blazoncek I am not sure what you mean by rebase as i am not a coder. My mate is very good at it and he saw what i was missing within about 30 sec. I just downloaded the FW1906 code off github and done the npm install and compiled to upload on my ESP32. And as i had designed my own 5ch driver board I followed the manufacturers data sheet and it runs perfectly. I do believe some of the RGBCCT strips might not be FW1906 ICs on them. But to answer your question. Yes I believe this is a viable PR for the next update. And also I see that the WS2805 RGBCCT IC has been released as a true 5ch chip not like the FW1906 6ch chip. So that could be another to add later aswell.

deece:
I believe this was intended for @Robert-github-com . I already rebased and squashed down to a minimal set (see my earlier comments), so hopefully they're isn't much to change from there.

> @blazoncek I am not sure what you mean by rebase as i am not a coder. My mate is very good at it and he saw what i was missing within about 30 sec. I just downloaded the FW1906 code off github and done the npm install and compiled to upload on my ESP32. And as i had designed my own 5ch driver board I followed the manufacturers data sheet and it runs perfectly. I do believe some of the RGBCCT strips might not be FW1906 ICs on them. But to answer your question. Yes I believe this is a viable PR for the next update. And also I see that the WS2805 RGBCCT IC has been released as a true 5ch chip not like the FW1906 6ch chip. So that could be another to add later aswell.

Suxsem:
Hi guys, I'm seeing more and more of those chips out there, any news on official support? Thank you

blazoncek:
0_15

mathojojo:
I just received my FW1906 RGB+CTT ledstrip, and I notice that it's not actually implemented in wled, but it's a work in progress here.
I'm not sure to understand the actual state of the work. Should it work if I build from branch 0_15 ? If yes, I can test it of course, but I'm not sure how to select the correct branch.

Using a wemos mini-S2 board, I already had to compile from GIT sources. But I never had to select a branch at any moment. Is the a git command to select another branch ?

Thank you for the work, and for your help.

deece:
@blazoncek I've created a PR against 0-15 here: https://github.com/Aircoookie/WLED/pull/3810

Suxsem:
Any news on the merge of this one? Thank you

deece:
@blazoncek With the merge of #3810, this can also be closed.

PR #3308, opened 2023-08-08
animated staircase control relay option

djasonic:
Adds control relay option to animated staircase, for staircase state to switch relay on and off for led power supply to save energy when staircase not lit.


blazoncek:
I was hoping you would not choose this path. 😞 

Instead of writing your custom relay logic it would be much better to utilise `toggleOnOff()` to switch WLED on and off instead of directly manipulating relay (and possibly breaking other internal behaviour).

djasonic:
good catch yes will do thank you.

PR #3311, opened 2023-08-09
Effect transitions

Aircoookie:
by @blazoncek 

Enables smooth blending from the previous effect if the effect is changed.


wled00/FX.h
```diff
@@ -381,6 +381,26 @@ typedef struct Segment {
     byte     *data; // effect data pointer
     static uint16_t maxWidth, maxHeight;  // these define matrix width & height (max. segment dimensions)
 
+    typedef struct TemporarySegmentData {
```
Aircoookie, line 384: This is not bad at all, so this is just a suggestion, but wouldn't it be more maintainable to make an anonymous `struct CoreSegmentData` and use it both in `Segment` and `Transition`? All the members could still be accessed like usual (e.g. `seg.speed`) since the anonymous struct has no name. Copying could probably also be simplified from how it is implemented now in `saveSegenv()` as the whole struct could be copied in one line (not sure if it is possible to do that with anonymous structs though).

wled00/FX_fcn.cpp
```diff
@@ -470,6 +557,7 @@ void Segment::setMode(uint8_t fx, bool loadDefaults) {
         sOpt = extractModeDefaults(fx, "mY");   if (sOpt >= 0) mirror_y  = (bool)sOpt; // NOTE: setting this option is a risky business
         sOpt = extractModeDefaults(fx, "pal");  if (sOpt >= 0) setPalette(sOpt); //else setPalette(0);
       }
+      markForReset();
```
Aircoookie, line 560: This is a good change,  resetting mid-transition like in `main` led to quite a few issues (new effect allocating data just for it to reset and re-allocate mid-transition), so I will probably change it there too for b4. Should probably be un-commented in `WS2812FX::setMode` as well.

wled00/FX_fcn.cpp
```diff
@@ -1079,13 +1170,15 @@ void WS2812FX::service() {
   _segment_index = 0;
   Segment::handleRandomPalette(); // move it into for loop when each segment has individual random palette
   for (segment &seg : _segments) {
+    if (!seg.isActive()) continue;
```
Aircoookie, line 1173: Continuing before `seg.resetIfRequired()` causes segment data to not be freed if the segment is deleted. Could this be the cause of the leak you observed?

Aircoookie (commented):
Transitions look amazing!

softhack007:
@blazoncek I've tested the new code on my "portable 16x8" equipment - looks good 👍 

I just noticed a problem with the "Matrix" effect when "global LEDs buffer" is disabled. Only the top row is changing, all other rows  stay dark. Not sure if this is related to fx-blending, or maybe its a general bug. Will re-test with my bigger setup in the next days, to see if I can confirm the problem on a second ESP32.

blazoncek:
> I just noticed a problem with the "Matrix" effect when "global LEDs buffer" is disabled.

For fun create a gap file (to disable a few LEDs) and observe the outcome. 😁 

Matrix is flawed. In many ways.

softhack007:
> For fun create a gap file (to disable a few LEDs) and observe the outcome. 😁

I'll try that 😁 

Just took a quick look at the Matrix effect code, and it seems it will 100% fail without absolutely lossless getPixelColor().
https://github.com/Aircoookie/WLED/blob/ba1b6f321ede6b578ebc86c6ff8ee9032b1a2b32/wled00/FX.cpp#L5136-L5137

Maybe let's consider Matrix a hopeless case, and get back to it once our universes have imploded, or when one of us is totally bored 😝 whatever comes first ;-)

blazoncek:
> Maybe let's consider Matrix a hopeless case, and get back to it once our universes have imploded, or when one of us is totally bored 😝 whatever comes first ;-)

I want the lady in red and a steak. 😉

blazoncek:
@softhack007 as @Aircoookie is currently busy do you agree to merge this and then create a new beta release?

PR #3317, opened 2023-08-13
Add send and receive groups to json api

shlomozippel:
I had a requirement to be able to set the send and receive groups dynamically based on a switch so I added the send & receive group bitmasks to the API / serialized state

### DIFF
--- a/wled00/json.cpp
+++ b/wled00/json.cpp
@@ -347,7 +347,9 @@ bool deserializeState(JsonObject root, byte callMode, byte presetId)
 
   JsonObject udpn      = root["udpn"];
   notifyDirect         = udpn["send"] | notifyDirect;
+  syncGroups           = udpn["sgrp"] | syncGroups;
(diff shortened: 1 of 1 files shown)


blazoncek:
No UI? Just API?

shlomozippel:
There already is a UI for the group mask in the Sync settings screen

blazoncek:
I know, I wrote it. 😄
I was wondering why do you need this API?

shlomozippel:
This is for an art car at burning man - it has about 30 controllers on it running WLED. I wanted them all to have identical configuration and only change the send / recv group based on a switch. That way if controllers need to be swapped out it's easy to do and change their group with DIP switch or similar. Basically avoid requiring a phone / tablet / computer whenever possible.

This PR let me set the group mask with a preset that was activated by GPIO

blazoncek:
If you want this merged then please update WLED knowledge base as well.
[WLED-Docs](https://github.com/Aircoookie/WLED-docs)

PR #3320, opened 2023-08-14
Semi adjustable NTP polling interval.

### LINKED ISSUE #3285: Adjustable NTP polling interval
Sorry if I missed a config but assume NTP is used to set the internal HW time IRQ. I've noticed my animations briefly pause/freeze when NTP is enabled (found tip to disable NTP from QuinLED). Is it possible to allow the user provides a recommended NTP background polling interval in the form of hh:mm? I assume, after device boots, a fixed number of NTP requests will set time, then the user's hh:mm interval would kink in to maintain time. This would prevent the need to disable NTP to prevent animations from freezing/pausing. Please let me know if there is a different solution. Thanks

blazoncek:
Fixes #3285


WoodyLetsCode:
Thanks for your work! ❤️ 
I could not reproduce #3285.
However, I think this PR would not be a fix for #3285 because he said he _must somehow be in the second condition with aggressive NTP retry_.
I also think the hardcoded polling interval of 11 hours and 40 minutes should be more than enough.
I think a semi-adjustable NTP polling interval would result in most people not knowing what to set it to.
Even I don't know why I should change it.

blazoncek:
I agree with you completely. And there was no feedback on the PR so I may as well close it.

PR #3345, opened 2023-08-25
2D effect improvements: centered 1D->2D Arc mapping, restored 2DFire2012

apanteleev:
This PR contains two changes:

1. The Arc mapping of 1D to 2D effects was only drawing, in fact, an arc - and leaving some uninitialized pixels beyond its radius. This change makes it draw full circles concentric with the segment, as long as the segment dimensions are even.

2. Ported the 2DFire2012 effect from the Sound Reactive fork, added slider controls for speed, cooling and sparking. Looks like it was implemented here before but disappeared someday? Not sure, but I don't see a reason to remove it.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -4766,6 +4766,64 @@ uint16_t mode_2DDrift() {              // By: Stepko   https://editor.soulmateli
 static const char _data_FX_MODE_2DDRIFT[] PROGMEM = "Drift@Rotation speed,Blur amount;;!;2";
 
 
+/////////////////////////
+//     2D Fire2012     //
+/////////////////////////
+
+uint16_t mode_2Dfire2012(void) {                // Fire2012 by Mark Kriegsman. Converted to WLED by Andrew Tuline. Ported from WLED-SR by Alexey Panteleev.
+  if (!strip.isMatrix) return mode_static(); // not a 2D set-up
+
+  const uint8_t COOLING = SEGMENT.intensity;
+  const uint8_t SPARKING = SEGMENT.custom1;
+
+  CRGBPalette16 currentPalette  = CRGBPalette16( CRGB::Black, CRGB::Red, CRGB::Orange, CRGB::Yellow);
+
+  if (millis() - SEGENV.step >= ((256-SEGMENT.speed) >>2)) {
+    SEGENV.step = millis();
+    // static byte *heat = (uint16_t *)dataStore;
+
+    if (!SEGENV.allocateData(sizeof(byte) * 4096)) return mode_static(); //allocation failed
+    byte *heat = reinterpret_cast<byte*>(SEGENV.data);
+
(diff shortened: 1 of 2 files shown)


blazoncek:
There is no need for either IMO.
1. just toggle mirroring in segment
2. use bar expansion in Fire 2012 effect

softhack007:
I'm not sure about changing "ARC" mapping to something completely different. We would break existing effect presets, where people (rightfully) assumed that "ARC" produces an arc not a circle.
However I still find your improvements to avoid gaps interesting. It see you basically use a "thicker pen" (two pixels towards inside) for overwriting gaps?

Maybe its better to
1. improve "ARC" with parts of your proposal, however leave it as a real ARC coming from one corner. 
2. Add a new mapping mode like "round", that will be centered. @blazoncek is right this could be created also with ARC + mirror horiz + mirrow vert. But maybe 

Not sure about the back-ported effect - "Fire 2012" also works on 2D if you select "BAR" for mapping (and it does look good). Would the original "2D Fire 2012" still look much different? If yes, then maybe we could add it.

What do you think?

blazoncek:
IMO there is no need to add circular expansion as same can be achieved by mirroring in X and Y dimensions. Originally @ewoudwijma developed circular expansion but I intentionally reduced it to arc as circular can be achieved by mirroring.

@softhack007 I fail to see how "thicker" this is? It is just painting 4 quadrants. Besenham's algorithm is faster but introduces gaps. That would benefit by painting 2 pixels.

There is no need to port 2D Fire from SR IMO.

softhack007:
> @softhack007 I fail to see how "thicker" this is? It is just painting 4 quadrants. Besenham's algorithm is faster but introduces gaps. That would benefit by painting 2 pixels.

You could be right, it seems i've misunderstood what the code does here.

softhack007:
> There is no need to port 2D Fire from SR IMO.

I have compared your port of "Fire 2012 2D" to the "Fire 2012" effect with "bar" mapping - Screenshots (preview) below. 

Actually I see almost no visual difference. Just that the "vanilla WLED" version allows to select palettes, and it seems in "vanilla" there is an additional blur step that slightly reduces frame-rates.

@apanteleev could you clarify how your 


_vanilla WLED "Fire 2012" with 2D expansion mode `Bar`_
image


_Your port of "2D Fire 2012"_
image

blazoncek:
Yes, I added `blur()` to 2D variant to make it more visually pleasing.
We can add option to enable/disable it if you are interested in fastest FPS possible.

blazoncek:
Reviewing it for the second time I see no added value in this PR.
I propose to close it without merging.

apanteleev:
Sorry about the delay in replying to this PR; I was off the grid for some time. The reason this PR is adding duplicate functionality is simply that I missed that the mirroring mode exists, and somehow didn't test the 1D fire effect / didn't expect it to do what it does.

Closing.

PR #3348, opened 2023-08-27
staircase toggle power on/off option bool

djasonic:
Refactored to check offMode and call toggleOnOff() to make the led power relay switch on when starting swipe on and switch off upon completion of a swipe off, and also toggle back on if disabling the staircase and it was off at the time.


blazoncek (approved):
This seems ok to me.

PR #3351, opened 2023-08-29
show current local time in info page :timer_clock: 

### LINKED ISSUE #3284: Add device current time to Info page
Not sure why, but the NTP feature doesn't always seem to update to the correct time without doing multiple reboots. Adding a current device time to the Info page allows a quick time check without going all the way into the Time & Macros page.

PSandro:
resolves #3284

Adding the current local time to the info page :timer_clock: . This allows to check the device time quickly without having to navigate to the Time & Macros Settings page :tada: 

Best,
Sandro


blazoncek:
Please remove all (.h) files that have no connection with the PR.
This includes all html_*.h except html_ui.h

blazoncek (approved):
All good.
I would prefer shorter string instead of "Current local time". Perhaps just "Time"?

PSandro:
> All good. I would prefer shorter string instead of "Current local time". Perhaps just "Time"?

I would also prefer 'Time'. My initial idea was to keep it consistent with the 'Time & Macros' page. But if that is not important, then let's change it to 'Time' :+1:

PSandro:
image
This is how it looks now

softhack007:
Change looks ok for me, too. Would agree to merge.

PR #3356, opened 2023-09-05
Introduce Autobrightness Usermod

mmaciola:
Autobrightness usermod by Michal Maciola

The usermod will use an analog photoresistor for ambient light measurement. In response to reading changes, led brightness will be adjusted.

The autobrightness usermod is using an array of overlapping buckets that defines min and max lux values, along with suggested brightness level. When the ambient light changes, a new bucket is identified. The proportional difference, between the previous brightness and new suggested brightness value, is kept.
To minimize sensitivity to short-term changes in lighting, debounce functionality has been implemented.

### DIFF
--- a/usermods/Autobrightness_with_photoresistor/readme.md
+++ b/usermods/Autobrightness_with_photoresistor/readme.md
@@ -1,0 +1,27 @@
+# Autobrightness usermod
+
+Autobrightness usermod by Michal Maciola
+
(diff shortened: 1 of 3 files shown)


blazoncek (changes requested):
Still I'd like you to change a few things before it is ready to be merged.
Not using runtime configuration (compile time options) is a thing of the past and needs to be changed.
Please see other usermods how to implement runtime configuration.

Issue #3370, opened 2023-09-12
Ripple Effect broken on matrix


poelzi:
### What happened?

When the LEDs are configured as matrix, the ripple effect behaves flickery

### To Reproduce Bug

1. Configure strip as matrix
2. Use one segment with ripple effect

### Expected Behavior

1d & 2d strip should look similar

### Install Method

Self-Compiled

### What version of WLED?

ESP32

### Which microcontroller/board are you seeing the problem on?

ESP32

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

blazoncek:
Ripple effect has been upgraded to work on 2D as ripples on the water.
It is intended that way.

If you want the same behaviour set up segment as 1D.

blazoncek:
https://github.com/Aircoookie/WLED/assets/7722392/49f6d9e2-f16a-4e28-aed6-e199e72b733d

softhack007:
> the ripple effect behaves flickery

Ripple is a very "fast" effect, so quick changes in pixel brightness could appear as flickering, especially on a matrix setup.

If you have several segments, each segment has its own brightness slider. I've noticed that changing segment brightness (aka opacity) is sometimes stronger compared to using the global brightness slider. 

So one thing you could try: Set segment brightness to max, and use the main brightness slider for adjusting your lights. Make sure you have "Use global LED buffer" enabled in LED settings, and check if the "max current" for brightness limiter is adequate for your fixture.

blazoncek:
We may tweak 2D ripples to add trail to lessen the apparent flickering.
What do you think @softhack007 ?

softhack007:
> We may tweak 2D ripples to add trail to lessen the apparent flickering. What do you think @softhack007 ?

I can check the effect on my larger matrix tomorrow, but yeah why not add a small fade_out() as option, similar to what we did for the rollingBalls effect.

blazoncek:
@poelzi can you try https://github.com/Aircoookie/WLED/commit/87cc3fd714fe7c5a3dd806f8dbd529493f171d9d and see if it behaves any better?

blazoncek:
Fix available in b6.

PR #3373, opened 2023-09-15
2D Ripple and Meteor effect tweak

blazoncek:
for #3370

Renders less chaotically on smaller 2D matrices.


wled00/FX.cpp
```diff
@@ -2400,9 +2400,10 @@ uint16_t ripple_base()
 
       #ifndef WLED_DISABLE_2D
       if (SEGMENT.is2D()) {
+        propI /= 2;
         uint16_t cx = rippleorigin >> 8;
         uint16_t cy = rippleorigin & 0xFF;
-        uint8_t mag = scale8(cubicwave8((propF>>2)), amp);
+        uint8_t mag = scale8(sin8((propF>>2)), amp);
         if (propI > 0) SEGMENT.draw_circle(cx, cy, propI, color_blend(SEGMENT.getPixelColorXY(cx + propI, cy), col, mag));
```
softhack007, line 2407: What's the difference between cubicwave8() and sin8()?

blazoncek: As per documentation cubicwave is similar to sin but has a faster rise and stays ~1 longer (closer to square wave).
It seems to me that behaviour is better with sin in 2D variant (less flickering).

softhack007:
thanks, looks good 👍 
Will merge.

PR #3377, opened 2023-09-17
Fix ES8388Source & ES7243 initialization.

christianpatterson:
Update ES8388Source::initialize and ES7243::initialize method signatures to match I2SSource::initialize so that when initialize is called on a AudioSource pointer the child class's method is used.

### DIFF
--- a/usermods/audioreactive/audio_source.h
+++ b/usermods/audioreactive/audio_source.h
@@ -192,6 +192,7 @@ class I2SSource : public AudioSource {
     }
 
     virtual void initialize(int8_t i2swsPin = I2S_PIN_NO_CHANGE, int8_t i2ssdPin = I2S_PIN_NO_CHANGE, int8_t i2sckPin = I2S_PIN_NO_CHANGE, int8_t mclkPin = I2S_PIN_NO_CHANGE) {
+      DEBUGSR_PRINTLN("I2SSource:: initialize().");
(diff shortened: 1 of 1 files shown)


blazoncek:
IMO it should be the other way around.
You should remove two obsolete parameters in I2SSource.
Looks like we forgot to do that.

softhack007:
> You should remove two obsolete parameters in I2SSource.
> Looks like we forgot to do that.

I agree with @blazoncek. The only place where this specific `initialize` is called, is here (and likewise in `case 2` for es7243)

https://github.com/Aircoookie/WLED/blob/527e3d6cd07509fa50a06e16d3a99dd89148d54e/usermods/audioreactive/audio_reactive.h#L1164-L1169

The two extra parameters were meant for SDA and SDL (I2C) but with the chance to only use global I2C pins, these are not needed any more.

christianpatterson:
Updated, but I was only able to test it with ES8388 and Generic I2S since that's all the hardware I have.

blazoncek:
I think this is also a problem @netmindz was facing when moving upstream code to MM.

netmindz:
Move was actually in the other direction. Removal of I2C from source broke MM as I think the global I2C hadn't actually been merged into MM at that stage

softhack007:
thanks for this PR :-) well spotted 👍

PR #3380, opened 2023-09-21
Allow 9 LED pins with audioreactive

softhack007:
In response to user feedback - people seem to miss their last two led pins in the audioreactive version. 

NPB cannot use `I2S#0` because audioreactive needs it. But `I2S#1` is still availeable on esp32. So this PR adds`I2S#1` to the digital led driver, allowing for 9 instead of 8 pins.

@blazoncek please review, and if OK for you then I propose to merge it into main. I've (intensively) tested that running audio and in parallel using `I2S#1` for NPB does work.

### DIFF
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -44,8 +44,8 @@
       #define WLED_MIN_VIRTUAL_BUSSES 4
     #else
       #if defined(USERMOD_AUDIOREACTIVE)      // requested by @softhack007 https://github.com/blazoncek/WLED/issues/33
-        #define WLED_MAX_BUSSES 8
-        #define WLED_MIN_VIRTUAL_BUSSES 2
+        #define WLED_MAX_BUSSES 9
+        #define WLED_MIN_VIRTUAL_BUSSES 1
       #else
         #define WLED_MAX_BUSSES 10
         #define WLED_MIN_VIRTUAL_BUSSES 0
--- a/wled00/bus_wrapper.h
+++ b/wled00/bus_wrapper.h
@@ -1177,8 +1177,12 @@ class PolyBus {
       uint8_t offset = 0; //0 = RMT (num 0-7) 8 = I2S0 9 = I2S1
       #if defined(CONFIG_IDF_TARGET_ESP32S2)
       // ESP32-S2 only has 4 RMT channels
-      if (num > 4) return I_NONE;
-      if (num > 3) offset = 1;  // only one I2S
+      #if defined(USERMOD_AUDIOREACTIVE)      // softhack007 I2S#0 is reserved for audio input
+        if (num > 3) return I_NONE;
+      #else
+        if (num > 4) return I_NONE;
+        if (num > 3) offset = 1;  // only one I2S
+      #endif
       #elif defined(CONFIG_IDF_TARGET_ESP32C3)
       // On ESP32-C3 only the first 2 RMT channels are usable for transmitting
       if (num > 1) return I_NONE;
@@ -1189,8 +1193,13 @@ class PolyBus {
       //if (num > 3) offset = num -4; // I2S not supported yet
       #else
       // standard ESP32 has 8 RMT and 2 I2S channels
-      if (num > 9) return I_NONE;
-      if (num > 7) offset = num -7;
+      #if defined(USERMOD_AUDIOREACTIVE)      // softhack007 I2S#0 is reserved for audio input, but I2S#1 is availeavle as 9th LED pin
+        if (num > 8) return I_NONE;
+        if (num == 8) offset = 2;
+      #else
+        if (num > 9) return I_NONE;
+        if (num > 7) offset = num -7;
+      #endif
       #endif
       switch (busType) {
         case TYPE_WS2812_1CH_X3:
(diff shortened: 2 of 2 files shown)


blazoncek (commented):
Don't get me wrong but if it is possible I'd much rather see we do not add `#ifdef USERMOD_AUDIOREACTIVE` into lowest level file (wrapper).

In fact why would you want to restrain LED buses if audio is compiled in? It may be used as a network receiving device only without any I2S device attached.

Instead I would prefer that usermod would inform user (in settings) that he/she is using I2S device for LEDs which is prohibiting audio input functionality (if he/she is using too many outputs).
IMO that would be much cleaner approach which would allow same binary to be used with 10 LED outputs and no audio input or 9 LED outputs and audio input.

softhack007:
The reason for `#ifdef USSERMOD_AUDIOREACTIVE` in bus_wrapper is that I need to modify the rules, so bus nr 9 will go to `I2S#1` instead of `I2S#0` in normal WLED without AR.

> In fact why would you want to restrain LED buses if audio is compiled in? It may be used as a network receiving device only without any I2S device attached.

The problem is that audioreactive setup does not check if I2S#0 is in use for LEDs, so we decided last year to restrict the user from 10 to 8 led pins, to avoid conflicts. 

I don't know how AR setup could detect that NPB NeoEsp32I*2s0*800KbpsMethod is in use - maybe there is a way. I can't say exactly what will happen when AR initialized I2S#0 while in use for LEDs - maybe we will get a crash, or maybe leds on pin 9 will just switch of.

> IMO that would be much cleaner approach which would allow same binary to be used with 10 LED outputs and no audio input or 9 LED outputs and audio input.

Agreed it will be a cleaner solution, but with a lot more changes needed compared to this easy improvement. Maybe the clean solution could be implemented in the next major release.

softhack007:
Another idea: if when change the default order, so pin9=I2S#1 and pin10=I2S#0, then the would be no need for #ifdef in bus_wrapper.

wled00/const.h
```diff
@@ -44,8 +44,8 @@
       #define WLED_MIN_VIRTUAL_BUSSES 4
     #else
       #if defined(USERMOD_AUDIOREACTIVE)      // requested by @softhack007 https://github.com/blazoncek/WLED/issues/33
-        #define WLED_MAX_BUSSES 8
-        #define WLED_MIN_VIRTUAL_BUSSES 2
+        #define WLED_MAX_BUSSES 9
```
blazoncek, line 47: This `#ifdef` may be unnecessary as well. Just allow 10 digital buses.

A check in Audioreactive's `setup()` could be added as follows:
```c++
int digitalBuses = 0;
for (int i=0; i<busses.getNumbuses(); i++) if (busses[i]->getType() >= 16 && busses[i]->getType() <= 31) digitalBuses++;
if (digitalBuses>9) return;
```

softhack007: Thanks, that check is actually an easy way to verify. I would still need a new #ifdef in audioreactive, because a similar check for bus 5 is needed for -S2.

I'm still not sure it's a good solution from user experience point of view. Because AR would simply fail to initialize, without information what is wrong. Maybe I'll need to add a flag and put something into info page, but space there is limited so we needed a short but absolutely clear message text.

Need to think about it ...

blazoncek: There is no problem if usermod requires core dependency, but core should not be dependant on usermods. That's how I see it.
Info dialog can be used with a simple note: "AR Inactive: Too many LED outputs used."

And perhaps instead of returning in `setup()` set `enabled` to `false`.

softhack007:
@blazoncek I'm still wondering if its good to change UI and core behaviour like this, literally 1 week before a release is scheduled.
Maybe this PR is better suited for 0.15 🤔 ??

wled00/const.h
blazoncek: No, as you do not want SPI variants included.
const.h defines bus types and anything 16>=x<=31 is single pin digital.
This may change of course in the future, but such magic numbers are also used in JS for UI so there is no way around them.

softhack007:
> 0.14.1 may be better suited if you want to push it faster.

OK, then let's postpone this one, and add it to 0.14.1. The current situation with 8 digital LED pins is not optimal, however starting to revise user interactions now is also not a good idea IMHO.

wled00/const.h
netmindz: Even if a magic number needs to be defined in both C and JavaScript, they should still be constants not magic numbers

netmindz:
> > Still waiting for @netmindz response, but once @Aircoookie approves new collaborators (if he does) he may be able to work on it directly.

What are you waiting on me for?

softhack007:
> What are you waiting on me for?

I think @blazoncek was referring to his new proposal for audioreactive palettes.

wled00/const.h
blazoncek: The *magic* numbers are defined in const.h. Look at LED types `TYPE_xxxxx`.
JS does not have such definitions. Yet.

softhack007:
superseded by commit https://github.com/Aircoookie/WLED/commit/20ed81cd86a7eb3ee1ffcfad4d7c3b2cf84d2a3a (0_15 branch).

blazoncek:
@softhack007 I did not ask if you are ok with the change (removal of #ifdef) but I was considering using audioreactive as a standard build in a near future and remove non-audio binary. What do you think?

Issue #3400, opened 2023-09-29
Sun Times Return Zero


tonyn0:
### What happened?

Sometimes, sun times are calculated as zero (depending on location, I hear). Requesting that "use real math" is implemented as standard.

### To Reproduce Bug

Be in wrong place at wrong time. ;)

### Expected Behavior

Calculations are always valid.

### Install Method

Binary from WLED.me

### What version of WLED?

b6

### Which microcontroller/board are you seeing the problem on?

Other

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

blazoncek:
Partial workaround is by custom compiling using `-D WLED_USE_REAL_MATH`.
The drawback is that image size will increase and also a few bytes of precious RAM will be missing.

The former may be an ESP32 issue and the latter ESP8266 issue.

softhack007:
I'll propose a proper fix tomorrow. It seems that reverting `_t` math to standard functions in getSunriseUTC() will make it work. Reason is that _t functions are slightly faster, but only use deliver low accuracy values.

From Discord
> Wondering if I found a bug. This was working fine on v5. My sundown keeps reverting to 0's. I can get the sundown to pop up once I change something, but after I save it and leave the page, it reverts back.

![Screenshot_2023-09-28_095530](https://github.com/Aircoookie/WLED/assets/91616163/adccbdae-0c14-418f-b312-eccde5ecac5d)

softhack007:
fixed in 0.14.0

softhack007:
See https://github.com/Aircoookie/WLED/issues/3601.
A more robust solution will be availeable with release 0.14.1

PR #3406, opened 2023-09-30
fix for #3400

softhack007:
replace low_accuracy math functions (sint_t, cos_t, atan_t, ...) with standard libm functions that have higher accuracy.

Needs a bit more testing before merging into main.


blazoncek:
I cannot reject or approve as the decision to use custom math functions was done by @Aircoookie 
My original implementation used clib math functions.

softhack007:
Update: looks good after some testing - picked a few random locations, and compared to official sunrise/sunset times on the internet. Accurate up to 1 minute 👍

PR #3407, opened 2023-10-01
Add information about which branch is being used 

Moustachauve:
Add the information about which software branch, beta or stable, is being used on this device. This is accessible through both the JSON API and MQTT, just like the version string. This will automatically detect if it is a beta or not based on the release version configured in package.json to make sure it is always set properly.

This could be really useful for any consumer of the API that wants to make sure to check the correct channel to know potentially if a device is using the most recent version, for example and also to match it to the correct branch (beta, soundreactive, ...).

Recreated from a topic branch to follow the project's guidelines.

### DIFF
--- a/pio-scripts/set_version.py
+++ b/pio-scripts/set_version.py
@@ -6,3 +6,7 @@ PACKAGE_FILE = "package.json"
 with open(PACKAGE_FILE, "r") as package:
     version = json.load(package)["version"]
     env.Append(BUILD_FLAGS=[f"-DWLED_VERSION={version}"])
+    if "-b" in version:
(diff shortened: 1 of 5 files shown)


Moustachauve:
cc @blazoncek and @softhack007.

blazoncek:
Perhaps this should be combined with #3839

blazoncek:
> I think those are two different things. The same product (What #3839 is about) could be used for stable or beta branches.

That was exactly the point. Combine the two in one solution.

blazoncek:
This has been superseded by "release" key in "info" object in JSON API in 0.15.0-b3 or later.

Moustachauve:
Cool thanks. I couldn't find a PR for it, I thought there have been multiple ones in the past, I guess it was not linked to one?

PR #3414, opened 2023-10-03
added the max fan pct to the PWM-Fan usermod. 

DanCaveman:
This allows high speed (loud fans) like delta fans to cap at a maximum percentage to keep noise down.

### DIFF
--- a/usermods/PWM_fan/usermod_PWM_fan.h
+++ b/usermods/PWM_fan/usermod_PWM_fan.h
@@ -159,7 +159,7 @@ class PWMFanUsermod : public Usermod {
     void setFanPWMbasedOnTemperature(void) {
       float temp = getActualTemperature();
       float difftemp = temp - targetTemperature;
-      // Default to run fan at full speed.
+      // Default to run fan at full speed (or maxPWMPct Defined by the user).      
       int newPWMvalue = 255;
       int pwmStep = ((maxPWMValuePct - minPWMValuePct) * newPWMvalue) / (7*100);
       int pwmMinimumValue = (minPWMValuePct * newPWMvalue) / 100;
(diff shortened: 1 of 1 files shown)


blazoncek:
Will be included in 0.14.1

usermods/PWM_fan/usermod_PWM_fan.h
```diff
@@ -179,6 +181,8 @@ class PWMFanUsermod : public Usermod {
         newPWMvalue = pwmMinimumValue + 5*pwmStep;
       } else if (difftemp <= 3.0) {
         newPWMvalue = pwmMinimumValue + 6*pwmStep;
+      } else {
```
blazoncek, line 184: It would be much cleaner (and produce smaller code) to use something like:
```c++
int steps = 7;
if (difftemp <= ..) {
  steps = ...
} else if (...) {
  steps = ...
}
newPWMValue = pwmMinimumValue + steps * pwmStep;
```

DanCaveman: Great Idea - and I got to thinking that you could just calculate the steps without doing an if statement.

Would something like this be more beneficial or too complicated:

```cpp
    void setFanPWMbasedOnTemperature(void) {
      float temp = getActualTemperature();
      // dividing minPercent and maxPercent into equal pwmvalue sizes
      int pwmStepSize = ((maxPWMValuePct - minPWMValuePct) * _pwmMaxValue) / (_pwmMaxStepCount*100);
      int pwmStep = calculatePwmStep(temp - targetTemperature);
      // minimum based on full speed - not entered MaxPercent 
      int pwmMinimumValue = (minPWMValuePct * _pwmMaxValue) / 100;
      updateFanSpeed(pwmMinimumValue + pwmStep*pwmStepSize);
    }

    uint8_t calculatePwmStep(float diffTemp){
      if ((diffTemp == NAN) || (diffTemp <= -100.0)) {
        DEBUG_PRINTLN(F("WARNING: no temperature value available. Cannot do temperature control. Will set PWM fan to 255."));
        return _pwmMaxStepCount;
      }
      if(diffTemp <=0){
        return 0;
      }
      int calculatedStep = (diffTemp / _pwmTempStepSize)+1;
      // anything greater than max stepcount gets max 
      return (uint8_t)min((int)_pwmMaxStepCount,calculatedStep);      
    }

```

PR #3420, opened 2023-10-04
Remove duplicate `viewport` declaration

mountainash:
Removed some redundant code in the HTML

Reasons:
- width should resize to fit the viewport/device (as the next line allows)
- has no effect (next line overrides it)
- reduce filesize

Addition:
- removed unneeded self-closing slash
- [lang moved to the correct HTML element](https://github.com/Aircoookie/WLED/pull/3420/commits/b5751795b576ff9a4c10a1fbbc2cb11df72df14f)

### DIFF
--- a/wled00/data/settings.htm
+++ b/wled00/data/settings.htm
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
 	<meta charset="UTF-8">
-	<meta content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" name="viewport"/>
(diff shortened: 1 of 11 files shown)


mountainash:
@WLED I'm planning on doing a bit of a clean up of the HTML, is that OK? Should I base it off `main` branch, or is there a newer branch?

softhack007:
> @wled I'm planning on doing a bit of a clean up of the HTML, is that OK? Should I base it off `main` branch, or is there a newer branch?

Actually this won't make it into the upcoming release of 0.14.0, due to lack of testing on all possible environments.
So my suggestion is to do cleanup work based on `0_15` branch which is for new developments. @blazoncek what's your view?

blazoncek:
Seems ok to me.

PR #3423, opened 2023-10-05
CSS optimisations

mountainash:
- removed unit type on `0` values
- removed leading `0` on decimal values
- removed unused and single use var's
- value spacing consistency
- `border: none` replaced with `border: 0`
- `outline: none` replaced with `outline: 0`
- consistent tab indenting
- removed commented-out styles (they haven't been re-enabled for 2+ years)

### DIFF
--- a/wled00/data/simple.css
+++ b/wled00/data/simple.css
@@ -80,7 +80,7 @@ a, a:visited {
 }
 
 button {
-	outline: none;
+	outline: 0;
 	cursor: pointer;
 	background-color: transparent;
 	border: 0;
@@ -91,7 +91,7 @@ button {
 	min-height: 40px;
 }
 button:hover {
-    background: var(--c-4);
+	background: var(--c-4);
 }
 
 .label {
@@ -293,7 +293,7 @@ button:hover {
 }
 
 #ndlt {
-  margin: 12px 0;
+	margin: 12px 0;
 }
 
 #roverstar {
@@ -320,7 +320,6 @@ button:hover {
 }
 
 #kv, #kn {
-	/*max-width: 490px;*/
 	display: inline-block;
 }
 
@@ -330,7 +329,7 @@ button:hover {
 }
 
 #info td, #nodes td {
-  padding-bottom: 8px;
+	padding-bottom: 8px;
 }
 
 #info .btn {
@@ -346,7 +345,7 @@ button:hover {
 }
 
 #kn td {
-  padding-bottom: 12px;
+	padding-bottom: 12px;
 }
 
 #heart {
@@ -385,7 +384,9 @@ img {
 .sliderdisplay {
 	content:'';
 	position: absolute;
-	top: 10px; left: 8px; right: 8px;
+	top: 10px;
+	left: 8px;
+	right: 8px;
 	height: 8px;
 	background: var(--c-4);
 	border-radius: 16px;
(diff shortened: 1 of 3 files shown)


wled00/data/simple.css
```diff
@@ -180,15 +174,15 @@ button:hover {
...
+	border: 0;
+	transition: color .3s, background-color .3s;
 	font-size: 17px;
 	color: var(--c-c);
 	min-width: 44px;
 }
 
 .top button {
-	padding: var(--tbp);
+	padding: 14px 8px 10px;
```
blazoncek, line 185: This alone is dealbreaker. Please revert the change.

wled00/data/simple.css
```diff
@@ -492,26 +486,19 @@ input[type=range]::-moz-range-thumb {
...
 }
 
-.hd {
-	display: var(--bhd);
-}
-
 #briwrap {
 	float: right;
-	margin-top: var(--bmt);
+	margin-top: 0;
```
blazoncek, line 501: This one as well.

wled00/data/simple.css
```diff
@@ -19,19 +19,13 @@
...
-	--c-tbh: rgba(51, 51, 51, var(--t-b));
-	/*following are internal*/
+	--c-o: rgba(34, 34, 34, .9);
+	--c-tb : rgba(34, 34, 34, .5);
+	--c-tba: rgba(102, 102, 102, .5);
+	--c-tbh: rgba(51, 51, 51, .5);
 	--th: 70px;
 	--tp: 70px;
 	--bh: 63px;
-	--tbp: 14px 8px 10px;
```
blazoncek, line 31: Why do you want to remove these?

blazoncek (changes requested):
I welcome your effort to optimize CSS (even though linter & minimizer might do the same) but removing parametric styles is out of the question.

wled00/data/simple.css
mountainash: Happy to revert, but can you tell me the reason for this decision?

wled00/data/simple.css
mountainash: Optimisation - it's a process overhead (small) and filesize overhead (small too) for no point. It's also a developer readability overhead. These variables are only uses once. 

If they were reused multiple times - I'd have left them in.

mountainash:
When built, the `const uint16_t PAGE_index_L` value in `html_ui.h` drops from `32012` to `31940`. In `html_simple.h` it goes down from `18069` to `18027`.

wled00/data/simple.css
blazoncek: What makes you say that? Have you combed the sources?
They get modified during runtime by JS.

mountainash: Agghh yuk... I see that now. For the changes between light & dark mode... not nice that this is done in JS (it can/should be done solely in CSS).

Thank you 🙇🏼 for pointing that out (it's good to have a Maintainer that's aware of all the parts of the project).

I'm working on a new "look" 💅🏼 for the Web UI and will aim to remove/reduce these.

blazoncek:
Do not be deterred by my comments or complaints.
I am in favour of updating UI but it is a **major** task since UI is a fundamental part of what WLED is.

It has to be *perfect* on first try.

PR #3424, opened 2023-10-05
📦 Nodemon version bump & package-lock update

mountainash:
_See commit messages_

Moved nodemon to `devDependencies` as it's only used for local dev.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,11 +21,13 @@
     "url": "https://github.com/Aircoookie/WLED/issues"
   },
   "homepage": "https://github.com/Aircoookie/WLED#readme",
+  "devDependencies": {
+    "nodemon": "^2.0.20"
+  },
   "dependencies": {
     "clean-css": "^4.2.3",
     "html-minifier-terser": "^5.1.1",
     "inliner": "^1.13.1",
-    "nodemon": "^2.0.20",
     "zlib": "^1.0.5"
   }
 }
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "homepage": "https://github.com/Aircoookie/WLED#readme",
   "devDependencies": {
-    "nodemon": "^2.0.20"
+    "nodemon": "^3.0.1"
   },
   "dependencies": {
     "clean-css": "^4.2.3",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,277 +1,372 @@
 {
   "name": "wled",
   "version": "0.14.0-b6",
-  "lockfileVersion": 1,
+  "lockfileVersion": 3,
   "requires": true,
-  "dependencies": {
-    "abbrev": {
+  "packages": {
+    "": {
+      "name": "wled",
+      "version": "0.14.0-b6",
+      "license": "ISC",
+      "dependencies": {
+        "clean-css": "^4.2.3",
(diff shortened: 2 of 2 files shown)


softhack007:
What's the added value of this PR?
Also it seems you modified some integrity hashes, can you point us to the sites where we can verify the new hashes? Just to be sure that we did not accidentally import malware.

mountainash:
The added value is that it's the recommended way of declaring dependencies used solely for development.

The updates to the lock file were not manual. NPM (v10.1.0 - as installed with Node.JS v20.8.0) updated the lock file.

Skimming through the log file the it looks like only a few hashes to changed, from `sha1` to `sha512`. Most of the changes are due to the lock file version `lockfileVersion` changing from version 1 to 3. All the, non nodemon, URLs appear to pull from the same location; if this changed it would be more reason for concern. I don't [believe NPM allows](https://docs.npmjs.com/policies/unpublish) reuse of the same URLs when files change.

If you are concerned about change; I suggest using a Codespaces VM on the `main` branch, copy the `package.json` from this request (to update nodemon) and run `npm install`. Then compare the lockfile changes generated with this PR.

Another option/PR would be to replace Nodemon and just use Node `--watch` as per https://www.jamesqquick.com/blog/using-node-watch-instead-of-nodemon/

Moustachauve:
+1 for this PR. I was about to create the same PR.

package.json
```diff
@@ -21,11 +21,13 @@
     "url": "https://github.com/Aircoookie/WLED/issues"
   },
   "homepage": "https://github.com/Aircoookie/WLED#readme",
+  "devDependencies": {
+    "nodemon": "^3.0.1"
```
Moustachauve, line 25: Why only move nodemon to devdependencies? In theory, all of the dependencies are dev only, nothing gets put on the device itself after build.

Moustachauve (approved):

package.json
mountainash: There's a difference in the dependencies used for "development", and the tools used to "build". The dependencies left in the `dependencies` declaration (as opposed to `devDependencies` declaration), as used during the build process.

Moustachauve: That's one way of seeing it. It could work. However, I am more used to see it as`dependencies` is anything required in production environments while `devDependencies` is anything needed for the project development, which includes build tools.

In any cases, there is no real point in implementing a separation between devDependencies or normal dependencies right now as we do not use any library in production. All of them are either build tools or dev tools and so we do not require installing only a certain group. They will all not be included in prod.

mountainash: This change "grouping" could speed up the build, as 'nodemon' wouldn't be needed for the build - and it's likely the build tool isn't using the `devDependencies`.

This PR is also for a version bump of nodemon.

mountainash:
> What's the added value of this PR? Also it seems you modified some integrity hashes, can you point us to the sites where we can verify the new hashes? Just to be sure that we did not accidentally import malware.

Easy to check for yourself. `rm package-lock.json`, then `npm install` and you'll get similar updates - although now some of the non-pin sub-sub-dependencies will likely have been updated and you'll get new versions, with new hashes. So this doesn't guarantee that "did not accidentally import malware" - unless you've checked very dependency inuse by the project I doubt you could even say that this isn't already the case (depending on your boundires/definition of malware).

PR #3425, opened 2023-10-05
.gitignore Reordered & grouped

mountainash:
- logical groupings
- alphabetical sorting
- consistent directory format


blazoncek (changes requested):
Why are you removing these from GH repo and some other entries in .gitignore?
They are mandatory for building binaries and keeping local repos "local".

Unfortunately I cannot approve this.

mountainash:
Sorry... I miss-understood https://github.com/Aircoookie/WLED/pull/3421#issuecomment-1748045725 as I thought that the changes should not be included, but I guess changes should be included when appropriate. I've reverted the new filter, but kept the rest of the tidy-up for your approval.

blazoncek (changes requested):
I do not want to be seen as a prick but I still have to say that removal of items is not desired or approved.

If you want to reorder things, fine, if you want to add things, fine, but not remove as this will expose unwanted files to the next commit.

mountainash:
> removal of items is not desired or approved

Nothing has been removed. [32ab2ae](https://github.com/Aircoookie/WLED/pull/3425/commits/32ab2ae201b86da61a332edb339ee77021299f9c) reverts the removal. It's only ordering that's changed now.

softhack007:
what is the added value of re-ordering entries in this file , as its only needed for git internally?

mountainash:
> what is the added value of re-ordering entries in this file , as its only needed for git internally?

Readability

blazoncek (approved):
I'm ok with it.

PR #3426, opened 2023-10-05
Optimised images

mountainash:
Screenshot 2023-10-05 at 12 25 43 pm

No lossy quality changes, just stripping meta data.

## TODO
- [x] change the photos in `usermods/TTGO-T-Display/assets/` to JPEGs instead of PNGs


blazoncek:
I am not sure what is the purpose of this PR?
These images are only used for readme.md document visible in Github.

softhack007:
> I am not sure what is the purpose of this PR?

Some question from my side. Does your optimization change the size of WLED firmware ? if yes, how much?

mountainash:
File size reduction are generally a good thing - in this case it makes it faster/less expensive for other Developers and Git duplication services to download the project - with no visible loss. Also some of these images are used on https://kno.wled.ge/ which would make browsing faster for visitors (_they may need to be copied over to another branch?_).

Issue #3382, opened 2023-09-23
SyntaxError and sometimes a reset when changing presets in 2D.


MOVE1904:
### What happened?

I made an illuminated sign for the back under my jersey for my favorite club.

![1](https://github.com/Aircoookie/WLED/assets/43117741/452e5028-b19e-48ed-b7fa-e14329618ce8)

There are 540 LEDs arranged irregularly on the sign. I configured WLED so that I use this in a 2D Matrix.
When I choose a presets I often got a syntax error or a reset of the ESP32, even with the simplest configuration.

https://github.com/Aircoookie/WLED/assets/43117741/fbf7413b-02f8-4a3f-a8ca-c2a9140e20a4

### To Reproduce Bug

Use the attached cfg and preset file.

[preset_cfg.zip](https://github.com/Aircoookie/WLED/files/12706610/preset_cfg.zip)

Select the “Start” preset several times.

### Expected Behavior

I expected that there would be no reset or error when changing the preset.

### Install Method

Self-Compiled

### What version of WLED?

WLED v0.14.0-b5 "Hoshi" (Build 2309100)

### Which microcontroller/board are you seeing the problem on?

ESP32, ESP32-S3

### Relevant log/trace output

```shell
---- Opened the serial port COM4 ----
WS FileRead: /index.htm
Not-Found HTTP call:
URI: /skin.css
WS FileRead: /skin.css
JSON buffer locked. (17)
JSON buffer size: 7890 for request: 6
JSON content length: 5285
JSON buffer released. (17)
JSON buffer locked. (17)
JSON buffer size: 4801 for request: 8
JSON content length: 2184
JSON buffer released. (17)
JSON buffer locked. (17)
JSON buffer size: 11798 for request: 3
JSON content length: 6665
JSON buffer released. (17)
Not-Found HTTP call:
URI: /presets.json
WS FileRead: /presets.json
WS client connected.
JSON buffer locked. (12)
JSON buffer size: 10967 for WS request (6665).
heap 181664
Sending WS data to a single client.
JSON buffer released. (12)
JSON buffer locked. (14)
JSON buffer released. (14)
JSON buffer locked. (17)
JSON buffer size: 11799 for request: 3
JSON content length: 6673
JSON buffer released. (17)
JSON buffer locked. (12)
JSON buffer size: 10968 for WS request (6673).
heap 154840
Sending WS data to multiple clients.
JSON buffer released. (12)
Loop took 45ms.
Usermods took 0ms.
[...]

PR #3427, opened 2023-10-05
workaround for #3382

softhack007:
On esp32, its possible that two threads (async_tcp, looptask) are sending out JSON via AsyncWebServer in parallel 
- serveJson() responds to json/si, and 
- updateInterfaces() is sending JSON state info using sendDataWs(). 

Assumed root cause: parts of AsyncWebServer are not thread-safe, and the payloads from a second ->send() call may corrupt the previous one if still being sent out. The problem should be solved properly in AsyncWebServer, however until that happens, this PR reduces the frequency of errors.

The workaround uses the INTERFACE_UPDATE_COOLDOWN timer  to delay the second webSocket send command so it does not (usually) interfere with the first.

Changes tested on esp32 with the example config & preset from bug report #3382.
Testing on 8266 still to be done.


blazoncek:
@softhack007 this has been superseded by JSON buffer guard in 0.15. This PR should be no longer necessary.

PR #3428, opened 2023-10-05
Added "Akemi' PWA + Favicon

mountainash:
Added a `webmanifest` to add some extra style when user's attempt to add this "site" to their Homescreen (making it a "PWA"). The favicon has been updated to match the apple-touch-icon and includes both a 16x16 and 32x32 size for better display.

### Installed on iOS Homescreen
Screenshot 2023-10-05 at 3 40 29 pm
Also shows current WLED iOS App, and current WLED PWA "W"

### Share Plane on iOS Safari
WLED

### Share Plane on macOS Safari
Screenshot 2023-10-05 at 3 16 37 pm

### Favicon in Arc sidebar
Favicon in Arc sidebar

## Other changes
- trailing space removal
- consistent HTML link statements


mountainash:
Logo from https://github.com/Aircoookie/Akemi/blob/master/akemi/001_cheerful.png

softhack007:
Hi,

please undo the "trailing space removal" from this PR. We prefer to have a clear change set, as its simplifies code review.

blazoncek:
This is not going to pass unless @Aircoookie himself approves.

Aircoookie (changes requested):
Thank you for the PR!
I approve of adding PWA capability in principle, it is a useful addition.
However, please remove the extra whitespace changes.

Also, it might be better to scale down the Akemi icon to 192 or 256 to reduce storage use by a bit.
It should be supported to specify the icon with a base64 string, therefore saving an extra network request:
https://support.google.com/webmasters/thread/6112971/in-pwa-manifest-file-can-we-give-the-datauri-as-the-src-of-an-icon?hl=en

Please try this out if you have the time. Otherwise I might adjust the myself but it would likely take a while until i get around to it.

mountainash:
It's seems a bit of a contradiction that the request is to lower the image quality to reduce storage, but at the same time also attempt to inline an image using base64 which is known to increase the size of the image, and to a lesser degree the white-space changes (which can add to the file size unnecessarily).

In response to the 1 additional network request - I think you'll find that it's a request that's made very seldomly... once when the user chooses to add the site to their home-screen and occasionally after to re-cache. But I think you'll find that these requests are already happening anyway... but now they are returning a 404. 🤷🏼 

> It should be supported to specify the icon with a base64 string, therefore saving an extra network request:
> https://support.google.com/webmasters/thread/6112971/in-pwa-manifest-file-can-we-give-the-datauri-as-the-src-of-an-icon?hl=en

[In the comments that forum post](https://support.google.com/webmasters/thread/6112971?hl=en&msgid=6181056), the poster says that it doesn't work.

Either way, this PR allows for Maintainers to edit it as they see fit. I'll leave it as it is.

Issue #3404, opened 2023-09-30
Add legend or hover tooltip for filters


tablatronix:
These are a bit ambiguous, it would help new users to know what these meean. A simple alt tag might do the trick, or a legend somewhere in UI

![Screenshot 2023-09-30 at 9 25 26 AM](https://github.com/Aircoookie/WLED/assets/807787/c5e010b3-9334-4743-9d39-d736d5b63170)

Either with tooltip class

or just a title

```html
<label id="filterFreq" class="check fchkl" title="TITLE">♫ <input type="checkbox" data-flt="♫" onchange="filterFx(this)"> <span class="checkmark"></span> 
</label>
```

![Screenshot 2023-09-30 at 9 45 53 AM](https://github.com/Aircoookie/WLED/assets/807787/bd5c47fc-7395-4429-b16a-2969674a46d1)

tablatronix:
I am not even sure what these flags mean. Not sure its even documented fyi.

https://kno.wled.ge/features/effects/

blazoncek:
If you figure out correct CSS I will gladly add it.

tablatronix:
Looks like there is a tooltip class

```html
<span id="sliderLabel0" class="tooltiptext">Effect speed</span>
```

tablatronix:
This will require a lot of adjustments because of the sticky and overflow visibility and z index etc.

![Screenshot 2023-09-30 at 2 15 04 PM](https://github.com/Aircoookie/WLED/assets/807787/c8a4d0c3-2b6b-4586-aeb8-a763a2aa19ea)

blazoncek:
Fill the search field and test on iOs/Android as well, please.

blazoncek:
When you have it ready you can make a PR.
https://github.com/Aircoookie/WLED/wiki/How-to-properly-submit-a-PR

If you do not want to bother just submit what changed.
Thank you.

blazoncek:
Any updates?

tablatronix:
Doing it now

tablatronix:
Need to figure out how to compile first, having an issue with 
(project configuration file): 'No option 'ldscript_1m0m' in section: 'common''

tablatronix:
How tf do I upload the index.htm, I have been uploading spiffs and nothing changes..

blazoncek:
https://kno.wled.ge/advanced/custom-features/#changing-web-ui

PR #3439, opened 2023-10-13
Add legend or hover tooltip for filters (#3404)

tablatronix:
Adds tooltips to filter checkboxes

### DIFF
--- a/wled00/data/index.css
+++ b/wled00/data/index.css
@@ -461,7 +461,7 @@ button {
 
 .filter {
 	z-index: 1;
-	overflow: hidden;
+	overflow: visible;
 }
 
 /* Tooltip text */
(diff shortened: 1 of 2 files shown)


wled00/data/index.css
```diff
@@ -461,7 +461,7 @@ button {
 
 .filter {
 	z-index: 1;
-	overflow: hidden;
+	overflow: visible;
```
blazoncek, line 464: I am not sure this change is appropriate as we do not want filter bar to wrap into two rows.

tablatronix: Yeah I didn't this it was a good idea, but saw nothing break, alternative is to add another wrapper and change its size to be tall enough for the tooltips maybe but might have to then account for the top bar option toggle

blazoncek: Screenshot 2023-10-14 at 18 36 24
Not good.

blazoncek:
Correctly implemented in 0_15 branch.

PR #3441, opened 2023-10-14
Allow product information to be specified at build time

Moustachauve:
This allow people who make custom builds of WLED (Ex: QuinLED) to specify the details about the product at build time rather than having it hardcoded. The default values of WLED are unchanged.

This is similar to #3407 and has a similar end goal.

### DIFF
--- a/usermods/PIR_sensor_switch/usermod_PIR_sensor_switch.h
+++ b/usermods/PIR_sensor_switch/usermod_PIR_sensor_switch.h
@@ -299,8 +299,8 @@ void PIRsensorSwitch::publishHomeAssistantAutodiscovery()
     JsonObject device = doc.createNestedObject(F("device")); // attach the sensor to the same device
     device[F("name")] = serverDescription;
     device[F("ids")]  = String(F("wled-sensor-")) + mqttClientID;
-    device[F("mf")]   = "WLED";
-    device[F("mdl")]  = F("FOSS");
+    device[F("mf")]   = WLED_BRAND;
+    device[F("mdl")]  = WLED_PRODUCT_NAME;
     device[F("sw")]   = versionString;
     
     sprintf_P(buf, PSTR("homeassistant/binary_sensor/%s/config"), uid);
--- a/usermods/PIR_sensor_switch/usermod_PIR_sensor_switch.h
+++ b/usermods/PIR_sensor_switch/usermod_PIR_sensor_switch.h
@@ -299,8 +299,8 @@ void PIRsensorSwitch::publishHomeAssistantAutodiscovery()
     JsonObject device = doc.createNestedObject(F("device")); // attach the sensor to the same device
     device[F("name")] = serverDescription;
     device[F("ids")]  = String(F("wled-sensor-")) + mqttClientID;
-    device[F("mf")]   = WLED_BRAND;
-    device[F("mdl")]  = WLED_PRODUCT_NAME;
+    device[F("mf")]   = F(WLED_BRAND);
+    device[F("mdl")]  = F(WLED_PRODUCT_NAME);
     device[F("sw")]   = versionString;
     
     sprintf_P(buf, PSTR("homeassistant/binary_sensor/%s/config"), uid);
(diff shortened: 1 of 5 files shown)


wled00/wled.h
```diff
@@ -10,6 +10,19 @@
...
 #define VERSION 2310130
 
+// You can define custom product info from build flags.
+// This is useful to allow API consumer to identify what type of WLED version
+// they are interacting with.
+// Use like this:
+// -D WLED_BRAND="F(\"Custom Brand\")"
+// -D WLED_PRODUCT_NAME="F(\"Custom Product\")"
+#ifndef WLED_BRAND
+  #define WLED_BRAND F("WLED")
```
blazoncek, line 20: I would prefer F() macro was not used in #define but rather in JSON assignment.

usermods/PIR_sensor_switch/usermod_PIR_sensor_switch.h
```diff
@@ -299,8 +299,8 @@ void PIRsensorSwitch::publishHomeAssistantAutodiscovery()
     JsonObject device = doc.createNestedObject(F("device")); // attach the sensor to the same device
     device[F("name")] = serverDescription;
     device[F("ids")]  = String(F("wled-sensor-")) + mqttClientID;
-    device[F("mf")]   = "WLED";
-    device[F("mdl")]  = F("FOSS");
+    device[F("mf")]   = WLED_BRAND;
```
blazoncek, line 302: This may break existing set-ups (and MQTT integrations). It is not needed for the purposes of determining what configuration the binary was built with.

I would avoid changing MQTT configurations if possible.

blazoncek (changes requested):
IMO the goal is OK but the implementation is not. Changing existing behaviour will break functionality at some point.
I think a better approach would be a new API for this specific purpose.

Moustachauve:
> IMO the goal is OK but the implementation is not. Changing existing behaviour will break functionality at some point. I think a better approach would be a new API for this specific purpose.

The existing behavior is not changed at all in my opinion. The API result will be exactly the same as before. 
3rd party devices could already change that value at compile time if they wanted (that's why the value is in the API afterall)

Moustachauve:
I would argue that for most people editing these fields, making sure these are not detected as WLED would be a benefit and not a bug.

usermods/PIR_sensor_switch/usermod_PIR_sensor_switch.h
Moustachauve: MQTT configuration doesn't change for WLED. It stays mf = WLED and mdl = FOSS. 

This allows people who **do not want** mqtt to return WLED and FOSS. 
Also, this is in a usermod. 

You don't need to worry about this breaking anything, because nothing WLED will ever change this value. 
People who make custom build of WLED were already able to modify this value if they wished. I am just making it easier to do so. 

If someone had a custom build that already had a custom brand set, this will not break their integration. 
If someone had a normal build, this will not break any integration because it is the same value. Nothing changed for them.

Moustachauve:
> I also do not know which 3rd party systems use these fields for their purpose (HomeAssistant and possibly others?) in determining connected systems. Please research which 3rd party systems use "mf" and "mdl" JSON API fields.

mf and mdl are only in usermods. No third party rely on it for 99% of the cases. These values are already modifiable. Why prevent people from modifying them if they want to? This does not affect default builds of WLED in any way. I can add more documentation to say to be careful when modifying these values but that's not something default WLED ever has to worry about.

> Still, I cannot approve the change in MQTT as I know the fields are needed for correct classification of HomeAssistant devices detected. This will certainly break existing configurations.

I just checked both in the code of Home Assistant and by testing with my own instance of HA and I am happy to report to you that Home Assistant does not use MQTT at all for WLED and also the device has full functionality.

wled00/wled.h
Moustachauve: Done!

blazoncek (approved):
👍

PR #3445, opened 2023-10-16
Update response code when access is denied

Moustachauve:
Having the wrong pin would result in a server error (500). 
The more appropriate error code for that would be 401.
 This also changes the page that asks for users to login from 200 to 401.


blazoncek (approved):
Excellent. Thank you.

PR #3450, opened 2023-10-17
Added #HH for only Hour and ##MM for only Minute in Scrolling Text, s…

knarfd:
…o you can add the hour and minute where you want with maybe without "scrolling", if you want to have a steady show of the hour and minute (especially if you have only 16x16 2d Matrix it comes handy), otherweise it scrolls when hour >=10 even if the font size is the smallest one


blazoncek:
Thanks. Can you re-base this PR to 0_14_1 branch?

knarfd:
Hopefully done O.O

tobsinger:
is there a way to set the text to be displayed with leading "0"? 
so "08" instead of "8"?

blazoncek:
Append 0 to template. I.e. `#HH0`

PR #3460, opened 2023-10-19
Fixed broken Alphacoders hosted assest asset links

mountainash:
Plus:
- whitespace clean up
- removed past Easter event

_3 different commits so Maintainers can cherry-pick as desired_


blazoncek:
Thank you. Can you re-base this PR to 0_14_1 branch?

PR #3461, opened 2023-10-19
added github bug report button and copy debug information button, Closes #2120

### LINKED ISSUE #2120: Add a github report feature
It would be nice to have a way to copy version/platform info in your preferred format for pasting into github reports without having to adjust the user-select None setting manually. e.g. 

v0.13.0-b0 "Toki"
2106302
esp8266 2_7_4_7 (1)

WoodyLetsCode:
Hi!

This adds a _Submit bug report_ button and _Copy debug information_ button to the security settings at the bottom.
grafik

When you click on the button, it opens a new issue with the bug report template and the field **What version of WLED?** is already filled out. In the field **Relevant log/trace output** you can find debug information like this:

```
## Debug information ##

WLED version: 0.14.0 (build 2310130)
LEDs:
	Count: 144
	Power: 1111
	FPS: 5
	Max power: 2000
	Max segment: 32
UDP Port: 21324
WiFi:
	RSSI: -77
	Signal: 46
	Channel: 1
Architecture: esp32
Core version: v3.3.6-16-gcc5440f6a2
Free heap: 199124
Uptime: 221
Time: 2023-10-19, 00:51:42
Opt: 79
Brand: WLED
Product: FOSS
MAC Address: c8c9a3cb9f10
IP Address: 172.16.16.169
```
This is also the debug information you get when clicking on the _Copy debug information_ button.

grafik

This closes #2120


blazoncek:
I would prefer sending entire `cfg.json` and/or `presets.json` instead of parsed partial data.
Parsing adds quite a few bytes to total image size and may lack important information.

I'll let @Aircoookie decide.

softhack007:
> I would prefer sending entire `cfg.json` and/or `presets.json` instead of parsed partial data.

Agreed, cfg.json and pesets.json are needed in many cases to make a problem reproducible. 

From the data proposed by @WoodyLetsCode, I think that `architecture`, `core` and `heap` (free heap?) could be useful. `MAC adress` and `ip address` however are very sensitive, and most user would not share them for good reason.

WoodyLetsCode:
I think you're right, it would be better sending entire `cfg.json` and `presets.json`.
Another thing I've just thought about: Should `wledUiCfg` from the localStorage also be sent?

And should I wait for @Aircoookie 's feedback or should I change it now?

WoodyLetsCode:
@blazoncek Since @Aircoookie is not responding, what do you think about just changing it to send the entire `cfg.json` and `presets.json` and then merging this pull request? If @Aircoookie does not like it that way, we can always change it later.

But in general, he should like this PR since he added #2120 to the 1.0.0 milestone.

WoodyLetsCode:
I ended up not sending `presets.json` because if you have 3-5 presets it is too much for GitHub:
Screenshot 2024-02-01 at 00-57-49 Bad request · GitHub

Now i trimmed it down to:
```
WLED version: 0.15.0-a0 (build 2401270)
Architecture: esp32
Core: v3.3.6-16-gcc5440f6a2
Free heap: 183648



cfg.json:
{"rev":[1,0],"vid":2401270, ...
```

PR #3464, opened 2023-10-19
UI tooltips (interface improvements)

ajotanc:
- Improvement in the creation of tooltips
With this improvement, it is possible to create a tooltip using the tooltip="Message you want to show" attribute on any element, so JavaScript already understands it automatically, follow some examples below.

**Buttons, Inputs, span**

image

image
image
image

However, for other elements that already have the tooltip with `tooltiptext`, a problem occurs if the line is removed or commented out, when it is compiled it ends up not showing the input ranges of the effects and gives the error `r is null` and ends up breaking the code, the solution to use until you understand why is leaving the `tooltiptext` elements with the `hide` class, so it worked smoothly after being compiled.

image


blazoncek:
To remove `tooltiptext` span-s you will need to modify `setEffectParameters()` function in index.js.
It updates tooltips with appropriate content.

ajotanc:
> To remove `tooltiptext` span-s you will need to modify `setEffectParameters()` function in index.js. It updates tooltips with appropriate content.

Correction of the sliders in the `setEffectParameters()` function to eliminate the need to update the old tooltip span.

wled00/data/index.js
```diff
@@ -1505,13 +1505,8 @@ function setEffectParameters(idx)
 	let nSliders = 5;
 	for (let i=0; i<nSliders; i++) {
 		var slider = gId("slider" + i);
-		var label = gId("sliderLabel" + i);
 		// if (not controlDefined and for AC speed or intensity and for SR all sliders) or slider has a value
 		if ((!controlDefined && i < ((idx<128)?2:nSliders)) || (slOnOff.length>i && slOnOff[i] != "")) {
-			if (slOnOff.length>i && slOnOff[i]!="!") label.innerHTML = slOnOff[i];
```
blazoncek, line 1511: Removal of these is unacceptable.
Instead you should update the content of tooltip text.

blazoncek (changes requested):
You should not remove existing feature.

ajotanc:
> You should not remove existing feature.

Messages showing correctly according to their respective effect.

blazoncek:
Unfortunately options (checkmarks at the bottom of effects tab) are not yet using the new feature.
When you solve that, you can remove `.tooltiptext` from CSS IMO.

ajotanc:
> Unfortunately options (checkmarks at the bottom of effects tab) are not yet using the new feature. When you solve that, you can remove `.tooltiptext` from CSS IMO.

I'm aware of this, I haven't changed the others yet because I'm waiting for a response so I can remove all the `tooltiptext`, if you say it can, I'll remove it and leave just the new model.

ajotanc:
Corrections made, everything works as it should.

Remembering that for correct use:

`tooltip="Message"`

For input `checkbox`, add the attribute in `span.checkmark`
For `buttons`, `div,` `input range`, add the attribute to itself

readme.md
```diff
@@ -3,7 +3,7 @@
   <a href="https://github.com/Aircoookie/WLED/releases"><img src="https://img.shields.io/github/release/Aircoookie/WLED.svg?style=flat-square"></a>
   <a href="https://raw.githubusercontent.com/Aircoookie/WLED/master/LICENSE"><img src="https://img.shields.io/github/license/Aircoookie/wled?color=blue&style=flat-square"></a>
   <a href="https://wled.discourse.group"><img src="https://img.shields.io/discourse/topics?colorB=blue&label=forum&server=https%3A%2F%2Fwled.discourse.group%2F&style=flat-square"></a>
-  <a href="https://discord.gg/A2FpqFQaRS"><img src="https://img.shields.io/discord/473448917040758787.svg?colorB=blue&label=discord&style=flat-square"></a>
+  <a href="https://discord.gg/QAh7wJHrRM"><img src="https://img.shields.io/discord/473448917040758787.svg?colorB=blue&label=discord&style=flat-square"></a>
```
blazoncek, line 6: Why is this change?

readme.md
```diff
@@ -66,7 +66,7 @@ Credits [here](https://kno.wled.ge/about/contributors/)!
 
 Join the Discord server to discuss everything about WLED!
 
-<a href="https://discord.gg/A2FpqFQaRS"><img src="https://discordapp.com/api/guilds/473448917040758787/widget.png?style=banner2" width="25%"></a>
+<a href="https://discord.gg/QAh7wJHrRM"><img src="https://discordapp.com/api/guilds/473448917040758787/widget.png?style=banner2" width="25%"></a>
```
blazoncek, line 69: And this?

blazoncek (approved):
I fixed some things and made the change even better at smaller footprint.

readme.md
ajotanc: I hadn't even seen that this file was uploaded, it came from when I made the clone, I didn't change it... But I'm going to remove it, hehe

ajotanc:
Another thing... I left the tooltip in span.checkmark because of this... It is not exactly centered on the object because it has the icon, so it is in the middle of the check and icon.

image

blazoncek:
The multiplier does not work well across different browsers.

ajotanc:
> The multiplier does not work well across different browsers.

I tested it in Firefox and Opera and they were fine. We can look at another way, or leave the value fixed at 0.4, which for me I think is enough.

ajotanc:
Very cool, now it looks 100%, hahaha! It turned out really well, thanks for the adjustments!

PR #3466, opened 2023-10-20
Fix jumping of selected effect on refresh

WoodyLetsCode:
Hi!
This fixes the problem that the selected effect jumps to the bottom and back on a refresh.

For example, if you select the effect rainbow and there is a refresh, it jumps from this:
grafik
To this:
grafik
And then back.

To reproduce this, just select the rainbow effect, go to Info and click on the refresh button a few times. Then you should see it in the background.


blazoncek:
Thanks. Can you re-base this PR to 0_14_1 branch?

WoodyLetsCode:
Done

Issue #3467, opened 2023-10-21
Active preset after 0.14.0


R1DEN:
### What happened?

Before the update if I turned off the light (by using the button in the upper left corner, or via API), the previously selected preset was still active/selected/reported in Home Assistant.
After the update, as soon as I turn off the lights - the preset is disabled/inactive/not reported in Home Assistant.
The settings themselves are not changed, but the preset is deselected in the UI.

### To Reproduce Bug

1) pick any preset
2) turn wled off via the button

### Expected Behavior

The preset should still be active (both in UI and in Home Assistant) as it was previously.

### Install Method

Binary from WLED.me

### What version of WLED?

0.14.0

### Which microcontroller/board are you seeing the problem on?

ESP32

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

blazoncek:
Not a bug but intentional behaviour.
Since presets can contain "on" state any change in UI (or from API) that causes state change (including "on" state) is considered *preset invalidator* (active preset is no longer valid).

R1DEN:
@blazoncek so was it considered a bug in all the previous versions?

R1DEN:
It seems to me that reverting it to the way it was provides more flexibility to the API user, as you could always check if "on" is false or not to determine if the preset is actually working or not.
But without the preset being reported in the "off" state - you cannot predict what would be the active preset in case you turn WLED on. Furthermore when turning it of and on again the preset is not restored, despite the light output being the exact same as it was when the preset was active.

UrTekNerd:
I discovered a new issue with WLED 0.14.0. First I would like to say Thank you for WLED. I love it. The issue I discovered is when every you run a playlist it wipes out all your segments. If I stop the playlist and just set a color with an effect it's fine. I had three segments and as soon as I started the playlist all my segments were gone. I was hoping to run 4 different playlists on 4 different segments.  Thank you.

blazoncek:
> I had three segments and as soon as I started the playlist all my segments were gone. I was hoping to run 4 different playlists on 4 different segments.  Thank you.

Segments are stored in presets, not playlist. Playlist does not remove segments, presets can.

If you are using ESP8266 segments can be removed by core functionality due to low memory condition. In such case ESP may also diconnect all clients and reconnect to the WiFi.
Please check how much available heap there is using Info page.

PR #3479, opened 2023-10-23
don't consider turning off/on as state change for preset resetting

R1DEN:
should also revert to the behavior observed prior to 0.14.0

### DIFF
--- a/wled00/led.cpp
+++ b/wled00/led.cpp
@@ -63,7 +63,6 @@ void toggleOnOff()
     briLast = bri;
     bri = 0;
   }
-  stateChanged = true;
 }
 
 


blazoncek:
Unfortunately this doesn't get my approval. For the reasons posted in #3467

R1DEN:
@blazoncek sorry for being a bit stubborn, but what would be the core issue with reverting things the way they were? As I've described [here](https://github.com/Aircoookie/WLED/issues/3467#issuecomment-1774721722)

blazoncek:
Removing `stateChanged = true;` will prohibit interface update (including sending WS notifications, Alexa, etc).

R1DEN:
@softhack007 hey, since you added the discussion label, maybe you have any ideas? Given your knowledge or wled internal workings and this [comment](https://github.com/Aircoookie/WLED/issues/3467#issuecomment-1774721722)

danlex-corp:
@blazoncek I would like to add about saving the active preset when turned off. After upgrading to version 0.14, I observe the following behavior:
When you turn on the VLED through a button in the interface or API, the active preset is not restored as a parameter (ps), it remains “-1”, although the behavior of the tape itself is restored and as if the preset is active, but this is only visible visually.
It was very convenient to use the current active preset in Home Assistant scenarios, but when you turn it on/off, the preset is logically reset.

blazoncek:
@danlex-corp since preset *also stores* WLED on state when you turn WLED on or off using UI power button you *invalidate* the currently selected preset. Unfortunately there is no way of knowing if other parameters changed as well without introducing additional variables to store entire state and previous state. This is unacceptable on ESP8266.

danlex-corp:
@blazoncek I removed "on" state from the presets. This doesn't help either. As I understand it, now the active preset is not saved at all when turned off? How did this work before on version 0.13?

R1DEN:
Isn't it possible to just not invalidate the preset when turning it off/on? And invalidate it on all other changes? I don't see the need for additional variables (apart from the toggledOnOff in this PR) if the invalidation check can be done on the fly.
Can confirm this did work in 0.13

Aircoookie:
I prefer the pre-0.14 behavior too, on/off should stay in preset just as changing the master brightness does.

However, the implementation with `toggledOnOff` is incorrect, as when other state is updated simultaneously with on/off, preset should be reset, but is incorrectly kept. Just removing `stateChanged = true;` in `toggleOnOff()` as originally implemented in https://github.com/Aircoookie/WLED/pull/3479/commits/9fbbad50f2ebbf8631054d1383480b3151dbcbc8 should be fine though.

@blazoncek
> Removing `stateChanged = true;` will prohibit interface update (including sending WS notifications, Alexa, etc).

It shouldn't, I think? `bri != briOld` in [led.cpp L103](https://github.com/Aircoookie/WLED/blob/7e17011ebc8e5f9dacea22ec6462f24ba2e10295/wled00/led.cpp#L103) also covers the case of on/off.

blazoncek:
`led.cpp` is a mess and an burdensome legacy (and you know it). 😁 
And all the logic behind it is a relic from the past (`handleTransition()` & `stateUpdated()`). It took me a year to figure out that convoluted logic it harbours. 😂 

BTW `toggleOnOff()` will not initiate `stateUpdated()` so if you want to start a transition and do change brightness, you either need `stateChanged = true;` or call `stateUpdated()` from it.

Regardless I still think that when you toggle WLED on/off using UI power button you are invalidating a preset. Why? Have a preset with `win&T=0` (or `{"on":false}`) and use it instead of power button. What happens? Now change it to `win&T=2`. What happens then?

FYI In 0_15 you can override UI power button to actually call a preset instead of performing plain on/off.

blazoncek:
BTW when talking about global brightness (and on/off) I keep getting reminded of #2984 and the pain global brightness brings with it.

R1DEN:
Any chance this could be implemented?

blazoncek:
Not in this form.

R1DEN:
Sure, that's not an issue. Maybe you could provide me some more context, info, links to "hidden dev/architecture rules" that are not in the contribution guide or something so that I could remake this PR?

softhack007:
> links to "hidden dev/architecture rules" that are not in the contribution guide

Hi @R1DEN, actually we don't have these documents. however to keep the overall design and architecture understandable and maintainable, it's our job as maintainers to discuss your proposed PR with you, and make sure it fits to the overall architecture. 

In some industries this role is called "design authority". We don't have enough people to maintain an up-to-date architectural description document, so our way is to engage in design discussions when a PR is presented for review.

I haope you understand. It's not about avoiding mistakes, but about keeping the wled design consistent.

R1DEN:
@softhack007 perfectly understandable. So is the information already in the PR comment sufficient for me to try come up with a different solution? Or would you (or anybody else) add something?

softhack007:
> @softhack007 perfectly understandable. So is the information already in the PR comment sufficient for me to try come up with a different solution? Or would you (or anybody else) add something?

As led.cpp is not really "my area", I would pass this question on to @blazoncek and @Aircoookie.

I do agree with comments from @blazoncek that the code in led.cpp is quite tricky - any change may have unexpected side-effects - so maybe a different solution would be preferable from maintainability perspective.

blazoncek:
I veto this in any case.
For the reasons stated above.
But since WLED is not my project @Aircoookie may decide otherwise.

PR #3483, opened 2023-10-24
Update Pixel Magic Tool

ajotanc:
Reducing the size of HTML/CSS/Javascript


ajotanc:
- Correction in animation generation that was actually breaking
- Button to return to WLED if pxmagic.htm is accessed via url (http://), if accessed via external (file://) the button will not show
- Adjustments to CSS and some features

ajotanc:
- CSS adjustments
- Removal of some html
- Correction of CORS error when using the file locally, it was unable to generate the images that are saved in WLED
- Among others that I don't remember now

PR #3484, opened 2023-10-25
Show 2D Configuration button by default

WoodyLetsCode:
The problem was that when the settings page loaded, the "2D Configuration" button appeared a few milliseconds later than the other buttons. This meant that sometimes you would accidentally make incorrect input. If you were fast and wanted to press a button quickly, the button would appear just before you pressed it, causing the rest of the buttons to move as well, and you would press the wrong button.


blazoncek:
I prefer it the other way around: showing an option when it is present, not hiding an option when it is not present.
If you understand what I mean. 😁

PR #3489, opened 2023-10-27
Show clock overlay only if all LEDs are solid black

### LINKED ISSUE #984: Feature request: Automatically turn off clock overlay
Add a configuration option to turn off the analog clock overlay as soon as an effect (i.e. anything besides "all pixels black") is started.

WoodyLetsCode:
This adds an option to show clock overlay only when all LEDs are solid black.
grafik

When an effect is active, Clock Overlay is not displayed:
grafik

If the color is not black, the clock overlay is not shown:
grafik

If the effect is solid and the color is black, the clock overlay is shown:
grafik

This closes #984.

### DIFF
--- a/wled00/overlay.cpp
+++ b/wled00/overlay.cpp
@@ -89,6 +89,15 @@ void _overlayAnalogCountdown()
 
 void handleOverlayDraw() {
   usermods.handleOverlayDraw();
+  if (analogClockSolidBlack) {
+    Segment* segments = strip.getSegments();
+    for (uint8_t i = 0; i < strip.getActiveSegmentsNum(); i++) {
+      Segment segment = segments[i];
+      if (segment.mode > 0 || segment.colors[0] > 0) {
+        return;
+      }
+    }
+  }
   if (overlayCurrent == 1) _overlayAnalogClock();
 }
 
--- a/wled00/overlay.cpp
+++ b/wled00/overlay.cpp
@@ -91,8 +91,9 @@ void handleOverlayDraw() {
   usermods.handleOverlayDraw();
   if (analogClockSolidBlack) {
     const Segment* segments = strip.getSegments();
-    for (uint8_t i = 0; i < strip.getActiveSegmentsNum(); i++) {
+    for (uint8_t i = 0; i < strip.getSegmentsNum(); i++) {
       const Segment& segment = segments[i];
+      if (!segment.isActive()) continue;
       if (segment.mode > 0 || segment.colors[0] > 0) {
         return;
       }
--- a/wled00/overlay.cpp
+++ b/wled00/overlay.cpp
@@ -90,9 +90,9 @@ void _overlayAnalogCountdown()
 void handleOverlayDraw() {
   usermods.handleOverlayDraw();
   if (analogClockSolidBlack) {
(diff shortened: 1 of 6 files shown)


wled00/overlay.cpp
```diff
@@ -89,6 +89,15 @@ void _overlayAnalogCountdown()
 
 void handleOverlayDraw() {
   usermods.handleOverlayDraw();
+  if (analogClockSolidBlack) {
+    Segment* segments = strip.getSegments();
+    for (uint8_t i = 0; i < strip.getActiveSegmentsNum(); i++) {
+      Segment segment = segments[i];
```
blazoncek, line 95: That's unnecessary segment object copy. Use a reference instead.

blazoncek (commented):
It may not be my call but IMO a clock overlay should be moved into usermod.

wled00/overlay.cpp
WoodyLetsCode: Done

WoodyLetsCode:
Hi @blazoncek! I would like to ask what we should do with this PR here?
I implemented this because @Aircoookie added #984 to the 1.0.0 milestone.
And I think the reason you haven't merged it yet is because you would rather have the clock overlay as a usermod.
Who should decide if this becomes a usermod?
Personally, I don't really care if it becomes a usermod or not.
If you are not sure at the moment, maybe we merge it and then we can still move the clock overlay to usermods at some point in the future.
What do you think about that? 😄

wled00/overlay.cpp
```diff
@@ -89,6 +89,15 @@ void _overlayAnalogCountdown()
 
 void handleOverlayDraw() {
   usermods.handleOverlayDraw();
+  if (analogClockSolidBlack) {
+    const Segment* segments = strip.getSegments();
+    for (uint8_t i = 0; i < strip.getActiveSegmentsNum(); i++) {
```
blazoncek, line 94: This may be incorrect as you are iterating across raw `segments[]` vector/array.
The correct way would be:
```c++
for (int i=0; i<strip.getSegmentsNum(); i++) {
  const Segment &segment = segments[i];
  if (!segment.isActive()) continue;
...
}
```

blazoncek (changes requested):
Hi. I toyed with the clock overlay a day or two ago and found it *useless* as it is not precise enough (unless you have a large circle with LED count in multiples of 60). I played with and without effects and the output was unappealing unless the only thing displaying was the clock itself.
AFAIAC I will not use it ever but perhaps there are users who are using it and to my eye this PR makes sense as it is optional.

If you ask me, then I am in favour of merging it but it was not my design so I'd rather see @Aircoookie decides. If he doesn't respond in a week, we'll merge it anyway. Does that sound ok?

wled00/overlay.cpp
WoodyLetsCode: Sure you are right!
Sorry for that! 😞

WoodyLetsCode:
> If you ask me, then I am in favour of merging it but it was not my design so I'd rather see @Aircoookie decides. If he doesn't respond in a week, we'll merge it anyway. Does that sound ok?

That sounds like a good plan.

blazoncek (approved):

PR #3490, opened 2023-10-28
Add LDR sensor dusk/dawn preset control Usermod

JeffWDH:
This usermod will obtain readings from a Light Dependent Resistor (LDR) and will turn on/off specific presets based on those readings. This is useful for exterior lighting situations where you want the lights to only be on when it is dark out.

Tested on an ESP32 (HUZZAH32) with a generic LDR.

### DIFF
--- a/usermods/LDR_Dusk_Dawn_v2/README.md
+++ b/usermods/LDR_Dusk_Dawn_v2/README.md
@@ -1,0 +1,26 @@
+# LDR_Dusk_Dawn_v2
+This usermod will obtain readings from a Light Dependent Resistor (LDR) and will turn on/off specific presets based on those readings. This is useful for exterior lighting situations where you want the lights to only be on when it is dark out.
+
+# Installation
+Add "-D USERMOD_LDR_DUSK_DAWN" to your platformio.ini [common] build_flags and build.
+
+Example:
+```
+[common]
+build_flags =
+  -D USERMOD_LDR_DUSK_DAWN   # Enable LDR Dusk Dawn Usermod
+```
+
+# Usermod Settings
+Setting | Description | Default
+--- | --- | ---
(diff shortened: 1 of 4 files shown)


blazoncek:
It would be great if you could also update `const.h` and `usermod_list.cpp` to allow for simpler compile (compiler `-D` option).

JeffWDH:
> It would be great if you could also update `const.h` and `usermod_list.cpp` to allow for simpler compile (compiler `-D` option).

I have made the requested changes.

platformio.ini
```diff
@@ -134,9 +134,9 @@ build_flags =
   -D DECODE_LG=true
   ;-Dregister= # remove warnings in C++17 due to use of deprecated register keyword by the FastLED library ;; warning: this breaks framework code on ESP32-C3 and ESP32-S2
   -DWLED_USE_MY_CONFIG
-  ; -D USERMOD_SENSORSTOMQTT
-  #For ADS1115 sensor uncomment following
-  ; -D USERMOD_ADS1115
+  ;-D USERMOD_SENSORSTOMQTT   # Enable Send sensor data to MQTT Usermod
```
blazoncek, line 137: Changing `platformio.ini` is not wanted unless strictly necessary.

blazoncek (approved):

blazoncek:
Please do not force push changes next time.

JeffWDH:
Understood. Other projects specifically ask for all changes to be contained within a single commit, I'll try to remember for next time.

blazoncek:
> Understood. Other projects I've contributed to specifically ask for all changes to be contained within a single commit, I'll try to remember for next time.

We do have such guidelines as well but they are not strict rules, especially if changes are requested.
It makes reviewing code easier if commits are clear. But you can always branch, squash merge and make a new PR if that may be an issue.

PR #3501, opened 2023-11-02
A spookier version of Halloween Eyes

TripleWhy:
I wanted to use Halloween Eyes for Halloween this year, but when I tried it, it was less spooky than I remembered. ^^  
So I tried to write a spookier version. Maybe you like it, too.

Bonus: `if (seg.mode != FX_MODE_HALLOWEEN_EYES)` in FX_fcn.cpp isn't required anymore.


blazoncek:
Thank you. Can you provide a video of the modified effect in action?

TripleWhy:
[spookyEyes.webm](https://github.com/Aircoookie/WLED/assets/2760830/1eedb4a9-4b73-472e-b720-490a9c134c45)

The old effect did the following: Display two eyes in a random location, fade them out.

The video isn't great, but mine does this:
- rapidly fade in two eyes in a random location
- randomly blink (or not) for very short durations
- turn off

I also increased the maximum configurable time between eyes showing up and the maximum configurable duration that eyes show up for. For the video, I chose a low time between eyes. In a real setup I'd have a rather large time so that eyes pop up when you don't expect them.

wled00/FX.cpp
```diff
@@ -2600,67 +2600,136 @@ uint16_t mode_twinklecat()
...
+      [[fallthrough]];
+    }
+    case eyeState::on: {
+      uint16_t start2ndEye = data.startPos + HALLOWEEN_EYE_WIDTH + HALLOWEEN_EYE_SPACE;
+      duration = min(duration, static_cast<uint16_t>(128u + (SEGMENT.intensity*64u)));
+
+      constexpr uint32_t minimumOnTimeBegin = 1024u;
+      constexpr uint32_t minimumOnTimeEnd = 1024u;
+      const uint32_t fadeInAnimationState = elapsedTime * static_cast<uint32_t>(256*8) / duration;
+      const uint32_t backgroundColor = SEGCOLOR(1);
```
blazoncek, line 2659: That's not gonna work good with overlay. You will need to improve that.

wled00/FX.cpp
```diff
@@ -2600,67 +2600,136 @@ uint16_t mode_twinklecat()
...
-  if (stateTime == 0) stateTime = 2000;
-
-  if (state == 0) { //spawn eyes
-    SEGENV.aux0 = random16(0, maxWidth - eyeLength - 1); //start pos
-    SEGENV.aux1 = random8(); //color
-    if (strip.isMatrix) SEGMENT.offset = random16(SEGMENT.virtualHeight()-1); // a hack: reuse offset since it is not used in matrices
-    state = 1;
-  }
+  if (!SEGENV.allocateData(sizeof(EyeData))) return mode_static(); //allocation failed
+  EyeData& data = *reinterpret_cast<EyeData*>(SEGENV.data);
```
blazoncek, line 2633: Interesting approach.

blazoncek (commented):
I see you like to type a lot. 😁 
IMO too many `static_cast`s that do not add to code readability, but that is my personal preference.

I would also condense the code wherever possible and add a few more comments.

wled00/FX.cpp
TripleWhy: A few questions about that:

- what are overlays?
- did it work with the original halloween eyes?
- why doesn't it work?
- what is the correct approach?

wled00/FX.cpp
TripleWhy: the allocate/reinterpret is stuff I found in other effects. Is there a better way?

TripleWhy:
> I see you like to type a lot. 😁 IMO too many `static_cast`s that do not add to code readability, but that is my personal preference.

Yeah you are right, but they are also safer :)
What would you like better? C-style casts? using a regular enum instead of enum class? That would allow for more implicit casting.

wled00/FX.cpp
blazoncek: - Overlays: segments overlapping.
- in fact it did not as the principle was the same as in your code, use `getPixelColor()` instead if overlay is enabled

wled00/FX.cpp
blazoncek: No, I was amused by using a reference instead of pointer.

blazoncek:
Being old-ish, I prefer simplicity of C and first iterations of C++. But I like to learn new tricks nevertheless.
Keep the code as is, except if you can tune it (to be smaller and/or faster).

wled00/FX.cpp
TripleWhy: Hm. I don't think I get it. I tried looking at some usages of `getPixelColor`, but it's not really wide spread among other effects.
I tried to experiment with an effect that is simple and uses `getPixelColor`: Dissolve. But it doesn't even seem to work correctly when using it as a simple effect (without overlays), when the background color is anything other than black. Not even dark gray works.

Long story short: I'm willing to put some more work into this, but I'd need more input. Can you maybe point me to some documentation or to an effect I can study that demonstrates how things are done correctly?

blazoncek: Unfortunately there's no such documentation.
Basically `getPixelColor()` retrieves the color pixel is set with. When segments overlap it will retrieve pixel color set in underlying effect or if it was set by some other part of current effect. If the pixel was not yet modified it will get a color from previous frame. It works similarly as getting `leds[i]` when using FastLED.
You need to enable double buffering in LED settings to retrieve unmodified color otherwise the color may be slightly modified. This is due to brightness being applied prior to outputting color to LEDs.

In case of Halloween Eyes, if they are used as overlay over underlying segment you need to fade eyes into underlying color instead of `SEGCOLOR(1)` otherwise the underlying segment will have "incorrect" pixel when eyes fade out.

PR #3507, opened 2023-11-08
DMX: Correct single RGB & DRGB modes to explicitly utilize white

mxklb:
Using `availDMXLen` to choose to utilize white channel is messy because it has nothing to do with ability of the strip. It is the number of channels available in the E1.31/Artnet package. Additional this may lead to unexpected behavior if subsequent DMX start addresses for multiple WLED controllers shall be used.

The approach here is clear, if no white channel exist:

- Single RGB -> 3 DMX channels
- Single DRGB -> 4 DMX channels

And if white channel exists:

- Single RGB(W) -> 4 DMX channels
- Single DRGB(W) -> 5 DMX channels

So number of DMX channels for both modes is +1 if LEDs have white, else number of channels is as documented.

> Note: Should be changed in documentation as well, shall be named _Single RGB(W)_ and _Single DRGB(W)_   

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of DMX data for LED strips with or without a white channel, ensuring correct color rendering in specific DMX modes.
  - Enhanced validation to prevent issues when DMX data length is insufficient for the configured channel setup.


blazoncek:
Please re-base this PR to 0_15 branch.

wled00/e131.cpp
```diff
@@ -142,32 +142,34 @@ void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol){
       return;  // nothing to do
       break;
 
-    case DMX_MODE_SINGLE_RGB:   // 3 channel: [R,G,B]
+    case DMX_MODE_SINGLE_RGB:   // 3-4 channels: [R,G,B] + W (if present)
       if (uni != e131Universe) return;
-      if (availDMXLen < 3) return;
+      uint16_t dmxChannelCount;
+      dmxChannelCount = strip.hasWhiteChannel() ? 4 : 3;
```
blazoncek, line 148: This may not be correct approach as user may choose to allow realtime data only on main segment. And main segment may not contain white channel.
Please check `setRealtimePixel()` for behaviour.

You will need to differentiate according to `useMainSegmentOnly` global variable.

blazoncek (changes requested):
I do not use realtime data so cannot test this PR.
Still, as WLED allows realtime data to be displayed on main segment only, and main segment may not include white channel, it is mandatory to differentiate if the underlying canvas supports white properly.

Segment has light capabilities embedded (and recalculated on each change) and white channel presence can be determined by `Segment::hasWhite()` method.

wled00/e131.cpp
mxklb: Ah alright, I'll further test this the following days according to your comments.

Is `strip.hasWhiteChannel()` always `== true` for RGBW (and ` == false` for RGB) strip only segments?

Due to the utilization of segments to handle different strip types (RGB|RGBW) I think your right and it must be implemented differently. The `strip.hasWhiteChannel()` must be wrong. I haven't had this in mind .. 

Now that I've looked into `setRealtimePixel` method, I'm looking forward to directly implement realtime _Segment (D)RGB(W)_  modes, utilizing `Segment::hasWhite()` and `Segment::setPixelColor(pix, r, g, b, w)` method. 

Due to your sentences

> .. as WLED allows realtime data to be displayed on main segment only, .. 
> .. as user may choose to allow realtime data only on main segment ..

I'm confused. Maybe lack of knowledge of the main segment concept. Is it possible to call for example `Segment &seg = strip.getSegment(1)` and afterwards call `seg.setPixelColor(pix, r, g, b, w)` to realtime set pixels of segment 1? Actually I'm also unclear about this realtime lock statement
```
realtimeLock(realtimeTimeoutMs, mde);
if (realtimeOverride && !(realtimeMode && useMainSegmentOnly)) return;
```
For realtime _Segment (D)RGB(W)_ just leave out the `&& useMainSegmentOnly` is okay?

Maybe I'm wrong, but unfortunately can't find out much about the main segment usage in general.

Nevertheless I'll rebase to 0_15 branch, or maybe provide another PR. If so, I'll close this one.

blazoncek: Briefly as I do not have much time today:
- only one segment can be *main* segment, there is always one *main* segment
- any segment can be *main* segment (differentiated with a lighter background in UI)
- normally you do not need to care which segment is *main*
- each segment can contain/encompass outputs of different LED types
- segment keeps track of output capabilities (RGB, W, CCT, etc)
- `strip`'s capabilities are a union of all segment's capabilities
- realtime code can either use whole strip (disregarding segments) or *main* segment
- when using strip (`useMainSegmentOnly!=true`) `strip.hasWhiteChannel()` can be used to detect presence of white channel
- when using main segment (`useMainSegmentOnly==true`) use `mainseg.hasWhite()` (where `Segment &mainseg = strip.getMainSegment();`)
- realtime locking happens automatically when realtime data is received or if instructed so by external command
- realtime lock is released automatically when no data is received in certain amount of time unless externally locked
- relatime lock insures pixels are not updated by internal effects

mxklb: latest commit shall differentiate according to useMainSegmentOnly global variable.

wled00/e131.cpp
```diff
@@ -144,25 +145,26 @@ void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol){
...
     case DMX_MODE_SINGLE_RGB:   // 3-4 channels: [R,G,B] + W (if present)
       if (uni != e131Universe) return;
-      uint16_t dmxChannelCount;
-      dmxChannelCount = strip.hasWhiteChannel() ? 4 : 3;
-      if (availDMXLen < dmxChannelCount) return;
 
       realtimeLock(realtimeTimeoutMs, mde);
       if (realtimeOverride && !(realtimeMode && useMainSegmentOnly)) return;
 
+      dmxChannelCount = useMainSegmentOnly ? strip.getMainSegment().hasWhite() : strip.hasWhiteChannel() ? 4 : 3;
```
blazoncek, line 152: I would add parentheses here so there is no doubt of precedence.

wled00/e131.cpp
```diff
@@ -144,25 +145,26 @@ void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol){
...
 
     case DMX_MODE_SINGLE_DRGB:  // 4-5 channels: [Dimmer,R,G,B] + W (if present)
       if (uni != e131Universe) return;
-      dmxChannelCount = strip.hasWhiteChannel() ? 5 : 4;
-      if (availDMXLen < dmxChannelCount) return;
 
       realtimeLock(realtimeTimeoutMs, mde);
       if (realtimeOverride && !(realtimeMode && useMainSegmentOnly)) return;
+      
+      dmxChannelCount = useMainSegmentOnly ? strip.getMainSegment().hasWhite() : strip.hasWhiteChannel() ? 5 : 4;
```
blazoncek, line 166: And here as well.

blazoncek (approved):
Seems ok but please add parentheses. 😄

mxklb:
@blazoncek please check https://github.com/Aircoookie/WLED/pull/3507/commits/bdd667d02df49495b5c4654cf505be975a642f10

I think it could be faster when only iterating over differentiated length as well, shall I add these checks again?

blazoncek:
IDK. I have very little knowledge  about realtime apart from DDP implementation.
If you can, do some testing and make appropriate changes if necessary. When you think it is ok just say so and I'll merge.

mxklb:
I'll change back PR draft mode, when I've tested code changes with realtime DMX console e131 input.

softhack007:
> I'll change back PR draft mode, when I've tested code changes with realtime DMX console e131 input.

@mxklb did you make progress? This PR is in draft for 8 months now, are you still working on it?

mxklb:
I forgot to test it. Implementaion looks still good, is an improvement.

I'll try to test it the following week, and will change draft mode, when everything works as expected.

mxklb:
Found some time to test, should be save to merge. 

With this change, sending white channel `W` is possible in DMX realtime modes *Single RGB(W)* and *Single DRGB(W)*.

The number of DMX channels depends on the white color capability of the strip and the selected DMX mode:

| LED-Strip \ DMX Mode | Single RGB(W) | Single DRGB(W) |
|:-----------|:--------:|:-------------:|
| [~~W~~] doesn't have white  | 3   | 4   |
| [W] has white channel   | 4   | 5   |

> Breaking Change: Sending only RGB for RGBW strip will do nothing. In previous implementation `W` channel was set to 0, when sending RGB to RGBW strip. So MR changes makes using these realtime modes more restrictive but precise. Nevertheless sending additional `W` channel was possible, but it was an undocumented hidden feature.

Compared to pervious hidden `W` feature this makes `W` channel usage for those two realtime modes more clear.

PR #3508, opened 2023-11-09
Fix settings caching on some browsers

Moustachauve:
Some browser, especially Android webview, have more aggressive caching strategies and might sometimes cache the response of the settings JS.

This implements two strategy, one in the back-end and one in the front-end. Two might be overkill, so we could keep a single one if needed.

### Front-end strategy to prevent caching
By adding a random string to the path of the js file every time, we make sure to never hit the browser cache.

### Back-end strategy to prevent caching 
By setting the response header "Cache-Control" to "no-store" and setting "Expires" to 0, we make sure the browsers and place calling this never store it in cache.


blazoncek:
I would only approve the second even though we may never have exact control over browser caching on a particular Android/iOS/PC device.
Also, please use `0_14_1` or `0_15` branch for PR.

blazoncek:
@Moustachauve can you revert HTML changes and only do backend change?
That way I think we can merge it right away.
@softhack007 do you agree?

Moustachauve:
Unless @softhack007 has a strong opinion, I will revert the HTML changes tonight.

softhack007:
> @Moustachauve can you revert HTML changes and only do backend change?
> That way I think we can merge it right away.
> @softhack007 do you

Fine for me 👍

Moustachauve:
Revert for the html changes is done.

PR #3513, opened 2023-11-10
UI optimizations

WoodyLetsCode:
# index.css
- removed `scrollbar-width` property as it had an invalid value `6px`. Syntax: https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width#syntax

# index.html

- I removed the trailing slash on void elements as it [has no effect](https://github.com/validator/validator/wiki/Markup-%C2%BB-Void-elements#trailing-slashes-in-void-element-start-tags-do-not-mark-the-start-tags-as-self-closing) and [interacts badly with unquoted attribute values](https://github.com/validator/validator/wiki/Markup-%C2%BB-Void-elements#trailing-slashes-directly-preceded-by-unquoted-attribute-values)
- I changed `p` to `span` in top and bottom bar, because element `p` is not allowed as child of element `button` in this context
- I changed `div` to `span` in `pallist` and `fxlist`, because element `div` is not allowed as child of element `label` in this context

See: https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2

### DIFF
--- a/wled00/data/index.css
+++ b/wled00/data/index.css
@@ -53,7 +53,6 @@ body {
 	-ms-user-select: none;
 	user-select: none;
 	-webkit-tap-highlight-color: transparent;
-	scrollbar-width: 6px;
 	scrollbar-color: var(--c-sb) transparent;
 }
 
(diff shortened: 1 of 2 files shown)


WoodyLetsCode:
Why was this closed?

blazoncek:
Looks like because branch was deleted.

PR #3522, opened 2023-11-13
Fix repeatable action for ir json

martinez20m:
String.indexOf returns -1 if not found. This causes false repeat action.


blazoncek:
Actually that's not correct.

PR #3527, opened 2023-11-15
Usermod v2 http pull light control 0.0.4

roelbroersma:
Update of 0.0.3 but with SHA1 support for the HASH, which is better than the old FOWler hash which had a maximum of 10 digits.  SHA1 is supported on ESP32 and ESP8266 (even though it are different core libraries).

### DIFF
--- a/usermods/usermod_v2_HttpPullLightControl/readme.md
+++ b/usermods/usermod_v2_HttpPullLightControl/readme.md
@@ -17,7 +17,7 @@ The `usermod_v2_HttpPullLightControl` is a custom user module for WLED that enab
 * Specify the URL endpoint and polling interval.
 
 ## JSON Format and examples
-* The module sends a GET request to the configured URL, appending a unique identifier as a query parameter: `https://www.example.com/mycustompage.php?id=xxxxxxxx` where xxxxxxx is a hash of the MAC address combined with a given salt.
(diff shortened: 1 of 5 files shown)


blazoncek:
There are now 2 PRs for similar thing. Which is the correct one?

roelbroersma:
This is the new version v.0.0.4,  the previous one is v.0.0.3.

You can first integrate version 0.0.3 and then this 0.0.4, which has as better hash method.  Maybe in future there will be a v0.0.5 :)

roelbroersma:
> What would be the purpose of merging 0.0.3 first and then 0.0.4? Why not just 0.0.4?

Nothing, beside having some administrative history.
You can also close the request for v0.0.3 and do version v0.0.4.

roelbroersma:
Hi @blazoncek , what do you propose I'll do when in the meantime I have updates?  I first did a pull-request for v.0.0.3,  then (even before that pull-request was merged), I did a new pull-request for v0.0.4  (this one).  And actually...  today I just made a v0.0.5 which fixes some invalid JSON handling and always releasing the buffer (which could have not been done in rare circumstances).

Do you want me to do a pull request for that? 🙄😃

blazoncek:
If these are progressive updates I see no point in opening *different* PRs.
Just push your changes to your current branch and the PR will update automatically.

You can also close 0.0.3 for that matter.

roelbroersma:
Closed because of the new v0.0.5.
See: https://github.com/Aircoookie/WLED/pull/3560

Issue #3515, opened 2023-11-10
Using Scrolling Text with Matrix Panel and #HHMM format in a Segment, Time gets wrong with -8 hour offset once in a while


YvesLev:
### What happened?

This happens once every day sometime once every few days.

This happens on 2 differents Led Panel setup (WS2812b) consisting of 16 X 32 leds.
It seems to happen less often on the Panel in the Kitchen, But this one has a Usermod (BH1750) and a Node-Red automation to auto dim (change the intensity) according to ambiant light.  Also has MQTT enabled.  But the problem exists also.

One segment has scrolling text effect to display the Time with #HHMM format in the Segment description.
Segment 2 display the outdoor temperature using REST command from HA.

The time gets wrong by itself.  -8 Hour (-7h since daylight saving).  It is not exactly 8/7 hour.

Exemple: Panel shows 10:40.  Real time is 17:37.

It does not seems to correct by itself.
To correct the time, I go in the UI, click on info, I see the time is wrong, then, while in the info window, I changed the intensity, it corrects the time immediatly.
Going in any config menu and just pressing save with no modif. will correct the problem.
Also, re-selecting the preset that is already running will fix the time.

I'm using an ESP32-Mini on both clock panel.
My Time Zone US-EST/EDT.

### To Reproduce Bug

I cannot reproduce the bug.
It just happen.
I notice it when I look at the clock.

### Expected Behavior

WLED Internal Time gets off by about 8 hours (7 hours since daylight saving is active).

### Install Method

Binary from WLED.me

### What version of WLED?

0.14.0 Release

### Which microcontroller/board are you seeing the problem on?

ESP32

### Anything else?

I'm using my local NTP server (Which is my Router).
Before I was using another Local NTP server (A windows workstation).
I just put back the default ntp server (0.wled.pool.ntp.org) and will report back.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

blazoncek:
Scrolling Text uses system time, so if a display is incorrect it means system time is incorrect.
System time is determined by querying NTP server. You should diagnose your NTP servers.

If there is no response for NTP server, WLED will keep its internal count of time which may get skewed with long running without NTP as the internal crystal frequency is not guaranteed to be precise.

YvesLev:
> Scrolling Text uses system time, so if a display is incorrect it means system time is incorrect. System time is determined by querying NTP server. You should diagnose your NTP servers.

[...]

PR #3536, opened 2023-11-20
NTP validation, and rejecting malformed responses (related to #3515)

softhack007:
validate NTP responses (fixes https://github.com/Aircoookie/WLED/issues/3515)
* purge old (not yet processes) NTP responses
* validate server responses before updating WLED time
* purge receive buffer when package is rejected (avoids mem leak on esp32)
* bugfix: trigger NTP query directly when NTP was enabled from UI (previously there could be a delay of up to 12 hours)

@Aircoookie, @blazoncek: Please quickly check this change in your environment before merging into 0_14_1.

### DIFF
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -375,7 +375,8 @@
 #define SUBPAGE_JS              254
 #define SUBPAGE_WELCOME         255
 
-#define NTP_PACKET_SIZE 48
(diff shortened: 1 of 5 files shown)


blazoncek (approved):
Apart from some indentation it looks ok to me.

PR #3539, opened 2023-11-23
Added POV image effect

Liliputech:
Setup is really easy, after first boot and WiFi/LEDs setup: go to wled.local/edit and upload a couple image to WLed's filesystem. Only PNG is supported right now, further support for GIF is planned. The image should be as wide as the 1D segment you want to apply to.

When done, go to the Effect page on the UI, select "POV Image" effect.

There should be a new selector, near the Effect Speed slider. You can use that selector to set the image for display on POV.

You could also update the image with a post to the JSON-API like this: curl -X POST http://[wled]/json/state -d '{"seg":{"id":0,"fx":114,"f":"/axel.png"}}'

The segment should move at around 120RPM (that's 2revolutions per seconds) for an image to showup. More informations and pictures here : https://www.instagram.com/lumina.toys/

### DIFF
--- a/usermods/pov_display/usermod_pov_display.h
+++ b/usermods/pov_display/usermod_pov_display.h
@@ -1,0 +1,85 @@
+#pragma once
+#include "wled.h"
+#include <PNGdec.h>
+
+void * openFile(const char *filename, int32_t *size) {
+    f = WLED_FS.open(filename);
+    *size = f.size();
+    return &f;
+}
+
+void closeFile(void *handle) {
+    if (f) f.close();
+}
+
+int32_t readFile(PNGFILE *pFile, uint8_t *pBuf, int32_t iLen)
+{
+    int32_t iBytesRead;
+    iBytesRead = iLen;
+    File *f = static_cast<File *>(pFile->fHandle);
+    // Note: If you read a file all the way to the last byte, seek() stops working
+    if ((pFile->iSize - pFile->iPos) < iLen)
+	iBytesRead = pFile->iSize - pFile->iPos - 1; // <-- ugly work-around
+    if (iBytesRead <= 0)
+	return 0;
+    iBytesRead = (int32_t)f->read(pBuf, iBytesRead);
+    pFile->iPos = f->position();
+    return iBytesRead;
+}
+
+int32_t seekFile(PNGFILE *pFile, int32_t iPosition)
+{
+    int i = micros();
+    File *f = static_cast<File *>(pFile->fHandle);
+    f->seek(iPosition);
+    pFile->iPos = (int32_t)f->position();
+    i = micros() - i;
+    return pFile->iPos;
+}
+
+void draw(PNGDRAW *pDraw) {
+    uint16_t usPixels[SEGLEN];
+    png.getLineAsRGB565(pDraw, usPixels, PNG_RGB565_LITTLE_ENDIAN, 0xffffffff);
+    for(int x=0; x < SEGLEN; x++) {
+	uint16_t color = usPixels[x];
+	byte r = ((color >> 11) & 0x1F);
+	byte g = ((color >> 5) & 0x3F);
+	byte b = (color & 0x1F);
+	SEGMENT.setPixelColor(x, RGBW32(r,g,b,0));
+    }
+    strip.show();
+}
+
+uint16_t mode_pov_image(void) {
+    const char * filepath = SEGMENT.name;
+    int rc = png.open(filepath, openFile, closeFile, readFile, seekFile, draw);
+    if (rc == PNG_SUCCESS) {
+	rc = png.decode(NULL, 0);
+	png.close();
+	return FRAMETIME;
+    }
+    return FRAMETIME;
+}
+
+class PovDisplayUsermod : public Usermod
+{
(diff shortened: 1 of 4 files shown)


blazoncek:
Thank you for the contribution. Looks like a nice enhancement, however...

Before this gets merged, please do the following:
- move all file handling functions into `file.cpp`
- move all segment drawing functions into `FX.cpp`
- remove all unnecessary `*.h` files (the ones that parent *.htm didn't receive updates)
- remove `file` property from `Segment` class and use segment name instead.
- remove *file* extension from effect filters (there is no need for that)
- modify file selector to be displayed beneath segment name input field
- make sure only GIF and PNG files are displayed (do not hide just cfg.json and presets.json)
- make sure this works in AP mode as well (I am not sure /edit works without internet)
- please provide memory footprint for ESP8266 and ESP32 (flash and RAM)

I would personally suggest to make a usermod from this as including additional libraries may be problematic for ESP8266 as it is already on the verge of usability.

Liliputech:
Hello Kristan,
Alright, I'll work on all those separate points asap.
I was also thinking of making this a usermod.
Keep in touch

Liliputech:
I moved mostly everything to a Usermod, but still some work to do.

Liliputech:
Hi Kristan,
I still have a question, given my patch do change index.js file (even in its usermod version) am I not supposed to re-generate the html_ui.h file and add it to the changes?
Best,

Liliputech:
Building in release mode
Compiling .pio/build/esp32c3dev/src/wled00.ino.cpp.o
Retrieving maximum program size .pio/build/esp32c3dev/firmware.elf
Checking size .pio/build/esp32c3dev/firmware.elf
RAM:   [====      ]  44.5% (used 145964 bytes from 327680 bytes)
Flash: [========= ]  90.4% (used 1421534 bytes from 1572864 bytes)

blazoncek:
Sorry for late reply. Changing index.js for purposes of usermod is out of the question as the usermod is optional feature.
ATM I have no solution for this problem (selecting filenames in UI).

It is possible to hack into Info page and (with clever approach) use it to select a file. But it may be a dirty hack.
It is also possible to add a HTTP handler and serve custom HTML page for such purpose.

Will get back if I have any idea how to get those.

Liliputech:
Hi Kristan,
Ok, I'll make some changes to my commit with your advice and provide a better version.
Just thinking out loud, what if I create like "yet another web UI" for WLED dedicated to my purpose?
Just like PixelMagicTool did at the time?
Then I would be able to compile and install it from source as well as everyone else along with WLED, without the need for me to put too much crap in the original WLED repository? what do you think?
Like this one? https://github.com/Aircoookie/WLED/pull/3042
Best,

blazoncek:
Hi. please take a look at #3835 for idea how to use segment name as an image selector.
And update the PR for 0_15 branch.

Liliputech:
Hello Kristan,
I've lost the credentials to my github account ><
I've started the recovery procedure and will make some update to this branch ASAP :)
Best,
Arthur

----- Mail original -----

Manut38:
Just dropping in to say that I absolutely love this!! 😍

Liliputech:
Hi,
I've rebased my commit on top of 0_15 branch :)
I've also tried with https://github.com/Manut38/WLED-GifPlayer-html -> works like a charm except the filename sent is missing a "/" before the filename.
Hope this is going to be merged soon :)

platformio.ini
```diff
@@ -100,6 +100,8 @@ build_flags =
   -D DECODE_SAMSUNG=true
   -D DECODE_LG=true
   -DWLED_USE_MY_CONFIG
+  #For POV Display uncomment following
```
blazoncek, line 103: The changes to `platformio.ini` are not allowed so please revert and add sample to `platformio_override.sample.ini`.

usermods/pov_display/usermod_pov_display.h
```diff
@@ -0,0 +1,84 @@
+#pragma once
+#include "wled.h"
+#include <PNGdec.h>
+static const char _data_FX_MODE_POV_IMAGE[] PROGMEM = "POV Image@!;;;1";
+
+PNG png;
```
blazoncek, line 6: These two should be part of class (instantiation), not globals.

blazoncek (changes requested):
Please remove changes to platformio.ini and consider if you can implement suggestions.

usermods/pov_display/usermod_pov_display.h
Liliputech: This was moved accordingly to the comment

platformio.ini
Liliputech: done :)

blazoncek (approved):

Liliputech:
Thank you very much for your approval! Can't wait for this change to be merged :)

Scabattoir:
Hi Liliputech and all.
I spent the better part of my last week trying to make this POV usermod work but I couldn't.
Is there any chance it could be updated so it would work?
Unfortunately I'm not good at coding, it took me about five days to be able to install VSCode and do a successful compile / upload…
Thank you.

PR #3569, opened 2023-12-05
new buildenv for esp32s3dev_16MB_PSRAM_opi dev board (LilyGo T7-S3)

raymondhardy:
Adds new esp32s3dev_16MB_PSRAM_opi dev board to work with Lily Go T7_s3 ESP32-S3-WROOM-1-N16R8

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -11,8 +11,8 @@
 
 # CI binaries
 ; default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, esp32dev, esp32_eth # ESP32 variant builds are temporarily excluded from CI due to toolchain issues on the GitHub Actions Linux environment
-default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, esp32dev, esp32_eth, esp32dev_audioreactive, lolin_s2_mini, esp32c3dev, esp32s3dev_8MB, esp32s3dev_8MB_PSRAM_opi
-
+default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, esp32dev, esp32_eth, esp32dev_audioreactive, lolin_s2_mini, esp32c3dev, esp32s3dev_8MB, esp32s3dev_8MB_PSRAM_opi, esp32s3dev_16MB_PSRAM_opi
+; default_envs = esp32s3dev_16MB_PSRAM_opi
 # Release binaries
 ; default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, esp32dev, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_8MB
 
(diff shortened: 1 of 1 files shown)


blazoncek:
@raymondhardy thank you for the environment. However I do have some concerns.

@softhack007 & @Aircoookie I am not happy with the growth of `platformio.ini` file and inclusion of every possible dev board with different RAM/flash configuration there is.
I propose we move all but those included in release cycle into `platformio_override.ini.sample`. That way each and every person can have only their environments included without cluttering global `platformio.ini`.

softhack007:
@blazoncek You're right, I think we need a new concept for al these "slightly different" boards. Especially with -S3/-S2/-C3 support added (and -C6 might join next year), collecting all possible buildenvs in `platformio.ini` is very hard to maintain.

In this special case, the `extends` keyword could be used to make the real differences more obvious:
```ini
[env:esp32s3dev_16MB_PSRAM_opi]
extends = env:esp32s3dev_8MB_PSRAM_opi
board_build.partitions = tools/WLED_ESP32_16MB.csv
board_upload.flash_size = 16MB
````


Not sure what to do in the general case though ... its clear we cannot go on collecting buildenv directly in `platformio.ini`. The good old times when esp32 4MB was all that matters are in the past - the futures is more variants (classic, -S3, -S2, -C3, ...), more flash sizes (2MB, 4MB, 8MB, 16MB, ...) and more access modes (dio, qio, qio_qspi, qio_opi, opi_opi, ...) that are not compatible with each other. Maybe `platformio.ini` should only provide the basic configs and consider main differences. Then specific boards would need to be added in another file. 

See also:
* https://github.com/Aircoookie/WLED/issues/2953

File: platformio.ini [hunk 1]
```diff
@@ -524,6 +524,29 @@ board_build.f_flash = 80000000L
 board_build.flash_mode = qio
 monitor_filters = esp32_exception_decoder
 
+[env:esp32s3dev_16MB_PSRAM_opi]
```
softhack007, line 527: see my first posting on how to simplify the section, by leveraging on `extends`.

softhack007 (changes requested):
see details below.

In general, a maintainer decision is needed (as stated by @blazoncek) on how to collect board buildenvs in general. This clarification seems necessary before any approval can be made.

platformio.ini
```diff
@@ -524,6 +524,29 @@ board_build.f_flash = 80000000L
...
+  ; -D WLED_RELEASE_NAME=ESP32-S3_PSRAM
+  -D WLED_USE_PSRAM -DBOARD_HAS_PSRAM ; tells WLED that PSRAM shall be used
+  -D BOARD_HAS_PSRAM
+lib_deps = ${esp32s3.lib_deps}
+board_build.partitions = tools/WLED_ESP32_16MB.CSV
+board_upload.flash_size = 16MB
+board_build.f_flash = 80000000L
+board_build.flash_mode = qio
+monitor_filters = esp32_exception_decoder
+
```
softhack007, line 549: Please fix build errors and warnings:

1) `<command-line>: warning: ISO C++11 requires whitespace after the macro name`

2) `[.pio/build/esp32s3dev_16MB_PSRAM_opi/partitions.bin] Source 'tools/WLED_ESP32_16MB.CSV' not found`

platformio.ini
```diff
@@ -524,6 +524,29 @@ board_build.f_flash = 80000000L
...
+framework = arduino
+
+# Configure options for the N16R8V variant
+build_flags= ${common.build_flags} ${esp32s3.build_flags}
+  -D CONFIG_LITTLEFS_FOR_IDF_3_2 -D WLED_WATCHDOG_TIMEOUT=0
+  -D ARDUINO_USB_CDC_ON_BOOT=0  -D ARDUINO_USB_MODE=1 ;; for boards with serial-to-USB chip
+  ;-D ARDUINO_USB_CDC_ON_BOOT=1 -D ARDUINO_USB_MODE=1      ;; for boards with USB-OTG connector only (USBCDC or "TinyUSB")
+  ; -D WLED_RELEASE_NAME=ESP32-S3_PSRAM
+  -D WLED_USE_PSRAM -DBOARD_HAS_PSRAM ; tells WLED that PSRAM shall be used
+  -D BOARD_HAS_PSRAM
```
softhack007, line 542: this line is a duplicate of the previous `-D WLED_USE_PSRAM -DBOARD_HAS_PSRAM`

platformio.ini
```diff
@@ -11,7 +11,7 @@
 
 # CI binaries
 ; default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, esp32dev, esp32_eth # ESP32 variant builds are temporarily excluded from CI due to toolchain issues on the GitHub Actions Linux environment
-default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, esp32dev, esp32_eth, esp32dev_audioreactive, lolin_s2_mini, esp32c3dev, esp32s3dev_8MB, esp32s3dev_8MB_PSRAM_opi
+default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, esp32dev, esp32_eth, esp32dev_audioreactive, lolin_s2_mini, esp32c3dev, esp32s3dev_8MB, esp32s3dev_8MB_PSRAM_opi, esp32s3dev_16MB_PSRAM_opi
```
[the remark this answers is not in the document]
raymondhardy: Removed

platformio.ini
raymondhardy: Changed to use extends

platformio.ini
softhack007: Yes, most likely. The automated github build runs on a linux host, and linux is case-sensitive for filenames.

raymondhardy:
> ecision is needed (as stated by @blazoncek) on how to collect board buildenvs in general. This clarification seems necessary before any approval can be made.

Sounds good. Just wanted to state that i did a new build with these changes and it works for the board! (I tested it with another board i had without any code on it)

softhack007 (approved):

PR #3572, opened 2023-12-05
fix(esp32c3-2mb): correct flash size for c3 board with only 2MB

drasch:
Tested using [NWI1125](http://w2.electrodragon.com/Board-dat/NWI/NWI1125-DAT/NWI1125-DAT.md) from https://www.electrodragon.com/product/esp-led-strip-board/


blazoncek:
@softhack007 I've talked to @Aircoookie and he agrees to prune `platformio.ini` and move all non-default environments into sample file.
Will do that in 0_15 branch.

PR #3573, opened 2023-12-06
add env for esp32 pico d4 with valid ledpin to prevent bootloops

driemekasten:
In recent versions of WLED GPIO 16 is used as the default LEDPIN. This causes bootloops on ESP32 Pico D4 because GPIO 16 is connected to the embedded flash (see notes in ESP32 Pico D4 Datasheet Section 2.2: "Pins IO16, IO17, CMD, CLK, SD0 and SD1 are used to connect the embedded flash, and can not be used for other purposes. For details, please see Section 6 Schematics.").
Added a new environment for Pico D4 (esp32picod4) to platformio.ini with LEDPIN 4 defined as the default (gpio 4 matches my own hardware project, GPIO 2 will also cause no problems). Also changed upload Speed to max. for this env.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -403,6 +403,17 @@ lib_deps = ${esp32.lib_deps}
 monitor_filters = esp32_exception_decoder
 board_build.partitions = ${esp32.default_partitions}
 
+[env:esp32picod4]
+board = esp32dev
+upload_speed = 921600
+platform = ${esp32.platform}
+platform_packages = ${esp32.platform_packages}
+build_unflags = ${common.build_unflags}
+build_flags = ${common.build_flags_esp32} -D WLED_RELEASE_NAME=ESP32 -D LEDPIN=4 #-D WLED_DISABLE_BROWNOUT_DET # -D WLED_MAX_BUTTONS=6 
+lib_deps = ${esp32.lib_deps}
+monitor_filters = esp32_exception_decoder
+board_build.partitions = ${esp32.default_partitions}
+
 [env:esp32dev_audioreactive]
 board = esp32dev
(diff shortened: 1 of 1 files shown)


blazoncek:
See considerations in #3569

softhack007:
Hi,
Similar to @blazoncek, I don't think that adding another buildenv - that only differs in the default ledpin - is the right way to address this problem.

From looking at the datasheet, it seems that gpio 16+17 are not availeable for use, similar to esp32 wrover with PSRAM. 

My proposal would be
* use a custom buildenv (platformio_override.ini), with `board = pico32`
* in WLED, we can then check for pico board with `#ifdef ARDUINO_ESP32_PICO` - and change d3fault pins, as well as block  gpio16/17.


It means that only the source code parts shown below would need adjustment. Everything else will come due to `board = pico32` in your platformio_override.ini.


https://github.com/Aircoookie/WLED/blob/1dab26bcbcac051f2b7be47a2d5c757a9938bf1f/wled00/const.h#L461-L465

https://github.com/Aircoookie/WLED/blob/1dab26bcbcac051f2b7be47a2d5c757a9938bf1f/wled00/wled.cpp#L351
https://github.com/Aircoookie/WLED/blob/1dab26bcbcac051f2b7be47a2d5c757a9938bf1f/wled00/wled.cpp#L365-L368

blazoncek:
This should be no longer necessary as I removed (offending) GPIO16 from default values for ESP32 (it also clashed with C3).
the new default value is GPIO2 or GPIO1 for DMX, If GPIO1 is also occupied by debug build, GPIO3 is used instead..

PR #3574, opened 2023-12-07
WifiStatus usermod - indicate Wifi state with LED blinks

petrusvr:
New User mode v2 added. Wifi status user mode indicates Wifi state with different blink types on LED connected to configured PIN.

### DIFF
--- a/usermods/WifiStatus/readme.md
+++ b/usermods/WifiStatus/readme.md
@@ -1,0 +1,35 @@
+# Wifi Status on LED
+v2 Usermod to indicate Wifi connection state on LED connected to a GPIO.
+
+Three states are supported:
+
+1. Wifi **disconnected** - indicated by fast blinks of LED
+2. Wifi **connected** - LED on with very short off after 2 seconds
+3. Wifi in **Access Point mode** - LED blink twice time and pause for 1 second 
+
+## Installation
+
+Copy and update the example `platformio_override.ini.sample` 
+from the Rotary Encoder UI usermode folder to the root directory of your particular build.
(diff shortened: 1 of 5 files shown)


blazoncek:
Thank you.
You could also modify or enhance STATUSLED functionality within WLED core.

petrusvr:
> Thank you. You could also modify or enhance STATUSLED functionality within WLED core.

Oh, I haven't notice that functionality. But using WifiStatus usermod gives possibility to configure pins/enabled from UI. So, maybe it is not worthless.

blazoncek:
I am not very fond of duplicate code 🤷‍♂️  but otherwise have no objections.

blazoncek:
Please update PR and rebase to 0_14_2 branch.

blazoncek:
Similar to #3631

dosipod:
@petrusvr 
I tried the usermod and it works as expected , one thing i am curious about since you are making a usermod
is to have the ability to use something like ws2812b pixel for status leds and might be add colors 

In the current wled.cpp code the logic does not seem to work as expected and the onboard led will blink all the time even when wled is connected to wifi which does not make sense , only worked for me when i changed the logic  and also could not get it to work with normal pixel just the onboard led  , did you notice the same behavior when just using the  -D STATUSLED=2
Seems too important feature in this hard times of multiple reports of connectivity issues and i also think it should be core functionality 
image

softhack007:
closing - this PR had no activity since >20 month, its not compatible any more with WLED 16.0.0-beta, and the purpose / added value is not clear to me.

PR #3590, opened 2023-12-14
Send empty page if skin.css does not exist

WoodyLetsCode:
This has the advantage that we do not need to send the 404 error page every time.

I also had a problem with this in Firefox. Sometimes Firefox used the css of the 404 error page and then the ui looked like this:
grafik

This should no longer happen with this PR.


blazoncek:
You are a treasure trove. ❤️

PR #3596, opened 2023-12-19
Add own background image upload

WoodyLetsCode:
This adds the feature to upload your own background image.
If you select an image, it will be base64 encoded and then set as the background url.
grafik

**Limitations:**
The maximum size of the image depends on the maximum size of your browser's local storage. Usually browsers can store up to 5 MiB of local storage. If you try to use an image that is too large, you will get this error:
grafik

### DIFF
--- a/wled00/data/settings_ui.htm
+++ b/wled00/data/settings_ui.htm
@@ -118,10 +118,10 @@
 	function genForm(s) {
 		var str = "";
 		str = addRec(s,"",l);
+		oldUrl = "";
 		
 		gId('gen').innerHTML = str;
 		if (gId('theme_bg_rnd').checked) {
-			oldUrl = "";
 			toggle("Image");
 		} else if (gId('theme_bg_url').value.startsWith('data:')) {
 			gId("bg_url").classList.add("hide");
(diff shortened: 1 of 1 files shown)


blazoncek:
In last 30 days we added almost 2kB into UI alone.
10.12.2023
`const uint16_t PAGE_index_L = 32764;`
10.1.2023
`const uint16_t PAGE_index_L = 44491;`

WoodyLetsCode:
We moved simple UI to index page, this increased the size a lot.
On the other hand we have saved a lot by eliminating duplicate code in simple UI.
But I'll leave it up to you whether we merge this PR or not. 😁

blazoncek:
True, and iro and color picker too.

Aircoookie (approved):
I really like this as it allows for custom backgrounds without utilizing very limited ESP resources or having to host and download the image from some external server!
Merging.

WoodyLetsCode:
Thanks! Another point that came to my mind: Also, when using WLED in AP mode, you can't download the image from an external server because you don't have Internet access. So now you can also set your own background image in AP mode.

PR #3597, opened 2023-12-19
Inline iro.js and rangetouch.js

WoodyLetsCode:
Since #3511 has been merged, we can now include `iro.js` and `rangetouch.js` using the inliner again.

I also removed the unused feedback function as it was probably forgotten to remove in commit d6c0642a02ab2c43391ebb7db3ef353f21cbe5fb.

### DIFF
--- a/wled00/data/index.htm
+++ b/wled00/data/index.htm
@@ -7,49 +7,7 @@
 	<meta content="yes" name="apple-mobile-web-app-capable">
 	<link rel="shortcut icon" href="data:image/x-icon;base64,AAABAAEAEBAAAAEAGACGAAAAFgAAAIlQTkcNChoKAAAADUlIRFIAAAAQAAAAEAgGAAAAH/P/YQAAAE1JREFUOI1j/P//PwOxgNGeAUMxE9G6cQCKDWAhpADZ2f8PMjBS3QW08QK20KaZC2gfC9hCnqouoNgARgY7zMxAyNlUdQHlXiAlO2MDAD63EVqNHAe0AAAAAElFTkSuQmCC"/>
 	<title>WLED</title>
-	<script>
-	function feedback(){}
-	// instead of including [script src="iro.js"][/script] and [script src="rangetouch.js"][/script]
-	// (which would be inlined by nodeJS inliner during minimization and compression) we need to load them dynamically
-	// the following is needed to load iro.js and rangetouch.js as consecutive requests to allow ESP8266
-	// to keep up with requests (if requests happent too fast some may not get processed)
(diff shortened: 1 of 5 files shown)


wled00/data/index.htm
```diff
@@ -7,52 +7,9 @@
 	<meta content="yes" name="apple-mobile-web-app-capable">
 	<link rel="shortcut icon" href="data:image/x-icon;base64,AAABAAEAEBAAAAEAGACGAAAAFgAAAIlQTkcNChoKAAAADUlIRFIAAAAQAAAAEAgGAAAAH/P/YQAAAE1JREFUOI1j/P//PwOxgNGeAUMxE9G6cQCKDWAhpADZ2f8PMjBS3QW08QK20KaZC2gfC9hCnqouoNgARgY7zMxAyNlUdQHlXiAlO2MDAD63EVqNHAe0AAAAAElFTkSuQmCC"/>
 	<title>WLED</title>
-	<script>
```
blazoncek, line 10: I would keep this for future reference as a comment.

blazoncek (changes requested):
Generally approved from my side.
Removal of iro.js and rangetouch.js was needed for simplified UI.

You may want to edit cdata.json to comment out generation of .htm files and remove/comment references from wled_server.cpp for PR to be complete.

wled00/data/index.htm
WoodyLetsCode: I see no advantage in keeping that as a comment. If you want to look at the code again in the future, you can simply look at the Git history.
I think that would just make the code less readable and most people would probably wonder what this comment is for when they see it.

And what exactly would you want to leave as a comment? The complete `<script>` block or just the feedback function?

blazoncek: History is one thing, actual snippet embedded as a comment is another. Snippet can be used immediately by anyone while code from history is only viable option for those that know it was there. Snippet (as a comment) *does not reduce readability* of the code, even without any actual comment in it, on the contrary IMO it adds value for a possible future when such feature may be re-implemented.

Dynamically loading of JS files may come in handy in the future (for possible extensions) and at least POC code should be kept in place as a guideline for such case. Commented code does not increase binary code size or adds performance penalties.

I am in favour of reducing HTTP requests to ESP but not at the expense of removing *knowledge* from source files. Otherwise you can strip *every* comment from *all* source files.

WoodyLetsCode: I added a comment with referencte to how it was done in 0.14.0. What do you think of this?
I think this is better because we don't have this big block of commented out code (so the code looks cleaner) and we still preserve the knowledge.

blazoncek (approved):

Issue #3603, opened 2023-12-21
Cannot create a custom palette with only two colors.


imeszaros:
### What happened?

The palette files is uploaded, but WLED won't read it because of a wrong length check in code.

### To Reproduce Bug

Try to create a palette with only 2 colors.

### Expected Behavior

Palette being created and displayed.

### Install Method

Binary from WLED.me

### What version of WLED?

0.14.0

### Which microcontroller/board are you seeing the problem on?

ESP32

### Anything else?

I'll create a PR soon.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

imeszaros:
Nevermind I've just noticed it's already fixed.

PR #3604, opened 2023-12-21
Fix for #3603

imeszaros:


imeszaros:
It looks like it has been fixed already after 0.14.0.

PR #3606, opened 2023-12-21
Fix bug that welcome page was not shown

WoodyLetsCode:
I noticed that the welcome page was not shown anymore.
This is a fix for that.
Now the welcome page shows up again.


Aircoookie (approved):
Good catch! This is due to `initConnection()` no longer returning at AP init if WiFi is not configured as ESP-NOW init is also part of the function now. Your fix should do the trick just fine.

PR #3610, opened 2023-12-22
Allow PSRAM (SPI RAM) to be used for JSON buffer

blazoncek:
JSON buffer is by far the largest consumer of heap (RAM).
While it *is* used often its speed is not that important for JSON operations (SPI RAM is slower than DRAM).

This will allow ESP devices equipped with SPI RAM to use JSON buffer twice the size of non-SPI RAM devices. This in turn allows ledmaps twice the size (up to 2400 pixels) or more *wiggle* room for web or API requests.


blazoncek:
Seems to be working flawlessly.

Issue #3387, opened 2023-09-26
Changing presets corrupts presets.json


bscholer:
### What happened?

Whenever I change my presets on one specific device, and refresh the page, I get an error saying that the presets.json file is corrupt and cannot be loaded.

When I inspect the file on the edit page, I can find the syntax issue quickly. It seems like when I save a preset, it duplicates it without a key in the presets array. As you can't have keyless objects in JSON, this results in a corrupted file.

### To Reproduce Bug

1. Download [presets.zip](https://github.com/Aircoookie/WLED/files/12730450/presets.zip), unzip, and upload `presets.json` file to any WLED device, running 0.13.X or 0.14.X (maybe older too).
2. Navigate to home page, and see that everything is working as expected.
3. Make a change to an existing preset (overwrite with state)
4. Refresh the page after saving, and see that the presets.json file is now corrupted.

### Expected Behavior

I should be able to save changes to existing presets without corrupting the file.
Better yet, WLED should validate that the new JSON is valid before writing over the old one.

### Install Method

Binary from WLED.me

### What version of WLED?

0.13.X, 0.14.X

### Which microcontroller/board are you seeing the problem on?

ESP32

### Anything else?

This only seems to happen with this presets.json file, and I'm not really sure why. The file I linked in the "To reproduce bug" section is valid, and nothing seems to be wrong with it. Everything falls apart when you edit a preset though.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

blazoncek:
A known issue. Root cause still unknown.
0.14 should be less affected though.

bscholer:
Are there other GitHub issues about this? I'm guessing some amount of debugging has been done on it?

blazoncek:
Yes, plenty. I no longer have the energy for it.

bscholer:
#2793, closing

blazoncek:
Anyone stumbling on this: Presets on ESP32 get corrupted (99.999%) on save by incompatible flash mode used when flashing ESP for the first time.
To resolve the issue *you will need to flash new bootloader using DIO or DOUT mode*.

Issue #3347, opened 2023-08-26
issue loading your presets


JevVan:
### What happened?

I was changing up my presets and at some point it failed to load and gave me the issue loading presets error.

After some googling and a bunch of troubleshooting with backup and restore I think I got it to a repeatable situation.  It probably has something to do with my switching the leds from a couple strands on pins 2 & 4 to a matrix connected only to pin 2.  

### To Reproduce Bug

Using the following .json backup of the presets, if you delete the one in slot 55, then press Config, then press Back,  You get the error, "Failed to load presets"

Looking a the presets file, it is add a bunch of spacing in the .json when deleting the preset in spot 55.  Running it through a cleaner after removing number 55 the same file will work.  So I believe it is just a bunch of extra space.

I was thinking it was only happening when deleting the last preset, but I cannot make it occur again with new presets even if I put them in slot 55.  So, in the end, I'm not sure how I got here unless it's the change in configuration.

Here is the preset that will break when 55 is deleted:
[wled_presets_WorkingWith55.json.txt](https://github.com/Aircoookie/WLED/files/12446678/wled_presets_WorkingWith55.json.txt)

And here is a backup of the broken one after deleting item in slot 55
[wled_presets_BrokenWithout55.json.txt](https://github.com/Aircoookie/WLED/files/12446676/wled_presets_BrokenWithout55.json.txt)

### Expected Behavior

It just works!

### Install Method

Binary from WLED.me

### What version of WLED?

WLED version 0.14.0-b4

### Which microcontroller/board are you seeing the problem on?

ESP32-S2

### Anything else?

Not 100% sure on the controller.  The stick on back says NODEMCU     ESP32S      ver 1.1

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

blazoncek:
Duplicate of #3181 #2886 #3139 #2793 #2755 #1981 #2434 and possibly others.

A known bug with a few workarounds and remedies in place but it can still happen as root cause is still undetermined.

Phobos-7:
Unfortunately, this issue wasn't fixed in the latest 0.14.0.
I am not using an Android app, only a web interface.
I got "Sorry, there was an issue loading your presets!" error right after I backed up my presets to a json file...

JevVan:
JSON file gets corrupted I was able to deal with it by opening manually
finding the messed up parts and fixing it.

blazoncek:
[...]

Issue #3181, opened 2023-04-22
File system corruption when adding JSON Preset


werkstrom:
### What happened?

Presets get corrupted. This is the second time in a few days it has happened so, better report it: I was adding JSON presets both times, i.e. JSON commands. The last time I was adding this exact JSON: `{"pixart":{"anim": ""}}`. I then get this screen on the presets page: 
![image](https://user-images.githubusercontent.com/13434693/233786227-e4846294-bb91-45cc-996a-a8f440b0769d.png)
Reboot does not fix the issue (predictably) 
When looking in the JSON it is clearly corrupt with strange chars some places. Fixing the JSON, (manually) fixes the issue (predictably).

### To Reproduce Bug

Adding JSON presets.

### Expected Behavior

For the file system not to corrupt the file 

### Install Method

Self-Compiled

### What version of WLED?

2303240

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell
Get this on the serial:

.pio/libdeps/esp32dev/LittleFS_esp32/src/lfs.c:1076:error: Corrupted dir pair at {0x0, 0x1}
.pio/libdeps/esp32dev/LittleFS_esp32/src/lfs.c:1076:error: Corrupted dir pair at {0x0, 0x1}
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

blazoncek:
Known issue. The corruption has been mitigated as much as WLED can do it but can still happen in special circumstances.
Duplicate of #2793

werkstrom:
Allright. Not saying anything should be done then, but leaving this here *in case* anyone would like to address it in the future: Not sure how things are done in code today either so I could be out sailing, but: 

If we write to one single file we basically are out of luck should the write go wrong. We could use intermediate shadow files for editing and then swapping them in the end (at save) validating their integrity before committing. Something along then lines. On save:

1. Make a copy of (f.ex.)  preset.-json --> checkpreset.json
2. Add changes like today to preset.json (guessing arduinojson) 
3. Validate integrity of preset.json
- if OK delete checkpreset.json and continue on as usual. Job done should be 95% of cases.
- if NOT OK delete preset.json, copy checkpreset.json --> preset.json, start over at 2

Now, I realize I might be way out here so it's just a thought.

blazoncek:
The approach is good on paper but unfortunately it takes a lot of CPU cycles and requires enough free file system space.
[...]

PR #3611, opened 2023-12-22
Suspend strip during operations

blazoncek:
My testing showed that preset corruption (#3382, #3387, #3347, #3181, #2793 and others) happens mostly when strip is being updated (ISR is being used).

This PR adds semaphoring to `strip.service()` so that effect rendering is postponed as well as adds additional checks to preset saving to delay it while `strip` is being updated.


blazoncek:
Works correctly, though I cannot confirm fix for corrupt presets as I cannot encounter one.

Issue #3601, opened 2023-12-20
Sunset shows 00:00 with certain longitude values.


zowper:
### What happened?

A few days before December 7th (and including December 7th), if the longitude was set to 11.74, the sunset time displayed "00:00" and the sunset scheduling didn't work properly.  If I changed the longitude to 111 flat, it worked properly.  

A few days ago (around December 18th) 111 stopped working as well, but 110 now works.  It appears to be a calculation error that changes over time.

### To Reproduce Bug

As of last night, the bug can be recreated by setting the latitude to 40.40 and the longitude to 111.  The exact numbers that reproduce the error will likely change over time.  After setting that latitude and longitude and clicking "Save", you should see the sunset time display as "00:00".

### Expected Behavior

It should calculate an actual sunset time for any valid latitude and longitude (which should be around 17:00 in this case for locations in Utah, USA).

### Install Method

Binary from WLED.me

### What version of WLED?

0.14.0

### Which microcontroller/board are you seeing the problem on?

ESP32

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

blazoncek:
Duplicate of #3400 and possibly others.
This may be math rounding error. /cc @softhack007

softhack007:
@blazoncek I found this in wikipedia, is it the same calculation as in WLED? The wikipedia formula also seems to work above the artic circle.

https://en.m.wikipedia.org/wiki/Sunrise_equation

blazoncek:
I have no idea where I borrowed that algorithm from (perhaps some NOAA resource). Not from Wikipedia though.

softhack007:
Could become complicated ... 
I think the main numerical instability comes from using "tan". tan(angle) shoots towards infinity for angle values close to 90deg and 270deg. So a small inaccuracy is enough to "tip the boat over" and the math starts to produce wrong numbers. I don't think we can cure it by using `double` instead of `float`.

https://github.com/Aircoookie/WLED/blob/10faaaf531e5f6eb2c00e31cba0f2ecd18ef1fd3/wled00/ntp.cpp#L468-L469

The wikipedia formula avoids `tan()` and only uses `sin()` and `cos()`. So it might be a good alternative.

However some deeper thinking and lots of testing might be needed if we want to improve the calculations in WLED. Looks like something for 0_15, but I don't see a quick fix in 0_14_1.

softhack007:
A quick fix could be to avoid angles where tan() becomes unstable. Something like 
[...]

PR #3612, opened 2023-12-22
workaround for invalid sunrise/sunset times (#3601)

softhack007:
In case of invalid or impossible sunset/sunrise results, retry by using previous day. 
max 3 days back, to prevent infinite loops and too-far-away results.

With today's date (22.12.2023), CET timezone (UTC+1), the following is a good testcase:
latitude = 65.74 S
longitude = 110.8 W

Debug output:
> NTP recv, l=48
> * sunrise (minutes from UTC) = 2147483647
> * sunrise (minutes from UTC) = 2147483647
> * sunrise (minutes from UTC) = 2147483647
> * sunrise (minutes from UTC) = 448
> Sunrise: 08:28
> * sunset  (minutes from UTC) = 0
> * sunset  (minutes from UTC) = 0
> * sunset  (minutes from UTC) = 0
> * sunset  (minutes from UTC) = 435
> Sunset: 08:15

its antarctic summer, so sun down for only 12 minutes makes sense near the antarctic circle.

Edit: @blazoncek I think there is another "brain bug" in the code. The sunset time is returned in "UTC". so when the local time  is UTC-8 and sunrise happens to be at 08:00 localtime, then `getSunriseUTC()` also returns `0`, and in this case that's a valid results. So using `0` as error return value of `getSunriseUTC()`  is problematic. Maybe we should better use an impossible value (like `INT16_MIN`) for error.

In fact, my workaround works for this scenario, too. But we should revise the code in 0_15.

### DIFF
--- a/wled00/ntp.cpp
+++ b/wled00/ntp.cpp
@@ -448,7 +448,7 @@ void checkTimers()
 
 #define ZENITH -0.83
 // get sunrise (or sunset) time (in minutes) for a given day at a given geo location
-int getSunriseUTC(int year, int month, int day, float lat, float lon, bool sunset=false) {
+static int getSunriseUTC(int year, int month, int day, float lat, float lon, bool sunset=false) {
   //1. first calculate the day of the year
   float N1 = 275 * month / 9;
   float N2 = (month + 9) / 12;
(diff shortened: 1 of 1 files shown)


blazoncek (approved):
I wouldn't say "math instability" but rather "rounding errors". 😀 

The code looks good I am just wondering why you decided to test `minUTC` to `2*24*60` and not just `24*60` as anything above 1440 is incorrect result.

softhack007:
Thanks, i've updated the PR. I also fixed the "0 is a valid UTC sunrise time" problem as it was possible to do that locally.

PR #3617, opened 2023-12-26
Make palette editor more mobile friendly

imeszaros:
- adjusts the page layout to eliminate scrolling on mobile screens
- makes it possible to drag the color markers on the palette editor on mobile browsers


blazoncek:
Thanks. When time permits I'll test and comment.

ljmerza:
are you able to post a small gif of how the interaction has imporved?

imeszaros:
> are you able to post a small gif of how the interaction has imporved?

Sure, here's the original behavior:
original

And here's the fixed behavior:
fixed

PR #3618, opened 2023-12-26
Fix pixel magic page logo

imeszaros:
This pull request fixes the missing pixel magic tool logo, the inliner could not process it because of a leading whitespace character in the `src` attribute of the image.


blazoncek:
FYI PXM has been changed for 0_15. This may be obsolete.

PR #3620, opened 2023-12-27
Use canvas instead of CSS gradient for liveview

zanhecht:
Allows for "pixel perfect" liveview instead of diffused view, to better match what's shown at https://kno.wled.ge/features/effects/. This will make it easier to see what the LEDs are doing, although it may not be as accurate a representation for installations with diffused LEDs.


zanhecht:
As a side note, I could've added the black border between pixels as shown on the effects page, but it doesn't look great on smaller screens. The basic code to do so, added after the ```fillRect``` line, is
```
            context.strokeStyle = "black";
            context.lineWidth = 1;
            context.strokeRect ( i * w / len, 0, w, h);
```

If the borders are desired, it would be possible to test for a minimum value of "w/len" before drawing them, but it would lead to an inconsistent apparent brightness between devices so it's not a great solution.

blazoncek:
Thank you. Please rebase this PR to 0_15 branch as such change will not make it into current version.

zanhecht:
I think I messed up on the rebase. I'll close this PR and create a new one.

Issue #3399, opened 2023-09-29
Gamma Correction for Custom Palettes


BaconWithThat:
**Is your feature request related to a problem? Please describe.**
I am trying to implement custom palettes and the output doesn't match my expectation.  After some experimentation, the output matches what I see with gamma correction turned off.  After some chat on discord, it was confirmed that custom palettes are expected to be gamma corrected.

**Describe the solution you'd like**
I think the most straightforward solution would be to do gamma correction for custom palettes.

**Describe alternatives you've considered**
If I knew how to do the gamma correction manually, I could do that and save the corrected values as a custom palette.

**Additional context**
Thanks to Blazoncek in the discord for helping me.

blazoncek:
Just a clarification for further consideration:
Existing FastLED palettes are already *gamma corrected* (at least kind of) but custom palettes are not.
Custom palette editor (available since 0.14.0-b2) does not know of any gamma correction and uploads palette as entered in UI. Also, it is possible to enter palette using /edit maintenance page.
So, custom palettes will remain gamma unadjusted in the file. We can adjust gamma when loading such palette from file into RAM.
Unfortunately this will have performance impact as loading can happen thousands of times per second.

BaconWithThat:
Could we get instructions on how to do gamma correction as an article on the wiki?  Then users w/ custom palettes could put the gamma corrected colors into their custom palettes.  I'm unsure what the downside would be aside from it looking wonky on the computer screen.

blazoncek:
https://en.m.wikipedia.org/wiki/Gamma_correction

BaconWithThat:
I ended up using the values in the built in [Gamma correction table ](https://github.com/Aircoookie/WLED/blob/332be7edd69853f70aebf92b83a4f0f275bcfa96/wled00/colors.cpp#L305) to manually transpose my custom palette and it looks great.

blazoncek:
I've implemented gamma correction for custom palettes.
Will be available in 0.14.1

PR #3624, opened 2023-12-28
Show custom palettes with non-gamma-corrected colors on the UI

imeszaros:
This way it would work the same way as the automatic palettes created from the user-defined colors. On the UI the input colors are shown but the actual palettes are created from gamma corrected colors. The UI shows the palettes the same as created on the editor so users won't be confused.

Related to #3399

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -887,6 +887,7 @@ class WS2812FX {  // 96 bytes
 
   // end 2D support
 
+    bool loadCustomPalette(CRGBPalette16 &targetPalette, int pal, bool gammaCorrection); // load a custom palette, returns true if the file exists
     void loadCustomPalettes(void); // loads custom palettes from JSON
     CRGBPalette16 _currentPalette; // palette used for current effect (includes transition)
     std::vector<CRGBPalette16> customPalettes; // TODO: move custom palettes out of WS2812FX class
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1727,51 +1727,61 @@ void WS2812FX::printSize() {
 #endif
 
 void WS2812FX::loadCustomPalettes() {
-  byte tcp[72]; //support gradient palettes with up to 18 entries
   CRGBPalette16 targetPalette;
   customPalettes.clear(); // start fresh
   for (int index = 0; index<10; index++) {
-    char fileName[32];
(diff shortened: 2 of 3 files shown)


blazoncek:
This defies #3399 as the request was to display gamma corrected palette.

I am not in favour of this change as it deals with gamma in partial way. Gamma corrections should be tackled in a global way taking account all considerations.
Also, if you do still want to pursue gamma correction in a more consistent way, please rebase PR to 0_15 branch. It will not go into 0.14.x

imeszaros:
Thanks for having a look!

Maybe I don't get what's the point of gamma correction in WLED. My understanding is, that the perceived brightness of LEDs is not linear to their PWM level, i.e. R=127,G=127,B=127 is not actually "half bright" to our eyes. Therefore we need gamma correction, we use the original channel value and get a substitute value from a lookup table calculated using a correction value. For a given color `C`, we need to set `Cγ` to the LED for us to percive the color `C`.

Therefore, if my understanding is right, there is no point in showing gamma corrected colors on the UI, otherwise the actual "color" of the LEDs would be perceived as brighter as shown on the UI. Basic palettes like 'Color Gradient' and 'Colors 1&2' currently don't show gamma corrected colors. Additionally, I don't understand how is it useful that FastLED palettes are pre-corrected if they are shown on the UI as they are, instead of the pre-correction versions of them.

Could you please point me to the right direction if I'm wrong, and feel free to close this PR if it has no value.

blazoncek:
Regardless of the point of gamma correction, #3399 requested gamma corrected palettes and consistency with color slot handling and it was implemented that way.

As I pointed above, gamma handling isn't implemented correctly (or in consistent manner) everywhere and there were discussions how to tackle it though no attempt was made to properly implement overall gamma handling (UI, palettes, liveview, realtime data, colors, etc). If you do want to tackle this problem, search issues opened on Github, WLED forum and posts on Discord (there were plenty) then try to implement gamma handling that will be *transparent and consistent* for every use.

I do not want yet another partial approach as it only adds to the confusion.

imeszaros:
My pull request does not remove gamma correction for custom plalettes, just allows showing the custom palettes as the user who created them intended to create them. Imho this IS the way they are consistent with color slot handling (but not consistent with built-in palette handling). Imho this would reduce the confusion not increase, but I respect your decision.

PR #3631, opened 2024-01-01
feat: bus status usermod

casesolved-co-uk:
Works roughly similar to the STATUSLED functionality.

Now implemented as an effect usermod, description in the README

### DIFF
--- a/usermods/usermod_v2_busstatus/README.md
+++ b/usermods/usermod_v2_busstatus/README.md
@@ -1,0 +1,41 @@
+# Bus Status Usermod
+
+v2 usermod to display bus/port status on a configured LED segment
+Use instead of the builtin STATUSLED
+
+WLED uses the term `bus` to refer to a logical port. This may be a physical collection of gpio pins or a virtual
+network bus.
+
+This usermod finds the type of each specified bus, plus the device status, and displays the result on configured LEDs.
+The bus status colours are configurable and listed at the end.
+
+The device status colours are:
+* Red device element = Error value
+* AP mode = Blue
+* Connected to Wifi = Green
+* MQTT Connected = Cyan
+
+The status is also shown on the UI Info page.
+
+## Define Your Options
+
+* `USERMOD_BUSSTATUS`   - have this usermod included
+* `BUSSTATUS_PINS`      - (optional) Array of GPIO pin numbers to display the status of, in order. e.g. `-D BUSSTATUS_PINS='{1,2,-1}'`
+                        - `-1` displays the device status. Default: { -1, 1, 2}
+
+## Configure
+
+* Set up your pixel strips in `LED Preferences`. For a racitup.com v1 board the settings for the status bus are:
+  - Type = WS281x
+  - LED order = GRB
+  - Length = 3
+  - GPIO = 12
+  - Skip = 0
+  - `Make a segment for each output`: ticked
+  - Ensure the automatic brightness limiter is set up appropriately for your number of LEDs, or disable
+
+* Configure the associated segment. The bus type colours below are configurable:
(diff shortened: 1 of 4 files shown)


blazoncek:
The changes to core functionality are unacceptable just to support usermod unless done wisely with the possibility to be used in other usermods or in core. I do not see such usability here.

casesolved-co-uk:
> The changes to core functionality are unacceptable just to support usermod unless done wisely with the possibility to be used in other usermods or in core. I do not see such usability here.

Could you be more specific and make some suggestions please?

The changes to core functionality are fully backward compatible.

blazoncek:
If you want a *status* to be displayed on a segment, why bother with all this?

Just make a custom effect function, that can be applied to any segment. You can append your custom effect to regular effects using `strip.addEffect()` in your `setup()`.

Do not complicate things when they don't need to be complicated. 😉

blazoncek:
Similar to #3574

casesolved-co-uk:
> If you want a _status_ to be displayed on a segment, why bother with all this?
> 
> Just make a custom effect function, that can be applied to any segment. You can append your custom effect to regular effects using `strip.addEffect()` in your `setup()`.

Thanks for the tips. How do you suggest I make the segment:

- Always on, freeze doesn't appear to work when changing another segment
- Static, for example having the breathe effect on an adjacent pixel also makes the status fade

blazoncek:
Just a regular segment with regular (custom) effect applied. But I'm not sure I understand what you want.
And BTW *never* force push your changes. Do normal commits and normal pushes.

casesolved-co-uk:
> Just a regular segment with regular (custom) effect applied. But I'm not sure I understand what you want.

This is not true.

- If you try putting the Breathe and Solid effects on adjacent segments (and busses), the Solid effect is still faded in and out 
- Freeze of the Solid segment causes it to fade to black

version v0.14.1-b2
Should I start a new bug issue?

blazoncek:
> This is not true.

How so? Perhaps I am the only one who knows how WLED works as I wrote the whole damn thing regarding segments, transitions and effect blending. 😁 

>     * If you try putting the Breathe and Solid effects on separate segments (and busses), the Solid effect is still faded in and out

Faded in what way? Are we talking about fading aka transitions or displaying a WLED status (whatever that means to you) on a part of the strip (aka segment)?
If you are talking about transitions (when powering WLED on or off) then yes *all* segments are transitioned (faded) in and out. When WLED is in off state nothing is displayed on strips except if using STATUSLED (aka `setStatusPixel()`).
If you are talking about displaying separate content of segments then each and every segment can display separate effect.

>     * Freeze of the Solid segment causes it to fade to black

Freeze does not do that. It only stops updating segment with effect data.

> version v0.14.1-b2 Should I start a new bug issue?

IDK but it looks to me that maybe we are not talking about the same thing.
It will be very difficult to convince me to approve current implementation as I think the "status usermod" could be done without changing WLED core functions.

casesolved-co-uk:
> Faded in what way? Are we talking about fading aka transitions or displaying a WLED status (whatever that means to you) on a part of the strip (aka segment)?

I've posted a video showing the problems. The implementation of this usermod has now been changed to just be an effect as you requested. The issues I'm seeing are part of the core functionality because I used the official release build for the video; it did **not** include this usermod.

blazoncek:
Hardware (power) issue, not WLED functionality.

softhack007:
closing - this PR had no activity since >20 month, its not compatible any more with WLED 16.0.0-beta, and the purpose / added value is not clear to me.

Plus the PR is full of force-pushed commits, and a lot of information seems to be missing/lost from the review history.

Issue #3417, opened 2023-10-04
Enhanced Layering facilities


RadianM:
Combining multiple effects into one pattern _is relatively easily done_ by overlaying certain effects in overlapping Segments but it's a bit hit and miss. Only a handfull of effects actually cater for this. I think the potential for combining effects is underestimated.

Embracing "Photoshop like layering" would literally add many new dimensions to the ability of WLED to allow the creation of interesting composite patterns. For example, I hacked some existing patterns with code such as...

```
aRGB=effectPixel(i);   //pixel created by effect
if(overlay){
 aRGB.nscale8_video(SEGMENT.intensity);  //"Opacity"
 bRGB=SEGMENT.getPixelColor(i);          //underlying pixel
 bRGB|=aRGB;                             //"or" operator brings each channel up to the higher of the two values
 SEGMENT.setPixelColor(i, bRGB);
}
```
..to make them mix the RGB from the effect with the underlying RGB from lower numbered segments.

Using code tweaks such as the above I created a "Bonfire night party" with a "smoky sky" background in Segment 0 (Courtesy of Aurora) overlayed with licking bonfire flames (Coutesy of Fire 2012) and the fireworks effect. All pretty simple to arrange - but the results are so much more engaging than the effects in isolation,

Unfortunately the Segment controls don't make life quite so easy (although recent fixes to issues #3403 #3405 have helped). The Segment brightness slider is the main issue: as it's applied to the pixels written by the effect, it also affects the brightness of all layers beneath. For layering applications, as a crude solution, it could be exposed to the effect code which would apply it before merging with underlying pixel data. In this way it could operate more like "Layer Opacity". But ideally, it would be applied in interactions with the bus_manager.

One other unfortunate consequence of the layering as-is is that Mirror effect also mirrors all underlying segments as the mirror is applied to the pixels as they arrive via the setPixelColor (the same problem as with brightness I mentioned above) so it does seem to make it vital for layering to be handled deeper down in the architecture (as opposed to being up to each individual effect)

Then, along with a new drop-down menu for different "Layer mode" options (simulated in image below), it could be used to apply the overlay using some of the popular image blending techniques and make layering a possibility for all effects. How much of this could practically be rolled into the existing code I'm not sure. I'm still finding my way around it. 

Proposal for enhanced Segment controls:
![Screenshot 2023-10-04 09 55 09](https://github.com/Aircoookie/WLED/assets/13075243/8937d55c-e927-4405-bfa1-3c8a0b32b00d)

[...]

PR #3633, opened 2024-01-01
[RFC] Transition styles

tkadauke:
This is work in progress and I am really just looking for feedback suggesting if this is useful and the if the implementation looks ok. Notably, this lacks 2D support, mainly because I have no 2D led matrix (amazon order on the way).

My main goal is to have a transition where the lights in a 1d strip turn on left to right and then stay on. As I looked into how to do this, I noticed that WLED doesn't have native support for transitions other than crossfade. In online forums, the suggestion was to use playlists, but that turned out to be quite tricky to get right. Diving into the source code, I noticed that it's actually not that hard to do "properly": we can support multiple transition styles (this PR implements 5 so far in addition to the default "fade" style, see below), and we can transition every time anything changes, including color, brightness, and effect mode.

The styles implemented so far are:

* swipe right: transition pixels one by one from low index ("left") to high index ("right").
* swipe left: transition pixels one by one from high index ("right") to low index ("left").
* outside in: transition pixels from both ends of the segment going inward.
* inside out: transition pixels from the center of the segment going outward.
* fairy dust: transition pixels one by one in random (but stable) order.

It's possible to imagine a million more transition styles, and even more so in 2D. However, due to the way this is currently implemented, there are some limitations. For one, the transition is stateless, so it's calculated only using the current transition progress and the old and new pixel color for each pixel. Also due to the fact that transitions are calculated in the setPixelColor() method, we can't do fancy things like e.g. a transition where the old effect is pushed out of the segment to the right and the new one is coming in from the left. That would be doable, but we'd need to calculate the segment with both the old and the new settings and store the color values in separate buffers *before* applying the transition.

I'd love some feedback. Happy to provide a video that shows the transitional effects if that would help.


blazoncek:
Thanks for the effort and suggestion.

Unfortunately I have difficulty understanding how will this benefit effects that do not paint whole segment with *color* but leave portions black or are very dynamic. To my understanding it will also not *transition* effect display (the are still faded) but only color. With this approach only *static* effects benefit.

If you want to *properly transition* effects as well as color with different transition modes you would need to add a *mask* where each effect function can paint on (on segment). That would need to be stored in `_t` structure as it only applies while in transition. And the whole implementation needs to be in `service()` function.

To be honest with you, I've alredy tried to implement that (which would also allow effect layering) but it proved too complex and resource consuming. It would also not work on ESP8266 due to lack of memory.

softhack007:
Hi, interesting idea 👍 

I haven't looked into your PR source code yet, however the simples "fade in" in could imagine is by using the `Usermod::handleOverlayDraw()` callback. This function is called after all effects have completed their drawing. So to create a "fade-in from left", the overlay handler would simply overwrite pixels from right with BLACK; in the next iteration, overwrite fewer pixels. It would look like a stage curtain that opens.

This would even allow to make a usermod for fade-in and fade-out. The only tricky (but possible) part is to find the right trigger conditions.

Just my thoughts - not sure if you used the "overlay" mechanism already. I'll study your code after the weekend.

tkadauke:
Thank you for your comments, much appreciated. I quickly threw together a video that demos the transitions, hope that clarifies things: https://www.youtube.com/watch?v=tRhyBT0fvC0

> Unfortunately I have difficulty understanding how will this benefit effects that do not paint whole segment with color but leave portions black or are very dynamic. To my understanding it will also not transition effect display (the are still faded) but only color. With this approach only static effects benefit.

As you can see from the video, the transition styles work wherever color fading works today, including color, effect, and brightness. It is true that this approach fails for some effects, although I don't know why (example: Twinkle to Twinkle Fox), but what's interesting is that for the exact same transitions, the cross fade transition also doesn't work. I'm still digging into this some more.

> If you want to properly transition effects as well as color with different transition modes you would need to add a mask where each effect function can paint on (on segment). That would need to be stored in _t structure as it only applies while in transition. And the whole implementation needs to be in service() function.

Another way would be to have a buffer for both the before transition and after transition effects. I don't know if that is going to be prohibitively expensive though, especially for ESP8266. I'll look into what it'd take to do the masking approach.

tkadauke:
I just added a buffer for transitions. For now, it's basically a color buffer for each LED and it's always allocated. On my tiny setup (Dig2Go ESP32 with 300 LEDs), this takes up an additional 1.2 kb. For my largest controller, which is an 8 channel that powers 600 leds per channel, this would add another 20kb of memory usage. Given that I have over 170kb free heap space under load, this seems like an ok trade off to me, but I'd like to hear other opinions on the subject. There are ways to make this smaller, although they would come at a cost of runtime allocation/deallocation of buffer memory, which can lead to fragmentation.

I also added two more transition styles that wouldn't be possible without a buffer.

* push right: pushes the old effect to the right while pushing in the new effect from the left. This looks really cool and works well with most effects.
* push left: same but pushes things the other way around.

I also made sure that all of this works well with segments that use grouping and/or spacing.

I haven't tested this on an ESP8266, mainly because I don't have one :) Will get my hands on one soon I guess! If it doesn't work, it should be easy to disable this feature for ESP8266.

wled00/FX.h
```diff
@@ -903,6 +917,7 @@ class WS2812FX {  // 96 bytes
     uint16_t _length;
     uint8_t  _brightness;
     uint16_t _transitionDur;
+    std::vector<uint32_t> _buffer;
```
blazoncek, line 920: Why did you choose to allocate *segment* color buffer at strip level? I'm also curious why vector?

IMO a much better approach would be `static uint32_t *_buffer;` in `Segment` class with regular `malloc()/free()` if you wanted to keep only one buffer for all segments (as they are not used at the same time). The size would still need to be the size of largest possible segment which is `strip.getLengthTotal()`.

wled00/FX_fcn.cpp
```diff
@@ -739,7 +754,12 @@ void IRAM_ATTR Segment::setPixelColor(int i, uint32_t col)
     col = RGBW32(r, g, b, w);
   }
 
+  setPixelColorExact(i, col);
+}
+
+void Segment::setPixelColorExact(int i, uint32_t col) {
```
blazoncek, line 760: I do not think introduction of new function is necessary.

wled00/FX_fcn.cpp
```diff
@@ -760,16 +779,26 @@ void IRAM_ATTR Segment::setPixelColor(int i, uint32_t col)
         indexMir += offset; // offset/phase
         if (indexMir >= stop) indexMir -= len; // wrap
 #ifndef WLED_DISABLE_MODE_BLEND
-        if (_modeBlend) tmpCol = color_blend(strip.getPixelColor(indexMir), col, 0xFFFFU - progress(), true);
+        if (_modeBlend) {
```
blazoncek, line 782: This is entirely wrong approach.
`_modeBlend` is a flag to indicate that `setPixelColor()` needs to take into account underlying pixel color. If you want to implement correct *transitioning* you need to take into account that segments may overlap.

And you guessed it correctly, this is the place where your *color* transitions should apply.

wled00/FX_fcn.cpp
```diff
@@ -760,16 +779,26 @@ void IRAM_ATTR Segment::setPixelColor(int i, uint32_t col)
...
+#else
+        strip.setPixelColor(indexMir, col);
 #endif
-        strip.setPixelColor(indexMir, tmpCol);
       }
       indexSet += offset; // offset/phase
       if (indexSet >= stop) indexSet -= len; // wrap
 #ifndef WLED_DISABLE_MODE_BLEND
-      if (_modeBlend) tmpCol = color_blend(strip.getPixelColor(indexSet), col, 0xFFFFU - progress(), true);
+      if (_modeBlend) {
```
blazoncek, line 794: Same as above.

wled00/FX_fcn.cpp
```diff
@@ -837,13 +866,101 @@ uint32_t Segment::getPixelColor(int i)
   }
 #endif
 
+  return strip.getPixelColor(getPixelIndex(i));
+}
+
+uint32_t Segment::getBufferColor(int i) {
```
blazoncek, line 872: Unnecessary if you'd be using `Segment::_buffer[]`.

wled00/FX_fcn.cpp
```diff
@@ -837,13 +866,101 @@ uint32_t Segment::getPixelColor(int i)
...
+        uint32_t newCol = strip._buffer[index];
+
+        setPixelColorExact(i, transitionColor(i, oldCol, newCol));
+      }
+      break;
+    }
+  }
+}
+
+uint32_t Segment::transitionColor(int n, uint32_t oldCol, uint32_t newCol) {
```
blazoncek, line 925: This entire function would need to be incorporated into `currentColor()` to make it transparent.

wled00/FX_fcn.cpp
```diff
@@ -837,13 +866,101 @@ uint32_t Segment::getPixelColor(int i)
...
   if ((i >= stop) && (stop>0)) i -= length(); // avoids negative pixel index (stop = 0 is a possible value)
-  return strip.getPixelColor(i);
+  return i;
+}
+
+void Segment::blendTransition() {
+  switch (transitionStyle) {
+    case TRANSITION_STYLE_PUSH_RIGHT: {
+      uint16_t len = virtualLength();
+      uint16_t pos = (float(progress()) / float(0xFFFFU)) * len;
```
blazoncek, line 891: Avoid using `float` (or `double`).

wled00/FX_fcn.cpp
```diff
@@ -1201,15 +1319,20 @@ void WS2812FX::service() {
         [[maybe_unused]] uint8_t tmpMode = seg.currentMode();  // this will return old mode while in transition
         delay = (*_mode[seg.mode])();         // run new/current mode
 #ifndef WLED_DISABLE_MODE_BLEND
-        if (modeBlending && seg.mode != tmpMode) {
+        if (modeBlending && seg.isInTransition()) {
```
blazoncek, line 1322: Effect blending only happens when actual effects (mode) change.
This change is incorrect.

blazoncek (changes requested):
I tried to provide some constructive feedback for areas that need clarifying.

My suggestion for further development is to configure ESP with at least 2048-4096 LEDs (4 or more outputs) with and without matrix configuration. You do not need actual LEDs attached but they will help. For ESP8266 do not go beyond 256 LEDs in matrix configuration or beyond 1024 in strip configuration.

And of course, add 2D transitions as that will add a whole new level of complexity.

wled00/FX.h
tkadauke: I agree that we only need the buffer for the largest segment. I was just wondering what typically happens while WLED starts up. I'm imagining that on startup, WLED will initialize a single segment that contains all LEDs. Then as the boot preset it loaded, the largest segment might be smaller than that. Is that true?

Either way, personally I use many of my light strips sometimes with multiple segments and sometimes with a single segment. So quite often, the largest segment is the entire strip for me. The reason why I went with a single buffer at the strip level is that I only ever have to reallocate that buffer if the strip configuration changes, which should happen much more rarely than changes to segments.

The tradeoff here is using the largest amount of memory all the time vs. potentially having to reallocate the buffer, which can lead to memory fragmentation and also costs CPU time. I also wonder if we should reallocate every time the largest segment changes? Or only if the largest segment size *increases*? I'm new to this platform, so I don't know what it is usually optimized for, so happy to go with either suggestion.

As per `vector` vs. `new uint_32[...]` vs `malloc()` ... I don't really care other than that it's harder to introduce memory bugs with vectors. I realize there's a small overhead of two integers (capacity and size). OTOH, there are already 5 other vectors in the `WS2812FX` class. Curious, any specific reason you think that in this particular case, `malloc()` is preferable?

wled00/FX_fcn.cpp
tkadauke: The reason I added that function is because I call it above, but also while rendering the transitions. The `setPixelColor()` function corrects the color for brightness first before setting the pixel in the strip. However, when calculating transitions, that correction has already happened, since we have called `setPixelColor()`, either storing the corrected color in the strip or in the buffer. So when I start moving pixels around in the transition, I have to use the already-corrected color, which is what `setPixelColorExact()` does.

wled00/FX_fcn.cpp
tkadauke: Happy to decouple my implementation from `_modeBlend`. I was building on top of your blending implementation, since what you implemented is basically the default transition style in this PR. A lot of the effect transition code was reused in my implementation.

wled00/FX_fcn.cpp
tkadauke: I would still have to calculate the pixel index, right? So it would be `Segment::_buffer[getPixelIndex(i)]`. I tried to implement this as an inline function, but the compiler yelled at me. Happy to revisit this or make it a macro.

wled00/FX_fcn.cpp
tkadauke: Yeah that was still on my todo list. Thanks for reminding me! :)

wled00/FX_fcn.cpp
tkadauke: Yeah, since I'm hooking into your effect blending implementation, a change to this line is necessary (however, it doesn't have to be exactly like this). In a previous version, I wrote

```
if (modeBlending && seg.isInTransition() && (seg.mode != tmpMode || transitionStyle != TRANSITION_STYLE_FADE)
```

which did work, but I noticed that changes to effects would ignore the selected transition style. Now, it's debatable if this is a bug or a feature, so I'm happy to revert back to the above change. Or as mentioned above, implement transition styles independent of the effect blending implementation.

tkadauke:
Thanks for your comments, this is really helpful! I have a whole bunch of development hardware shipping from Amazon, so hopefully will be able to stress test this in different environments and also start working on 2d soon.

wled00/FX.h
blazoncek: Yes, the approach is correct, but why in `WS2812FX` class? I think it fits better in `Segment` class as a `static` as it is only used for segment functions.

The purpose of vectors used in `WS2812FX` is to simplify dynamic *expansion* while maintaining simplicity of arrays. Unless heap becomes low, memory is never deallocated from vector. In your case you should only allocate memory once and never release or reallocate it.

wled00/FX_fcn.cpp
blazoncek: You got that wrong. This line does not need to change.
The code within `if` needs to execute only if effect blending is actually happening. The actual change needs to happen in the blending part (in `Segment::setPixelColor()`) but that might be challenging without introduction of "layering" (see my other posts on that matter). It will also require a bit more resources (and hence ESP32 only).

wled00/FX_fcn.cpp
blazoncek: I have reconsidered and this should replace `SEGCOLOR()` macro in some smart way.
`SEGCOLOR()` macro is used throughout effect functions to get currently selected colors (you'd need to implement transitions for all 3 color slots) which ATM `WS2812FX::service()` packs into `strip` global/public variables.

wled00/FX_fcn.cpp
blazoncek: See my other comment regarding `SEGCOLOR()` macro.
With current *mode blending* I managed to do actual effect transitions from one effect to another in a smooth way without any change to effect functions. The goal of this PR should be the same while allowing different transition styles.
And these styles need to apply to color (possibly taken from palette as well) and to effects.

So you effectively have two transitioning methods to implement.

wled00/FX_fcn.cpp
blazoncek: Effect will do something like this:
```c++
uint32_t col1 = SEGCOLOR(1);
uint32_t col0 = SEGMENT.color_from_palette(palEntry, false, PALETTE_SOLID_WRAP, 0);
for (unsigned i=0; i<SEGLEN; i++) SEGMENT.setPixelColor(i, color_blend(col1, col0, progress);
```
or
```c++
SEGMENT.setPixelColor(i, color_blend(SEGCOLOR(1), SEGMENT.color_from_palette(palEntry, false, PALETTE_SOLID_WRAP, 0), progress));
```
And from this snippet you can immediately see where the problem lies. `SEGCOLOR()` is unaware of the pixel that is being retrieved. Same goes for `SEGMENT.color_from_palette()`. The only place where relation to pixel index happens is in `Segment::setPixelColor()` and `setPixelColor()` receives the exact color that it needs to paint.
Unfortunately you do not know if this color is from *color slot* (`SEGCOLOR()`) or from *palette* or from somewhere else. How would you transition it such case?
IMO not without effect layering and then blend two layers according to transition mode. Each layer would need to be painted separately with original and transitioned colors/palettes.

tkadauke:
Based on your feedback, I think I found a way to do all of this without a buffer.

During transitions, we render the new effect/color/brightness first, and then the old one. Your blending feature looks at the `progress()` in `setPixelColor()` and then blends the colors. I think I can change the code so that in `setPixelColor()`, we take the transition style into account right away, just like you're doing with the color blending.

* For the fade transition style, nothing changes.
* For the swipe/inside out/outside in transition styles, in the end we either want the old or the new color for a particular pixel. In `setPixelColor()` we can determine if that pixel will end up with the old or the new color. We can simply not call `strip.setPixelColor()` for the color that isn't going to end up there.
* For the push left/push right transition styles, we need to translate pixel coordinates. Again, we already know how a pixel will be translated when `setPixelColor()` is called, so we can simply adjust the coordinate.

I can imagine transition styles that we *really* can't do without a buffer (for example if we do coordinate transformations other than translate), but I think those are outside of the scope of this PR.

Does that sound like what you've been suggesting?

blazoncek:
You are effectively saying that "transitions" (as you call them) are always effect "blending" (as I call it) regardless of the fact that effect does not change.

Yes, with such approach this could work and that was at the back of my mind all along. Such approach will inevitably reduce frame rate at large pixel count as each frame will need to be calculated twice during transitions.

But hopefully not much more than it is now.

I am still baffled how you'll handle palettes and all three color slots.

tkadauke:
FWIW, this is roughly the approach I'm using now, with the difference that I render the pre-transition frame into a buffer and then mix the colors together at once. The reason I'm changing the line in `service()` is so that we can apply all transition styles for every change that would trigger a transition (color, brightness, effect), except palette (for now). Not sure if you've watched the video, but it demonstrates much of that.

It's possible to get rid of the buffer at the expense of code complexity, so I'm not sure if that's the right approach.

Have you had a chance to try my changes?

wled00/FX_fcn.cpp
tkadauke: It took me a while but I think I understand your point now. What I call "color" here and what is generally known as "color" are different things. Here, it's the final pixel color after the transition style is applied. Generally, "color" is the color slot that the user picks in the UI. Clearly, this method is poorly named. I'll rename it.

blazoncek:
No unfortunately not yet. Working on #3641

tkadauke:
I tried implementing transition styles without the buffer (see https://github.com/Aircoookie/WLED/pull/3633/commits/fcf190e52972f073237c3e205059562dfe28ae94), but when I added support for 2D transition styles, it became clear quickly that because of the way many of the effects are implemented, this is not going to work. There are really two problems: 1) many 2D effects do not draw every pixel and 2) many 2D effects alter the previous frame in some way, e.g. through fading. You can actually see the problem on the current release build when you enable effect blending and then switch between Akemi and Black Hole back and forth (make the transition time long to make it more obvious). If we really want to have transitions that look great under all circumstances, we actually need two buffers. One for the old mode and one for the new mode. Each need to be rendered into a buffer and at the end we combine them into the final display.

In https://github.com/Aircoookie/WLED/pull/3633/commits/3e6e25d21a26774faa1eaf45542f4387f19a682b I implemented double-buffered transitions. I realize that this adds 8 bytes per pixel, which might be too much in some circumstances. I plan on gating that behind an option that is disabled by default to allow people to save memory. That said, for 4800 LEDs (which is the maximum I have on one controller), it'd take up 38kb of memory. On the plus side, this fixes the issue with effect blending.

I currently don't see a way where we can reuse the buffer across segments. The reason is 2) above. In the general case, we need to retain the previous frame in order for the next frame to be correct. This is not just a theoretical concern. Transitions between some effects actually look pretty obviously wrong if we clear the previous frame or reuse the buffer across segments.

blazoncek:
Please re-read my comments in #3417.

Doing what you are trying to do will not help and will only slow down FPS. Of course I may be wrong as I have yet to thoroughly review your code but my experience wasn't that good.
I still keep layer blending branch locally and can push it for you to examine. Unfortunately it is not in working condition.

As mentioned in #3417 there are several problems to blend different effects/transitions/colors with current implementation of effect and segment logic including the ones you describe. That is, if you want to make blending work in any situation (you haven't touched overlapping segments).

Sometime early in 0.14 I introduced `Segment::setUpLeds()` which was used as a segment buffer for effects that relied on exact previous pixel values but was removed in favour of global LED buffer. It could be re-introduced but that will require updating any effect that misbehaves. The downside is memory fragmentation and memory demands the upside is that segment has its own canvas to draw on, unrelated to actual LED output.

tkadauke:
I understand your concerns as this introduces a bunch of resource requirements. But I'd like to invite you to try it and see for yourself. When I finally got it working last night, the transitions (especially push left/right) looked magical. As for FPS, I tested 16x16, 32x32, and 64x64, and transition styles didn't slow things down noticeably. However, I don't know exactly how to measure FPS beyond the info screen, so would appreciate some pointers.

wled00/FX_2Dfcn.cpp
```diff
@@ -192,29 +197,113 @@ void IRAM_ATTR Segment::setPixelColorXY(int x, int y, uint32_t col)
...
+  uint16_t width = virtualWidth();
+  uint16_t height = virtualHeight();
+
+  // Sometimes the timing works out such that "pos" calculated below doesn't hit
+  // all the possible coordinates. This can happen e.g. when the frame rate is low.
+  // As a result, there can be some pixels left in a transitional state. This is a
+  // problem if the target effect doesn't update all the pixels. In order to ensure
+  // there are no artifacts left from the transition, we need to clear the segment
+  // before we render the transition.
+  fill(BLACK);
```
blazoncek, line 232: Can you think of a better way to handle this?
`fill()` uses `setPixelColor()/setPixelColorXY()`.

wled00/FX_fcn.cpp
```diff
@@ -92,9 +93,15 @@ Segment::Segment(const Segment &orig) {
   _t = nullptr; // copied segment cannot be in transition
   name = nullptr;
   data = nullptr;
+  buffer1 = nullptr;
+  buffer2 = nullptr;
   _dataLen = 0;
   if (orig.name) { name = new char[strlen(orig.name)+1]; if (name) strcpy(name, orig.name); }
   if (orig.data) { if (allocateData(orig._dataLen)) memcpy(data, orig.data, orig._dataLen); }
+
+  uint16_t len = virtualLength();
```
blazoncek, line 102: Although `Segment` content was copied it may be clearer to use `orig.length()`.
Do not use `virtualLength()` as its return value changes when using grouping, spacing and mirroring.

wled00/FX_fcn.cpp
```diff
@@ -299,6 +308,12 @@ void Segment::startTransition(uint16_t dur) {
   _t->_cctT           = cct;
 #ifndef WLED_DISABLE_MODE_BLEND
   if (modeBlending) {
+    uint16_t len = virtualLength();
```
blazoncek, line 311: Same as above, use `length()`

blazoncek (commented):
Looks good on paper. Though I still have to test it.
Let's ask @softhack007 and @Aircoookie for their opinion.

wled00/FX_2Dfcn.cpp
tkadauke: Curious, what's the problem with that? The goal is to clear all pixels, taking segment grouping, spacing and mirroring into account. Is there a simpler or more efficient way?

wled00/FX_fcn.cpp
tkadauke: In the latest version, I am still using `virtualLength()` but I'm dynamically growing the buffer if needed.

tkadauke:
Alright, I think this is taking shape now. I just added dynamic loading of available transition styles, just like it's done with effects. This way, if effect blending is disabled, we don't show the the drop down in the UI. I'm also filtering out styles that are 2D-only if the strip is 1D.

I also added palette transitions now. It's not using the same mechanism as palette blending, since that wouldn't work with some of the transition styles.

I think at this point, this covers the full scope of the PR. I'll add a more transition styles in future PRs. So I'll mark this as ready for review.

wled00/FX.h
```diff
@@ -962,6 +972,13 @@ class WS2812FX {  // 96 bytes
       setUpSegmentFromQueuedChanges(void);
 };
 
+inline uint32_t hashInt(uint32_t s) {
```
blazoncek, line 1024: Please move this function into `util.cpp` as it may be used in the same manner as `crc16()`

wled00/FX.h
```diff
@@ -320,11 +320,15 @@
 #define TRANSITION_STYLE_FADE            0
 #define TRANSITION_STYLE_SWIPE_RIGHT     1
 #define TRANSITION_STYLE_SWIPE_LEFT      2
-#define TRANSITION_STYLE_PUSH_RIGHT      3
-#define TRANSITION_STYLE_PUSH_LEFT       4
-#define TRANSITION_STYLE_OUTSIDE_IN      5
-#define TRANSITION_STYLE_INSIDE_OUT      6
-#define TRANSITION_STYLE_FAIRY_DUST      7
+#define TRANSITION_STYLE_SWIPE_UP        3
```
blazoncek, line 323: I'd group 2D transitions together.

wled00/FX_fcn.cpp
```diff
@@ -152,6 +153,27 @@ Segment& Segment::operator= (Segment &&orig) noexcept {
   return *this;
 }
 
+// allocate and/or resize effect buffers if necessary
+void Segment::allocateBuffers() {
+  uint16_t len = virtualLength();
```
blazoncek, line 165: This will yield unnecessary heap fragmentation. Please use `length()` instead as `length()` does not change when changes spacing, grouping or mirroring.

wled00/FX_fcn.cpp
```diff
@@ -692,6 +716,9 @@ void Segment::setPalette(uint8_t pal) {
   if (pal < 245 && pal > GRADIENT_PALETTE_COUNT+13) pal = 0; // built in palettes
   if (pal > 245 && (strip.customPalettes.size() == 0 || 255U-pal > strip.customPalettes.size()-1)) pal = 0; // custom palettes
   if (pal != palette) {
+#ifndef WLED_DISABLE_MODE_BLEND
```
blazoncek, line 740: Not sure I like this. Palettes have their own option for blending and this introduces inconsistency.

blazoncek (commented):
I think you should really use `length()` for buffer allocation and try to keep settings consistent.

blazoncek:
Oh, one more thing.
This will not get merged into 0.14. Please rebase to 0_15 branch.

wled00/FX_fcn.cpp
tkadauke: The original palette blending still works and takes precedence over this if enabled.

wled00/FX_2Dfcn.cpp
blazoncek: Well, IMO `fill()` is an effect function and may (in the future) employ different principles of filling a segment. If you look at the definition you'll see that it does not *fill* every pixel in every instance. It may skip pixels if spacing is used.
I think it would be wiser to employ classic buffer cleaning using a `for` loop or `memset()`.

wled00/FX_fcn.cpp
blazoncek: If `strip.paletteFade` is `false` and `modeBlending` is `true`, the palette will still be blended. This is not a desired outcome.

wled00/FX_fcn.cpp
```diff
@@ -1198,19 +1350,44 @@ void WS2812FX::service() {
...
+          Segment::renderToBuffer(seg.buffer2);
+          uint16_t d2 = (*_mode[tmpMode])();  // render old mode into buffer
+
+          // Clean up
           seg.restoreSegenv(_tmpSegData);     // restore mode state (will also update transitional state)
           delay = MIN(delay,d2);              // use shortest delay
           Segment::modeBlend(false);          // unset semaphore
+          Segment::renderToBuffer(0);
+
+          // Mix everything together
```
blazoncek, line 1380: It would be cleaner to just use one function and do `if (is2D()) {...}` in it.

blazoncek:
I think it only needs polishing touches. And testing on ESP8266.

tkadauke:
> This will not get merged into 0.14. Please rebase to 0_15 branch.

Do I need to open a new PR for that?

wled00/FX_fcn.cpp
tkadauke: That depends on the transition style. Yes, we start a transition if `strip.paletteFade` is `false` and `modeBlending` is `true`, but `currentPalette` was modified so that the palette is not blended, instead just like with `currentBri` and `currentColor` we get the transitional palette if `_modeBlend` is true. With the Fade style, this looks similar to palette blending. But with all the other styles the looks are very different.

Before I added this change, the transitions didn't take palette changes into account and looked quite odd.

wled00/FX_fcn.cpp
tkadauke: If you feel strongly about it, happy to combine. As of right now, it'd be a pretty long function. As I'm adding more transition styles, that'll only get worse. I'm thinking of ways to split up the implementation of the transitions, perhaps the same way as it's done with effects, where there's one function per effect. That way, we'd also get rid of the switch statement.

blazoncek: No, I do not have *strong feelings*. 😁 It just seemed cleaner.

wled00/FX_2Dfcn.cpp
blazoncek: ~~And then consider user changing spacing on segment without changing anything else. You'll get artifacts remaining.~~

I have checked the code and I actually added segment clearing prior to changing spacing, so all is good.
I had to mentally run through the implementation and it is indeed ok. At first I thought that overlapping segments would be overwritten but now I see that `setPixelColor()` is reused to do the actual painting in `renderTransition()`.

tkadauke:
I did the merge as you requested, but I can't set the target branch. Would appreciate if you could do that.

tkadauke:
I've got a rebased version of my changes. I could either force-push here or open a new PR.

blazoncek:
I'd prefer a new PR as this still needs ESP8266 verification and most likely exclusion as having 4+ LED buffers will be to much for it.
Why 4? NeoPixelBus has 2, BusManager uses one (if selected) and this will use 2 additional.

tkadauke:
Closing in favor of #3669.

PR #3639, opened 2024-01-03
Improve ETag Caching

WoodyLetsCode:
This also adds support for ETag caching for the settings pages, specifically for the following pages ETag caching is now enabled:
- settings_wifi
- settings_leds
- settings_ui
- settings_sync
- settings_time
- settings_sec
- settings_dmx
- settings_um
- update
- settings_2D
- settingsCss
- welcome
- settings
- liveviewws2D
- liveview
- settingsCss
- favicon
- usermod
- index
- pixart
- pxmagic
- cpal
- 404

I also fixed some issues with the previous caching not being [RFC 7232](https://datatracker.ietf.org/doc/html/rfc7232) compliant.
In [Section 4.1 of RFC 7232](https://datatracker.ietf.org/doc/html/rfc7232#section-4.1) it is defined that a _304 (Not Modified) status code indicates that a conditional GET or HEAD request has been received and would have resulted in a 200 (OK) response_. This is why I added a check if the status code was 200.
Another thing that the previous implementation did not do was that _the server generating a 304 response MUST generate any of the following header fields that would have been sent in a 200 (OK) response to the same request: Cache-Control, Content-Location, Date, ETag, Expires, and Vary._ This is why I'm now calling the `setStaticContentCacheHeaders` function in the `handleIfNoneMatchCacheHeader` function.

To eliminate duplicate code, I created the function `handleStaticContent`. I added a comment that explains what this function does. Basically, now instead of having blocks like this:
```c++
server.on("/page.htm", HTTP_GET, [](AsyncWebServerRequest *request){
    if (handleFileRead(request, "/page.htm")) return;
    if (handleIfNoneMatchCacheHeader(request)) return;
    AsyncWebServerResponse *response = request->beginResponse_P(200, "text/html", PAGE_page, PAGE_page_L);
    response->addHeader(FPSTR(s_content_enc),"gzip");
    setStaticContentCacheHeaders(response);
    request->send(response);
});
```
We now have this:
```c++
server.on("/page.htm", HTTP_GET, [](AsyncWebServerRequest *request) {
    handleStaticContent(request, "/page.htm", 200, "text/html", PAGE_page, PAGE_page_L);
});
```
Which is much shorter. 😄 

If you are wondering what the `eTagSuffix` is used for in the `handleStaticContent` function, it is currently not used, but I added it to have more control over caching. In the future this could be used if we want to use ETag caching for `fxdata` or `effects`.

I also fixed a small bug in `cdata.js` that caused `favicon_length` to have the wrong value.


wled00/wled_server.cpp
```diff
@@ -15,8 +15,9 @@
  * Integrated HTTP web server page declarations
  */
 
-bool handleIfNoneMatchCacheHeader(AsyncWebServerRequest* request);
-void setStaticContentCacheHeaders(AsyncWebServerResponse *response);
+bool handleIfNoneMatchCacheHeader(AsyncWebServerRequest* request, int code, const String &eTagSuffix = "");
```
blazoncek, line 18: Could you avoid String?

blazoncek (approved):
Well done! 👍

wled00/wled_server.cpp
WoodyLetsCode: Sure, we could also use `uint16_t`. I'll change it later.

blazoncek: I know that String has some advantages but if used improperly it may fragment memory.
Do whatever you want.

PR #3640, opened 2024-01-03
can bus implementation in a usermod file

nor24o:


nor24o:
add the implementation to main branch

PR #3654, opened 2024-01-07
Upgrade the MPU6050 usermod

willmmiles:
Update the MPU6050 usermod with a number of new features:

- Configurable interrupt pin (and polling mode)
- Configurable calibration values
- UMData output, for consumption by other subsystems
- Clean up some of the compile warnings due to overlapping defines
- Shim the info output, so as to work with the current web UI
- Add a demo usermod that applies the gyro data in a simple way

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -198,6 +198,8 @@ lib_deps =
   #For ADS1115 sensor uncomment following
     ; adafruit/Adafruit BusIO @ 1.13.2
     ; adafruit/Adafruit ADS1X15 @ 2.4.0
+  #For MPU6050 IMU uncomment follwoing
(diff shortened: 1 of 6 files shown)


blazoncek:
I am sure @ewoudwijma will appreciate this and will most likely make something out of the enhancement.
Unfortunately I cannot test but I guess that is not necessary. 😉

ewoudwijma:
Yeah probably. Made one gyro / accelerometer reactive effect in the past. And wanted to make the Pong game using G/A.
But never finished it. Feel free to finish this
(Code is in WLED MM, but would be nice to merge it in WLED AC if Pong is working with this 🙂)

Issue #3607, opened 2023-12-22
Usermod improvements Animated_Staircase: 


Maveric2005:
Usermod improvements Animated_Staircase: 

Here are a few suggestions for better customisation

- Adjustability whether the direction of movement should be after the animation is completed (i.e. when the staircase slowly goes out). Whether the same running direction as when it starts. Always downwards or always upwards.

- Option to set the fade time

- Different times can be set for animation upwards or downwards (depending on which sensor is triggered)

blazoncek:
Please talk to original author.
Usermods are author's responsibilty unless core team members take that responsibility.

PR #3657, opened 2024-01-08
Animated_Staircase: swipe off in the direction of the sensor detected

gannaramu:
current Animated Staircase swipes off in a fixed direction irrespective of which sensor triggered it. this was also mentioned in #3607


blazoncek:
@srg74 could you test (or your clients) this update?

srg74:
> @srg74 could you test (or your clients) this update?

Can’t confirm due to absence of testing rig.

PR #3658, opened 2024-01-08
Fix that color picker is not centered on screens with width between 1024px and 1249px

WoodyLetsCode:
This fixes that the Color Picker is not centerd on screens with width between 1024px and 1249px.

| Before | Now |
|--------|--------|
| Screenshot 2024-01-08 at 14-31-20 (L) WLED | Screenshot 2024-01-08 at 14-31-13 WLED |


blazoncek:
Interesting. FF for mac renders it centered.

PR #3666, opened 2024-01-09
Ignore autogenerated files with binary array for the Web UI and fix caching bug

WoodyLetsCode:
This PR removes all the `html_*.h` files and add them to `.gitignore`. As a result, we no longer encounter merge conflicts related to these files. PlatformIO automatically regenerates these files when the project is opened.


.github/workflows/wled-ci.yml
```diff
@@ -39,6 +39,11 @@ jobs:
         environment: ${{ fromJSON(needs.get_default_envs.outputs.environments) }}
     steps:
     - uses: actions/checkout@v3
+    - name: Set up Node.js
+      uses: actions/setup-node@v4
+      with:
+        cache: 'npm'
+    - run: npm install
```
softhack007, line 46: Maybe add "npm audit fix" as another step?

It tries to update components with know security vulnerabilities, unless there are "breaking changes".

WoodyLetsCode: I would rather manually execute `npm audit fix` and then commit the new `package-lock.json`, so that everyone has the fixes for security vulnerabilities.
I was planning to rework `cdata.js` more thoroughly in the future anyway and update all dependencies. This would also give me the opportunity to run an `npm audit fix`. But I would have done that in a separate PR.

WoodyLetsCode:
With commit [693fb97](https://github.com/Aircoookie/WLED/commit/693fb97fa6d8f6058803436c55faf7094ac3178b) I fixed the bug that if you change inlined files like index.js, the script does not recognize the changes. Now we don't have the cache file `cdataCache.json` anymore. The script now simply rebuilds everything if a file in `wled00/data` has changed. Or more precisely, if the mtime of any file in `wled00/data` is greater than the lowest mtime of all html_*.h files, then the script rebuilds the files.

blazoncek:
@WoodyLetsCode can you resolve conflicts?

WoodyLetsCode:
@blazoncek done

PR #3670, opened 2024-01-10
Disable custom CSS by default

WoodyLetsCode:
This will disable custom CSS by default.
I see no reason to have this enabled by default since there is no default skin.css file.
This just costs an unnecessary request on every load.
If someone wants to use custom CSS then they can enable it.


blazoncek:
The idea of loading custom CSS by default is just that: load custom CSS regardless of configuration.
I use it daily on most of my devices to distinguish what purpose they serve.

PR #3673, opened 2024-01-11
Add ANIMartRIX usermod

netmindz:
Creating fresh PR as the previous one had something very odd going on with git, the changes were shown but was missing the first commit that actually added the files. Something clearly gone wrong with the parent branch.

See https://github.com/Aircoookie/WLED/pull/3222


netmindz:
One thing I forgot to mention on the PR before was that we can't just add to the dependencies folder as this code is under a different license

PR #3674, opened 2024-01-11
Better sorting and naming of static palettes in the Custom Palette Editor

WoodyLetsCode:
This alphabetically sorts the static palettes of the Custom Palette Editor.
Also the hover text was changed from `Copy slot [id] palette to editor` to `Copy [name] palette to editor`.
| Before | Now | Pallete sorting in UI |
|--------|--------|--------|
| Screenshot 2024-01-11 233204 | Screenshot 2024-01-11 233123 | Screenshot 2024-01-11 at 23-37-41 WLED Tester | 

Now it should be easier to get the correct palette for editing.
This should be a _fix_ for the _bug_ that was reported in the discord  ⁠#beta-testing channel.

### DIFF
--- a/wled00/data/cpal/cpal.htm
+++ b/wled00/data/cpal/cpal.htm
@@ -201,6 +201,7 @@
   var gradientLength = rect.width;
   var mOffs = Math.round((gradientLength / 256) / 2) - 5;
   var paletteArray = []; //Holds the palettes after load.
+  var paletteName = []; // Holds the names of the palettes after load.
   var svgSave = '<svg style="width:25px;height:25px" viewBox="0 0 24 24"><path fill=#fff d="M22,12A10,10 0 0,1 12,22A10,10 0 0,1 2,12A10,10 0 0,1 12,2A10,10 0 0,1 22,12M7,12L12,17V14H16V10H12V7L7,12Z"/></svg>'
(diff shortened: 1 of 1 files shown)


wled00/data/cpal/cpal.htm
```diff
@@ -520,8 +521,10 @@ <h1 style="display: flex; align-items: center;">
     if (hst.length > 0 ) {
       try {
         var arr = [];
-        const response = await fetch('http://'+hst+'/json/info');
-        const json = await response.json();
+        const responseInfo = await fetch('http://'+hst+'/json/info');
```
blazoncek, line 524: Perhaps also use `getURL()` logic from `index.js` which will work behind reverse proxy.

blazoncek (commented):
Thank you.

wled00/data/cpal/cpal.htm
WoodyLetsCode: I did it like it was done before. But it you would prefer it we could also use `getURL()`.

WoodyLetsCode: The only problem I see is that if we do this we have to copy the code from index.js to cpal.htm and then we have created duplicate code again. We could wait with this until we have a `global.js` file that contains functions like `getURL()` that are used on every page, so we do not have to maintain the duplicate code. But I'll let you decide. Should I just copy the code now, or should we wait until we have a `global.js` file?

PR #3681, opened 2024-01-14
Make rainbow effects more colorful

TripleWhy:
I was playing around with some rainbow effects and was wondering why there was no real yellow color. Cyan and magenta are also mysteriously underrepresented. I think rainbow effects should use the full rainbow spectrum.

If you agree, that is what this PR does.  
Here are two images displaying the Rainbow effect on 256 LEDs, captured using the web UI:

rainbows

Top: 0_15 (874b24fb320c6dc16e509514b70965bb4ca7964b)  
Bottom: mine (760a1d45124903070a62944faf6f0f3e95e9bbaf)

This works by replacing the linear transitions R -> G -> B in `color_wheel` by the HSV transition 0° -> 360°.

## Summary by CodeRabbit

* **New Features**
  * Updated the color wheel to provide smoother and more consistent color transitions by rotating through HSV hues.

* **Bug Fixes**
  * Removed the "Use 'rainbow' color wheel" option from the LED settings in the user interface to streamline configuration.

* **Refactor**
  * Simplified color handling by removing the rainbow color wheel setting and related configuration options.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1082,7 +1082,7 @@ uint32_t Segment::color_wheel(uint8_t pos) const {
   if (palette) return color_from_palette(pos, false, false, 0); // only wrap if "always wrap" is set
   uint8_t w = W(getCurrentColor(0));
   uint32_t rgb;
-  hsv2rgb(CHSV32(pos, 255, 255), rgb);
+  hsv2rgb(CHSV32(static_cast<uint16_t>(pos << 8), 255, 255), rgb);
   return rgb | (w << 24); // add white channel
 }
 
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1076,26 +1076,23 @@ void Segment::blur(uint8_t blur_amount, bool smear) const {
 /*
  * Put a value 0 to 255 in to get a color value.
  * The colours are a transition r -> g -> b -> back to r
- * Inspired by the Adafruit examples.
+ * Rotates the color in HSV space, where pos is H. (0=0deg, 256=360deg)
  */
 uint32_t Segment::color_wheel(uint8_t pos) const {
   if (palette) return color_from_palette(pos, false, false, 0); // never wrap palette
   uint8_t w = W(getCurrentColor(0));
-  pos = 255 - pos;
-  if (useRainbowWheel) {
-    CRGB rgb;
-    hsv2rgb_rainbow(CHSV(pos, 255, 255), rgb);
-    return RGBW32(rgb.r, rgb.g, rgb.b, w);
-  } else {
-    if (pos < 85) {
-      return RGBW32((255 - pos * 3), 0, (pos * 3), w);
-    } else if (pos < 170) {
-      pos -= 85;
-      return RGBW32(0, (pos * 3), (255 - pos * 3), w);
-    } else {
-      pos -= 170;
-      return RGBW32((pos * 3), (255 - pos * 3), 0, w);
-    }
+  // These h and f values are the same h and f you have in the regular HSV to RGB conversion.
+  // The whole funciton really is just a HSV conversion, but assuming H=pos, S=1 and V=1.
+  const uint32_t h = (pos * 3) / 128;
+  const uint32_t f = (pos * 6) % 256;
+  switch (h) {
+    case 0: return RGBW32(255    , f      , 0      , w);
+    case 1: return RGBW32(255 - f, 255    , 0      , w);
+    case 2: return RGBW32(0      , 255    , f      , w);
+    case 3: return RGBW32(0      , 255 - f, 255    , w);
+    case 4: return RGBW32(f      , 0      , 255    , w);
+    case 5: return RGBW32(255    , 0      , 255 - f, w);
+    default: return 0;
   }
 }
 
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1079,21 +1079,11 @@ void Segment::blur(uint8_t blur_amount, bool smear) const {
  * Rotates the color in HSV space, where pos is H. (0=0deg, 256=360deg)
  */
 uint32_t Segment::color_wheel(uint8_t pos) const {
-  if (palette) return color_from_palette(pos, false, false, 0); // never wrap palette
+  if (palette) return color_from_palette(pos, false, false, 0); // only wrap if "always wrap" is set
   uint8_t w = W(getCurrentColor(0));
-  // These h and f values are the same h and f you have in the regular HSV to RGB conversion.
-  // The whole funciton really is just a HSV conversion, but assuming H=pos, S=1 and V=1.
-  const uint32_t h = (pos * 3) / 128;
-  const uint32_t f = (pos * 6) % 256;
-  switch (h) {
-    case 0: return RGBW32(255    , f      , 0      , w);
-    case 1: return RGBW32(255 - f, 255    , 0      , w);
-    case 2: return RGBW32(0      , 255    , f      , w);
-    case 3: return RGBW32(0      , 255 - f, 255    , w);
-    case 4: return RGBW32(f      , 0      , 255    , w);
-    case 5: return RGBW32(255    , 0      , 255 - f, w);
-    default: return 0;
-  }
+  uint32_t rgb;
+  hsv2rgb(CHSV32(pos, 255, 255), rgb);
+  return rgb | (w << 24); // add white channel
 }
 
 /*
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -518,7 +518,6 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
   CJSON(briMultiplier, light[F("scale-bri")]);
   CJSON(paletteBlend, light[F("pal-mode")]);
   CJSON(strip.autoSegments, light[F("aseg")]);
-  CJSON(useRainbowWheel, light[F("rw")]);
 
   CJSON(gammaCorrectVal, light["gc"]["val"]); // default 2.2
   float light_gc_bri = light["gc"]["bri"];
@@ -1040,7 +1039,6 @@ void serializeConfig(JsonObject root) {
   light[F("scale-bri")] = briMultiplier;
   light[F("pal-mode")] = paletteBlend;
   light[F("aseg")] = strip.autoSegments;
-  light[F("rw")] = useRainbowWheel;
 
   JsonObject light_gc = light.createNestedObject("gc");
   light_gc["bri"] = (gammaCorrectBri) ? gammaCorrectVal : 1.0f;  // keep compatibility
--- a/wled00/data/settings_leds.htm
+++ b/wled00/data/settings_leds.htm
@@ -905,7 +905,6 @@ Swap: <select id="xw${s}" name="XW${s}">
 			<option value="3">None (not recommended)</option>
 		</select><br>
 		Use harmonic <i>Random Cycle</i> palette: <input type="checkbox" name="TH"><br>
-		Use &quot;rainbow&quot; color wheel: <input type="checkbox" name="RW"><br>
 		Target refresh rate: <input type="number" class="s" min="0" max="250" name="FR" oninput="UI()" required> FPS
 		<div id="fpsNone" class="warn" style="display: none;">&#9888; Unlimited FPS Mode is experimental &#9888;<br></div>
 		<div id="fpsHigh" class="warn" style="display: none;">&#9888; High FPS Mode is experimental.<br></div>
--- a/wled00/set.cpp
+++ b/wled00/set.cpp
@@ -348,7 +348,6 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
     t = request->arg(F("TP")).toInt();
     randomPaletteChangeTime = MIN(255,MAX(1,t));
     useHarmonicRandomPalette = request->hasArg(F("TH"));
-    useRainbowWheel = request->hasArg(F("RW"));
 
     nightlightTargetBri = request->arg(F("TB")).toInt();
     t = request->arg(F("TL")).toInt();
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -623,7 +623,6 @@ WLED_GLOBAL unsigned long transitionStartTime;
 WLED_GLOBAL bool          jsonTransitionOnce       _INIT(false);  // flag to override transitionDelay (playlist, JSON API: "live" & "seg":{"i"} & "tt")
 WLED_GLOBAL uint8_t       randomPaletteChangeTime  _INIT(5);      // amount of time [s] between random palette changes (min: 1s, max: 255s)
 WLED_GLOBAL bool          useHarmonicRandomPalette _INIT(true);   // use *harmonic* random palette generation (nicer looking) or truly random
-WLED_GLOBAL bool          useRainbowWheel          _INIT(false);  // use "rainbow" color wheel instead of "spectrum" color wheel
 
 // nightlight
 WLED_GLOBAL bool nightlightActive _INIT(false);
--- a/wled00/xml.cpp
+++ b/wled00/xml.cpp
@@ -380,7 +380,6 @@ void getSettingsJS(byte subPage, Print& settingsScript)
     printSetFormValue(settingsScript,PSTR("TL"),nightlightDelayMinsDefault);
     printSetFormValue(settingsScript,PSTR("TW"),nightlightMode);
     printSetFormIndex(settingsScript,PSTR("PB"),paletteBlend);
-    printSetFormCheckbox(settingsScript,PSTR("RW"),useRainbowWheel);
     printSetFormValue(settingsScript,PSTR("RL"),rlyPin);
     printSetFormCheckbox(settingsScript,PSTR("RM"),rlyMde);
     printSetFormCheckbox(settingsScript,PSTR("RO"),rlyOpenDrain);


blazoncek:
Bravo! That boggled me as well.
Approved.

blazoncek:
One more thing: As to avoid angering certain users please make the new color wheel calculation optional.
Make an entry in LED settings (something like "Use HSV transition for color wheel") and a global selector (`WLED_GLOBAL bool hsvColorWheel _INIT(false);` ).

blazoncek:
Update: Please check `colors.cpp` and function `colorHStoRGB()`. You could optimize the code to reuse that function (which itself could be optimised to not use `float`).

blazoncek:
Tested this today but have to say that I fail to see any real difference on my LEDs.

TripleWhy:
Hm. You are right. There *is* a difference for me, but I can't really say that one better than the other. My version certainly is a bit brighter (obviously, now that I think about it). I have the cheapest LEDs I could find though.

Possibly there might be a greater difference with higher quality or future LEDs or setups. I mean there clearly is a difference on an LCD screen.

TripleWhy:
I saw that. Some more explicit feedback:

- I think I wouldn't keep both versions at the same time.
- Configuration option: I have no idea how this works in WLED, so I can't really comment on that.
- The setting itself is pretty non-descriptive, so I'd want some more information as a user.
- `colorHStoRGB()`: Looks good. Probably doesn't change much though ^^

TripleWhy:
Another option would be to keep the original function for ESP8266, and use the hsv function for ESP32.

I made some comparisons between the color computation of these functions:

### D1 Mini
|             | original | hsv    |  delta | % |
| ----------- | -------- | ------ | ------ | --------- |
| exec/ns     |   429.43 | 546.26 | 116.83 |  127.21 % |
| flash/bytes |   874987 | 875035 |     48 |  100.01 % |

### WT32-ETH01
|             |  original |     hsv |  delta | % |
| ----------- | --------- | ------- | ------ | --------- |
| exec/ns     |   159.56  |  170.03 |  10.47 |  106.56 % |
| flash/bytes |   1290933 | 1290957 |     24 |  100.00 % |

blazoncek:
> I made some comparisons between the color computation of these functions

Are you talking about `color_wheel()` and `colorHStoRGB()` or about *old* vs. *new* `color_wheel()`?

TripleWhy:
`color_wheel` original vs. mine. The versions of the function I benchmarked didn't contain the first two lines that read `palette` and `currentColor`, that's why I said this is about the color conversion only.

blazoncek:
Perhaps change `% 256` into `& 0xFF` and test again. It should provide same result.

TripleWhy:
[According](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGISRqkrgAyeAyYAHI%2BAEaYxBLSAA6oCoRODB7evv6ByamOAqHhUSyx8VK2mPYFDEIETMQEmT5%2BAZXV6XUNBEWRMXEJtvWNzdltCsM9YX2lA1IAlLaoXsTI7BzmAMxhyN5YANQmm27IE/iCR9gmGgCC1zcEAJ6JmFhU%2B15hBAAcAPoE%2B2ijwImCOVluWze032ACUAOIWADqmzMEDIwFI0VIAHd5vsIJ9BCj/lAgSCILjDscjm59mZJHiTAB2WkQMmYNGM6nHfZcABsjJZ%2BPZ0C5bhp%2B2%2BgtZIui83l83u90JBGJALQtBIv2xCEwVV%2BCAUADcCV8/gC8oLwTd9raPmb/vtsVSACL7DRg%2B52/YAeh9%2BwAKrqlPsEPtDOh9u8jWIvJgFOHiJh9gRdfsFKxk2GI1H9o9lqGmEbk2EU2mk8AvAZiPsABJCABqKdQsIR%2BzQDGLxFSAgAdF67X7A2mdZ5k1QvAxkIQBPsk2JaI99ngE9ovBNw3XG%2B2BF2ewxSICvACmAoFD4wsA60cXXlD0Ib1xwwxIw3H/3bt6OxuVWrQ678TyfYACp9k2PF/S4Mxvk9CEzG2d4AFkAHkXQASQAMQATQHW1vwBX8zEdd4b0AlIQP2AVDjMABWOkaL5WC7ngqolFwncGB/L4/xIzY3QgIDQKo8w%2BXdVQMIwpitlcPAqHYhRsUIZAwwgBArXY0QQw0EA50wAgVgYVskRRCBaLou1D3eb19kPDRrJsp1FU2a0v1PZMuB0pN9OIQz4WM1EzP2ABaKND0Ciz3Xsw9cSY1yQzMTy9IMozkVROzrLCmjzNtSyosc2K7U05NNkS7zfIRVKIHS71MrokKqFqmr8ucjS3P2SRSuSvzKqsprqoi8KcualzCramjOp8lKTMGhz%2BqGwL6uipyRttN4mCrAgJsMj0Ws/O1mRdJUmRdDhFloTgaN4PwOC0UhUE4cVLGsdNllWZMth4UhNpu07FgAaxAFFez5DQ%2BW%2BJkaK4b5NihyQ%2BQATn0ThJCuzReHujheAUEBAm%2BrRFjgWAkDQFhEjoOJyEoEmyfoeJ40SFFgDMDQuHh4gzD4OgQW7ShojRzEwgaR5OE%2B6JBeIR5kOibRMAcEXeBJthBGQhhF35rBoi8YA3AXbHuF4LAWEMYBxB%2B0h8CTBw8GLPXbswVRZePdZPq%2BKp%2BdoPBomIIWPCwfmCGIPAWHlxYqAMYAFAbPBMGxZCXmuz7%2BEEEQxHYCok/kJQ1H53QOYMIwUGsax9E97HIEWVBEhqPWgqoBhUCCxpHGCuuG/ttYq/SBMgrOG9gAYLwi8sLgnyCxF66CpgCADz2ne78eG6nmfNZBburyC5C182YLEQXYL67OD3oh35JXeISfA9THeJ7iYgSBvSd11eIKV57vSgpjQOmGiehr7b4g77EAfgPJQ6AP4NDwN/X%2BQUWCaiMKIWgtBu5UAnFOGob8rYCGQegKeTAMHVxbhMQODgP6eCnuTF%2BhAaBVHQPPG%2BAD758SwIkS2U9n5YF2A0chWC/5BVvowl0j9QG10nJgwyY8J5COfokLhbBua8NSP3IK1MGjJlrhMJgyA/pBRYagEEDgSAt3bpgTuPCx78KAXxJMJAsA1iCi6FC6FsKPnRl2QOWAy4QEWHYWWNQXAvlGH4LgQQXy9BKGUPQeQ0gCECZElI0SGBhP6OUdovjOiTFicEnxYiuiNCSbMFJGiRieBaHoIpUxijJIkN416axqlIw4JdUg11boY32AADRBJxJg%2BxsBCGUCifYwBkDIElL2MwvY7KmRZmzMweIIC4BnDWD68xeB41OqQVSKBUCk3JmQCgEBqZ7JAPTRmzNWbs05rQbm2M2T8zFswCW8sBaPMltLXxzzFaMAICrNWZsNZax1ogvWn1DbG1NrdC2aSbb83to7EEzzXbnTNkfb2EtfbrFujPYO%2BtQ7h0jtHWO8dnkZxTuIdOshFAqHUGbXQwT87GCHjYI%2BniK6mM4pwWuE8m54BbjfVQHcCE9wIOgPuA8mUj14UvQOK94xSunjKuewV16b2VdvMee8goHxFUfE%2BqAz4X0IGGCR/9AHAKfmA1%2BSgCDgK/j/NRC8%2BEMMsYIkBz9P6QPtcFWBAhhm6xbqgsR%2BCu4txwfUYNZiqBELwCQo0ZDHD0EoQQahtBaG8IsTeZhrCQRgI4dWbhnF03OvNcIwNNReFSLATI72ci4gKLwEolRSZCH1C0Tou%2B%2BiiB2KoMY9l88M1WMwDYut9jHGYRwnxVpbi8AePgN4qoaTnAQFcJkkJ6B8kROCVEmoq7t3pA3XMVJOSMklOyFkhdx7ugHsKSerIQShhXumOEuYNSVh1K4GdC6qMzZtM6YwDMvT%2BmDOGaM744zJn4nObM%2BZiyu3URhqsr6aN5SkABpsTYvYoa0Xht8GiLMNC0WhnyBpKNSDBy4BoQILT0acCxjjJDP0CYwEQCAfAKDKbUFkGStO0gM5Uuziiw0ehaCGjY1QJ4Lx6MiaUKoAOTBK5%2BMxqQYgQmOYqYUGJiT7BAjqftnJhT6RaOfsad%2B1pnAXSyXeFHGOda/3dMAwMswQyRljImVMqD7M8RGgTHZgDfTHPOdA%2BB9zMzPNrOQ/9QGkhxkUZomYPkMMaJMm%2BPDSQTISOmZo0p7GuMIsNLMJlu6tGGP40WHuQzkggA%3D%3D%3D) to godbolt's compiler explorer, that produces the same binary

blazoncek:
> [According](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGISRqkrgAyeAyYAHI%2BAEaYxBLSAA6oCoRODB7evv6ByamOAqHhUSyx8VK2mPYFDEIETMQEmT5%2BAZXV6XUNBEWRMXEJtvWNzdltCsM9YX2lA1IAlLaoXsTI7BzmAMxhyN5YANQmm27IE/iCR9gmGgCC1zcEAJ6JmFhU%2B15hBAAcAPoE%2B2ijwImCOVluWze032ACUAOIWADqmzMEDIwFI0VIAHd5vsIJ9BCj/lAgSCILjDscjm59mZJHiTAB2WkQMmYNGM6nHfZcABsjJZ%2BPZ0C5bhp%2B2%2BgtZIui83l83u90JBGJALQtBIv2xCEwVV%2BCAUADcCV8/gC8oLwTd9raPmb/vtsVSACL7DRg%2B52/YAeh9%2BwAKrqlPsEPtDOh9u8jWIvJgFOHiJh9gRdfsFKxk2GI1H9o9lqGmEbk2EU2mk8AvAZiPsABJCABqKdQsIR%2BzQDGLxFSAgAdF67X7A2mdZ5k1QvAxkIQBPsk2JaI99ngE9ovBNw3XG%2B2BF2ewxSICvACmAoFD4wsA60cXXlD0Ib1xwwxIw3H/3bt6OxuVWrQ678TyfYACp9k2PF/S4Mxvk9CEzG2d4AFkAHkXQASQAMQATQHW1vwBX8zEdd4b0AlIQP2AVDjMABWOkaL5WC7ngqolFwncGB/L4/xIzY3QgIDQKo8w%2BXdVQMIwpitlcPAqHYhRsUIZAwwgBArXY0QQw0EA50wAgVgYVskRRCBaLou1D3eb19kPDRrJsp1FU2a0v1PZMuB0pN9OIQz4WM1EzP2ABaKND0Ciz3Xsw9cSY1yQzMTy9IMozkVROzrLCmjzNtSyosc2K7U05NNkS7zfIRVKIHS71MrokKqFqmr8ucjS3P2SRSuSvzKqsprqoi8KcualzCramjOp8lKTMGhz%2BqGwL6uipyRttN4mCrAgJsMj0Ws/O1mRdJUmRdDhFloTgaN4PwOC0UhUE4cVLGsdNllWZMth4UhNpu07FgAaxAFFez5DQ%2BW%2BJkaK4b5NihyQ%2BQATn0ThJCuzReHujheAUEBAm%2BrRFjgWAkDQFhEjoOJyEoEmyfoeJ40SFFgDMDQuHh4gzD4OgQW7ShojRzEwgaR5OE%2B6JBeIR5kOibRMAcEXeBJthBGQhhF35rBoi8YA3AXbHuF4LAWEMYBxB%2B0h8CTBw8GLPXbswVRZePdZPq%2BKp%2BdoPBomIIWPCwfmCGIPAWHlxYqAMYAFAbPBMGxZCXmuz7%2BEEEQxHYCok/kJQ1H53QOYMIwUGsax9E97HIEWVBEhqPWgqoBhUCCxpHGCuuG/ttYq/SBMgrOG9gAYLwi8sLgnyCxF66CpgCADz2ne78eG6nmfNZBburyC5C182YLEQXYL67OD3oh35JXeISfA9THeJ7iYgSBvSd11eIKV57vSgpjQOmGiehr7b4g77EAfgPJQ6AP4NDwN/X%2BQUWCaiMKIWgtBu5UAnFOGob8rYCGQegKeTAMHVxbhMQODgP6eCnuTF%2BhAaBVHQPPG%2BAD758SwIkS2U9n5YF2A0chWC/5BVvowl0j9QG10nJgwyY8J5COfokLhbBua8NSP3IK1MGjJlrhMJgyA/pBRYagEEDgSAt3bpgTuPCx78KAXxJMJAsA1iCi6FC6FsKPnRl2QOWAy4QEWHYWWNQXAvlGH4LgQQXy9BKGUPQeQ0gCECZElI0SGBhP6OUdovjOiTFicEnxYiuiNCSbMFJGiRieBaHoIpUxijJIkN416axqlIw4JdUg11boY32AADRBJxJg%2BxsBCGUCifYwBkDIElL2MwvY7KmRZmzMweIIC4BnDWD68xeB41OqQVSKBUCk3JmQCgEBqZ7JAPTRmzNWbs05rQbm2M2T8zFswCW8sBaPMltLXxzzFaMAICrNWZsNZax1ogvWn1DbG1NrdC2aSbb83to7EEzzXbnTNkfb2EtfbrFujPYO%2BtQ7h0jtHWO8dnkZxTuIdOshFAqHUGbXQwT87GCHjYI%2BniK6mM4pwWuE8m54BbjfVQHcCE9wIOgPuA8mUj14UvQOK94xSunjKuewV16b2VdvMee8goHxFUfE%2BqAz4X0IGGCR/9AHAKfmA1%2BSgCDgK/j/NRC8%2BEMMsYIkBz9P6QPtcFWBAhhm6xbqgsR%2BCu4txwfUYNZiqBELwCQo0ZDHD0EoQQahtBaG8IsTeZhrCQRgI4dWbhnF03OvNcIwNNReFSLATI72ci4gKLwEolRSZCH1C0Tou%2B%2BiiB2KoMY9l88M1WMwDYut9jHGYRwnxVpbi8AePgN4qoaTnAQFcJkkJ6B8kROCVEmoq7t3pA3XMVJOSMklOyFkhdx7ugHsKSerIQShhXumOEuYNSVh1K4GdC6qMzZtM6YwDMvT%2BmDOGaM744zJn4nObM%2BZiyu3URhqsr6aN5SkABpsTYvYoa0Xht8GiLMNC0WhnyBpKNSDBy4BoQILT0acCxjjJDP0CYwEQCAfAKDKbUFkGStO0gM5Uuziiw0ehaCGjY1QJ4Lx6MiaUKoAOTBK5%2BMxqQYgQmOYqYUGJiT7BAjqftnJhT6RaOfsad%2B1pnAXSyXeFHGOda/3dMAwMswQyRljImVMqD7M8RGgTHZgDfTHPOdA%2BB9zMzPNrOQ/9QGkhxkUZomYPkMMaJMm%2BPDSQTISOmZo0p7GuMIsNLMJlu6tGGP40WHuQzkggA%3D%3D%3D) to godbolt's compiler explorer, that produces the same binary

Nice to learn something new. Quite a different ASM to Z80. ;)

So simpler C code produces 27% overhead. Whou would've thought.

TripleWhy:
Is this still PR needed? It looks like there is now an alternative implementation with the same intention in place. However there are some key differences:

- The implementation in main has a configuration option. So does this PR in the current state, but not in the original state.
- The implementation in main uses a more generic approach and *may* result in smaller image size (assuming both versions use the configuration option).
- The implementation in main is *probably* far less optimized for speed.
- The implementation in main inverts the hue, so instead of r->g->b you suddenly get r->b->g when you change the configuration option; while this PR aims to have both rainbows go in the same direction.

DedeHai:
what is wrong with using fastled HSVtoRGB? I wrote an optimized version of that, stuck in PR limbo.

TripleWhy:
Apart from the direction of H?
Nothing, but this version can skip a lot of steps because it can assume S=V=1.

DedeHai:
Then I still dont understand why adding another function that does the same is needed? Is it used in any effects where the gained speed matters? If so why not use rainbow palette (IIRC that is fewer operations)?

TripleWhy:
It's not adding a full function, just some specialized code.
But I was asking an honest question. Is this still needed/wanted or should the PR be closed?

DedeHai:
is the initial assesment still true? i.e. are the colorwheel colors bad?

TripleWhy:
In the default configuration, yes. With the rainbow color wheel option enabled, it now uses an hsv conversion with the caveats described [above](https://github.com/wled/WLED/pull/3681#issuecomment-3139517694).

TripleWhy:
> If so why not use rainbow palette (IIRC that is fewer operations)?

I forgot to address that: The rainbow color palette simply looks different. If you look at the actual values, it also isn't really an hsv sweep (or at least it wasn't when I last looked at it).

DedeHai:
I checked this and IMHO yes this is still relevant. However I think adding another version of a rainbow is not a good solution, it should just use fastled HSV to RGB and not add another option for users to choose: 0.16 changes color handling significantly and reading the comments above, the difference is minute in reality (please correct me if I am wrong).
edit:
I see that that is already implemented. In order to not add that config option to not change existing setups: simply changing
`hsv2rgb_rainbow(CHSV(pos, 255, 255), rgb);`
to
`hsv2rgb_rainbow(CHSV(255-pos, 255, 255), rgb);`
and removing the old version would render this PR obsolete. right?
edit2:
if you agree with the above @TripleWhy can you provide a new PR with that change including removing the checkmark from the config?

blazoncek:
In all honesty I was using code from this PR for several months and though I tried I never really noticed huge difference on displays I used. So I forgot about it.

However at some point @DedeHai came up with this topic and it turned out FastLED had "rainbow" color wheel for a long time. So reusing it seemed natural. I do not expect there are thousands of users that would want three variants of color wheel as the difference might be imperceptible on many LEDs. But I am not against giving users the option since the code is relatively compact and small.

If you can integrate current implementation and add your take (perhaps via dropdown "Select color wheel: Standard/Rainbow/Colorful" or radio button) I think it can be integrated.

TripleWhy:
I believe a quick recap is in order.

- There are now 3 alternative implementations of color_wheel, let's call them `original`, `rainbow` and `PR`.
- All of them try to achieve the same thing, as far as I can tell.
- Therefore, I don't think anyone asks for a configuration switch between 3 versions.

1. `original` does a linear transition from R to G to B to R. Fast.
2. `rainbow` handles the input as HSV hue and uses fastled to convert from an hsv color to an rgb color. This (looks like it) is by far the slowest version.
3. `PR` interprets the input as HSV hue and skips most of the HSV conversion steps. This is slightly slower than `original`, but (likely) magnitudes faster than `rainbow`.

At the time, this PR was not merged because of performance considerations. Instead the much slower `rainbow` was added at some point.

So much for what happened so far. So no, I don't think an option to choose between 3 versions is a good idea. Instead we should decide on one. I have no problem closing this PR. But I would also like to find the best rainbow. :)

&nbsp;

Now for some new insights. Because there are now 3 versions, I wanted to create an update of the image in the original PR description, and noticed a few issues:

rainbows3

The first (single) line is what gimp thinks a hue sweep should look like.

The other double lines are `original`, `rainbow`, `PR`, in that order. The first single ("raw") line of each double line are the raw rgb colors, that each function produces, saved as srgb png. The second single ("UI") line of each double line is a screenshot of the peek line in the wled UI. It doesn't matter if gamma correction is enabled or not, the line will look the same in the UI. Is there is some other color transformation going on?

Conclusions from this image: of the raw lines, only `PR` produces all 6 pure colors, and it is basically identical to the gimp version.
Of the UI lines, `original` looks closest to the gimp version, but it fails to produce pure yellow, cyan, magenta. `rainbow` produces all pure colors, but its center is skewed to the right, and the transition from cyan to blue looks a bit funky to me. `PR` produces all pure colors, but the mixed ones are too wide.

<details>
<summary>So I made a version that looks perfect in the UI</summary>

```c++
  const uint32_t h = (pos * 3) / 128;
  const uint32_t f = ((pos * 6) % 256) * 7 / 12;
  switch (h) {
    case 0: return RGBW32(255    , f      , 0      , w);
    case 1: return RGBW32(148 - f, 255    , 0      , w);
    case 2: return RGBW32(0      , 255    , f      , w);
    case 3: return RGBW32(0      , 148 - f, 255    , w);
    case 4: return RGBW32(f      , 0      , 255    , w);
    case 5: return RGBW32(255    , 0      , 148 - f, w);
    default: return 0;
  }
```
</details>
but on real world LEDs it looks like you would expect the code, so I'm going to assume that the peek UI is not a reliable source of information.

DedeHai:
nice comparison. I assume you read the fastled article about why they do it that way?

TripleWhy:
No, otherwise I wouldn't have wondered about it so much I assume ^^

Where is that article?

DedeHai:
https://github.com/FastLED/FastLED/wiki/FastLED-HSV-Colors

TripleWhy:
I see, thanks. I would say they fail their stated goals. The yellow is even more dull than the one in `original`. (After all, the absence of yellow in both default and rainbow palettes is what prompted me to look into this in the first place.) And the off-center cyan looks just broken to me in physical setups. But I guess that's just an opinion.

Anyway, that fastled hsv rainbow is actually exactly the rainbow palette. (or at least what it used to be? I think I read something about gamma and modified palettes recently)

So I would argue that there is no need for a `hsv2rgb_rainbow`-based `color_wheel`.

TripleWhy:
Oh btw. if this is what fastled uses for "spectrum" hsv<->rgb conversions:

fastled hsv

that also disagrees with what wikipedia says and what `PR` implements:

fastled hsv

DedeHai:
you are raising some good points.
I would suggest this:
- use rainbow palette in colorwheel as it is the one looking more natural and less technical, at least from your comparison above, assuming it does look the same as the fastled rainbow in non-dulled version
- replace the fastled hsv2rgb with the proper implementation (which I already did, see `colors.cpp` and improved in https://github.com/wled/WLED/pull/4615/)

the fastled hsv rainbow version was created for LED use without gamma correction, hence why you see it all dull when applying gamma to it, so we should now ditch that entirely.

edit:
on another note: 
I think my suggestion in the other issue to make the new gamma correction optional would be a good way to mitigate some of the issues arising with it. I will add that to the PR if I can find a way to properly do it.

TripleWhy:
Well, I want a technical looking palette, that's the whole point of this PR. ^^

For a more natural one there is always the rainbow palette. I therefore wouldn't get rid of it.

Adding options for gamma correction, and even more gamma correction, and then for where to apply it is nice, but if you need a certain set of options enabled for one palette to look good, and then another set of options enabled for a different palette to look good, it doesn't help, does it. That would only work if those options were parts of the preset instead of the configuration.

blazoncek:
@TripleWhy please start afresh with new PR. Keep both existing "rainbows" however pleasing they may look to you.
And, please, do not force-push in the future.

In the end it is only this that matters:
```c++
const uint32_t h = (pos * 3) / 128;
const uint32_t f = (pos * 6) % 256;
switch (h) {
  case  0: return RGBW32(255    , f      , 0      , w);
  case  1: return RGBW32(255 - f, 255    , 0      , w);
  case  2: return RGBW32(0      , 255    , f      , w);
  case  3: return RGBW32(0      , 255 - f, 255    , w);
  case  4: return RGBW32(f      , 0      , 255    , w);
  case  5: return RGBW32(255    , 0      , 255 - f, w);
  default: return 0;
}
```
Everything else is cosmetics.

TripleWhy:
Well force-pushing was the cleanest way to get this PR up to date, because it contained commits from yourself that were not supposed to be merged anymore.

As I said, both rainbows still exist, one in the form of an hue sweep (color_wheel), the other in the form of the rainbow palette, which is (or used to be) identical to hsv2rgb_rainbow.

I don't see a reason for a new PR, but I can of course create one if that makes you happy. I would however first like to know if we can agree on a solution, otherwise I don't need to open yet another PR.

DedeHai:
> Well, I want a technical looking palette, that's the whole point of this PR. ^^

"I was playing around with some rainbow effects and was wondering why there was no real yellow color" was your initial statement, that is what the rainbow palette provides and the technical HSV2RGB does not, at least from your images shown above. 

>Adding options for gamma correction, and even more gamma correction, and then for where to apply it is nice, but if you need a certain set of options enabled for one palette to look good, and then another set of options enabled for a different palette to look good, it doesn't help, does it. 

that is not what I was proposing. I will try to implement it.

DedeHai:
I ran some tests on this matter and here are my insights:
- the comparisons shown above are a valid representation on what it looks like on actual LEDs
- hsv2rgb_rainbow is a very close match to the rainbow palette, the differences are negligible
- the proposed hsv2rgb produces a more "technical" rainbow with evenly distributed colorbands

my code from https://github.com/wled/WLED/pull/4798 has a too aggressive hue preservation, if I fix that, the gamma correction works very well in all cases, I currently see no need for the "apply gamma after effects" option I proposed earlier.

Bottom line is: I do agree with the overall proposed changes in this PR: 
- remove the option `Use "rainbow" color wheel` as it is virtually identical to rainbow palette which can be saved in presets rather than the settings, making it more versatile
- use hsv2rgb as the new default colorwheel as it produces more evenly spread color-bands compared to "original"

I would propose one change: instead of "copying" the already existing hsv2rgb code just use the function. Its slightly slower but in practice I would not expect it to matter much as its at least as fast as `color_from_palette()`

Here is my proposed solution:

``` C++
  if (palette) return color_from_palette(pos, false, false, 0); // only wrap if "always wrap" is set
  pos = 255 - pos; // legacy color order
  uint8_t w = W(getCurrentColor(0));
  uint32_t rgb;
  hsv2rgb(CHSV32(pos, 255, 255), rgb);
  return rgb | (w << 24); // add white channel
}
```
Also corrected the comment about wrapping.

TripleWhy:
Sounds good to me, assuming that version of `hsv2rgb` does the same thing I do here... Should I update the PR or would you like add that code yourself?

However, the direction of the hue gradient should match the original color_wheel, not the rainbow color wheel IMO. I.e. remove `pos = 255 - pos;`. (Should that line be necessary to match the original color_wheel, I believe `hsv2rgb` is wrong. :)

As a side note I just noticed `CHSV32` with the 16 bit hue for the first time, and immediately questioned the call with 8 bit `pos`. I do see that there are two constructors, one for 16 bit hue and one for 8 bit hue, and I find that a questionable construct, but that's unrelated to this PR.

DedeHai:
you can update this PR if you like, since its already 95% there.

I believe its correct the way it is now. But please test it.

since when is a constructor overload a questionable construct?

TripleWhy:
Done, tested, looks fine to me.

Anything that makes a reader stop and wonder or look something up is questionable.

An overload isn't questionable on its own, but providing more than one overload for an integer type is problematic, especially if both expect different value ranges. If both expected the range 0-255 it would be fine, but there would also be no point in providing both. If we had strong type guarantees that would also be fine, but c++ casts freely and implicitly between many primitive types. You could of course provide some 8 and 16 bit range types that don't allow implicit casting, to make this safe, but that would just complicate things. Here are some examples where it's not immediately obvious or even misleading which function is called:

```cpp
CHSV32(255, s, v);

int h;
CHSV32(h, s, v);

uint8_t h;
CHSV32(2*h, s, v);

float h;
CHSV32(h * 255, s, v);
```

DedeHai:
> Done, tested, looks fine to me.

did you also test the color order against the original?

re: overload: ah, I see. what's the best alternative? a wrapper? or leave it to the caller and not provide it at all? IIRC my idea was to be 1:1 compatible to CHSV but that may have been a bad idea to start with.

TripleWhy:
> did you also test the color order against the original?

yes, it's the same order

> re: overload: ah, I see. what's the best alternative? a wrapper?

I just noticed that none of the above examples compile, precisely because they are ambiguous, but I guess you get the gist ^^
There is no great alternative. You either provide just the one constructor with more features, and leave the rest up to the user, (however that would make the code above actually compile... I suppose you could explicitly delete other overloads). Optionally you can also provide a static `CHSV32::from8Bit` (or similar). Or you provide no public constructor, just `CHSV32::from16Bit` and `CHSV32::from8Bit` for maximum consistency and clarity.

Edit: you could also provide just the 16 bit constructor and allow implicit casting from CHSV, but I don't think that would improve anything

DedeHai:
> yes, it's the same order

compared against main? I remember it being flipped. That would mean hsv2rgb is also flipped against the _rainbow version of fastled. Or I remember incorrectly.

So I think I will remove the ambiguous overload at one point. Can you therefore please provide a 16bit value so its one less thing to clean up later?

TripleWhy:
> > yes, it's the same order
> 
> compared against main? I remember it being flipped. That would mean hsv2rgb is also flipped against the _rainbow version of fastled. Or I remember incorrectly.

the rainbow color wheel was flipped, because the `pos = 255 - pos;` [line](https://github.com/wled/WLED/blob/93e011d4039d1a1e8c4a70503f2d79cf4ac44c3b/wled00/FX_fcn.cpp#L1084) that belongs to the old conversion was left unconditional before the if, instead of moving it with the old conversion into the else. That means that the original color wheel was r>g>b, and the rainbow wheel was r>b>g. The version I just pushed does not use that line and is therefore r>g>b, the same as the original wheel.

> So I think I will remove the ambiguous overload at one point. Can you therefore please provide a 16bit value so its one less thing to clean up later?

Ok ^^

DedeHai (approved):

blazoncek:
@DedeHai you are being inconsistent. In #4798 you are arguing for speed and reimplementing `color_fade()` when the code is used in a single pass over entire strip, while here you are suggesting the use of `hsv2rgb()` adding unnecessary computations for function that is called for each pixel which may be painted over several times.

If you insist on preferring speed over compact code, please be consistent. @TripleWhy 's implementation of HSV to RGB was faster as it didn't deal with S & V.

DedeHai:
I know I am being inconsistend and I gave the reason why I think its acceptable. Please read it again.
If you think its not acceptable, please give me some numbers.

PR #3682, opened 2024-01-14
Use strip.now instead of millis() in effects

blazoncek:
AKA "supersync" from WLED MM fork


MoonModules:
Hi @blazoncek , cool, you want to add supersync to upstream ? thumbsup

PR #3684, opened 2024-01-14
Fix missing header from HTTP requests

blazoncek:
Add HTTP header to requests made for JSON.

### DIFF
--- a/wled00/data/index.css
+++ b/wled00/data/index.css
@@ -920,6 +920,7 @@ select.sel-p, select.sel-pl, select.sel-ple {
 	margin: 5px 0;
 	width: 100%;
 	height: 40px;
+	padding: 0 20px 0 8px;
 }
 div.sel-p {
 	position: relative;
--- a/wled00/data/index.js
+++ b/wled00/data/index.js
@@ -1,4 +1,5 @@
 //page js
+const jsonH = {"Content-type": "application/json; charset=UTF-8"};
 var loc = false, locip, locproto = "http:";
 var isOn = false, nlA = false, isLv = false, isInfo = false, isNodes = false, syncSend = false/*, syncTglRecv = true*/;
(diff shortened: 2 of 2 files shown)


Aircoookie (approved):
Nice, good practice to always include the header even if not technically required.
Also, kudos on getting rid of `html_*.h` on GitHub!
I swear, that needlessly caused 95% of merge conflicts in PRs.

blazoncek:
> Also, kudos on getting rid of `html_*.h` on GitHub! I swear, that needlessly caused 95% of merge conflicts in PRs.

That was @WoodyLetsCode . Node is beyond my capabilities.

He earned a special status with all the PRs he made IMO.

WoodyLetsCode:
Thanks for your work! ❤️ But I have a few questions about this PR:
Why do you want to add this header for GET requests? Isn't that totally unnecessary for GET requests? Or what should the WLED server be able to do with this additional information from the header? Also in [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.5) there is nothing about using the `Content-Type` header for GET requests, so I don't understand the point why this should be good practice? The [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.5) only states that messages containing a payload body should generate a `Content-Type` header field.
I therefore only see an increase in traffic by a few bytes. Or am I seeing this wrong? 🤔

blazoncek:
I guess you know better than I. 😄

WoodyLetsCode:
So unless there is a specific reason, I would not add this header to the GET request. Same for the `charset` header. 🙂 
But that's just my opinion. Do what you think is best. 😃

blazoncek:
If that sounded sarcastic it was not. I meant it.
You are way better than me as far as that goes.

WoodyLetsCode:
All good! 😄  I didn't take your message sarcastically.

PR #3696, opened 2024-01-16
Limit number of LEDS shown in Liveview to pixel width of canvas

zanhecht:
Avoids aliasing errors on narrow screens (such as on mobile). Follow-on to discussion in #3621 , but broken out into a separate pull request.

For example, when displaying the colors from the C9 palette:
image

The current 0_15 branch code looks like the following with 1024 LEDs on a mobile screen that's 915 pixels wide:
image

Limiting it to the width of the canvas results in the following:
image

### DIFF
--- a/wled00/data/liveview.htm
+++ b/wled00/data/liveview.htm
@@ -25,10 +25,11 @@
     var ctx;
     function draw(start, skip, leds, fill) {
       c.width = d.documentElement.clientWidth;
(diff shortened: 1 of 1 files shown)


blazoncek:
Removing a part of LED strip from peek would be more annoying than seeing moire.
Not in favour of this PR.

zanhecht:
Fair enough. I'll open another PR for another method to reduce moire without cutting off the end.

PR #3702, opened 2024-01-18
Trim platformio.ini & add info to sample

blazoncek:
Reduce complexity of default `platformio.ini`  to only include release builds.
Add additional information into sample `platformio_override.ini.sample`


Moustachauve:
This is great. LGTM!

softhack007:
Great :+1 

@blazoncek I'll give you feedback on this weekend. Actually we could simplify many buildenvs by using "extends", which also ensures that all standard flags propagate to user envs. Maybe I'll enhance your PR a bit into that direction.

WoodyLetsCode:
Would it be possible to add a debug environment with faster upload speed and `WLED_DEBUG` enabled?

Moustachauve:
@WoodyLetsCode It should be made in a separate PR

blazoncek:
> Would it be possible to add a debug environment with faster upload speed and `WLED_DEBUG` enabled?

No, I want people te *request* debug build.

WoodyLetsCode:
How about creating an `extra_envs.ini` file and putting all non-release builds there? Then we add that file to `extra_configs` in `platformio.ini`. This would allow them to be used directly (without having to copy anything in `platformio_override.ini`) and we still have a clean `platformio.ini`.

platformio.ini
```diff
@@ -62,7 +31,7 @@ arduino_core_2_7_4 = espressif8266@2.6.2
 arduino_core_3_0_0 = espressif8266@3.0.0
 arduino_core_3_2_0 = espressif8266@3.2.0
 arduino_core_4_1_0 = espressif8266@4.1.0
-arduino_core_3_1_2 = espressif8266@4.2.0
+arduino_core_3_1_2 = espressif8266@4.2.1
```
WoodyLetsCode, line 34: Should we remove unused versions?

platformio.ini
WoodyLetsCode, line 1: Should we also move [this](https://github.com/Aircoookie/WLED/pull/3702/files#diff-4446afd728a4f34cbcddc306a9cb6be845d1a61c216076a295683bcc9c106724R134-R135) to sample?

```ini
# fast upload speed (remove ';' when building for development use)
; upload_speed = 921600
```

WoodyLetsCode, line 1: What about moving `esp32_idf_V4` to sample?

platformio.ini
```diff
@@ -9,39 +9,8 @@
...
-; default_envs = wemos_shield_esp32
-; default_envs = m5atom
-; default_envs = esp32_eth
-; default_envs = esp32dev_qio80
-; default_envs = esp32_eth_ota1mapp
-; default_envs = esp32s2_saola
-; default_envs = esp32c3dev
-; default_envs = lolin_s2_mini
+# CI/release binaries
+default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, esp32dev, esp32_eth, esp32dev_audioreactive, lolin_s2_mini, esp32c3dev, esp32s3dev_8MB, esp32s3dev_8MB_PSRAM_opi, esp32_wrover
```
WoodyLetsCode, line 13: `esp32_wrover` as default env?
Is it used that much?
We didn't have it as a binary for release 0.14.1? But also `lolin_s2_mini` was not in release 0.14.1.
Why do we keep them?

WoodyLetsCode (commented):
I took a look at this.
It actually looks quite good, I just have a few questions / comments. 😄

platformio.ini
blazoncek: IMO no, as you can still use those in `platformio_override.ini` if you choose so.

platformio.ini
blazoncek: Indeed I introduced it here as there were quite a few complaints in the past about crashing ESP32 chips (Pico in particular) since WROVER uses GPIO16 for PSRAM access.
It should become standard in the future for such builds.

ESP32-S2 and ESP32-S3 do not share PSRAM GPIO with anything else so are safe as is.

platformio.ini
blazoncek: It may also require updated comment. 😄 
IMO that is very dependant on user's environment and particular serial chip used.

blazoncek:
> How about creating an `extra_envs.ini` file

Never explored that because I am annoyed with unnecessary build environments present in VSC PIO sidebar. I want to keep it simple and relevant to *my* environment.

*my* not being mine particularly but each user's own environment, defined in `platformio_override.ini` which is excluded from Git and hence never updated by pulling.

WoodyLetsCode:
> Never explored that because I am annoyed with unnecessary build environments present in VSC PIO sidebar. I want to keep it simple and relevant to _my_ environment.

I assume you mean this sidebar:
grafik
But you could simply select your environment here and then you would no longer have all the build environments in the sidebar:
grafik
grafik

platformio.ini
WoodyLetsCode: Why would it need an updated comment?

I would have thought of something like that:
```ini
...
# Please visit documentation: https://docs.platformio.org/page/projectconf.html

[platformio]
default_envs = WLED_tasmota_1M

[env]
# fast upload speed (remove ';' when building for development use)
upload_speed = 921600

#----------
# SAMPLE
#----------
...
```

platformio.ini
WoodyLetsCode: IMO, if we only need them for `platformio_override.ini`, we should move them there as well.
But let's do it the way you like it better.

Moustachauve: > IMO no, as you can still use those in `platformio_override.ini` if you choose so.

Could they be added in `platformio_override.ini` instead if they are only needed there?

platformio.ini
Moustachauve: @blazoncek Does the API have any distinction between normal ESP32 and ESP32 wrover?

platformio.ini
```diff
@@ -62,7 +31,7 @@ arduino_core_2_7_4 = espressif8266@2.6.2
 arduino_core_3_0_0 = espressif8266@3.0.0
 arduino_core_3_2_0 = espressif8266@3.2.0
 arduino_core_4_1_0 = espressif8266@4.1.0
```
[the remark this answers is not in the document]
blazoncek: Fixed.

platformio.ini
WoodyLetsCode: Yes they could also be added in `platformio_override.ini`

blazoncek:
> > Never explored that because I am annoyed with unnecessary build environments present in VSC PIO sidebar. I want to keep it simple and relevant to _my_ environment.
> 
> I assume you mean this sidebar: 

yes

> But you could simply select your environment here and then you would no longer have all the build environments in the sidebar

Unfortunately I *always* need 4. ESP32, ESP32-S2, ESP32-C3 and ESP8266 with debug. So I keep them expanded.
Otherwise I've made `default_envs`which I switch when bulk building.

WoodyLetsCode:
Okay accepted! 😄 
And can we rename `platformio_override.ini.sample` to `platformio_override.sample.ini`, so that VS Code automatically shows the correct code highlighting?

Issue #2845, opened 2022-10-23
Two or more WiFi settings


Juris3D:
Would be very handy to have at least two WiFi settings for WLED.
Very inconvenient when work-in-progress device needs to be set up in several physical places (for example, home and my workplace)

chatziko:
The most flexible solution would be to accept a list of wifi configurations, each of which could be by SSID or BSSID (#2151), with a preference between them (possibly implicit preference by their order). WLED should connect to the first available one.

Another useful and relatively common use case: Many people have multiple access points, and WLED often fails to connect to the closest one. I'd like to setup:
- one configuration to a specific AP (BSSID), to be used by preference
- a fallback configuration with only SSID, to use any available AP if the main one is down.

Something like what esphome allows: https://esphome.io/components/wifi.html#connecting-to-multiple-networks

BigBill2000:
Plus one on this - I have WLED on my motorcycle so my need is a little unique. When I am at home I want to connect to my home network for updates and config etc. But when I ride with my friend who also has WLED lights, I want them to connect together and sync over WIFI. It would be nice to save both SSID's so it will be easy to connect together. They both fallback to their own AP when not at home, so I will have to change one to connect to the other. Going to impress some folks with the motorcycles side by side showing the same light show will be awesome! Thanks

JPZV:
Hi! I just added this feature in my latest PR: #3703

Please, let me know if this works to you, and don't hesitate to report me any issue that you find ;)

Juris3D:
Wow that's is great, if so! (I mean added double Wi-Fi) I will report back as soon as I try (a bit busy now, again, with Unreal Engine projects)

Joeboyc2:
Completely missed the plus button under the wifi ip address boxes lol, that's for the work on this

Aaroneisele55:
Is there the possibility to set which WiFi the device should prefer?
I have a backup WiFi and a main WiFi and want it to only use the backup one when the main is down....

blazoncek:
The order of entries determines preference.
There is a second level of preference though: signal strength.
If the signal strength of lower entry is several magnitudes stronger, it will be preferred.

Issue #1228, opened 2020-10-06
Request - Wifi AP list


abyssdj:
Just a passing thought, but it would be very handy in some cases to be able to have a list of multiple APs rather than just one, possibly with a HTP type priority, so that nodes will attempt to connect to multiple access points before falling back to WLED's own AP mode. The lights I have built are sometimes run on home wifi, sometimes on a hotspot, sometimes standalone, so it would be really nice to see this integrated some time rather than having to switch the wifi settings each time.

hell1337hound:
keeeen

Aircoookie:
I agree, this would come in handy for all lights which aren't stationary :)
Closing since we already have #852

Issue #852, opened 2020-04-16
Multiple SSID support


scargill:
WLED is clearly very good at what it does, but one thing stands out. Unlike my own early ESP8266 endeavours and, for example, Tasmota, WLED does not appear to support multiple SSIDs.  After tryng out various modes and having a GREAT time with this in my home office today, I made the mistake of attempting to demo a little board _ LED ring to my wife this evening on her PC. Immediately - a hurdle - my home office is out of range of the house, which is single storey and wide and so has it's own access points. So, my 2 minute demo went no-where - opportunity lost. Simple adding a second access point would further enhance the flexibiliy of this excellent software. I assume that setting up demos in one location and using them in another is commonplace.

chrisl8:
I would also love to be able to move my WLED device easily between wifi networks.

I'm willing to dig into the code if someone can point me in the right direction.

scargill:
> I would also love to be able to move my WLED device easily between wifi networks.
> 
> I'm willing to dig into the code if someone can point me in the right direction.

I would but I'm a WLED newby - good luck - Happy to be a guinea pig...

Aircoookie:
I agree, this could definitely be added, at least for a seond SSID.
The way it would work is to attempt connecting to the primary SSID for the first 10 seconds, then attempting the secondary, then the first again and so forth. The time spent attempting to connect could be lengthened after multiple attempts in case some networks take longer to connect (usually never the case). All the WiFi connection logic is currently in `wled.cpp`.

austwhite:
If you do this, we would still need a fall back if it can't connect to both.
Maybe 15 seconds attempt to come next, some APs are slow, and try for 3 times on each SSID then fall back to AP mode?
Just wanted to put my name in to watch this too 🙂

scargill:
> I agree, this could definitely be added, at least for a seond SSID.
> The way it would work is to attempt connecting to the primary SSID for the first 10 seconds, then attempting the secondary, then the first again and so forth. The time spent attempting to connect could be lengthened after multiple attempts in case some networks take longer to connect (usually never the case). All the WiFi connection logic is currently in `wled.cpp`.

Sorry about taking so long to come back on this, guys, do I take it we have a winner? 

[...]

PR #3703, opened 2024-01-18
Added multi WiFi saving feature

JPZV:
## TL;DR:

This PR adds the Multi-Wifi saving feature, which was requested and discussed in #2845, #2974, #852, #1228, and other issues that I may not found.

### User's side
For the user I tried to make it as simple as possible: Now the WiFi setting page will have three "new" sections:
- _Connect to a new network_: Here the user can scan for a nearby network or add it manually with the option "_Other network..._"
- _Saved networks_: Here the user can manage its saved network by changing the SSID, the Password, or deleting it.
- _Addresses settings_: Here is the old DHCP/IP and mDNS settings. Sadly, this will affect to all networks, so, for example, setting a static IP to 192.168.0.64/24 while on _networkA,_ this will be used for every other network, even if the subnet of _networkB_ starts from 192.168.1.0/24. So, DHCP is fully recommended from now on if using multiple networks.

### In back-end
Nowadays WLED saves the WiFi setting in two global vars: _clientSSID_ and _clientPass_. Both are strings (well... Char array). Now, I removed these vars and created three new ones: _clientNetsSSID_, _clientNetsPass_ and _clientSavedNets_. The first two are almost the same as the old vars, but now they're Strings Array (or Char Matrix, or Char Array Array, or whatever you want to call it), which can save up to 5 Networks in ESP32, or 3 in ESP8266 (those limits were selected arbitrarily). While _clientSavedNets_ is a Byte that stores how many networks are saved and loaded into memory.

When WLED wants to connect to a network, it first scan for every nearby networks and start looking for the nearest saved network. Then, tries to connect to it, and if it can't, then try with the next near one. Each attempt has like 5 seconds before failing. This amount of time was selected because "Last reconnect too old" message was being fired immediately with anything greater than 5 seconds and with just one network.

### Backward compatibility
Now WLED will check if the config file has a network saved with the last scheme, and if there's, then it will add it as the _first network_ and then it'll save the config file with the new scheme

### Drawbacks
Because I'm not familiar with **Improv** and **EEPROM saving** in WLED, and because of backward compatibility concerns, in both cases it'll set/load the first network. For example:

```diff
-  readStringFromEEPROM(  0, clientSSID, 32);
-  readStringFromEEPROM( 32, clientPass, 64);
-  readStringFromEEPROM( 96,      cmDNS, 32);
+  readStringFromEEPROM(  0, clientNetsSSID[0], 32);
+  readStringFromEEPROM( 32, clientNetsPass[0], 64);
+  readStringFromEEPROM( 96,             cmDNS, 32);
```

This shouldn't be a big issue for most users, except for those who uses Improv/EEPROM and wanted to save multiple WiFi networks.

Also, as mentioned before, now the IP Settings will affect to every network, even if the subnets differs com each others.

## Files Changed:

- **cfg.cpp**: Now the WiFi settings will be saved as `ssid%d` and `psk%d`, where `%d` goes from 0 to 4, instead of `ssid` and `psk` respectively
- **const.h**: Added `WLED_MAX_SAVED_NETWORKS`, `SUBPAGE_MAIN_SETTINGS`, and `SUBPAGE_WIFI_NETWORKS` constant
- **data/settings_wifi.htm**: Added three new sections: _Connect to a new network_, _Saved networks_, and _Addresses settings_. Also, changed the `loadJS()` function to letting load more than one JS.
- **data/style.css**: Added `input[type="password"]` to the same `input[type="text"]` style.
- **fcn_declare.h**: Changed the declaration of `getSettingsJS` to `void getSettingsJS(byte subPage, char* dest, byte subSettings);`, so it can load sub-settings from the same page.
- **improv.cpp**, **set.cpp**, **wled.cpp** and **wled_eeprom.cpp**: Added support for multiple WiFi settings.
- **wled.h**:
  - Replaced `clientSSID` and `clientPass` global vars with `clientNetsSSID` and `clientNetsPass` respectively.
  - Added `clientSavedNets` global var.
  - Now `WLED_WIFI_CONFIGURED` will be true if the first network is configured, or if `clientSavedNets` is greater than zero.
- **wled_server.cpp**: Now the function `serveSettingsJS` will accept the `s` argument for loading sub-settings from the same page.
  - In case `s` is zero (default behavior). It'll return the `GetV` JS function. Otherwise, it'll return the `GetW` one.
- **xml.cpp**: Added support for `SUBPAGE_WIFI_NETWORKS` sub-setting in `SUBPAGE_WIFI` page. In this case, it'll return a tuple of three where the first item is the amount of saved networks, while the second item is the maximum amount of networks that can be saved, and the last item is an array of every saved network.


blazoncek:
Thank you.
Please re-base this for 0_15 branch.

blazoncek:
Something went wrong as there are 88 files changed. I cannot review that.
Can you please follow https://github.com/Aircoookie/WLED/wiki/How-to-properly-submit-a-PR and open a new PR? Select 0_15 branch when doing so.

JPZV:
> Something went wrong as there are 88 files changed. I cannot review that. Can you please follow https://github.com/Aircoookie/WLED/wiki/How-to-properly-submit-a-PR and open a new PR? Select 0_15 branch when doing so.

Hi, yes! The rebase went extremely wrong, so I had to clone the 0_15 branch and patch all the files by hand. Right now I'm half way.

I'll open a new PR, so for now I'll close this one as it got corrupted. Sorry for that

blazoncek:
Thank you!

PR #3709, opened 2024-01-20
Implement multiple WiFi

blazoncek:
- similar/alternative to #3705
- solves #2845, #2974, #852

### DIFF
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -40,21 +40,39 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
   linked_remote[12] = '\0';
 #endif
 
-  JsonObject nw_ins_0 = nw["ins"][0];
-  getStringFromJson(clientSSID, nw_ins_0[F("ssid")], 33);
-  //int nw_ins_0_pskl = nw_ins_0[F("pskl")];
-  //The WiFi PSK is normally not contained in the regular file for security reasons.
-  //If it is present however, we will use it
-  getStringFromJson(clientPass, nw_ins_0["psk"], 65);
+  size_t n = 0;
+  JsonArray nw_ins = nw["ins"];
+  if (!nw_ins.isNull()) {
+    // as password are stored separately in wsec.json when reading configuration vector resize happens there, but for dynamic config we need to resize if necessary
+    if (nw_ins.size() > 1 && nw_ins.size() > multiWiFi.size()) multiWiFi.resize(nw_ins.size()); // resize constructs objects while resizing
+    for (JsonObject wifi : nw_ins) {
+      JsonArray ip = wifi["ip"];
+      JsonArray gw = wifi["gw"];
+      JsonArray sn = wifi["sn"];
+      char ssid[33] = "";
+      char pass[65] = "";
+      IPAddress nIP = (uint32_t)0U, nGW = (uint32_t)0U, nSN = (uint32_t)0x00FFFFFF; // little endian
+      getStringFromJson(ssid, wifi[F("ssid")], 33);
+      getStringFromJson(pass, wifi["psk"], 65); // password is not normally present but if it is, use it
+      for (size_t i = 0; i < 4; i++) {
+        CJSON(nIP[i], ip[i]);
+        CJSON(nGW[i], gw[i]);
+        CJSON(nSN[i], sn[i]);
+      }
+      if (strlen(ssid) > 0) strlcpy(multiWiFi[n].clientSSID, ssid, 33); // this will keep old SSID intact if not present in JSON
+      if (strlen(pass) > 0) strlcpy(multiWiFi[n].clientPass, pass, 65); // this will keep old password intact if not present in JSON
+      multiWiFi[n].staticIP = nIP;
+      multiWiFi[n].staticGW = nGW;
+      multiWiFi[n].staticSN = nSN;
+      if (++n >= WLED_MAX_WIFI_COUNT) break;
+    }
+  }
 
-  JsonArray nw_ins_0_ip = nw_ins_0["ip"];
-  JsonArray nw_ins_0_gw = nw_ins_0["gw"];
-  JsonArray nw_ins_0_sn = nw_ins_0["sn"];
-
-  for (byte i = 0; i < 4; i++) {
-    CJSON(staticIP[i], nw_ins_0_ip[i]);
-    CJSON(staticGateway[i], nw_ins_0_gw[i]);
-    CJSON(staticSubnet[i], nw_ins_0_sn[i]);
+  JsonArray dns = nw[F("dns")];
+  if (!dns.isNull()) {
+    for (size_t i = 0; i < 4; i++) {
+      CJSON(dnsAddress[i], dns[i]);
+    }
   }
 
   JsonObject ap = doc["ap"];
@@ -212,7 +230,7 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
   JsonObject btn_obj = hw["btn"];
   bool pull = btn_obj[F("pull")] | (!disablePullUp); // if true, pullup is enabled
   disablePullUp = !pull;
-  JsonArray hw_btn_ins = btn_obj[F("ins")];
+  JsonArray hw_btn_ins = btn_obj["ins"];
   if (!hw_btn_ins.isNull()) {
     for (uint8_t b = 0; b < WLED_MAX_BUTTONS; b++) { // deallocate existing button pins
       pinManager.deallocatePin(btnPin[b], PinOwner::Button); // does nothing if trying to deallocate a pin with PinOwner != Button
@@ -433,7 +451,7 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
   if (e131Port == DDP_DEFAULT_PORT) e131Port = E131_DEFAULT_PORT; // prevent double DDP port allocation
   CJSON(e131Multicast, if_live[F("mc")]);
 
-  JsonObject if_live_dmx = if_live[F("dmx")];
+  JsonObject if_live_dmx = if_live["dmx"];
   CJSON(e131Universe, if_live_dmx[F("uni")]);
   CJSON(e131SkipOutOfSequence, if_live_dmx[F("seqskip")]);
   CJSON(DMXAddress, if_live_dmx[F("addr")]);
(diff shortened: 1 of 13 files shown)


JPZV:
@blazoncek If I wanted to make a change to your solution, how do you like I make it? By a patch file or by a PR to multiwifi branch?

blazoncek:
> @blazoncek If I wanted to make a change to your solution, how do you like I make it? By a patch file or by a PR to multiwifi branch?

Start off by reviewing my PR.
When we agree on a change just branch off of my branch and PR back to it.

blazoncek:
> I just sent you a review btw.

I do not see any comment or review on Github. Where did you send it to?

JPZV:
> Where did you send it to?

Right here in this PR:

image

blazoncek:
> Right here in this PR:
> 

You need to finish/submit it for me to see it. I cannot see pending review.

JPZV (changes requested):

wled00/wled.cpp
```diff
@@ -915,6 +917,7 @@ void WLED::handleConnection()
     if (now - lastReconnectAttempt > ((stac) ? 300000 : 18000) && WLED_WIFI_CONFIGURED) {
       if (improvActive == 2) improvActive = 3;
       DEBUG_PRINTLN(F("Last reconnect too old."));
+      if (++selectedWiFi >= multiWiFi.size()) selectedWiFi = 0;
```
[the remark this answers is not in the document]
blazoncek: Networks are retried every 18s in current implementation (after unsuccessful attempt) with 3 possible configured networks a connection will be established within a minute. It may not be the fastest but it will still work as intended and there are no consequences.

The problem with your code is that it tries to duplicate existing connectivity approach. That should not be done as there need to be exactly one location a connection is established as many things rely on that (including notification to usermods).

If you really do not want to wait for a minute on your third network, you could adapt the scan to proactively select appropriate network (`selectedWiFi`) and submit a PR to this branch (co-authoring this PR). I am also not sure if scanning for available networks constantly is really the best approach as it must take away a few CPU cycles.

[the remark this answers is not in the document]
JPZV: You're trying to connect to a list of network following its index instead of which is nearest, which results not only the third one not being connected immediately, but also connecting to the worst network in most cases.

Let's think that I have a bracelet (which I'm making and that's why I made my PR) and I have saved both my wifi and my neighbor's. As of course my WiFi will be the first on the list, every time I turn on my bracelet, it'll try to connect to my WiFi, and then to my neighbor's if it fails. The issue comes when I go to my neighbor house, because my WiFi will still be visible and connectable, but with almost no signal, making almost not reachable by my phone.

Making things worse, by normal behavior, the ESP will not recognize it lost WiFi connection after a while, maybe not an entire minute, but here every seconds matter with your approach.

And, making this even worse, as `WLED_MAX_WIFI_COUNT` can be configurable, I can set it to 5, or to 10 if I have a board that support it, so now it's 18 * 10 = 180 seconds = 3 minutes to connect to the last WiFi network, which is unacceptable by any mean.

If you don't want to loose CPU cycles with `Scan()`, then make a WiFi list that it's being updated every time `selectedWiFi` reach 0, with that, we can assure to connect not only to a reachable network, but also to the nearest.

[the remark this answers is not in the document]
blazoncek: That's *unacceptable* for your particular case, not any case. But I hear you.

Feel free to optimise.

wled00/cfg.cpp
```diff
@@ -665,19 +683,23 @@ void serializeConfig() {
 #endif
 
   JsonArray nw_ins = nw.createNestedArray("ins");
+  for (size_t n = 0; n < multiWiFi.size(); n++) {
+    JsonObject wifi = nw_ins.createNestedObject();
+    wifi[F("ssid")] = multiWiFi[n].clientSSID;
```
JPZV, line 688: Can't save the SSIDs when configuring the networks, it'll always be the default one (for the first network) or empty (for the second and others).
The password seems to be saving well, so the issue seems to be either here or in `set.cpp`

wled00/data/settings_wifi.htm
```diff
@@ -52,13 +52,13 @@
 				}
 				scanLoops = 0;
 
-				let cs = gId("CS");
-				if (cs) {
+				let cs = d.querySelectorAll("#wifi_entries input[type=text]#CS0");
```
JPZV, line 55: The scan button should work on every network, not just on the first one

JPZV (changes requested):
I added two more issues that I found while testing your code.

Due to the first one on `set.cpp` or `cfg.cpp`, I cannot continue testing it, so maybe there're more issues.

wled00/wled.cpp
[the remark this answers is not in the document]
blazoncek: Latest commit adds prioritisation of WiFi-s according to signal strength and order of importance.

wled00/cfg.cpp
blazoncek: Fixed.

wled00/data/settings_wifi.htm
blazoncek: Fixed.

wled00/cfg.cpp
JPZV: > Fixed.

Still not working. I cannot connect to my network because the SSID will not save at all.

Please watch this [video](https://streamable.com/n8956l) about the issue. Also note, it doesn't care if the password is the right one or not, the SSID will still be the same ("Your_network" for the first one, and empty for the rest)

Tested with https://github.com/Aircoookie/WLED/pull/3709/commits/1bebf3d3d6f76a7afb2f7258425842f3af72bfef

JPZV: Confirm, it's working now. Well done!

wled00/data/settings_wifi.htm
```diff
@@ -70,37 +71,70 @@
...
+		}
+		function btnWiFi(i) {
+			gId("wifi_add").style.display = (i<maxNetworks) ? "inline":"none";
+			gId("wifi_rem").style.display = (i>1) ? "inline":"none";
+		}
+		function addWiFi(ssid="",pass="",ip=0,gw=0,sn=0x00ffffff) { // little endian
+			var i = gId("wifi_entries").childNodes.length;
+			if (i >= maxNetworks) return;
+			var b = `<div id="net${i}"><hr class="sml">
+Network name (SSID${i==0?", empty to not connect":""}):<br><input type="text" id="CS${i}" name="CS${i}" maxlength="32" value="${ssid}" ${i>1?"required":""}><br>
```
JPZV, line 120: The part of `${i>1?"required":""}` should be `${i>=1?"required":""}` because CS0 is the main WiFi, while CS1 and up are the "optional" ones.
Without that, the second WiFi's SSID can be set to empty

wled00/wled.cpp
```diff
@@ -697,7 +701,48 @@ bool WLED::initEthernet()
...
+    int rssi = -9999;
+    for (int o = 0; o < status; o++) {
+      DEBUG_PRINT(F(" WiFi available: ")); DEBUG_PRINT(WiFi.SSID(o));
+      DEBUG_PRINT(F(" RSSI: ")); DEBUG_PRINT(WiFi.RSSI(o)); DEBUG_PRINTLN(F("dB"));
+      for (unsigned n = 0; n < multiWiFi.size(); n++)
+        if (!strcmp(WiFi.SSID(o).c_str(), multiWiFi[n].clientSSID)) {
+          // find the WiFi with the strongest signal (but keep priority of entry if signal difference is not big)
+          if ((n < selectedWiFi && WiFi.RSSI(o) > rssi-10) || WiFi.RSSI(o) > rssi) {
+            rssi = WiFi.RSSI(o);
+            selectedWiFi = n;
```
JPZV, line 735: This will always select the nearest WiFi, but the code will enter in a loop if the pass is wrong for that WiFi. Even if the second-nearest network has everything right.

In other words, if the nearest WiFi fails to establish a connection, then it should fallback to the second-nearest WiFi, and then to the third and go-on.

wled00/wled.cpp
```diff
@@ -697,7 +701,48 @@ bool WLED::initEthernet()
...
+  if (status >= 0) { // status contains number of found networks
+    DEBUG_PRINT(F("WiFi scan completed: ")); DEBUG_PRINTLN(status);
+    int rssi = -9999;
+    for (int o = 0; o < status; o++) {
+      DEBUG_PRINT(F(" WiFi available: ")); DEBUG_PRINT(WiFi.SSID(o));
+      DEBUG_PRINT(F(" RSSI: ")); DEBUG_PRINT(WiFi.RSSI(o)); DEBUG_PRINTLN(F("dB"));
+      for (unsigned n = 0; n < multiWiFi.size(); n++)
+        if (!strcmp(WiFi.SSID(o).c_str(), multiWiFi[n].clientSSID)) {
+          // find the WiFi with the strongest signal (but keep priority of entry if signal difference is not big)
+          if ((n < selectedWiFi && WiFi.RSSI(o) > rssi-10) || WiFi.RSSI(o) > rssi) {
```
JPZV, line 733: By default, `WiFi.scanNetworks()` (either Async or no) will always "return" the networks already ordered from nearest to farthest. There's no need to sort them again loosing CPU cycles.

JPZV (changes requested):
I've added three more issues that I found.

One related to the webpage, and the others two to the wled.cpp

wled00/wled.cpp
blazoncek: I thought about that but decided it is best to leave AP open in such case to indicate incorrect set-up.
IMO people will use the functionality mostly when wifis are not available at the same time. But I may be wrong.

wled00/wled.cpp
blazoncek: There is no sorting involved. By nearest you are referring to strongest signal I assume.
Even though the strongest signal may be, in your opinion, the best choice I have a use case where it is not, so the order of entered WiFis in settings takes precedence *if the difference in signal strength is low (10dB)*.

wled00/wled.cpp
JPZV: IMO, we should always check every network, because sometimes AP doesn't work that fine as we would expect (as strange as it sounds, I have two cases of ESP32-S3 where I couldn't get the AP to work after a while. The WiFi/STA was working tho, just the AP).

The AP always must be the last resort for connection, it consumes a lot of power and heat.

blazoncek: In such case just move a little. ;)

Joke aside, ESP has the capacity to be configured to connect to multiple defined APs.
I may implement that at some point.
https://docs.espressif.com/projects/arduino-esp32/en/latest/api/wifi.html#wifimulti

JPZV: In your first comment I thought that by leaving the AP open you meant about leaving the AP Mode open.

Instead of implementing ESP's wifimulti (which is only available in ESP32 AFAIK), we just should check the others networks if the first one doesn't works, as I said before.

willmmiles: > Joke aside, ESP has the capacity to be configured to connect to multiple defined APs.  I may implement that at some point.

You guys just did.  I took a quick look at the internal implementation and it's [pretty close to the code in this PR.](https://github.com/espressif/arduino-esp32/blob/cceebb58f170c34a2356f0d22106dd6f0d141023/libraries/WiFi/src)/WiFiMulti.cpp#L119

It might be worth checking if it's easy to vendor WiFiMulti in to ESP8266 builds - I don't see anything glaringly obvious that would prevent it.

blazoncek:
@JPZV I think this is now complete.

JPZV:
> @JPZV I think this is now complete.

Okey! I'll test it right now and I'll give you my feedback today :)

wled00/set.cpp
```diff
@@ -19,21 +19,52 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
...
+      char gw[5] = "GW"; gw[2] = 48+n; gw[4] = 0; //GW address
+      char sn[5] = "SN"; sn[2] = 48+n; sn[4] = 0; //subnet mask
+      if (request->hasArg(cs)) {
+        if (n >= multiWiFi.size()) multiWiFi.push_back(WiFiConfig()); // expand vector by one
+        char oldSSID[33]; strcpy(oldSSID, multiWiFi[n].clientSSID);
+        char oldPass[65]; strcpy(oldPass, multiWiFi[n].clientPass);
+
+        strlcpy(multiWiFi[n].clientSSID, request->arg(cs).c_str(), 33);
+        if (strlen(oldSSID) == 0 || !strncmp(multiWiFi[n].clientSSID, oldSSID, 32)) {
+          forceReconnect = true;
```
JPZV, line 36: Seems that setting the SSID to empty while connected to the WiFi doesn't triggers this ForceReconnect.

It's a silly issue because after a restart it'll behavior as intended, but I wonder how deep would be this issue

JPZV (commented):
It looks nice. I only found a minor bug. Not sure if it needs a fix, so I'll leave it as a comment

JPZV:
@blazoncek I just checked that I wasn't mentioned in [v0.15.0-b1](https://github.com/Aircoookie/WLED/releases/tag/v0.15.0-b1) credits

blazoncek:
How would you want to be credited?

JPZV:
As a Contributor related to the Multi-Wifi feature, as I worked along with you to make it possible.

blazoncek:
Will be included in next changelog update.

jasmohan-narula:
@JPZV I thank you for helping implementing this. Even without credit, I appreciate your efforts.

mellbergsimon:
Is there a limitation on which platforms this works? Esp8266 supported? There is no info in wifi setup how to enable multi wifi. The documentation seems to be outdated in that regard also. How can we show new users how to use this feature?

blazoncek:
Screenshot 2025-01-20 at 06 45 31
It seems obvious. Same approach as adding a LED output.

mellbergsimon:
You are correct. I did not notice the + sign. Thank you for the clarification. I should have noticed!

blazoncek:
You can, if you wish, and contribute by updating KB to clarify that. See [WLED-Docs](https://github.com/Aircoookie/WLED-Docs).

PR #3715, opened 2024-01-22
Allow nodes buttons to be opened in new target or copy ip address

### LINKED ISSUE #3708: Enhancement, allow nodes buttons to be opened in new target, or copy ip address
Enhancement, Nodes buttons to behave like proper href links, So i can open in new, or copy link etc.

I see no reason why these links are coded like this.
I cannot open in new target, copy link or anything logical for an external link.

Screenshot 2024-01-20 at 1 26 46 PM

Please adopt some basic styleguide and accessibility guidlines in this library for all UI

WoodyLetsCode:
This makes nodes buttons to proper href links.
This fixes #3708.

### DIFF
--- a/wled00/data/index.css
+++ b/wled00/data/index.css
@@ -863,6 +863,15 @@ input[type=range]::-moz-range-thumb {
 	margin: 0;
 	white-space: nowrap;
 }
+a.btn {
+	display: block;
+	white-space: nowrap;
+	text-align: center;
+	padding: 8px 32px;
+	position: relative;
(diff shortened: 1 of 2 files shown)


blazoncek:
On FF for macOS the font family is different. `<button>` does not respect/inherit `<body>`'s family while `<a>` does.
The difference is miniscule though (1px).

WoodyLetsCode:
I noticed that too, but I thought this would be better because now we use the font defined in `body`.

PR #3716, opened 2024-01-23
Improved WiFi Connection (PR #3709)

JPZV:
(This PR is for #3709)

With this change, the ESP will only try to connect to the nearest saved WiFi networks instead to every saved ones.

Also, I fixed and issue where `forceReconnect` was still on `true` after reconnecting, making it an infinite loop.

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -752,20 +752,32 @@ void WLED::initConnection()
     // convert the "serverDescription" into a valid DNS hostname (alphanumeric)
     char hostname[25];
     prepareHostname(hostname);
-    int16_t nets = WiFi.scanNetworks();
-    bool netFound = false;
-    for (int wifi = 0; wifi < nets; wifi++) {
-      for (int i = 0; i < multiWiFi.size(); i++) {
-        // We check the current selected WiFi just in case it still in range
-        // and maybe it's the nearest. 
-        if (!strcmp(multiWiFi[selectedWiFi].clientSSID, WiFi.SSID(wifi).c_str())) {
-          WiFi.begin(multiWiFi[selectedWiFi].clientSSID, multiWiFi[selectedWiFi].clientPass);
-          netFound = true;
-          break;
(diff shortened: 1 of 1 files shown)


wled00/wled.cpp
```diff
@@ -752,7 +752,22 @@ void WLED::initConnection()
     // convert the "serverDescription" into a valid DNS hostname (alphanumeric)
     char hostname[25];
     prepareHostname(hostname);
-    WiFi.begin(multiWiFi[selectedWiFi].clientSSID, multiWiFi[selectedWiFi].clientPass);
+    int16_t nets = WiFi.scanNetworks();
```
blazoncek, line 755: The unfortunate problem with this call is that it is a blocking call.
While it is scanning LEDs are not updated.

JPZV: I made the scanNetwork() async in [5cd582cf](https://github.com/Aircoookie/WLED/pull/3716/commits/5cd582cfd16c9da9c6a984154d5c8fddff300dd3)

blazoncek:
Unfortunately I was working on my implementation so didn't see what you came up with afterwards.
It seems like you insist on calling `WiFi.begin()` from within scanning function which I'd like to avoid.

I also noticed that the code you provided sometimes fails to identify all available networks. Don't know why exactly but while I was working on my implementation I noticed that `WiFi.disconnect()` caused scan to fail and produce empty result.

I also added WiFi signal strength prioritisation so the WiFi with strongest signal is chosen instead of the first discovered.

JPZV:
Well, you implemented almost the same so I'll close this PR.

But I recommend you to take a look at [8afa427](https://github.com/Aircoookie/WLED/pull/3716/commits/8afa427a2400523c8cf03e6093e392bda279e915) because there's a bug in your implementation.

Right now I'm not in my computer but as soon as I connect to it I'll add a review in #3709 along with any other issue that I find

blazoncek:
I have to rethink my development strategy. In my dev branch I have the following...
image

PR #3718, opened 2024-01-24
Update GitHub Actions workflow

WoodyLetsCode:
This updates the GitHub Actions workflow and all the GitHub Actions used.

This was necessary because the old workflow was throwing warnings:
grafik

I also optimized the caching for PlatformIO. Now building is much faster:
| Before | Now |
|--------|--------|
| grafik | grafik | 
| grafik | grafik | 

I also removed some unnecessary code, like caching the python setup with the `actions/cache` action. Here we can just use the built-in cache function of the `actions/setup-python` action.
The `GITHUB_TOKEN` is also not needed. You would only need a `GITHUB_TOKEN` if you want to create releases in other repositories.

In the previous version, we had `build_output/firmware/*.bin.gz` and `build_output/firmware/*.bin` files in the artifacts. I see no reason to do this. It just combines the disadvantages of uncompressed files and the disadvantages of compressed files.

When creating a new tag, I also see no reason to have the files from `build_output/firmware` and `build_output/release`. They are the same files, just with different filenames. So now I have changed it so that we always use the release files for artifacts.

Creating a draft release with built files whenever a new tag is created remains the same.


dosipod:
@WoodyLetsCode As usual excellent work , i was wondering why keep the gz as it is only useful for esp02 and the size  does add up 
image

WoodyLetsCode:
@dosipod Thanks for the hint on ESP02! I didn't know we needed bin.gz files for ESP02.
I also found out that @softhack007 asked for this here: https://github.com/Aircoookie/WLED/issues/3390#issuecomment-1742161200

So I will change this and add bin.gz files for ESP02 to the artifacts as well.
Are bin.gz files only needed for ESP02 or also for ESP01 or ESP8266?

dosipod:
@WoodyLetsCode  No gz is only needed  for esp02 ( 2MB )  when using OTA  , ESp01 1M does not work with OTA so gz is useless . The normal esp8266 4MB OTA will work without gz

WoodyLetsCode:
So now for ESP02 we have two files in the artifact:
- WLED_0.15.0-a0_ESP02.bin
- WLED_0.15.0-a0_ESP02.bin.gz

PR #3721, opened 2024-01-25
Temporary AP

blazoncek:
Adds additional AP mode which is available only for 5min after boot.

Can be used as a (weak) security measure on devices that are intended to be used in public areas without WiFi coverage while still allowing configuration within first few minutes of startup.

At first I intended to replace original AP mode (No connection after boot) but that seemed too problematic for first time WLED users as AP would disappear and prohibit configuration if not connected to within 5 minutes.
This could be amended by checking if SSID is "Your_Network" and prohibit disengaging AP mode. In such case no new AP mode is necessary.


wled00/json.cpp
```diff
@@ -470,6 +470,19 @@ bool deserializeState(JsonObject root, byte callMode, byte presetId)
...
+  JsonObject wifi = root[F("wifi")];
+  if (!wifi.isNull()) {
+    bool apMode = getBoolVal(wifi[F("ap")], apActive);
+    if (!apActive && apMode) WLED::instance().initAP();  // start AP mode immediately
+    else if (apActive && !apMode) { // stop AP mode immediately
+      dnsServer.stop();
+      WiFi.softAPdisconnect(true);
+      apActive = false;
+    }
+    //bool restart = wifi[F("restart")] | false;
```
robhaswell, line 482: Remove commented out code.

wled00/const.h
```diff
@@ -157,6 +157,7 @@
 #define AP_BEHAVIOR_NO_CONN               1     //Open when no connection (either after boot or if connection is lost)
 #define AP_BEHAVIOR_ALWAYS                2     //Always open
 #define AP_BEHAVIOR_BUTTON_ONLY           3     //Only when button pressed for 6 sec
+#define AP_BEHAVIOR_BOOT_NO_CONN_5MIN     4     //Open AP when no connection after boot but only for 5min
```
robhaswell, line 160: I don't think the parameters of the mode should be baked into the mode's identifier. I suggest `AP_BEHAVIOR_BOOT_TEMPORARY`.

wled00/wled.cpp
```diff
@@ -918,8 +913,18 @@ void WLED::handleConnection()
       initConnection();
     }
     if (!apActive && now - lastReconnectAttempt > 12000 && (!wasConnected || apBehavior == AP_BEHAVIOR_NO_CONN)) {
-      DEBUG_PRINTLN(F("Not connected AP."));
-      initAP();
+      if (!(apBehavior == AP_BEHAVIOR_BOOT_NO_CONN_5MIN && now > 300000)) {
```
robhaswell, line 916: Ideally `30000` would be a constant, removing duplication and allowing it to be overridden.

wled00/data/settings_wifi.htm
```diff
@@ -186,10 +186,12 @@ <h3>Configure Access Point</h3>
...
 		<select name="AB">
-		<option value="0">No connection after boot</option>
-		<option value="1">Disconnected</option>
-		<option value="2">Always</option>
-		<option value="3">Never (not recommended)</option></select><br>
+			<option value="0">No connection after boot</option>
+			<option value="1">Disconnected</option>
+			<option value="2">Always</option>
+			<option value="3">Never (not recommended)</option>
+			<option value="4">No connection after boot (5min)</option>
```
robhaswell, line 193: Temporary?

wled00/json.cpp
blazoncek: It is a hint for future.

PR #3722, opened 2024-01-25
Usermod Klipper Percentage: Fix

lost-hope:
Due to the difference handling of slash vs. backslash on different OS, there was an issue with the Discord Bot


blazoncek:
I wanted to get this into 0_15 as I won't be fixing 0.14 any more.

PR #3728, opened 2024-01-27
added harmonic random palette generator

DedeHai:
an advanced random palette generator based on harmonic color theory


wled00/FX_fcn.cpp
```diff
@@ -462,6 +458,110 @@ CRGBPalette16 &Segment::currentPalette(CRGBPalette16 &targetPalette, uint8_t pal
   return targetPalette;
 }
 
+CRGBPalette16 Segment::generateRandomPalette(){
```
blazoncek, line 461: As generating palettes has no connection with segment class I'd recommend moving this function out of Segment class and make it global *color utility* function (`color.cpp`).

wled00/FX_fcn.cpp
```diff
@@ -462,6 +458,110 @@ CRGBPalette16 &Segment::currentPalette(CRGBPalette16 &targetPalette, uint8_t pal
...
+  /*  
+	generates a random palette based on color theory
+  */
+
+  CHSV palettecolors[4]; //array of colors for the new palette
+  uint8_t keepcolorposition = random8(4); //color position of current random palette to keep
+  palettecolors[keepcolorposition] = rgb2hsv_approximate(_randomPalette.entries[keepcolorposition*5]); //read one of the base colors of the current palette
```
blazoncek, line 468: This will need to be passed as a parameter.

blazoncek (changes requested):
Thank you. This seems like a great addition.

Nevertheless I'd like you to move the function out of Segment class and please check coding style (i.e. `for (...) {` add space between `for` and (. There is also speed benefit if using `int` instead of `int8_t` for counters.

And please rebase for `0_15` branch.

PR #3729, opened 2024-01-27
added function to generate random palette based on harmonic color theory

DedeHai:
looks really nice :)


DedeHai:
not sure if adding #include "FastLED.h" fcn_declare.h is the correct way

blazoncek:
> not sure if adding #include "FastLED.h" fcn_declare.h is the correct way

Better to place it into "wled.h"

wled00/colors.cpp
```diff
@@ -91,6 +91,106 @@ void setRandomColor(byte* rgb)
   colorHStoRGB(lastRandomIndex*256,255,rgb);
 }
 
+/*  
+ *generates a random palette based on color theory
+ */
+
+CRGBPalette16 generateRandomPalette(CRGBPalette16* basepalette)
```
blazoncek, line 98: You could have used a reference here to avoid ugly pointers. 😄

blazoncek (approved):
Thank you very much. Code looks good.
Will test ASAP.

wled00/colors.cpp
DedeHai: ugly to you, elegant and fast to me ;)

blazoncek: What's nicer: `structPtr->var` or `structRef.var`? 😀

wled00/colors.cpp
```diff
@@ -91,6 +91,106 @@ void setRandomColor(byte* rgb)
...
+
+  //now set the hues
+  int j=0;
+  for (int i = 0; i<4; i++) {
+    if(i==keepcolorposition) continue; //skip the base color
+    palettecolors[i].hue = harmonics[j];
+    j++;
+   }
+
+  return CRGBPalette16( palettecolors[0],
```
softhack007, line 187: Currently our palettes are assumed to be gamma-corrected.  You can use gamma8(uint8_t) or gamma32(uint32_t) to achieve that.

wled00/wled.h
```diff
@@ -177,6 +177,7 @@ using PSRAMDynamicJsonDocument = BasicJsonDocument<PSRAM_Allocator>;
 #define PSRAMDynamicJsonDocument DynamicJsonDocument
 #endif
 
+#include "FastLED.h"
 #include "const.h"
```
softhack007, line 181: To avoid inconsistent configured fastled instances, it's better to use the same fastled flags as in FX.h

https://github.com/Aircoookie/WLED/blob/2b0fad87d21e3a1f5538b8e238e92594d303da9c/wled00/FX.h#L34-L36

wled00/colors.cpp
DedeHai: oh, I did not know you can do that in C++, I usually code in C.
ok, you are right, would be better.

wled00/colors.cpp
```diff
@@ -91,6 +91,106 @@ void setRandomColor(byte* rgb)
...
+
+CRGBPalette16 generateRandomPalette(CRGBPalette16* basepalette) 
+{
+  CHSV palettecolors[4]; //array of colors for the new palette
+  uint8_t keepcolorposition = random8(4); //color position of current random palette to keep
+  palettecolors[keepcolorposition] = rgb2hsv_approximate(basepalette->entries[keepcolorposition*5]); //read one of the base colors of the current palette
+  palettecolors[keepcolorposition].hue += random8(20)-10; // +/- 10 randomness
+  //generate 4 saturation and brightness value numbers
+  //only one saturation is allowed to be below 200 creating mostly vibrant colors
+  //only one brigthness value number is allowed to below 200, creating mostly bright palettes
```
softhack007, line 106: small typo: "brigthness" -> brightness

wled00/colors.cpp
```diff
@@ -91,6 +91,106 @@ void setRandomColor(byte* rgb)
...
+    uint8_t j = random8(0, i + 1);
+    //swap [i] and [j]
+    uint8_t temp = palettecolors[i].value;
+    palettecolors[i].value = palettecolors[j].value;
+    palettecolors[j].value = temp;
+  }
+
+  //now generate three new hues based off of the hue of the chosen current color
+  uint8_t basehue = palettecolors[keepcolorposition].hue;
+  uint8_t harmonics[3]; //hues that are harmonic but still a littl random
```
softhack007, line 135: small typo: "littl" -> little

blazoncek:
Tested on ESP32 and it performs ok.
There is a slight FPS drop when using FPS >80 and several segments. Not sure if it is related to the change.

blazoncek:
@DedeHai please re-test after my change.

blazoncek:
Do not force push.

wled00/FX_fcn.cpp
```diff
@@ -77,8 +77,8 @@ uint16_t Segment::_usedSegmentData = 0U; // amount of RAM all segments use for t
 uint16_t Segment::maxWidth = DEFAULT_LED_COUNT;
 uint16_t Segment::maxHeight = 1;
 
-CRGBPalette16 Segment::_randomPalette = generateRandomPalette(&_randomPalette);
-CRGBPalette16 Segment::_newRandomPalette = generateRandomPalette(&_randomPalette);
+CRGBPalette16 Segment::_randomPalette = generateRandomPalette(_randomPalette);
```
blazoncek, line 80: Recursive initialisation is unacceptable IMO.

blazoncek (changes requested):
I've had a second look after yesterday's testing, which involved monitoring palette transitions for several minutes.

I must say that random palettes didn't look any more pleasing than the old ones. Perhaps the issue lies in randomisation of palette entries but I cannot say as I am no color theory specialist. What I would expect from this PR is to add (apart from greater complexity) a more visually pleasing result (best observed using Palette or Rainbow effects).

There are some other issues addressed separately as in-line comments.

P.S. And please follow the guidelines from Wiki on how to manage PRs. I've made some changes yesterday which were overwritten by force-pushing your updates.

wled00/colors.cpp
```diff
@@ -91,6 +91,113 @@ void setRandomColor(byte* rgb)
...
 
+/*  
+ *generates a random palette based on color theory
+ */
+
+CRGBPalette16 generateRandomPalette(CRGBPalette16 &basepalette) 
+{
+  CHSV palettecolors[4]; //array of colors for the new palette
+  uint8_t keepcolorposition = random8(4); //color position of current random palette to keep
+  palettecolors[keepcolorposition] = rgb2hsv_approximate(basepalette.entries[keepcolorposition*5]); //read one of the base colors of the current palette
```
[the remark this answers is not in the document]
DedeHai: it is. if(i==keepcolorposition) continue; //skip the base color 
see line 182

wled00/FX_fcn.cpp
DedeHai: can use any palette available. is there a global one that can be used?

wled00/FX_fcn.cpp
```diff
@@ -223,12 +223,7 @@ CRGBPalette16 IRAM_ATTR &Segment::loadPalette(CRGBPalette16 &targetPalette, uint
     case 1: {//periodically replace palette with a random one
       unsigned long timeSinceLastChange = millis() - _lastPaletteChange;
       if (timeSinceLastChange > randomPaletteChangeTime * 1000U) {
-        _randomPalette = _newRandomPalette;
-        _newRandomPalette = CRGBPalette16(
-                        CHSV(random8(), random8(160, 255), random8(128, 255)),
-                        CHSV(random8(), random8(160, 255), random8(128, 255)),
-                        CHSV(random8(), random8(160, 255), random8(128, 255)),
-                        CHSV(random8(), random8(160, 255), random8(128, 255)));
+        _newRandomPalette = generateRandomPalette(_randomPalette);
```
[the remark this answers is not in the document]
DedeHai: if not passing the base palette, the code for getting one random color needs to be in a separate function. did this for ease of use.

wled00/colors.cpp
```diff
@@ -91,6 +91,113 @@ void setRandomColor(byte* rgb)
...
+  }
+  //allow one to be lower
+  palettecolors[3].saturation = random8(80,255);
+  palettecolors[3].value = random8(50,255);
+
+  //shuffle the arrays using Fisher-Yates algorithm
+  for (int i = 3; i > 0; i--) {
+    uint8_t j = random8(0, i + 1);
+    //swap [i] and [j]
+    uint8_t temp = palettecolors[i].saturation;
```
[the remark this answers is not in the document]
DedeHai: good point. was looking for this, did not find it.

DedeHai:
> I've had a second look after yesterday's testing, which involved monitoring palette transitions for several minutes.
> 
> I must say that random palettes didn't look any more pleasing than the old ones. Perhaps the issue lies in randomisation of palette entries but I cannot say as I am no color theory specialist. What I would expect from this PR is to add (apart from greater complexity) a more visually pleasing result (best observed using Palette or Rainbow effects).
> 
> There are some other issues addressed separately as in-line comments.
> 
> P.S. And please follow the guidelines from Wiki on how to manage PRs. I've made some changes yesterday which were overwritten by force-pushing your updates.

I disagree. I did compare it to the original and it is much better. but that is taste I guess.  Not saying it could not be improved more but it is a start. maybe get some other's opinions?

blazoncek:
> I disagree. I did compare it to the original and it is much better. but that is taste I guess. Not saying it could not be improved more but it is a start. maybe get some other's opinions?

What about auric color pair palettes, complementary color palettes, tertiary color palettes, etc? Not randomised as seen in this code, but still random as in random base hue. That's what I was thinking of.

I would love to see other people feedback as well. We should ask beta testers on Discord for help.

wled00/colors.cpp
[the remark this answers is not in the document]
blazoncek: what about other swaps? what's their use then?

wled00/colors.cpp
```diff
@@ -91,6 +91,113 @@ void setRandomColor(byte* rgb)
...
+    //swap [i] and [j]
+    uint8_t temp = harmonics[i];
+    harmonics[i] = harmonics[j];
+    harmonics[j] = temp;
+  }
+
+  //now set the hues
+  int j=0;
+  for (int i = 0; i<4; i++) {
+    if(i==keepcolorposition) continue; //skip the base color
```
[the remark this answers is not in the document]
blazoncek: IDK but I guess there must be some sort of *autocorrect*.

DedeHai:
> What about auric color pair palettes, complementary color palettes, tertiary color palettes, etc?

that IS exactly what it does. I dont get what you mean...

wled00/FX_fcn.cpp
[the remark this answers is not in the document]
DedeHai: could do that, but it is less accurate, as there are only 4 positions passed, so it will take an in-between color and shift that. but maybe good enough? not a bad Idea and much simpler. I will try it.

wled00/colors.cpp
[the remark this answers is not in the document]
DedeHai: to make the palette more random as the colors are picked from the color wheel in an ordered fashion, you would never get a greenish color next to a blueish one if picking 4 colors for example. it would always look 'rainbowy', which is not what I was going for with a random generator. you can try to remove the hue shuffeling and see if it is more to your taste?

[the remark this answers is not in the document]
blazoncek: > it would always look 'rainbowy', which is not what I was going for

Oh! And my expectation was exactly the "rainbowy" type of palette. That clarifies it.
What if you'd implement a random decision about non-"rainbowy" palette? So that sometimes you'd get *smooth* palette and sometimes not?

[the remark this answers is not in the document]
DedeHai: sure, right now it would happen quite rarely. I will make it 50:50, shuffle or no shuffle and see what I get.
I was also thinking of adding a 10% chance of desaturating all colors more for a pastel color look.

wled00/FX_fcn.cpp
DedeHai: I commented this before the single color one :) still need to try that but should be fine

wled00/FX_fcn.cpp
```diff
@@ -466,6 +462,15 @@ CRGBPalette16 IRAM_ATTR &Segment::currentPalette(CRGBPalette16 &targetPalette, u
 void Segment::handleRandomPalette() {
   // just do a blend; if the palettes are identical it will just compare 48 bytes (same as _randomPalette == _newRandomPalette)
   // this will slowly blend _newRandomPalette into _randomPalette every 15ms or 8ms (depending on MIN_SHOW_DELAY)
+  // if palette transitions is enabled, blend it according to Transition Time (if longer than minimum given by service calls) 
+  if(strip.paletteFade)
+  {
+    if((millis()&0xFFFF) - _lastPaletteBlend < strip.getTransition()>>7) //assumes that 128 updates are needed to blend a palette, so shift by 7 (can be more, can be less)
```
blazoncek, line 468: Actually my measurements were 255 blends for full palette blend if colors were on the opposite ends.
I do not like this even though it adds predictability. After all do you really need *exact* time for two palettes to entirely blend? IMO not and I'd rather have CPU cycles spent elsewhere.

DedeHai: yes, 'worst case' is 255 but typical much lower but since palette blending can be anything, it is better to pick an average value. also makes it easy to use bitshifts and save some ram. 
your comment is unclear to me. this does not add predictability but adds the possibility to fade more slowly. not needed for disco style lighting but very useful for slow ambient lights.

blazoncek: This code:
```c++
  if (doShow) {
    yield();
    Segment::handleRandomPalette();
    show();
  }
```
introduced in one of recent commits ensures that palettes will be entirely blended in exactly 255 frames (which is ~6s assuming 42 FPS). Or, using your assumption of not entirely divergent palettes, in 128 frames (~3s).
If you reduce FPS you'll get even longer blend times. Assuming ambient light doesn't need high FPS.

If you think this is too fast then perhaps we need to introduce some scaling mechanism? As most users (until they become experienced) will use default transition time which is 0.7s making your code almost useless and producing almost instantaneous palette change. I do not want that.

blazoncek:
> that IS exactly what it does. I dont get what you mean...

I do not know what *randomness* ESP took yesterday, but today palettes look less "jumpy" and more pleasing.
So I have to take that back.

wled00/FX_fcn.cpp
DedeHai: you are wrong on that ;) 
the code delays it by skipping the  nblendPaletteTowardPalette() if transition time is set higher than function call intervals. if it set lower, it behaves exactly as before, so transitioning in ~128 frames. it will never speed it up.

blazoncek: "*ensures that palettes will be entirely blended in exactly 255 frames*"

A call to `show()` determines 1 frame. So the above statement is as true as it can be as each call to `Segment::handleRandomPalette()` will only blend 1/255th of intended blend.

`nblendPaletteTowardPalette()` will internally only compare 48 bytes when nothing needs to be done using `memcmp()`. This is *always* performed.

There is the other question/desire we need to clarify. What I want is a smooth blend of two random palettes which may change at user's desired time (anything between 1s and 65535s) and you want this smooth blend to happen within the *transition* time as specified in UI. These two desires have nothing in common (except for the smoothness).

You are mentioning slow, ambient illumination, blend which would also require longer period between random palette change (you wouldn't want to change random palette ever second and have transition time of 6553 seconds). So these two would need to be correlated somehow in such case.

As for CPU cycles: `loadPalette()` is called each time `color_from_palette()` is called which may be in 1000s of times *per frame*. You added two divisions, bitwise AND and subtraction to it. `Segment::handleRandomPalette()` is called once per frame.

DedeHai: you make good points. remember I requested to have a separate value in the LED config for palette transitions? you are describing exactly my thought process for that request. palette change time and transition time should not be correlated I agree, I would much rather have a separate palette transition value, it would also decouple FX transitions from color/palette transitions. 
as for CPU cycles: you are absolutely correct. but the code was really bad to begin with, I just made it a tad worse.
I see no reason why that if statement needed to be in the `loadPalette()` fuction. 
So I now moved the if statement to the `Segment::handleRandomPalette()` function, so only called once per frame. It works just fine that way. Still need to implement the other updates discussed. will push it as soon as it is done.

blazoncek: Discussion is the best way to solve problems (though it can make some as well). 😄

softhack007:
> I disagree. I did compare it to the original and it is much better. but that is taste I guess. Not saying it could not be improved more but it is a start. maybe get some other's opinions?

I found some time yesterday to try out the changes (still without gamma correction). I think the new random palettes do look nice :+1 just sometimes they seems to be a bit too dark, especially when you run at 50% or les global brightness.

Compared to the old random palettes, they are more colourful and fewer pastel colours (pale horses). However for me the also lack any "white".

As the results are so different, why not let users decide? It means we could have two random palettes in the list, one "classic" and the other one "harmonic". What do you think?

wled00/colors.cpp
```diff
@@ -91,6 +91,122 @@ void setRandomColor(byte* rgb)
...
+  }
+
+  //apply saturation & gamma correction
+  CRGB RGBpalettecolors[4];
+  for (int i = 0; i < 4; i++) {
+    if(makepastelpalette && palettecolors[i].saturation > 180) { 
+      palettecolors[i].saturation -= 160; //desaturate all four colors
+    }    
+    RGBpalettecolors[i] = (CRGB)palettecolors[i]; //convert to RGB
+    RGBpalettecolors[i] = gamma32((uint32_t)RGBpalettecolors[i]);
```
softhack007, line 190: This one is a bit tricky, because the uint32_t operator of CRGB returns _RGBA_ (A=alpha channel) and not RGBW (W=white channel). This might be a problem for sk6812.  So better to cut off the 'A', for example uint32_t(RGBpalettecolors[i]) & 0x00FFFFFF`

softhack007:
@blazoncek @DedeHai I'm wondering why this PR - which is about adding a new way of making random palettes - is also modifying the transitions code? 

I have to admit that I did not read through your long discussion... just wouldn't it be better to separate the topics, and move chances to the transition code into a separate PR?

blazoncek (approved):

wled00/FX_fcn.cpp
```diff
@@ -466,6 +456,21 @@ CRGBPalette16 IRAM_ATTR &Segment::currentPalette(CRGBPalette16 &targetPalette, u
 void Segment::handleRandomPalette() {
   // just do a blend; if the palettes are identical it will just compare 48 bytes (same as _randomPalette == _newRandomPalette)
   // this will slowly blend _newRandomPalette into _randomPalette every 15ms or 8ms (depending on MIN_SHOW_DELAY)
+  // if palette transitions is enabled, blend it according to Transition Time (if longer than minimum given by service calls)
```
softhack007, line 459: @blazoncek I was referring to this part (modifying transitions/blending)

blazoncek: Ah. This refers to "respect transition time" while blending new random palette into old/displayable random palette.

My (or original) implementation relied on periodic (with constant period) calls to `handleRandomPalette()` to do the blend, (due to bug) this happened rather quickly (and not consistent between different ESPs)

The bug has been mitigated and the blend was consistent but still dependent on FPS. This change (although I'm not very fond of it) somehow mitigates that.
Reality is that it would need a separate timer and be independently controllable.

softhack007: Thanks for explaining 👍 
Indeed transition time should - in best case - not depend on framerates, but use millis() or other timers so it always looks smooth, and same time will be needed no matter how high or low user set their "target fps".

As its "your" code, you're the best person to know if a second PR is needed or not.

softhack007: @blazoncek another thought:
We could measure "time elapsed" (in millis) since the last blend step. then calculate how many steps correspond to elapsed time (using 40 FPS as the baseline for changes). Put the blending into a loop. If steps needed > 0, perform several blends at once. 
This is similar to how audioreactive mitigates "hickups" in userloop activity.

DedeHai: @softhack007 I already added in something like that but I think it needs to be polished a little more. It is not a question of coding it but more a question of how it should behave. 
My opinion is, that the user should be given the option to set the blend speed, which I added in by tying it to 'Transition Speed' but IMHO this should be a separate value. Next question is then, does this blend speed control random palette only or is it also used for normal palette blending or will that still depend on transition time. Adding too many options is also bad for usability, it should be somehow logical which value controls what. Right now (without my changes but with @blazoncek fix for random blend) the  Transition Time (which also controls palette blending) can be set to 5000ms for example, but random palettes will still blend at default speed, which is ~150 frames. To me as a user this is quite inconsistent.

wled00/colors.cpp
DedeHai: ok, I was struggling a bit as I was looking for a function that could do this in HSV, but then decided not to mess with it and just convert to RGB but was then wondering why it is 32bit not 24bit. thx for pointing this out. will fix it.

DedeHai:
> I found some time yesterday to try out the changes (still without gamma correction => code from Sunday). I think the new random palettes do look nice 👍 just sometimes they seems to be a bit too dark, especially when you run at 50% or less global brightness.
> 
> Compared to the old random palettes, they are more colourful and fewer pastel colours (pale horses). However for me they also lack any "white".
> 
> As the results are so different, why not let users decide? It means we could have two random palettes in the list, one "classic" or simple, and the other one "harmonic". What do you think?

I tweaked it already a little, made it a tad brighter and upon request from @blazoncek added a 50% chance of hue shuffeling, giving it a more 'ranbowy' and added a 10% chance to generate a 'pastel' palette (experimental, not sure if this should be kept).
There are only so much 'color theory' I can pack in there, the rest is personal taste. You like more white, another more dark, another wants more pastel, another more rainbow etc. However, the parameters are easy to tweak but it may be hard to find a set of parameters we all can agree on :) but I will do my best and I agree with you about 'more white'. 
In the latest commit I added the 'fully random palette' back in as a function (which I needed to initialize) so it would be real easy now to have a boolean/tick in the LED config that activates 'harmonic random palette' or leaves it the way it was and some users may be used to.

wled00/FX_fcn.cpp
blazoncek: If we are discussing it then let me try to explain "Transition Time" as far as blending palettes go. "Transition Time" is the time taken to switch (smoothly without visible steps) from one palette to the other. Random palette is exactly that - a single randomly generated palette. A single palette which is constantly changing so no transition is triggered when random change happen. If you want to consider this random change to be considered as a trigger for "Transition Time" then the approach taken above works until transition time is shorter than minimum time needed for full blend of two palettes (255 iterations) this is the inconsistency I am not fond of. I am not saying it will not work or that it may impact users much, I just want for things to be logical.

User can specify the amount of time between random changes (between 1s and 65535s), we could reuse that time to determine the time needed to transition from one random palette to the other without introducing yet another parameter or rely on "Transition Time". The connection can be linear, logarithmic or some other non-linear function. I.e. if a user specifies random palette time of 30s then the actual transition could take 1/3rd of time (or 10 seconds). This correlation will ensure no conflicts arise from improper user selection. I am open to suggestions regarding this approach.

In any case I'm ok with proposed solution taking `strip.getTransition()` into account.

DedeHai: I completely agree, to me the current solution (iwithout this PR) with frame based random palette blends is inconsistent as well. If I set transitions to a slow 10s then change between palettes I get a smooth transition. If I switch to random palette, suddenly transitions are much faster. 
I just try to imagine differen use cases and what a user would expect. In my case, I use it for 2D matrix displays and ambient lighting. For the display case, a frame based 'static' change would be fine. For ambient not so much (as I mentioned before), here I would want slow transitions or at least have some control over it.
Then I see lots of people using it for quite flashy displays, mostly sound reactive stuff. I think here a fast transition would be expected if transition time is set low, even if the random change is set to say 5 minutes, so tying the random transition time to the change interval may be a bad idea. These are the scenarios I was thinking about.
There is no 'one fits all' solution I think without adding another config parameter. 
The currently proposed solution in this PR may be the best compromise. Some people may not like it, some may find it much better, most probably don't even care or notice ;)
The most consistent approach would be to adjust the random palette blend so it (approximately) fits the transition time, even if that is set lower than the current frame based transition. I could do that by adjusting the amount of blending done in 'handleRandomPalette' but keep the approach to just call that function once every frame.

blazoncek:
@DedeHai please verify if my latest changes broke anything. If not, I am merging as is.

DedeHai:
> @DedeHai please verify if my latest changes broke anything. If not, I am merging as is.

thx for the code cleanup. I see no issues, works just like before.

PR #3730, opened 2024-01-28
Using the connected strip as a timer

ale1800:
Hello! This is a usermod to use the strip as a timer with some functions based on MQTT messages. When the timer starts, the strip is filled with the color of the solid effect.
Led are turned off according to the remaining time of the timer.


blazoncek:
What's wrong with using Percent effect?

ale1800:
Hello @blazoncek! I can see that is possible to have a sort of progress bar with that effect tinkering a bit, but it doesn't look like something strictly related to a timer. I think that a timer with dedicated functions with few Mqtt messages could be way easier to use. Obviously I haven't searched on the internet all the possible way to obtain a visual timer with wled, so if you can show me something already built-in I missed I would be very happy to use it!

usermods/Bar_Timer/usermod_timer_bar.h
```diff
@@ -0,0 +1,210 @@
...
+
+          if (timer_status == RUNNING) {
+
+            helper = 1;
+            strip.trigger();
+            strip.show();
+
+            for (int y = number_of_leds - 1; y >= number_of_leds - led_turn_off; y--) {
+
+              strip.setPixelColor(y, 0, 0, 0, 0);
```
softhack007, line 94: Even the "solid" effect redraws very often  - each frame in case users have "off refresh" enabled.

If your intention is to draw on the stripe and you want to be sure that normal effects don't draw afterwards, you have two possibilities:
a/ only draw in handleOverlayDraw() - this function is called when all effects are finished
b/  add your own effect within the usermod. An example with some info how to do it: https://github.com/MoonModules/WLED/tree/mdev/usermods/usermod_v2_weather

In both cases, you don't need an explicit `strip.trigger(); strip.show();`.

platformio.ini
```diff
@@ -137,7 +137,9 @@ build_flags =
   -DWLED_USE_MY_CONFIG
   ; -D USERMOD_SENSORSTOMQTT
   #For ADS1115 sensor uncomment following
-  ; -D USERMOD_ADS1115
+  ;-D USERMOD_ADS1115
```
blazoncek, line 140: Please revert anything in `platformio.ini`

usermods/Bar_Timer/usermod_timer_bar.h
```diff
@@ -0,0 +1,210 @@
...
+          led_turn_off = 0;
+        }
+    
+        void loop (){
+
+          if (timer_status == RUNNING) {
+
+            helper = 1;
+            strip.trigger();
+            strip.show();
```
blazoncek, line 90: Never call `strip.show()` from usermod. It can mess other things.
`strip.trigger()` should be enough to force LED update on next cycle. And IMO even that is unnecessary if other best practices are followed.

usermods/Bar_Timer/usermod_timer_bar.h
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+
+        /**
+         * This function is executed before every strip.show(). This timer visualization uses the solid effect. When leds are turned off, the main
+         * WLED software try to restore the color picked from the web UI, so time by time all the turned off leds light up again and then turned off
+         * in the loop function of this usermod. So using this function we can set again the led off before the show() function
+         * 
+         */
+        void handleOverlayDraw()
```
blazoncek, line 109: As @softhack007 mentioned, this is the only place where usermod should interact with LED output.

The other option is custom effect function.

usermods/Bar_Timer/usermod_timer_bar.h
```diff
@@ -0,0 +1,210 @@
...
+     * handling of MQTT message
+     * topic only contains stripped topic (part after /wled/MAC)
+     */
+        bool onMqttMessage(char* topic, char* payload) {
+
+          // check if we received a command
+
+          if (strlen(topic) == 9 && strncmp_P(topic, PSTR("/timerbar"), 9) == 0) {
+
+            if (strcmp("STOP", payload) == 0){
```
blazoncek, line 133: Please be consistent, use `..._P()` functions and strings in PROGMEM if possible.

usermods/Bar_Timer/usermod_timer_bar.h
```diff
@@ -0,0 +1,210 @@
...
+              uint16_t add_time = atoi(payload + 1);
+
+              switch(timer_status) {
+
+                case STOPPED:
+
+                  number_of_leds = strip.getLengthTotal();
+
+                  millis_wait_time = (add_time / (float) (number_of_leds-1)) * 1000;
+                  Serial.print(millis_wait_time);
```
blazoncek, line 149: Really needed? If you wanted debug output, use `DEBUG_PRINT...()` macros.

usermods/Bar_Timer/usermod_timer_bar.h
```diff
@@ -0,0 +1,210 @@
...
+
+                  
+
+                  main = strip.getSegment(0);
```
blazoncek, line 153: Use by reference otherwise you are operating on a segment copy. Needlessly using RAM and causing other side effects.

usermods/Bar_Timer/usermod_timer_bar.h
```diff
@@ -0,0 +1,210 @@
...
+
+                  //Save current strip configuration
+                  settings[0] = effectCurrent;
+                  settings[1] = effectSpeed;
+                  settings[2] = effectPalette;
+
+                  strip.setMode(0, FX_MODE_STATIC);
```
blazoncek, line 160: Don't mix `strip` and `segment`. Use one or the other.

blazoncek (changes requested):
I've tried to provide som insightful feedback.
It may be better to rethink the approach taken IMO.

ale1800:
@softhack007 @blazoncek thank you very much for all this suggestion! I will modify all the things you pointed out

ale1800 (commented):
Hello! I changed the approach just using the handleOverlayDraw() function. Then I tried to organize better the timer and its variables. I also added gradual change to the strip color from green at the beginning to red at the end. Please let me know if the code needs other changes. Thanks in advance!

platformio.ini
blazoncek: This still needs addressing. Please remove `platformio.ini` from the PR.
It would be best to also rebase to 0_15 branch.

ale1800: @blazoncek sorry what do you mean? Do you want me to fork the 0_15 branch and modify that repo instead of the main one?

blazoncek (approved):

platformio.ini
blazoncek: I would suggest you follow [this](https://github.com/Aircoookie/WLED/wiki/How-to-properly-submit-a-PR) Wiki article and use 0_15 branch as your base.

In essence:
```
git fetch
git checkout -b 0_15 upstream/0_15 --no-track
git pull
git checkout -b strip-timer
<edit changes>
git commit -a -m "comment"
git push origin
```
And then make a new PR from your "strip-timer" branch.

If you do not mind *polluting* your repository (as you made changes in "main" branch) then just pull `upstream/0_15` and change destination branch of this PR.

PR #3731, opened 2024-01-29
Fix brightness/transition states on UDP exit (Hyperion etc)

TheHolyRoger:
Fixes: https://github.com/Aircoookie/WLED/issues/3680

Possibly fixes https://github.com/Aircoookie/WLED/issues/3700 too


blazoncek:
That's fixed (or supposed to be) in 0_15.
Please use that and/or rebase this for 0_15 branch.

TheHolyRoger:
@blazoncek you're right all working in 0_15 without this fix thanks!

PR #3732, opened 2024-01-30
UM Audioreactive: add two compiler options

wled-install:
Add two additional compiler options:
UM_AUDIOREACTIVE_ENABLE: makes usermod default enabled
UM_AUDIOREACTIVE_DYNAMICS_LIMITER_OFF: disables rise/fall limiter default


softhack007:
Actually I'd call it "dynamics limiter" instead of rise-fall-limiter. Otherwise OK for me 👍

wled-install:
Good point, renamed to UM_AUDIOREACTIVE_DYNAMICS_LIMITER_OFF

softhack007 (approved):

PR #3741, opened 2024-02-03
Switch to willmmiles fork of AsyncWebServer (0_15)

willmmiles:
As for #3740, but on 0_15.

Switch WLED to use the willmmiles fork of AsyncWebServer, with the following improvements:

- Merge some open PRs from upstream
- Fix longstanding use-after-free in AsyncWebServerRequest::_removeNotInterestingHeaders
- Move many static strings to PROGMEM, reducing RAM usage by ~1k
- Add a simple response queue, reducing peak RAM usage
   -  Note: this is not a hard limit unfortunately - memory usage is still unbounded with a SYM flood or large POST, as I have yet to find a practical solution to throttle inbound traffic.
- Some small reductions in allocator usage during operation

This should be considered an alpha-quality work in progress; I'm offering it here for early testing and feedback.  Version number is not pinned yet.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -144,7 +144,7 @@ lib_deps =
     fastled/FastLED @ 3.6.0
     IRremoteESP8266 @ 2.8.2
     makuna/NeoPixelBus @ 2.7.5
-    https://github.com/Aircoookie/ESPAsyncWebServer.git @ ~2.0.7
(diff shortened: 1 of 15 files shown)


willmmiles:
Closing this one - I think it'll be better managed in smaller updates.

PR #3752, opened 2024-02-07
2DGEQ: Getting same sized bar width on 32x32 display

adrianschroeter:
before a x=32 (n times of 16) had not equal sized bars, but first was a single pixel and later a bar had 3 pixel width. This solves it to have always 2 pixel sized bars.

I have to admit that I did not test with other pixel dimensions.


blazoncek:
It is easy to test, just resize segment.

softhack007:
> I have to admit that I did not test with other pixel dimensions.

PR  looks reasonable, however I'm currently "away from WLED" (holidays) and cannot test.

@adrianschroeter please test with segments of different width, and inform us about results. At least the following should be tested:
* width=2
* width=8
* width=15
* width=16
* width=17
* width=23
* width=24
* width=32

Also please try a few different numbers of bars (slider), in addition to the standard of 16 bars.

blazoncek:
@adrianschroeter are you willing to pursue this PR further? If so, please rebase it for 0_15 branch.

wled00/FX.cpp
```diff
@@ -7229,7 +7229,7 @@ uint16_t mode_2DGEQ(void) { // By Will Tatam. Code reduction by Ewoud Wijma.
   if ((fadeoutDelay <= 1 ) || ((SEGENV.call % fadeoutDelay) == 0)) SEGMENT.fadeToBlackBy(SEGMENT.speed);
 
   for (int x=0; x < cols; x++) {
-    uint8_t  band       = map(x, 0, cols-1, 0, NUM_BANDS - 1);
+    uint8_t  band       = map(x, 0, cols, 0, NUM_BANDS);
```
softhack007, line 7232: This is in line with the common workaround for know inaccuracies of `map` would be the exact opposite: add `1` to in_max and out_max.

However - as already stated - it would be good to test with common segment sizes before merging.
@adrianschroeter you can simply change the segment size to test your proposed modification.

softhack007 (changes requested):
While the code change is aligned to "common practice" for avoiding inaccuracies of `map`, we still need confidence by testing with several segment width and "number of bars" settings.

adrianschroeter:
Hi, I finally rebased and tested your suggested segment sizes.

Apart from width = 2 it made all sense to me. The 17, 23, 24 had different sized bars, but I think that is expect.

width=2 showed only 2 led's to me (not 2x2). However, not sure if this resolution is so useful.

softhack007 (approved):

softhack007:
@adrianschroeter thanks for the testing 👍 

I still see the PR is targeted at our 'main' branch. Please rebase it for `0_15`, then we can merge.

adrianschroeter:
0_15 merge in new request

PR #3756, opened 2024-02-10
feat: KITT usermod effect

casesolved-co-uk:
v2 usermod for replicating the effect found on KITT, the car from the 80's TV show Knightrider.
The original effect was in Red, however, the usermod also supports palettes

Has 4 controls:

- Speed
- Tail length
- Delay before start
- Dual

Similar to Sinelon but uses triangular wave not sine.

## Summary by CodeRabbit

- **New Features**
  - Introduced a dynamic lighting effect inspired by the iconic Knight Rider KITT. This effect now offers multiple customizable color palettes and refined transitions.
  
- **Documentation**
  - Added detailed setup instructions to guide users in enabling and configuring the new effect.

### DIFF
--- a/usermods/usermod_v2_KITT/README.md
+++ b/usermods/usermod_v2_KITT/README.md
@@ -1,0 +1,8 @@
+# KITT Usermod Effect
+
+v2 usermod for replicating the effect found on KITT, the car from the 80's TV show Knightrider.
+The original effect was in Red, however, the usermod also supports palettes
+
+## Include the usermod
+
+* `-D USERMOD_KITT`
--- a/usermods/usermod_v2_KITT/usermod_v2_KITT.h
+++ b/usermods/usermod_v2_KITT/usermod_v2_KITT.h
@@ -1,0 +1,69 @@
+#pragma once
+#include "wled.h"
+
+#ifndef FX_MODE_KITT
+#define FX_MODE_KITT    255  // Auto-allocate number
+#endif
+
+class KITTUsermod : public Usermod {
+  private:
+    // strings to reduce flash memory usage (used more than twice)
+    static const char _FX_MODE_KITT[];
+
+    /*
+     * Uses speed, intensity (tail), custom1 (delay) and option1 checkbox for dual mode
+     */
+    static uint16_t mode_kitt(void) {
+        if (SEGLEN == 1) {
+            SEGMENT.fill(SEGCOLOR(0));
+            return FRAMETIME;
+        }
+        if (SEGMENT.call == 0) {
+            SEGMENT.aux0 = 0;
+            SEGMENT.step = 0;
+        }
+        SEGMENT.fade_out(255 - SEGMENT.intensity); // Fade to SEGCOLOR(1) - BG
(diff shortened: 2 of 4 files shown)


blazoncek:
That seems similar to K.I.T.T (Scanner and Scanner dual) existing effects.

casesolved-co-uk:
> That seems similar to K.I.T.T (Scanner and Scanner dual) existing effects.

Yes, but there is no delay feature and this combines two effects into one

blazoncek:
> Yes, but there is no delay feature and this combines two effects into one

I managed to test the effect but apart from the delay on one side of the scan it is no different than the mentioned Scanner.
If you do want "delay" why not on both sides of the scan? I admit this might be personal preference.

I was also wondering about modulus (%) used in the effect. It is quite inefficient.

What if we incorporate the delay into original effect?

casesolved-co-uk:
In the original show it was quite common to see intermittent scans, but the scan always did a full loop not one way. Plus the effect gets quite tiring if it is continuous.

Optimisation vs maintenance. A single division operation per segment once every 23ms? How many microseconds are saved, 0.01us?

I'd have thought changing the behaviour of an existing effect would be the last thing to suggest for backward compatibility.

softhack007:
Can we close this PR in favour of #3763 ? Or is this usermod still "sufficiently different" so to keep it open?

casesolved-co-uk:
> Can we close this PR in favour of #3763 ? Or is this usermod still "sufficiently different" so to keep it open?

I was going to add support for 2D matrices but since blaz doesn't seem to like collaboration maybe he can add it to his. I don't have time.

blazoncek:
> blaz doesn't seem to like collaboration

I did not know this. Thanks for enlightening me.

mishu28nmv:
Can someone please write this effect as found when you save the presets as json file?
I tried the Scanner dual and is nothing like KITT. Maybe something is wrong with my ESP32, but please someone, just upload this as this ex. for Xmas `"115":{"on":true,"bri":125,"transition":7,"mainseg":0,"seg":[{"id":0,"start":0,"stop":150,"grp":1,"spc":0,"of":0,"on":true,"frz":false,"bri":125,"cct":127,"set":0,"col":[[255,0,0,0],[8,255,0,0],[255,255,255,0]],"fx":54,"sx":13,"ix":0,"pal":0,"c1":8,"c2":20,"c3":16,"sel":true,"rev":false,"mi":false,"o1":false,"o2":false,"o3":false,"si":0,"m12":1},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0}],"n":"Xmas Classic R/G/W","ql":"🎅"},
`

PR #3760, opened 2024-02-15
LockedJsonResponse: Release early if possible

willmmiles:
Release the json buffer lock as soon as we've finished serializing.  This should slightly reduce the number of lock collisions by releasing sooner - the response class isn't destructed until after the last packet is ack'd.

### DIFF
--- a/wled00/json.cpp
+++ b/wled00/json.cpp
@@ -1023,14 +1023,26 @@ void serializeModeNames(JsonArray arr)
 
 // Global buffer locking response helper class (to make sure lock is released when AsyncJsonResponse is destroyed)
 class LockedJsonResponse: public AsyncJsonResponse {
+  bool _holding_lock;
   public:
   // WARNING: constructor assumes requestJSONBufferLock() was successfully acquired externally/prior to constructing the instance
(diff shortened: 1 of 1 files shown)


willmmiles:
Apologies for the incorrect initial push, I grabbed the wrong version of this commit.

blazoncek:
One question: Is it really necessary to release buffer lock early?
And: What's the time difference? I wouldn't expect it to be more than a few 100us earlier.

willmmiles:
I don't have exact numbers, but it's a couple of round trip times -- it can be 10s of ms.   (Data goes out, ack comes back, fin goes out, fin-ack comes back).  If a second request comes in and the lock is contended, then we hit a minimum of 1s retry delay, which really hurts.  (This is not the same as the lock delay - a 503 response has a minimum 1s backoff delay.)  This was happening routinely during my ;ocal tests when a browser tried to send out multiple requests in parallel during initial page load.

The biggest impact of this patch is on requests that can be served in a single outbound packet buffer (eg <1k); in that case, the lock is released before the callback even exits, as the first packet is sent in the `send` call.  This eliminates the possiblity of contention altogether, dramatically improving responsiveness.

PR #3763, opened 2024-02-17
Effect: modified KITT (Scanner)

blazoncek:
- add delay
- add dual checkmark
- rename Fade rate to Trail (inverse)
- makes Scanner Dual obsolete

Alternative to #3756. Handles long strips better and uses predictable speed. Adds option to have delay on both sides of scan.


blazoncek:
@Aircoookie thoughts?

blazoncek:
@Aircoookie another, pleading, ping. 😄

PR #3777, opened 2024-02-26
Add HUB75 support

netmindz:


platformio.ini
```diff
@@ -461,3 +461,19 @@ build_flags = ${common.build_flags} ${esp32s2.build_flags} -D WLED_RELEASE_NAME=
   -D HW_PIN_MISOSPI=9
 ;  -D STATUSLED=15
 lib_deps = ${esp32s2.lib_deps}
+
+
+[env:esp32dev_hub75]
```
blazoncek, line 466: I will require this to be moved into sample file. I want platformio.ini clean and only used for default official builds.

wled00/bus_manager.cpp
```diff
@@ -605,6 +605,198 @@ void BusNetwork::cleanup() {
...
+  mxconfig.chain_length = max((u_int8_t) 1, min(bc.pins[0], (u_int8_t) 4)); // prevent bad data preventing boot due to low memory
+
+  if(mxconfig.mx_width >= 64 && (bc.pins[0] > 1)) {
+    DEBUG_PRINTF("WARNING, only single panel can be used of 64 pixel boards due to memory")
+    mxconfig.chain_length = 1;
+  }
+
+  // mxconfig.driver   = HUB75_I2S_CFG::SHIFTREG;
+
+#if defined(ARDUINO_ADAFRUIT_MATRIXPORTAL_ESP32S3) // MatrixPortal ESP32-S3
```
blazoncek, line 640: As mentioned in another comment, this should in fact be a runtime option.
I know it will/may be a lot of work/selection options, but I hate the idea that such configuration is compile-time based.

wled00/bus_manager.cpp
```diff
@@ -605,6 +605,198 @@ void BusNetwork::cleanup() {
...
+
+
+  DEBUG_PRINTF("MatrixPanel_I2S_DMA config - %ux%u length: %u\n", mxconfig.mx_width, mxconfig.mx_height, mxconfig.chain_length);
+
+  // OK, now we can create our matrix object
+  display = new MatrixPanel_I2S_DMA(mxconfig);
+
+  this->_len = (display->width() * display->height());
+
+  pinManager.allocatePin(mxconfig.gpio.r1, true, PinOwner::HUB75);
```
blazoncek, line 730: There is also an `allocateMultiplePins()` method available.

wled00/const.h
```diff
@@ -280,6 +280,9 @@
 #define TYPE_LPD8806             52
 #define TYPE_P9813               53
 #define TYPE_LPD6803             54
+
+#define TYPE_HUB75MATRIX         100 // 100 - 110
```
blazoncek, line 284: This may need revisiting as it somehow clashes with @Aircoookie 's bit oriented approach.

wled00/data/settings_leds.htm
```diff
@@ -23,6 +23,20 @@
 		function isD2P(t) { return t > 47 && t < 64; }                      // is digital 2 pin type
 		function is16b(t) { return t == 26 || t == 29 }                     // is digital 16 bit type
 		function isVir(t) { return t >= 80 && t < 96; }                     // is virtual type
+		function isHub75(t) { return t >= 100 && t <= 110; }                // is Hub75 type
+		function hideHub75() {
```
blazoncek, line 27: I would prefer the other way around. Instead of hiding Hub75 types, insert them into SELECT.
It will require  more effort in xml.cpp and some modified code here but should allow for enhancements in the future.

I will tackle this problem and push modifications when I'm done and the solution is viable. No need for you to do anything.

blazoncek (changes requested):
Thanks, this already looks great, though I have some remarks.

The most important:
- no modifications in platformio.ini
- if possible avoid compile time differentiation

Unfortunately I do not own Hub75 panels but I'll try to work on some enhacnements using logic.

wled00/bus_manager.cpp
netmindz: I'm a bit surprised by this comment as I messaged you during the development and specifically discussed the challenges relating to the usability of configuration when so many pins are needed and that they are non configurable if you are using one of the ready made boards such as the Adafruit Matrix Portal S3

wled00/const.h
netmindz: Can you explain this comment further?

wled00/bus_manager.cpp
blazoncek: I'm sorry if I forgot, I interact with several (possibly in range of 50) users and developers at any time.
So things may slip my mind and I do not heave the time to revisit all communication made. I apologise.

Nevertheless users may build clones or try their own PCBs following HW layout of the existing boards. So it would be more natural to select any Hub75 compatible board instead of compile-time choice. _Similar as it is with Ethernet_.

After all, the only difference I can see is in GPIO allocation.

wled00/const.h
blazoncek: I tried to hide the fact that this was a bitmasked value in 0.15 macros (check marcos in 0.14 or previous). But I would still like it to follow the original idea of particular bit defining LED type if possible.
There is no need for you to do anything at this point. It was more of a reminder for myself.

wled00/bus_manager.cpp
netmindz: The ideal solution would be a total refactor whereby the front-end requests what config the bus requires, rather than the current mess of having a large number of elements that you then need to work out what to hide for that bus type. Then you can remove all the isType calls.

But this is of course a big change and should be a PR in it's own right

blazoncek:
> Also, if you have PSRAM (and what new ESP32 doesn't these days?) are you really limited to one panel?

PSRAM is slow. Too slow for anything LED related.

wled00/bus_manager.cpp
blazoncek: I agree that it is a mess. Still, a little better than it was.

netmindz:
> Wow! I'm shocked at how cleanly that slots in. Nice.
Thanks :)

> I have both a Matrix S3 and a Mesmerizer (nightdriverled.com) so I may try spinning a build. What kind of frame rates are you seeing on the 2D animations? (We have a lot of the same effects in NightDriver and I've flirted with using the Faptastic HUB75 layer, too...)

I'm having real issues with wifi with the Matrix S3, so will be interesting to see if you see the same.
 
> Why are ESP32_FORUM_PINOUT, ARDUINO_ADAFRUIT_MATRIXPORTAL_ESP32S3, etc. compile-time flags (and not defined in a derivation of esp32dev_hub75) and not runtme-selectable flags like APA102 vs. WS2814 vs WS2812, etc.?

If you are using the matrix portal, you can't change the pins, so no point asking the user to enter 14 pins. For other boards, just lazy ;)

> Given how differently -S3 and -Nothing have to be built, can you really rely on board = esp32dev? Or is this a "draw the rest of the owl" that needs to have another custom_platform_mps3.ini written that derives from an appropriate S3 with the 8MB flash, 2MB PSRAM combination? (https://github.com/platformio/platform-espressif32/blob/master/boards/adafruit_matrixportal_esp32s3.json ?) Do you need help with that, perhaps?

The example env just shows what options you need to enable this feature. If you aren't using a regular esp32 then you need to add the extra options to the env for your board

> Would this be likely to plan nicely with the Virtual Display (chained panels - limited only by color depth and refresh rates) and UDP/DDP inputs and such?

In theory yes you could use DDP then Hub75 output or mix of Hub75 and other output types
 
> You can see I'm pretty excited to see this as I've traveled a lot of similar paths. I'm a bit engaged in the short term, but may be able to help land this if you need extra eyes and fingers on it.

Great

netmindz:
> Also, if you have PSRAM (and what new ESP32 doesn't these days?) are you
really limited to one panel? 

You are not limited to single panels. At the moment it will allow you to select up to 4 panels of 32x32 in horizontal configuration. We can add other layouts later

> The ESP32-S3s always have octal PSRAM  and can
DMA to and from it

The library being used does allow PSRAM use

platformio.ini
netmindz: Done

wled00/bus_manager.cpp
netmindz: It's a struct for the gpio pins, not array so can't really swap over easily

netmindz:
The simple option @robertlipe would just be to use this ready-made bin

https://github.com/netmindz/WLED/actions/runs/8056668690/artifacts/1276931005

netmindz:
If you are trying to build from source, you need to run

`npm install`

netmindz:
@robertlipe  - I suspect the wrong version of the platform is being used. We have the option to build on V4, which is what I use for the https://github.com/MoonModules/WLED/pull/108

lboue:
Thanks @netmindz. I also have Adafruit MatrixPortal S3 board. I will try this.

lboue:
> The simple option @robertlipe would just be to use this ready-made bin
> 
> https://github.com/netmindz/WLED/actions/runs/8056668690/artifacts/1276931005
Should I use the following command?

`esptool.py write_flash 0x10000 adafruit_matrixportal_esp32s3.bin`

netmindz:
> > The simple option @robertlipe would just be to use this ready-made bin
> > 
> > https://github.com/netmindz/WLED/actions/runs/8056668690/artifacts/1276931005
> 
> Should I use the following command to flash Adafruit MatrixPortal S3 board?
> 
> `esptool.py write_flash 0x10000 adafruit_matrixportal_esp32s3.bin`

Not sure about that, I just tend to goto the manual ota page of an existing WLED install and upload the bin

lboue:
> Not sure about that, I just tend to goto the manual ota page of an existing WLED install and upload the bin

Which firmware file did you use for WLED install  before going to the manual ota page and upload your HUB75 version?
https://github.com/Aircoookie/WLED/releases/tag/v0.14.1

- WLED_0.14.1_ESP32-S3_16MB_PSRAM_opi.bin
- WLED_0.14.1_ESP32-S3_8MB.bin
- WLED_0.14.1_ESP32-S3_8MB_PSRAM_opi.bin
- WLED_0.14.1_ESP32-S3_8MB_PSRAM_qspi.bin

netmindz:
> > Not sure about that, I just tend to goto the manual ota page of an existing WLED install and upload the bin
> 
> Which firmware file did you use for WLED install before going to the manual ota page and upload your HUB75 version? https://github.com/Aircoookie/WLED/releases/tag/v0.14.1
> 
> * WLED_0.14.1_ESP32-S3_16MB_PSRAM_opi.bin
> * WLED_0.14.1_ESP32-S3_8MB.bin
> * WLED_0.14.1_ESP32-S3_8MB_PSRAM_opi.bin
> * WLED_0.14.1_ESP32-S3_8MB_PSRAM_qspi.bin

Ah, for the matrix3 I didn't actually try a regular install, just went directly to building from source. Does the web installer give a basic working setup? https://install.wled.me/

lboue:
@netmindz 
Any news about this PR?

netmindz:
There is support for hub75 in the latest MoonModules builds, just use V4_S or V4_M builds

As for support in AirCookie, @blazoncek said he would be updating the PR with runtime vs compile time pin configuration 
https://github.com/Aircoookie/WLED/pull/3777#discussion_r1503852407

blazoncek:
@netmindz I've tried to come up with a better solution regarding conditional UI elements for setting options for Hub panels.
Unfortunately I was not successful as that would require redesign of whole LED settings page and bus manager logic together with available LED types. I tried that but ended efforts as there was too little gain for too much work in the end.

I can see 2 ways forward as a result of this: a) make Hub 75 panels regular option and not a conditional compile or b) not support Hub 75 at all

For a) that would probably increase code size which most users would rather see used by other usermods instead (current flash utilisation is already above 95% while we are still with ESP32 core 3.5.0, newer cores add substantial amount to this).

There is of course still possibility to keep PR as is and hide Hub panels in 99.9% of installations at the expense of few wasted bytes for never used UI elements. I cannot decide either way so I'll leave the decision to @Aircoookie. I have marked the PR with `keep` label so it will not become stale.

For anyone else interested please upvote this PR so we can see how much interest is there and decide upon it.

netmindz:
Swapping to add the options to the drop-down rather than removing is the smaller problem to solve. 

As we have discussed, the bigger issue is that of the pin configuration. 

The "proper" fix would be for the UI to query the bus manager to ask what the config choices should be. This would both remove the need for the UI to selectively hide/rename the config choices and also mean that adding new options like say which predefined combination of pins you are using for hub75 could come from the specific bus type class

blazoncek:
> The "proper" fix would be for the UI to query the bus manager to ask what the config choices should be. This would both remove the need for the UI to selectively hide/rename the config choices and also mean that adding new options like say which predefined combination of pins you are using for hub75 could come from the specific bus type class

Yes, that's what I was referring to. There is no such way ATM (to query) and hence needs hardcoded info in HTML files.

The other problem is also the increase of binary files themselves as mentioned which must not be underestimated as there are several enhancements in the pipeline that require quite a few more bytes.

netmindz:
I have started the updating of the LED preferences to be build from the BusManager and it's enabled Bus @blazoncek 

https://github.com/Aircoookie/WLED/pull/4056

netmindz:
You  can see the "clean" version of the PR, without the changes in these two PRs designed to help simplify adding new output types such as this https://github.com/Aircoookie/WLED/pull/4056 and https://github.com/Aircoookie/WLED/pull/4058

at this link - https://github.com/netmindz/WLED/compare/LED-Prefs-BusManager-config...netmindz:WLED:HUB75-AC

Are you happy that it's starting to look like a more suitable set of changes @blazoncek ?

blazoncek:
Thank you @netmindz ! When I can, I'll join the effort.

> Are you happy that it's starting to look like a more suitable set of changes @blazoncek ? Note how it requires no change to the settings_leds.htm at the moment.

Happiness has nothing to do with this but yes, I am happy. 😄 

So what is the main difference between #4056 and #4058? What would be the needed order of merges?

netmindz:
> Thank you @netmindz ! When I can, I'll join the effort.

Great 😃 

> > Are you happy that it's starting to look like a more suitable set of changes @blazoncek ? Note how it requires no change to the settings_leds.htm at the moment.
> 
> Happiness has nothing to do with this but yes, I am happy. 😄

Great 😃 
> 
> So what is the main difference between #4056 and #4058? What would be the needed order of merges?

Rather than one mega PR that does a whole load of changes, I thought would be easier to do the very low risk #4056 first, get that merged in. I can then be working on #4058 alongside this PR so we have the dynamic config we need to support all of the options we would want to see for HUB75, which would ultimately include drop-downs for things like the layout of multiple panels

netmindz:
I've temporarily changed the target of the merge to `bus-config`  so we can see just my HUB75 changes, not the bus-config changes as well @blazoncek

wled00/const.h
netmindz: Once we have the proper UI changes to support non-standard config options then we should only need one maybe two IDs to handle all the different panel setups. At this point, we can swap to the proper ID(s) rather than my placeholder of 101 -> 110

blazoncek: Network types are 80-95 and 2 pin digital 48-63, so IMO Hub75 types can span 64-71 (8 types) or 64-79 (16 types). Would that be ok?
I would not consider Hub75 to be a *virtual* type as it needs GPIO pins.
And if you want additional *virtual* types those can go all the way to 126. I would reserve 127 for usermod type as discussed elsewhere. The usermod type *can* reserve GPIO pins as usermod instead of as a bus.

blazoncek:
@netmindz one thing popped up while I looked at the code (which now looks much nicer!!). How do you intend to honour ABL since you removed `count` field in UI?

netmindz:
To be honest I wasn't planning on being part of ABL.

Firstly the panels you connect might have any number of possibly LED types due to the huge variety of the number of panels out there in the market. 

The second is that HUB75 works more like a CRT monitor where you are refreshing the pixels by scanning over them at high speed and relying on the persistence of vision - you never actually light up all LEDs at the same time, only two or four of the rows.

Therefore the calculation of the current draw is not possible to calculate on the same basis as addressable LEDs, any calculation used would need to be based on the width of the matrix and then half scan or quarter scan panel type

blazoncek:
Good. Then it just needs to be mentioned somewhere.

blazoncek (approved):

wled00/data/settings_leds.htm
```diff
@@ -339,17 +342,25 @@
...
-							return; // do not check conflicts
 					} else {
 						LC.max = d.max_gpio;
 						LC.min = -1;
 					}
 				}
+				if (isHub75(t) && (nm=="L0" || nm=="L1")) {
+					// Matrix width and height
+					LC.max = 128;
+					LC.min = 16;
```
blazoncek, line 353: you can add `LC.step=16;` to limit entries to multiples of 16.
Be sure to clear step otherwise.

netmindz: Actual values supported are 16,32,64,128

This an example of where I want the bus config to be able to supply a enum

blazoncek: Let's expedite @willmmiles "End oappend" and then this will be easier to implement in JS and backend code.

netmindz: Yeah I thought that would help. I've also been playing with a few ideas as how to reduce repetition but also allow per ledType customisations 

E.g we don't want to have to pass "data pin" for every led and "clock pin" for every one that features clock - we already have the cover with the D and 2P type.

For the exception, like for Hub75 we can supply the optional "overrides" for any labels, input fields active, input types etc

blazoncek: We could do something similar as it is done for usermods: `addDD()`.

wled00/bus_manager.h
```diff
@@ -344,7 +392,7 @@ struct BusConfig {
   {
     refreshReq = (bool) GET_BIT(busType,7);
     type = busType & 0x7F;  // bit 7 may be/is hacked to include refresh info (1=refresh in off state, 0=no refresh)
-    size_t nPins = Bus::getNumberOfPins(type);
+    size_t nPins = OUTPUT_MAX_PINS;
```
netmindz, line 395: We are using the "pins" in the config file for more than just GPIO, so always copy the full array

netmindz:
There are many more config options that the library supports that are not (yet) exposed through the WLED UI, but my feeling is to merge what is there, then have a later follow up PR(s) that tackle specific items, like non-horizontal layout of multiple panels

netmindz:
... and a label indicating which pinout configuration is active. But that's blocked by the pr that deals with the extra ledType data from the bus

blazoncek (commented):
There are a few minor issues but otherwise seems ok.
Will need to take more time to go through code more thoroughly.

wled00/bus_manager.cpp
```diff
@@ -784,6 +813,293 @@ void BusNetwork::cleanup() {
...
+  if(this->_len >= MAX_LEDS) {
+    DEBUG_PRINTLN("MatrixPanel_I2S_DMA Too many LEDS - playing safe");
+    return;
+  }
+
+  DEBUG_PRINTLN("MatrixPanel_I2S_DMA created");
+  // let's adjust default brightness
+  display->setBrightness8(25);    // range is 0-255, 0 - 0%, 255 - 100%
+
+  delay(24); // experimental
```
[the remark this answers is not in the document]
netmindz: @softhack007 ?

wled00/bus_manager.cpp
```diff
@@ -784,6 +813,293 @@ void BusNetwork::cleanup() {
...
+  if (_ledBuffer != nullptr) DEBUG_PRINTLN(F("MatrixPanel_I2S_DMA LEDS buffer enabled."));
+  if (_ledsDirty != nullptr) DEBUG_PRINTLN(F("MatrixPanel_I2S_DMA LEDS dirty bit optimization enabled."));
+  if ((_ledBuffer != nullptr) || (_ledsDirty != nullptr)) {
+    DEBUG_PRINT(F("MatrixPanel_I2S_DMA LEDS buffer uses "));
+    DEBUG_PRINT((_ledBuffer? _len*sizeof(CRGB) :0) + (_ledsDirty? getBitArrayBytes(_len) :0));
+    DEBUG_PRINTLN(F(" bytes."));
+  }
+}
+
+void __attribute__((hot)) BusHub75Matrix::setPixelColor(uint16_t pix, uint32_t c) {
```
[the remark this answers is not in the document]
netmindz: @softhack007

wled00/bus_manager.h
```diff
@@ -1,6 +1,13 @@
 #ifndef BusManager_h
 #define BusManager_h
 
+#ifdef WLED_ENABLE_HUB75MATRIX
+
+#include <ESP32-HUB75-MatrixPanel-I2S-DMA.h>
+#include <ESP32-VirtualMatrixPanel-I2S-DMA.h>
+#include <FastLED.h>
```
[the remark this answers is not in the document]
netmindz: Was added for one of the functions or possibly CRGB. I'll check if it's still needed

wled00/bus_manager.cpp
```diff
@@ -784,6 +813,293 @@ void BusNetwork::cleanup() {
...
+  // mxconfig.driver = HUB75_I2S_CFG::ICN2038S;  // experimental - use specific shift register driver
+  //mxconfig.latch_blanking = 3;
+  // mxconfig.i2sspeed = HUB75_I2S_CFG::HZ_10M;  // experimental - 5MHZ should be enugh, but colours looks slightly better at 10MHz
+  //mxconfig.min_refresh_rate = 90;
+  //mxconfig.min_refresh_rate = 120;
+  mxconfig.clkphase = bc.reversed;
+
+  fourScanPanel = nullptr;
+
+  if(bc.type == TYPE_HUB75MATRIX_HS) {
```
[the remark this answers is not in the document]
netmindz: updated

wled00/bus_manager.h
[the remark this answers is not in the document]
netmindz: Yes it's still required - see the usage of CRGB

wled00/bus_manager.h
```diff
@@ -313,6 +322,39 @@ class BusNetwork : public Bus {
     bool      _broadcastLock;
 };
 
+#ifdef WLED_ENABLE_HUB75MATRIX
+class BusHub75Matrix : public Bus {
+  public:
+    BusHub75Matrix(BusConfig &bc);
+    bool hasRGB() { return true; }
```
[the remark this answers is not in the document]
netmindz: You can't actually override as inline, bu that suggestion helped find and fix that bug

wled00/bus_manager.h
netmindz: I have checked the places the array is created and replaced the magic number 5 with the proper constant

wled00/bus_manager.cpp
```diff
@@ -784,6 +813,293 @@ void BusNetwork::cleanup() {
...
+      int y = pix / _panelWidth;
+      display->drawPixelRGB888(int16_t(x), int16_t(y), r, g, b);
+    }
+  }
+}
+
+uint32_t BusHub75Matrix::getPixelColor(uint16_t pix) const {
+  if (!_valid || pix >= _len) return IS_BLACK;
+  if (_ledBuffer)
+    return uint32_t(_ledBuffer[pix].scale8(_bri)) & 0x00FFFFFF;  // scale8() is needed to mimic NeoPixelBus, which returns scaled-down colours
```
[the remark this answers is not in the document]
netmindz: @softhack007

wled00/bus_manager.h
netmindz: If I don't change, then we *would* see an out of bounds as the number of GPIO pins used is 16 but there are only 3 "pins" used in the config

wled00/bus_manager.cpp
```diff
@@ -776,6 +805,334 @@ void BusNetwork::cleanup() {
...
+  _valid = false;
+  _panelWidth = 0;
+  deallocatePins();
+  DEBUG_PRINTLN("HUB75 output ended.");
+
+  //if (virtualDisp != nullptr) delete virtualDisp;  // warning: deleting object of polymorphic class type 'VirtualMatrixPanel' which has non-virtual destructor might cause undefined behavior
+  delete display;
+  display = nullptr;
+  virtualDisp = nullptr;
+  if (_ledBuffer != nullptr) free(_ledBuffer); _ledBuffer = nullptr;
```
blazoncek, line 1110: Missing braces.

wled00/bus_manager.cpp
```diff
@@ -776,6 +805,334 @@ void BusNetwork::cleanup() {
...
+    DEBUG_PRINTF("MatrixPanel_I2S_DMA = unsupported color order %u\n", bc.colorOrder);
+  }
+
+  DEBUG_PRINTF("MatrixPanel_I2S_DMA config - %ux%u length: %u\n", mxconfig.mx_width, mxconfig.mx_height, mxconfig.chain_length);
+  DEBUG_PRINTF("R1_PIN=%u, G1_PIN=%u, B1_PIN=%u, R2_PIN=%u, G2_PIN=%u, B2_PIN=%u, A_PIN=%u, B_PIN=%u, C_PIN=%u, D_PIN=%u, E_PIN=%u, LAT_PIN=%u, OE_PIN=%u, CLK_PIN=%u\n",
+                mxconfig.gpio.r1, mxconfig.gpio.g1, mxconfig.gpio.b1, mxconfig.gpio.r2, mxconfig.gpio.g2, mxconfig.gpio.b2,
+                mxconfig.gpio.a, mxconfig.gpio.b, mxconfig.gpio.c, mxconfig.gpio.d, mxconfig.gpio.e, mxconfig.gpio.lat, mxconfig.gpio.oe, mxconfig.gpio.clk);
+
+  // OK, now we can create our matrix object
+  display = new MatrixPanel_I2S_DMA(mxconfig);
```
blazoncek, line 941: (std::nothrow)

wled00/bus_manager.h
```diff
@@ -315,6 +324,37 @@ class BusNetwork : public Bus {
     bool      _broadcastLock;
 };
 
+#ifdef WLED_ENABLE_HUB75MATRIX
+class BusHub75Matrix : public Bus {
+  public:
+    BusHub75Matrix(BusConfig &bc);
+    void setPixelColor(unsigned pix, uint32_t c) override;
```
blazoncek, line 331: This is the place for hot attribute as the compiler will know to optimise everywhere.

netmindz:
@coderabbitai review

netmindz:
Hi @robertlipe 

Yeah trying to get into upstream WLED has stalled.

It is however well supported in WLED-MM which is actually way better suited to using HUB75 with as that supports significantly higher pixel counts and improved framerates.

@softhack007 has been very busy recently, but at some stage he hopefully will get some time to address these comments so that users of regular WLED can at least experiment with say single 64x64 panel before swapping to WLED-MM for bigger setups

wled00/bus_manager.h
netmindz: Added

spiro-c:
I was just testing this PR and i was trying to figure out if i can use two 64x32 panels but look like when ever i set the chain length to 2 the esp is going in the boot loop 
<details>
<summary>Debug log</summary>
<br>

```
---DEBUG INFO---
Runtime: 30000
Unix time: 30,001
Free heap: 211640
TX power: 34/34
Wifi state: 3
ESP-NOW state: 0
State time: 0
NTP last sync: 999000000
Client IP: 192.168.10.177
Loops/sec: 1471
Loop time[ms]: 0/664
UM time[ms]: 0/3
Strip time[ms]:0/32
Segments: 1 -> 196/0B
  Seg: 30,1 [A=1, 2D=0, RGB=1, W=0, CCT=0]
Modes: 4*218=872B
Data: 4*218=872B
Map: 2*0=0B
Web server status: WS client disconnected.
settings resp 0
Loop delayed more than 8ms.
settings resp 2
JSON buffer locked. (6)
JSON buffer released. (6)
Bus: Config (0-30, type:65, CO:0, rev:0, skip:0, AW:0 kHz:0, mA:55/0)
# of buses: 2
PIN ALLOC: Pin 0 successfully allocated by 0x85.
PIN ALLOC: 0x0000000000000lX.
Loop delayed more than 17ms.
Re-init busses.
Removing all.
Digital Cleanup.
Image playback end called
Removing all.
Maximum LEDs on a bus: 0
Digital buses: 0
Bus: Adding bus (p:0 v:0)
Huidu HD-WF2 ESP32-S3 pinout
PIN ALLOC: Pin 2 allocated by 0x8E.
PIN ALLOC: Pin 6 allocated by 0x8E.
PIN ALLOC: Pin 10 allocated by 0x8E.
PIN ALLOC: Pin 3 allocated by 0x8E.
PIN ALLOC: Pin 7 allocated by 0x8E.
PIN ALLOC: Pin 11 allocated by 0x8E.
PIN ALLOC: Pin 39 allocated by 0x8E.
PIN ALLOC: Pin 38 allocated by 0x8E.
PIN ALLOC: Pin 37 allocated by 0x8E.
PIN ALLOC: Pin 36 allocated by 0x8E.
PIN ALLOC: Pin 21 allocated by 0x8E.
PIN ALLOC: Pin 33 allocated by 0x8E.
PIN ALLOC: Pin 35 allocated by 0x8E.
PIN ALLOC: Pin 34 allocated by 0x8E.
PIN ALLOC: 0x0000000000000lX.
MatrixPanel_I2S_DMA = unsupported color order 0
MatrixPanel_I2S_DMA config - 64x32 length: 2
R1_PIN=2, G1_PIN=6, B1_PIN=10, R2_PIN=3, G2_PIN=7, B2_PIN=11, A_PIN=39, B_PIN=38, C_PIN=37, D_PIN=36, E_PIN=21, LAT_PIN=33, OE_PIN=35, CLK_PIN=34
Length: 4096
MatrixPanel_I2S_DMA created
heap usage: 252
MatrixPanel_I2S_DMA begin ok
heap usage: 37244
MatrixPanel_I2S_DMA clear ok
MatrixPanel_I2S_DMA allocate memory
MatrixPanel_I2S_DMA allocate memory ok
MatrixPanel_I2S_DMA started, width=16320, 4096 pixels.
MatrixPanel_I2S_DMA LEDS buffer enabled.
MatrixPanel_I2S_DMA LEDS dirty bit optimization enabled.
MatrixPanel_I2S_DMA LEDS buffer uses 12800 bytes.
Heap after buses: 162896
Loading custom palettes
Loading custom ledmaps
strip buffer size: 16384B
Heap after strip init: 146632B
Segment geometry: 0,4096 -> 0,1 [1,0]
-- Started transition: S=0x3fced458 T(0x3fced3d0) O[0x3fcb82cc] OP[0x3fcb8328]
Writing settings to /wsec.json...
JSON buffer locked. (4)
JSON buffer released. (4)
Writing settings to /cfg.json...
JSON buffer locked. (2)
Cfg: Saving bus #0
  (0-4096, type:65, CO:1, rev:0, skip:0, AW:0 kHz:0, mA:0/0)
JSON buffer released. (2)
Loop took 315ms.
Usermods took 0ms.
Strip took 8ms.
Image playback end called
Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.

Core  1 register dump:
PC      : 0x40375def  PS      : 0x00060e30  A0      : 0x82021ad5  A1      : 0x3fcebcf0
=> 0x40375def: MatrixPanel_I2S_DMA::updateMatrixDMABuffer(unsigned short, unsigned short, unsigned char, unsigned char, unsigned char) at .pio/libdeps/HD-WF2_ESP32_S3/ESP32 HUB75 LED MATRIX PANEL DMA Display/src/ESP32-HUB75-MatrixPanel-I2S-DMA.cpp:330
A2      : 0x00000000  A3      : 0x00003fbf  A4      : 0x0000001f  A5      : 0x000000ff
A6      : 0x000000a0  A7      : 0x00000000  A8      : 0x00003fbf  A9      : 0x3fced36c
A10     : 0xfffffffe  A11     : 0x0000001f  A12     : 0x00000000  A13     : 0xffc13fbf
A14     : 0x00003fc0  A15     : 0x00000000  SAR     : 0x00000000  EXCCAUSE: 0x0000001c
EXCVADDR: 0x000000a0  LBEG    : 0x40056f5c  LEND    : 0x40056f72  LCOUNT  : 0x00000000
=> 0x40056f5c: ?? ??:0
=> 0x40056f72: ?? ??:0


Backtrace: 0x40375dec:0x3fcebcf0 0x42021ad2:0x3fcebd20 0x42022143:0x3fcebd50 0x4202324f:0x3fcebd80 0x4201d513:0x3fcebda0 0x4201d6f0:0x3fcebe00 0x4204c901:0x3fcebe30 0x4204cdfe:0x3fcebef0 0x42067805:0x3fcebf10
=> 0x40375dec: MatrixPanel_I2S_DMA::updateMatrixDMABuffer(unsigned short, unsigned short, unsigned char, unsigned char, unsigned char) at .pio/libdeps/HD-WF2_ESP32_S3/ESP32 HUB75 LED MATRIX PANEL DMA Display/src/ESP32-HUB75-MatrixPanel-I2S-DMA.cpp:329
=> 0x42021ad2: MatrixPanel_I2S_DMA::drawPixelRGB888(short, short, unsigned char, unsigned char, unsigned char) at .pio/libdeps/HD-WF2_ESP32_S3/ESP32 HUB75 LED MATRIX PANEL DMA Display/src/ESP32-HUB75-MatrixPanel-I2S-DMA.h:926
      (inlined by) VirtualMatrixPanel::drawPixelRGB888(short, short, unsigned char, unsigned char, unsigned char) at .pio/libdeps/HD-WF2_ESP32_S3/ESP32 HUB75 LED MATRIX PANEL DMA Display/src/ESP32-VirtualMatrixPanel-I2S-DMA.h:465
=> 0x42022143: BusHub75Matrix::show() at wled00/bus_manager.cpp:1074
      (inlined by) BusHub75Matrix::show() at wled00/bus_manager.cpp:1052
=> 0x4202324f: BusManager::show() at wled00/bus_manager.cpp:1321 (discriminator 2)
=> 0x4201d513: WS2812FX::show() at wled00/FX_fcn.cpp:1638
=> 0x4201d6f0: WS2812FX::service() at wled00/FX_fcn.cpp:1261
=> 0x4204c901: WLED::loop() at wled00/wled.cpp:127 (discriminator 5)
=> 0x4204cdfe: loop() at wled00/wled_main.cpp:23
=> 0x42067805: loopTask(void*) at C:/Users/blago/.platformio/packages/framework-arduinoespressif32@src-5da4143405621532219ebb625b78dd75/cores/esp32/main.cpp:50




ELF file SHA256: 0000000000000000

Rebooting...
ESP-ROM:esp32s3-20210327
```

</details>

In the log when setting chain to two i notice only this different `MatrixPanel_I2S_DMA started, width=16320, 4096 pixels.` what is setting the width of the panel to `16320=64*255` so i dont know if that is issue 

Setting to just one 64x32 panels is working correct ... no crashes and  `MatrixPanel_I2S_DMA started, width=64, 2048 pixels.` is correct

robertlipe:
I'm AFK so missing details but in your Platformio add a filter for
esp32monitor, I think it's called.nThis will translate that pile of numbers
in the crash to a decoded symbolic stack with function names, arguments,
and line numbers to aid in debugging.

platformio_override.ini
```diff
@@ -0,0 +1,88 @@
+# Example PlatformIO Project Configuration Override
+# ------------------------------------------------------------------------------
+# Copy to platformio_override.ini to activate overrides
+# ------------------------------------------------------------------------------
+# Please visit documentation: https://docs.platformio.org/page/projectconf.html
+
+[platformio]
+default_envs = esp32dev_hub75, adafruit_matrixportal_esp32s3, esp32S3_PSRAM_HUB75, esp32dev_hub75_forum_pinout
+
+[env:esp32dev_hub75]
```
netmindz: TODO: move this to the platform_override_sample.ini once testing complete

netmindz:
The main issue I am aware of at the moment is while the value you enter for the panel width is saved and applied, it's not loaded back properly for the settings page so always shows as empty and has to be re-entered every time

wled00/bus_manager.cpp
```diff
@@ -21,8 +21,37 @@
 #include "bus_wrapper.h"
 #include "bus_manager.h"
 
+// functions to get/set bits in an array - based on functions created by Brandon for GOL
+//  toDo : make this a class that's completely defined in a header file
+bool getBitFromArray(const uint8_t* byteArray, size_t position) { // get bit value
```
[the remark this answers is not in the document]
netmindz: @softhack007 ?

wled00/bus_manager.cpp
```diff
@@ -745,6 +774,347 @@ void BusNetwork::cleanup() {
...
+    return;
+  }
+
+#if defined(CONFIG_IDF_TARGET_ESP32) || defined(CONFIG_IDF_TARGET_ESP32S2)// classic esp32, or esp32-s2: reduce bitdepth for large panels
+  if (mxconfig.mx_height >= 64) {
+    if (mxconfig.chain_length * mxconfig.mx_width > 192) mxconfig.setPixelColorDepthBits(3);
+    else if (mxconfig.chain_length * mxconfig.mx_width > 64)  mxconfig.setPixelColorDepthBits(4);
+    else mxconfig.setPixelColorDepthBits(8);
+  } else mxconfig.setPixelColorDepthBits(8);
+#endif
```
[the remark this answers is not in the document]
netmindz: You are correct, I have fixed in 5629ed8b9

wled00/bus_manager.cpp
```diff
@@ -760,6 +790,350 @@ void BusNetwork::cleanup() {
...
```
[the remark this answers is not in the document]
netmindz: You are correct, I have fixed in 5629ed8b9

wled00/bus_manager.cpp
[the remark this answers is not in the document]
netmindz: I'm not sure the reason, but as it's only short and only during initialization I'm tempted to keep for now

wled00/bus_manager.cpp
```diff
@@ -760,6 +790,352 @@ void BusNetwork::cleanup() {
...
+  display->setBrightness8(25);    // range is 0-255, 0 - 0%, 255 - 100%
+
+  delay(24); // experimental
+  DEBUGBUS_PRINT(F("heap usage: ")); DEBUGBUS_PRINTLN(lastHeap - ESP.getFreeHeap());
+  // Allocate memory and start DMA display
+  if( not display->begin() ) {
+      DEBUGBUS_PRINTLN("****** MatrixPanel_I2S_DMA !KABOOM! I2S memory allocation failed ***********");
+      DEBUGBUS_PRINT(F("heap usage: ")); DEBUGBUS_PRINTLN(lastHeap - ESP.getFreeHeap());
+      return;
+  }
```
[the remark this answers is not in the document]
netmindz: no need to release the pins as it's not like you can reuse them for any other purpose if they are connected to the hub75 panel

wled00/bus_manager.cpp
```diff
@@ -760,6 +790,352 @@ void BusNetwork::cleanup() {
...
+    DEBUGBUS_PRINTLN("MatrixPanel_I2S_DMA allocate memory ok");
+
+    if (_ledsDirty == nullptr) {
+      display->stopDMAoutput();
+      delete display; display = nullptr;
+      _valid = false;
+      DEBUGBUS_PRINTLN(F("MatrixPanel_I2S_DMA not started - not enough memory for dirty bits!"));
+      DEBUGBUS_PRINT(F("heap usage: ")); DEBUGBUS_PRINTLN(lastHeap - ESP.getFreeHeap());
+      return;  //  fail is we cannot get memory for the buffer
+    }
```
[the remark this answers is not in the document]
netmindz: no need to release the pins as it's not like you can reuse them for any other purpose if they are connected to the hub75 panel

wled00/bus_manager.cpp
```diff
@@ -760,6 +790,352 @@ void BusNetwork::cleanup() {
...
+                mxconfig.gpio.r1, mxconfig.gpio.g1, mxconfig.gpio.b1, mxconfig.gpio.r2, mxconfig.gpio.g2, mxconfig.gpio.b2,
+                mxconfig.gpio.a, mxconfig.gpio.b, mxconfig.gpio.c, mxconfig.gpio.d, mxconfig.gpio.e, mxconfig.gpio.lat, mxconfig.gpio.oe, mxconfig.gpio.clk);
+
+  // OK, now we can create our matrix object
+  display = new MatrixPanel_I2S_DMA(mxconfig);
+  if (display == nullptr) {
+      DEBUGBUS_PRINTLN("****** MatrixPanel_I2S_DMA !KABOOM! driver allocation failed ***********");
+      DEBUGBUS_PRINT(F("heap usage: ")); DEBUGBUS_PRINTLN(lastHeap - ESP.getFreeHeap());
+      return;
+  }
```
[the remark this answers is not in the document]
netmindz: no need to release the pins as it's not like you can reuse them for any other purpose if they are connected to the hub75 panel

rhamblen:
am i missing something?  How do I enable the HUB75 support.  In the MM release it was a setting of the LED, but I cant find that setting in v16.  Tried a couple of different f/w  builds on esp32

netmindz:
@rhamblen see the sample environments in the platformio override example file

rhamblen:
thanks
just to confirm... I dont use the nightly build I have to compile my own build using platformio override example file rather than the default platformio file

PR #3790, opened 2024-03-03
Fix preset sorting

WoodyLetsCode:
This also adds the _better search_ for presets without the bug that the elements get reordered when you click on the search field.

| Before | Now |
|--------|--------|
| grafik | grafik|

### DIFF
--- a/wled00/data/index.htm
+++ b/wled00/data/index.htm
@@ -133,7 +133,7 @@
 		<p class="labels hd" id="pall"><i class="icons sel-icon" onclick="tglHex()">&#xe2b3;</i> Color palette</p>
 		<div id="palw" class="il">
 			<div class="staytop fnd">
-				<input type="text" placeholder="Search" oninput="search(this,'pallist')" onfocus="search(this,'pallist')" />
+				<input type="text" placeholder="Search" oninput="search(this,'pallist')" />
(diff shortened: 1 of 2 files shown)


wled00/data/index.htm
```diff
@@ -133,7 +133,7 @@
 		<p class="labels hd" id="pall"><i class="icons sel-icon" onclick="tglHex()">&#xe2b3;</i> Color palette</p>
 		<div id="palw" class="il">
 			<div class="staytop fnd">
-				<input type="text" placeholder="Search" oninput="search(this,'pallist')" onfocus="search(this,'pallist')" />
+				<input type="text" placeholder="Search" oninput="search(this,'pallist')" />
```
blazoncek, line 136: Why did you remove `onfocus` event? Now, if something forces a redraw of palettes (or effects or presets, see below) when you navigate to search field that already contains a value needs additional keystroke to update list.

blazoncek (commented):
Thank you. Generally approve but I have reservations regarding removal of `onfocus` events.

wled00/data/index.htm
WoodyLetsCode: I thought for quite a long time about what this was needed for, but I couldn't think of it. That's why I thought it wasn't needed. Now, thanks to your comment, I've thought of a case where it is needed. Thank you! I have changed it!

PR #3794, opened 2024-03-05
New Staircase usermod with position measurements

danielBreitlauch:
This is a user-mod for illuminating staircases. There is already one such user-mod but this one uses a ultrasound sensor to measure the position of a person walking the stairs.
Then only the path walked + some stairs ahead are turned on. This may help to not get blinded by turning on all lights ahead.
The important settings are all editable from the webUI but I guess this can also help others to build something similar.

Lmk what you guys think.

## Summary by CodeRabbit

- **New Features**
	- Introduced a smart staircase lighting control system that dynamically adjusts LED animations based on motion and distance detection. This feature integrates with smart home systems, allowing real-time sensor monitoring and automated light management.

- **Documentation**
	- Provided comprehensive installation and configuration instructions, including sensor setup guidelines and integration steps for an optimized staircase lighting experience.


blazoncek:
Thanks for the effort and contribution.

As I am not the user of staircase usermods (there are currently two of them) I may not know every desire people may have with such usermod.
AFAIK the Animated Staircase usermods also uses ultrasonic sensors as well as PIR sensors for activation. How come you didn't modify the existing one? It may be difficult to justify yet another staircase usermod if there are no mayor differences between them.

danielBreitlauch:
I started out just modifying the existing one but then realized it is too different what they do.
The existing usermod uses either PIR or ultrasound sensors to trigger an animation. My usermod uses PIR for triggering the animation and additionally an ultrasound sensor to measure the position of someone walking on the stairs.
On the existing mod every stair is activated by time. On my mod every stair is activated if the person is standing on it.

PR #3798, opened 2024-03-06
Esp32 s3 touch fix

DedeHai:
fixes touch for S2 and S3 boards. As touch does not work at all without this fix, even though not thoroughly tested, it cannot make things any worse I think.

### DIFF
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -255,10 +255,10 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
         }
         //if touch pin, enable the touch interrupt on ESP32 S2 & S3
         #ifdef SOC_TOUCH_VERSION_2    // ESP32 S2 and S3 have a fucntion to check touch state but need to attach an interrupt to do so
-          else if ((buttonType[s] == BTN_TYPE_TOUCH || buttonType[s] == BTN_TYPE_TOUCH_SWITCH)) 
-          {
-            touchAttachInterrupt(btnPin[s], touchButtonISR, 256 + (touchThreshold << 4)); // threshold on Touch V2 is much higher (1500 is a value given by Espressif example, I measured changes of over 5000)
-          }
+        if ((buttonType[s] == BTN_TYPE_TOUCH || buttonType[s] == BTN_TYPE_TOUCH_SWITCH)) 
+        {
+          touchAttachInterrupt(btnPin[s], touchButtonISR, 256 + (touchThreshold << 4)); // threshold on Touch V2 is much higher (1500 is a value given by Espressif example, I measured changes of over 5000)
+        }
         #endif 
         else
       #endif
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -257,7 +257,7 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
         #ifdef SOC_TOUCH_VERSION_2    // ESP32 S2 and S3 have a fucntion to check touch state but need to attach an interrupt to do so
           else if ((buttonType[s] == BTN_TYPE_TOUCH || buttonType[s] == BTN_TYPE_TOUCH_SWITCH)) 
           {
-            touchAttachInterrupt(btnPin[s], touchButtonISR, touchThreshold<<2); //threshold on Touch V2 is much higher (TODO: may need shift by 3 if very noisy)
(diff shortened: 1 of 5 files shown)


blazoncek:
Thanks. Can you also update `set.cpp`?

DedeHai:
> Thanks. Can you also update `set.cpp`?
> 
> EDIT: From lines 239 onward.

I see no change necessary, the digitalPinToTouchChannel() works as is for S2 and S3 (according to comment above the function): " this function works for ESP32, ESP32-S2 and ESP32-S3 - including the C3, it will return -1 for any pin"

blazoncek:
> I see no change necessary,

But where does then ISR get attached when user sets up a new button? He/she would need to bounce WLED. I do not want that.

DedeHai:
> > I see no change necessary,
> 
> But where does then ISR get attached when user sets up a new button? He/she would need to bounce WLED. I do not want that.

ISR does not actively need to be attached, its a native interrupt vector. pinmode() should take care of that. it works on my S3 so it seems interrupt is set correctly.
what do you mean by 'bounce WLED'?

wled00/cfg.cpp
```diff
@@ -252,8 +253,16 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
           btnPin[s] = -1;
           pinManager.deallocatePin(pin,PinOwner::Button);
         }
+        //if touch pin, enable the touch interrupt on ESP32 S2 & S3
+        #ifdef SOC_TOUCH_VERSION_2    // ESP32 S2 and S3 have a fucntion to check touch state but need to attach an interrupt to do so
+          else if ((buttonType[s] == BTN_TYPE_TOUCH || buttonType[s] == BTN_TYPE_TOUCH_SWITCH)) 
+          {
+            touchAttachInterrupt(btnPin[s], touchButtonISR, touchThreshold<<4); //threshold on Touch V2 is much higher (1500 is a value given by Espressif example)
```
blazoncek, line 260: I am talking about this. There is no counterpart in `set.cpp`.

DedeHai: oh, guess I was wrong. will check.

DedeHai: ok, I think I have it fixed now. I even added interrupt detach in the pin de-allocation function. touch now work immediately after saving button settings.

wled00/pin_manager.cpp
```diff
@@ -32,6 +32,11 @@ bool PinManagerClass::deallocatePin(byte gpio, PinOwner tag)
     return false;
   }
 
+  #ifdef SOC_TOUCH_VERSION_2 // ESP32 S2 and S3 have a fucntion to check touch state, detach any previous assignments
```
blazoncek, line 35: I do not think I like this approach here. PinManager never attached any interrupts/interrupt handlers to GPIO so it shouldn't detach any as well.
If you need to detach an interrupt handler it should be done at the place where such handler is attached (`set.cpp` only as there is no counterpart for boot sequence).

wled00/set.cpp
```diff
@@ -264,7 +265,7 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
           #ifdef SOC_TOUCH_VERSION_2 // ESP32 S2 and S3 have a fucntion to check touch state but need to attach an interrupt to do so
```
blazoncek, line 265: You will need to add logic to detect if previous button was a touch button and detach interrupt handler somewhere in this place.

blazoncek (changes requested):
Deataching interrupt handlers in PinManager is only an option if PinManager was also responsible for attaching such handlers.

IMO you can work to incorporate attaching handler into PinManager or remove detaching from PinManager and detach handler in `set.cpp` in the same place where attaching happens (I would prefer this second option).

wled00/set.cpp
DedeHai: I took another look at this. 
the previous assignment of a pin does not need to be checked, the interrupt can just be detached (the function will actually just set the callback to a null pointer and set the threshold to max so the interrupt never triggers).
I see one issue with removing the 'detach interrupt' from the deallocate function: if a pin is changed from a touch input to a non-button pin (say an LED output), the interrupt should also be disabled. what would be a good place to do that?

blazoncek: The same place attachment occurred IMO.

blazoncek (approved):

PR #3799, opened 2024-03-07
0 15 asyncwebserver 2.1.0 features

willmmiles:
This PR leverages the new features in the proposed AsyncWebServerWLED 2.1.0 build.  In particular:

- AsyncWebServerWLED 2.1.0 exposes its PROGMEM ContentType constants; use them where appropriate
- AsyncWebServerWLED 2.1.0 accepts Strings for `server.on()` requests, so we don't have to cast PROGMEM to `const char *` and rely on the alignment fault handler to correctly construct;
- AsyncWebServerWLED 2.1.0 introduces a SharedBuffer type that can be explicitly constructed and moved without the extra heap allocation and indirection of AsyncWebSocketMessageBuffer.  This eliminates the deprecation warnings for that type.

### DIFF
--- a/wled00/fcn_declare.h
+++ b/wled00/fcn_declare.h
@@ -434,7 +434,6 @@ void handleSerial();
 void updateBaudRate(uint32_t rate);
 
 //wled_server.cpp
-String getFileContentType(String &filename);
 void createEditHandler(bool enable);
 void initServer();
 void serveMessage(AsyncWebServerRequest* request, uint16_t code, const String& headl, const String& subl="", byte optionT=255);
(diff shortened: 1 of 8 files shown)


willmmiles:
@blazoncek  You were asking about the warning fixes - these are the relevant patches.

blazoncek:
Thanks.

So basically it is
```c++
AsyncWebSocketSharedBuffer newBuff(size);
...
ws->text(std::move(newBuff));
```

If I understand this (new to me) C++ syntax it says "use move constructor" when calling `text()` method? Is this correct?

willmmiles:
Superseded by #3828

PR #3800, opened 2024-03-07
Pin ArduinoFFT version

willmmiles:
The upstream repository has updated the develop branch to v2.0 which is not API compatible.  Pin the last commit before the new version was merged.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -172,7 +172,7 @@ lib_deps =
   # SHT85
     ;robtillaart/SHT85@~0.3.3
   # Audioreactive usermod
-    ;https://github.com/kosme/arduinoFFT#develop @ ^1.9.2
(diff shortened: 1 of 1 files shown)


blazoncek:
IMO we're better off accepting new API and modify AR usermod. @softhack007 your thoughts?

willmmiles:
> IMO we're better off accepting new API and modify AR usermod.

I agree that we should update to the new version soon. I suggested this patch as a safe, short term fix as the GitHub CI system is now reporting build failures for all PRs, and clean checkouts won't build anymore either unless your local PlatformIO has already cached the old version.

blazoncek:
Resolved by using release version 2.0.0 of ArduinoFFT.

willmmiles:
Good enough for me - I didn't have hardware on hand to test any changes, I just wanted the CI to work :)

PR #3828, opened 2024-03-16
ESPAsyncWebServer 2.2.0 + features

willmmiles:
Update to ESPAsyncWebServer 2.2.0 and leverage the new features.

- (ESP8266) Use the ContentType symbols from the AsyncWebServer instance, reducing PROGMEM duplication
- (ESP8266) Pass PROGMEM types to server.on(), so the String class will use the correct constructors instead of relying on the exception handler
- Use explicit allocation for AsyncWebSocket handling, eliminating unneeded object allocations and frees; also fixes deprecation warnings
- serveLiveLeds: avoid unneeded stack->buffer copy, and fix buffer overflow

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -144,7 +144,7 @@ lib_deps =
     fastled/FastLED @ 3.6.0
     IRremoteESP8266 @ 2.8.2
     makuna/NeoPixelBus @ 2.7.5
-    https://github.com/Aircoookie/ESPAsyncWebServer.git @ ^2.1.0
+    https://github.com/Aircoookie/ESPAsyncWebServer.git @ ^2.2.0
   # for I2C interface
     ;Wire
   # ESP-NOW library
--- a/wled00/fcn_declare.h
+++ b/wled00/fcn_declare.h
@@ -434,7 +434,6 @@ void handleSerial();
 void updateBaudRate(uint32_t rate);
 
 //wled_server.cpp
-String getFileContentType(String &filename);
 void createEditHandler(bool enable);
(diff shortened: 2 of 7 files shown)


blazoncek:
@willmmiles is this ready to be merged? I will blindly believe you. 😄

willmmiles:
I think so, but the whole point of a PR is to get a second pair of eyes in case I made some easy to spot mistake!   The changes are pretty shallow, with the intent to make it easier to review - it's probably easier to go through them one commit at a time.

There are a couple places where the code could maybe be cleaner at the expense of the patches being harder to follow.  One example is [factoring out the now-duplicate names in wled_server](https://github.com/Aircoookie/WLED/blob/df6c271830f567e47b310a53a46909ae20cf6922/wled00/wled_server.cpp#L21).  I had planned to send a second PR with cleanup later after you'd checked that the logic was sound.

The last commit (5f2480c) is probably the most critical, as it both fixes an actual buffer overflow when `(used/n)>226` and improves performance by avoiding the stack-to-heap copy.  Looking at it now, there's still room for improvement there too - it works, but we could shrink the dynamic allocation to fit the actually needed size instead of using a fixed value.

blazoncek:
Code inspection reveals no discrepancies but I have not tested it yet.
In any case everything is logical so if `AsyncWebSocketBuffer` or `DynamicBuffer` code have flaws it would need to be addressed there.

willmmiles:
I went ahead and did those two code quality improvements.  Might as well save on the testing effort!

PR #3836, opened 2024-03-19
Add SSD1309_64 I2C Support for FLD Usermod

THATDONFC:
Adds support for SSD1309_64 I2C display to the FLD Usermod.

Fixes an issue where palettes with an ID > 58 but not a custom palette did not display the palette name on the display.

There are quite a bit of whitespace diffs caused by trailing spaces being removed before commit.

### DIFF
--- a/usermods/usermod_v2_four_line_display_ALT/readme.md
+++ b/usermods/usermod_v2_four_line_display_ALT/readme.md
@@ -1,4 +1,4 @@
-# I2C 4 Line Display Usermod ALT
+# I2C/SPI 4 Line Display Usermod ALT
 
 Thank you to the authors of the original version of these usermods. It would not have been possible without them!
 "usermod_v2_four_line_display"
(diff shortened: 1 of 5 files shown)


blazoncek (changes requested):
I would avoid breaking changes if possible. Even at the expense of awkward UI order if necessary (but I think it will be ok).

THATDONFC:
> I would avoid breaking changes if possible. Even at the expense of awkward UI order if necessary (but I think it will be ok).

Got it. So move the new display to the end of the list and make it #9 and change the existing ones back to their original values?

My OCD is going to hate that but I understand wanting to avoid breaking changes wherever possible.

blazoncek:
Yes. Thank you.
And I completely understand your feelings as well.

THATDONFC:
This should revert the breaking change. Let me know if you want me to move the new display to the end of the list in each place I added it. I think this is a good compromise to keep the UI order organized while retaining backwards compatibility.

blazoncek (approved):

PR #3842, opened 2024-03-21
fix 12 typos

RoboSchmied:
fix typo: endianess --> endianness

### DIFF
--- a/wled00/src/dependencies/json/ArduinoJson-v6.h
+++ b/wled00/src/dependencies/json/ArduinoJson-v6.h
@@ -6621,28 +6621,28 @@ T& move(T& t) {
 }
 #endif
 #if ARDUINOJSON_LITTLE_ENDIAN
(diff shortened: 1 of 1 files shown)


softhack007:
Hi, it looks like your PR changes the name of one function in ArduinoJson-v6. Actually this file is not a creation of WLED authors, but just a copy&paste dependency library. The original lib is here: https://github.com/bblanchon/ArduinoJson

Can you confirm that your change is agreed by the authors of ArduinoJSON? 

I'd rather not get into a situation where we might want to update a dependency, but it's not possible any more  because in WLED the name of one function was changed/refactored to correct the spelling.

blazoncek:
I see no added value from this PR. Variable and function names may not be English at all.

softhack007:
Closed, as this PR creates a risk of becoming incompatible with upstream ArduinoJson.

PR #3848, opened 2024-03-23
Add ESP32 S2 Mini Board environment

LDprg:
Very much the same as lolin_s2_mini. However some HW pins mismatched and the board type too.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -11,7 +11,7 @@
 
 # CI binaries
 ; default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, esp32dev, esp32_eth # ESP32 variant builds are temporarily excluded from CI due to toolchain issues on the GitHub Actions Linux environment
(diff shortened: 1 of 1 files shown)


LDprg:
Actually the lovely s2 mini config just works fine (only from source not from precompiled binary).

eadmaster:
> only from source not from precompiled binary

you mean the binaries in the [releases page]( https://github.com/Aircoookie/WLED/releases)?

LDprg:
yes

eadmaster:
No problem, should i flash 
https://github.com/Aircoookie/WLED/releases/download/v0.14.4/WLED_0.14.4_ESP32-S2.bin

or

https://github.com/Aircoookie/WLED/releases/download/v0.15.0-b3/WLED_0.15.0-b3_ESP32-S2.bin

?

Last time i've tried the web server was working fine, but led controls were not working. Not sure if it is a wiring/voltage level issue with these small boards, but the same setup worked fine witb another esp8266 board.

LDprg:
I did flash 0.15.0-b3.

eadmaster:
the web server works fine, but i'm  getting some issues with the led control (blinking/cycling colors with no reason), are you using some level shifter?

I have a 30 leds strip and was working fine without level shifting on the other esp8266 board i have.

PR #3863, opened 2024-03-29
Use correct start index for Art-Net in proxy mode

askask:
Art-Net data doesn't contain a start byte. This is currently only taken into account when using regular LED output.
With this commit it also works when using DMX proxy mode.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed DMX data alignment across Art-Net and E1.31 lighting protocols for proper channel synchronization and correct data handling when using multiple protocol paths.

### DIFF
--- a/wled00/e131.cpp
+++ b/wled00/e131.cpp
@@ -93,11 +93,14 @@ void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol){
     return;
   }
 
+  // DMX data in Art-Net packet starts at index 0, for E1.31 at index 1
+  int8_t artnetOffset = (protocol == P_ARTNET) ? -1 : 0;
+
   #ifdef WLED_ENABLE_DMX
   // does not act on out-of-order packets yet
   if (e131ProxyUniverse > 0 && uni == e131ProxyUniverse) {
     for (uint16_t i = 1; i <= dmxChannels; i++)
-      dmx.write(i, e131_data[i]);
+      dmx.write(i, e131_data[i+artnetOffset]);
     dmx.update();
   }
   #endif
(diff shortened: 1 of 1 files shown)


troyhacks:
I think I understand it - it's just aligning incoming Art-Net/E1.31 data so channels from an Art-Net packet hits DMX output on the correct channel.

I haven't used this feature, but I think it's safe to merge.  Good catch, @askask!

wled00/e131.cpp
```diff
@@ -135,9 +138,8 @@ void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol){
     availDMXLen = (dmxChannels - DMXAddress) + dmxLenOffset;
   }
 
-  // DMX data in Art-Net packet starts at index 0, for E1.31 at index 1
-  if (protocol == P_ARTNET && dataOffset > 0) {
-    dataOffset--;
+  if (dataOffset > 0) {
```
softhack007, line 141: The next line should be guarded with `#ifdef WLED_ENABLE_DMX`, to avoid compile errors in "normal" builds  without full DMX support.

netmindz:
I'm not sure WLED is really appropriate for trying to be a an ArtNet/sACN to DMX (wired) convertor, but if has that feature it should certainly map the right values to the right channel.

If we are looking for confirmation of fix before merge with actual hardware I can probably rustle something up.

For future reference I would always recommend using a dedicated esp32 if you want such bridge from wireless to wired DMX, lighting desk send a universe at up to 44 Hz which WLED can't provide so you will lose frames and moving heads might behave erratically when getting a slow and inconsistent update

softhack007:
@coderabbitAI review

softhack007:
> If we are looking for confirmation of fix before merge with actual hardware I can probably rustle something up.

@netmindz any news?

askask:
This issue was fixed as part of #5547 (0f30c54ca6d621feed1cb2ccedd06d527d36f5d4)

PR #3868, opened 2024-03-31
added possible multi PIR usermod

rommo911:
added support for multi pir usermod

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -326,6 +326,12 @@ lib_deps =
 # WLED BUILDS
 # ------------------------------------------------------------------------------
 
+[env:nodemcuv2_160_pir]
+extends = env:nodemcuv2
+board_build.f_cpu = 160000000L
+build_flags = ${common.build_flags_esp8266} -D WLED_RELEASE_NAME=ESP8266_160 -DWLED_DISABLE_2D -D USERMOD_PIRSWITCH -D PIR_SENSOR_MAX_SENSORS=3 -D PIR_SENSOR_OFF_SEC=3
+
+
 [env:nodemcuv2]
 board = nodemcuv2
 platform = ${common.platform_wled_default}
--- a/usermods/PIR_sensor_switch/usermod_PIR_sensor_switch.h
+++ b/usermods/PIR_sensor_switch/usermod_PIR_sensor_switch.h
@@ -33,7 +33,7 @@ class PIRsensorSwitch : public Usermod
 {
 public:
   // constructor
-  PIRsensorSwitch() {}
+  PIRsensorSwitch(const char* nm) : _name(nm) {}
   // destructor
   ~PIRsensorSwitch() {}
 
@@ -72,7 +72,7 @@ private:
   bool HomeAssistantDiscovery = false;        // is HA discovery turned on
 
   // strings to reduce flash memory usage (used more than twice)
-  static const char _name[];
+  const char *_name;
   static const char _switchOffDelay[];
   static const char _enabled[];
   static const char _onPreset[];
@@ -185,7 +185,6 @@ public:
 };
 
 // strings to reduce flash memory usage (used more than twice)
-const char PIRsensorSwitch::_name[]           PROGMEM = "PIRsensorSwitch";
 const char PIRsensorSwitch::_enabled[]        PROGMEM = "PIRenabled";
 const char PIRsensorSwitch::_switchOffDelay[] PROGMEM = "PIRoffSec";
 const char PIRsensorSwitch::_onPreset[]       PROGMEM = "on-preset";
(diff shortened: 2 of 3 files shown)


blazoncek:
This will break existing integrations and MQTT workflows.
If you need multiple PIR sensors to trigger WLED the 0.15 has multi-pin PIR usermod.

rommo911:
> This will break existing integrations and MQTT workflows. If you need multiple PIR sensors to trigger WLED the 0.15 has multi-pin PIR usermod.

Alright, i could not actually understand how it will break MQTT, its been working fine for me, but then as u mentioned the 0.15 seems has better integration of this feature. thanks

rommo911:
actuall the multi pin pir usermod in 0.15 does only 'multi pin pir ' so only one PIR finally . 
mu usecase actually is having two seperate PIRs, each one discovered by hass as individiual PIR sesnor 
( using 3 PIRS under bed to do some cutom motion sesing and automation ion HASS  )

platformio.ini
```diff
@@ -326,6 +326,12 @@ lib_deps =
 # WLED BUILDS
 # ------------------------------------------------------------------------------
 
+[env:nodemcuv2_160_pir]
```
blazoncek, line 329: The change to `platformio.ini` is unacceptable. Please revert.

wled00/usermods_list.cpp
```diff
@@ -245,7 +245,8 @@ void registerUsermods()
   #endif
 
   #ifdef USERMOD_PIRSWITCH
-  usermods.add(new PIRsensorSwitch());
+  usermods.add(new PIRsensorSwitch("PIRsensorSwitch0"));
```
blazoncek, line 248: If you want to use a string, use a PROGMEM string and keep original name.

usermods/PIR_sensor_switch/usermod_PIR_sensor_switch.h
```diff
@@ -272,7 +271,7 @@ void PIRsensorSwitch::publishMqtt(const char* state)
   //Check if MQTT Connected, otherwise it will crash the 8266
   if (WLED_MQTT_CONNECTED) {
     char buf[64];
-    sprintf_P(buf, PSTR("%s/motion"), mqttDeviceTopic);   //max length: 33 + 7 = 40
+    sprintf_P(buf, PSTR("%s/motion_%s"), mqttDeviceTopic,_name);   //max length: 33 + 7 = 40
```
blazoncek, line 274: This will break existing integrations (mine included). Unacceptable change,

usermods/PIR_sensor_switch/usermod_PIR_sensor_switch.h
```diff
@@ -283,21 +282,21 @@ void PIRsensorSwitch::publishHomeAssistantAutodiscovery()
...
     doc[F("pl_off")] = "off";
-    sprintf_P(uid, PSTR("%s_motion"), escapedMac.c_str());
+    sprintf_P(uid, PSTR("%s_%s_motion"), escapedMac.c_str(),_name);
     doc[F("uniq_id")] = uid;
     doc[F("dev_cla")] = F("motion");
     doc[F("exp_aft")] = 1800;
 
     JsonObject device = doc.createNestedObject(F("device")); // attach the sensor to the same device
-    device[F("name")] = serverDescription;
+    device[F("name")] = String(serverDescription) ;
```
blazoncek, line 299: Why String and unnecessary fragment heap?

blazoncek (changes requested):
There are several things I do not like about this PR.
First, it should be based on the 0_15 branch as [requested](https://github.com/Aircoookie/WLED/blob/main/CONTRIBUTING.md).
If you want multiple MQTT targets, please modify existing 0.15 usermod that includes multi-pin extension.
Use strings stored in flash as the usermod is used also on ESP8266 which is short on RAM.

platformio.ini
rommo911: Noted , error from my side

wled00/usermods_list.cpp
rommo911: Noted

usermods/PIR_sensor_switch/usermod_PIR_sensor_switch.h
rommo911: I am not familiar with why this breaks it , although I will find a better solution for a better integration .

usermods/PIR_sensor_switch/usermod_PIR_sensor_switch.h
rommo911: My bad , I work all the time woth esp32 so I don't pay lot of attention to limitation of esp8266. But noted

rommo911:
Thank

> There are several things I do not like about this PR. First, it should be based on the 0_15 branch as [requested](https://github.com/Aircoookie/WLED/blob/main/CONTRIBUTING.md). If you want multiple MQTT targets, please modify existing 0.15 usermod that includes multi-pin extension. Use strings stored in flash as the usermod is used also on ESP8266 which is short on RAM.

Yes sure thing. I will close this PR, maybe I'll rebase from 0_15 and do better next time. Thanks

rommo911:
To be rebased an fixed

Issue #3855, opened 2024-03-25
Websocket connections not working anymore for >0.14.1


smhex:
### What happened?

I am using a third-party library (wled-client) to access my WLED using websockets. After upgrading my WLED to any version newer than 0.14.1 the library stopped working. I replaced it with a native websocket implementation and it didn't work either. I always get ECONNRESET when trying to open the socket. After downgrading to 0.14.1. everything works again. The web interface is accessible without any issues. Http API is also working as expected.

### To Reproduce Bug

Use any websocket client and connect to ws://<ipofwled>/ws. It should return immediately the wled state upon connect. In my case it returns nothing.

### Expected Behavior

Upon successful connection the state object should be returned.

### Install Method

Binary from WLED.me

### What version of WLED?

0.14.2

### Which microcontroller/board are you seeing the problem on?

ESP32

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Jedden19:
Same with ESP8266

blazoncek:
Please post your WS connection details and @willmmiles may have solution.

willmmiles:
Replicated on 0.14.2, works on 0.15.   Investigating.

willmmiles:
Seems to work on 0.14.3 as well.  I suspect this may have been fixed by the websocket memory management fixes in the newer version of AsyncWebServer.

smhex:
In the meantime I am able to catch a serial dump, however my WLED version does not include debug symbols. So nothing here to see :-(. I tested with 0.15.0-b1 and fails again. The connection request triggers a reboot of my device. If someone can provide a [WLED_0.15.0-b1_ESP32.bin](https://github.com/Aircoookie/WLED/releases/download/v0.15.0-b1/WLED_0.15.0-b1_ESP32.bin) or any other affected version with debug symbols enabled I am happy to test.

How do I get 0.14.3?

blazoncek:
Both versions available on Discord.

smhex:
Thx, will download and continue testing...

smhex:
Okay, here is the serial dump from a 0.14.3 debug build. As soon as my client connects, a reboot is triggered. I did this several times (see attached log)

[putty.log](https://github.com/Aircoookie/WLED/files/14812270/putty.log)

Is there anything more I can test/provide?

willmmiles:
[...]

PR #3873, opened 2024-04-01
Update to AsyncWebServer v2.2.1

willmmiles:
Fix use-after-free issue and slightly improve code size.  Note that the version is changed to a hard pin, so future updates can be validated before getting picked up by new clones, and old version builds are reproducible.


willmmiles:
Fixes #3855 and possibly some other AWS>=2.1 crash issues.

PR #3876, opened 2024-04-03
Create new stale.yml

w00000dy:
This uses the new [Close Stale Issues](https://github.com/marketplace/actions/close-stale-issues) GitHub Action.

To see if everything is working correctly, I have enabled [`debug-only`](https://github.com/marketplace/actions/close-stale-issues#debug-only), which means that **no GitHub API calls will be made that could change your issues and pull requests**.

This action will run every day at 12:00 UTC. You can also [run it manually](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow).

After the action has run, we can look at the log, and if everything looks fine, we can remove `debug-only: true` to let stale do its work.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,27 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '0 12 * * *'
+  workflow_dispatch:
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          days-before-stale: 120
+          days-before-close: 7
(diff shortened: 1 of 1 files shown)


softhack007:
@Aircoookie, @blazoncek If I remember correctly, we had a number of user complaints after activating the "stale bot" a few years ago? 

Do you think its a good idea to activate the [Close Stale Issues](https://github.com/marketplace/actions/close-stale-issues) GitHub action now and give it a try?

blazoncek:
I am all in favour of closing stale PRs or issues. We do have `keep` tag if we want to retain open PRs or issues.

blazoncek:
@w00000dy please check the output and let me know when to remove debug.

PR #3877, opened 2024-04-03
Effect blending styles

blazoncek:
Implements 14 different transitions/blends between different effects/modes.

Alternative to #3669 by @tkadauke with less memory fragmentation and utilisation.
Implemented using clipping rectangles which may not produce same results as original PR (#3669).

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -84,6 +84,10 @@ uint16_t      Segment::_lastPaletteBlend  = 0; //in millis (lowest 16 bits only)
 
 #ifndef WLED_DISABLE_MODE_BLEND
 bool Segment::_modeBlend = false;
+uint16_t Segment::_clipStart = 0;
+uint16_t Segment::_clipStop = 0;
+uint8_t  Segment::_clipStartY = 0;
+uint8_t  Segment::_clipStopY = 1;
 #endif
 
 // copy constructor
@@ -413,12 +417,17 @@ void Segment::restoreSegenv(tmpsegd_t &tmpSeg) {
 
 uint8_t IRAM_ATTR Segment::currentBri(bool useCct) {
   uint32_t prog = progress();
+  uint32_t curBri = useCct ? cct : (on ? opacity : 0);
   if (prog < 0xFFFFU) {
-    uint32_t curBri = (useCct ? cct : (on ? opacity : 0)) * prog;
-    curBri += (useCct ? _t->_cctT : _t->_briT) * (0xFFFFU - prog);
+    uint8_t tmpBri = useCct ? _t->_cctT : (_t->_segT._optionsT & 0x0004 ? _t->_briT : 0);
+#ifndef WLED_DISABLE_MODE_BLEND
+    if (blendingStyle > BLEND_STYLE_FADE) return _modeBlend ? tmpBri : curBri; // not fade/blend transition, each effect uses its brightness
+#endif
+    curBri *=  prog;
+    curBri += tmpBri * (0xFFFFU - prog);
     return curBri / 0xFFFFU;
   }
-  return (useCct ? cct : (on ? opacity : 0));
+  return curBri;
 }
 
 uint8_t IRAM_ATTR Segment::currentMode() {
@@ -431,16 +440,23 @@ uint8_t IRAM_ATTR Segment::currentMode() {
 
 uint32_t IRAM_ATTR Segment::currentColor(uint8_t slot) {
   if (slot >= NUM_COLORS) slot = 0;
+  uint32_t prog = progress();
+  if (prog == 0xFFFFU) return colors[slot];
 #ifndef WLED_DISABLE_MODE_BLEND
-  return isInTransition() ? color_blend(_t->_segT._colorT[slot], colors[slot], progress(), true) : colors[slot];
+  if (blendingStyle > BLEND_STYLE_FADE) return _modeBlend ? _t->_segT._colorT[slot] : colors[slot]; // not fade/blend transition, each effect uses its color
+  return color_blend(_t->_segT._colorT[slot], colors[slot], prog, true);
 #else
-  return isInTransition() ? color_blend(_t->_colorT[slot], colors[slot], progress(), true) : colors[slot];
+  return color_blend(_t->_colorT[slot], colors[slot], prog, true);
 #endif
 }
 
 CRGBPalette16 IRAM_ATTR &Segment::currentPalette(CRGBPalette16 &targetPalette, uint8_t pal) {
   loadPalette(targetPalette, pal);
   uint16_t prog = progress();
+#ifndef WLED_DISABLE_MODE_BLEND
+  if (prog < 0xFFFFU && blendingStyle > BLEND_STYLE_FADE && _modeBlend) targetPalette = _t->_palT; // not fade/blend transition, each effect uses its palette
+  else
+#endif
   if (strip.paletteFade && prog < 0xFFFFU) {
     // blend palettes
     // there are about 255 blend passes of 48 "blends" to completely blend two palettes (in _dur time)
@@ -456,9 +472,9 @@ CRGBPalette16 IRAM_ATTR &Segment::currentPalette(CRGBPalette16 &targetPalette, u
 void Segment::handleRandomPalette() {
   // is it time to generate a new palette?
   if ((millis()/1000U) - _lastPaletteChange > randomPaletteChangeTime) {
-        _newRandomPalette = useHarmonicRandomPalette ? generateHarmonicRandomPalette(_randomPalette) : generateRandomPalette();
-        _lastPaletteChange = millis()/1000U;
-        _lastPaletteBlend = (uint16_t)(millis() & 0xFFFF)-512; // starts blending immediately
+    _newRandomPalette = useHarmonicRandomPalette ? generateHarmonicRandomPalette(_randomPalette) : generateRandomPalette();
+    _lastPaletteChange = millis()/1000U;
+    _lastPaletteBlend = (uint16_t)(millis() & 0xFFFF)-512; // starts blending immediately
   }
 
   // if palette transitions is enabled, blend it according to Transition Time (if longer than minimum given by service calls)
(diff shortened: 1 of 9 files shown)


blazoncek:
I'd love to hear your thoughts @tkadauke.

tkadauke:
I certainly like how it's less complex :)

I'm out of town this week, but can test on the weekend. If this produces roughly the same results as my PR, I would definitely prefer it over mine.

blazoncek:
@willmmiles could you take a look as well?

willmmiles:
OK, will do!

wled00/FX_fcn.cpp
```diff
@@ -662,6 +680,32 @@ uint16_t IRAM_ATTR Segment::virtualLength() const {
...
+  if (_clipStart != _clipStop && blendingStyle > BLEND_STYLE_FADE) {
+    bool invert    = _clipStart > _clipStop;
+    unsigned start = invert ? _clipStop : _clipStart;
+    unsigned stop  = invert ? _clipStart : _clipStop;
+    if (blendingStyle == BLEND_STYLE_FAIRY_DUST) {
+      unsigned len = stop - start;
+      if (len < 2) return false;
+      unsigned shuffled = hashInt(i) % len;
+      unsigned pos = (shuffled * 0xFFFFU) / len;
+      return progress() <= pos;
```
willmmiles, line 698: This needs to be `return (progress() <= pos) ^ _modeBlend;`, otherwise we're using the same mask for old and new.

wled00/FX_fcn.cpp
```diff
@@ -662,6 +680,32 @@ uint16_t IRAM_ATTR Segment::virtualLength() const {
...
+    }
+    const bool iInside = (i >= start && i < stop);
+    if (!invert &&  iInside) return _modeBlend;
+    if ( invert && !iInside) return _modeBlend;
+    return !_modeBlend;
```
willmmiles, line 703: Consider using xor here - `return !iInside ^ invert ^ _modeBlend;`

willmmiles (commented):
Other notes:  with this patch, when I set blending mode to "fade", changing color settings fades from one to the next.  However, if I change the blending mode, changing colors just hard switches over - the fade is disabled but I don't get the new transition style.  I think we should be consistent and apply the transition style to color transitions as well, not just effect mode changes.

wled00/FX_fcn.cpp
blazoncek: Logic was never my best friend.

blazoncek:
> However, if I change the blending mode, changing colors just hard switches over - the fade is disabled but I don't get the new transition style.

This is what it is supposed to be as old effect has to run with its color and new effect with a new set of colors (or palette).
If you are using 1D strip then only to 6 transition/blend modes are available. However if you do not see "Swipe left" or "Swipe right" or "Fairy Dust" transitions, then something is amiss.

wled00/FX_fcn.cpp
willmmiles: Lol fair enough!  Xor in particular is the weird one I rarely use.   In this context, xor can be thought of as a "conditional invert", ie. `a ^ b` == `if (b) return !a; else return a;`

willmmiles:
> > However, if I change the blending mode, changing colors just hard switches over - the fade is disabled but I don't get the new transition style.
> 
> This is what it is supposed to be as old effect has to run with its color and new effect with a new set of colors (or palette). If you are using 1D strip then only to 6 transition/blend modes are available. However if you do not see "Swipe left" or "Swipe right" or "Fairy Dust" transitions, then something is amiss.

I definitely don't -- the other transition styles are only applied if `seg.mode != tmpMode`, not when any other parameter (colors, palette, etc.) is changed.

Honestly we might want to map out the API - how do `fadeTransition`, `modeBlending`, `blendingStyle` and `transitionTime` (which might be 0 aka no transition) all inter-relate for each type of change (mode change, non-mode parameter change).   It might be cleaner to drop `fadeTransition` and `modeBlending` entirely in favor of `blendingStyle` and using `transitionTime == 0` to indicate "no transition", since at least some clients have adopted that approach anyways.

willmmiles:
On another axis: if we wanted to completely integrate fading and other styles, instead of `isPixelClipped` we could have a `pixelBlendAlpha` call that could return any value from 0 (100% old) to 255 (100% new).  This could allow some more effect styles that combine approaches, and the infrastructure could also be used in the future to implement multi-effect overlays.

blazoncek:
Transitions or blending or fading, whatever you want to call it, progressively evolved from simple two-color blend and on/off fade into what it is now. The first iteration even didn't have universal (across all segments) color blending and was limited to first N segments.

So, a lot of flags are legacy based and could be integrated into `transitionDelay` as ultimately if that is 0, no transitions/blending/fading is performed and it (blending) does not take much (well, at least the amount it did in the past) CPU cycles.

The problem with `transitionDelay` is that it is a bit convoluted. There is also `transitionDelayDefault` and `jsonTransitionOnce` and perhaps some other hidden logic somewhere.

AFAIK the `fadeTransition` is used for on/off transition (entirely separate process) and color transition/blend, `modeBlend` just enables/disables effect transitions/blending (independent of color blending) and `blendingStyle` determines the kind of effect blend/transition which is new in this PR.

tkadauke:
Ok, so I finally got a chance to test this. As I said, I like how simple this is, and that it doesn't use 8 extra bytes per pixel, like my PR. That said, there are some gaps.

* As @willmmiles pointed out, only effect transitions are using the selected blend style. I think the real magic comes out when color/palette/brightness changes also trigger the transition. This would also make it much easier to test the transitions for correctness. It's hard to be sure, but I suspect that both "Pinch-out" and "Inside-out" are not working as intended, since I saw some abrupt brightness changes.
* Push Left/Push Right for 1D transitions aren't implemented. I think that wouldn't be too hard to add. In addition to clipping, you'd add a translation step to both the old and the new effect, which simply moves the rendered pixel by an offset (or x,y vector).
* In my PR, only transitions that work with the current strip style (1D vs 2D) are shown in the UI.
* Palette changes happen abruptly here. When I switch from solid white to Aurora using "Fairy Dust", I see some pixels fade out completely, until at the very end the palette switches to green from one frame to the next.
* Related to the above, if I disable "Effect Blending" in the LED settings, the UI still shows the blend selector (which I think it shouldn't), but the good news is that it actually is disabled.

I also noticed that as with my own PR, toggling the power state doesn't trigger transitions.

If I were to go from here, I'd also consider using masks for transitions. This would help simplify the Fairy Dust blend mode, and would remove the special case in the `isPixelClipped()` method. Masks could either be one bit per pixel, or a number from 0 to 255 to enable alpha blending, similar to what @willmmiles suggested above.

willmmiles:
> If I were to go from here, I'd also consider using masks for transitions. This would help simplify the Fairy Dust blend mode, and would remove the special case in the isPixelClipped() method. Masks could either be one bit per pixel, or a number from 0 to 255 to enable alpha blending, similar to what @willmmiles suggested above.

Heh, great minds think alike!  I thought about it but stopped short of suggesting full on alpha mask generation.  I had the crazy idea of using some of the existing FX code to generate an alpha mask to blend other FX.  The real coup would be tying it to some of the audio reactive code, I think it'd make some really neat composite effects..

As a core feature, though, it runs in to the code/CPU vs RAM tradeoff -- my understanding is that RAM is at a premium for many existing use cases, so contiguous space for extra buffers on the scale of pixel counts can't be assumed to be available.  The "Fairy Dust" implementation in this PR is a good example, it trades up doing a bunch of extra calculations for every render (something like 3x shift, 3x xor, 4*multiply, 2*divide -- all twice per pixel!) to avoid needing to allocate (and worse, manage!) a temporary buffer of pixel status.  It's difficult to strike a balance between supporting all features on large installations vs faster implementations when space permits - I don't envy @blazoncek for having to make those tough calls!  My $0.02 would be ship this approach now, and keep the idea of blending mask buffers for the next iteration.

wled00/FX.cpp
```diff
@@ -1211,28 +1211,29 @@ static const char _data_FX_MODE_COMET[] PROGMEM = "Lighthouse@!,Fade rate;!,!;!"
  */
 uint16_t mode_fireworks() {
   if (SEGLEN == 1) return mode_static();
-  const uint16_t width  = SEGMENT.is2D() ? SEGMENT.virtualWidth() : SEGMENT.virtualLength();
-  const uint16_t height = SEGMENT.virtualHeight();
+  const unsigned width  = SEGMENT.is2D() ? SEGMENT.virtualWidth() : SEGMENT.virtualLength();
```
blazoncek, line 1214: FFS! These changes were unintended for PR. 🤦‍♂️

blazoncek:
@willmmiles & @tkadauke I've finally got around to look into a few of the issues you brought up.
The push variants are still not implemented as in #3669 and UI may still need refinements.

blazoncek:
@tkadauke FYI On/Off blending style implemented.

The only thing remaining (compared to #3669) is "push" variants.

blazoncek:
@tkadauke I have push variants (all 8 of them) ready and will be available tonight.

tkadauke:
This is amazing. I am out of town today and tomorrow but happy to give this a try on Thursday if that sounds ok to you.

blazoncek:
Take your time. It may not be perfect still and I would be happy with help in debugging.

blazoncek:
Is there still any interest in this or do we abandon and close the PR?

willmmiles:
I think this is a valuable feature.  Other than correctness testing, what else needs to be done for it?

blazoncek:
> Other than correctness testing, what else needs to be done for it?

It just needs to be tested if there are any bugs or behaviour that is unexpected/unintended. I've seen occasion where transition was interrupted (and stopped). But only when switching effects or other parameters during an existing (long) transition. Unfortunately I could not catch it nor consistently reproduce.

blazoncek:
@DedeHai what did you do!?!?

blazoncek:
Closing this PR as @DedeHai made unintentional push with irrelevant changes.
Will also delete branch and push it again from my local copy.

PR #3892, opened 2024-04-09
Send ArtnetPollReply for Art-Net proxy universe

askask:
This commit adds sending an ArtnetPollReply for the e131ProxyUniverse, which is currently not implemented.
It also checks that this universe doesn't overlap with the regular e131Universe, so no duplicate replies are sent.

### DIFF
--- a/wled00/e131.cpp
+++ b/wled00/e131.cpp
@@ -346,7 +346,6 @@ void handleArtnetPollReply(IPAddress ipAddress) {
 
   switch (DMXMode) {
     case DMX_MODE_DISABLED:
-      return;  // nothing to do
       break;
 
     case DMX_MODE_SINGLE_RGB:
@@ -391,9 +390,17 @@ void handleArtnetPollReply(IPAddress ipAddress) {
       break;
   }
 
-  for (uint16_t i = startUniverse; i <= endUniverse; ++i) {
-    sendArtnetPollReply(&artnetPollReply, ipAddress, i);
+  if (DMXMode != DMX_MODE_DISABLED) {
+    for (uint16_t i = startUniverse; i <= endUniverse; ++i) {
+      sendArtnetPollReply(&artnetPollReply, ipAddress, i);
+    }
   }
+
+  #ifdef WLED_ENABLE_DMX
+    if (e131ProxyUniverse > 0 && (DMXMode == DMX_MODE_DISABLED || (e131ProxyUniverse < startUniverse || e131ProxyUniverse > endUniverse))) {
+      sendArtnetPollReply(&artnetPollReply, ipAddress, e131ProxyUniverse);
+    }
+  #endif
 }
 
 void prepareArtnetPollReply(ArtPollReply *reply) {


blazoncek:
@mxklb @troyhacks @netmindz as you have more DMX experience can you verify/review this PR, please?

netmindz:
Is there an associated issue detailing what the issue is that this PR seeks to resolve?

It's hard to review a solution without a clear understanding of the intention

wled00/e131.cpp
```diff
@@ -391,9 +390,17 @@ void handleArtnetPollReply(IPAddress ipAddress) {
...
 
-  for (uint16_t i = startUniverse; i <= endUniverse; ++i) {
-    sendArtnetPollReply(&artnetPollReply, ipAddress, i);
+  if (DMXMode != DMX_MODE_DISABLED) {
+    for (uint16_t i = startUniverse; i <= endUniverse; ++i) {
+      sendArtnetPollReply(&artnetPollReply, ipAddress, i);
+    }
   }
+
+  #ifdef WLED_ENABLE_DMX
```
netmindz, line 399: Why the ifdef?

netmindz (changes requested):

troyhacks:
I think it's a feature enhancement. 

With some of the existing WLED code we can automatically detect Art-Net endpoints, and this appears to extend this to the Art-Net proxy system.

I haven't used this functionality but I don't think there's any risk in adding it. 

Art-Net users at this level are few, I think. I'm one of them and I'm still kinda like ... whut? 😁

wled00/e131.cpp
troyhacks: As far as I can tell, this is allowing Art-Net to be translated/proxied to DMX512, like some of the hardware boxes I've been using (Pknight, for example).

This (and the previous Art-Net PR) seems to extend this functionality.

askask:
> I think it's a feature enhancement.
> 
> With some of the existing WLED code we can automatically detect Art-Net endpoints, and this appears to extend this to the Art-Net proxy system.

Yes, that's correct. This is a nice feature to get a quick overview (for example in QLC) of the available Art-Net nodes in the network and it would be useful if it also worked for proxied DMX outputs.

wled00/e131.cpp
askask: The variable `e131ProxyUniverse` used in the ifdef block is only available if `WLED_ENABLE_DMX` is defined ([here](https://github.com/Aircoookie/WLED/blob/main/wled00/wled.h#L414))

troyhacks:
I'm good with adding this. Thanks, @askask

wled00/e131.cpp
netmindz: yeah easier to see from my laptop than looking on my phone. I can see now that you are not moving any existing code into the ifdef, just adding extra which does depend on a value only present based on existing ifdef

wled00/e131.cpp
```diff
@@ -346,7 +346,6 @@ void handleArtnetPollReply(IPAddress ipAddress) {
 
   switch (DMXMode) {
     case DMX_MODE_DISABLED:
-      return;  // nothing to do
```
netmindz, line 349: while this code may well be redundant, code cleanup in a PR that is meant to be adding new features muddies the waters a bit. Better to keep the PR to only the changes required for the task

netmindz (changes requested):

mxklb:
Looking good. I'm not using this artnet poll feature, so can't test it, but looks like an improvement for DMX proxy. 

Save to merge

wled00/e131.cpp
askask: I replaced the return with an if condition because otherwise the code I added wouldn't be called if DMX mode is set to disabled

netmindz: ah, yes I follow what you are doing now

netmindz (approved):

PR #3897, opened 2024-04-12
reset usermod TetrisAI back to initial version

muebau:
With this usermod TetrisAI compiles again.


blazoncek:
As you are using `static` variables in your effect function: How does that scale up to playing Tetris on multiple segments?

softhack007:
So this basicially  reverts some of the last commits of the usermod (please correct me if I misunderstood)?

Wouldn't it be smarter to solve the compiler errors, instead of  stripping out things until it compiles?

usermods/TetrisAI_v2/usermod_v2_tetrisai.h
```diff
@@ -96,6 +94,8 @@ void drawGrid(TetrisAIGame* tetris, TetrisAI_data* tetrisai_data)
 ////////////////////////////
 uint16_t mode_2DTetrisAI()
 {
+  static unsigned long lastTime = 0;
```
softhack007, line 97: Don't use static variables in an effect function. 
You can either use `SEGENV.aux0`, `SEGENV.aux1`, `SEGENV.step`, or allocate persistent data by using `SEGENV.allocateData(dataSize)`.

usermods/TetrisAI_v2/usermod_v2_tetrisai.h
```diff
@@ -116,14 +116,16 @@ uint16_t mode_2DTetrisAI()
   //range 0 - 16
   tetrisai_data->colorInc = SEGMENT.custom2 >> 4;
 
-  if (!tetrisai_data->tetris || (tetrisai_data->tetris.nLookAhead != nLookAhead
+  static TetrisAIGame tetris(cols < 32 ? cols : 32, rows, 1, piecesData, numPieces);
```
softhack007, line 119: Please don't use static variables in an effect function (see previous comment).

usermods/TetrisAI_v2/usermod_v2_tetrisai.h
```diff
@@ -151,48 +153,48 @@ uint16_t mode_2DTetrisAI()
...
+      drawGrid(&tetris, tetrisai_data);
+      lastTime = millis();
+      tetris.poll();
     }
   }
-  else if (tetrisai_data->tetris.state == TetrisAIGame::ANIMATE_GAME_OVER)
+  else if (tetris.state == TetrisAIGame::ANIMATE_GAME_OVER)
   {
-    if (millis() - tetrisai_data->lastTime > msDelayGameOver)
+    if (millis() - lastTime > msDelayGameOver)
```
softhack007, line 173: Please use `strip.now` instead of `millis()`

softhack007 (changes requested):
Sorry I cannot approve this PR as it is now.
- please avoid using static or global variables in your effect function.
- please use strip.now in your effect function
- I might have misunderstood - but reverting commits "until it compiles" is usually not a good solution. I would suggest to look at the compiler errors in detail, understand the root cause in the code, and then improve the source code to make the compiler happy.

usermods/TetrisAI_v2/usermod_v2_tetrisai.h
softhack007: You can find many good examples for this in wled00/FX.cpp

blazoncek (changes requested):
I have to agree with @softhack007 and his review comments.

I also tried to provide a hint for better solution than this (which IMO is unacceptable). When you'll be updating the code, please also try to avoid `double` variables as they are slow compared to `float` counterparts (ESP32 has HW floating point but only for single precision, not double).

usermods/TetrisAI_v2/usermod_v2_tetrisai.h
muebau: I started to create a 'non blocking' version of the AI part. The commits I took back were related to that.

I need to have a look in the 'static' part.

usermods/TetrisAI_v2/usermod_v2_tetrisai.h
muebau: OK it should be very easy to change that.

muebau:
OK I need to test if float is good enough for the AI part.

blazoncek:
@muebau is there any interest in bringing this usermod up to speed and adjust code as requested or do we pull the plug and remove it from WLED?
The code that does not compile is of no use and code that does not meet requirements cannot be included.

muebau:
> @muebau is there any interest in bringing this usermod up to speed and adjust code as requested or do we pull the plug and remove it from WLED? The code that does not compile is of no use and code that does not meet requirements cannot be included.

Well I will try to find time to fix at least the easy parts ( millis() ).

I am still not sure how to target the problem with the state:
`  static TetrisAIGame tetris(cols < 32 ? cols : 32, rows, 1, piecesData, numPieces);`

I need an instance of `TetrisAIGame` for each segment, and it needs to be permanent. Every time the size of the segment (or other parameters) changes, I need to destroy the current instance and create a new one to be able to 'work' on the internal game grid.

I would appreciate suggestions from developers with more insight into WLED and C++.

muebau:
I think this should fix the issues.
What do you think?

robertvb83:
Nice update... Tested on my 32x32 Matrix with 4 segments
PXL_20240525_200225594~2

muebau:
Love to see it is working as it does here. 😂

muebau:
BTW is this a WS2812 4X16X16 setup or something else?

robertvb83:
> BTW is this a WS2812 4X16X16 setup or something else?

Yes exactly this, with an added 3d printed grid and diffuser foil

usermods/TetrisAI_v2/usermod_v2_tetrisai.h
```diff
@@ -143,35 +148,37 @@ uint16_t mode_2DTetrisAI()
...
     tetrisai_data->tetris = TetrisAIGame(gridWidth, gridHeight, nLookAhead, piecesData, numPieces);
+    tetrisai_data->tetris.state = TetrisAIGame::States::INIT;
     SEGMENT.fill(SEGCOLOR(1));
   }
 
   if (tetrisai_data->intelligence != SEGMENT.custom1)
   {
     tetrisai_data->intelligence = SEGMENT.custom1;
-    double dui = 0.2 - (0.2 * (tetrisai_data->intelligence / 255.0));
+    float dui = 0.2 - (0.2 * (tetrisai_data->intelligence / 255.0));
```
blazoncek, line 158: To reduce storage requirements add `f` to floating constants.

blazoncek (approved):
Apart from a few missing `f` for float constants this is ok for me.
@softhack007 do you have any objections?

usermods/TetrisAI_v2/usermod_v2_tetrisai.h
muebau: good finding. I fixed it and found a few more.

usermods/TetrisAI_v2/gridbw.h
```diff
@@ -112,6 +112,17 @@ class GridBW
     {
         return pixels[y] == (uint32_t)((1 << width) - 1);
     }
+
+    void reset()
+    {
+        if (width > 32)
+        {
+            throw std::invalid_argument("maximal width is 32");
```
softhack007, line 120: throw & catch are not supported on 8266. So this "thow" could cause a crash.

muebau: Ok what is your suggestion?
What would be a good alternative?

softhack007: Btw, why is the max width 32 pixels? Just lack of testing, or math breaks down (overflows), or too much memory needed? We have users with very large fixtures...

If your effect has some limits, the normal way is `return mode_static();`

muebau: The internal engine works on uint32 rows (32 pixels of max. width).
This way the grid has this limitation.

So I will return mode_static(); then.

muebau: I clamped the value to 32 and added a section to the readme

softhack007: 👍

softhack007: Looks like you still need to change the second "throw" in line 120

muebau: indeed. fixed.

softhack007 (approved):

usermods/TetrisAI_v2/gridbw.h
blazoncek: One more thing regarding the size. If the segment is larger than 32 pixels, just center everything and leave black bars on the sides.

blazoncek: You can, but that is a bad UX.
With centered game and leaving black bars, you are hinting a user that there is nothing of interest on the sides (or top & bottom) while still having a meaningful display.

blazoncek: You need to calculate offset manually.
```c++
int offsetX = cols > 32 ? (cols - width) / 2 : 0;
int offsetY = rows > 32 ? (rows - height) / 2 : 0;
...
SEGMENT.setPixelColorXY(x + offsetX, y + offsetY, color);
```
Where `width` and `height` are your desired/calculated width/height of the effect display.

muebau: Yes thank you. In the meantime, I have already solved it in exactly the same way.

muebau: I think it should have been resolved but while testing I found that 'peek' seems to scale pixels 'down' if I try to define something like 36 x 260.

Is there a way to simulate WLED (eg. vitual matrix, X86 simulator, debug window, etc.)?

muebau: The effect gets centered now. The maximal height gets also checked now. The 'next piece' and the border are placed smarter. You are right it was worth to have another look.

usermods/TetrisAI_v2/readme.md
```diff
@@ -6,7 +6,7 @@ Version 1.0
 
 ## Installation 
 
-Just activate the usermod with `-D USERMOD_TETRISAI` and the effect will become available under the name 'Tetris AI'.
+Just activate the usermod with `-D USERMOD_TETRISAI` and the effect will become available under the name 'Tetris AI'. If you are running out of flash memory, use a different memory layout (e.g. https://github.com/MoonModules/WLED/blob/mdev/tools/WLED_ESP32_4MB_256KB_FS.csv).
```
blazoncek, line 9: I would prefer you do not reference other repositories as their content may change without notice.
There are equivalent partitions included in this repository.

blazoncek (changes requested):
I would prefer you stick to this repository for completeness.

usermods/TetrisAI_v2/readme.md
softhack007: confirmed ... the 256KB file is also in the AC repository since a few months.
see https://github.com/Aircoookie/WLED/tree/0_15/tools for all partition files.

muebau: Cool there is one in the repo. I copied the link from the issue and missed the ones in the repo.

blazoncek (approved):

Issue #2932, opened 2022-12-08
Reboot with the new OS7.50 from the Fritzbox 7590


PatricckW:
### What happened?

The WLED restarted after 59 seconds, since the new software 7.50 is on the Fritzbox and it was rebooted once.
I have see this bug still on the Beta Versions from the Router, with the old OS7.29 I have no problems.

### To Reproduce Bug

Install the Version 7.50 and reboot the WLED, than it will restart after 59 Sec

### Expected Behavior

No reboots 

### Install Method

Binary from WLED.me

### What version of WLED?

WLED 0.13.3 (build 2208222)

### Which microcontroller/board are you seeing the problem on?

ESP32

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

softhack007:
Actually WLED does not know what kind of router it is connecting to, as we use the standard Wifi driver from espressif.

To better understand what happens, we first need to see if you get error messages on serial interface. Please connect your device to a Computer (using USB serial), and tell us what kind of messages you get on serial monitor (for example, you can use Arduino IDE serial monitor).

PatricckW:
That is very difficult vor me, it is from Cod.m without USB interface, only Wifi.
I use a Fritzbox 7590 from AVM, one of the most used Routers for Privat in Germany
![wled](https://user-images.githubusercontent.com/111278002/206523132-7b1e1653-542f-414a-b218-2c14bb620bb2.JPG)
![Unbenannt](https://user-images.githubusercontent.com/111278002/206523232-029cd93c-25a0-4ece-96fc-cc76a1e79e8e.JPG)

softhack007:
Just to check if the current version works better,

please install WLED 0.14 (beta 0) from one of these sites:
* https://wled-install.github.io
* or from https://github.com/srg74/WLED-wemos-shield/tree/master/resources/Firmware/WLED_wemos_shield/v0.14.0-b0

PatricckW:
I try version 14 but it is the same Problem

softhack007:
ok, then best to try another ESP32.

PatricckW:
I wait for a new one yes.
I geht the problem as I try the beta Version from AVM for the Router, wenn I go back to the Release all work fine, no restarts

PatricckW:
I try it today with an Devolo Repeater, which I have converted as an access point with own WLAN SSID
No it work fine, no Reboots yet

SmartHomeExpert:
[...]

PR #3902, opened 2024-04-15
update esp32 platform to arduino-esp32 v2.0.9

softhack007:
This PR upgrades the arduino-esp32 framework from v2.0.6 to 2.0.9.
I would suggest staying with this version, as later versions of arduino-esp32 (i.e. 2.0.10 thru 2.0.14) have new bugs and they create new problems with audioreactive.

I've used v2.0.9 on my development boards for some time now,  and did not find new issues. 

arduino-esp32 v2.0.9 fixes some very nasty memory leaks in the `WiFiUDP` class, that lead to low memory and spurious exception crashes.

In principle, we could even upgrade to `platform = espressif32@ 6.6.0` while still forcing arduino-esp32 2.0.9 with platform_packages=... -> to be discussed. Just that I haven't tested the combination of platform 6.6.0 / arduino-esp32 2.0.9.

<br/>

#### esp32 "V4"
```
PLATFORM: Espressif 32 (6.3.2)
PACKAGES:
 - framework-arduinoespressif32 @ 3.20009.0 (2.0.9)
 - tool-esptoolpy @ 1.40501.0 (4.5.1)
 - toolchain-xtensa-esp32 @ 8.4.0+2021r2-patch5
```

#### esp32-S3
```
PLATFORM: Espressif 32 (6.3.2)
PACKAGES:
 - framework-arduinoespressif32 @ 3.20009.0 (2.0.9)
 - tool-esptoolpy @ 1.40501.0 (4.5.1)
 - toolchain-riscv32-esp @ 8.4.0+2021r2-patch5
 - toolchain-xtensa-esp32s3 @ 8.4.0+2021r2-patch5
```

#### esp32-S2
```
PLATFORM: Espressif 32 (6.3.2)
PACKAGES:
 - framework-arduinoespressif32 @ 3.20009.0 (2.0.9)
 - tool-esptoolpy @ 1.40501.0 (4.5.1)
 - toolchain-riscv32-esp @ 8.4.0+2021r2-patch5
 - toolchain-xtensa-esp32s2 @ 8.4.0+2021r2-patch5
 ```

#### esp32-C3
```
PLATFORM: Espressif 32 (6.3.2)
PACKAGES:
 - framework-arduinoespressif32 @ 3.20009.0 (2.0.9)
 - tool-esptoolpy @ 1.40501.0 (4.5.1)
 - toolchain-riscv32-esp @ 8.4.0+2021r2-patch5
 ```

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -233,8 +233,8 @@ AR_lib_deps = kosme/arduinoFFT @ 2.0.0
 ;;
 ;; please note that you can NOT update existing ESP32 installs with a "V4" build. Also updating by OTA will not work properly.
 ;; You need to completely erase your device (esptool erase_flash) first, then install the "V4" build from VSCode+platformio.
-platform = espressif32@5.3.0
-platform_packages =
+platform = espressif32@ ~6.3.2
+platform_packages = platformio/framework-arduinoespressif32 @ 3.20009.0    ;; select arduino-esp32 v2.0.9 (arduino-esp32 2.0.10 thru 2.0.14 are buggy so avoid them)
 build_flags = -g
   -Wshadow=compatible-local ;; emit warning in case a local variable "shadows" another local one
   -DARDUINO_ARCH_ESP32 -DESP32
@@ -248,8 +248,8 @@ lib_deps =
 
 [esp32s2]
 ;; generic definitions for all ESP32-S2 boards
-platform = espressif32@5.3.0
-platform_packages =
+platform = espressif32@ ~6.3.2
+platform_packages = platformio/framework-arduinoespressif32 @ 3.20009.0    ;; select arduino-esp32 v2.0.9 (arduino-esp32 2.0.10 thru 2.0.14 are buggy so avoid them)
 default_partitions = tools/WLED_ESP32_4MB_1MB_FS.csv
 build_flags = -g
   -DARDUINO_ARCH_ESP32
@@ -267,8 +267,8 @@ lib_deps =
 
 [esp32c3]
 ;; generic definitions for all ESP32-C3 boards
-platform = espressif32@5.3.0
-platform_packages =
+platform = espressif32@ ~6.3.2
+platform_packages = platformio/framework-arduinoespressif32 @ 3.20009.0    ;; select arduino-esp32 v2.0.9 (arduino-esp32 2.0.10 thru 2.0.14 are buggy so avoid them)
 build_flags = -g
   -DARDUINO_ARCH_ESP32
   -DARDUINO_ARCH_ESP32C3
@@ -284,8 +284,8 @@ lib_deps =
 
 [esp32s3]
 ;; generic definitions for all ESP32-S3 boards
-platform = espressif32@5.3.0
-platform_packages =
+platform = espressif32@ ~6.3.2
+platform_packages = platformio/framework-arduinoespressif32 @ 3.20009.0    ;; select arduino-esp32 v2.0.9 (arduino-esp32 2.0.10 thru 2.0.14 are buggy so avoid them)
 build_flags = -g
   -DESP32
   -DARDUINO_ARCH_ESP32
--- a/platformio.ini
+++ b/platformio.ini
@@ -221,6 +221,7 @@ build_flags = -g
 tiny_partitions = tools/WLED_ESP32_2MB_noOTA.csv
 default_partitions = tools/WLED_ESP32_4MB_1MB_FS.csv
 extended_partitions = tools/WLED_ESP32_4MB_700k_FS.csv
+big_partitions = tools/WLED_ESP32_4MB_256KB_FS.csv     ;; 1.8MB firmware, 256KB filesystem, coredump support
 large_partitions = tools/WLED_ESP32_8MB.csv
 extreme_partitions = tools/WLED_ESP32_16MB_9MB_FS.csv
 lib_deps =
--- a/platformio.ini
+++ b/platformio.ini
@@ -174,7 +174,7 @@ lib_deps =
   # SHT85
     ;robtillaart/SHT85@~0.3.3
   # Audioreactive usermod
-    ;kosme/arduinoFFT @ 2.0.0
+    ;kosme/arduinoFFT @ 2.0.1
 
 extra_scripts = ${scripts_defaults.extra_scripts}
 
@@ -218,14 +218,18 @@ build_flags = -g
   #use LITTLEFS library by lorol in ESP32 core 1.x.x instead of built-in in 2.x.x
   -D LOROL_LITTLEFS
   ; -DARDUINO_USB_CDC_ON_BOOT=0 ;; this flag is mandatory for "classic ESP32" when building with arduino-esp32 >=2.0.3
+tiny_partitions = tools/WLED_ESP32_2MB_noOTA.csv
 default_partitions = tools/WLED_ESP32_4MB_1MB_FS.csv
+extended_partitions = tools/WLED_ESP32_4MB_700k_FS.csv
+large_partitions = tools/WLED_ESP32_8MB.csv
+extreme_partitions = tools/WLED_ESP32_16MB_9MB_FS.csv
 lib_deps =
   https://github.com/lorol/LITTLEFS.git
   https://github.com/pbolduc/AsyncTCP.git @ 1.2.0
   ${env.lib_deps}
 # additional build flags for audioreactive
 AR_build_flags = -D USERMOD_AUDIOREACTIVE
-AR_lib_deps = kosme/arduinoFFT @ 2.0.0
+AR_lib_deps = kosme/arduinoFFT @ 2.0.1
 
 [esp32_idf_V4]
 ;; experimental build environment for ESP32 using ESP-IDF 4.4.x / arduino-esp32 v2.0.5
@@ -250,7 +254,6 @@ lib_deps =
 ;; generic definitions for all ESP32-S2 boards
 platform = espressif32@ ~6.3.2
 platform_packages = platformio/framework-arduinoespressif32 @ 3.20009.0    ;; select arduino-esp32 v2.0.9 (arduino-esp32 2.0.10 thru 2.0.14 are buggy so avoid them)
-default_partitions = tools/WLED_ESP32_4MB_1MB_FS.csv
 build_flags = -g
   -DARDUINO_ARCH_ESP32
   -DARDUINO_ARCH_ESP32S2
@@ -296,7 +299,6 @@ build_flags = -g
   -DCO
   ;; please make sure that the following flags are properly set (to 0 or 1) by your board.json, or included in your custom platformio_override.ini entry:
   ;; ARDUINO_USB_MODE, ARDUINO_USB_CDC_ON_BOOT
-
 lib_deps =
   https://github.com/pbolduc/AsyncTCP.git @ 1.2.0
   ${env.lib_deps}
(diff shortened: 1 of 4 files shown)


willmmiles:
What are the chances you should find this now!  I was chasing a heap corruption issue under web server load over the weekend, and it kept faulting in the WifiUDP code (though this was with the old 1.x framework).  I was not very impressed by WifiUDP - allocating and freeing a buffer just to /poll/ if there's a packet seems really wasteful to me - but I wouldn't want to fork it just to fix that.

I hadn't seen `new(std::nothrow)` before.  There's probably a few places in the wled code this could also be used to more gracefully handle low memory conditions.

blazoncek:
Compiles ok, image size doesn't increase too much. 👍

blazoncek:
@softhack007 why are we not switching to v4 for ESP32?

softhack007:
> @softhack007 why are we not switching to v4 for ESP32?

Currently I think there are two blocking points

* 300kb increased firmware size with V4
* users need to erase their filesystem when installing V4. We had several reports that littlefs filesystem must be re-formatted, otherwise there are lots of errors plus slow UI responses.

willmmiles:
FWIW I tried platform @~5.3.0 with my ESP32 and WROVER a while back and I saw a bunch of filesystem issues; but my tests with platform @~6.3.2 over the last couple of days have given me no trouble at all upgrading from 3.5.0.

I do still see a 150kb code size increase from 3.5.0 to 6.3.2 on my base ESP32; I was surprised to see only a 5kb increase on the WROVER.  (Note my unit is a v3, so I've been disabling the compiler workaround for older chips.)

blazoncek:
> Note my unit is a v3, so I've been disabling the compiler workaround for older chips.

WROVER rev.1 will add ~100kB to the image size. Tested with 3.5.0, not yet with 6.

willmmiles:
> WROVER rev.1 will add ~100kB to the image size. Tested with 3.5.0, not yet with 6.

Yes.  I only mention that I'm disabling that feature in case someone is trying to replicate the metrics I'm reporting.  I haven't tested the `-mfix-esp32-psram-cache-issue` case, so in addition to the base size increase, it might also result in a different code size change between 3.5.0 and 6.3.2 as well.

blazoncek:
I am receiving compile errors for LittleFS:
```
.pio/libdeps/wemos_d1_mini32_debug/LittleFS_esp32/src/LITTLEFS.cpp: In member function 'virtual bool LITTLEFSImpl::exists(const char*)':
.pio/libdeps/wemos_d1_mini32_debug/LittleFS_esp32/src/LITTLEFS.cpp:44:28: error: no matching function for call to 'LITTLEFSImpl::open(const char*&, const char [2])'
     File f = open(path, "r");
                            ^
In file included from .pio/libdeps/wemos_d1_mini32_debug/LittleFS_esp32/src/LITTLEFS.cpp:17:
/Users/blaz/.platformio/packages/framework-arduinoespressif32/libraries/FS/src/vfs_api.h:38:17: note: candidate: 'virtual fs::FileImplPtr VFSImpl::open(const char*, const char*, bool)'
     FileImplPtr open(const char* path, const char* mode, const bool create) override;
                 ^~~~
/Users/blaz/.platformio/packages/framework-arduinoespressif32/libraries/FS/src/vfs_api.h:38:17: note:   candidate expects 3 arguments, 2 provided
Compiling .pio/build/wemos_d1_mini32_debug/libfde/LittleFS_esp32/esp_littlefs.c.o
*** [.pio/build/wemos_d1_mini32_debug/libfde/LittleFS_esp32/LITTLEFS.cpp.o] Error 1
.pio/libdeps/wemos_d1_mini32_debug/LittleFS_esp32/src/esp_littlefs.c:19:2: warning: #warning ("Use the built-in LITTLEFS library") [-Wcpp]
 #warning("Use the built-in LITTLEFS library")
  ^~~~~~~
.pio/libdeps/wemos_d1_mini32_debug/LittleFS_esp32/src/esp_littlefs.c: In function 'esp_littlefs_init':
.pio/libdeps/wemos_d1_mini32_debug/LittleFS_esp32/src/esp_littlefs.c:624:23: warning: declaration of 'err' shadows a previous local [-Wshadow=compatible-local]
             esp_err_t err;
                       ^~~
.pio/libdeps/wemos_d1_mini32_debug/LittleFS_esp32/src/esp_littlefs.c:520:15: note: shadowed declaration is here
     esp_err_t err = ESP_FAIL;
```

Does anyone know what's wrong? I only modified `platform = ${esp32_idf_V4.platform}` and `build_flags = ${common.build_flags_esp32_V4} ...`

blazoncek:
Found the issue. Had to update `lib_deps` as well.
Now my test debug build just crosses 100% flash size by 0.1%. It was at 99.8% prior to switch to 6.3.2.

There are plenty of "shadow" variables all over the place as well. 😄

blazoncek:
So, here I am with ESP32 platform 6.3.2 and a few usermods:
```
  -D WLED_DEBUG
  -D WLED_DISABLE_ALEXA
  -D WLED_DISABLE_HUESYNC
  -D WLED_DISABLE_LOXONE
  -D USERMOD_DALLASTEMPERATURE
  -D USERMOD_FOUR_LINE_DISPLAY
  -D USERMOD_ROTARY_ENCODER_UI
  -D USERMOD_PIRSWITCH
  -D USERMOD_MULTI_RELAY
  -D USERMOD_AUDIOREACTIVE
```
produces
```
RAM:   [==        ]  18.3% (used 59876 bytes from 327680 bytes)
Flash: [========= ]  92.4% (used 1574425 bytes from 1703936 bytes)
```

willmmiles:
Might be a good time to try LTO?

blazoncek:
WROVER has better results with fewer usermods but compared to 3.5.0 the size decreased.
```
RAM:   [          ]   4.2% (used 55444 bytes from 1310720 bytes)
Flash: [==========]  98.0% (used 1541125 bytes from 1572864 bytes)
```

softhack007:
> Might be a good time to try LTO?

Agreed - this could help solve the first problem (firmware size). Another possibily would be to use the "700kb filesystem" partition file instead of the 1MB layout.

We'd still have to tackle the second problem category "device slow and very unresponsive" that was reported by some users - for example https://github.com/Aircoookie/WLED/issues/2932#issuecomment-1902571336. Unfortunately I cannot reliably reproduce that one...

softhack007:
> What are the chances you should find this now! I was chasing a heap corruption issue under web server load over the weekend, and it kept faulting in the WifiUDP code (though this was with the old 1.x framework). I was not very impressed by WifiUDP - allocating and freeing a buffer just to /poll/ if there's a packet seems really wasteful to me - but I wouldn't want to fork it just to fix that.

😄 I know exactly what you mean with "not very impresed". In the MM fork, we have started to hunt down memory leaks from WifiUDP some time ago, and we saw that some problems "magically disappeared" with arduino-esp32 2.0.9.

Another strange thing: when you do a `Udp.parsePacket()` but don't do `Udp.read()` because the size is unexpected, then the next `parsePacket()` starts to behave strangely and `Udp.flush()` does help for some reason.. in fact flush() seems to clear the input buffer, which is the opposite of what happens on 8266 where flush() waits until the output buffer is sent,

willmmiles:
> We'd still have to tackle the second problem category "device slow and very unresponsive" that was reported by some users - for example [#2932 (comment)](https://github.com/Aircoookie/WLED/issues/2932#issuecomment-1902571336). Unfortunately I cannot reliably reproduce that one...

If I'm reading it correctly, it looks like that issue was being reported with the older v4.x or v5.x platforms though -- maybe things have improved by v6.3 that we're testing here?  🤞  In my local development I've found that heap corruption can look an awful lot like "device slow and unresponsive" when the micro keeps unexpectedly rebooting in the middle of a request.

willmmiles:
> 😄 I know exactly what you mean with "not very impresed". In the MM fork, we have started to hunt down memory leaks from WifiUDP some time ago, and we saw that some problems "magically disappeared" with arduino-esp32 2.0.9.

I took a look at the WifiUDP changes between versions - they definitely fixed a leak there.  The problem I was chasing was definitely a heap corruption issue though (my money'd be on a use-after-free) ... but it also just went away with arduino-esp32 2.0.9.  I think WifiUDP was just the victim for me as it was doing a lot of repeated heap operations.  I have a sneaky suspicion there's some other fix in between versions there that was the real helper for my problem.  Either way, strong indication that we can improve stability by updating.

> Another strange thing: when you do a Udp.parsePacket() but don't do Udp.read() because the size is unexpected, then the next parsePacket() starts to behave strangely and Udp.flush() does help for some reason.. in fact flush() seems to clear the input buffer, which is the opposite of what happens on 8266 where flush() waits until the output buffer is sent.

Right, it's got all that extra logic that throws away the packet size and puts the data in a ring buffer -- ie. it tries to treat the incoming data as a byte stream, not a packet stream.   I can see that being useful for some use cases, but any protocol where the datagram boundaries has meaning is a poor fit for WifiUDP.   Maybe we *should* sketch a better one - it's not particularly large or complex.

wled-install:
We could build custom arduino core excluding some stuff that nobody uses (like bluetooth or SPI ethernet drivers)  to reduce flash and RAM usage. This was the solution for me in other software to be able to update from 2.0.2 to 2.0.14. Custom core will be eventually also the only solution to handle the next monster comming called arduino core 3.x.x. (but this is quite remote future)

blazoncek:
@softhack007 I was able to update WROVER (rev.1) model (actually 3 of them) using OTA with version 4 environment and Arduino 2.0.9 without issues and running better with smaller flash footprint.
Screenshot 2024-04-20 at 21 24 15

Unfortunately my debug build for WROOM model is too big to successfully update OTA even though it builds at 99.8% of flash. Regular builds may be ok with just one or two usermods.

blazoncek:
Agreed. We can start with WROVER and build from there.

softhack007:
> Agreed. We can start with WROVER and build from there.

👍 I like this idea. So I'll upgrade the WROVER buildenv to use V4, then merge this PR.

blazoncek:
@softhack007 I would also like to remove the use of `${common.build_flags_esp32}` and replace it with `${common.build_flags} ${esp32.build_flags}` to be in line with S2, S3 and C3.

I also talked with @Aircoookie and he agrees to include Audioreactive in default build (no more separate build). We will refer to it not as a usermod but an optional core extension (to differentiate it from usermods).

I would also like that for Multi-relay and PIR sensor and DallasTemperature usermods as several available boards feature capabilities for those and were requested in Discord poll.

My recommendation regarding S3 would also be to separate builds for 16MB OPI, 8MB OPI and 4MB QSPI until @lost-hope figures out how to provide the choice to select flash size independently.

softhack007:
@blazoncek thanks, let me try to multi-answer your multiple points

> @softhack007 I would also like to remove the use of `${common.build_flags_esp32}` and replace it with `${common.build_flags} ${esp32.build_flags}` to be in line with S2, S3 and C3.

I agreed - I can look into this tomorrow.

> 
> I also talked with @Aircoookie and he agrees to include Audioreactive in default build (no more separate build). We will refer to it not as a usermod but an optional core extension (to differentiate it from usermods).

👍 

What we need to look into: should audioreactive still allocate all ressources at setup() even when it's not activated?
- default pins only "clained" when AR is set to enabled
- what to do with I2S#0 - in principle user should be able to use 10 led pins, however today we try to avoid conflicts by limiting user to 9 led pins. Could be solved if pinmanager is also used for managing allocation of  I2S units.


> 
> I would also like that for Multi-relay and PIR sensor and DallasTemperature usermods as several available boards feature capabilities for those and were requested in Discord poll.
> 

Generally makes sense for me.

Maybe we need to re-shuffle the usermods folder for this, to make it obvious that we have to classes of usermods - "core_extensions" and "contrib" maybe.

> My recommendation regarding S3 would also be to separate builds for 16MB OPI, 8MB OPI and 4MB QSPI until @lost-hope figures out how to provide the choice to select flash size independently.

👍

willmmiles:
> Maybe we need to re-shuffle the usermods folder for this, to make it obvious that we have two types of usermods - "core_extensions" and "contrib" (or any better name).

Put my $0.02 behind this proposal!  I'd like to push towards moving some of the other conditionally compiled core subsystems over to the module framework as well.

DedeHai:
Since your are on discord sabbatical (that's a good thing) let me just leave this here:
I was looking into code size optimisations recently and found a few things. did you already look into this topic?  there seem to be a few ways, best would be link-time optimization (but not officially supported). 
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/performance/size.html  has some compile flags for lower code size
the project https://github.com/letscontrolit/ESPEasy/  uses concatinating script for ESP8266. I wrote to TD-er to ask about LTO (-flto flag) and he replied that they have it working except for ESP8266.

PR #3904, opened 2024-04-15
added improvements to color scaling and blurring

DedeHai:
- made blurring faster by not writing the color and then reading it back but keeping it as a variable: on a C3, FX black hole goes from 55FPS to 71FPS 
- added optional parameter to blur function (smear) that can be used in combination with SEGMENT.clear(), blurring the frame without dimming the current display (repeated calls without clearing will flood the segment). this is useful to blur without 'motion blurring' being added.
- scale8 is inlined and repeated calls uses flash, plus it is slower than native 32bit, so I added 'color_scale' function which is native 32bit and scales 32bit colors (RGBW).
- bonus: changes save roughly 600bytes of flash memory

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -582,7 +582,7 @@ typedef struct Segment {
     #endif
     uint32_t getPixelColor(int i);
     // 1D support functions (some implement 2D as well)
-    void blur(uint8_t);
+    void blur(uint8_t, bool smear = false);
     void fill(uint32_t c);
     void fade_out(uint8_t r);
     void fadeToBlackBy(uint8_t fadeBy);
@@ -610,7 +610,7 @@ typedef struct Segment {
     inline void setPixelColorXY(float x, float y, byte r, byte g, byte b, byte w = 0, bool aa = true) { setPixelColorXY(x, y, RGBW32(r,g,b,w), aa); }
     inline void setPixelColorXY(float x, float y, CRGB c, bool aa = true)                             { setPixelColorXY(x, y, RGBW32(c.r,c.g,c.b,0), aa); }
     #endif
-    uint32_t getPixelColorXY(uint16_t x, uint16_t y);
+    uint32_t getPixelColorXY(int x, int y);
     // 2D support functions
     inline void blendPixelColorXY(uint16_t x, uint16_t y, uint32_t color, uint8_t blend) { setPixelColorXY(x, y, color_blend(getPixelColorXY(x,y), color, blend)); }
     inline void blendPixelColorXY(uint16_t x, uint16_t y, CRGB c, uint8_t blend)         { blendPixelColorXY(x, y, RGBW32(c.r,c.g,c.b,0), blend); }
@@ -619,8 +619,8 @@ typedef struct Segment {
     inline void addPixelColorXY(int x, int y, CRGB c, bool fast = false)                             { addPixelColorXY(x, y, RGBW32(c.r,c.g,c.b,0), fast); }
     inline void fadePixelColorXY(uint16_t x, uint16_t y, uint8_t fade)                               { setPixelColorXY(x, y, color_fade(getPixelColorXY(x,y), fade, true)); }
     void box_blur(uint16_t i, bool vertical, fract8 blur_amount); // 1D box blur (with weight)
-    void blurRow(uint16_t row, fract8 blur_amount);
-    void blurCol(uint16_t col, fract8 blur_amount);
+    void blurRow(uint32_t row, fract8 blur_amount, bool smear = false);
+    void blurCol(uint32_t col, fract8 blur_amount, bool smear = false);
     void moveX(int8_t delta, bool wrap = false);
     void moveY(int8_t delta, bool wrap = false);
     void move(uint8_t dir, uint8_t delta, bool wrap = false);
@@ -655,8 +655,8 @@ typedef struct Segment {
     inline void addPixelColorXY(int x, int y, CRGB c, bool fast = false)          { addPixelColor(x, RGBW32(c.r,c.g,c.b,0), fast); }
     inline void fadePixelColorXY(uint16_t x, uint16_t y, uint8_t fade)            { fadePixelColor(x, fade); }
     inline void box_blur(uint16_t i, bool vertical, fract8 blur_amount) {}
-    inline void blurRow(uint16_t row, fract8 blur_amount) {}
-    inline void blurCol(uint16_t col, fract8 blur_amount) {}
+    inline void blurRow(uint32_t row, fract8 blur_amount, bool smear = false) {}
+    inline void blurCol(uint32_t col, fract8 blur_amount, bool smear = false) {}
     inline void moveX(int8_t delta, bool wrap = false) {}
     inline void moveY(int8_t delta, bool wrap = false) {}
     inline void move(uint8_t dir, uint8_t delta, bool wrap = false) {}
--- a/wled00/FX_2Dfcn.cpp
+++ b/wled00/FX_2Dfcn.cpp
@@ -275,8 +275,7 @@ uint32_t IRAM_ATTR Segment::getPixelColorXY(int x, int y) {
 
 // blurRow: perform a blur on a row of a rectangular matrix
 void Segment::blurRow(uint32_t row, fract8 blur_amount, bool smear){
-  if (!isActive() || blur_amount == 0)
-    return; // not active
+  if (!isActive() || blur_amount == 0) return; // not active
   const uint_fast16_t cols = virtualWidth();
   const uint_fast16_t rows = virtualHeight();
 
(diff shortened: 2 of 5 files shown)


blazoncek (changes requested):
In general I approve of the changes with minor issues regarding formatting and unnecessary modifications.

I would avoid introducing new function `color_scale()` when `color_fade()` is intended to do the same. If you find the *fade* misleading (as opposed to *scale*) then please change the original function name instead. There are only a handful of references used, but I would recommend against it.

DedeHai:
Thanks for the review (and for spotting that bug in fx_fcn, it should indeed be 255, leftover from testing). I will fix the unnecessary changes, that was my auto-format plugin that sometimes does this if I am not careful. The new function should indeed be integrated into color_fade, this resulting code is what came from a lot of code changes that did not make it into the final version.

blazoncek (approved):

blazoncek:
FYI I do not see any performance gain on ESP32 or ESP8266.

softhack007:
> FYI I do not see any performance gain on ESP32 or ESP8266.

@DedeHai thanks for this PR 😃 I've also merged it into the MoonModules fork. It does give me the performance gain you describe, at least on larger fixtures (tested with 32x32 on 4 led pins). The main speedup comes from your optimizations of 2D blur, looks like color_add only has a minor part in the speedup.  Good improvement, as blur() and fade_out are indeed on a very critical path👍

DedeHai:
@softhack007 you are welcome. I discovered this 'insufficiency' in my particle system ventures where I noticed that accessing getPixelColorXY() and setPixelColorXY() are quite slow due to the many checks (and conversions) that need to be performed and blurring did it twice per pixel which I changed to only once per pixel, making it much faster.
The other changes give minor speed boosts, removing access to scale8() is not that much faster but uses way less code (scale8 is inlined each time). I did not change color_add() but color_fade(), saving some CPU cycles on all color fade functions.
there are many other places where the fastled functions that are optimized for 8bit ATMEGAs could be replaced with native 32bit functions which would save some more code but it is probably not worth the effort. 
Further speed improvements could be made by rendering blur/fade to a local RGB buffer and then transferring that to the outputbuffer (which I am doing in the ParticleSystem, roughly cutting blur computation time in half).

softhack007:
> Further speed improvements could be made by rendering blur/fade to a local RGB buffer and then transferring that to the outputbuffer (which I am doing in the ParticleSystem, roughly cutting blur computation time in half).

I think this is already happening in `0_15` if you set "Use global LED buffer" in LED settings - Pixels are rendered into a buffer,  and GetPixelColor will return what's in the buffer. The complete buffer is transferred to the NPB driver on busses.show() in one shot.

🤔 Maybe conversions and several layers of abstractions - that happen between Segment::setPixelColor() and BusDigital::setPixelColor() - are still slowing down the process, so you see a 2X speedup when adding your own buffer on top... 

<br/>


**(off-topic)** 

just for comparison, you could try to run your particle code in the [MoonModules fork](https://github.com/MoonModules/WLED)
We are a few months behind the mainstream "Aircoookie" code, however we have a number of experimental optimizations to improve speed on larger setups. A few differences you would need to consider:
* _disable_ global LED buffer, to enable buffering on segment level
* your effects must explicitly request buffering with `if (SEGENV.call == 0) SEGENV.setUpLeds();`
* use a buildenv like `env:esp32_4MB_S` which uses optimization for speed (i.e. "gcc -O2" instead of "gcc -Os"). 
* back up your cfg.json and presets.json - some MM build have a different partition table, so your old filesystem might get erased.
* experiment with enabling [this line](https://github.com/MoonModules/WLED/blob/6c93250cd8515b36bbe40596078ed5aacfb96e04/wled00/FX_2Dfcn.cpp#L234) - it prevents re-writing pixels of the same color (kind of dangerous, and some effects will not like it)


I would be interested to know if the MM code performs better for you. maybe we find the pieces that could be brought back into "upstream AC" to further optimize speed.

blazoncek:
FYI a future version of NPB does away with double/tripple buffering and `GetPixelColor()` will return unmodified original `SetPixelColor()` value which will remove the need for global LED buffer. I am in talks with @Makuna for testing that. There are several optimisations in NPB as a result of this transition.

If we reintroduce my infamous `setUpLeds()` is debatable and quite possible as a remedy for numerous translations along the way of `setPixelColor()`. It will not be CRGB though.

blazoncek:
> Using a local buffer with a known size can get away without all those checks and just straight forward copy the data, which is a lot faster. the blur function acts on a known segment size and could safely do so.

That's what `setUpLeds()` was all about. But it was incompatible with Harry Baas segment set-up so it had to go.

Makuna:
> > Maybe conversions and several layers of abstractions - that happen between Segment::setPixelColor() and BusDigital::setPixelColor() - are still slowing down the process, so you see a 2X speedup when adding your own buffer on top...
> 
> the issue is ... the checking that setPixelColor() (rightfully) does. 
the Pixels() method returns a pointer to buffer directly, You can use that to bypass the range checking.

blazoncek:
@Makuna we have translation from Segment (a virtual canvas) into pixel strip. Segments *can* overlap and can also encompass inexistent pixels so quite a few translations have to be made to set/get exact pixel.

DedeHai:
> overlapping Segments will wreck havoc if you do not blend local buffers correctly. Which you can't from within Segment environment (effect function).

how so? OR: how are overlapping segments handled in general? What I am wondering: with overlapping segments setPixelColor() will overwrite what the segment put there or is there a check for that and it will blend?

blazoncek:
Segments are individual and do not share any data between themselves - except actual LED color returned by `getPixelColor()` if they overlap and _non-local_ LED buffer is used.

_Non-local_ LED buffer is either _global_ buffer or NPB's _edit_ buffer. The problem with (current) NPB's _edit_ buffer is the modified LED data if brightness is set to anything other than 255 (or 65535 for 16bit) as when you apply `setBrightness()` every `setPixelColor()` is scaled according to brightness and stored in _edit_ buffer as such. When retrieved with `getPixelColor()` it is no longer equal to what was set with `setPixelColor()`. That's why _global_ buffer was introduced.

`setUpLeds()` introduced _local_ LED buffer (prior to introducing _global_ buffer) so that effect that relied on previous pixel values worked correctly (several Audioreactive/FastLED effects). With _local_ buffer every Segment was totally independent from any other Segment. This was removed when _global_ buffer was added as _Harry Baas_ segment set-up didn't work with _local_ buffer. This was a demand from SR team (and MM fork implemented several of their own solutions to tackle this).

How overlapping segments behave (with _non-local_ buffer) very much depends on how the effects are written. Some will take into account previous/underlying pixel values some will not. So you can expect anything from only one effect showing to uncontrolled flickering and since segments are stored in a vector their order may get shifted when vectors are reallocated in memory (though that should be rare the case).

DedeHai:
Ok, I think I understand how it's done. I just tested overlapping segments and FX that use additive colors do overlap nicely and FX that use setPixelColor just overwrite the lower numbered segment.
I don't see a good solution that would always work except using a local persistent buffer per segment and then blending the overlapping segments by adding the colors before transferring it to the NPB or global buffer. At least for RGB this would work but since I do not have a full picture on all the variation you already tried this may just be what you had with setUpLeds().

Makuna:
> A global rendering function would have to take care of blending the segments, as described above.

Out of band question.  When you render your buffers are you thinking it would be multiple being blended into a destination?  How many source buffers?  And the destination buffer by something like a GetPixelColor?  I was just experimenting with a two source buffers for a render (NeoPixelBus terms which are similar to PC game graphics terms), and was considering how many sources would be useful.

blazoncek:
FastLED implemented + operator  (and +=) to simplify _blending_ of two colors (it just *add*ed respective RGB channels) so many FastLED examples will exhibit something like:

```c++
CRGB leds[200];
CRGB col1 = CRGB::RED;
CRGB col2 = CRGB::GREEN;
CRGB yellow = col1 + col2;
...
leds[100] = CRGB::BLUE;
leds[100] += yellow;
```

So the + operator was actually `SetPixelColor(x, blend(GetPixelColor(x), somenewcolor);` in NPB terms. Where `blend()` would be a function that somehow combines two RgbxColor objects.

Currently in WLED we don't do any special blending (though I did have a branch that blended two _local_ buffers) as Segments are rendered one after the other and effect functions are all over the place regarding how they treat existing pixels (some will `fill(BLACK)`, etc). As said, mostly FastLED ported effects relied on previous pixel values but not exclusively.

Makuna:
Thanks, enough with my side bar conversations.
JFYI:  In NPB, blending is more flexible in the amounts of each source to blend by supplying a progress value.
```
    static RgbColor LinearBlend(const RgbColor& left, const RgbColor& right, float progress);
   static RgbColor LinearBlend(const RgbColor& left, const RgbColor& right, uint8_t progress);
```
I wonder if that snippet you provided is a bilinear blend?
```
    static RgbColor BilinearBlend(const RgbColor& c00, 
        const RgbColor& c01, 
        const RgbColor& c10, 
        const RgbColor& c11, 
        float x, 
        float y);
```

PR #3920, opened 2024-04-21
Relay open drain output (allow esp32 to drive 5v relay modules)

Suxsem:
This PR adds a new boolean option to the relay settings: open drain.

image

When select, relay pin mode is set to OUTPUT_OPEN_DRAIN instead of normal OUTPUT.

> OUTPUT_OPEN_DRAIN: an open-drain or open-collector output. HIGH (1) leaves the output in high impedance state, LOW (0) pulls the output low. Typically used with an external pull-up resistor to allow any of multiple devices to set the value low safely.

**This allow driving 5v relay modules (cheap super common Arduino ones) just powering them with 5V and adding an external pull-up resistor between the module input pin and 5V, without risking the esp32 pins (that are not 5v-tolerant) and without additional level converters.**

image

Fully tested

### DIFF
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -335,12 +335,16 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
   CJSON(irApplyToAllSelected, hw["ir"]["sel"]);
 
   JsonObject relay = hw[F("relay")];
+
+  if (relay.containsKey("odrain")) {
+    rlyOpenDrain = relay["odrain"];
+  }
   int hw_relay_pin = relay["pin"] | -2;
   if (hw_relay_pin > -2) {
     pinManager.deallocatePin(rlyPin, PinOwner::Relay);
     if (pinManager.allocatePin(hw_relay_pin,true, PinOwner::Relay)) {
       rlyPin = hw_relay_pin;
-      pinMode(rlyPin, OUTPUT);
+      pinMode(rlyPin, rlyOpenDrain ? OUTPUT_OPEN_DRAIN : OUTPUT);
     } else {
       rlyPin = -1;
     }
@@ -868,6 +872,7 @@ void serializeConfig() {
   JsonObject hw_relay = hw.createNestedObject(F("relay"));
   hw_relay["pin"] = rlyPin;
   hw_relay["rev"] = !rlyMde;
+  hw_relay["odrain"] = rlyOpenDrain;
 
   hw[F("baud")] = serialBaud;
 
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -336,9 +336,7 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
 
   JsonObject relay = hw[F("relay")];
 
-  if (relay.containsKey("odrain")) {
-    rlyOpenDrain = relay["odrain"];
-  }
+  rlyOpenDrain  = relay[F("odrain")] | rlyOpenDrain;
   int hw_relay_pin = relay["pin"] | -2;
   if (hw_relay_pin > -2) {
     pinManager.deallocatePin(rlyPin, PinOwner::Relay);
@@ -872,7 +870,7 @@ void serializeConfig() {
   JsonObject hw_relay = hw.createNestedObject(F("relay"));
   hw_relay["pin"] = rlyPin;
   hw_relay["rev"] = !rlyMde;
-  hw_relay["odrain"] = rlyOpenDrain;
+  hw_relay[F("odrain")] = rlyOpenDrain;
 
   hw[F("baud")] = serialBaud;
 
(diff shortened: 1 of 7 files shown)


blazoncek:
I have exactly the same (visually) relays in daily use and never experienced any issues with normal OUTPUT setting.
What was your motive or issue to add open-drain option for the GPIO in use?

Suxsem:
@blazoncek do you power it from 5V or 3.3V?
If 5V, mine just doesn't work with a 3.3v logic input voltage: both LOW and HIGH turns the relay on.
If 3.3V...those relays are supposed to be powered from 5V to reliably turn on. I have 5 of them and 2 of them and when powered with 3.3V often bounces on-off when nudged.

Both problems are fixed when powered with 5V and controlled by a 5V logic level.

blazoncek:
5V.

wled00/cfg.cpp
```diff
@@ -335,12 +335,16 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
   CJSON(irApplyToAllSelected, hw["ir"]["sel"]);
 
   JsonObject relay = hw[F("relay")];
+
+  if (relay.containsKey("odrain")) {
```
blazoncek, line 339: Why not use common ArduinoJson construct `rlyOpenDrain  = relay[F("odrain")] | rlyOpenDrain;` or  `CJSON(rlyOpenDrain, relay[F("odrain")] )`?
It does the same.

wled00/cfg.cpp
```diff
@@ -868,6 +872,7 @@ void serializeConfig() {
   JsonObject hw_relay = hw.createNestedObject(F("relay"));
   hw_relay["pin"] = rlyPin;
   hw_relay["rev"] = !rlyMde;
+  hw_relay["odrain"] = rlyOpenDrain;
```
blazoncek, line 875: use F()

blazoncek (changes requested):
I've checked my old notes and there was a problem with one particular relay (which I attached to pre-release prototype of Dig-Uno) that didn't work. Since then Dig-Unos include level shifter for relays as should any hardware that mixes 3.3V logic with 5V logic.

That does not mean anything though. 😄

wled00/cfg.cpp
Suxsem: I just used the same syntax of existing "rev" setting. Should I change that too?

wled00/cfg.cpp
Suxsem: ok, should I change "pin" and "rev" also?

softhack007:
> When select, relay pin mode is set to OUTPUT_OPEN_DRAIN instead of normal OUTPUT.

Question: does it make sense to combine both inverted/non-inverted with open_drain? If not, then it might be better to have a drop-down with 3 options: normal/inverted/open_drain, instead of two independent checkboxes?

dosipod:
@Suxsem  Very good ,I have a lot of those exact relays and they  would only work if powered from 3.3V ( which as you 
stated is not ideal )  or if a level shifter is added  , tested your PR and they all work now when powered from 5V .

wled00/cfg.cpp
blazoncek: No. Strings that are short and have many repetitions should not be enclosed in F() macro as they are optimised by linker.

blazoncek:
> Question: does it make sense to combine both inverted/non-inverted with open_drain? If not, then it might be better to have a drop-down with 3 options: normal/inverted/open_drain, instead of two independent checkboxes?

That is a very good observation!

wled00/cfg.cpp
blazoncek: That one has a negation included (for backwards compatibility).

Suxsem:
> > When select, relay pin mode is set to OUTPUT_OPEN_DRAIN instead of normal OUTPUT.
> 
> Question: does it make sense to combine both inverted/non-inverted with open_drain? If not, then it might be better to have a drop-down with 3 options: normal/inverted/open_drain, instead of two independent checkboxes?

@softhack007 @blazoncek I think they should be kept separate: all 4 combinations of inverted / open drain are valid. 3.3V relay modules should not use open drain but some may have an active-low logic and others may have active-high logic. 5V relays should use the open drain option with the pull up resistor but some of them uses an NPN transistor and others a PNP one so open-drain and inverted could be useful

blazoncek (approved):
I'm ok with changes.

wled00/data/settings_leds.htm
```diff
@@ -822,7 +823,7 @@ <h3>Hardware setup</h3>
 		Apply IR change to main segment only: <input type="checkbox" name="MSO"><br>
 		<div id="json" style="display:none;">JSON file: <input type="file" name="data" accept=".json"><button type="button" class="sml" onclick="uploadFile('/ir.json')">Upload</button><br></div>
 		<a href="https://kno.wled.ge/interfaces/infrared/" target="_blank">IR info</a><br>
-		Relay GPIO: <input type="number" min="-1" max="48" name="RL" onchange="UI()" class="xs"> Invert <input type="checkbox" name="RM"><span style="cursor: pointer;" onclick="off('RL')">&nbsp;&#x2715;</span><br>
+		Relay GPIO: <input type="number" min="-1" max="48" name="RL" onchange="UI()" class="xs"> Invert <input type="checkbox" name="RM"> Open drain <input type="checkbox" name="RO"><span style="cursor: pointer;" onclick="off('RL')">&nbsp;&#x2715;</span><br>
```
blazoncek, line 826: Can you move the X (span) next to input field and then I'll merge.

dosipod:
Something might be you could also fix if possible is the  multi-relay usermod

wled00/data/settings_leds.htm
Suxsem: @blazoncek sure, done!

PR #3921, opened 2024-04-21
Add GIF Player UI extension (/gifplayer.htm)

Manut38:
I wrote a little UI extension page for the new GIF support and I would love to have it included in this project.

Thanks to @ajotanc, whose [PixelMagicTool](https://github.com/ajotanc/PixelMagicTool) I used as a base for my layout and structure!

I also included a "GIF" button right next to the pxmagic.htm button.\
This is merely an idea and of course I'm happy to take suggestions about the placement or about the implementation in general.

Source: https://github.com/Manut38/WLED-GifPlayer-html

Thank you all so much for your work on WLED, it's honestly amazing!!\
I've been using it for the last couple of years now and I love to see how it developed over that time. 

---

<p align="left">
  
</p>

A WLED UI extension for an easy control of the awesome new "Image" effect, capable of displaying animated GIFs on your LED matrix.

It is using the ESP filesystem editor at `/edit` to list and change files on the local ESP storage.

## Features

- Click to change currently displayed image
- Upload/Delete image files
- Save images as presets
- Forced caching of images and request limiting (ESP crashes when there are too many concurrend HTTP requests)
- Button to refresh all images, useful when an image was replaced with another one of the same name

## Limitations

> [!WARNING]  
> **Only upload images in your matrix native resolution! (e.g. 16x16)**
> 
> Due to the limited flash storage space and processing power, .gif files of larger dimensions will make the effect lag and kill WLED performance

- Currently the "Image" effect only supports .gif files
- All image files will be uploaded to/read from the root path of WLEDs LittleFS because it does not support subfolders
- SD card support untested

## Screenshots

Screenshot Image List
Screenshot Image Select

### DIFF
--- a/wled00/data/gifplayer/gifplayer.htm
+++ b/wled00/data/gifplayer/gifplayer.htm
@@ -1,0 +1,1132 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>GIF Player</title>
+    <style>
+      :root {
+        --s-thumb: #0006;
+        --s-background: #0003;
+        --overlay: rgba(0, 0, 0, 0.5);
+        --background: #111;
+        --text: #ddd;
+        --gray-dark: #222;
+        --gray-medium: #333;
+        --gray-light: #eeeeee;
+        --blue-dark: #284352;
+        --blue-medium: #3b708b;
+        --blue-light: #48a;
+        --success-dark: #03440c;
+        --success-medium: #548c2f;
+        --success-light: #8ac926;
+        --error-dark: #a80a0a;
+        --error-medium: #c9262b;
+        --error-light: #ff595e;
+        --warning-dark: #dc2f02;
+        --warning-medium: #e85d04;
+        --warning-light: #f48c06;
+      }
+
+      ::-webkit-scrollbar {
+        width: 6px;
+      }
+
+      ::-webkit-scrollbar-track {
+        background: transparent;
+      }
+
+      ::-webkit-scrollbar-thumb {
+        background: var(--s-thumb);
+        opacity: 0.2;
+        border-radius: 5px;
+      }
+
+      ::-webkit-scrollbar-thumb:hover {
+        background: var(--s-background);
+      }
+
+      ::selection {
+        background: var(--blue-light);
+      }
+
+      * {
+        font-family: Helvetica, Verdana, sans-serif;
+        box-sizing: border-box;
+        margin: 0;
+        padding: 0;
+      }
+
+      body {
+        margin: 0;
+        display: flex;
+        justify-content: center;
+        align-items: flex-start;
+        min-height: 100vh;
+        background: var(--background);
+        overflow: scroll;
+      }
+
+      footer {
+        display: flex;
+        justify-content: center;
+        margin-block: 12px;
+        padding-inline: 8px;
+        font-size: 0.75rem;
+        color: var(--text);
+        font-weight: 600;
+      }
+
+      a {
+        text-decoration: none;
+        color: var(--blue-light);
+      }
+
+      a:is(:hover, :focus, :active) {
+        color: var(--blue-medium);
+      }
+
+      .container {
+        width: 100%;
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        flex-direction: column;
+      }
+
+      .content {
+        width: min(730px, calc(100% - 40px));
+        margin-inline: 20px;
+      }
+
+      .row {
+        display: flex;
+        flex-wrap: nowrap;
+        justify-content: space-between;
+        align-items: flex-end;
+      }
+
+      .row-dialog-button {
+        display: flex;
+        justify-content: center;
+        margin-block: 10px;
+      }
+
+      .col {
+        flex-basis: calc(50% - 10px);
+        position: relative;
+        padding-inline: 5px;
+      }
+
+      .col-full {
+        flex-basis: 100%;
+        position: relative;
+        padding-inline: 5px;
+      }
+
+      .col-small {
+        flex-basis: 25%;
+        position: relative;
+        padding-inline: 5px;
+      }
+
+      .col-medium {
+        flex-basis: 60%;
+        position: relative;
+        padding-inline: 5px;
+      }
+
+      .col-dialog-button {
+        min-width: 180px;
+      }
+
+      .header {
+        position: relative;
+        display: flex;
+        flex-direction: column;
+        padding-block: 20px;
+      }
+
+      .header .title {
+        font-size: 2.5rem;
+        line-height: 2.5rem;
+        font-weight: 800;
+        color: var(--gray-light);
+        text-align: center;
+        text-wrap: wrap;
+      }
+
+      .header .title .title-wled {
+        image-rendering: pixelated;
+        image-rendering: crisp-edges;
+        height: 1.8rem;
+        margin-inline: 15px;
+      }
+
+      .header .rainbow {
+        background: linear-gradient(
+          to right,
+          #ef5350,
+          #f48fb1,
+          #7e57c2,
+          #2196f3,
+          #26c6da,
+          #43a047,
+          #eeff41,
+          #f9a825,
+          #ff5722
+        );
+        background-clip: text;
+        -webkit-background-clip: text;
+        -webkit-text-fill-color: transparent;
+        background-size: 200% 200%;
+        animation: rainbow 5s linear infinite;
+        transition: background-position 0.5s ease;
+      }
+
+      .text-info {
+        font-size: 0.8rem;
+        color: var(--text);
+      }
+
+      .image-list-info {
+        margin-top: 5px;
+        text-align: center;
+      }
+
+      label {
+        display: flex;
+        margin-bottom: 5px;
+        font-weight: bold;
+        color: var(--text);
+        align-items: center;
+      }
+
+      .margin-y {
+        margin-top: 20px;
+      }
+
+      input[type="text"],
+      input[type="number"],
+      select,
+      textarea {
+        width: 100%;
+        padding: 10px;
+        border-radius: 50px;
+        background-color: var(--gray-medium);
+        border: 1px solid var(--gray-medium);
+        outline: none;
+        color: var(--gray-light);
+        font-size: 0.875rem;
+        text-align: center;
+      }
+
+      .dropzone {
+        border: 1px dashed var(--gray-light);
+        background-color: var(--gray-dark);
+        color: var(--gray-light);
+        text-align: center;
+        padding: 40px 10px;
+        border-radius: 8px;
+        transition: all 0.5s ease-in-out;
+        margin-inline: 5px;
+      }
+
+      .dropzone:hover {
+        cursor: pointer;
+        color: var(--gray-dark);
+        background-color: var(--gray-light);
+        border-color: var(--gray-dark);
+      }
+
+      .dropzone.dragover {
+        background-color: var(--gray-medium);
+      }
+
+      #toast-container {
+        position: fixed;
+        bottom: 20px;
+        right: 20px;
+        z-index: 9999;
+      }
+
+      .toast {
+        display: flex;
+        align-items: center;
+        width: auto;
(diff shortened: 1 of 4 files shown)


dosipod:
Works okay with one note ( may be not related directly to the html page ) is that the storage size  reported of 241/983 KB is not accurate and it will only report the correct number once you reboot the esp which is the same behavior on the filesystem  info page  .

softhack007:
Does your PR build on top of the PR from Aircoookie for GIF support, or is it independent?

* https://github.com/Aircoookie/WLED/pull/3835

Manut38:
@dosipod Yes, I did notice that behavior too. I'm just getting this info straight from the /json/info endpoint. I don't know of any other way to check for the current values.

@softhack007 This is essentially a separate interface, specialized in controlling the new "Image" effect, added by Aircoookie in the gif branch. So it is built on top of that functionality.\
Right now it is using the FX ID 53 for sending the requests.
`const FX_MODE_IMAGE = 53; // set effect mode "Image"`
This needs to be adjusted, if the ID is changed before the gif branch is merged to main

Manut38:
I made a few changes to the code since I took the screenshots that I would love to hear your opinion on.

1. I added a **pause** and **stop** button to the UI
My reasoning for this is that it allows for a much faster control the displayed image, without having to go back to the main WLED UI. My intention for this page is to basically build an clean interface specialized for the new "Image" effect, so I felt that it made sense to include this functionality; Especially if displaying images is the primary use of your LED matrix.
2. The image buttons are now aware of already existing presets that hold the same image filename in their segment config. Users can now e.g. customize the playback speed of the gif using a preset and still expect GifPlayer to load the correct preset.

Another thing I want to note is that I am very aware that this is building on top of a feature that is still currently in-development. Especially because of that I would be interested in @Aircoookie's opinion, because as far as I know he was the one mostly working on the "Image" effect.
I could also see how adding more and more isolated htmls (pixart.htm, pxmagic.htm, gifplayer.htm) might not be the way to go forward with this, however I'm very curious about your all of your vision here.

Manut38:
Updated the images to reflect the changes to the button layout.
Any news on this? 😄

dosipod:
@Manut38 I am personally not in favor of adding such external  tools to wled  at build time just from the points that future updates to wled might break things( this has been seen with other tools a lot )  and if your tool is separate then it could be updated faster which we also seen first hand . Does not harm to maintain your page though , please post about the same on wled discord 2D channel 

At the moment we use a lot of tools like that and we just download the latest version and copy under edit , i am not sure what  is wrong with that if any    . That is beside the fact that gif is not merged yet

Manut38:
@dosipod I kind of agree with this honestly. This pull request was motivated by the fact that there are some other tools that have been integrated into the source directly and also got mentioned on the official wiki.
Of course this way additional tools like this would get a lot more exposure and users.

I saw that the "older" [WLED-PixelArtConverter](https://github.com/werkstrom/WLED-PixelArtConverter/) [pixartmin.htm] got a separate entry in the wiki at https://kno.wled.ge/features/pixel-art-converter/
Maybe this could be a way to present some links to tools like this on the official wiki?
Because I remember that I found these extensions pretty much randomly on GitHub rather than on an official post.
What is your take on this?
Thanks for all the amazing work btw 👍

blazoncek:
@Manut38 when (or if) GIF player gets merged and @Aircoookie decides that this UI extension is usable and merges this PR you can make a PR for WLED documentation so that it gets attention of users.

ajotanc:
The idea was really good, huh? hahah! I'll test it and let you know what I think, I'm out of date... I didn't even know it had gif support already. It looks very promising.

If possible @Manut38 , can I adapt what you did to my project? Looking at yours gave some very interesting ideas!

Manut38:
@ajotanc Oh I'm sorry, I missed this reply! 
Yeah, your idea definitely was great, and I've been using your solution before I found the on-the-fly decoder that they wrote here.
Thank you a lot for your pioneering on this!

However I'm pretty sure, that whenever the gif branch will be merged it will be the more versatile and faster solution for animations.
I'm honestly still super stoked about this 😁

Of course you can take any inspiration you like from my code, it's open source after all and it's also the exact same thing I did with your code before 😄

wled00/wled.h
```diff
@@ -81,6 +81,8 @@
     #define WIFI_MODE_AP WIFI_AP
     #include <QuickEspNow.h>
   #endif
+  #undef  WLED_DISABLE_GIF
+  #define WLED_DISABLE_GIF
```
netmindz, line 85: please remove

wled00/fcn_declare.h
```diff
@@ -257,7 +257,6 @@ void handleIR();
 #include "ESPAsyncWebServer.h"
 #include "src/dependencies/json/ArduinoJson-v6.h"
 #include "src/dependencies/json/AsyncJson-v6.h"
-#include "FX.h"
```
netmindz, line 260: please revert this change

wled00/wled_server.cpp
```diff
@@ -9,6 +9,9 @@
 #ifndef WLED_DISABLE_PXMAGIC
   #include "html_pxmagic.h"
 #endif
+#ifndef WLED_DISABLE_GIFPLAYER
```
netmindz, line 12: Please wrap with a check for WLED_ENABLE_GIF is defined as no point having the player if you don't have GIF  support

DedeHai:
this is a nice addition to the gif player. I looked at flash usage: this UI uses more flash than the gif player itself... just wondering if it can be slimmed down a little by not using all those fancy graphics.

Manut38:
Thank you for your considerations to add this!
I'll be quite busy for the rest of the week, but I'll promise I'm going to take a look at the proposed changes as soon as I can.

DedeHai:
IMHO this addition makes the gif player much more user friendly but if it is added by default it must be as slim as possible, ideally 1k or less of flash.
Just an idea: have the most basic version possible built in and if internet is available, load an extended version from git repo?

wled00/wled.h
manuelfocus: I'm not sure if you meant me with this, as it's not part of the extension.

wled00/fcn_declare.h
manuelfocus: Again, I'm not sure what you meant by that

manuelfocus:
I have made a few smaller changes that could be incorporated into a potential release.
@netmindz check if you want to update your fork for MM as well :)

wled00/wled.h
willmmiles: Your PR includes a commit that adds these lines, which - as you note - are unrelated to the extension you're contributing.  Please add a commit that undoes this change (and the other like it), so the final merge doesn't include any of these unnecessary or unrelated changes.

manuelfocus:
@DedeHai 
In the beginning I thought the same, but then I realized that the other existing UI mods are also somewhere around 5-10 KB and ultimately it will be the actual gifs that you will upload using it that fill up the flash.
I'm running around 80 gifs on my 16x16 matrix without problem and they're all 2-10 KB each.
With that I'm at 536 KB / 983 KB on a common ESP32 WROOM DevKit.

manuelfocus:
The biggest issue I see right now is actually the performance of the webserver. I intentionally let the client load every image sequentially because the whole ESP crashes whenever multiple requests are made at the same time.

Not sure if this is something that is known or will be fixed. I think it's because of the way AsyncTCP and LittleFS play together.

willmmiles:
> The biggest issue I see right now is actually the performance of the webserver. I intentionally let the client load every image sequentially because the whole ESP crashes whenever multiple requests are made at the same time.
> 
> Not sure if this is something that is known or will be fixed. I think it's because of the way AsyncTCP and LittleFS play together.

You will probably need some kind of limit in practice, though the device shouldn't crash -- it's expected to return 503s if overloaded, and I believe that the filesystem can trigger sending 404s if the file handles are exhausted, though I haven't chased that down yet.  Under extreme pressure, connections may also just be dropped.  If you've got a repeatable crash, the stack traces (and information about what specific board you're testing with) would be very helpful.

wled00/wled.h
manuelfocus: @willmmiles Ah I see, but these came from the merge of netmindz. I did not introduce them, but I can remove them.

DedeHai:
> ultimately it will be the actual gifs that you will upload using it that fill up the flash.

different partition. if you manage to load this into the filesystem by default, I agree.

wled00/data/gifplayer/gifplayer.htm
```diff
@@ -0,0 +1,1341 @@
...
+        </div>
+      </div>
+    </dialog>
+    <dialog id="settings-dialog">
+      <div class="dialog-content">
+        <h3 id="">Settings</h3>
+        <form id="formHostname" novalidate>
+          <div class="row-dialog-button">
+            <div class="col-dialog-button" validate style="position: relative">
+              <label for="hostname">Hostname</label>
```
netmindz: We don't need configuration of the hostname if we are serving directly from the device

wled00/data/gifplayer/gifplayer.htm
```diff
@@ -0,0 +1,1341 @@
...
+    const GIF_SPEED = 128; // set effect speed to 128, original gif speed
+    const ALLOWED_EXT = ["gif", "bmp", "png"];
+    const urlEdit = "/edit";
+    const urlJsonState = "/json/state";
+    const urlJsonInfo = "/json/info";
+    const urlPresets = "/presets.json";
+
+    // dynamic variables
+    const d = document;
+    const hostname = gId("hostname");
```
netmindz: no need for configurable host

netmindz (changes requested):
If you could please remove the redundant hostname configuration

wled00/data/gifplayer/gifplayer.htm
manuelfocus: I agree, this is only useful, if you want to use the static .htm to control a remote WLED instance.
I will remove it for the included version.

manuelfocus:
I'm currently in the process of a small refactoring, but I'm having an issue with something in the code that messes up the included cdata.js minifier and leads to a blank page when served from WLED directly.
I'll update the code as soon as possible and incorporate your changes.

netmindz:
One thought, is is possible to refactor this code so it shares rather than duplicating pxmagic code or possibly even combine into one?

manuelfocus:
@netmindz I have addressed your comments by removing the settings dialog altogether and moving the button to clear cache to the footer.
I am satisfied with the state of the UI now.

I don't think combining this with pxmagic makes a lot of sense at this point as the only code they are sharing are a few css blocks. Either way it would not give us a lot of benefits regarding the file size.
The current revision gets compressed into 6954 bytes by the minfier, which I think is not too bad. 
Anyway I still understand the argument that these "external tools" might not belong into the main build at all, but of course I'm not the one in charge of making a decision about that. 

Another thing I have noticed is that currently there is quite a strong stutter when crossfading from a "traditional" effect mode to the "image" effect. I'm not sure if this has been addressed before somewhere.

manuelfocus:
@willmmiles 
I can still replicate the aforementioned issue with the web server.
Every time too many requests are made (e.g. loading the images from the esp) the web server code seems to be running into an indefinite loop.

The ESP does not crash (no crash dump) and I was unable to get any useful information out of a my debug build using WLED_DEBUG and DEBUG_ESP_HTTP_SERVER.
As this happens somewhat regularly when reloading the GIF player too quickly, it would be helpful to fix what is causing that to happen.

I have already put in a limit of the max concurrent request made during image fetching.
The current value is 2, but even then it still happens occasionally.
I can't do much more from the UI side as it just stops responding, not even returning an 404/50x to me.

netmindz:
Has this now been superceded @DedeHai ?

DedeHai:
yes, the pixel forge gif tool is very loosely based on work in this PR but packs many more features at less than half the size (IIRC this PR uses about 6k, the gif tool from pixelforge is less than 3k)

netmindz:
Thank you for your submission @Manut38 

We won't be merging your tool, but it has been an inspiration for what is now included in WLED so your efforts were not in vain and certainly useful for helping to test the gif playback feature

manuelfocus:
Thank for the recognition!
I agree, the new tool has a lot more very useful features.
I yet have to try it out, and after that I will decide if I will keep maintaining my version 😆

PR #3925, opened 2024-04-22
Reserve "FOSS" product name for official builds

jamesmyatt:
If the "FOSS" product name is reserved for official builds, then it becomes much easier to identify devices that have official builds from the Aircoookie repository.

* WLED_BRAND_NAME = "WLED"  --> Implies it uses the Aircoookie version of the source code. Forks should use a different value.
* WLED_PRODUCT_NAME = "FOSS" --> Implies it's an official build. Unofficial ones should use a different one. The default for custom builds is now "DIY".

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -12,6 +12,11 @@
 # CI/release binaries
 default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, esp32dev, esp32_eth, esp32dev_audioreactive, lolin_s2_mini, esp32c3dev, esp32s3dev_8MB, esp32s3dev_8MB_PSRAM_opi, esp32s3_4M_PSRAM_qspi, esp32_wrover
 
+# Note that WLED_BRAND_NAME='"WLED"' identifies the source code as coming from the Aircoookie repository. Forks should use a different value.
+# Official builds from this repository have WLED_PRODUCT_NAME='"FOSS"'. Unofficial builds can have other values.
+# Note that third-party components may use these values in order to OTA update your device's firmware using these builds,
+# overwriting any custom firmware that you have loaded.
+
 src_dir  = ./wled00
 data_dir = ./wled00/data
 build_cache_dir = ~/.buildcache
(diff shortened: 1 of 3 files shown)


blazoncek:
IMO that's as good as nothing. It does not change anything on forked repositories as anyone is discouraged to modify `platformio.ini` and use `platformio_override.ini` instead. So the outcome will be the same regardless of which branch is used in such repositories when using default environment builds.

And if a user/vendor chooses `platformio_override.ini` there are still no guarantees that he/she doesn't copy entire PIO environment.

jamesmyatt:
As you've said before, you can't control the behaviour of people who make forks, especially with an open hobby project like WLED, so there are no guarantees ever. But you can make the default behaviour follow what you would like, give them the hooks to do do the same and have consequences if they don't. Here, the consequences are that third-parties like Home Assistant might try to OTA your device with the official version of the firmware.

As an initial step, this PR changes the default behaviour to be that the product name is different depending on whether you use a build config from the `platformio.ini` or from the `platformio_override.ini`, which is not nothing.

It also asks people nicely to change WLED_BRAND_NAME (in `const.h`) if they make any changes to the source code. Which is also not nothing.

platformio.ini
```diff
@@ -12,6 +12,11 @@
 # CI/release binaries
 default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, esp32dev, esp32_eth, esp32dev_audioreactive, lolin_s2_mini, esp32c3dev, esp32s3dev_8MB, esp32s3dev_8MB_PSRAM_opi, esp32s3_4M_PSRAM_qspi, esp32_wrover
 
+# Note that WLED_BRAND_NAME='"WLED"' identifies the source code as coming from the Aircoookie repository. Forks should use a different value.
+# Official builds from this repository have WLED_PRODUCT_NAME='"FOSS"'. Unofficial builds can have other values.
+# Note that third-party components may use these values in order to OTA update your device's firmware using these builds,
+# overwriting any custom firmware that you have loaded.
```
jamesmyatt, line 18: There's almost certainly a better place to put this text. Where is that?

jamesmyatt:
Ah. That's not how MoonModules works at all! I was misled by looking at the wrong source code file.

[`json.cpp`](https://github.com/MoonModules/WLED/blob/mdev/wled00/json.cpp#L1158) suggests that they use "brand" = "WLEDMM" or similar, but they actually use "WLED" and have the "product" set to ["MoonModules"](https://github.com/MoonModules/WLED/blob/bf6932ecb79e8fdab73505dddd147f442121afb8/wled00/wled.h#L31) instead.

So problem 1 has to be solved with a combination of "brand" and "product" together, rather than just "brand".

PR #3933, opened 2024-04-26
Improved brightness change via long button presses

gaaat98:
I made a small enhancement on how the brightness is handled when long pressing physical buttons. 
Instead of looping from 0 to 255 and using the byte overflow to return to 0, we now start by increasing the brightness and maxing out at 255. To decrease the brightness, the user must release the button and long press it again; it will gradually decrease to 1 and stay there.
The next long press will increase the brightness again, and so on; this allows for a better selection of brightness (with the previous version, I found it difficult to set to extreme values such as 255 or 1). I also slightly decreased the time between repeated actions to make dimming quicker; a cool addition could be making it proportional to the transition time.

The direction is shared between buttons because it made more sense to me, but it can be easily changed.

I copied the behaviour from the dimmable lights that I encountered where I live, I hope it makes sense to other people!

### DIFF
--- a/wled00/button.cpp
+++ b/wled00/button.cpp
@@ -10,7 +10,7 @@
 #define WLED_LONG_REPEATED_ACTION   400 // how often a repeated action (e.g. dimming) is fired on long press on button IDs >0
 #define WLED_LONG_AP               5000 // how long button 0 needs to be held to activate WLED-AP
 #define WLED_LONG_FACTORY_RESET   10000 // how long button 0 needs to be held to trigger a factory reset
-#define WLED_LONG_BRI_STEPS          16 // how long to wait before increasing/decreasing brightness on long press
+#define WLED_LONG_BRI_STEPS          16 // how much to increase/decrease the brightness with each long press repetition
 
 static const char _mqtt_topic_button[] PROGMEM = "%s/button/%d";  // optimize flash usage
 static bool buttonBriDirection = false; // true: increase brightness, false: decrease brightness
--- a/wled00/button.cpp
+++ b/wled00/button.cpp
@@ -13,6 +13,7 @@
 #define WLED_LONG_BRI_STEPS          16 // how long to wait before increasing/decreasing brightness on long press
 
 static const char _mqtt_topic_button[] PROGMEM = "%s/button/%d";  // optimize flash usage
+static bool buttonBriDirection = false; // true: increase brightness, false: decrease brightness
 
 void shortPressAction(uint8_t b)
 {
@@ -41,7 +42,6 @@ void longPressAction(uint8_t b)
     switch (b) {
       case 0: setRandomColor(col); colorUpdated(CALL_MODE_BUTTON); break;
       case 1: 
-        // increase bri on true, decrease on false
         if(buttonBriDirection) {
           if (bri == 255) break; // avoid unnecessary updates to brightness
           if (bri >= 255 - WLED_LONG_BRI_STEPS) bri = 255;
--- a/wled00/button.cpp
+++ b/wled00/button.cpp
@@ -7,9 +7,10 @@
 #define WLED_DEBOUNCE_THRESHOLD      50 // only consider button input of at least 50ms as valid (debouncing)
 #define WLED_LONG_PRESS             600 // long press if button is released after held for at least 600ms
 #define WLED_DOUBLE_PRESS           350 // double press if another press within 350ms after a short press
-#define WLED_LONG_REPEATED_ACTION   300 // how often a repeated action (e.g. dimming) is fired on long press on button IDs >0
+#define WLED_LONG_REPEATED_ACTION   400 // how often a repeated action (e.g. dimming) is fired on long press on button IDs >0
 #define WLED_LONG_AP               5000 // how long button 0 needs to be held to activate WLED-AP
 #define WLED_LONG_FACTORY_RESET   10000 // how long button 0 needs to be held to trigger a factory reset
+#define WLED_LONG_BRI_STEPS          16 // how long to wait before increasing/decreasing brightness on long press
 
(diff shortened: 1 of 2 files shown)


blazoncek:
Can you make `buttonBriDirection` local static instead of global?
Next time please use different branch so I can more easily test.

gaaat98:
I can move ```buttonBriDirection``` in button.cpp and remove the WLED_GLOBAL but it would still need to have global scope since I need to access it from both ```longPressAction``` and ```handleButton```. 
The other possible solution I thought of would be declaring it in ```handleButton``` as static but then I would have to change  ```void longPressAction(uint8_t b)``` to ```void longPressAction(uint8_t b, bool direction)```. 
Which solution is preferred?


> Next time please use different branch so I can more easily test.
Sorry about that!

wled00/button.cpp
```diff
@@ -7,11 +7,13 @@
 #define WLED_DEBOUNCE_THRESHOLD      50 // only consider button input of at least 50ms as valid (debouncing)
 #define WLED_LONG_PRESS             600 // long press if button is released after held for at least 600ms
 #define WLED_DOUBLE_PRESS           350 // double press if another press within 350ms after a short press
-#define WLED_LONG_REPEATED_ACTION   300 // how often a repeated action (e.g. dimming) is fired on long press on button IDs >0
+#define WLED_LONG_REPEATED_ACTION   400 // how often a repeated action (e.g. dimming) is fired on long press on button IDs >0
```
blazoncek, line 10: Why did you change this? This is supposed to be ms.

wled00/button.cpp
```diff
@@ -7,11 +7,13 @@
...
 #define WLED_LONG_AP               5000 // how long button 0 needs to be held to activate WLED-AP
 #define WLED_LONG_FACTORY_RESET   10000 // how long button 0 needs to be held to trigger a factory reset
+#define WLED_LONG_BRI_STEPS          16 // how long to wait before increasing/decreasing brightness on long press
```
blazoncek, line 13: That is a large step. Perhaps take a look at `ir.cpp` how it is implemented there.
IMO, if you keep it compile time dependant, it would be better if this would be lower (i.e. 8) to allow more gradual brightness.

I also find comment misleading. It has nothing to do with "time".

wled00/button.cpp
```diff
@@ -284,10 +298,12 @@ void handleButton()
...
       if (now - buttonPressedTime[b] > WLED_LONG_PRESS) { //long press
-        if (!buttonLongPressed[b]) longPressAction(b);
-        else if (b) { //repeatable action (~3 times per s) on button > 0
+        if (!buttonLongPressed[b]) {
+          buttonBriDirection = !buttonBriDirection; //toggle brightness direction on long press
+          longPressAction(b);
+        } else if (b) { //repeatable action (~5 times per s) on button > 0
           longPressAction(b);
-          buttonPressedTime[b] = now - WLED_LONG_REPEATED_ACTION; //333ms
+          buttonPressedTime[b] = now - WLED_LONG_REPEATED_ACTION; //200ms
```
blazoncek, line 306: According to #define that is 400ms.

blazoncek (commented):
Can you please comment on the decisions made?

wled00/button.cpp
gaaat98: The rationale behind this choice is that the higher the value the greater the number of repetitions per seconds, WLED_LONG_REPEATED_ACTION does not actually mean "repeat the action every XXX milliseconds", I don't know how to explain with words so I'll make this example:
- When WLED_LONG_REPEATED_ACTION = 300, suppose now = 1000:
```buttonPressedTime[b] = now - WLED_LONG_REPEATED_ACTION ----> buttonPressedTime[b] = 700```
next call of handleButton we assume now = 1201
```if (now - buttonPressedTime[b] > WLED_LONG_PRESS) ----> if(1201 - 700 > 600) ----> if(501 > 600)``` ---> repeated action DOES NOT trigger


- When WLED_LONG_REPEATED_ACTION = 400, suppose now = 1000:
```buttonPressedTime[b] = now - WLED_LONG_REPEATED_ACTION ----> buttonPressedTime[b] = 600```
next call of handleButton we assume now = 1201
```if (now - buttonPressedTime[b] > WLED_LONG_PRESS) ----> if(1201 - 600 > 600) ----> if(601 > 600)``` ---> repeated action DOES trigger

The greater the value of ```WLED_LONG_REPEATED_ACTION``` is, the greater we set ```buttonPressedTime[b]``` back in time, therefore increasing the number of repetitions per second, with the actual delay between repetitions being ```WLED_LONG_PRESS - WLED_LONG_REPEATED_ACTION``` 
(now - buttonPressedTime[b] > WLED_LONG_PRESS equals to 
now_new - now_old + WLED_LONG_REPEATED_ACTION > WLED_LONG_PRESS that is 
now_new - now_old > WLED_LONG_PRESS - WLED_LONG_PRESS)

Hope I did not make myself a fool by totally misunderstanding the code 😅

wled00/button.cpp
gaaat98: I chose to double the step since I found the previous stepping to be too slow, and this with the new timing of the action was a good trade off for me, 16 different levels of brightness are more than enough (IMHO) when controlling through a physical button, as a user I'd rather go from 1 to max quicker. 
But obviously it can be reverted to 8!

I agree, the comment is wrong and misleading, I must have turned my brain off when writing it. Will fix it asap

wled00/button.cpp
gaaat98: See comment below

wled00/button.cpp
blazoncek: Ok, IDK who implemented this (will need to check git blame, I hope it was not me) but it is un-intuitive. 😄 It looks like it is roughly 1000/WLED_LONG_REPEAT_ACTION*100 ms.

PR #3940, opened 2024-04-29
WIP: enhance Serial API JSON

JonnyTech:
The Serial API for JSON currently does not support writing to cfg.json - this is an attempt to add it with code taken from wled_server.cpp - feedback and guidance is needed please because as I am non-developer there is something wrong with my implementation that I cannot figure out.

### DIFF
--- a/wled00/wled_serial.cpp
+++ b/wled00/wled_serial.cpp
@@ -123,7 +123,10 @@ void handleSerial()
             releaseJSONBufferLock();
             return;
           }
-          verboseResponse = deserializeState(pDoc->as<JsonObject>());
+          JsonObject root = pDoc->as<JsonObject>();
+          verboseResponse = deserializeState(root);
+          //if JSON contains networking key then initiate saving of config
+          if (root.containsKey("nw")) doSerializeConfig = true;
           //only send response if TX pin is unused for other purposes
           if (verboseResponse && (!pinManager.isPinAllocated(hardwareTX) || pinManager.getPinOwner(hardwareTX) == PinOwner::DebugOut)) {
             pDoc->clear();


JonnyTech:
Hello. I am still trying to figure this out. Are there any developers who can provide some input, please. FYI there is a discourse chat too: https://wled.discourse.group/t/11728 if that is easier, and there is interest from at least another user here https://wled.discourse.group/t/11683/16/ - thank you in advance.

JonnyTech:
Is anybody able to help or comment, please? This is currently a showstopper for me to use WLED.

blazoncek:
Why don't you use Improv? It is used to configure WiFi and may be repurposed to configure Ethernet.

BTW I do not use JSON over serial so can't help you there.

joosthoi1:
I think this should be tackled and should be expanded. At the moment, you are missing alot of important features with json over serial. All get routes (https://kno.wled.ge/interfaces/json-api/#api) are unavailable except State, Info, Effects and Palettes. (By setting "v": true in a post request). And you can't choose to get only some of this data.

Furthermore (And this seems an issue for the normal json api aswell), presets aren't exposed by the json api and are instead requested by calling /presets.json, making a change can then be done by settings the current state to something and posting psave to the json api. This seems clunky to me and you cant create a preset without altering the state, like you can with playlists. It also sets n (name), ib, sb and sc (various booleans) Which seem undocumented. Now, presets.json Can't seem to be called from json over serial, which seems like a clear oversight. Why would I want to cycle to a different preset when I don't even know which presets exists. Idem for creating playlists

blazoncek:
> json api is extremely coupled with a request object

AFAIK it is not. Can you show an example?

As for serial: It is a stream not a packet. So you need to address it as such and all is good.

Issue #3937, opened 2024-04-27
Audio reactive effects for PWM RGB Strips


gaaat98:
Hi! I was trying to replace some cheap TUYA RGB PWM controllers (specifically [this](https://aliexpress.com/item/1005001621143902.html)) with Audioreactive WLED, but after spending the evening wiring everything up I realized that the audioreactive effects are only available for addressable strips (as also confirmed by #3494).

My question/feature request now is: it's possible to implement some simple effects such as changing brightness/color/effect speed based on sound? The cheap Tuya controller did it, and I think also WLED should be able to do it!

If I were to implement these effects myself, where should I start?

Thanks in advance

dosipod:
@gaaat98  I do not use any analog strips but a  trick we used for a long time to test is just create fake w281x output ( even setup a matrix if you like 2D audio effects ) . If you do that then you would have access to those effects in ui . Not exactly sure how good those effects would be on your setup but I just tested that with esp32 onboard led and it kinda works with audio effects. Better then nothing :)

gaaat98:
> @gaaat98 I do not use any analog strips but a trick we used for a long time to test is just create fake w281x output ( even setup a matrix if you like 2D audio effects ) . If you do that then you would have access to those effects in ui . Not exactly sure how good those effects would be on your setup but I just tested that with esp32 onboard led and it kinda works with audio effects. Better then nothing :)

Thanks, this seems to be working also with my analog strips! I wonder if there's a way to turn this workaround into an actual feature

blazoncek:
If effect works with a single pixel it should expose capability using metadata. That's all.

gaaat98:
So would it be welcome if I updated the metadata of the sound reactive effects that still do something "cool" on single pixels despite being designed for addressable strips or is it better not to?

blazoncek:
Just be conservative and it will be ok. 😉

PR #3942, opened 2024-04-29
Enabled some audioreactive effects for single pixel strips/segments

gaaat98:
"Solution" to #3937 
Enabled Juggles, Noisefire, Plasmoid, DJ Lights, Freqmatrix, Freqwave, Noisemove, Rocktaves, and Waterfall effects (I chose the ones that looked the best to me) to be selected for single pixel segments/LED strips.

Some effects (noisefire, plasmoid) needed just the metadata to be updated, the others also needed the ```if (SEGLEN == 1) return mode_static();``` check to be disabled, I verified that removing the check does not create problems in the functions (hopefully 😅, see comments in code).

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -6643,7 +6643,7 @@ static const char _data_FX_MODE_GRAVIMETER[] PROGMEM = "Gravimeter@Rate of fall,
 //   * JUGGLES      //
 //////////////////////
 uint16_t mode_juggles(void) {                   // Juggles. By Andrew Tuline.
-  if (SEGLEN == 1) return mode_static();
+  //if (SEGLEN == 1) return mode_static();
   um_data_t *um_data;
   if (!usermods.getUMData(&um_data, USERMOD_ID_AUDIOREACTIVE)) {
     // add support for no audio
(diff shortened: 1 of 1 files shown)


blazoncek:
I think it is ok, but let's wait what @softhack007 says.

blazoncek:
@softhack007 could you look into this, please. IMO the changes are benign and I'd like to include them in b3 release scheduled for next week.

softhack007:
> @softhack007 could you look into this, please. IMO the changes are benign and I'd like to include them in b3 release scheduled for next week.

I'm on it 😃 

overall looks good, thanks @gaaat98 for doing the "zero D" checks. I'll try them on my own setup again, plus review the code to be sure there is no danger of running into invalid pointers or div/0 situations.

wled00/FX.cpp
```diff
@@ -6643,7 +6643,7 @@ static const char _data_FX_MODE_GRAVIMETER[] PROGMEM = "Gravimeter@Rate of fall,
 //   * JUGGLES      //
 //////////////////////
 uint16_t mode_juggles(void) {                   // Juggles. By Andrew Tuline.
-  if (SEGLEN == 1) return mode_static();
+  //if (SEGLEN == 1) return mode_static();
```
softhack007, line 6646: Please delete these commented-out checks. No need to keep them in the code - github always has the history.

softhack007 (changes requested):
Thanks, looks good :-)

Please just delete the commented-out checks like `if (SEGLEN == 1) return mode_static();`.

wled00/FX.cpp
gaaat98: Done!

softhack007 (approved):
👍

PR #2736, opened 2022-07-30
Usermod: Analog clock

lordjaxom:
I have moved the analog clock overlay functionality to a user mod and added following features:

- Colors of hours, minutes and seconds are configurable via Usermod Settings
- Possibility to have different LEDs display the seconds than the hours and minutes
- Can turn off seconds altogether
- Switch to turn on blending / mixing the colors of the pointers with each other or existing effects
- Fading effect for the seconds pointer
- Runs much smoother due to frequent updates even if there is no effect running

If desired, I could also remove the now obsolete "analog clock overlay" from the base code.

PR #3943, opened 2024-04-30
Analog clock 2d branch0 15

bandi13:
Added a 2D Analog Clock effect. It needed some antialiased lines and circles, so I've implemented that as well. On my 48x48 pixel display it looks like this: 
IMG_20240430_001047063_HDR

Also cleaned up some function naming to be more consistent with the majority of the source code.

## Summary by CodeRabbit

* **New Features**
  * Introduced a 2D analog clock visual effect for matrix setups, displaying hour, minute, and optional second hands with customizable colors.
* **Other Changes**
  * The new 2D Analog Clock effect replaces the previous Halloween effect at effect ID 53.


blazoncek:
Please revert all unnecessary changes (renamed functions).

blazoncek:
You are not requested to change existing code unless necessary.
There is no need to change existing functions.

softhack007:
Hi, thanks for taking the time to add this nice new effect:-)

If you need anti-aliased pixel drawing, you might want to look into existing variant of setPixelColorXY that take a "float" argument. These functions are already able to perform anti-aliased plotting.

> There aren't any unnecessary changes. I needed to create 2 new functions `drawCircleAntialiased` and `drawLineAntialiased`. The coding style for the circle functions were an anomaly compared to the rest of the code so I modified them to be like the rest of the code base.

Please focus on the feature you would like to add - I agree with @blazoncek that correcting coding style is not needed. There are other forks that sometimes like to cherry-pick single commits; unfortunately the "coding style fix" will make it very hard to do so for future commits that come after your  PR.

bandi13:
I don't agree with that mentality, but not my repo so I will do as you ask. As for the setPixelColorXY my drawCircleAntialiased function does use that.

Do you want me to add `-DWLED_USE_AA_PIXELS` into platformio.ini's `build_flags` as this effect depends on those gates being enabled.

wled00/FX_2Dfcn.cpp
```diff
@@ -513,6 +541,33 @@ void Segment::drawLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint3
...
+  if (!isActive()) return; // not active
+  const uint16_t cols = virtualWidth();
+  const uint16_t rows = virtualHeight();
+  if (x0 >= cols || x1 >= cols || y0 >= rows || y1 >= rows) return;
+  const int16_t dx = abs(x1-x0), sx = x0<x1 ? 1 : -1;
+  const int16_t dy = abs(y1-y0), sy = y0<y1 ? 1 : -1;
+  int16_t err = (dx-dy)/2, e2, x2;
+  int16_t ed = (dx+dy == 0) ? 1 : sqrt((float)dx*dx + (float)dy*dy);
+
+  for (;;) {
```
softhack007, line 555: Coding style: please re-write this for loop into a `while{}` or `do{}while` loop. The gcc optimizer will thank you ;-)

wled00/FX_2Dfcn.cpp
```diff
@@ -513,6 +541,33 @@ void Segment::drawLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint3
...
+//line function
+void Segment::drawLineAntialiased(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint32_t c) {
+  if (!isActive()) return; // not active
+  const uint16_t cols = virtualWidth();
+  const uint16_t rows = virtualHeight();
+  if (x0 >= cols || x1 >= cols || y0 >= rows || y1 >= rows) return;
+  const int16_t dx = abs(x1-x0), sx = x0<x1 ? 1 : -1;
+  const int16_t dy = abs(y1-y0), sy = y0<y1 ? 1 : -1;
+  int16_t err = (dx-dy)/2, e2, x2;
+  int16_t ed = (dx+dy == 0) ? 1 : sqrt((float)dx*dx + (float)dy*dy);
```
softhack007, line 553: you can use `sqrtf()` which is a bit faster on float.
Or look into sqrt16() from fastLED.

wled00/FX_2Dfcn.cpp
```diff
@@ -513,6 +541,33 @@ void Segment::drawLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint3
   }
 }
 
+//line function
+void Segment::drawLineAntialiased(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint32_t c) {
+  if (!isActive()) return; // not active
+  const uint16_t cols = virtualWidth();
+  const uint16_t rows = virtualHeight();
+  if (x0 >= cols || x1 >= cols || y0 >= rows || y1 >= rows) return;
+  const int16_t dx = abs(x1-x0), sx = x0<x1 ? 1 : -1;
```
softhack007, line 550: Try to use `roundf()` or `lround()` -  it could improve accuracy and make the expression simpler because you don't need special handling for x1<x0

wled00/FX.cpp
```diff
@@ -4861,6 +4861,49 @@ static const char _data_FX_MODE_FLOWSTRIPE[] PROGMEM = "Flow Stripe@Hue speed,Ef
...
+uint16_t mode_2DAnalogClock(void) {                  // By Andras Fekete (bandi13)
+  if (!strip.isMatrix) return mode_static(); // not a 2D set-up
+
+  const uint16_t cols = SEGMENT.virtualWidth();
+  const uint16_t rows = SEGMENT.virtualHeight();
+  const uint16_t centerX = cols >> 1;
+  const uint16_t centerY = rows >> 1;
+  const uint16_t radius = min(centerX, centerY);
+
+  auto hours_in_day = (localTime / (60 * 60)) % 12;
```
softhack007, line 4877: Please use `time_t` instead of `auto`

blazoncek: Ouch! @willmmiles will spank you. 😉 We are too old for this auto madness.

wled00/FX_2Dfcn.cpp
bandi13: This function's style is nearly identical to the existing `drawLine()` including the `for (;;)` loop.

wled00/FX_2Dfcn.cpp
bandi13: Not sure I follow.

softhack007: sorry my mistake. I was assuming that `x0` and `x1` are float.

wled00/FX.cpp
willmmiles: Hehe! :) I'm a big AAA fan - few things annoy me more than having to remind the compiler a type name it already f-ing knows - but I'll also be the first to say that *consistency in the local code style* is much more important.

bandi13:
Do you need anything from me to merge this PR?

blazoncek:
> Do you need anything from me to merge this PR?

I will be adding some optimisations and modifications later. Once I'm done, you'll check if everything is still working and then it can be merged.

blazoncek:
I've had a thorough look and test of this new effect and changes made to WLED.

I do not like `drawCircleAntialiased()` and the issue it creates when you want to draw a thin anti-aliased circle. It cannot do it.
I will update `drawCircle()` and `drawLine()` to allow anti-aliased circles and lines that are thin (1 pixel) without using floating point math.

I also do not like the idea that an effect is conditionally compiled. This is unacceptable for single effect IMO.
Another problem with anti-aliased approach taken in this effect is that it looks awful on anything less than 20x20 sized segment so I added an option to disable anti-aliasing while testing and it looked acceptable on 10x10 segments (without seconds hand).

I'll move the PR into draft until AA circle and line are finished and included in 0_15 branch.

blazoncek (approved):
Reworked & approved

wled00/FX.cpp
```diff
@@ -4861,6 +4861,53 @@ static const char _data_FX_MODE_FLOWSTRIPE[] PROGMEM = "Flow Stripe@Hue speed,Ef
 #define XY(x,y) SEGMENT.XY(x,y)
 
 
+/////////////////////////////
+//    2D Analog Clock      //
+/////////////////////////////
+uint16_t mode_2DAnalogClock(void) {                  // By Andras Fekete (bandi13)
+  if (!strip.isMatrix || !SEGMENT.is2D()) return mode_static(); // not a 2D set-up
+
+  const int cols = SEGMENT.virtualWidth();
```
bandi13, line 4870: Not sure why these need to be signed.

wled00/FX.cpp
```diff
@@ -4861,6 +4861,53 @@ static const char _data_FX_MODE_FLOWSTRIPE[] PROGMEM = "Flow Stripe@Hue speed,Ef
...
+
+  SEGMENT.fadeToBlackBy(240);
+  SEGMENT.drawCircle(centerX, centerY, radius, DARKGREY, soft);
+  if (radius > 8) SEGMENT.drawCircle(centerX, centerY, radius-1, DARKGREY, soft); // thicker circle
+
+  time_t hours_in_day = (localTime / (60 * 60)) % 12;
+  time_t minutes_in_day = (localTime / (60)) % 60;
+  time_t seconds_in_day = (localTime) % 60;
+
+  float hour_angle = radians(30.0f * (hours_in_day + minutes_in_day / 60.0f) - 90.0f);
```
bandi13, line 4885: You could have left the original numbers in for code clarity. The preprocessor would have done the simplified math anyway.

wled00/FX.cpp
```diff
@@ -4861,6 +4861,53 @@ static const char _data_FX_MODE_FLOWSTRIPE[] PROGMEM = "Flow Stripe@Hue speed,Ef
...
+    int second_y = centerY + second_len * sin_t(second_angle);
+    SEGMENT.drawLine(centerX, centerY, second_x, second_y, SEGCOLOR(2), soft);
+  }
+
+  SEGMENT.drawLine(centerX, centerY, minute_x, minute_y, SEGCOLOR(1), soft);
+  SEGMENT.drawLine(centerX, centerY, hour_x, hour_y, SEGCOLOR(0), soft);
+
+  return FRAMETIME;
+} // mode_2DAnalogClock()
+static const char _data_FX_MODE_2DANALOGCLOCK[] PROGMEM = "Analog Clock@,,,,,Seconds,Soft;Hour,Minute,Second;;2;o1=1,o2=1";
```
bandi13, line 4908: Any way to make the default colors actually correct? On startup, Hour=white, Minute=black, Second=black for me.

wled00/FX.cpp
```diff
@@ -4861,6 +4861,53 @@ static const char _data_FX_MODE_FLOWSTRIPE[] PROGMEM = "Flow Stripe@Hue speed,Ef
...
 
+/////////////////////////////
+//    2D Analog Clock      //
+/////////////////////////////
+uint16_t mode_2DAnalogClock(void) {                  // By Andras Fekete (bandi13)
+  if (!strip.isMatrix || !SEGMENT.is2D()) return mode_static(); // not a 2D set-up
+
+  const int cols = SEGMENT.virtualWidth();
+  const int rows = SEGMENT.virtualHeight();
+  const int centerX = (cols-!(cols%2)) >> 1;  // use odd sized circle
```
bandi13, line 4872: Not terribly a fan of omitting the last row and/or column, but I can see that the circle drawn is nicer this way

wled00/FX.cpp
```diff
@@ -4861,6 +4861,53 @@ static const char _data_FX_MODE_FLOWSTRIPE[] PROGMEM = "Flow Stripe@Hue speed,Ef
...
+  const int centerY = (rows-!(rows%2)) >> 1;  // use odd sized circle
+  const int radius = min(centerX, centerY);
+  const bool soft = radius > 6 && SEGMENT.check2;
+
+  SEGMENT.fadeToBlackBy(240);
+  SEGMENT.drawCircle(centerX, centerY, radius, DARKGREY, soft);
```
bandi13, line 4878: This ultimately makes a bright spot at the 3,6,9, and 12 o'clock positions. Just looks odd. Only seems to happen when brightness=10.

wled00/FX.cpp
```diff
@@ -4861,6 +4861,53 @@ static const char _data_FX_MODE_FLOWSTRIPE[] PROGMEM = "Flow Stripe@Hue speed,Ef
...
+    int second_len = (radius * 9) / 10;
+    int second_x = centerX + second_len * cos_t(second_angle);
+    int second_y = centerY + second_len * sin_t(second_angle);
+    SEGMENT.drawLine(centerX, centerY, second_x, second_y, SEGCOLOR(2), soft);
+  }
+
+  SEGMENT.drawLine(centerX, centerY, minute_x, minute_y, SEGCOLOR(1), soft);
+  SEGMENT.drawLine(centerX, centerY, hour_x, hour_y, SEGCOLOR(0), soft);
+
+  return FRAMETIME;
```
bandi13, line 4906: Why not calculate this once per second? No need to do it any more or less.

bandi13 (commented):
I left a few comments, most are for my understanding. Seems to work though the default colors are terrible.

wled00/FX.cpp
blazoncek: prevents compiler warnings when doing comparison with int
all that is needed are 15 bits so it fits in int

wled00/FX.cpp
blazoncek: can be added into comment if you think so.
compiler is unable to optimise `360 * (hours_in_day + minutes_in_day / 60.0) / 12` as it needs to mix integer and floating point maths. with integer maths precedence is important as you are loosing precision when dividing before multiplication.
that may not hold true with every compiler though.

wled00/FX.cpp
blazoncek: Unfortunately you can't but I guess you are *not* using *every* effect using *default* values. Or are you?
If you are not, you should consider using presets which save every configuration item.

wled00/FX.cpp
blazoncek: with even sized (%2 == 0) circles there is no correct center of the circle as it falls between two pixels. in such case the hands become ugly on anything smaller than 32x32 pixel

wled00/FX.cpp
blazoncek: I did not observe such behaviour on up to 20x20 (which I happen to have).

Consider changing segment size between 8x8 (non AA) and 32x32 (or larger if you have it). Then observe AA circle. Some pixels will be dimmer than others. And yes brightness plays a role when restoring/sampling a pixel using `getPixelColor()` if not using global buffer.

But I thought you wanted AA circle. Otherwise you should've used `fillCircle()` to get arbitrary sized border of uniform brightness. I did experiment with it but ultimately rejected it as it was obvious to me you wanted a smooth look.

wled00/FX.cpp
blazoncek: Agreed that there should be no need to execute it more (or less) often than every 1000ms. But default FPS is 42 which does not divide 1000 nicely (without remainder). So there will be situations where seconds hand will behave erratic if you'd return 1000 (that is the amount of time needed until next execution) and not `1000 / FPS` (which is FRAMETIME when using default FPS).

Try it on small dimension. And with TM1814 LEDs.
Each effect should return FRAMETIME to allow seamless transitions. The return value is deprecated and is kept for backwards compatibility until all effects are rewritten to be executed at the user specified FPS.

wled00/FX.cpp
softhack007: Effect metadata format is documented here: 
https://kno.wled.ge/interfaces/json-api/#effect-metadata

It looks like you can define default colors in metadata, too.

bandi13: The default values should be ones that work for most cases out of the box. When you buy a car/phone/computer/anything it should work out of the box without you having to edit parameters to make it functional. Note that I didn't say customized, just functional out of the gate.

Maybe you could have a special case, that if the Hour=White and Minute=Second=Black that it would set the colors to Red, Green, Blue respectively?

wled00/FX.cpp
bandi13: Why not go full Monty and do `radians(30.0f * hours_in_day + minutes_in_day / 2.0f - 90.0f)`? Divide by 2 is more efficient than another number (just decrement the mantissa IIRC).

A comment is definitely warranted. Though we could go back to the original equation and have all the constants be `float` which should let the preprocessor figure out that it needs to do `360 / 12`.

blazoncek: > Why not go full Monty and do `radians(30.0f * hours_in_day + minutes_in_day / 2.0f - 90.0f)`? Divide by 2 is more efficient than another number (just decrement the mantissa IIRC).

You can do that if you want. Anything that reduces CPU clock cycles is welcome.

wled00/FX.cpp
blazoncek: When you buy it, yes. If you get it for free, it is just a nice touch.

> Maybe you could have a special case, that if the Hour=White and Minute=Second=Black that it would set the colors to Red, Green, Blue respectively?

You can do that in effect initialisation code that starts as:

```c++
if (SEGENV.call == 0) {
  SEGMENT.colors[0] = <hour_color>;
...
}
```

But I would only change those if they a black.
There are examples elsewhere in the code.

The other option is to enhance metadata. That will be a lot more work IMO.

bandi13:
I put in the default colors. Seems to work well enough.

Sorry it took some time, I'm only doing this as a hobby. Weekends = family time. Strictly enforced. ;-)

blazoncek:
One more thing occurred to me tonight. WLED has "analog clock overlay" feature (for strips) and since segments now can support overlapping (if coded properly) it would be ok IMO to move that overlay code into this effect and make it 1D + 2D.

This would add 2 benefits to analog clock overlay:
1) it could be turned on/off using segment power button (or JSON API) instead of in settings
2) it could only encompass part of the strip instead of entire strip

@softhack007 what do you think?

BTW @bandi13 do you think I'm doing WLED professionally? No, I do not. My sleep and spare time suffer so that I can do WLED. 😀

bandi13:
Should that overlay code refactor be a separate PR?

blazoncek:
IMO if this effect (analog clock) can replace existing overlay, no. I'll remove overlay code once effect can handle 1D and 2D.

@softhack007 @Aircoookie your opinion?

blazoncek:
Reference usermod: #2736

PR #3944, opened 2024-04-30
Using brightness in analog clock overlay

paspiz85:
I want to adjust the brightness of the analog clock. I used the existing brightness setting to decrease the intensity of the 3 LEDs related to the analog clock overlay.
I tested on my ESP8266 Wemos D1 mini

### DIFF
--- a/wled00/overlay.cpp
+++ b/wled00/overlay.cpp
@@ -11,7 +11,6 @@ void _overlayAnalogClock()
   {
     _overlayAnalogCountdown(); return;
   }
-  uint8_t brightness = strip.getBrightness();
   float hourP = ((float)(hour(localTime)%12))/12.0f;
   float minuteP = ((float)minute(localTime))/60.0f;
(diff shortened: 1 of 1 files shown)


blazoncek:
That's not how it should go.
Use `color_fade()` available in 0_15 branch.

wled00/overlay.cpp
```diff
@@ -11,6 +11,7 @@ void _overlayAnalogClock()
   {
     _overlayAnalogCountdown(); return;
   }
+  uint8_t brightness = strip.getBrightness();
```
softhack007, line 14: @blazoncek not sure what is better to retrieve the global brightness - simply use global `bri`, or use `‎WS2812FX::getBrightness‎()`?

blazoncek: Yes. That would be in line with other code that uses it.

paspiz85: ok

PR #3945, opened 2024-04-30
Add Webpage shortcuts and Fix resizing bug

w00000dy:
This adds Web page shortcuts.

Now you can use these URLs to go directly to (Colors), Effects, Segments, and Presets:
http://[ip]/#Colors
http://[ip]/#Effects
http://[ip]/#Segments
http://[ip]/#Presets

Also, when you click through the container, the URL in your browser changes accordingly, so users know they can use those URLs.
grafik

I also fixed a small bug. The bug was that when resizing the window, it always jumped to the Colors tab instead of staying on the currently selected tab.

### DIFF
--- a/wled00/data/index.js
+++ b/wled00/data/index.js
@@ -272,6 +272,7 @@ function onLoad()
 
 	selectSlot(0);
 	updateTablinks(0);
+	handleLocationHash();
 	cpick.on("input:end", () => {setColor(1);});
 	cpick.on("color:change", () => {updatePSliders()});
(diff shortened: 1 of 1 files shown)


blazoncek:
@w00000dy please verify if my tablet fix still works for you.

w00000dy:
What was not working before your fix?

w00000dy:
> @w00000dy please verify if my tablet fix still works for you.

I tested it and it works! Thank you very much! 👍🏼

PR #3949, opened 2024-05-01
Correct mA/LED value for WS2815

romosborne:
Currently it's not possible to save an LED strip set as WS2815 because the mA/LED value falls outside the allowed range.  The form validation throws an error because the input element isn't focusable because it's hidden by default and only shown when a custom mA/LED value is entered

This PR fixes that by setting the correct value for mA/LED for WS2815 strips


blazoncek:
The proposed fix is incorrect. Sorry to disappoint you.
Proper fix is implemented in 0_15 which has a new, per output ABL.

PR #3003, opened 2023-01-05
Usermod Battery 🔋 Added Support for different battery types, Optimized file structure

itCarl:
Currently adding new features / restructurings Things.
The main purpose for this PR is so that can @softhack007 can see what i already made.

PR #3959, opened 2024-05-06
Update Usermod: Battery

adamsthws:
Issue:
When taking the initial voltage reading after first powering on, voltage hasn't had chance to stabilize so the reading can be inaccurate, which in turn may incorrectly trigger the low-power preset. (Manifests when the user has selected a low read interval and/or is using a capacitor).

Resolution:
A non-blocking, fixed 10 second delay has been added to the initial voltage reading to give the voltage time to stabilize.


blazoncek:
Please rebase this for 0_15 branch and perhaps coordinate this PR with @itCarl who is the author and has [another PR in queue](https://github.com/Aircoookie/WLED/pull/3003).

itCarl:
Looks like a reasonable update.
I would suggest that this PR is merged after merging PR #3003

PR #3961, opened 2024-05-07
Added Pinwheel Expand 1D Effect

Brandon502:
Requested on discord to bring this effect from MM to main branch. Reworked and optimized by @softhack007.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -738,16 +738,17 @@ void IRAM_ATTR Segment::setPixelColor(int i, uint32_t col)
         float centerY = roundf((vH-1) / 2.0f);
         // int maxDistance = sqrt(centerX * centerX + centerY * centerY) + 1;
         float angleRad = (max(vW,vH) > Pinwheel_Size_Medium) ? float(i) * Int_to_Rad_Big : float(i) * Int_to_Rad_Med; // angle in radians
-        float cosVal = cosf(angleRad);
-        float sinVal = sinf(angleRad);
+        float cosVal = cos_t(angleRad);
+        float sinVal = sin_t(angleRad);
 
         // draw line at angle, starting at center and ending at the segment edge
         // we use fixed point math for better speed. Starting distance is 0.5 for better rounding
-        constexpr int_fast32_t Fixed_Scale = 512;  // fixpoint scaling factor
-        int_fast32_t posx = (centerX + 0.5f * cosVal) * Fixed_Scale; // X starting position in fixed point
-        int_fast32_t posy = (centerY + 0.5f * sinVal) * Fixed_Scale; // Y starting position in fixed point
-        int_fast16_t inc_x = cosVal * Fixed_Scale; // X increment per step (fixed point)
-        int_fast16_t inc_y = sinVal * Fixed_Scale; // Y increment per step (fixed point)
+        // int_fast16_t and int_fast32_t types changed to int, minimum bits commented
+        constexpr int Fixed_Scale = 512;  // fixpoint scaling factor 18 bit
+        int posx = (centerX + 0.5f * cosVal) * Fixed_Scale; // X starting position in fixed point 18 bit
+        int posy = (centerY + 0.5f * sinVal) * Fixed_Scale; // Y starting position in fixed point 18 bit
+        int inc_x = cosVal * Fixed_Scale; // X increment per step (fixed point) 10 bit
+        int inc_y = sinVal * Fixed_Scale; // Y increment per step (fixed point) 10 bit
 
         int32_t maxX = vW * Fixed_Scale; // X edge in fixedpoint
         int32_t maxY = vH * Fixed_Scale; // Y edge in fixedpoint
@@ -885,8 +886,8 @@ uint32_t IRAM_ATTR Segment::getPixelColor(int i)
         float centerX = (vW - 1) / 2.0f;
         float centerY = (vH - 1) / 2.0f;
         float angleRad = (max(vW,vH) > Pinwheel_Size_Medium) ? float(i) * Int_to_Rad_Big : float(i) * Int_to_Rad_Med; // angle in radians
-        int x = roundf(centerX + distance * cosf(angleRad));
-        int y = roundf(centerY + distance * sinf(angleRad));
+        int x = roundf(centerX + distance * cos_t(angleRad));
+        int y = roundf(centerY + distance * sin_t(angleRad));
         return getPixelColorXY(x, y);
         break;
       }
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -638,15 +638,14 @@ uint16_t IRAM_ATTR Segment::nrOfVStrips() const {
 }
 
 // Constants for mapping mode "Pinwheel"
-constexpr int Pinwheel_Steps_Small = 72;       // no holes up to 16x16;
+constexpr int Pinwheel_Steps_Small = 72;       // no holes up to 16x16
 constexpr int Pinwheel_Size_Small = 16;       
-constexpr int Pinwheel_Steps_Medium = 192;     // no holes up to 32x32;  60fps
+constexpr int Pinwheel_Steps_Medium = 192;     // no holes up to 32x32
 constexpr int Pinwheel_Size_Medium = 32;       // larger than this -> use "Big"
-constexpr int Pinwheel_Steps_Big = 296;        // no holes expected up to  58x58; 40fps
-constexpr float Int_to_Rad_Small = (DEG_TO_RAD * 360) / Pinwheel_Steps_Small;  // conversion: from 0...208 to Radians
-constexpr float Int_to_Rad_Med = (DEG_TO_RAD * 360) / Pinwheel_Steps_Medium;   // conversion: from 0...208 to Radians
-constexpr float Int_to_Rad_Big = (DEG_TO_RAD * 360) / Pinwheel_Steps_Big;      // conversion: from 0...360 to Radians
-int prevRay = INT_MIN;
+constexpr int Pinwheel_Steps_Big = 296;        // no holes expected up to  56x56
+constexpr float Int_to_Rad_Small = (DEG_TO_RAD * 360) / Pinwheel_Steps_Small;  // conversion: from 0...72 to Radians
+constexpr float Int_to_Rad_Med = (DEG_TO_RAD * 360) / Pinwheel_Steps_Medium;   // conversion: from 0...192 to Radians
+constexpr float Int_to_Rad_Big = (DEG_TO_RAD * 360) / Pinwheel_Steps_Big;      // conversion: from 0...296 to Radians
 
 // 1D strip
 uint16_t IRAM_ATTR Segment::virtualLength() const {
@@ -738,6 +737,7 @@ void IRAM_ATTR Segment::setPixelColor(int i, uint32_t col)
         for (int y = 0; y <  i; y++) setPixelColorXY(i, y, col);
         break;
       case M12_sPinwheel: {
+        static int prevRay = INT_MIN; // previous ray number
         // i = angle --> 0 - 296  (Big), 0 - 192  (Medium), 0 - 72 (Small)
         float centerX = roundf((vW-1) / 2.0f);
         float centerY = roundf((vH-1) / 2.0f);
@@ -761,9 +761,9 @@ void IRAM_ATTR Segment::setPixelColor(int i, uint32_t col)
         int32_t maxY = vH * Fixed_Scale; // Y edge in fixedpoint
 
         // Odd rays start further from center if prevRay started at center.
-        if (i % 2 == 1 && (i - 1 == prevRay || i + 1 == prevRay)) {
-          posx = (posx + inc_x * (vW/4));
-          posy = (posy + inc_y * (vH/4));
+        if ((i % 2 == 1) && (i - 1 == prevRay || i + 1 == prevRay)) {
+          posx += inc_x * (vW/4);
+          posy += inc_y * (vH/4);
         }
         prevRay = i;
 
@@ -772,13 +772,8 @@ void IRAM_ATTR Segment::setPixelColor(int i, uint32_t col)
           // scale down to integer (compiler will replace division with appropriate bitshift)
           int x = posx / Fixed_Scale;
           int y = posy / Fixed_Scale;
-#if 1
           // set pixel
           if (x != lastX || y != lastY) setPixelColorXY(x, y, col);  // only paint if pixel position is different
-#else
-          // experimental: only set pixel if color is different (trade getPC performance against setPC)
-          if ((x != lastX || y != lastY) && (getPixelColorXY(x, y) != col)) setPixelColorXY(x, y, col); // only paint if pixel color is different
-#endif
           lastX = x;
           lastY = y;
           // advance to next position
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -640,13 +640,13 @@ uint16_t IRAM_ATTR Segment::nrOfVStrips() const {
 // Constants for mapping mode "Pinwheel"
 constexpr int Pinwheel_Steps_Small = 72;       // no holes up to 16x16;
 constexpr int Pinwheel_Size_Small = 16;       
-constexpr int Pinwheel_Steps_Medium = 200;     // no holes up to 32x32;  60fps
+constexpr int Pinwheel_Steps_Medium = 192;     // no holes up to 32x32;  60fps
 constexpr int Pinwheel_Size_Medium = 32;       // larger than this -> use "Big"
 constexpr int Pinwheel_Steps_Big = 296;        // no holes expected up to  58x58; 40fps
 constexpr float Int_to_Rad_Small = (DEG_TO_RAD * 360) / Pinwheel_Steps_Small;  // conversion: from 0...208 to Radians
 constexpr float Int_to_Rad_Med = (DEG_TO_RAD * 360) / Pinwheel_Steps_Medium;   // conversion: from 0...208 to Radians
 constexpr float Int_to_Rad_Big = (DEG_TO_RAD * 360) / Pinwheel_Steps_Big;      // conversion: from 0...360 to Radians
-
+int prevRay = INT_MIN;
 
 // 1D strip
 uint16_t IRAM_ATTR Segment::virtualLength() const {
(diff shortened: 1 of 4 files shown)


blazoncek (changes requested):
Looks fine to me, except the use of sinf/cosf which should be replaced by sin_t/cos_t.

BTW I found out that there is no need to use `int_fast32_t` unless you strictly need 32 bits and the code is ported to 16 bit environments (or `int_fast16_t`). `int` would be cleaner (or `unsigned` where appropriate) and the compiler treats it as such, but that's just my preference.

softhack007:
Hi, let me answer as the one who did the optimization part of the new mapping mode.

> Looks fine to me, except the use of sinf/cosf which should be replaced by sin_t/cos_t.

sin_t/cos_t are slow and less accurate. 

I guess you intend to reduce 8266 memory footprint, by not pulling in the dependency to std math lib?

> 
> BTW I found out that there is no need to use `int_fast32_t` unless you strictly need 32 bits and the code is ported to 16 bit environments (or `int_fast16_t`). `int` would be cleaner (or `unsigned` where appropriate) and the compiler treats it as such, but that's just my preference.

Not sure. In fact we do fixed-point math in setPixelColor(). As the x and y position is increment until we hit the border,  9 bit of "fraction" just leaves about 1-2bit of slack when one of the matrix dimensions is close to the maximum of 256 (imagine a matrix 196x4). Unsigned is not possible, because we have negative increments when sin() gets negative. 

So considering that x and y can at most 256, we'd need 18bit of resolution for good results (one bit for sign, 8bit integer part, 9bit fraction). Maybe we can  squeeze out 2bits of fraction so everything fits into 16bit. I'd need to do some more testing to see if reducing the resolution creates holes in the mapping.

We might be able to implement your suggestion at least in the getpixelcolor() part, which is inaccurate any way.

What do you think?

softhack007:
(Changing hat - now speaking as a maintainer 🎩 ) 

the pinwheel mapping works with many effects, however it has a few open ends:
- not explictly supporting "virtual 2D strip" effects like fire or popcorn
- getpixelColor is sometimes inaccurate, as it returns color of a pixels with correct angle, however on the "outer circle", which was possibly overwritten by another "ray". This could be improved by using a pixel from the segment edge, but that's still not perfect. To really solve this issue, the papping mode would need to allocate it's own buffer on segment level.

However I think these two limiations are still acceptable for a first shot, just there is room for improvement in the future.

blazoncek:
`sin_t`/`cos_t` implementation can be overridden using `-D WLED_USE_REAL_MATH`, `sinf`/`cosf` cannot.
They were introduced by @Aircoookie and I insist they are used across core WLED until he changes his mind. And yes, they were introduced to reduce RAM usage on ESP8266 despit being slightly inaccurate at extreme edges.

`int_fast16_t` is just a typedef of `int` on ESP platform (as it is 32 bit). So is `int_fast32_t`.
At least that's what compiler and VSC tell me.
It does matter with non fast versions. What compiler does is it adds `& 0xFFFF` to every instruction when using `int16_t` (non fast).

softhack007:
> despit being slightly inaccurate at extreme edges.

Ok - just let me add that sin_t and cos_t are not just inaccurate -  on esp32 platforms they are noticeably slower than sinf/cosf. 

@Brandon502 would you make the changes as request by @blazoncek ? For the int_fast16_t and int_fast32_t types, please change them to `int` and add a comment about min required number of bits ( I.e. 18 for int32, 10 for int16).

softhack007 (approved):
Approved from my side, assuming that changes already requested by @blazoncek get implemented.

Adding the "documentation" tag, because I think that limitations of the pinwheel mapping - it may work well with some effects - should be documented. But that's a different task.

softhack007:
@blazoncek do you think this can go into the upcoming 0.15.0-b3, or should we better wait and merge this PR after -b3 is out?

blazoncek:
I think this will be included in b3 release this weekend.

Brandon502:
@softhack007 I never tested large matrices after your rework just looked at the code. I loaded up latest MM version on a virtual 64x64 matrix. Up to 30x30 works fine. 31x31 and 32x32 produces holes. 33x33 up to 50x50 also works correctly. 51x51 and above also produces holes on my end. Not sure if I'm setting up something wrong or if some things need to be tweaked a bit.

softhack007:
> @softhack007 I never tested large matrices after your rework just looked at the code. I loaded up latest MM version on a virtual 64x64 matrix. Up to 30x30 works fine. 31x31 and 32x32 produces holes. 33x33 up to 50x50 also works correctly. 51x51 and above also produces holes on my end. Not sure if I'm setting up something wrong or if some things need to be tweaked a bit. This applies to v15 version as well for me.

@Brandon502 thanks for testing. Actually, my main test environments are 24x32 and 52x52 where I didn't see holes.
Holes are always possible when the number of pixels gets very hight. Its mainly about finding the best "tuning" that avoids holes without blowing up runtime too much. 

I can take a look on Sunday. If you have time, maybe test out these ideas:

* For the 32x32 case, please change the 32 in `constexpr int Pinwheel_Size_Medium = 32;` to 30.
* for the >50x50 case, I need to see if rounding helps with avoiding holes. I was hoping to get away without rounding for speed reasons. Maybe lines 758 + 759 need to include rounding, i.e. `int x = (posx + Fixed_Scale/2) /  Fixed_Scale;`.

blazoncek (approved):
I am a bit perplexed by complexity of what should be a simple expansion.
I did not try to compile so can't judge the efficiency of the code but for testing I'd recommend setting up 4 segments (16x16) with pinwheel expansion and see what FPS you get.

Otherwise the code looks ok.

EDIT: Do you have a video of the expansion in action?

Brandon502:
> I am a bit perplexed by complexity of what should be a simple expansion. I did not try to compile so can't judge the efficiency of the code but for testing I'd recommend setting up 4 segments (16x16) with pinwheel expansion and see what FPS you get.
> 
> Otherwise the code looks ok.
> 
> EDIT: Do you have a video of the expansion in action?


https://github.com/Aircoookie/WLED/assets/105077712/fe369c27-4e42-4a00-a5de-32c815c824f9

virtual matrix so liveview is a bit laggy. Shows a stable 20fps running 4 16x16 segments (Rainbow, Two Dots, Rainbow Runner, Washing Machine). A single 32x32 effect runs around ~28fps.

blazoncek:
> Shows a stable 20fps running 4 16x16 segments (Rainbow, Two Dots, Rainbow Runner, Washing Machine). A single 32x32 effect runs around ~28fps.

I thought it would be slow. Unfortunate as that is.

@Brandon502 & @softhack007 if you can spot a place for speed optimisation this would need it.

A word about missing pixels: Unless you increase the "resolution" of your expansion (i.e. slow it down), you'll always get them on certain dimensions. Larger dimensions more chance for them as you are dealing with sin/cos.

softhack007:
> @Brandon502 & @softhack007 if you can spot a place for speed optimisation this would need it.

Yes there are two options:
* add rounding (see my previous comment) and reduce the number of "rays"
* add two more "steps" by finding the optimal number of rays for: size <= 16 (~160 Ray's maybe less); 30 <= size <= 40 (~256 rays may work)

I'm always from my esp32 today - most likely can test that tomorrow.

Brandon502:
> > @Brandon502 & @softhack007 if you can spot a place for speed optimisation this would need it.
> 
> Yes - the limiting factor is the number of setPixelColor() calls needed for the full pinwheel, so I see two options:
> 
>     * add rounding (see my previous comment) and reduce the number of "rays"
> 
>     * add two more "steps" by finding the optimal number of rays for common sizes: size <= 16 (~160 rays maybe less); 30 <= size <= 40 (~256 rays may work)
> 
> 
> I'm always from my esp32 today - most likely can test that tomorrow.

Created a custom effect so I can control rays and turn rounding on/off. Rounding seems to be much worse, at least the method you mentioned above. Here's the amount of rays needed for various sizes without rounding.

    8x8 30 rays
16x16 72 rays
24x24 128 rays
32x32 200 rays
40x40 272 rays
48x48 296 rays
56x56 296 rays
64x64 448 rays

This may not be 100% accurate, but it's a decent baseline.

blazoncek:
Dear Lord! 😁 For each pixel you draw 200 lines on 32x32? No wonder it is slow.

You need to optimize loop to only call `setPixelColor()` when either x or y actually changes from previous iteration.
Next optimisation would be to have dynamic "rays" that start at certain distance from center instead of all of them starting at center and overwriting same pixel over and over again.

Brandon502:
I added small pinwheel option for matrices 16x16 and below. The 4 16x16 effects on 32x32 grid went from ~20 fps to ~33 fps. Pinwheel seems to be higher fps then arc on effects that I've tested.

blazoncek:
> Pinwheel seems to be higher fps then arc on effects that I've tested.

Yes arc is slow as it iterates in polar space (increasing angle) instead of using Bresenham's circle drawing algorithm as that produces empty pixels. That is also in need of optimisation.

Brandon502:
> Dear Lord! 😁 For each pixel you draw 200 lines on 32x32? No wonder it is slow.
> 
> You need to optimize loop to only call `setPixelColor()` when either x or y actually changes from previous iteration. Next optimisation would be to have dynamic "rays" that start at certain distance from center instead of all of them starting at center and overwriting same pixel over and over again.

200 lines total on 32x32. setPixelColor seems to be called 3604 times on 32x32. So that can definitely be reduced by avoiding overwriting near the center like you said. I'll play around with trying to reduce it.

Brandon502:
@softhack007 Here's a couple stats I got earlier from the effect. May help optimize. 
On 32x32 with 200 lines, setPixelColor was called 3604 times. 212 of those calls were back to back duplicates within the same line. 16x16 with 72 lines, setPixelColor was called 652 times. 28 of those calls were back to back duplicates within the same line.

One way I tried reducing is by having every odd line start half way to the edge instead of the center. This took about 1/3 of the calls away. Works great and looks pretty much identical on effects that fill the entire matrix like rainbow, flow, etc. But on effects that scan or produce thinner lines it makes them look much worse. 

Hopefully you'll be able to find another method to optimize without changing the overall effect too much.

softhack007:
> Dear Lord! 😁 For each pixel you draw 200 lines on 32x32?

Not really 😉 actually we map each "pixel" to a single ray coming from the centre. So for 32x32 we have a total of ~200 rays for ~200 virtual pixels - each virtual pixel is one ray, and for a full frame all rays are needed.

The ray algo is as as fast as it can be - we only need four integer operations to get the next pixel of a ray 🚀, plus 4 compares to detect if we left the canvas.

> You need to optimize loop to only call `setPixelColor()` when either x or y actually changes from previous iteration.

Good idea :-) yes this might help, and it will not change the look at all. I'll try it tomorrow.

Brandon502:
We can drop the number of rays for 32x32 even lower to 192. 
The shorter odd rays I mentioned above can work if there is a global variable that keeps track of the last ray number drawn. Then you can check if a ray is odd and the previous ray was one before it. Saves quite a few calls and this can probably be expanded further if you can't think of another method. With this method FPS went up to 35 with the 4 16x16 effects and FPS went from 29 to 33 on a single 32x32 Rainbow.

blazoncek:
> Not really 😉 actually we map each "pixel" to a single ray coming from the centre. So for 32x32 we have a total of ~200 rays for ~200 virtual pixels - each virtual pixel is one ray, and for a full frame all rays are needed.

Exactly, for each of virtual 200 pixels you need to draw a line (aka ray) from center to the edge (that's from 16 to 22 physical pixels). That would total roughly 3800 pixels painted (average). On a 1024 pixel matrix!

Another optimisation could be to keep a 2D array of already painted pixels. That would reduce `setPixelColor()` calls dramatically.
It is a temporary array, kept on stack (about 1k for 32x32 if using bool and less if using bitfields).

softhack007:
> Another optimisation could be to keep a 2D array of already painted pixels. That would reduce `setPixelColor()` calls dramatically.

I don't think this will work in the "general case" where only some pixels are re-drawn by effects - like fireworks, popcorn, chase, meteor, gravimeter or freqmap. Also we will need something like "segment::startFrame()" to reset the array of painted pixels... So might need some more thinking ...

softhack007:
@Brandon502 i think our best chance - in addition to my optimizations to avoid double painting of pixels - is to find optimized numbers of rays for common sizeses -> 16x16, 32x32

Brandon, I'm not sure I fully understand your "odd rays" idea - basically you draw even numbered ones from centre (outwards) while drawing odd-numbered ones from the edge (inwards) - is this correct?

I want to try a similar one later today - by just moving the start of "odd"  rays left by 1/2 pixel, we possibly get a slightly better coverage and less holes.

Please test my latest commit with your "test setup" because i'm curious if you see a speedup.

Brandon502:
> @Brandon502 i think our best chance - in addition to my optimizations to avoid double painting of pixels - is to find optimized numbers of rays for common sizeses -> 16x16, 32x32 (we already have that)
> 
> I'm not sure I fully understand your "odd rays" idea - basically you draw even numbered ones from centre (outwards) while drawing odd-numbered ones from the edge (inwards) - is this correct?
> 
> later today I want to try a similar idea - by just moving the start of "odd" rays left by 1/2 pixel, we possibly get a slightly better coverage and less holes.
> 
> Please test my latest commit with your "test setup" because i'm curious if you see a speedup.

Here's the fps on my setup.

38 FPS on 4 16x16 effects
31 FPS on 1 32x32 Rainbow

Experimental method:
34 FPS on 4 16x16 effects
26 FPS on 1 32x32 Rainbow

Combining your lastX/Y and my odd ray method
40 FPS on 4 16x16 effects
35 FPS on 1 32x32 Rainbow

I pushed my odd ray code. You can likely change the math and make it much quicker. It still goes outward but I just increment it a few times before entering the loop so it starts about midway to the edge instead of the center. When you keep track of the prevRay number you can use this method without breaking effects that don't cover full screen. 

`posx = (posx + inc_x * (vW/4));`

This is how I'm jumping away from the center. On square matrices you can go even further from the center, but rectangles will start producing holes. Can likely find a better spot to jump and make it quicker.

softhack007:
> Combining your lastX/Y and my odd ray method
> 40 FPS on 4 16x16 effects
> 35 FPS on 1 32x32 Rainbow

Thanks, I think that's a great result - we are almost twice as fast now, compared to the original code. And still no holes 😃 

I'll look at your "odd ray" code later tonight - and maybe add some tuning for the math.

blazoncek:
I'll wait with b3.

Brandon502:
@softhack007 can't test the 52x52 hole yet, but the holes on 24x32, 12x16, and 14x16 are from the prevRay method it seems. Need to tweak the jump distance a bit more.

Brandon502:
May have messed up while committing. Forgot to pull before committing. https://github.com/Aircoookie/WLED/pull/3961/commits/9e0b91ac17f4ad84978be399d7c8f066a4508b85 contains the fix. Still need to test on 52x52 though. The jump needed to be the same for each axis. This improves fps slightly for a single 32x32 effect.

Brandon502:
oddRays
EvenRays

Quick visualization showing only the odd rays / even rays. May help show areas for further improvement.

Brandon502:
@softhack007 fixed music mode effects. Got tired of trying ways to fix the weird rounding issues causing wrong pixel to be used in getPixelColor so I just used the exact same method as setting. Seems like it works. Changed small and medium pinwheel sizes back to the smaller values. I haven't spotted any missing holes.

softhack007:
> Got tired of trying ways to fix the weird rounding issues causing wrong pixel to be used in getPixelColor so I just used the exact same method as setting

Hi @Brandon502, maybe that's the better solution 👍 . 
A part of me keeps saying "DRY - don't repeat yourself" but most important is "it works", and the ray algo is fast enough to re-run it (without setting pixels) in getpixelcolor. 

The other benefit is that we get higher resolution in getpPixelColor now, because we always use an edge pixel not a circle inside the segment frame - that's an improvement especially on non-square layouts.

@blazoncek I think we are good now for  beta3. Maybe just wait if something comes up until the evening, then merge the PR.

softhack007:
Hi guys, for me it's good now 😃 
@Brandon502 maybe you can do a final cross-check before we merge?


An intersting goodie (and we should keep that) is the behaviour of "Wavesins" with pinwheel mapping. 
Its producing a moirè pattern that alternates with every "round"
Wavesins_Pinwheel

Brandon502:
> Hi guys, for me it's good now 😃 @Brandon502 maybe you can do a final cross-check before we merge?
> 
> An intersting goodie (and we should keep that) is the behaviour of "Wavesins" with pinwheel mapping. Its producing a moirè pattern that alternates (black/orange) with every "full scan" round. Looks a bit like a radar screen 😉
> 

Hey @softhack007, I did a run through of all the effects last night. The only "broken" effects are now ones that use a "stripe" pattern. Like wavesins (above) and traffic light along with a few others. With these you can see the jump on odd rays. Even if you remove the jump they still are somewhat broken since close rays still overwrite each other. Playing with the sliders on wavesins does produce so nice patterns with the jump visible.

I'm not sure why you're getting holes on your large setup. If you want to reduce the count of rays even further can you try these values....

```
constexpr int Pinwheel_Steps_Small = 64;       
constexpr int Pinwheel_Steps_Medium = 160;
constexpr int Pinwheel_Steps_Big = 296; 
```

```
//    8x8  ->  24 steps    36x36  ->  192 steps
//  12x12  ->  48 steps    40x40  ->  216 steps 
// *16x16  ->  64 steps    44x44  ->  256 steps
//  20x20  ->  88 steps    48x48  ->  256 steps
//  24x24  -> 112 steps    52x52  ->  296 steps
//  28x28  -> 136 steps   *56x56  ->  296 steps
// *32x32  -> 160 steps    60x60  ->  408 steps
```

I've doubled check all of these values, so if they aren't working for you on >32x32 I can't trust my setup anymore lol.

softhack007:
> I'm not sure why you're getting holes on your large setup.

One difference is I'm using an -S3 in my setup, which requires a newer framework version (esp-idf 4.4.4, arduino-esp 2.0.9). Could be that my additional holes comes from slightly different accuracies of sinf/cosf in the new framework. I had two clear holes in 51x52 and 52x52.


> If you like data here's the optimal rays for every grid size and the amount of setPixelColor calls with and without jump needed for each

Thanks 👍 will put them into excel later tonight - maybe there's a pattern. I'd say let's do a second optimization round when -beta3 is out.

Brandon502:
> > If you like data here's the optimal rays for every grid size and the amount of setPixelColor calls with and without jump needed for each
> 
> Thanks 👍 will put them into excel later tonight - maybe there's a pattern. I'd say let's do a second optimization round after -beta3 is out.

Sounds good. Just pushed a small change to format getPinwheelAngle the same as getPinwheelLength removing the long single line method. 

Final FPS numbers:
41 FPS on 4 16x16
37 FPS on 32x32 Rainbow

wled00/FX_fcn.cpp
```diff
@@ -640,23 +640,27 @@ uint16_t IRAM_ATTR Segment::nrOfVStrips() const {
...
 
 // Pinwheel helper function: pixel index to radians
 static float getPinwheelAngle(int i, int vW, int vH) {
   int maxXY = max(vW, vH);
-  return (maxXY > Pinwheel_Size_Small) ? ((maxXY > Pinwheel_Size_Medium) ? ((maxXY > Pinwheel_Size_Big) ? float(i) * Int_to_Rad_XL : float(i) * Int_to_Rad_Big) : float(i) * Int_to_Rad_Med) : float(i) * Int_to_Rad_Small;
+  if (maxXY <= Pinwheel_Size_Small)  return float(i) * Int_to_Rad_Small;
+  if (maxXY <= Pinwheel_Size_Medium) return float(i) * Int_to_Rad_Med;
+  if (maxXY <= Pinwheel_Size_Big)    return float(i) * Int_to_Rad_Big;
+  // else
+  return Pinwheel_Steps_XL;
```
softhack007, line 663: small oopsie - the final return should be `return float(i) * Int_to_Rad_XL`

Brandon502: Can't believe I didn't see that. Thanks, fixed.

PR #3962, opened 2024-05-08
ESP8266 audioreactive UDP sync ported from MoonModules/WLED

### LINKED ISSUE #3960: Audio Data UDP Sync (receive only) on ESP8266
I needed to sync the effects from an ESP32 with an I2S mic to an ESP8266 (actually, an Athom controller running on an ESP8285, but I believe it's almost the same). I was wondering why this feature is not implemented in mainline WLED, especially since the code is basically already there and does not require any particular dependency (I think).

Then I found MoonModules/WLED/pull/60 and successfully ported it to 0_15... It does the job and seems to be stable, so I am asking if there is any particular reason not to add this feature also here.

Thank you for your time, and pardon my inexperience!

gaaat98:
Tries to resolve #3960 

However, it seems to be unstable on RGB(W) PWM strips at higher clock speeds, causing frequent reboots

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -195,6 +195,7 @@ build_flags =
   ; decrease code cache size and increase IRAM to fit all pixel functions
   -D PIO_FRAMEWORK_ARDUINO_MMU_CACHE16_IRAM48 ;; in case of linker errors like "section `.text1' will not fit in region `iram1_0_seg'"
   ; -D PIO_FRAMEWORK_ARDUINO_MMU_CACHE16_IRAM48_SECHEAP_SHARED ;; (experimental) adds some extra heap, but may cause slowdown
+  -D USERMOD_AUDIOREACTIVE
 
 lib_deps =
   #https://github.com/lorol/LITTLEFS.git
--- a/platformio.ini
+++ b/platformio.ini
@@ -196,7 +196,6 @@ build_flags =
   ; decrease code cache size and increase IRAM to fit all pixel functions
   -D PIO_FRAMEWORK_ARDUINO_MMU_CACHE16_IRAM48 ;; in case of linker errors like "section `.text1' will not fit in region `iram1_0_seg'"
   ; -D PIO_FRAMEWORK_ARDUINO_MMU_CACHE16_IRAM48_SECHEAP_SHARED ;; (experimental) adds some extra heap, but may cause slowdown
-  -D USERMOD_AUDIOREACTIVE
 
 lib_deps =
   #https://github.com/lorol/LITTLEFS.git
--- a/usermods/audioreactive/audio_reactive.h
+++ b/usermods/audioreactive/audio_reactive.h
@@ -119,10 +119,6 @@ static uint8_t inputLevel = 128;              // UI slider value
 #else
   uint8_t sampleGain = SR_GAIN;               // sample gain (config value)
 #endif
-//static uint8_t soundAgc = 1;                  // Automagic gain control: 0 - none, 1 - normal, 2 - vivid, 3 - lazy (config value)
-//static uint8_t audioSyncEnabled = 0;          // bit field: bit 0 - send, bit 1 - receive (config value)
-//static bool udpSyncConnected = false;         // UDP connection status -> true if connected to multicast group
-
 // user settable options for FFTResult scaling
 static uint8_t FFTScalingMode = 3;            // 0 none; 1 optimized logarithmic; 2 optimized linear; 3 optimized square root
 
@@ -146,24 +142,7 @@ const float agcSampleSmooth[AGC_NUM_PRESETS]  = {  1/12.f,   1/6.f,  1/16.f}; //
 // AGC presets end
 
 static AudioSource *audioSource = nullptr;
-//static volatile bool disableSoundProcessing = false;      // if true, sound processing (FFT, filters, AGC) will be suspended. "volatile" as its shared between tasks.
 static bool useBandPassFilter = false;                    // if true, enables a bandpass filter 80Hz-16Khz to remove noise. Applies before FFT.
-
-// audioreactive variables shared with FFT task
-// static float    micDataReal = 0.0f;             // MicIn data with full 24bit resolution - lowest 8bit after decimal point
-// static float    multAgc = 1.0f;                 // sample * multAgc = sampleAgc. Our AGC multiplier
-// static float    sampleAvg = 0.0f;               // Smoothed Average sample - sampleAvg < 1 means "quiet" (simple noise gate)
-// static float    sampleAgc = 0.0f;               // Smoothed AGC sample
-
-// peak detection
-// static bool samplePeak = false;      // Boolean flag for peak - used in effects. Responding routine may reset this flag. Auto-reset after strip.getMinShowDelay()
-// static uint8_t maxVol = 31;          // Reasonable value for constant volume for 'peak detector', as it won't always trigger (deprecated)
-// static uint8_t binNum = 8;           // Used to select the bin for FFT based beat detection  (deprecated)
-// static bool udpSamplePeak = false;   // Boolean flag for peak. Set at the same time as samplePeak, but reset by transmitAudioData
-// static unsigned long timeOfPeak = 0; // time of last sample peak detection.
-// static void detectSamplePeak(void);  // peak detection function (needs scaled FFT results in vReal[])
-// static void autoResetPeak(void);     // peak auto-reset function
-
 
 ////////////////////
 // Begin FFT Code //
@@ -176,17 +155,12 @@ void FFTcode(void * parameter);      // audio processing task: read samples, run
 static void runMicFilter(uint16_t numSamples, float *sampleBuffer);          // pre-filtering of raw samples (band-pass)
 static void postProcessFFTResults(bool noiseGateOpen, int numberOfChannels); // post-processing and post-amp of GEQ channels
 
-//#define NUM_GEQ_CHANNELS 16                                           // number of frequency channels. Don't change !!
-
 static TaskHandle_t FFT_Task = nullptr;
 
 // Table of multiplication factors so that we can even out the frequency response.
 static float fftResultPink[NUM_GEQ_CHANNELS] = { 1.70f, 1.71f, 1.73f, 1.78f, 1.68f, 1.56f, 1.55f, 1.63f, 1.79f, 1.62f, 1.80f, 2.06f, 2.47f, 3.35f, 6.83f, 9.55f };
 
 // globals and FFT Output variables shared with animations
-//static float FFT_MajorPeak = 1.0f;              // FFT: strongest (peak) frequency
-//static float FFT_Magnitude = 0.0f;              // FFT: volume (magnitude) of peak frequency
-//static uint8_t fftResult[NUM_GEQ_CHANNELS]= {0};// Our calculated freq. channel result table to be used by effects
 #if defined(WLED_DEBUG) || defined(SR_DEBUG)
 static uint64_t fftTime = 0;
 static uint64_t sampleTime = 0;
--- a/usermods/audioreactive/audio_reactive.h
+++ b/usermods/audioreactive/audio_reactive.h
@@ -1,6 +1,9 @@
 #pragma once
 
 #include "wled.h"
+
+#ifdef ARDUINO_ARCH_ESP32
+
 #include <driver/i2s.h>
 #include <driver/adc.h>
 
@@ -8,11 +11,9 @@
   #error This audio reactive usermod is not compatible with DMX Out.
 #endif
 
-#ifndef ARDUINO_ARCH_ESP32
-  #error This audio reactive usermod does not support the ESP8266.
 #endif
 
-#if defined(WLED_DEBUG) || defined(SR_DEBUG)
+#if defined(ARDUINO_ARCH_ESP32) && (defined(WLED_DEBUG) || defined(SR_DEBUG))
 #include <esp_timer.h>
 #endif
 
@@ -57,6 +58,50 @@
 
 #define MAX_PALETTES 3
 
+static volatile bool disableSoundProcessing = false;      // if true, sound processing (FFT, filters, AGC) will be suspended. "volatile" as its shared between tasks.
+static uint8_t audioSyncEnabled = 0;          // bit field: bit 0 - send, bit 1 - receive (config value)
+static bool udpSyncConnected = false;         // UDP connection status -> true if connected to multicast group
+
+#define NUM_GEQ_CHANNELS 16                                           // number of frequency channels. Don't change !!
+
+// audioreactive variables
+#ifdef ARDUINO_ARCH_ESP32
+static float    micDataReal = 0.0f;             // MicIn data with full 24bit resolution - lowest 8bit after decimal point
+static float    multAgc = 1.0f;                 // sample * multAgc = sampleAgc. Our AGC multiplier
+static float    sampleAvg = 0.0f;               // Smoothed Average sample - sampleAvg < 1 means "quiet" (simple noise gate)
+static float    sampleAgc = 0.0f;               // Smoothed AGC sample
+static uint8_t  soundAgc = 0;                   // Automagic gain control: 0 - none, 1 - normal, 2 - vivid, 3 - lazy (config value)
+#endif
+//static float    volumeSmth = 0.0f;              // either sampleAvg or sampleAgc depending on soundAgc; smoothed sample
+static float FFT_MajorPeak = 1.0f;              // FFT: strongest (peak) frequency
+static float FFT_Magnitude = 0.0f;              // FFT: volume (magnitude) of peak frequency
+static bool samplePeak = false;      // Boolean flag for peak - used in effects. Responding routine may reset this flag. Auto-reset after strip.getMinShowDelay()
+static bool udpSamplePeak = false;   // Boolean flag for peak. Set at the same tiem as samplePeak, but reset by transmitAudioData
+static unsigned long timeOfPeak = 0; // time of last sample peak detection.
+static uint8_t fftResult[NUM_GEQ_CHANNELS]= {0};// Our calculated freq. channel result table to be used by effects
+
+// TODO: probably best not used by receive nodes
+//static float agcSensitivity = 128;            // AGC sensitivity estimation, based on agc gain (multAgc). calculated by getSensitivity(). range 0..255
+
+// user settable parameters for limitSoundDynamics()
+#ifdef UM_AUDIOREACTIVE_DYNAMICS_LIMITER_OFF
+static bool limiterOn = false;                 // bool: enable / disable dynamics limiter
+#else
+static bool limiterOn = true;
+#endif
+static uint16_t attackTime =  80;             // int: attack time in milliseconds. Default 0.08sec
+static uint16_t decayTime = 1400;             // int: decay time in milliseconds.  Default 1.40sec
+
+// peak detection
+#ifdef ARDUINO_ARCH_ESP32
+static void detectSamplePeak(void);  // peak detection function (needs scaled FFT reasults in vReal[]) - no used for 8266 receive-only mode
+#endif
+static void autoResetPeak(void);     // peak auto-reset function
+static uint8_t maxVol = 31;          // (was 10) Reasonable value for constant volume for 'peak detector', as it won't always trigger  (deprecated)
+static uint8_t binNum = 8;           // Used to select the bin for FFT based beat detection  (deprecated)
+
+#ifdef ARDUINO_ARCH_ESP32
+
 // use audio source class (ESP32 specific)
 #include "audio_source.h"
 constexpr i2s_port_t I2S_PORT = I2S_NUM_0;       // I2S port to use (do not change !)
@@ -74,18 +119,10 @@ static uint8_t inputLevel = 128;              // UI slider value
 #else
   uint8_t sampleGain = SR_GAIN;               // sample gain (config value)
 #endif
-static uint8_t soundAgc = 1;                  // Automagic gain control: 0 - none, 1 - normal, 2 - vivid, 3 - lazy (config value)
-static uint8_t audioSyncEnabled = 0;          // bit field: bit 0 - send, bit 1 - receive (config value)
-static bool udpSyncConnected = false;         // UDP connection status -> true if connected to multicast group
+//static uint8_t soundAgc = 1;                  // Automagic gain control: 0 - none, 1 - normal, 2 - vivid, 3 - lazy (config value)
+//static uint8_t audioSyncEnabled = 0;          // bit field: bit 0 - send, bit 1 - receive (config value)
+//static bool udpSyncConnected = false;         // UDP connection status -> true if connected to multicast group
 
-// user settable parameters for limitSoundDynamics()
-#ifdef UM_AUDIOREACTIVE_DYNAMICS_LIMITER_OFF
-static bool limiterOn = false;                 // bool: enable / disable dynamics limiter
-#else
-static bool limiterOn = true;
-#endif
-static uint16_t attackTime =  80;             // int: attack time in milliseconds. Default 0.08sec
-static uint16_t decayTime = 1400;             // int: decay time in milliseconds.  Default 1.40sec
 // user settable options for FFTResult scaling
 static uint8_t FFTScalingMode = 3;            // 0 none; 1 optimized logarithmic; 2 optimized linear; 3 optimized square root
 
@@ -109,23 +146,23 @@ const float agcSampleSmooth[AGC_NUM_PRESETS]  = {  1/12.f,   1/6.f,  1/16.f}; //
 // AGC presets end
 
 static AudioSource *audioSource = nullptr;
-static volatile bool disableSoundProcessing = false;      // if true, sound processing (FFT, filters, AGC) will be suspended. "volatile" as its shared between tasks.
+//static volatile bool disableSoundProcessing = false;      // if true, sound processing (FFT, filters, AGC) will be suspended. "volatile" as its shared between tasks.
 static bool useBandPassFilter = false;                    // if true, enables a bandpass filter 80Hz-16Khz to remove noise. Applies before FFT.
 
 // audioreactive variables shared with FFT task
-static float    micDataReal = 0.0f;             // MicIn data with full 24bit resolution - lowest 8bit after decimal point
-static float    multAgc = 1.0f;                 // sample * multAgc = sampleAgc. Our AGC multiplier
-static float    sampleAvg = 0.0f;               // Smoothed Average sample - sampleAvg < 1 means "quiet" (simple noise gate)
-static float    sampleAgc = 0.0f;               // Smoothed AGC sample
+// static float    micDataReal = 0.0f;             // MicIn data with full 24bit resolution - lowest 8bit after decimal point
+// static float    multAgc = 1.0f;                 // sample * multAgc = sampleAgc. Our AGC multiplier
+// static float    sampleAvg = 0.0f;               // Smoothed Average sample - sampleAvg < 1 means "quiet" (simple noise gate)
+// static float    sampleAgc = 0.0f;               // Smoothed AGC sample
 
 // peak detection
-static bool samplePeak = false;      // Boolean flag for peak - used in effects. Responding routine may reset this flag. Auto-reset after strip.getMinShowDelay()
-static uint8_t maxVol = 31;          // Reasonable value for constant volume for 'peak detector', as it won't always trigger (deprecated)
-static uint8_t binNum = 8;           // Used to select the bin for FFT based beat detection  (deprecated)
-static bool udpSamplePeak = false;   // Boolean flag for peak. Set at the same time as samplePeak, but reset by transmitAudioData
-static unsigned long timeOfPeak = 0; // time of last sample peak detection.
-static void detectSamplePeak(void);  // peak detection function (needs scaled FFT results in vReal[])
-static void autoResetPeak(void);     // peak auto-reset function
+// static bool samplePeak = false;      // Boolean flag for peak - used in effects. Responding routine may reset this flag. Auto-reset after strip.getMinShowDelay()
+// static uint8_t maxVol = 31;          // Reasonable value for constant volume for 'peak detector', as it won't always trigger (deprecated)
+// static uint8_t binNum = 8;           // Used to select the bin for FFT based beat detection  (deprecated)
+// static bool udpSamplePeak = false;   // Boolean flag for peak. Set at the same time as samplePeak, but reset by transmitAudioData
+// static unsigned long timeOfPeak = 0; // time of last sample peak detection.
+// static void detectSamplePeak(void);  // peak detection function (needs scaled FFT results in vReal[])
+// static void autoResetPeak(void);     // peak auto-reset function
 
 
 ////////////////////
@@ -139,7 +176,7 @@ void FFTcode(void * parameter);      // audio processing task: read samples, run
 static void runMicFilter(uint16_t numSamples, float *sampleBuffer);          // pre-filtering of raw samples (band-pass)
 static void postProcessFFTResults(bool noiseGateOpen, int numberOfChannels); // post-processing and post-amp of GEQ channels
 
-#define NUM_GEQ_CHANNELS 16                                           // number of frequency channels. Don't change !!
+//#define NUM_GEQ_CHANNELS 16                                           // number of frequency channels. Don't change !!
 
 static TaskHandle_t FFT_Task = nullptr;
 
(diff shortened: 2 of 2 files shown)


gaaat98:
I have left the default ones

immagine

softhack007:
Screenshot_20240509-133320_Samsung Internet

@blazoncek question as I'm not very experienced with 8266 - sync receive polls for new UDP multicast packets each 20ms. This works well on esp32.

Maybe 8266 can  have a problem with it? Do you think that adding 'yield()' before/after UDP calls might help?

blazoncek:
I think @willmmiles or @pbolduc will know much more about AsyncUDP/TCP than me.

`fftBuff` is allocated on stack as far as I can see. That shouldn't be a problem. @gaaat98 please provide crash dump using exception decoder with your WDT crash.

gaaat98:
Excuse the dumb question, but how do I enable stack trace printing? The only thing I see on serial while resets occur is this:
```
Loop delayed more than 3ms.
Loop delayed more than 3ms.

 ets Jan  8 2013,rst cause:4, boot mode:(3,7)

wdt reset
load 0x4010f000, len 3424, room 16
tail 0
chksum 0x2e
load 0x3fff20b8, len 40, room 8
tail 0
chksum 0x2b
csum 0x2b
v000e0460
~ld
���'�{��n<�$�l b�|{�l�o��o�l ��{�d�l

�
```

blazoncek:
-D WLED_DEBUG

(which you have) and

monitor_filters = esp8266_exception_decoder

in your PIO environment.

But perhaps 8266 does not support WDT exceptions and does not print stack dump.
It looks to me that it may be possible that there is some loop that iterates across too many iterations (like -1 becoming 65535 when doing int16_t conversion to uint16_t).

pbolduc:
One of the biggest issues is the Espressif networking stack has problems with high UDP traffic rates.  It has been a while since I looked at the networking. I found this issue https://github.com/espressif/arduino-esp32/issues/4104 which talks about errors with the parsePacket function.  I need a reminder where the WifiUDP library comes from in WLED.

willmmiles:
> I need a reminder where the WifiUDP library comes from in WLED.

WifiUDP is part of the Arduino framework on ESP8266 and ESP32.   (The E131 receiver elsewhere in the project is using the ESPAsyncUDP library, which is *not* part of the framework.)

My quick read of the WifiUDP framework code on ESP8266 suggests it's reasonably well written.  By contrast, the ESP32 version is awful, heap allocating and freeing full 1460 byte packet buffer every time it polls to see if there's a new packet(!!). 

I pulled this PR in to my web server stability working branch and I haven't been able to reproduce the fault yet.  It's neat to see the AR effects working on an 8266 again, though it's noticably laggy compared to the other ESP8266 running as a DDP segment beside it.

@gaaat98 would you mind posting your config, in case there's something non-obvious I've got configured differently?

softhack007:
> I pulled this PR in to my web server stability working branch and I haven't been able to reproduce the fault yet

@willmmiles I think the problem is is happening with analog (PWM) strips only - may also happening with bitbang ws2812 drivers. So I suspects it's a timing issue, or simply the 8266 is overloaded when doing PWM + UDP polling + other stuff.

gaaat98:
@willmmiles [Here](https://github.com/Aircoookie/WLED/files/15282194/wled_cfg_ESP8285.json)'s the configuration on my athom controller (esp8285), on this board the crashes happen after a few updates when the clock speed is set to values higher than 'Slowest' (the effect I tested was DJ Light but I assume also the others will do the same).

[This](https://github.com/Aircoookie/WLED/files/15282198/wled_cfg_ESP8266.json) is the config of a ESP8266 board I used for testing, here the crashes are a bit less frequent but still happen. I noticed that if I open the /liveview page on a browser the number of reboots noticeably increases, which suggests that the boards crash due to overload.

These are both configured as PWM strips, I also have another athom controller connected to SK6812 strip with UDP sync enabled that shows no such issues

gaaat98:
@blazoncek That's basically what I did to obtain this log, the only difference is that I used a NodeMCU ESP8266 instead of D1 Mini, is that relevant?
> ```
> Loop delayed more than 3ms.
> Loop delayed more than 3ms.
> 
>  ets Jan  8 2013,rst cause:4, boot mode:(3,7)
> 
> wdt reset
> load 0x4010f000, len 3424, room 16
> tail 0
> chksum 0x2e
> load 0x3fff20b8, len 40, room 8
> tail 0
> chksum 0x2b
> csum 0x2b
> v000e0460
> ~ld
> ���'�{��n<�$�l b�|{�l�o��o�l ��{�d�l
> 
> �
> ---WLED 0.15.0-b3 2405030 INIT---
> esp8266 @ 80MHz.
> Core: 3.1.2
> FLASH: 2 MB
> heap 29072
> PIN ALLOC: Pin 1 successfully allocated by 0x89 (137)
> Registering usermods ...
> heap 28984
> ```
Before this Serial only showed other `Loop delayed more than 3ms.` and other generic debug info that was printed periodically.

The configuration used is the one I uploaded earlier:
> [This](https://github.com/Aircoookie/WLED/files/15282198/wled_cfg_ESP8266.json) is the config of a ESP8266 board I used for testing, here the crashes are a bit less frequent but still happen. I noticed that if I open the /liveview page on a browser the number of reboots noticeably increases, which suggests that the boards crash due to overload.

I repeated that once again and collected other logs [here: crash-log.txt](https://github.com/Aircoookie/WLED/files/15282894/crash-log.txt), clock speed is set to 'Fastest', WLED crashes after a while if I set to effect DJ Light from the Android app (the NodeMCU is receiving sound data from an esp32 which is listening to music), or crashes immediately if I enable the Peek option from the app.

willmmiles:
> > I pulled this PR in to my web server stability working branch and I haven't been able to reproduce the fault yet
> 
> @willmmiles I think the problem is is happening with analog (PWM) strips only - may also happening with bitbang ws2812 drivers. So I suspects it's a timing issue, or simply the 8266 is overloaded when doing high freq PWM (interrupt driven) + UDP polling + WiFi + other stuff.

I tried both bitbang ws2812 (actually my default ESP8266 test, due to a poor pin selection in a project I built before I discovered WLED!) and RGBW PWM in "normal" speed.  No crashes for me after 12h of operation.

> 
> Besides UDP polling, audioreactive runs a linear filter (no loops) then simply copies received data to the variables used in effects. So audioreactive on 8266 does "almost nothing" in addition to UDP.

I suspect the polling latency might be the source of the visible lag.  On 8266 calling the poll looks to be "basically free" in terms of CPU usage if there's nothing pending; we could probably poll every loop with no impact there (not so much on ESP32, though!!).  Stability is more important than performance, though, so I'll investigate this later.

willmmiles:
> clock speed is set to 'Fastest', WLED crashes after a while if I set to effect DJ Light from the Android app

Aha, there it is!  Just got one with these settings.

gaaat98:
@willmmiles during your test with the PWM strip did you select a sound reactive effect? Did you have another device sending the audio data? To have the sound reactive effects you need to make sure you pulled [PR#3942](https://github.com/Aircoookie/WLED/pull/3942) otherwise they are disabled.

Sorry to ask but it's stange since I have almost immediate crashes

willmmiles:
> @willmmiles during your test with the PWM strip did you select a sound reactive effect? Did you have another device sending the audio data? To have the sound reactive effects you need to make sure you pulled https://github.com/Aircoookie/WLED/pull/3942 otherwise they are disabled.

Yes and yes.  For the initial testing I had the same effect (GravCentric) running on three WS2812 strips - one on the ESP32 AR node directly, one on an ESP8266 over DDP, and one on an ESP8266 using AR sync.  That's how I could tell the AR sync was laggy compared to DDP. ;)

For the overnight test I reconfigured that last unit to add PWM output on some unconnected pins.  It ran OK for me using the GravCentric effect at 'Normal' speed.  Dunno if it was the speed, the DJ light effect, or peek, but it faulted right away when I set it up with that config ... though it hasn't faulted since.

blazoncek:
> ---WLED 0.15.0-b3 2405030 INIT---
> esp8266 @ 80MHz.

@gaaat98 try 160MHz builds.

gaaat98:
Same behaviour, I'm trying to establish reliable steps to reproduce the crashes but at the moment it seems that for me is enough to have the ESP8266 receiving data, the Dj Lights effect active, clock set to Fastest, and then just wait for maybe less than a minute while making some noises near the sending device...

Logs for the 160MHz build [here](https://github.com/Aircoookie/WLED/files/15283629/crash-log2.txt)

willmmiles:
I managed to get the crash to reliably reproduce, but I haven't yet pinpointed the root cause.  Here's what I've learned so far - might be old hat for some of you, sorry I'm still new to this environment:

- I was able to reproduce the crash using non-AR FX ('Breathe' in particular) on the PWM segment, so long as the brightness got above a certain point
- It won't reproduce without the UDP traffic, though
- `-D DEBUG_ESP_HWDT` gives a more helpful crash dump on WDT timer crashes such as these, but the user and system stacks are all munged together
- `-D DEBUG_ESP_HWDT_NOEXTRA4K` gives clean stack traces, but costs 4K of RAM
- Adding `DEBUG_PRINT(F("Reset for: ")); DEBUG_PRINTLN(ESP.getResetInfo());` to `WLED::setup()` gives yet more clues to what was going on when the system faulted
- None of these have left me with a smoking gun, or at least I haven't isolated it yet - the faulting instruction pointers seem to be all over the system and frequently "benign" (eg. register arithmetic, nothing that would obviously lead to a fault).
- .. though there's a frequent occurrence of a ROM function at the top of the system stack, `ram_rxiq_get_mis`, which I am guessing is for retrieving packets from the wifi module??

 I'm on vacation and away from home for the next couple of weeks, so I won't be able to do any more digging until I get back, but I wanted to leave some tracks in case anyone else was investigating.

Sorry I don't have more answers!

flux242:
please condisder the following implementation of the `receiveAudioData` function:
- The buffer is static now and its size is 83 bytes (size of the v1 packet)
- Fixed possible stack crash attack if somebody would send big UDP packets
- Added a while loop to make sure that if packets are sent too often it wouldn't cause out of ram crash.

```
    bool receiveAudioData()   // check & process new data. return TRUE in case that new audio data was received. 
    {
      if (!udpSyncConnected) return false;
      bool haveFreshData = false;

      size_t packetSize = 0u;
      while ( (packetSize = fftUdp.parsePacket()) !=0 )
      {
        if ( (packetSize == sizeof(audioSyncPacket)) || (packetSize == sizeof(audioSyncPacket_v1)) ) {
          //DEBUGSR_PRINTLN("Received UDP Sync Packet");
          static uint8_t fftBuff[sizeof(audioSyncPacket_v1)];
          fftUdp.read(fftBuff, packetSize);

          // VERIFY THAT THIS IS A COMPATIBLE PACKET
          if ( isValidUdpSyncVersion((const char *)fftBuff)) {
            decodeAudioData(packetSize, fftBuff);
            //DEBUGSR_PRINTLN("Finished parsing UDP Sync Packet v2");
            haveFreshData = true;
            receivedFormat = 2;
          }
          else if (isValidUdpSyncVersion_v1((const char *)fftBuff)) {
            decodeAudioData_v1(packetSize, fftBuff);
            //DEBUGSR_PRINTLN("Finished parsing UDP Sync Packet v1");
            haveFreshData = true;
            receivedFormat = 1;
          }
          else receivedFormat = 0; // unknown format
        }
      }
      return haveFreshData;
    }
```

softhack007:
> I managed to get the crash to reliably reproduce, but I haven't yet pinpointed the root cause. Here's what I've learned so far - might be old hat for some of you

🤔 so the only "smoking guns" we have so far:
- it's a watchdog reset -> so something took too long time or the wifi driver did not get it's regular time slice 
- it happens with PWM leds -> pwm on 8266 created a load of interrupts, because it does not have the "ledc" unit of esp32. 8266 basicially uses a timer that creates interrupts whenever the output needs to toggle (up to ~50 times per millisecond). PWM toggles most frequently when i density is around 50%
- there is no real "pattern" as to where it crashes -> does not look like a single loop taking g too much time

@willmmiles @gaaat98 can you try if adding a "yield()" before or after the wifiudp calls helps? I.e. somewhere here: 

https://github.com/gaaat98/WLED/blob/cec72419863a0ce8f19a02aa05bf9073b1302dcd/usermods/audioreactive/audio_reactive.h#L1086-L1090

receiving happens in "arduino loop" context so `yield()` is allowed.

softhack007:
> please condisder the following implementation of the `receiveAudioData` function:
> 

Thanks for sharing your ideas. Maybe before we look into memory management, let me explain a bit how the audio sync feature works:

- the sender creates a UDP _muticast_ channel. Multicast = one-to-many, but not the same as broadcast (one-to-all)
- the sender sends o e packet of 44 bytes each 20 millis
- only one sender should exist in your network!
- many receivers attach to the multicast channel.
- receivers poll for new packets each 10 millis (normally) so any "backlog" - if that exists - will be consumed quickly
- timing is important. If the receiver would unpack too many packets at once, this would result in "lagging" of effects
- so we are not sharing raw audio data, but audio processing results that's why 44 bytes each 20ms is enough



> * The buffer is static now and its size is 83 bytes (size of the v1 packet)

Sounds reasonable, however the buffer does not need to be `static` - fixed size means that it will be put onto the stack, while variable sized arrays are allocated on the heap

> * Fixed possible stack crash attack if somebody would send big UDP packets


Honestly, there are lots of easy ways to attack and crash an 8266 running WLED. I think we can concentrate on reasonable cases of normal operation. This includes "only one sender" and "no rogue packets".

> * Added a while loop to make sure that if packets are sent too often it wouldn't cause out of ram crash.


I think we have a timing problem (see my previous post) not mainly an issue with availeable heap. @willmmiles I'm not sure if wifiUdp will "perserve" unread packages ? Or do we always get the latest one from the wifi driver and old ones are discarded? Maybe on 8266 we need to "flush" the receive buffer after  each ` parsepacket() > 0`? Similar to what `udp.flush()` does on esp32...

gaaat98:
> @willmmiles @gaaat98 can you try if adding a "yield()" before or after the wifiudp calls helps? I.e. somewhere here:
> 
> https://github.com/gaaat98/WLED/blob/cec72419863a0ce8f19a02aa05bf9073b1302dcd/usermods/audioreactive/audio_reactive.h#L1086-L1090
> 
 I have added yield() after `size_t packetSize = fftUdp.parsePacket();` and also after `fftUdp.read(fftBuff, packetSize);` unfortunately the controller still crashes shortly after setting a sound reactive effect.

willmmiles:
Short update: I am still working on this!  I added some code to track interrupts and other events across HWDT crashes; unfortunately I'm still not seeing any consistency in the traces from one crash to the next.

I'm really wishing for that JTAG debugger, but I don't think I can get that working without rebuilding the whole toolchain. :(

willmmiles:
> @willmmiles I'm not sure if wifiUdp will "preserve" unread multicast packages? Or do we always get the latest one from the wifi driver and old ones are discarded? Maybe on 8266 we need to manually "flush" the receive buffer after each `parsepacket() > 0`? Similar to what `udp.flush()` does on esp32... for example by repeating wifiudp::read() until it returns "zero bytes read".

I removed the time check limiter in the poll function to see what happens.  There's definitely some evidence of malfeasance.  The udp packets seem to be released to us in bursts of five-- so there's definitely queuing going on somewhere -- and unfortunately they are not delivered very frequently.

willmmiles:
> The udp packets seem to be released to us in bursts of five-- so there's definitely queuing going on somewhere -- and unfortunately they are not delivered very frequently.

Ugh, it's worse than that.  I turned on the arduino core debugging, they're being delivered by lwip to WifiUDP in batches.  It will queue at max 5, but sometimes it's dropping two or three.  Our loop() function runs to completion many times in between these bursts, so we aren't to blame for hogging the CPU.  The interrupt traces show wifi interrupts also seem to be somewhat bursty.  I'll see if I can figure out how to tickle the network stack to get it to deliver things more smoothly.

willmmiles:
> > The udp packets seem to be released to us in bursts of five-- so there's definitely queuing going on somewhere -- and unfortunately they are not delivered very frequently.
> 
> Ugh, it's worse than that. I turned on the arduino core debugging, they're being delivered by lwip to WifiUDP in batches. It will queue at max 5, but sometimes it's dropping two or three. Our loop() function runs to completion many times in between these bursts, so we aren't to blame for hogging the CPU. The interrupt traces show wifi interrupts also seem to be somewhat bursty. I'll see if I can figure out how to tickle the network stack to get it to deliver things more smoothly.

OK, this turned out to be a wrinkle in multicast handling on wifi networks: the buffering was being done by my AP, not the ESP.  Apparently this is what happens if you allow the "modem sleep", so for any AR sync users, it probably needs to be turned off.  I don't know if we should maybe find some way to indicate that in the UI somewhere, or have the AR usermod page also set that setting.  Possibly we could also detect when this is happening (ie. when we receive two packets in one poll) and provide some indication in the info panel.

Unfortunately it gets worse, though.  For multicast packets, if **any** node on the SSID enables wifi power saving (aka modem sleep), the wifi standard mandates that *all* multicast packets must be buffered and batched, so they don't get missed by the one sleeping node.  The only workaround for this is to ask your AP to perform "multicast-to-unicast" conversion, if it supports it.  (Fortunately mine does!)

Chasing that down has been a great learning experience - I've learned a lot about the internals of xtos, wifi drivers, and the wifi standard - but alas the PWM crashes don't seem to be directly related and still happen even with correctly timed packet reception.  Onward!

softhack007:
>  there's definitely queuing going on somewhere -- and unfortunately they are not delivered very frequently.

> Unfortunately it gets worse, though. For multicast packets, if **any** node on the SSID enables wifi power saving (aka modem sleep), the wifi standard mandates that _all_ multicast packets must be buffered and batched, so they don't get missed by the one sleeping node. The only workaround for this is to ask your AP to perform "multicast-to-unicast" conversion, if it supports it. (Fortunately mine does!)

@willmmiles wow that's unexpected - thanks for the good research 😃 
It actually matches with another experience - we have this old note in our [sound sync wiki](https://mm.kno.wled.ge/soundreactive/sync/) about "make sure you have gone into the WiFi Preferences and clicked on 'Disable WiFi sleep' at the bottom of the page. It has caused us innumerable problems in the past". At least we finally know why that happens.

I'll try to find some time in the next week so I can join debugging 8266. My bets are still on timing issues, and maybe we need to carefully place a few `yield()` so that the wifi stack is not dying.

tablatronix:
I am running 015 on shelly rgbw2, and and having reboots also, currently debugging. I am using nothing special either, just fresh config, defaults. I tried testing all interfaces off, still sporadic reboots.

softhack007:
> I am running 015 on shelly rgbw2, and and having reboots also

Are you shelly bulbs "PWM" (analogue), or ws2812b?

tablatronix:
Oh sorry the shelly rgbw2 is 4 ch pwm controller

tablatronix:
I am going to have to add some debugging, there is nothing useful logged when this occurs..

willmmiles:
Today's update: I'm about 99% confident the crashes are caused by a bug in the PWM core.  I finally instrumented the timer NMI in the PWM code, and the event trace leading up to the crash shows that it's getting stuck continuously triggering the timer NMI in a 1000-cycle loop without ever advancing the main program, with all the wifi interrupt actively pending.

Having reviewed the framework code a bit, I'm not super surprised - using the NMI is clever, but required them to hand roll a lock-free update mechanism, a task fraught with peril for the best of us.  There's some code smells in there too - the Arduino API wrapper around their core functions is implemented in a way that forces two synchronizations for every PWM pin update.  And since we can't mask the NMI, "synchronization" means "put something in the update mailbox, then try to trigger an interrupt and yield() loop until it's handled". :(

Anyways, now that I can get some kind of logging out of it, I should be able to probe the internal state and pin it down completely.  The good news is that they've abstracted the PWM interface using weak symbols, so we're free to replace it locally without having to wait for an upstream patch.  🤞there could be some real progress on the weekend!

tablatronix:
Thanks, after exhausting everything related to build features, and as random as this seems, I was also looking into the PWM driver, nothing else has given me any reproducibility.

Is this in BusPwm::BusPwm(BusConfig &bc)? Or abstracted from one of the dependancies?

also let me know if anything needs testing, I am sitting here with a full A/B test running... bout to pull my hair out..

0DBAE349-61A5-43CC-99C2-C0E0307E2B92_1_105_c

blazoncek:
If I summarize all of the above and draw a line, there is nothing wrong with this PR and the crashing issue only occurs when using PWM output. Is this correct?

If yes, then I think we can still proceed with this PR. Is this fine with you @softhack007 ? Anyone else has anything else to say?

I am sure @willmmiles will figure out PWM bug as he has done so many other times.

willmmiles:
There were a couple of unrelated issues I ran in to along the way:
- Packet structure packing was different between ESP32 and ESP8266 compiles, so I couldn't rx any packets as sizeof() didn't match.  (Rather suprised this hasn't come up for anyone else?!?)  I have a [patch for this](https://github.com/willmmiles/WLED/commit/341b49fce30adeb6136b89d830b7741c9ea82a84) that explicitly defines the packet padding to enforce member alignment across architectures.
- I also wrote a version of the fixed-buffer-size patch [mentioned above](https://github.com/Aircoookie/WLED/pull/3962#issuecomment-2121876915).  I don't think my WIP will work correctly with the ESP32 WifiUDP, though -- IIRC the ESP32 implementation requires clients to flush() or read() all bytes to avoid getting out of sync.
- Strictly speaking, casting to a struct type from a stack allocated packet buffer of char or uint8 type isn't safe; it can violate the member alignment rules.  [I wrote the classical fix for this](https://github.com/willmmiles/WLED/commit/f4ec68e72b8ad45f28754a3982b62b40134e3468)  (memcpy from buffer to the struct type) in my debug branch, but it would save an extra copy to WifiUDP.read() directly in to a struct pointer.

Any or all of those could be a separate PR, though.

softhack007:
> * Packet structure packing was different between ESP32 and ESP8266 compiles, so I couldn't rx any packets as sizeof() didn't match.  (Rather suprised this hasn't come up for anyone else?!?)

@willmmiles I was not sure that packet sizes really match, however "it seemed to work". Actually there is a change that's still on my list to bring it back into AC wled:

https://github.com/MoonModules/WLED/blob/2edfcb33437386d32d8327a3c91b1c54c8df68e1/usermods/audioreactive/audio_reactive.h#L1050-L1062

With this, ssizeof(audioSyncPacket) should be 44 on both platforms. Can you double-check please?

softhack007:
Honestly speaking, memory management in the sound sync code (originally from atuline/wled) was always like "this doesn't look look right, but let's not touch it as everything might break down" 😆 so I'm not really surprised that the code has issues. I'll look into your patches after the weekend. Thanks 👍

willmmiles:
> https://github.com/MoonModules/WLED/blob/2edfcb33437386d32d8327a3c91b1c54c8df68e1/usermods/audioreactive/audio_reactive.h#L1050-L1062
> 
> With this, sizeof(audioSyncPacket) should be 44 on both platforms. Can you double-check please?

Confirmed this works too.  I see there's a couple of new fields in the old padding - any plans to upstream those features?

willmmiles:
Latest PWM update:  I think I've finally gotten to the bottom of it.  It looks like an actual, honest-to-goodness CPU bug.

The problem seems to be with the 'rfi 3' instruction used to return from the NMI used by the PWM timer.  Sometimes this seems to hang the CPU instead of correctly returning.  I don't know what causes it -- my best guess is contention with some other interrupt.

After the fault, the CPU stalls until the next NMI (timer event) occurs.  When it does so, EPC3, the "return address" from the NMI, is set pointing to the NMI's own rfi instruction.  When it tries to return from the "nested" interrupt, it goes in to a loop just resetting the instruction pointer back to the same 'return from interrupt' instruction, forever and ever, until the hard WDT trips.  The timer NMIs actually still fire -- so PWM still works in this stuck state! -- but since the return address is at the end of the NMI handler, it just goes back to its loop when the ISR completes.

The good news is that the timer NMIs can maybe be used to recover the processor!  I found a workaround by having the NMI handler back up the last "known good" return address, and restore it if the ISR finds the return address points to itself.  My initial tests of this workaround have been very encouraging; it hasn't crashed on me yet, when it would typically fail in seconds.

What's left to do is find the cleanest way to integrate this fix in to WLED.  My development approach involves duplicating and rewriting the processor's base interrupt vector table in assembly.  This is great for tracing interrupts, but pretty invasive and finicky.  The easiest place to do it would be the in platform PWM code - it'd be just a few lines of C, actually - but I don't relish forking the platform, and I'm not sure I could convince the upstream guys of the CPU fault: obviously the current code works well for them!   Alternatively, the platform core uses some linker tricks to allow overriding the PWM code; so we could pull the ESP8266 PWM code in to our repo with only a minimal space penalty, though we'd have to pay some extra attention to monitoring upstream changes when we update the platform.

I'm somewhat leaning towards the latter solution.   Anybody else got an opinion?

I'd also note that we the platform offers a phase-locked PWM core we should probably switch to - the way they've implemented the base PWM system produces pathologically bad performance if you change several outputs back to back, like we do for RGB(WW) outputs.  Each channel tries to keep its own phase, so you can often get separate nearly-back-to-back interrupts for every channel. :(   The phase-locked implementation should never generate more than analogFreq interrupts per second, while the base solution could theoretically generate as many as (analogFreq * active PWM channels)!

Thanks for bearing with me.  It's been quite an adventure getting this deep in to the ROM and platform code.

blazoncek:
@willmmiles ESP8266 is nearly at the end of its lifespan. Just put a band aid on it and be done with it.

Nevertheless, it would be nice to inform upstream of the findings. WLED used Tasmota framework in the past and @Jason2866 was very kind to us on many occasions so we could change to that if they incorporated such bugfix.

Jason2866:
We do not use core 3.x for esp8266. Tried the first releases. Was not stable for us. So we stayed on a modified v2.7.4 version which works well for us. There is no 3.x version from us and I won't do a 3.x based one. Tasmota based on esp8266 will end with our modified v2.7.x version

blazoncek:
Thank you @Jason2866 .
So the option is to revert core to 2.7.x but use newer toolchain (as we need newer C++ due to requirements from NeoPixelBus).
I do not know how to do that correctly so any help would be appreciated. There is a hint (I believe by @softhack007 or perhaps by me) in our `platformio.ini` for reverting to older version.

Jason2866:
@blazoncek Why should you want to revert? If the current version works well for you.
With a Platformio script you can patch the current core with the changes you want/need.

softhack007:
> I'd also note that we the platform offers a phase-locked PWM core we should probably switch to - the way they've implemented the base PWM system produces pathologically bad performance if you change several outputs back to back

From software perspektive, I would like to use the phase-locked PWM generator - looks like it can be activated by adding `-DPIO_FRAMEWORK_ARDUINO_WAVEFORM_LOCKED_PHASE` to build_flags? Another option seems to be just call ` enablePhaseLockedWaveform();`in our setup function.

Just one but - from electrical perspective. Phase-locked means that all PWM channels will be aligned, so more often we'll see all channels switch from 0 to 1 at the same time. This could become a problem for users with high current PWM setups using lots of color channels. If all PWM channels switch at the same time, its a sudden rise of power demand from the PSU. This may  lead to stronger "coil whining" tones emitted from cheap PSU models. So we should habe some beta tester (with long PWM strips having many color channels) feddback, before going that route.

softhack007:
@willmmiles just an idea about the "nearly-back-to-back interrupts for every channel". What if we slightly delay the starting of each PWM channel? I'm thinking of something like `analogWrite(RED, new_red); delay(1); analogWrite(GREEN, new_green); delay(3); analogWrite(BLUE, new_blue);`. If I'm right, this would bring the PWM channels a bit more "out of phase", making the "nearly-back-to-back interrupts" scenario less likely to happen?

blazoncek:
Regarding phase shifted PWM: I've had a discussion about bringing that to WLED with @jw2013. IMO that would be great addition.
But this is steering away from this PR. Shall we open a new issue for tracking PWM progress?

Jason2866:
We introduced lowering power supply load by "PWM auto-phasing" with this PR https://github.com/arendst/Tasmota/pull/14590 for Tasmota (for esp32x).

softhack007:
> But this is steering away from this PR. Shall we open a new issue for tracking PWM progress?

@blazoncek agreed. Either a new issue ticket, or maybe a new (draft) PR to also share related code.

For "8266 sound sync" we still need to know if special handling in the UDP code is necessary to prevent PWM crashes. 
Or maybe we need an automatic "kill switch" feature, that disables the usermod when PWM analog leds are used. 

For the "singled logical LED" PWM scenario, the 8266 could also be controlled by DDP and a sperate segment assigned to the one-led remote 8266. So audio processing would entirely be done on the esp32 having the microphone, while 8266 can handle PWM outputs without troubles.

softhack007:
> We introduced lowering power supply load by "PWM auto-phasing" with this PR [arendst/Tasmota#14590](https://github.com/arendst/Tasmota/pull/14590) for Tasmota (for esp32x).

nice 😄 could be something for our esp32 PWM driver, too.

@Jason2866 If i understand correctly, checking if several PWM outputs are actually "phase-locked" just requires a 12€ logic analyzer and   sigrok pulseview ?

Jason2866:
> > We introduced lowering power supply load by "PWM auto-phasing" with this PR [arendst/Tasmota#14590](https://github.com/arendst/Tasmota/pull/14590) for Tasmota (for esp32x).
> 
> nice 😄 could be something for our esp32 PWM driver, too.
> 
> @Jason2866 If i understand correctly, checking if several PWM outputs are actually "phase-locked" just requires a 12€ logic analyzer and sigrok pulseview ?

Yes, since you can test with a low PWM frequency (in our case 977Hz)

tablatronix:
I dropped my target fps to 30 and my stability has improved exponentially, might be anecdotal.. maybe it helps someone shrug

willmmiles:
> Either a new issue ticket, or maybe a new (draft) PR to also share related code.

I'll open a new PR tonight with a minimal PWM crash fix, vendoring in the soft PWM core currently in use.  We can follow it up with a second PR regarding analog phase improvements.  I believe the 8266 phase-locked core could be trivially modified to either (a) introduce a fixed phase offset for each line, or (b) introduce a "maximum transitions per tick" parameter.  Either way, separate PR.

>  just an idea about the "nearly-back-to-back interrupts for every channel" problem. What if we slightly delay the starting of each PWM channel? I'm thinking of something like analogWrite(RED, new_red); delay(1); analogWrite(GREEN, new_green); delay(3); analogWrite(BLUE, new_blue);. Use prime numbers for the delays, to avoid creating harmonics between shifted phases.

My $0.02 would be yes to phase shifting or spread spectrum, no to `delay()`.  The 8266 soft PWM code can let us specify the phase directly without losing FPS.  The PWM code from Tasmota linked above demonstrates how to do it for ESP32.
 
> For "8266 sound sync" we still need to know if special handling in the UDP code is necessary to prevent PWM crashes.

I'm fairly confident that the PWM problems could be triggered without the sound sync code at all, I believe that any kind of network traffic risks it.  The fast regular traffic from the sync code is exacerbating a pre-existing bug.  I'll post a beta PWM fix tonight and hopefully we can head off these problems.

platformio.ini
```diff
@@ -196,6 +196,7 @@ build_flags =
   ; decrease code cache size and increase IRAM to fit all pixel functions
   -D PIO_FRAMEWORK_ARDUINO_MMU_CACHE16_IRAM48 ;; in case of linker errors like "section `.text1' will not fit in region `iram1_0_seg'"
   ; -D PIO_FRAMEWORK_ARDUINO_MMU_CACHE16_IRAM48_SECHEAP_SHARED ;; (experimental) adds some extra heap, but may cause slowdown
+  -D USERMOD_AUDIOREACTIVE
```
blazoncek, line 199: I will require this to be removed and then we can proceed with PR.

gaaat98: Done!

softhack007 (approved):
@gaaat98 thanks, good job 👍 

Approved from my side.

usermods/audioreactive/audio_reactive.h
```diff
@@ -1049,9 +1090,12 @@ class AudioReactive : public Usermod {
...
 
       size_t packetSize = fftUdp.parsePacket();
-      if (packetSize > 5) {
+#ifdef ARDUINO_ARCH_ESP32
+      if ((packetSize > 0) && ((packetSize < 5) || (packetSize > UDPSOUND_MAX_PACKET))) fftUdp.flush(); // discard invalid packets (too small or too big) - only works on esp32
+#endif
+      if ((packetSize > 5) && (packetSize <= UDPSOUND_MAX_PACKET)) {
         //DEBUGSR_PRINTLN("Received UDP Sync Packet");
-        uint8_t fftBuff[packetSize];
+        static uint8_t fftBuff[UDPSOUND_MAX_PACKET+1] = { 0 }; // static buffer for receiving, to reuse the same memory and avoid heap fragmentation
```
softhack007, line 1098: NB: this line would also work without `static` - putting the buffer into stack, instead of using global space.
The important point is just to make it a fixed-size array, because variable size arrays are allocated from the heap.

I was not sure if it's better to use heap (no `static`), or global space (with `static`). @blazoncek I'll leave it up to you to decide what's  better for 8266.

blazoncek: Use stack if possible. It costs nothing and the array size is not problematic.

blazoncek:
I have not tested the code, but skimming it reveals no major issues. It has my "go".

@softhack007 merge it when you're comfortable with it.

netmindz:
Great to see we are almost there with this. Good work guys

blazoncek:
I'd like to add addendum to this PR.

My recent PR #4066 exploits ESP-NOW for WiFi-less syncing between devices. If audio sync would use ESP-NOW (in addition to UDP) that would make large-scale audio deployments a possibility.

netmindz:
I have a PoC of audio sync over ESP-NOW, which I was working on for the case where users had issue with multicast on their network @blazoncek 

https://github.com/MoonModules/WLED/pull/148/files

blazoncek:
@netmindz yes, something like that but hooked into framework provided in upstream implementation.

PR #3964, opened 2024-05-09
Update Usermod: Battery

adamsthws:
When taking the initial voltage reading after first powering on, voltage hasn't had chance to stabilize so the reading can be inaccurate, which in turn may incorrectly trigger the low-power preset. (Manifests when the user has selected a low read interval and/or is using a capacitor).

Resolution: A non-blocking, fixed 10 second delay has been added to the initial voltage reading to give the voltage time to stabilize.

Thankyou!

### DIFF
--- a/usermods/Battery/battery_defaults.h
+++ b/usermods/Battery/battery_defaults.h
@@ -14,6 +14,12 @@
   #endif
 #endif
 
+// The initial delay before the first battery voltage reading after power-on.
+// This allows the voltage to stabilize before readings are taken, improving accuracy of initial reading.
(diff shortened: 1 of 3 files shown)


adamsthws:
Quick notes...

This is a reworked version of the (now closed) PR here: https://github.com/Aircoookie/WLED/pull/3959

Thankyou to @blazoncek and @itCarl for your suggestions...

- Rebased the update for 0_15.

- Added a constant so the delay can be modified via my_config.h.

- Small adjustments to make the PR compatible again after the recent restructuring in this PR: (https://github.com/Aircoookie/WLED/pull/3003).

PR #3965, opened 2024-05-10
Fix missing conversions of bme280 values

LordMike:
The BME280 usermod uses a multiply-round-divide approach to cap the temperature/humidity/pressure values to some number of decimals. But the divide-part was missing in a few instances.

This shows up as the temperature/pressure being a 10-factor too high, depending on the config chosen in my case (BMP280 I2C chip).

Turns out the divide-by-10 was just missing from a few statements.


blazoncek:
Please rebase this for 0_15 branch.

LordMike:
Rebased. I also changed the target to `0_15` as well.

PR #3966, opened 2024-05-10
Add changeable i2c address to BME280 usermod

LordMike:
The BME280 usermod assumes an address of 0x76 which is incorrect for the BMP280 chip, which defaults to 0x77. This change puts the i2c address into the config and defaults to 0x76. 

Notes:
* ~~This builds on #3965. I'll rebase when that's merged.~~
* ~~As of right now, WLED must be rebooted for the address change to take effect. I'm not much of a C++ developer to know the ins and outs of reinitializing the `BME280I2C` instance .. do I stop it? delete it? .. If that was fixed, rebooting wasn't necessary.~~
* The config shows up as a text field with integers. It's my understanding that I can't manipulate with how the configs are shown, f.ex. to make a dropdown. Should it instead be a textual field which is then parsed in the code as a hex string -- or is decimals fine? ..

New config view. The `I2CAddress` is new - here set to 0x77.

image

### DIFF
--- a/usermods/BME280_v2/usermod_bme280.h
+++ b/usermods/BME280_v2/usermod_bme280.h
@@ -28,6 +28,7 @@ private:
   bool UseCelsius = true;                 // Use Celsius for Reporting
   bool HomeAssistantDiscovery = false;    // Publish Home Assistant Device Information
   bool enabled = true;
+  BME280I2C::I2CAddr i2cAddress = BME280I2C::I2CAddr_0x76;  // Default i2c address for BME280
 
   // set the default pins based on the architecture, these get overridden by Usermod menu settings
   #ifdef ESP8266
@@ -48,7 +49,7 @@ private:
       BME280I2C::I2CAddr_0x76 // I2C address. I2C specific. Default 0x76
   };
 
-  BME280I2C bme{settings};
+  BME280I2C bme;
 
   uint8_t sensorType;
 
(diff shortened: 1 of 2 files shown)


blazoncek:
Please do not force push changes.

usermods/BME280_v2/usermod_bme280.h
```diff
@@ -186,6 +187,9 @@ class UsermodBME280 : public Usermod
   {
     if (i2c_scl<0 || i2c_sda<0) { enabled = false; sensorType = 0; return; }
     
+    settings.bme280Addr = i2cAddress;
+    bme = BME280I2C(settings);
```
blazoncek, line 191: That is not correct way to initialise the sensor.
What this does is it uses default copy constructor and overwrites original BME object without de-initialising it first.

You should follow dynamic sensor start & stop as seen in other usermods.

blazoncek (changes requested):
I'm not happy with how sensor is initialised.

usermods/BME280_v2/usermod_bme280.h
LordMike: I figured out how to do this. This also fixes the issue with having to reboot. Now it correctly contacts the proper chip address and resets between config updates.

LordMike:
@blazoncek i just remembered. Am I not right in thinking the settings instance is not needed as a separate field?

To update it, I have to use setSettings() which means a copy is made in the bme instance. So having a copy of the settings in this class is just double the data. 

So I can remove the field with settings and create the settings every time I need it instead. Yes? :)

blazoncek:
According to the library documentation that may be sufficient.

LordMike:
Seems to work as well.

blazoncek (approved):

PR #3971, opened 2024-05-13
Update FX.cpp because of compile issue.

larryl79:
solution for ambiguous overload for 'operator==' error message

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -5224,7 +5224,7 @@ uint16_t mode_2Dmatrix(void) {                  // Matrix2D. By Jeremy Williams.
         oldSpawnColor = SEGMENT.getPixelColorXY(SEGENV.aux0, SEGENV.aux1);  // find color of previous spawns
         SEGENV.aux1 ++;                                                     // our sample pixel will be one row down the next time
     }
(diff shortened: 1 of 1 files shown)


blazoncek:
Please rebase this for 0_15 branch.

softhack007:
I'm wondering if you are using some exotic build environment or not recommended library versions. I've never seen this warning.

'CRGB::Black' is an enum from the 'CRGB' object, casting it to "CRGB" again should have zero effect.

softhack007:
CRGB::Black is not used in the matrix effect of `0_15` any more.


Closing this PR as its not applicable in 0_15. The reported problem is not happening in official builds of `main`, Any errors reported in FX.cpp would not depend on usermods in a different compilation unit.

larryl79:
> I'm wondering if you are using some exotic build environment or not recommended library versions. I've never seen this warning.
> 
> 'CRGB::Black' is an enum from the 'CRGB' object, casting it to "CRGB" again should have zero effect. Same for `pix` and `oldSpawnColor` - they are all "CRGB" already.

No, I just freshly downloaded the repo, and tried compile. I had about 400 warnings, and 2 errors as mentioned. Then I got the platformio override file from usermods folder. When I done it I never thought to publicate it, so I didn't done any pics, or evidences. After when I finished with the compile with usermods then I thought to add it.

PR #3977, opened 2024-05-13
Add a usermod for AHT10, AHT15 and AHT20 temperature/humidity sensors

LordMike:
This pull request introduces a new usermod for integrating the AHT10, AHT15, and AHT20 humidity and temperature sensors into WLED. It uses the [enjoyneering/AHT10](https://registry.platformio.org/libraries/enjoyneering/AHT10) library to handle the wire stuff, and implements settings like:

- I2CAddress: The i2c address in decimal. Set it to either 56 (0x38, the default) or 57 (0x39).
- SensorType: Numeric 0-2 for each of the valid sensor chips
- CheckInterval: Number of seconds between readings
- Decimals: Number of decimals to put in the output

Info view.
image

Settings view.
image


usermods/AHT10_v2/README.md
```diff
@@ -0,0 +1,34 @@
...
+extends = env:esp32dev
+build_flags =
+  ${common.build_flags} ${esp32.build_flags}
+  -D USERMOD_AHT10
+  ; -D USERMOD_AHT10_DEBUG ; -- add a debug status to the info modal
+lib_deps = 
+  ${esp32.lib_deps}
+  enjoyneering/AHT10@~1.1.0
+  Wire
+```
```
blazoncek, line 34: Please add your contact information.

usermods/AHT10_v2/usermod_aht10.h
```diff
@@ -0,0 +1,201 @@
...
+      if (_initDone)
+      {
+        // Reloading config
+        initializeAht();
+      }
+
+      _initDone = true;
+      return configComplete;
+    }
+};
```
blazoncek, line 199: Please also implement `getId()` method. I know it is not mandatory ATM but may be required in future enhancements.

usermods/AHT10_v2/usermod_aht10.h
```diff
@@ -0,0 +1,201 @@
...
+      if (currentTime - _lastLoopCheck < _checkInterval)
+        return;
+      _lastLoopCheck = currentTime;
+
+      _lastStatus = _aht->readRawData();
+      
+      if (_lastStatus == AHT10_ERROR)
+      {
+        // Perform softReset and retry
+        DEBUG_PRINTLN("AHTxx returned error, doing softReset");
```
blazoncek, line 72: Please enclose strings with `F()` macro. The exception are short strings used multiple times.

blazoncek (changes requested):
There are a couple of indentation errors as well as the in-line commented issues.

You might also want to add build flags and libraries to `platformio_override.sample.ini`

mike-csis:
@blazoncek is there an `.editorconfig` or other kind of editor support for proper indentation?

blazoncek:
VSC AFAIK detects it and uses appropriate indentation by default. You can enter anything, though.

LordMike:
I belive I've handled the issues. I used format document in vscode based on an editorconfig it came up with.

Quick note - the CI isn't building my usermod, right?..

LordMike:
Should I implement MQTT publishing, guarded by the "if mqtt compiled in" ifdefs?

blazoncek:
> Quick note - the CI isn't building my usermod, right?.. If you want, I could add a sample override file directly in the usermod folder, which could be referenced by a CI job for usermods. Something like the below.

Indeed. That would be nice as currently none of the usermods go through compilation process.

One more observation: Is it necessary to have "AHT Temperature" and "AHT Humidity"? There is no reason for "AHT" IMO.

blazoncek (approved):

blazoncek:
> Should I implement MQTT publishing, guarded by the "if mqtt compiled in" ifdefs?

That would be welcome by many users. Including HA autodiscovery broadcast.

LordMike:
I was not sure what to put there. In my setup, I'll have both the BME280, AHT20 and an INA226 (tbd). As the BME280 will provide "Temperature", I wasn't sure if also providing Temperature through the AHT was going to be an issue.

LordMike:
As I read it, the two usermods would both be creating the `u` dictionary in json and adding named keys to it. If both add "Temperature" .. as I read the JSON code now, the second caller would get a reference to the same array that the first one made. So in the end, the two usermods would put in 4 values - `[Temp 1, "C", Temp 2, "C"]`..

blazoncek:
Why on earth would you want two temperature sensors on one, LED driving, device?

LordMike:
Well. The device I got from _china_ had both in it to provide me with 3 values (temp, humidity and pressure).. So _yay me_ .. :D

Hilarious part is that a BMxxxx chip variant has all three values, but ... 🤷

blazoncek:
You can always disable one of them...

LordMike:
> You can always disable one of them...

Entirely true. Either code it so that temp isn't read from one of them, either a checkbox or "255" decimals meaning "don't". Or discard pressure and keep temp+humidity. :P

I'll have a look at the mqtt stuff. As for the naming of the ui json .. I can name it without AHT, but if the two usermods are run at the same time it's .. undefined..

LordMike:
So I:

- Added MQTT publishing
- HASS MQTT discovery publishing (stole code from BME280)
- Added settings that mimic that of BME280
- If MQTT is disabled, all of the above disappear
- I removed 5 instance bools in favor of a flags bitfield which is controlled through two macros

usermods/AHT10_v2/usermod_aht10.h
```diff
@@ -0,0 +1,320 @@
...
+
+#define AHT10_SUCCESS 1
+
+#define STATE_SETTING_ENABLED 0b00000001
+#define STATE_SETTING_MQTT_PUBLISH 0b00000010
+#define STATE_SETTING_MQTT_PUBLISHALWAYS 0b00000100
+#define STATE_SETTING_MQTT_HOMEASSISTANT 0b00001000
+#define STATE_INITDONE 0b00010000
+
+#define STATE_ISSET(stateFlags, state) (((stateFlags) & (state)) == (state))
```
blazoncek, line 14: Why on Earth would you want the ancient method of bitfield handling via macros? What's wrong with unnamed bitfield struct?

```c++
struct {
  bool enabled : 1;
  bool mqtt : 1;
  ...
}
...
if (enabled) {
  ...
  if (mqtt) {
    ...
  }
}
```
This code is much easier to follow.

usermods/AHT10_v2/usermod_aht10.h
```diff
@@ -0,0 +1,320 @@
...
+
+class UsermodAHT10 : public Usermod
+{
+private:
+  static const char _name[];
+
+  unsigned long _lastLoopCheck = 0;
+
+  // To avoid storing N bools, this flags field will handle all of those
+  uint8_t _stateFlags = 0;
```
blazoncek, line 25: Use unnamed bitfield struct instead.

usermods/AHT10_v2/usermod_aht10.h
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    if (_lastStatus == AHT10_SUCCESS)
+    {
+      float temperature = truncateDecimals(_aht->readTemperature(AHT10_USE_READ_DATA));
+      float humidity = truncateDecimals(_aht->readHumidity(AHT10_USE_READ_DATA));
+
+#ifndef WLED_DISABLE_MQTT
+      // Push to MQTT
+      mqttPublishIfChanged(F("temperature"), _lastTemperature, temperature);
```
blazoncek, line 164: That's an interesting approach. I've yet to encounter a sensor (apart from switch or button) that is not reporting measured values on a periodic base.

usermods/AHT10_v2/usermod_aht10.h
```diff
@@ -0,0 +1,320 @@
...
+    if (currentTime - _lastLoopCheck < _checkInterval)
+      return;
+    _lastLoopCheck = currentTime;
+
+    _lastStatus = _aht->readRawData();
+
+    if (_lastStatus == AHT10_ERROR)
+    {
+      // Perform softReset and retry
+      DEBUG_PRINTLN("AHTxx returned error, doing softReset");
```
blazoncek, line 147: F()?

usermods/AHT10_v2/usermod_aht10.h
```diff
@@ -0,0 +1,320 @@
...
+      char subuf[128];
+      snprintf_P(subuf, 127, PSTR("%s/%s"), mqttDeviceTopic, (const char *)topic);
+      mqtt->publish(subuf, 0, false, String(state).c_str());
+    }
+  }
+
+  // Create an MQTT Sensor for Home Assistant Discovery purposes, this includes a pointer to the topic that is published to in the Loop.
+  void mqttCreateHassSensor(const String &name, const String &topic, const String &deviceClass, const String &unitOfMeasurement)
+  {
+    String t = String(F("homeassistant/sensor/")) + mqttClientID + F("/") + name + F("/config");
```
blazoncek, line 93: No F() in "/" needed.

blazoncek (commented):
I'll neither approve nor request changes as technically the code looks correct (as far as logical analysis goes), but IMO it could be made clearer by using unnamed struct bitfields.

Also reexamine all strings, please. There are advantages and disadvantages when using F() macro.

You could also avoid inlining (reduced code size) by moving function definitions outside of class definition. See other usermods.

usermods/AHT10_v2/usermod_aht10.h
LordMike: I'm from a C# world - C++ is .. not my forté :)

usermods/AHT10_v2/usermod_aht10.h
LordMike: In my experience with other projects, they report values over MQTT when there are changes despite them doing periodic stuff. It reduces the load on other components. If you noticed, the BME280 usermod offers the same thing. I'm a bit worried on the comparison side though - I've seen it not report values (so float==float worked), but it is theoretically possible that `1.0 != 1.0`. I didn't have time to make a comparison with an epsilon boundary though.

usermods/AHT10_v2/usermod_aht10.h
LordMike: Missed that. I made a regex to find all strings - it obviously missed this one -.-'

usermods/AHT10_v2/usermod_aht10.h
blazoncek: Fair. Look at the code for Segment class for example.

usermods/AHT10_v2/usermod_aht10.h
blazoncek: depending on required precision consider roundf() or > and <

usermods/AHT10_v2/usermod_aht10.h
blazoncek: There is a penalty for using F() though. So strings (short) need to be searched throughout entire WLED code to see if they need F().

usermods/AHT10_v2/usermod_aht10.h
LordMike: I used the accuracy from the chips. It makes way more sense. :D

usermods/AHT10_v2/usermod_aht10.h
```diff
@@ -5,24 +5,22 @@
...
-  uint8_t _stateFlags = 0;
+  struct
+  {
+    bool settingEnabled : 1;    // Enable the usermod
+    bool mqttPublish : 1;       // Publish mqtt values
+    bool mqttPublishAlways : 1; // Publish always, regardless if there is a change
+    bool mqttHomeAssistant : 1; // Enable Home Assistant docs
+    bool initDone : 1;          // Initialization is done
+    unsigned : 3;
+  } _stateFlags;
```
blazoncek, line 23: FYI, there is no need to have struct object to use bitfields.

LordMike: Oh. Neat.

LordMike: Using these packed bits makes for way easier code. :)

Issue #3978, opened 2024-05-13
Palettes: Bad performance on 0.14.* and later


jw2013:
### What happened?

I discovered, that previous WLED binaries on ESP8266 were running at 80 MHz instead of 160 MHz.
Using it a lot, so I tested the newer _160 releases for my d1-mini's, but they were actually much slower!
String with 600 LEDs (WS2815) was now running at **13 FPS**, instead of the **20 FPS** using e.g. the 0.13.3 release.

So I curiously compiled 0.13.3 using board_build.f_cpu = 160000000L, and performance went up to **25 FPS**. Nice!

But why do the new releases perform so badly? I compiled basically all 0.14.* sources, and discovered that performance went downhill going from 0.14.1-b2 to 0.14.1-b3. Only real code change here was in json.cpp, and that one indeed causes the performance issue.

### To Reproduce Bug

Install firmware 0.14.1-b2 and 0.14.1-b3 to compare.
Setup LEDs with power limiter (WS2815), 600 LEDs (WS281x), Effect Glitter. Check FPS in info.

### Expected Behavior

Expected to keep the nice FPS rate.

### Install Method

Self-Compiled

### What version of WLED?

WLED 0.14.1-b2 and WLED-0.14.1-b3

### Which microcontroller/board are you seeing the problem on?

ESP8266

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jw2013:
@blazoncek , why did you add the "cannot reproduce" label? It is super simple to reproduce this!
For your convenience, please find attached example cfg.json and presets.json files.

[wled_presets.json](https://github.com/Aircoookie/WLED/files/15310270/wled_presets.json)
[wled_cfg.json](https://github.com/Aircoookie/WLED/files/15310269/wled_cfg.json)

600 LEDs, WS281X, using Glitter effect, on a regular ESP8266EX D1 mini module.

Using the release binaries provided here, I get the following results:

WLED_0.13.3_ESP8266.bin => 21 FPS
WLED_0.14.1-b2_ESP8266.bin => 12 FPS
WLED_0.14.1-b3_ESP8266.bin => 8 FPS

(self compiled WLED_0.13.3_ESP8266_160.bin => 25 FPS)

This is not just a minor performance issue, I hope you agree?
And replacing all ESP8266 and ESP8285 with ESP32 modules is NOT the correct fix.

jw2013:
[...]

PR #3979, opened 2024-05-15
Fix for the bad palette performance #3978

jw2013:
Fix for the bad performance #3978 caused by Segment::color_from_palette creating loading/creating palettes on each call.

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -91,6 +91,7 @@
 //#define SEGLEN           strip._segments[strip.getCurrSegmentId()].virtualLength()
 #define SEGCOLOR(x)      strip.segColor(x) /* saves us a few kbytes of code */
 #define SEGPALETTE       strip._currentPalette
+#define SEGPALETTEIDX    strip._currentPaletteIndex
 #define SEGLEN           strip._virtualSegmentLength /* saves us a few kbytes of code */
 #define SPEED_FORMULA_L  (5U + (50U*(255U - SEGMENT.speed))/SEGLEN)
 
@@ -571,7 +572,7 @@ typedef struct Segment {
     uint8_t  currentMode(void);                 // currently active effect/mode (while in transition)
     uint32_t currentColor(uint8_t slot);        // currently active segment color (blended while in transition)
     CRGBPalette16 &loadPalette(CRGBPalette16 &tgt, uint8_t pal);
-    CRGBPalette16 &currentPalette(CRGBPalette16 &tgt, uint8_t paletteID);
+    inline CRGBPalette16 &currentPalette(uint8_t paletteID);
 
     // 1D strip
     uint16_t virtualLength(void) const;
@@ -707,6 +708,7 @@ class WS2812FX {  // 96 bytes
 #endif
       // semi-private (just obscured) used in effect functions through macros
       _currentPalette(CRGBPalette16(CRGB::Black)),
+      _currentPaletteIndex(255),
       _colors_t{0,0,0},
       _virtualSegmentLength(0),
       // true private variables
@@ -902,6 +904,7 @@ class WS2812FX {  // 96 bytes
 
     void loadCustomPalettes(void); // loads custom palettes from JSON
     CRGBPalette16 _currentPalette; // palette used for current effect (includes transition)
+    uint8_t _currentPaletteIndex;
     std::vector<CRGBPalette16> customPalettes; // TODO: move custom palettes out of WS2812FX class
 
     // using public variables to reduce code size increase due to inline function getSegment() (with bounds checking)
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -438,18 +438,21 @@ uint32_t IRAM_ATTR Segment::currentColor(uint8_t slot) {
 #endif
 }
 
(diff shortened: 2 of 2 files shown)


wled00/FX.h
```diff
@@ -91,6 +91,7 @@
 //#define SEGLEN           strip._segments[strip.getCurrSegmentId()].virtualLength()
 #define SEGCOLOR(x)      strip.segColor(x) /* saves us a few kbytes of code */
 #define SEGPALETTE       strip._currentPalette
+#define SEGPALETTEIDX    strip._currentPaletteIndex
```
blazoncek, line 94: Wrong.
Current palette is not a `strip` property, it is an effect property and needs to be stored in Segment object.
See explanation below.

wled00/FX.h
```diff
@@ -571,7 +572,7 @@ typedef struct Segment {
     uint8_t  currentMode(void);                 // currently active effect/mode (while in transition)
     uint32_t currentColor(uint8_t slot);        // currently active segment color (blended while in transition)
     CRGBPalette16 &loadPalette(CRGBPalette16 &tgt, uint8_t pal);
-    CRGBPalette16 &currentPalette(CRGBPalette16 &tgt, uint8_t paletteID);
+    inline CRGBPalette16 &currentPalette(uint8_t paletteID);
```
blazoncek, line 575: Inlining will increase code size which is unacceptable as we are already at 98.5% flash size.

wled00/FX_fcn.cpp
```diff
@@ -438,18 +438,21 @@ uint32_t IRAM_ATTR Segment::currentColor(uint8_t slot) {
 #endif
 }
 
-CRGBPalette16 IRAM_ATTR &Segment::currentPalette(CRGBPalette16 &targetPalette, uint8_t pal) {
-  loadPalette(targetPalette, pal);
-  unsigned prog = progress();
+inline CRGBPalette16 &Segment::currentPalette(uint8_t pal) {
+  if ( SEGPALETTEIDX != pal ) {
```
blazoncek, line 442: This will not work during mode (aka effect) blending. Each mode (effect) can have its own palette that needs to be blended during execution of mode (effect).

wled00/FX_fcn.cpp
```diff
@@ -438,18 +438,21 @@ uint32_t IRAM_ATTR Segment::currentColor(uint8_t slot) {
...
+  }
+  uint16_t prog = progress();
   if (strip.paletteFade && prog < 0xFFFFU) {
     // blend palettes
     // there are about 255 blend passes of 48 "blends" to completely blend two palettes (in _dur time)
     // minimum blend time is 100ms maximum is 65535ms
-    unsigned noOfBlends = ((255U * prog) / 0xFFFFU) - _t->_prevPaletteBlends;
-    for (unsigned i=0; i<noOfBlends; i++, _t->_prevPaletteBlends++) nblendPaletteTowardPalette(_t->_palT, targetPalette, 48);
-    targetPalette = _t->_palT; // copy transitioning/temporary palette
+    uint16_t noOfBlends = ((255U * prog) / 0xFFFFU) - _t->_prevPaletteBlends;
```
blazoncek, line 451: `uint16_t` will unnecessarily increase code size and slow execution down as it adds `& 0xFFFF` to every operation.

blazoncek (changes requested):
Unfortunately I have to disapprove this PR as it negates effect and palette blending.

wled00/FX.h
jw2013: The idea was to use SEGPALETTE, which resolves to strip._currentPalette, as the palette cache.
Alternative would be to declare a new global variable for this cache, as I did before in the example.
(i thought that SEGPALETTE was still referred to, but not in active use, so wanted to 'reuse' it efficiently)

wled00/FX_fcn.cpp
jw2013: Good catch, we can keep both vars as 'unsigned'

jw2013:
@blazoncek , the change does not negate effect and palette blending, but just acts as a cache for loadPalette.

softhack007:
> > Unfortunately I have to disapprove this PR as it negates effect and palette blending.

> the change does not negate effect and palette blending, but just acts as a cache for loadPalette.


Hi @blazoncek and @jw2013, maybe before jumping directly into a possible solution, we could do some good old "design engineering"? 

From what I understood, 
* the problem can be described as "re-loading fixed palettes on each Segment::color_from_palette is inefficient, especially if loading from flash"
* Segment::color_from_palette is used very frequently in some effects
* the proposal is to add a caching mechanism, to prevent repeated loading of exactly the same palettes
* WLED has some features that lead to very frequently changing palettes: effect blending, palette blending, dynamic palettes (eg audioreactive palettes)
* personally, I thing that frequently re-loading dynamic palettes in the middle of drawing an effect is a bit of overkill.

<br/>

Assuming that some caching is possible there are design decisions to be made
 - where to place the cache memory (global, strip, segment) ?
 - when to update the cache (on-the fly in  color_from_palette, upfront once per frame in strip.service) ?
 - what are scenarios where caching must be skipped (like segment in transition)?
 - are there side-effects on other segment functions - copy&move constructors, transition object, etc
 

✏️  Opinions and comments, please!

blazoncek:
Obsolete by 77e6ea8

PR #3980, opened 2024-05-15
Palette performance fix #3978 backport to 0.14

### LINKED ISSUE #3978: Palettes: Bad performance on 0.14.* and later

jw2013:
Using that fix restores the performance issue created by color_from_palette recreating palette objects on each call.

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -92,6 +92,7 @@
 //#define SEGLEN           strip._segments[strip.getCurrSegmentId()].virtualLength()
 #define SEGCOLOR(x)      strip.segColor(x) /* saves us a few kbytes of code */
 #define SEGPALETTE       strip._currentPalette
+#define SEGPALETTEIDX    strip._currentPaletteIndex
 #define SEGLEN           strip._virtualSegmentLength /* saves us a few kbytes of code */
 #define SPEED_FORMULA_L  (5U + (50U*(255U - SEGMENT.speed))/SEGLEN)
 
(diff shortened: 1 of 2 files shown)


blazoncek:
Fundamental changes will not be ported to 0.14
Development has been stopped for 0.14.

PR #3984, opened 2024-05-16
Add USERMOD RC Switch - for controlling RC transmitter

JuMalIO:


usermods/rc_switch_v2/rc_switch_v2.h
```diff
@@ -0,0 +1,72 @@
+#pragma once
+
+#include "wled.h"
+
+#include <RCSwitch.h>
+
+#ifndef USERMOD_RC_SWITCH_PIN
+#ifdef ARDUINO_ARCH_ESP32
+#define USERMOD_RC_SWITCH_PIN 22
```
JuMalIO, line 9: Not sure which pins to use?

usermods/rc_switch_v2/rc_switch_v2.h
```diff
@@ -0,0 +1,72 @@
...
+    
+    void loop() {}
+
+    void readFromJsonState(JsonObject& root)
+    {
+      if (!initDone) return;  // prevent crash on boot applyPreset()
+
+      JsonObject usermod = root[FPSTR(_name)];
+      if (!usermod.isNull()) {
+        mySwitch.send(usermod["rcswitch"]);
```
JuMalIO, line 50: Also not sure if json `usermod["rcswitch"]` will be available and will it be a `char *` type as needed?

blazoncek: Cast it using `.to<const char *>()` method. It will return null pointer if it is not string.

usermods/rc_switch_v2/rc_switch_v2.h
blazoncek: Check online resources to determine available GPIO.
One would be RandomnerdTutorials.com

blazoncek (commented):
It would be nice if you'd provide dynamic configuration to allow run-time selection of GPIO.

usermods/rc_switch_v2/rc_switch_v2.h
JuMalIO: how exactly do I cast it?

`const char * test = usermod["rcswitch"].to<const char *>()`

I am getting error:

error: no matching function for call to 'ArduinoJson6181_90::MemberProxy<ArduinoJson6181_90::ObjectRef, const char*>::to()

blazoncek: Ah. sorry. It is `.as<const char *>()`.
Please consult ArduinoJson documentation.

JuMalIO:
Tested with 
```
POST /json/state
{"RcSwitch":{"payload":"1010001111101110010110000101110100001111000000000100001111110101"}}
```
And its working! Would be good to have this in master... 0_15 does not seem to build so I need to copy code each time...

dosipod:
@JuMalIO  what an excellent idea , do you have any specific RC transmitter link so we  could buy the same ?

JuMalIO:
> @JuMalIO what an excellent idea , do you have any specific RC transmitter link so we could buy the same ?
> 
> 433MHz ? I though RC uses 2.4GH , if that is RF433MHz then we have that even with a remote salvged from analog controllers , even blz wanted to buy one

I just bought cheap 433mhz transmitter module from Ali and it is working with my motorized roller blinds.

JuMalIO:
> It would be nice if you'd provide dynamic configuration to allow run-time selection of GPIO.

Why would you need to change a PIN? You still need to compile usermod any way?

blazoncek:
> Why would you need to change a PIN? You still need to compile usermod any way if you want this mod?

Because there are some people (and bots) that provide pre-compiled binaries. They will not do it for *every* GPIO possible for a single usermod.

And it will not get my approval until it gets this option.

dosipod:
@JuMalIO Not sure you have closed this to open a new one or you no longer interested in that .If you are still interested then we are able to use an  this RF433  usermod  https://github.com/athom-tech/Sound-Reactive-WLED/blob/master/usermods/RF433/RF433_V2.h And it is working as expected with on/off functionality 
image

Unfortunately i tried your usermod mainly due to mqtt but it was not working for me , if you closed this for that reason then we might give it a shot to adept mqtt  ( we may or may not submit a PR as that is a different matter then actually having a usermod working and available to the community )   .

PR #3986, opened 2024-05-17
Add INA226 usermod for reading current and power over i2c

LordMike:
This pull request introduces a new usermod for integrating the INA226 voltage, current and power sensors into WLED. It uses the [wollewald/INA226_WE](https://registry.platformio.org/libraries/wollewald/INA226_WE) library to handle the wire stuff, and implements settings like:

- **I2CAddress**: The I2C address in decimal. Default is 64 (0x40).
- **CheckInterval**: Number of seconds between readings. This should be higher than the time it takes to make a reading, determined by the two next options.
- **INASamples**: The number of samples to configure the INA226 to use for a measurement. Higher counts provide more accuracy. See the 'Understanding Samples and Conversion Times' section for more details.
- **INAConversionTime**: The time to use on converting and preparing readings on the INA226. Higher times provide more precision. See the 'Understanding Samples and Conversion Times' section for more details.
- **ShuntResistor**: Shunt resistor value in milliohms. An R100 shunt resistor should be written as "100", while R010 should be "10".
- **CurrentRange**: Expected maximum current in milliamps (e.g., 5 A = 5000 mA).

It also implements MQTT publishing and HA discovery documents.

**Images:**
- [Settings](https://github.com/Aircoookie/WLED/assets/1027111/761124d1-5b23-4f4a-9069-2fd96b2fd542)
- [Info view](https://github.com/Aircoookie/WLED/assets/1027111/a0071d2e-e35c-4a46-bd07-ad8a9b909ca6)
- [Info view in debug](https://github.com/Aircoookie/WLED/assets/1027111/c00948bc-973f-4057-8412-d209caf5268b)

### DIFF
--- a/usermods/INA226_v2/README.md
+++ b/usermods/INA226_v2/README.md
@@ -1,0 +1,36 @@
+# Usermod AHT10
+This Usermod is designed to read a `AHT10`, `AHT15` or `AHT20` sensor and output the following:
+- Temperature
+- Humidity
+
+Configuration is performed via the Usermod menu. The following settings can be configured in the Usermod Menu:
+- I2CAddress: The i2c address in decimal. Set it to either 56 (0x38, the default) or 57 (0x39).
+- SensorType, one of:
+  - 0 - AHT10
+  - 1 - AHT15
+  - 2 - AHT20
+- CheckInterval: Number of seconds between readings
+- Decimals: Number of decimals to put in the output
+
+Dependencies, These must be added under `lib_deps` in your `platform.ini` (or `platform_override.ini`).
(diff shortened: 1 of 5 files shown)


LordMike:
As this is likely to be used to measure currents and powers of the lights WLED controls, would it make sense to react to changes in the effects, on/off, brightness and then trigger a measurement, regardless of what interval is set?

F.ex. if I set the brightness to something new, then ~2 seconds later I get a measurement - even if I just got one on an interval of 60 seconds.

LordMike:
No notes? Wild. :)

PR #3993, opened 2024-05-20
Update wled.h regarding OTA Password

gsieben:
All important settings can be set in my_config.ini. Except the OTA password. This is always the default password. Here is the suggested correction.


blazoncek:
There was a talk on Discord about that possibility a while ago and the conclusion was to not implement it.
That is due to the fact that many new/inexperienced users are unaware of the possibility of different builds and it would make troubleshooting when you need AP mode difficult and frustrating for them if the default AP password was changed..

blazoncek:
Exactly the point!

As far as WLED goes there is no need for custom passwords. That's how I see it from the support POV.

What you do in your custom build/fork is up to you. The problem arises when random people try random build and then request help on the official channels with no clue what factory reset will do.

blazoncek:
I am just debating why it wasn't already implemented. I am not saying that PR is not being considered. You chose to close the PR, not me.

> And having the same OTA password everywhere is not recommended anyway.

And how do you propose to do that on default installations while keeping uneducated users "happy"?
AFAIK changing passwords from *default* ones is the first security minded task *anyone* should do.

Other than that, WLED is not intended to be used on public networks (and that's clearly stated). OTA lock, PIN or settings lock are there to prevent *accidental* change or firmware updates, they are not meant to "secure" the device as WLED lacks any kind of SSL or TLS implementation to encrypt the transmission of sensitive data.

PR #3994, opened 2024-05-20
Usermod BME68X Sensor Implementation rebased for WLED 0.15

gsieben:
Here is again my pull request for the user mod "BMW68X" sensor implementation. Now as requested for wled version 0.15.


blazoncek (changes requested):
Thanks. Well written and complete. But...

There's a code that does not belong in a usermod PR. You might re-open the original PR for that instead.

blazoncek (approved):

PR #4010, opened 2024-06-01
fix: esp32 s3 zero rev v0.2 wifi not working

cstruck:
This PR solves problems with the newly added support for `esp32s3_4M_qspi` for some esp32-s3 chip versions, where the wifi won't work.

I have two different esp32-s3 zero dev boards.

- ESP32-S3 (QFN56) (revision v0.2) seems to need the `LOLIN_WIFI_FIX` to get the wifi working.
- ESP32-S3 (QFN56) (revision v0.1) seems to work without the fix but also works perfectly fine with it.

It's a simple fix since it already exists for some other esp32 chip types by adding the `LOLIN_WIFI_FIX` build variable.


bentodman:
I just want to make a note here for anyone that finds this page. 
The Seeed XIAO ESP32S3 seems to require this fix for SOME of its boards. Ive had 30 of them and nearly half needed this fix to work which made it seem like a hardware issue to start with. 

If your wifi doesn't connect or show an AP try this!

PR #4011, opened 2024-06-03
Add new options for Artnet / sACN

Bauteiltoeter:
This PR adds two new options to the available options for sACN / Artnet Interface.

The two new options make it possible to use Artnet to switch between Preset selection and and pixel-control. 
For this the first channel is used. 
If channel 1 is between 0 and 126 WLED is in Preset mode.
In this case:
Channel 2: Preset ID
Channel 3: Brightness

If channel 1 is between 127 and 255 WLED is in direct pixel mode (RGB or RGBW).
In this case:
Channel 2-4: unused
Starting from channel 5: RGB / RGBW pixel data.


blazoncek:
Please rebase your PR for 0_15 branch.

netmindz:
It's rather unusual for a dmx fixture to change it's whole personality based on a dmx value, that's what RDM is for. 

If we were to make this change then it would be nice if we could handle all possible modes than just preset/rgb

Bauteiltoeter:
I didn't see it as "changing the personality" (which in fact is unusual), but I see it as "integrated effect" vs "cell / pixel mode." 
There are a lot of fixtures that have an macro channel / mode (brightness + internal effect type) additional to the single cell access. This is why I implemented the switch between "single cell" and "predefined effects".

I'll have to rebase this PR which is a bit of a pain because of the autogenerated files.

willmmiles:
I'm really not a fan of the code duplication.  (Sometimes I think copy-and-paste ought to be banned from IDEs!)

I'll second the suggestion regarding supporting all possible modes instead of limiting to one or the other.  Instead of special casing a couple, maybe try something like:

```c++
auto activeDMXMode = DMXMode;
if (DMXMode == DMX_MODE_SELECTABLE) {
    activeDMXMode = e131_data[dataOffset] / 10;   // default case will catch invalid values
    dataOffset += 1;
    availDMXLen -= 1;
}

switch (activeDMXMode) { 
   // existing mode logic here
}
```

PR #4013, opened 2024-06-05
Add LD2410 sensor usermod

wesleygas:
Adds a usermod that reports from a LD2410 movement/presence sensor via MQTT and (optionally) presents it to Home Assistant discovery


thepuncake:
Any chance we can get more customization options for the sensor?

PR #4018, opened 2024-06-08
Added Cube Mapping Tool

Brandon502:
Excel file that can also be used in google sheets to help map 3 panels (8x8 or 16x16) into a cube. Input your wiring over the current example and copy the ledmap created into WLED/edit.


netmindz:
Why does this need a rebase @softhack007 if it's just adding an excel file?

PR #4033, opened 2024-06-26
Usermod Updated: Internal Temperature V2

adamsthws:
Hello!

Please may I request to add the following improvement to the usermod: 'Internal Temperature V2'...

### Added high temperature indicator/action

- A configurable preset is activated when the internal temperature raises above a configurable threshold temperature.

- When the internal temperature falls back below the threshold, the previously active playlist/preset/state is re-activated.

- If the `preset to trigger` is set to ` 0 `, no preset is activated when temperature rises above the threshold.

- A small temperature buffer prevents frequent toggling between states when the temperature is close to the threshold.
   (The reset threshold is automatically calculated to be two degrees lower than whatever the activation threshold is set to).

- To prevent the user setting the loop interval too low, a minimum allowable interval has been added.

### To do

- In a later phase I plan to add ` isAboveThreshold ` to MQTT.

Thankyou!
Adam Matthews


usermods/Internal_Temperature_v2/readme.md
```diff
@@ -1,17 +1,48 @@
 # Internal Temperature Usermod
-This usermod adds the temperature readout to the Info tab and also publishes that over the topic `mcutemp` topic.
 
-## Important
-A shown temp of 53,33°C might indicate that the internal temp is not supported.
 
-ESP8266 does not have a internal temp sensor
+<p align="left">
```
blazoncek, line 4: Please remove HTML formatting. This is a Markdown document.

usermods/Internal_Temperature_v2/usermod_internal_temperature.h
```diff
@@ -6,21 +6,34 @@ class InternalTemperatureUsermod : public Usermod
 {
 
 private:
+  static const unsigned long minLoopInterval = 1000;  // minimum allowable interval (ms)
```
blazoncek, line 9: This could be `constexpr`.

usermods/Internal_Temperature_v2/usermod_internal_temperature.h
```diff
@@ -6,21 +6,34 @@ class InternalTemperatureUsermod : public Usermod
...
   unsigned long loopInterval = 10000;
   unsigned long lastTime = 0;
   bool isEnabled = false;
   float temperature = 0;
+  int presetToActivate = -1;            // Preset to activate when temp goes above threshold (-1 = disabled)
+  float activationThreshold = 95.0;     // Temperature threshold to trigger high-temperature actions
```
blazoncek, line 15: Use `float` constants. I.e. `95.0f`

usermods/Internal_Temperature_v2/usermod_internal_temperature.h
```diff
@@ -6,21 +6,34 @@ class InternalTemperatureUsermod : public Usermod
...
   static const char _enabled[];
   static const char _loopInterval[];
+  static const char _activationThreshold[];
+  static const char _presetToActivate[];
 
   // any private methods should go here (non-inline method should be defined out of class)
   void publishMqtt(const char *state, bool retain = false); // example for publishing MQTT message
 
+  //  Makes sure the measurement interval can't be set too low
+  void setSafeLoopInterval(unsigned long newInterval) {
```
blazoncek, line 29: There is no benefit to this as it is inlined by compiler.

usermods/Internal_Temperature_v2/usermod_internal_temperature.h
```diff
@@ -6,21 +6,34 @@ class InternalTemperatureUsermod : public Usermod
...
+    loopInterval = max(newInterval, minLoopInterval);
+    }
+
 public:
   void setup()
   {
+    setSafeLoopInterval(loopInterval);  // Initialize with a safe loop interval
```
blazoncek, line 36: No need for this as it is done in `readFromConfig()`.

usermods/Internal_Temperature_v2/usermod_internal_temperature.h
```diff
@@ -41,6 +55,30 @@ class InternalTemperatureUsermod : public Usermod
...
 
+    // Check if temperature has gone above the threshold
+    if (temperature >= activationThreshold) {
+      // Update the state flag if not already set
+      if (!isAboveThreshold){
+        isAboveThreshold = true;
+        }
+      // Activate the 'over-threshold' preset if it's not already active
+      if (presetToActivate != -1 && currentPreset != presetToActivate) {
+        saveTemporaryPreset();  // Save the current preset to allow re-activation later
```
blazoncek, line 66: This does not save *preset ID*. It only saves current state in temporary preset.

usermods/Internal_Temperature_v2/usermod_internal_temperature.h
```diff
@@ -41,6 +55,30 @@ class InternalTemperatureUsermod : public Usermod
...
+      }
+    // Check if temperature is back below the threshold
+    else if (temperature <= (activationThreshold - resetMargin)) {
+      // Update the state flag if not already set
+      if (isAboveThreshold){
+        isAboveThreshold = false;
+        }
+      // Revert back to the original preset
+      if (currentPreset == presetToActivate){
+        applyTemporaryPreset(); // Restore the previously stored active preset
```
blazoncek, line 78: This will not restore previous preset ID.

usermods/Internal_Temperature_v2/usermod_internal_temperature.h
```diff
@@ -80,15 +118,30 @@ class InternalTemperatureUsermod : public Usermod
...
 
+    // Append useful info to the usermod settings gui
+    void appendConfigData()
+    {
+    // Display 'ms' next to the 'Loop Interval' setting
+    oappend(SET_F("addInfo('Internal Temperature:Loop Interval', 1, 'ms');"));
+    // Display '°C' next to the 'Activation Threshold' setting
+    oappend(SET_F("addInfo('Internal Temperature:Activation Threshold', 1, '°C');"));
+    // Display '-1 = Disabled' next to the 'Preset To Activate' setting
+    oappend(SET_F("addInfo('Internal Temperature:Preset To Activate', 1, '-1 = disabled');"));
```
blazoncek, line 133: *Disabled*? Perhaps *unused*.

FYI there is no preset with ID 0.

blazoncek (changes requested):
There are some things that need to be changed.

When you want to temporarily store *current* preset, you are better of by saving `currentPreset` variable. If that is 0 then it is safe to save current state using `saveTemporaryPreset()`.

I'd also like to invite @lost-hope to do a peer review.

usermods/Internal_Temperature_v2/usermod_internal_temperature.h
adamsthws: I saw this done in the Battery usermod, would you like me to remove it?

blazoncek: Usermod's `loop()` is executed as many times as possible with no predetermined frequency. It is up to you to "schedule" *regular* processing within it.

With this comment I was referring to the fact that any function definition in the definition of a class is inlined as if you would add `inline` keyword. To avoid this you can define the function outside class definition or, in this case, simply use assignment in places where function is called (this not being wise always).

blazoncek:
There ar still a few `<br>` in readme. You missed some float constants and preset id could be `uint8_t` to save a bit of RAM (not mandatory).

adamsthws:
> There ar still a few `<br>` in readme. You missed some float constants and preset id could be `uint8_t` to save a bit of RAM (not mandatory).

 Sorry I missed these... Thanyou for being so thorough with your review!

usermods/Internal_Temperature_v2/usermod_internal_temperature.h
adamsthws: Thankyou for the helpfull explination Blaz... Much appreciated! :)

adamsthws:
> There are some things that need to be changed.
> 
> When you want to temporarily store _current_ preset, you are better of by saving `currentPreset` variable. If that is 0 then it is safe to save current state using `saveTemporaryPreset()`.
> 
> I'd also like to invite @lost-hope to do a peer review.

In addition to storing current preset, if a playlist is active, is it also possible to store the current playlist (and position in playlist?) to resume from the same place later?

blazoncek (approved):

blazoncek:
> In addition to storing current preset, if a playlist is active, is it also possible to store the current playlist (and position in playlist?) to resume from the same place later?

Yes, use `currentPlaylist` which is set if non-zero.

adamsthws:
> > In addition to storing current preset, if a playlist is active, is it also possible to store the current playlist (and position in playlist?) to resume from the same place later?
> 
> Yes, use `currentPlaylist` which is set if non-zero.

This is now implimented, thankyou for your guidance and patience Blaz!

blazoncek:
@lost-hope is this ok with you?

lost-hope:
> @lost-hope is this ok with you?

Yep looks fine to me.

PR #4035, opened 2024-06-28
Fix PWM crashes on ESP8266

willmmiles:
Vendor in the ESP8266 Arduino core PWM library, with a fix for a nasty NMI crash bug.  Sometimes the NMI return instruction seems to fail, resulting in an infinite loop as the PC gets stuck "returning" to the NMI handler itself.  Work around this by backing up and restoring the PC (and PS) from the last known good target values if needed.

I've opted to place the code in the platformio "libs" folder, which makes it easy to conditionally include for ESP8266 builds via platformio.ini.

Fixes PWM crash issues issues described in #3962.

### DIFF
--- a/lib/ESP8266PWM/src/core_esp8266_waveform_pwm.cpp
+++ b/lib/ESP8266PWM/src/core_esp8266_waveform_pwm.cpp
@@ -1,0 +1,711 @@
+/* esp8266_waveform imported from platform source code
+   Modified for WLED to work around a fault in the NMI handling,
+   which can result in the system locking up and hard WDT crashes.
+*/
+
+/*
+  esp8266_waveform - General purpose waveform generation and control,
+                     supporting outputs on all pins in parallel.
+
+  Copyright (c) 2018 Earle F. Philhower, III.  All rights reserved.
+
+  The core idea is to have a programmable waveform generator with a unique
+  high and low period (defined in microseconds or CPU clock cycles).  TIMER1
+  is set to 1-shot mode and is always loaded with the time until the next
+  edge of any live waveforms.
+
+  Up to one waveform generator per pin supported.
+
+  Each waveform generator is synchronized to the ESP clock cycle counter, not
+  the timer.  This allows for removing interrupt jitter and delay as the
+  counter always increments once per 80MHz clock.  Changes to a waveform are
+  contiguous and only take effect on the next waveform transition,
+  allowing for smooth transitions.
+
+  This replaces older tone(), analogWrite(), and the Servo classes.
+
+  Everywhere in the code where "cycles" is used, it means ESP.getCycleCount()
+  clock cycle count, or an interval measured in CPU clock cycles, but not
+  TIMER1 cycles (which may be 2 CPU clock cycles @ 160MHz).
+
+  This library is free software; you can redistribute it and/or
+  modify it under the terms of the GNU Lesser General Public
+  License as published by the Free Software Foundation; either
+  version 2.1 of the License, or (at your option) any later version.
+
+  This library is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


blazoncek:
@tablatronix @softhack007 @gaaat98 can you test this PR please?

gaaat98:
Hi! Sorry for the late reply but I was travelling. I did not do any stress test, but I can say that with the fix, the ESP8266 PWM light did not crash when UDP synced with an ESP32, even with the clock speed set to fastest. Before, it crashed in a few seconds, but now I was able to have it working for more than 10 minutes with sound reactive effects going! Thanks a lot for your great work and research effort!!

blazoncek:
@softhack007 any second thoughts before merging?
@tablatronix were you able to do any testing?

platformio.ini
```diff
@@ -202,6 +202,7 @@ lib_deps =
   #https://github.com/lorol/LITTLEFS.git
   ESPAsyncTCP @ 1.2.2
   ESPAsyncUDP
+  ESP8266PWM
```
softhack007, line 205: I think that code from "lib" is even compiled when it's not in `lib_deps` (but maybe I'm wrong...)

lib/ESP8266PWM/src/core_esp8266_waveform_pwm.cpp
```diff
@@ -0,0 +1,711 @@
...
+  This library is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+  Lesser General Public License for more details.
+
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, write to the Free Software
+  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA
+*/
+
```
softhack007, line 45: You could add `#ifndef ARDUINO_ARCH_ESP32` ... `#endif` around the file, to make sure it's never used on esp32.

lib/ESP8266PWM/src/core_esp8266_waveform_pwm.cpp
```diff
@@ -0,0 +1,711 @@
+/* esp8266_waveform imported from platform source code
+   Modified for WLED to work around a fault in the NMI handling,
+   which can result in the system locking up and hard WDT crashes.
+*/
```
softhack007, line 4: Please add a URL or exact release name for the original file. This might save a few hair later, in case we want to follow an updated arduino-8266 release...

It might also help if you mark your changes with a comment like ` //  willmiles restore lost processor context`, or put your changes between `// begin patch` ..... `// end patch`

platformio.ini
willmmiles: It is not!  That's expressly why I put it there, to avoid having to add #ifdef guards so as to keep the source as close as possible to the upstream version.  The local `libs` folder is merely the first place in the `lib_deps` search path, so local libraries will shadow upstream ones.  It looks like it was intended as an easy place for vendoring in forked libraries.

lib/ESP8266PWM/src/core_esp8266_waveform_pwm.cpp
willmmiles: Done.  Thanks for the review!

softhack007 (approved):
👍 looks good for me, thanks

tablatronix:
I have not, my rgb lights are now in my pool, but I just got some small ones for bench testing. Will try to try to wire a tester up

Moonbase59:
FYI: This apparently also fixed my Athom 15W bulbs crashing every few seconds when controlled by LedFx. Installed WLED_0.15.0-b5_ESP02_160.bin.gz on these, and crashing problems are gone. Thanks!

PR #4036, opened 2024-06-29
Enabled pin 18-19 as LED output for ESP32-C3

zalatnaicsongor:
The ESP32-C3 has its usb pins mapped to GPIO 18-19. However, these pins can be freely reused if not used for USB purposes.

This makes WLED compatible with the SP530E controller sold by BTF-Lighting and other sellers on Aliexpress.

Documentation on how to flash the SP530E is to follow.

### DIFF
--- a/wled00/pin_manager.cpp
+++ b/wled00/pin_manager.cpp
@@ -243,7 +243,6 @@ bool PinManagerClass::isPinOk(byte gpio, bool output) const
   #if defined(CONFIG_IDF_TARGET_ESP32C3)
     // strapping pins: 2, 8, & 9
     if (gpio > 11 && gpio < 18) return false;     // 11-17 SPI FLASH
-    if (gpio > 17 && gpio < 20) return false;     // 18-19 USB-JTAG
   #elif defined(CONFIG_IDF_TARGET_ESP32S3)
     // 00 to 18 are for general use. Be careful about straping pins GPIO0 and GPIO3 - these may be pulled-up or pulled-down on your board.
     if (gpio > 18 && gpio < 21) return false;     // 19 + 20 = USB-JTAG. Not recommended for other uses.


Noschvie:
> Documentation on how to flash the SP530E is to follow.

... interested in this, thanks!

zalatnaicsongor:
> > Documentation on how to flash the SP530E is to follow.
> 
> ... interested in this, thanks!

https://github.com/Aircoookie/WLED-Docs/pull/214

wled00/pin_manager.cpp
```diff
@@ -243,7 +243,6 @@ bool PinManagerClass::isPinOk(byte gpio, bool output) const
   #if defined(CONFIG_IDF_TARGET_ESP32C3)
     // strapping pins: 2, 8, & 9
     if (gpio > 11 && gpio < 18) return false;     // 11-17 SPI FLASH
-    if (gpio > 17 && gpio < 20) return false;     // 18-19 USB-JTAG
```
softhack007, line 246: This should only be disabled in case of ARDUINO_USB_CDC_ON_BOOT=0
Otherwise debugging over native USB will not work any more (purpose of ARDUINO_USB_CDC_ON_BOOT=1)

softhack007:
@zalatnaicsongor why did you close your PR? 🤔 


⚠️  Now we have an article in the KB explaining how to configure gpio19 on this controller, however gpio19 cannot be configured on c3, even when building the firmware with -DARDUINO_USB_CDC_ON_BOOT=0`

blazoncek:
@softhack007 I modified C3 behaviour to enable GPIO 18 & 19 if CDC is disabled.

softhack007:
@blazoncek thanks, your right it seems -  I've overlooked this recent change 😅 thanks for clarifying 👍 

https://github.com/Aircoookie/WLED/blob/ea608cecb0a5ac11626a82df4156f87e81ffc8c1/wled00/pin_manager.cpp#L243-L248

PR #4038, opened 2024-06-30
Restore ability to configure multi-pin bus from build flags

PaoloTK:
I have tweaked finalizeInit() to support multi-pin busses (clocked strips, analog 2ch+). My understanding is that this was working at one point in the past since some default environments (e.g. env:Athom_RGBCW) have flags for led types that require multiple pins.  

My implementation uses a function to calculate the number of required pins for the given bus type and then divides the number of pins passed at compile time by this amount to figure out the amount of busses to create. 

I have tested all possible combinations I could think of. Here's what happens based on the amount of data pins provided and the bus type:
1. If the bus type requires more pins than provided, no output gets configured.
2. If the bus type requires a fraction of the pins provided, outputs get configured until you run out of pins. So if you have 4 pins and a type of LED that requires 2 pins, output 1 will use the first and second pins, output 2 will use the third and fourth. If you have 5 pins configured, the 5th pin is not used since you'd need 6 to configure 3x of an LED type that requires 2 pins.
3. If the bus type requires the same amount of pins provided, one output gets configured with all the provided pins.

All LED types are supported with the exception of network ones. The pins are used instead to build the IP, but this is unwanted behavior, especially because currently they return TRUE on IS_2PIN (since their type is > 47).

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -801,6 +801,7 @@ class WS2812FX {  // 96 bytes
       getFirstSelectedSegId(void),
       getLastActiveSegmentId(void),
       getActiveSegsLightCapabilities(bool selectedOnly = false),
+      getNumPins(uint8_t busType),
       setPixelSegment(uint8_t n);
 
     inline uint8_t getBrightness(void) { return _brightness; }
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1109,11 +1109,16 @@ void WS2812FX::finalizeInit(void)
     DEBUG_PRINTLN(F("No busses, init default"));
     const uint8_t defDataPins[] = {DATA_PINS};
     const uint16_t defCounts[] = {PIXEL_COUNTS};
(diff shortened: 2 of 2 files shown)


blazoncek:
Please rebase for 0_15 branch first.

PaoloTK:
> Please rebase for 0_15 branch first.

I was actually hoping to get this bug fixed for the 0.14 branch since the reason I submitted this PR is to aid with a preassembled Quinled analog device that runs a C3 which currently works better on 0.14. However if 0.14 development has ceased completely I will rebase.

blazoncek:
There will not be any functionality changes in 0.14.
However 0_15 already has similar multipin fix in place.

PaoloTK:
> There will not be any functionality changes in 0.14. However 0_15 already has similar multipin fix in place.

I would consider this a bugfix more than a functionality change, but yeah I've taken a look at the 0.15 implementation and it makes sense to backport that instead regardless if needed. Thank you.

blazoncek:
See my post above. No more changes to 0.14 will be made.
Especially for compile time options as those have no influence on released binaries.

PR #4045, opened 2024-07-07
platform.ini cleanup + version bumps

TripleWhy:
## Goals

1. Make the platform.ini file easier to understand and read.
2. Being able to use more recent compiler and c++ versions (in all environments, instead of most environments). This will enable future improvements, that I am working on.

## Changes

- make formatting more uniform
- organize environments hierarchically
- remove duplications
- upgrade platform versions
- remove some obsolete dependencies
- upgrade some dependencies while at it
- upgrade c++ version from gnu++11 to gnu++17 (fastled doesn't compile without gnu extensions)
- remove debug symbols from the esp32 release builds

### Platform Version Upgrade
Most environments were using relatively recent platform versions, but the base esp32 configuration was not. Upgrading it does increase image size. But that is a problem we have anyway, because we do have envs that use more recent platform versions. This change just results in fewer differences between environments. It also increases the minimum gcc version the code has to compile with, therefore enabling better features and more recent c++ versions.

### Dependencies:
The original LittleFS dependency was removed, because all platforms are now recent enough to include it anyway.

The other removed dependency is `framework-arduinoespressif32 @ https://github.com/Aircoookie/arduino-esp32.git#1.0.6.4` that was used by the base esp32 configuration, but not all environments. I removed it because it is incompatible with the more recent platform version. 
I could however not figure out what it was good for originally, so hopefully this doesn't break anything? 
As far as I can tell, WLED runs fine without it on the ESP-WROOM-32 board. (Most other boards didn't use it.)

## Further Upgrade Options:
This PR brings the minimum GCC version from 5.2 to 8.4. It is possible to get it up to 12.2 by adding a line such as:
```
platform_packages = toolchain-xtensa-esp32@~12.2
```
That works, I tried it. It would however require specifying different toolchains for each architecture (I think).

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -19,107 +19,37 @@ extra_configs =
   platformio_override.ini
 
 [common]
-# ------------------------------------------------------------------------------
-# PLATFORM:
-#   !! DO NOT confuse platformio's ESP8266 development platform with Arduino core for ESP8266
-#
-#   arduino core 2.6.3 = platformIO 2.3.2
-#   arduino core 2.7.0 = platformIO 2.5.0
-# ------------------------------------------------------------------------------
-arduino_core_2_6_3 = espressif8266@2.3.3
-arduino_core_2_7_4 = espressif8266@2.6.2
-arduino_core_3_0_0 = espressif8266@3.0.0
-arduino_core_3_0_2 = espressif8266@3.2.0
-arduino_core_3_1_0 = espressif8266@4.1.0
-arduino_core_3_1_2 = espressif8266@4.2.1
-
-# Development platforms
-arduino_core_develop = https://github.com/platformio/platform-espressif8266#develop
-arduino_core_git = https://github.com/platformio/platform-espressif8266#feature/stage
-
-# Platform to use for ESP8266
-platform_wled_default = ${common.arduino_core_3_1_2}
-# We use 2.7.4.7 for all, includes PWM flicker fix and Wstring optimization
-#platform_packages = tasmota/framework-arduinoespressif8266 @ 3.20704.7
-platform_packages = platformio/toolchain-xtensa @ ~2.100300.220621 #2.40802.200502
-                    platformio/tool-esptool #@ ~1.413.0
-                    platformio/tool-esptoolpy #@ ~1.30000.0
-
-## previous platform for 8266, in case of problems with the new one
-## you'll need  makuna/NeoPixelBus@ 2.6.9 for arduino_core_3_0_2, which does not support Ucs890x
-;; platform_wled_default = ${common.arduino_core_3_0_2}
-;; platform_packages = tasmota/framework-arduinoespressif8266 @ 3.20704.7
-;;                    platformio/toolchain-xtensa @ ~2.40802.200502
-;;                    platformio/tool-esptool @ ~1.413.0
-;;                    platformio/tool-esptoolpy @ ~1.30000.0
+# ldscript (available ldscripts at https://github.com/esp8266/Arduino/tree/master/tools/sdk/ld)
+#    ldscript_2m1m (2048 KB) = 1019 KB sketch, 4 KB eeprom, 1004 KB spiffs, 16 KB reserved
+#    ldscript_4m1m (4096 KB) = 1019 KB sketch, 4 KB eeprom, 1002 KB spiffs, 16 KB reserved, 2048 KB empty/ota?
+ldscript_1m128k = eagle.flash.1m128.ld
+ldscript_2m512k = eagle.flash.2m512.ld
+ldscript_2m1m = eagle.flash.2m1m.ld
+ldscript_4m1m = eagle.flash.4m1m.ld
 
 # ------------------------------------------------------------------------------
 # FLAGS: DEBUG
 # esp8266 : see https://docs.platformio.org/en/latest/platforms/espressif8266.html#debug-level
 # esp32   : see https://docs.platformio.org/en/latest/platforms/espressif32.html#debug-level
 # ------------------------------------------------------------------------------
-debug_flags = -D DEBUG=1 -D WLED_DEBUG
-  -DDEBUG_ESP_WIFI -DDEBUG_ESP_HTTP_CLIENT -DDEBUG_ESP_HTTP_UPDATE -DDEBUG_ESP_HTTP_SERVER -DDEBUG_ESP_UPDATER -DDEBUG_ESP_OTA -DDEBUG_TLS_MEM ;; for esp8266
+debug_flags =
+  -g
+  -DDEBUG=1
+  -DWLED_DEBUG
+  -DDEBUG_ESP_WIFI
+  -DDEBUG_ESP_HTTP_CLIENT
+  -DDEBUG_ESP_HTTP_UPDATE
+  -DDEBUG_ESP_HTTP_SERVER
+  -DDEBUG_ESP_UPDATER
+  -DDEBUG_ESP_OTA
+  -DDEBUG_TLS_MEM
   # if needed (for memleaks etc) also add; -DDEBUG_ESP_OOM -include "umm_malloc/umm_malloc_cfg.h"
   # -DDEBUG_ESP_CORE is not working right now
 
 # ------------------------------------------------------------------------------
-# FLAGS: ldscript (available ldscripts at https://github.com/esp8266/Arduino/tree/master/tools/sdk/ld)
-#    ldscript_2m1m (2048 KB) = 1019 KB sketch, 4 KB eeprom, 1004 KB spiffs, 16 KB reserved
-#    ldscript_4m1m (4096 KB) = 1019 KB sketch, 4 KB eeprom, 1002 KB spiffs, 16 KB reserved, 2048 KB empty/ota?
-#
-# Available lwIP variants (macros):
-#    -DPIO_FRAMEWORK_ARDUINO_LWIP_HIGHER_BANDWIDTH  = v1.4 Higher Bandwidth (default)
-#    -DPIO_FRAMEWORK_ARDUINO_LWIP2_LOW_MEMORY       = v2 Lower Memory
-#    -DPIO_FRAMEWORK_ARDUINO_LWIP2_HIGHER_BANDWIDTH = v2 Higher Bandwidth
-#    -DPIO_FRAMEWORK_ARDUINO_LWIP2_HIGHER_BANDWIDTH_LOW_FLASH
-#
-# BearSSL performance:
-#  When building with -DSECURE_CLIENT=SECURE_CLIENT_BEARSSL, please add `board_build.f_cpu = 160000000` to the environment configuration
-#
-# BearSSL ciphers:
-#   When building on core >= 2.5, you can add the build flag -DBEARSSL_SSL_BASIC in order to build BearSSL with a limited set of ciphers:
-#     TLS_RSA_WITH_AES_128_CBC_SHA256 / AES128-SHA256
-#     TLS_RSA_WITH_AES_256_CBC_SHA256 / AES256-SHA256
-#     TLS_RSA_WITH_AES_128_CBC_SHA / AES128-SHA
-#     TLS_RSA_WITH_AES_256_CBC_SHA / AES256-SHA
-#  This reduces the OTA size with ~45KB, so it's especially useful on low memory boards (512k/1m).
-# ------------------------------------------------------------------------------
-build_flags =
-  -DMQTT_MAX_PACKET_SIZE=1024
-  -DSECURE_CLIENT=SECURE_CLIENT_BEARSSL
-  -DBEARSSL_SSL_BASIC
-  -D CORE_DEBUG_LEVEL=0
-  -D NDEBUG
-  -Wno-attributes ;; silence warnings about unknown attribute 'maybe_unused' in NeoPixelBus
-  #build_flags for the IRremoteESP8266 library (enabled decoders have to appear here)
-  -D _IR_ENABLE_DEFAULT_=false
-  -D DECODE_HASH=true
-  -D DECODE_NEC=true
-  -D DECODE_SONY=true
-  -D DECODE_SAMSUNG=true
-  -D DECODE_LG=true
-  -DWLED_USE_MY_CONFIG
-
-build_unflags =
-
-ldscript_1m128k = eagle.flash.1m128.ld
-ldscript_2m512k = eagle.flash.2m512.ld
-ldscript_2m1m = eagle.flash.2m1m.ld
-ldscript_4m1m = eagle.flash.4m1m.ld
-
-[scripts_defaults]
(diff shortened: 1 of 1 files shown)


softhack007:
Hi @TripleWhy,

About the esp32 platform version - sorry this is a no-go from my point of view. Same for changing gcc version to diverge from the versions included in respective arduino-esp32 release.

 For esp32 4MB, we first need a solution to reduce firmware size before upgrading the framework.

platformio.ini
```diff
@@ -195,325 +173,295 @@ build_flags =
...
   ESPAsyncTCP @ 1.2.2
   ESPAsyncUDP
-  ${env.lib_deps}
 
 [esp32]
-#platform = https://github.com/tasmota/platform-espressif32/releases/download/v2.0.2.3/platform-espressif32-2.0.2.3.zip
-platform = espressif32@3.5.0
-platform_packages = framework-arduinoespressif32 @ https://github.com/Aircoookie/arduino-esp32.git#1.0.6.4
-build_flags = -g
+platform = espressif32@6.7.0
```
softhack007, line 185: Sorry but the ESP-IDF version included in the "standard" platform 6.7.0 has known bugs, we cannot use it.
The latest usable arduino-esp32 versions are 2.0.9 (well tested) and 2.0.14 (seems to work). Everything that came after 2.0.14 is hot garbage.

platformio.ini
```diff
@@ -19,107 +19,37 @@ extra_configs =
...
-
-# Development platforms
-arduino_core_develop = https://github.com/platformio/platform-espressif8266#develop
-arduino_core_git = https://github.com/platformio/platform-espressif8266#feature/stage
-
-# Platform to use for ESP8266
-platform_wled_default = ${common.arduino_core_3_1_2}
-# We use 2.7.4.7 for all, includes PWM flicker fix and Wstring optimization
-#platform_packages = tasmota/framework-arduinoespressif8266 @ 3.20704.7
-platform_packages = platformio/toolchain-xtensa @ ~2.100300.220621 #2.40802.200502
```
softhack007, line 44: @blazoncek @willmmiles I seems to remember that this was added on purpose, especially to prevent very strange problems observed by users making their own custom builds? I think we need the `platform_packages`  section for 8266?

platformio.ini
```diff
@@ -195,325 +173,295 @@ build_flags =
...
   ;; ARDUINO_USB_MODE, ARDUINO_USB_CDC_ON_BOOT
-lib_deps =
-  https://github.com/pbolduc/AsyncTCP.git @ 1.2.0
-  ${env.lib_deps}
+
+[esp32c3]
+;; generic definitions for all ESP32-C3 boards
+extends = esp32
+build_flags =
+  ${esp32.build_flags}
```
softhack007, line 252: Sorry, just no.

softhack007:
@TripleWhy maybe you dis not recognize this - the platformio.ini "high level" structure is correct, and it was chosen carefully. We need a base environment for all supported chip models - 8266, esp32, esp32-s3, esp32-s2, esp32-c3. Two years ago we tried to base all esp32 "variants" on the same build_flags and lib_deps, and it was a huge disaster. This is why we have largely independent top level environments for  [esp32], [esp32-S3], etc. Please do not touch this part.

You may of course propose optimizations for the lower level [env:....] environments. Especially the use of "extends" is very welcome. Also minor readability improvements are welcome.

If you change the main environments (without "env" prefix), please also perform functional testing on the environments derives from them. "It compiles" is not good enough as a justification 😉 when changing the main sections.

platformio.ini
willmmiles: We'd removed an unversioned `platform_packages = platformio/framework-arduinoespressif8266` call because it was overriding the `platform = ${common.platform_wled_default}` selection.  The unversioned reference was picking up old random versions from people's build systems instead of the specific framework version intended by the `platform` entry.

I can't speak for specific the specific toolchain version selections - it's "always been that way" for the few months I've been working with WLED.

blazoncek:
Any change to `platformio.ini` for platforms or packages or toolchains or boards or base environments will require thorough testing on several different boards of all supported MCU variants.
I cannot approve this for 0.15 in any form but it may be useful for 0.16 when we start working on it.

TripleWhy:
Thanks for the insight everyone. I should not have mixed refactoring and version changes, so that there can be independent discussions about those topics.

Therefore I propose the following:  
If I can produce a new platform.ini without version changes that I can prove produces the same result as the current one, would you accept that?

I would based on that, I would then later open a new discussion about versions. (I am deliberately not replying to version related comments here.)

softhack007:
> Therefore I propose the following:
> If I can produce a new platform.ini without version changes that I can prove produces the same result as the current one, would you accept that?

Hi @TripleWhy thanks for your proposal, this makes a lot sense.

Before we go again into building "head with nails" (instead of nails with heads), maybe let's see what are our needs.

From my perspective - as the evil genius who drove new MCU support for some time - I think we need this as a generic structure.

* [common] -> flags and libs for WLED in general
* |- [esp8266] -> everything specific to 8266
* |- [esp32_base] (or any better name) flags, libs, partitions, that are the same for all esp32 variants. Maybe include platform. Also collect flags like `AR_lib_deps`
* |--> next 6 targets inherit from [esp32_base]
*   |--- [esp32] -> classic esp32, arduino-esp32 v1.x (legacy)
*   |--- [esp32_V4] -> classic esp32, but using newer arduino-esp32 v2.x
*   |--- [esp32s3]
*   |--- [esp32s2]
*   |--- [esp32c3]
*   |--- [esp32c6] (maybe supported in WLED 0.16.)

The last 6 targets can be the baseline for user's specific buildenvs.

The reason is that some MCUs will always need special flags. If you follow a strict top-down design, this will easily lead into conflicting flags. So I prefer to have esp32, s3, s2, c3 (and later c6 maybe) all side-by-side on the second or third level.

_Definition of done_
- standard firmware is the same when build withhold or new platformIO.ini (binary compare)
- randomly picked custom buildenvs for the previous platformIO.ini still compile without errors using the new platdormio.ini
- maybe also check if 3rd party build tools - like the tool from @wled-install - are still working correctly.

Comments welcome 😃

platformio.ini
```diff
@@ -195,325 +173,295 @@ build_flags =
...
-  ; -D PIO_FRAMEWORK_ARDUINO_MMU_CACHE16_IRAM48_SECHEAP_SHARED ;; (experimental) adds some extra heap, but may cause slowdown
+  -DPIO_FRAMEWORK_ARDUINO_MMU_CACHE16_IRAM48 ;; in case of linker errors like "section `.text1' will not fit in region `iram1_0_seg'"
+  ; -DPIO_FRAMEWORK_ARDUINO_MMU_CACHE16_IRAM48_SECHEAP_SHARED ;; (experimental) adds some extra heap, but may cause slowdown
 
 lib_deps =
-  #https://github.com/lorol/LITTLEFS.git
+  ${env.lib_deps}
   ESPAsyncTCP @ 1.2.2
   ESPAsyncUDP
-  ${env.lib_deps}
```
softhack007, line 205: This order - specific libs first, `${env.lib_deps}` last - was chosen on purpose.
The reason is that asyncWebserver imports an old ESPAsyncTCP release. so the only way to ensure that only the version we want is used, was to put "our" ESPAsyncTCP before asyncWebserver which is part of `${env.lib_deps}`.

Pease keep the previous order.

platformio.ini
```diff
@@ -165,18 +95,66 @@ lib_deps =
...
+# BearSSL ciphers:
+#   When building on core >= 2.5, you can add the build flag -DBEARSSL_SSL_BASIC in order to build BearSSL with a limited set of ciphers:
+#     TLS_RSA_WITH_AES_128_CBC_SHA256 / AES128-SHA256
+#     TLS_RSA_WITH_AES_256_CBC_SHA256 / AES256-SHA256
+#     TLS_RSA_WITH_AES_128_CBC_SHA / AES128-SHA
+#     TLS_RSA_WITH_AES_256_CBC_SHA / AES256-SHA
+#  This reduces the OTA size with ~45KB, so it's especially useful on low memory boards (512k/1m).
+# ------------------------------------------------------------------------------
+build_flags =
+  -std=gnu++17
```
softhack007, line 133: Why force gnu++17 ? 
I'd prefer we stay with what espressif has set in their platform files.

platformio.ini
```diff
@@ -165,18 +95,66 @@ lib_deps =
...
+  -DDECODE_HASH=true
+  -DDECODE_NEC=true
+  -DDECODE_SONY=true
+  -DDECODE_SAMSUNG=true
+  -DDECODE_LG=true
+  -DWLED_USE_MY_CONFIG
+  ; ${common.debug_flags}
+
+build_unflags =
+  -std=gnu++11
```
softhack007, line 151: No.

platformio.ini
```diff
@@ -195,325 +173,295 @@ build_flags =
...
-  ${env.lib_deps}
-
-[esp32c3]
-;; generic definitions for all ESP32-C3 boards
-platform = espressif32@ ~6.3.2
-platform_packages = platformio/framework-arduinoespressif32 @ 3.20009.0    ;; select arduino-esp32 v2.0.9 (arduino-esp32 2.0.10 thru 2.0.14 are buggy so avoid them)
-build_flags = -g
-  -DARDUINO_ARCH_ESP32
-  -DARDUINO_ARCH_ESP32C3
-  -DCONFIG_IDF_TARGET_ESP32C3=1
```
softhack007, line 274: Where did you put this line? (`-DCONFIG_IDF_TARGET_ESP32C3=1`)

Its critical for ensuring a proper build on -S3.

softhack007 (changes requested):
Thanks for your effort to make our platformIO.ini easier to maintain.

As already explained, I would prefer that this PR gets a "fresh start" with all maintainers first stating their needs.

softhack007 (changes requested):
Thanks for your effort to make our platformIO.ini easier to maintain.

As already explained, I would prefer that this PR gets a "fresh start" with all maintainers first stating their needs.

platformio.ini
```diff
@@ -195,325 +173,295 @@ build_flags =
...
+  ${env.lib_deps}
   ESPAsyncTCP @ 1.2.2
   ESPAsyncUDP
-  ${env.lib_deps}
 
 [esp32]
-#platform = https://github.com/tasmota/platform-espressif32/releases/download/v2.0.2.3/platform-espressif32-2.0.2.3.zip
-platform = espressif32@3.5.0
-platform_packages = framework-arduinoespressif32 @ https://github.com/Aircoookie/arduino-esp32.git#1.0.6.4
-build_flags = -g
```
softhack007, line 211: What is the problem with '-g'? firmware.bin seems to be 100% the same with or without, however firmware.elf (needed for stackdumps) would be missing parts that make crash dumps readable.

TripleWhy: yeah I noticed too that there is another -ggdb flag coming from somewhere. My intention here was to make the release build smaller by removing debug symbols  and putting them in the debug_flags so that you can manually add them when debugging and not when building a release for publishing.
To achieve that, we would also have to unflag -ggdb.
However, I don't actually know if the debug information is actually in the final binary or if it is stripped by a later build step.

platformio.ini
TripleWhy: It's in the `[esp32c3]` section obviously :)

platformio.ini
TripleWhy: I consider the c++ version a version discussion, too, so let's discuss this later.

TripleWhy:
> The reason is that some MCUs will always need special flags. If you follow a strict top-down design, this will easily lead into conflicting flags. So I prefer to have esp32, s3, s2, c3 (and later c6 maybe) all side-by-side on the second or third level.

I see the point, but:

1. My impression was that you are supposed to create an override file instead of modifying the platform.ini to customize your build. So in the end is only matters what is effectively inside the `env:` sections. For that it doesn't matter if the content came from an `extend` or from repetition. While that doesn't matter for users who want to create their own builds, fewer repetitions (almost) always makes the maintainer's job easier.
2. Even with a hierarchical structure, you can always overwrite the settings anyway either by repeating, or for compiler flags, you can simply append a contradicting flag, as only the latest one is considered.

That being said, the hierarchy you suggest might actually make sense from a logical point of view, if you consider e.g. ESP32 and ESP32-C3 different chips. (I guess they are.) Turns out the issue here is simply the naming. I therefore suggest renaming the `env:esp32` (and `env:esp32_v4`) to something like `env:esp32_classic(_v4)`. Or maybe to find a more technical term wroom or lx6 instead of classic (would that be correct?).

platformio.ini
willmmiles: Debug information is stripped from the binaries uploaded to the devices in the step that creates .bin from .elf.  The binaries would be way way way too big otherwise.

Jason2866:
Just sharing my experience with Platformio setup for a big project. Do not change many stuff in one step. Platformio is highly complex under the hood and has its bugs.
The extensive use of extends is in my experience not what I would do. Been there and had "fun" and simplified the very elegant looking setup. Using scripts has its bugs (in Platformio) too. I avoid to change a working setup.
In general "overriding" previous settings can have very strange unwanted side effects.

If strange compile errors occur (header files not found and such stuff) after a Platformio setup change. There is a problem "somewhere" in Platformio. The issue is completely unrelated!!

platformio.ini
Jason2866: espressif32@3.5.0 is Arduino core 1.0.6
A change to espressif32@6.7.0 is a change to core 2.0.16
Major version change!! Btw. 2.0.16 has bugs. Only latest version in branch release/v2.x in repo should be used (not released).
Core 2.0.14 is working not bad at all, if a release version is needed

softhack007:
Thanks @Jason2866 what you say makes a lot of sense to me.

> The extensive use of extends is in my experience not what I would do. Been there and had "fun" and simplified the very elegant looking setup.

While I like how "extends" improves readability, its also true it makes understanding what's really going on a nightmare sometimes. 

From my own experience: I had a platformio.ini with lots of inheritance by _extends=_. The problem was, in one build env there was a flags that should not be there, while two other flags - that were visible in build_flags - were missing. I was like _WTF_ and it took me two days to understand what's going on under the hood. In the end, I used `advanced -> compilation database` just to see the flags that were actually passed to gcc. It turned out that one inheritance path was wrong (explaining the extra build flag), and a deliberate "build_unflags=" from another inheritance path removed the flags that should be there. :shipit:

blazoncek:
`extends` are bane and boon at the same time.

I used to avoid them but it was PITA to maintain `platformio_override.ini` with 20+ environments.

I am now using them extensively but have carefully crafted hierarchy in my overrides. So my take would be to avoid `extends` in base `platformio.ini` but give commented examples in sample override file.

Disclaimer: I am not a developer and never delved deep into how PIO works so all my knowledge comes from working or failed experiments.

Issue #3396, opened 2023-09-28
Implement QuickDebug library for debugging


blazoncek:
https://github.com/gmag11/QuickDebug

Will require updates to every DEBUG_...() call to set appropriate level of debugging.

softhack007:
👍  +1

blazoncek:
@ewoudwijma @netmindz this may benefit MoonModules for as well if you are willing to help.

netmindz:
Looks a good library to use

softhack007:
closing - we will replace QuickESPNow, and there is a more generic concept in discussion for debug messages.

PR #4050, opened 2024-07-10
Debug refactor

blazoncek:
Separate different debug components]

- generic (WLED_DEBUG)
- FX (WLED_DEBUG_FX)
- bus manager (WLED_DEBUG_BUS)
- pin manager (WLED_DEBUG_PINMANAGER)
- usermods (WLED_DEBUG_USERMODS)
- math (WLED_DEBUG_MATH)
- all (WLED_DEBUG_ALL)

In an attempt to reduce flash usage while still getting a few debug messages.

It would still be preferable to implement #3396 instead or alongside.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -58,7 +58,7 @@ platform_packages = platformio/toolchain-xtensa @ ~2.100300.220621 #2.40802.2005
 # esp8266 : see https://docs.platformio.org/en/latest/platforms/espressif8266.html#debug-level
 # esp32   : see https://docs.platformio.org/en/latest/platforms/espressif32.html#debug-level
 # ------------------------------------------------------------------------------
-debug_flags = -D DEBUG=1 -D WLED_DEBUG
+debug_flags = -D DEBUG=1 -D WLED_DEBUG_ALL ;; or -D WLED_DEBUG -DWLED_DEBUG_FS -DWLED_DEBUG_FX -DWLED_DEBUG_BUS -DWLED_DEBUG_PINMANAGER -DWLED_DEBUG_USERMODS -DWLED_DEBUG_MATH
   -DDEBUG_ESP_WIFI -DDEBUG_ESP_HTTP_CLIENT -DDEBUG_ESP_HTTP_UPDATE -DDEBUG_ESP_HTTP_SERVER -DDEBUG_ESP_UPDATER -DDEBUG_ESP_OTA -DDEBUG_TLS_MEM ;; for esp8266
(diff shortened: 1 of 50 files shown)


netmindz:
From a brief look - QuickDebug has more features

With this MR I think it would be all to eat to use the wrong value going forward with new development, then lost time with people not seeing the logs they expect etc. It's one thing to do a find and replace, it's quite another to ensure correct usage going forward

netmindz:
Out of interest, how much smaller is that the image without the logs?

blazoncek:
> From a brief look - QuickDebug has more features

True, but it also uses more RAM and flash and is not a drop-in replacement. I would still prefer to use it, though, if we can manage to change entire codebase.

> It's one thing to do a find and replace, it's quite another to ensure correct usage going forward

The debug options are divided into separate groups which makes debug output even less crowded. I.e. if you want to debug segment and effect handling then you'd use `WLED_DEBUG_FX` and your output would not be overwhelmed by other unnecessary debug information. Or, if you are working on enhancing bus manager you would only want to see `WLED_DEBUG_BUS` output.
There are no restrictions to keep using `WLED_DEBUG` (or add `WLED_DEBUG_ALL`) for anything an uneducated user wants.

> Out of interest, how much smaller is that the image without the logs?

Instead of receiving 100.1% to 105% flash usage on my development builds I can use debug I need and still be around 98.9% of flash usage. Non debug builds yield about 97%.

netmindz:
> > From a brief look - QuickDebug has more features
> 
> True, but it also uses more RAM and flash and is not a drop-in replacement. I would still prefer to use it, though, if we can manage to change entire codebase.

Hopefully not so much that it's a blocker

> > It's one thing to do a find and replace, it's quite another to ensure correct usage going forward
> 
> The debug options are divided into separate groups which makes debug output even less crowded.

Yes I understand what you are trying to achieve, it's just the code discipline and chance for user error when coding to use the wrong DEBUGXX_PRINTLN by mistake. That's where QuickDebug has a big advantage 

> > Out of interest, how much smaller is that the image without the logs?
> 
> Instead of receiving 100.1% to 105% flash usage on my development builds I can use debug I need and still be around 98.9% of flash usage. Non debug builds yield about 97%.

A fair chunk of space then. I think QuickDebug will produce smaller builds based on the log level, I'm not sure if changing the log level for specific tags would also reduce the filesize

blazoncek:
Just for the reference, I will not be updating this branch any more. I have moved all of the codebase into my fork and merged it into my development branch as I could not get debug builds to fit into flash by enabling all features/build options that I depend on.

If anyone ever revisits this or implements QuickDebug I will be monitoring still.

PR #4058, opened 2024-07-15
Fetch supported LED config from Bus

netmindz:
Following on from this PR https://github.com/Aircoookie/WLED/pull/4056

Replace convoluted logic in settings_leds.htm with config requirements supplied in the LEDType data

Motivation - https://github.com/Aircoookie/WLED/pull/3777

### DIFF
--- a/wled00/bus_manager.h
+++ b/wled00/bus_manager.h
@@ -108,7 +108,7 @@ struct LEDType {
   uint8_t id;
   String type;
   String name;
-  String config = F("{}");
+  String config = "{}";
 };
 
 //parent class of BusDigital, BusPwm, and BusNetwork


blazoncek:
@netmindz please take a look at https://github.com/Aircoookie/WLED/commit/6f3267aee98095bab3381956fa1949f5c09cb2db
Implements similar as this PR but does not change Bus class. Instead it utilises BusManager and uses LED type convention to select appropriate GPIO text.

blazoncek:
Except for GPIO field names I think this PR has become obsolete.
Will you expand on GPIO or move that into HUB75 PR? I would prefer latter.

netmindz:
I'll still like to remove things like IP address from the html, but certainly anything in here needs to be redone against the final bus-config

netmindz:
Idea added in other commits

PR #4060, opened 2024-07-17
boot-up delay to fix wifi not starting in some setups

DedeHai:
adding a new #define option to allow users to add a delay before hardware init
use
-D WLED_BOOTUPDELAY=500
in platformio env definition to add 500ms of delay before hardware init

the idea is to let power stabilise for a little longer before hardware initialisation, this seems to help with some setups that have issues with wifi not connecting at bootup. On two of my setups that have issues with wifi at boot, this helps.


softhack007:
@DedeHai thanks,

It looks like you picked the wrong branch -  could you redo your PR and make it against our `0_15` branch, instead of targeting `main`?

softhack007:
From my side - OK to merge 👍

PR #4061, opened 2024-07-18
Fix Lissajous effect

pipe01:
The Lissajous effect's metadata string has an extra semicolon, which is wrong according to [the documentation](https://kno.wled.ge/interfaces/json-api/#effect-metadata).


blazoncek:
Please rebase for 0_15 branch.

pipe01:
Seems like it's already fixed there, my bad.

PR #4071, opened 2024-07-23
Check the pinManager for allocated usermod pins.

axlan:
Proposed fix for https://github.com/Aircoookie/WLED/issues/4070 .

Tested on a esp32s3.


blazoncek:
This is unnecessary if usermods correctly store and use configuration in "cfg.json".

PR #4073, opened 2024-07-25
Add example runArgs to devcountainer to access serial port.

axlan:
I found this capability to be useful for being able to upload directly from the dev container.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -12,6 +12,13 @@
 		}
 	},
 
+	// To give the container access to a device serial port, you can uncomment one of the following lines.
+	// You can explicitly forward the port. The docker user needs to be able to access this port, and this will only work
+	// if the device is plugged in from the start without reconnecting.
+	// "runArgs": ["--device=/dev/ttyACM0", "--group-add", "dialout"],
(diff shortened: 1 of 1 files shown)


blazoncek:
You should base any work off of 0_15 branch.
I have changed destination branch for this PR but you should be careful next time.

softhack007:
@axlan please add a sentence about security implications of the proposed commands. If I understand correctly, it means that the container runtime get permission to access any "dialout" device.

> *dialout* Full and direct access to serial ports. Members of this group can reconfigure the modem, dial anywhere, etc.

https://wiki.debian.org/SystemGroups#Other_System_Groups

axlan:
@softhack007 , that's not entirely correct.
For ` "runArgs": ["--device=/dev/ttyACM0", "--group-add", "dialout"],` the container is still sandboxed, so only the serial device `/dev/ttyACM0` is exposed. The group is added since typically access to a serial port is restricted to this group. The other ports aren't even visible.

Running with `--privileged` does actually expose the full set of serial devices, so I can call that out more explicitly, but it's also important to remember that even in that case, the exposure is less then just running outside of a dev container normally. These sandboxing concerns are usually more relevant for running systems that are exposing servers to the network, rather then just providing an environment for building.

Issue #3953, opened 2024-05-03
BTNPIN compile time option sets unused buttons to GPIO 0


robertvb83:
### What happened?

Setting 4 buttons works as expected (buttons GPIO are only set but still disabled) -D BTNPIN=0,4,34,35

However if you set less than 4 buttons, the remaining buttons are set to GPIO 0 automatically. Leading to double entries for 0
-D BTNPIN=0,4

### To Reproduce Bug

See above

### Expected Behavior

Set other buttons to "unused"

### Install Method

Self-Compiled

### What version of WLED?

WLED 0.15-b3

### Which microcontroller/board are you seeing the problem on?

ESP32

### Anything else?

![Screenshot_20240503-141637](https://github.com/Aircoookie/WLED/assets/36081149/55eaebc4-c860-4d9b-b762-4398d4e69532)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

blazoncek:
Just define BTNPIN=0,4,-1,-1
You can also set BTNTYPE if you wish.

robertvb83:
thank you, that is, what i was looking for.

how is BTNTYPE set? i use 2 push buttons and 1 analog potentiometer, 4th button should be GPIO35 but disabled by default

BTNPIN=0,4,34,35
BTNTYPE = PUSH, PUSH, ANALOG, -1

blazoncek:
Check `const.h` for button types.

robertvb83:
> Check `const.h` for button types.

Thanks for pointing me in that direction. It works exactly as it should

PaoloTK:
> > I'm not really happy with the fix proposed in #4074, because it will break again when user set `-D WLED_MAX_BUTTONS=x` in their custom build environments.
> 
> I second that.

I am a bit confused by this. I initially was exploring counting the passed pins to set the rest up to WLED_MAX_BUTTONS to -1. However once I saw that this very issue was marked as "won't fix" with the workaround of setting unused pins to -1, I simply replicated that in code. 

I agree that it's not the best solution because a custom compile can break it by setting a different WLED_MAX_BUTTONS value, but many things can be broken by improper flag configuration.

blazoncek:
It just means you need to think harder how to solve the underlying issue. Not just provide a partial "band aid".

PaoloTK:
> It just means you need to think harder how to solve the underlying issue. Not just provide a partial "band aid".

[...]

PR #4074, opened 2024-07-25
Fix default button pin assignment for esp32

PaoloTK:
When using the BTNPIN compile flag to assign button pins in 0.15 you must define your pins and then add leading "-1" until WLED_MAX_BUTTONS, which is expected behavior (https://github.com/Aircoookie/WLED/issues/3953). However if you don't define BTNPIN at compile a similar issue occurs on ESP32 due to the default def only having 2 pins. I added an if statement to define 4 default pins in case we're compiling for ESP32. 

I also changed the BTNTYPE def. It currently has a def for 2 pins with the second being BTN_TYPE_NONE. However since BTN_TYPE_NONE = 0 any type not configured will automatically evaluate to BTN_TYPE_NONE.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -12,6 +12,13 @@
 		}
 	},
 
+	// To give the container access to a device serial port, you can uncomment one of the following lines.
+	// You can explicitly forward the port. The docker user needs to be able to access this port, and this will only work
+	// if the device is plugged in from the start without reconnecting.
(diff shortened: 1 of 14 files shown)


softhack007:
Hi,

I think this solution is not good for the generic case, as it simply "postpones" the problem because users may want WLED_MAX_BUTTONS >4 (12, or maybe even 16). See [my comment](https://github.com/Aircoookie/WLED/issues/3953#issuecomment-2250333363) in #3953.

PR #4078, opened 2024-07-27
Add a parameter for blur amount in the Fire 2012 effect

apanteleev:
With a larger 2D matrix, blur becomes very expensive in terms of performance. On ESP32 with a 32x24 matrix, the Fire 2012 effect with blur runs at 13 FPS, and without blur at about 30 FPS. Sometimes more, but I couldn't figure out why.

This PR adds a user setting for the blur amount so that you can set it to 0 or any other value for aesthetic reasons.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -2106,14 +2106,17 @@ uint16_t mode_fire_2012() {
   for (unsigned stripNr=0; stripNr<strips; stripNr++)
     virtualStrip::runStrip(stripNr, &heat[stripNr * SEGLEN], it);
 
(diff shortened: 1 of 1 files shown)


blazoncek:
Please rebase for 0_15 branch.

apanteleev:
Rebased.

blazoncek:
I will merge this but just out of curiosity, did you check speed with 0_15 branch? There were some optimisations done regarding blur.

apanteleev:
> did you check speed with 0_15 branch? There were some optimisations done regarding blur.

Not when submitting the PR, but now I did. 29 FPS with blur, 43 FPS without. So, much faster than 0.14 - thanks for letting me know!

PR #4081, opened 2024-08-01
Allow lower values for touch threshold

RobinMeis:
Solved issue https://github.com/Aircoookie/WLED/issues/4076 by allowing lower values for touch threshold

### DIFF
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -296,7 +296,7 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
           #ifdef SOC_TOUCH_VERSION_2    // ESP32 S2 and S3 have a function to check touch state but need to attach an interrupt to do so
           else
           {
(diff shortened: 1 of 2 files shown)


blazoncek:
Once @dedehai gets back from vacation I hope he'll review and approve.

DedeHai (commented):
Need to test, may lead to ghost detections

RobinMeis:
> Need to test, may lead to ghost detections

For me it doesn't but shouldn't be a too big issue as you already can get ghost detections on ESP32-Wroom with inappropriate settings.

DedeHai:
I tested the change, low values lead to ghost detections but that is probably ok, I did not notice any crashes in my tests (tested on S2). 
@RobinMeis your PR is missing changes to the second interrupt attachment here:
https://github.com/Aircoookie/WLED/blob/f6c47ac19cd9fc2b6b2861289b86f2ef963a329a/wled00/set.cpp#L292

apart from that I am ok with removing the 256 as a minimum base value.

RobinMeis:
I have fixed this. For me, I also get ghost detections for very low values but didn't get any ghost detections within the last two weeks on two lamps with reasonable values.

DedeHai:
looks good, gets my approval.

PR #4082, opened 2024-08-01
Fix incorrect PWM bit depth on Esp32 with XTAL clock

PaoloTK:
Fixes incorrect bit depth for ESP32 with XTAL support (new "wave" of ESP32 models like C3).

Currently 0.15 has hardcoded bit depths for ESP8266 and ESP32. Since Arduino 2.0.2, ESP32 models that support external XTAL clock  (C3 for example) are [set to use it instead of the built in APB clock](https://github.com/espressif/arduino-esp32/blob/2.0.2/cores/esp32/esp32-hal-ledc.c#L25). XTAL runs at half the frequency (40 MHz rather than 80 MHz) and so the pin fails to assign correctly due to incorrect bit depth.
My approach was to calculate the correct bit depth on the fly based on what clock the microprocessor supports (including ESP8266). I have tested the code on a regular ESP32, a C3 and an ESP8266 on all currently possible clock settings and everything looks in line with the previous values.

A few notes:
1. I tested performance and it's less than 0.5 ms extra in the worst case I've seen.
2. There's no frequency validation. This can technically cause issues (for example the new ESP32 boards have a smaller range of bit depths compared to the original) but since currently the frequencies are hardcoded and not user configurable I don't think it's a problem.
3. Even before this PR, ESP8266 was configurable with frequencies above 1000 Hz. In my testing (and it's commonly agreed) 8266 cannot reliably produce frequencies above 1000 Hz, causing flickering and poor dimming performance. I think it's worth considering removing these options.

### DIFF
--- a/wled00/bus_manager.cpp
+++ b/wled00/bus_manager.cpp
@@ -378,6 +378,31 @@ void BusDigital::cleanup() {
 }
 
 
+#ifdef ESP8266
+  // 1 MHz clock
+  #define CLOCK_FREQUENCY 1000000UL
+#else
+  // Use XTAL clock if possible to avoid timer frequency error when setting APB clock < 80 Mhz
+  // https://github.com/espressif/arduino-esp32/blob/2.0.2/cores/esp32/esp32-hal-ledc.c
+  #ifdef SOC_LEDC_SUPPORT_XTAL_CLOCK
+    #define CLOCK_FREQUENCY 40000000UL
+  #else
+    #define CLOCK_FREQUENCY 80000000UL
+  #endif
+#endif
+
+#ifdef ESP8266
+  #define MAX_BIT_WIDTH 10
+#else
+  #ifdef SOC_LEDC_TIMER_BIT_WIDE_NUM
+    // C6/H2/P4: 20 bit, S2/S3/C2/C3: 14 bit
+    #define MAX_BIT_WIDTH SOC_LEDC_TIMER_BIT_WIDE_NUM 
+  #else
+    // ESP32: 20 bit (but in reality we would never go beyond 16 bit as the frequency would be to low)
+    #define MAX_BIT_WIDTH 20
+  #endif
+#endif
+
 BusPwm::BusPwm(BusConfig &bc)
 : Bus(bc.type, bc.start, bc.autoWhite, 1, bc.reversed)
 {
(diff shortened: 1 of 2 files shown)


wled00/bus_manager.cpp
```diff
@@ -384,24 +397,17 @@ BusPwm::BusPwm(BusConfig &bc)
   if (!IS_PWM(bc.type)) return;
   unsigned numPins = NUM_PWM_PINS(bc.type);
   _frequency = bc.frequency ? bc.frequency : WLED_PWM_FREQ;
+  // duty cycle resolution (_depth) can be extracted from this formula: CLOCK_FREQUENCY > _frequency * 2^_depth
+  _depth = uint8_t(log((float)CLOCK_FREQUENCY / (float)_frequency) / log(2.0));
```
blazoncek, line 401: Why floating point and `log()`? You can get away using integer math and shifting. 
I.e. `_depth = (CLOCK_FREQUENCY/_frequency) >> 8;`
And then devise actual bit depth by counting active bits.

Or use `for` loop to get required bits by shifting dynamically.
I.e. `for (_depth=12; _depth>8; _depth--) if (((CLOCK_FREQUENCY/_frequency)>>_depth) > 0) break;`

blazoncek (changes requested):
I do not like floating point and the dreaded `log()`.
You can get away without.

PaoloTK:
> I do not like floating point and the dreaded `log()`. You can get away without.

Thank you, fixed. I also implemented a platform specific max bit width value but kept the lower limit at 8 bit. 

I've been wondering about the best strategy in regard to esp8266. The issue stems from the fact that WLED_PWM_FREQ is set to a middle ground value on ESP32 that strikes a good balance between frequency and bit depth, but on 8266 it is set to the highest possible frequency before dimming stutters and similar issues occur.
An option could be to set WLED_PWM_FREQ to a value such that the fastest preset is = 880 Hz, but that doesn't seem great because in the future presets might change (or go away).
We could disable the options above Normal for ESP8266, which would limit the highest frequency to WLED_PWM_FREQ.

wled00/const.h
```diff
@@ -521,6 +521,35 @@
...
+
+#ifndef MAX_BIT_WIDTH
+  #ifdef ESP8266
+    #define MAX_BIT_WIDTH 10
+  #else
+    #ifdef SOC_LEDC_TIMER_BIT_WIDE_NUM
+      // C6/H2/P4: 20 bit, S2/S3/C2/C3: 14 bit
+      #define MAX_BIT_WIDTH SOC_LEDC_TIMER_BIT_WIDE_NUM 
+    #else
+      // ESP32: 32 bit
```
blazoncek: https://www.espressif.com/sites/default/files/documentation/esp32_technical_reference_manual_en.pdf#ledpwm

wled00/bus_manager.cpp
```diff
@@ -40,19 +40,6 @@ uint8_t realtimeBroadcast(uint8_t type, IPAddress client, uint16_t length, byte
   #define DEBUG_PRINTF_P(x...)
 #endif
 
-// ESP8266 has 1 MHz clock
```
blazoncek: I would keep these constant here.

blazoncek (approved):
Haha! I came to almost exactly the same code, except I didn't move constants into `const.h` and didn't use floats.
Since the constants are not needed anywhere else perhaps it makes no sense to put them into `const.h`.

PaoloTK:
> Haha! I came to almost exactly the same code, except I didn't move constants into `const.h` and didn't use floats. Since the constants are not needed anywhere else perhaps it makes no sense to put them into `const.h`.

No surprise there, I tested the solutions you suggested and found the second to work better (first had issues with rounding).

Thank you for all the unvaluable feedback and for merging with the fixes you suggested!

PR #4086, opened 2024-08-03
Add Test mode usermod to load pattern on multi-button hold

PaoloTK:
This usermod allows you to load a specific test pattern (brightness, color, effect, palette) by holding down a designated number of buttons.

It was developed for quality control on pre-assembled QuinLED boards with PWM outputs to control "dumb" LEDs.
To test a digital output you can simply connect it to an addressable LED, which will either work properly or not.
This doesn't work with analog outputs, because the MOSFET can fail "closed" and thus still light up the LED. The whole dimming range must be tested.

The test pattern is triggered by holding down multiple buttons, and then automatically disables itself so that it can be used on first boot on the assembly line and then be invisible to the user. If for some reason the user wants to test the outputs with the pattern, they can manually enable the usermod again.

### DIFF
--- a/usermods/Test_Mode/README.md
+++ b/usermods/Test_Mode/README.md
@@ -1,0 +1,26 @@
+# Usermod Test Mode
+This usermod allows you to load a specific test pattern (brightness, color, effect, palette) by holding down a designated number of buttons.
+
+It was developed for quality control on pre-assembled QuinLED boards with PWM outputs to control "dumb" LEDs.   
+To test a digital output you can simply connect it to an addressable LED, which will either work properly or not.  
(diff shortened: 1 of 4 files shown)


blazoncek:
You can test brightness ramp using default behaviour of button 1 and color selection using button 0.
No additional code needed.

PaoloTK:
> You can test brightness ramp using default behaviour of button 1 and color selection using button 0. No additional code needed.

The built in functionality wasn't enough for our needs. It requires multiple button presses which is complicated on the assembly line and increases cost, and the color cycle of button 0 doesn't include colors that light up the dedicated white chip(s) of RGBW or RGBCCT lights.

PR #4090, opened 2024-08-04
Rename wled00.ino to wled_main.cpp

willmmiles:
PlatformIO's Arduino compatibility shim forces an intermediate .cpp file to be recreated and rebuilt on every build.  Apart from being wasteful of time, this can also cause problems when batch building at the command line when the VSCode IDE is trying to build the cpp plugin configuration, resulting in failed builds as the intermediate file is rewritten while being built.

As the Arduino IDE is no longer supported by this project, it should be safe now to rename the .ino to eliminate these small headaches.


softhack007:
Thanks @willmmiles I think its reasonable to cut the last ropes for arduino IDE - @blazoncek what do you think?

blazoncek:
TBH IDK. 🤷‍♂️
It does not bother me.

softhack007:
> Not to lose focus, but if renaming it is on the table, why the "00" at all? Would a _better_ name be considered?

@robertlipe the "00" is a relic from ArduinoIDE - Arduino expects to find the main sketch in `<Folder>/<Folder>.ino`. wled00.ino is basically a stub that calls WLED::instance().setup() and WLED::instance().loop(). We could rename the file as `wled_main.cpp` if that is more descriptive. What do you think?

softhack007:
🤔 my first local build ended with "NeoPixelBusLg.h not found". After deleting `.pio` and `HOME/.buildcache` and `.vscode/cpp_properties.json`, the build was successful.

@willmmiles @blazoncek can you double-check that local builds still work for you?

PR #4094, opened 2024-08-09
Improve and introduce measures for consistent code style

Mattstir:
I'm highly passionate about clean code and a consistent code style, so I see this as a great opportunity to enhance WLED. 😉

By introducing ESLint, we can automatically enforce code style during each CI run (and hint at problems during a local `npm build`), ensuring that future changes adhere to the established guidelines. For now, I've aligned the rules with the suggestions in WLED's `CONTRIBUTING.md` and made them mandatory. This is now only a basic start, these rules can be easily expanded in the future to include more than just style.

I've also corrected the existing style issues in the `.js` files and js snippets in html files.

Note: This PR is purely stylistic with no logic changes. I hope you like it! 😄

### DIFF
--- a/.github/workflows/wled-ci.yml
+++ b/.github/workflows/wled-ci.yml
@@ -81,7 +81,7 @@ jobs:
 
 
   testCdata:
-    name: Test cdata.js
+    name: Test & Lint
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -92,3 +92,4 @@ jobs:
           cache: 'npm'
       - run: npm ci
       - run: npm test
+      - run: npm lint
--- a/.github/workflows/wled-ci.yml
+++ b/.github/workflows/wled-ci.yml
@@ -92,4 +92,4 @@ jobs:
           cache: 'npm'
       - run: npm ci
       - run: npm test
-      - run: npm lint
+      - run: npm run lint
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,8 @@ Please make all PRs against the `0_15` branch.
 When in doubt, it is easiest to replicate the code style you find in the files you want to edit :)
 Below are the guidelines we use in the WLED repository.
 
+For javscript content you can use `npm lint` (or just the normal `npm run build` that also executes lint) to make sure you ather to our style.
+
 #### Indentation
 
 We use tabs for Indentation in Web files (.html/.css/.js) and spaces (2 per indentation level) for all other files.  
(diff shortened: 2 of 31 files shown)


Mattstir:
Also note that each style rule got introduced in its own commit so we could easily revert one of them if not wanted ;)

softhack007:
@Mattstir interesting idea 😃

Two questions come to my mind:
* are these rules only applied to JS and html? Because we already have a problem with flash size, and additional "beautification" in the embedded webpages could lead to increased flash size (bad).
* do you have a list of the rules? Because I think it's better to first agree on the rule set, and then automating checks as a second step.

Mattstir:
@softhack007 Thanks for your interest :D
 
* With this pr applied to `.js` files and the js snippets that are inside the html files. Of course, bundle size is a very important issue and I was thinking about that when I added this. After each rule I added, I ran `npm build`, but nothing changed in the built files. So I assume that the minifying tool you guys use `html-minifier-terser` is already stripping whitespaces and comments (I haven't fully researched how you configure it now, but as I basically only changed whitespaces and these seem to be stripped again, I'm optimistic 🤩).

* You can find the rules I added in the configuration file I added `eslint.config.mjs`.
First, I extended the recommended rules from https://eslint.org/docs/latest/rules/ (see the rules that have ✅), but I switched some of them to off or just to warn not to flag any current code with errors.
Then I added some style rules from https://eslint.style/rules 
```
    "@stylistic/js/function-call-spacing": ["error", "never"],
    "@stylistic/js/key-spacing": "error",
    "@stylistic/js/keyword-spacing": "error",
    "@stylistic/js/no-extra-semi": "error",
    "@stylistic/js/no-whitespace-before-property": "error",
    "@stylistic/js/space-before-blocks": "error",
    "@stylistic/js/space-infix-ops": "error",
    "@stylistic/js/spaced-comment": "error"
```
I selected these based on the desired style in the `CONTRIBUTING.md` file.

ingDIY:
> I'm highly passionate about clean code and a consistent code style, so I see this as a great opportunity to enhance WLED. 😉
> 
> By introducing ESLint, we can automatically enforce code style during each CI run (and hint at problems during a local `npm build`), ensuring that future changes adhere to the established guidelines. For now, I've aligned the rules with the suggestions in WLED's `CONTRIBUTING.md` and made them mandatory. This is now only a basic start, these rules can be easily expanded in the future to include more than just style.
> 
> I've also corrected the existing style issues in the `.js` files and js snippets in html files.
> 
> Note: This PR is purely stylistic with no logic changes. I hope you like it! 😄

WOW code formatting and cleaning is something I also would like to implement in WLED, but also for C and H files.
we could use [uncrustify](https://github.com/uncrustify/uncrustify).

I'll follow this thread!

PR #4103, opened 2024-08-15
Remove repeating code to fetch audio data

netmindz:
Remove pointless code duplication


blazoncek:
This optimisation is great, however..

Instead of modifying usermod manager with a *single usermod* extension it would be much better IMO to add a `static` function in `FX.cpp` to work arround the underlying issue.

I.e..
```c++
static um_data_t *getAudioData() {
  um_data_t *um_data;
  if (!usermods.getUMData(&um_data, USERMOD_ID_AUDIOREACTIVE)) {
    // add support for no audio
    um_data = simulateSound(SEGMENT.soundSim);
  }
  return um_data;
}
```

netmindz:
Updated to suggested form, I was going to add to utils.cpp at first, the top of FX.cpp makes more sense though

This version is also more performant than when I moved to the UserModManager so definitely good suggestion @blazoncek

PR #4111, opened 2024-08-20
replaced PWM LUT with calculation

DedeHai:
Calculation is a close match to the original CIE brightness formula discussed in [issue 2767](https://github.com/Aircoookie/WLED/issues/2767) : calculated values deviate a maximum of 2 and zero for the lowest 22 brightness values. 
The calculation is optimized to use as few floating point calculations as possible and even takes different bit depths.
Checked values for 8,9,10,11,12,13,14 bits, they are all calculated correctly. Tested dimming on ESP32 C3 and it works just the same as before.
Saves about 400bytes of flash compared to the LUT.


wled00/bus_manager.cpp
```diff
@@ -497,45 +497,22 @@ uint32_t BusPwm::getPixelColor(uint16_t pix) {
...
+  #else  // use CIE brightness formula
+  unsigned pwmBri = (unsigned)_bri * 100;  
+  if(pwmBri < 2040) pwmBri = ((pwmBri << _depth) + 115043) / 230087; //adding '0.5' before division for correct rounding
+  else {  
+    pwmBri += 4080;
+    float temp = (float)pwmBri / 29580;
+    temp = temp * temp * temp * (1<<_depth) - 1; 
+    pwmBri = (unsigned)temp;
+  }
+  Serial.println(pwmBri);
```
blazoncek, line 515: Please comment out or use DEBUG_PRINTLN.

blazoncek (commented):
Why did you choose not to use the same for ESP8266?

wled00/bus_manager.cpp
DedeHai: ups... :)

DedeHai:
> Why did you choose not to use the same for ESP8266?

I thought there was a reason for ESP8266 doing it differently. but sure can.

blazoncek:
I'll tag @mcer12 to review and verify if the thing works as he envisioned.

blazoncek:
I see no downside for the implementation and if @mcer12 will find a flaw we can adjust later.

PR #4124, opened 2024-09-01
Copy Segment FX

DedeHai:
Added an FX that copies the selected source segment:
- 1D and 2D supported
- brightness of segment is relative to source segment
- optionally shifts the color hue
- invert, transpose, mirror etc. work
- if source and target do not match in size, smallest size is copied
- unused pixels fade to black (allows overlapping segments)
- if invalid source ID is set, segment just fades to black
- added a rgb2hsv conversion function as the fastled variant is inaccurate and buggy

note: 1D to 2D and vice versa is not supported

## Summary by CodeRabbit

- **New Features**
  - Introduced a new lighting effect that allows one segment to copy colors from another segment, with options for color adjustments and axis flipping.
  - Added advanced color adjustment capabilities, enabling hue shifting, brightness enhancement, and saturation reduction for copied colors.

- **Improvements**
  - Enhanced effect customization with new parameters for fine-tuning color appearance during segment copying.


DedeHai:
just realised: if the source segment is inverted, the targed does not invert (which is ok) but if the source is mirrored, only half of the pixels are copied unless the mirror settings in the target segments are the same.
Two options: 
-this is a feature, i.e. leave it up to the user to make the settings identical
-silently copy the 'mirror' and 'transpose' setting from the source segment (without updating the UI)

willmmiles:
> just realised: if the source segment is inverted, the targed does not invert (which is ok) but if the source is mirrored, only half of the pixels are copied unless the mirror settings in the target segment are the same.

My quick suggestion would be to use `.width()`, etc. on the source segment instead of `.virtualWidth()`.   This should essentially copy the rendered output, including invert, mirror, transpose, etc.  Continue to use `virtualWidth()` for sizing the destination segment, though, and let the destination segment options be applied as normal.

I don't have a good setup for testing this but I'll give it a go when I can.

DedeHai:
I did not know about this distinction, that is actually what I was looking for :)
will change and test.

DedeHai:
that does not seem to work as intended. the numbers are correct (i.e. the width is not halfed if mirrored) but one half still stays black. 
I am not sure what the best way to proceed is. 
From a user perspective, I would expect that 'copy' means 'do the same as the source segment' meaning also the mirroring and grouping/spacing should be copied.

willmmiles:
> that does not seem to work as intended.

We are being foiled by `getPixelColor()`'s internal processing, I think, it's still doing some logical to physical pixel conversion.  Possibly we'll need to do our own iteration using the raw segment coordinates (start, stop, spacing, group).

wled00/FX.cpp
```diff
@@ -93,7 +94,58 @@ uint16_t mode_static(void) {
...
+      for (unsigned x = 0; x < cx; x++) {
+        for (unsigned y = 0; y < cy; y++) {
+          sourcecolor = strip._segments[sourceid].getPixelColorXY(x, y);   
+          if(SEGMENT.custom2 > 0) // color shifting enabled
+          {
+            CHSV pxHSV = rgb2hsv(sourcecolor); //convert to HSV
+            pxHSV.h += SEGMENT.custom2; // shift hue
+            hsv2rgb_spectrum(pxHSV, sourcecolor); // convert back to RGB 
+            
+          }
```
willmmiles, line 121: This gets copy and pasted below - probably it wants to be a little static function instead.

DedeHai: indeed, can you help out? I do not know what the best way to implement it is. just adding a function?

DedeHai:
I tested just copying the source settings by doing
`SEGMENT.mirror = strip._segments[sourceid].mirror;`  for example. that works, even UI gets updated (though not consistently)
by doing this I found a nasty bug.
if both x and y mirroring are set in a segment, the flipping is done incorrectly. the error is in line 214 of FX_2Dfcn.cpp (sorry, don't know how to reference that here directly). what I have is a 16x16 matrix, two segments, one x= 0-7, one x=8-15, both have y=0-15. if I set x and y mirroring in the right hand segment (the on going x=8-15) one quarter gets transferred to the first segment. that is independend of my copy function. here is what it looks like (line is the split between segments)
image
I lack the underlaying logic to fix it.

blazoncek:
@DedeHai looks like start offsets are missing.
```c++
        strip.setPixelColorXY(start + width() - xX - 1, startY + height() - yY - 1, tmpCol);
```

DedeHai:
> looks like start offsets are missing.
correct, can confirm that adding `start` fixes it.

DedeHai:
did some testing on the latest commit. it works fine with one exception: setting spacing other than 0 seems to screw things up. the target segment does weird things and I have no idea how to fix it.

willmmiles:
> I tested just copying the source settings by doing
SEGMENT.mirror = strip._segments[sourceid].mirror; for example. that works, even UI gets updated (though not consistently)

This feels wrong to me - I want to set my own mirroring or spacing settings on the destination segment and they should be respected, not overridden.  However I expect the 'source data' to be the rendered output, including the source's mirroring/reversing/grouping etc.

Spacing is where things get tough, though.   Spacing+offset is for creating interleaved segments.   My gut feeling is to pretend the gaps don't exist when copying -- leave it to the destination segment to decide how to render those.  I think this might be a bit tricky; I'll sketch some code.

wled00/FX.cpp
DedeHai: oh, I though you meant some fancy C++ macro magic ;) 
but good point, will add this to color utils

DedeHai:
> This feels wrong to me - I want to set my own mirroring or spacing settings on the destination segment and they should be respected, not overridden. However I expect the 'source data' to be the rendered output, including the source's mirroring/reversing/grouping etc.

it does feel wrong, but you cannot copy data that just is not there. like if source is mirrored, target cannot be 'unmirrored' without re-calculating the FX. 'reverse' is still possible (and is not copied).
offset is (currently) not copied and seems to work fine.

what I did not test at all yet: what happens if global buffer is disabled.

willmmiles:
> it does feel wrong, but you cannot copy data that just is not there. like if source is mirrored, target cannot be 'unmirrored' without re-calculating the FX. 'reverse' is still possible (and is not copied).

I don't think that's true - your original logic with `virtualLength()` and `getPixelColor()` calls operates on an unmirrored/unreversed/ungrouped "virtual pixel" index space, so that the FX code itself doesn't need to know about those flags.  `getPixelColor` knows how to "undo" all of those features, since they're performed in `setPixelColor`.

That was the original concern that spawned this discussion, I thought; we wanted the copy to operate on the "rendered" mirrored/reversed/grouped FX output, so if the destination segment had no flags set, it would look completely identical to the source segment.  I would then expect that any flags set on the destination are applied normally to the rendered copy.   To implement this, it looks like we'd need a new "getRenderedPixelColor()" that has slightly different index calculation logic.   I have a working 1D version at home, sorry I didn't get time to publish it last night.

That said: if we don't want to go that way, I think the reasonable alternative is to ask the user to manually harmonize the segment flags they want.  For example, if the source segment is 8x16 (mirrored, so the FX calculates 8x8), I think it's reasonable for the copy operation to treat that as an 8x8 source (eg. exactly your original code); and then it's up to the user to set the mirror flag on the destination segment if they also want it mirrored.

To sum up: I think either we should copy from either the "rendered" space, for which we need new getPixelColor flavors; or from the "FX" space, ignoring the source segment flags, as your original code does.  In either case the destination segment render flags should be set by the user, I really don't think we should override them.

Note I've completely passed over "spacing" and "offset" here -- IMO these options should be ignored by the copy operation as their purpose is to specify which pixels are /not/ part of the segment, and thus shouldn't be considered for the copy.

DedeHai:
I tried using .length() instead of virtual and confirmed the loop went over the full index range but if mirror of source is active, it sill only gets half of the data, the rest is still black (which is weird).  
also if transpose was active, not using virtual did not result in a correct copy. 
Anyway: the source buffer will contain the 'mirrored' FX data, so it does not matter whether the user sets the destination to 'unmirrored', reversing will do the same thing.
I also played a little with manual setting of spacing and grouping (which also requires to use `virtualLength()` to be used) and found that it actually works but did not work if I automatically assigned the value.
I am not sure adding a new 'getPixelColor' will be necessary but I am curios on what that would look like.
I also have a solution for the 'copy source settings': since this is an FX: it has user selectable checkboxes. I added one already but the code to copy the settings is not done yet. 
I also found that using `fadeToBlackBy()` is really slow (and maybe buggy?). It worked in 2D but when I switched to 1D with just 128+128 LEDs the FPS would (sometimes) drop to 8FPS (!). That went away without the `fadeToBlackBy()` so I am pretty sure that was the cause (I have no explanation on why it would be so slow, and only in 1D).
I also may have fixed the need for that `shift_hue()` function: 
there is no need to distinguish the code between 1D and 2D, a 1D strip can be treated as a nx1 matrix (unless that is really slow for large strips? need to test some more. For `WLED_DISABLE_2D` there can be an ifdef with different code.

when you say 'rendered space' that is the buffer that is sent out and 'FX space' is that same buffer but with 'mapped' access right? I still don't fully grasp on what is mapped how as in the code it is really hard to grasp where a pixel color actually gets written to (or read from).

blazoncek:
> we'd need a new "getRenderedPixelColor()"

I did implement "raw" get/setPixelColor (in one of my POCs) that did no translations but it proved wrong and there were always ways that it did not work properly.
I'm not fond of copying segments, but that does not matter. I would suggest you rely on regular getPixelColor() and operate from destination segment. If you will exceed coordinates from source segment you'll get black so you'll copy only "active" pixels from source (however many there are).
"active" are what virtualLength, virtualWidth and virtualHeight are returning. Those are the pixels that effect operated on.
mirroring, reversing and transposing are post rendering activities, hidden from effect functions.

willmmiles:
> I tried using .length() instead of virtual and confirmed the loop went over the full index range but if mirror of source is active, it sill only gets half of the data, the rest is still black (which is weird).
> Anyway: the source buffer will contain the 'mirrored' FX data, so it does not matter whether the user sets the destination to 'unmirrored', 'reverse' will do the same thing (flips one axis).

This is the expected behavior of `getPixelColor` - it's not accessing the source buffer directly, it's accessing the "virtual pixel" space, described by `virtualLength` etc.  (what I called "FX" space earlier) .  Mirrors/reverse/etc. do not apply to `getPixelColor`, you can think of them as being applied later.

> when you say 'rendered space' that is the buffer that is sent out and 'FX space' is that same buffer but with 'mapped' access right? I still don't fully grasp on what is mapped how as in the code it is really hard to grasp where a pixel color actually gets written to (or read from).

I'd defined "render space" as "the pixels that are actually written to" without any gaps.  This might just be my idiosyncratic way of of thinking about it though...

The way I think of it is:

`Segment.setPixelColor()`/`Segment.getPixelColor()` accept indexes in to "virtual pixel" space for each segment.  These are the pixel indexes that are used by FX.

[ 1, 2, 3, 4 ]

FX space is mapped to a "render space" via mirror, transpose, reverse, grouping.  (There are currently no functions to access this; I had proposed it as new concept to use as a copy source.)

mirror - note length is doubled from FX space
[ 1, 2, 3, 4, 4, 3, 2 1]

grouping=2, mirror, reverse - length x2 for mirror, x2 again for grouping
[ 4, 4, 3, 3, 2, 2, 1, 1, 1, 1, 2, 2, 3, 3, 4, 4]

"Render space" is then mapped to "strip space" with start, stop, spacing and offset.  Note that spacing is applied in between groups.  "strip space" can be accessed via `strip.setPixelColor()` and `strip.getPixelColor()`.  (Xes are pixel indexes that are not part of the example segment.)

start=3, spacing = 1, offset = 0, mirror
[ X, X, X, 1, X, 2, X, 3, X, 4, X, 4, X, 3, X, 2, X, 1, X]

start=0, spacing = 1, offset = 1, mirror
[ X, 1, X, 2, X, 3, X, 4, X, 4, X, 3, X, 2, X, 1]

start=0, spacing = 1, offset = 0, grouping=2, mirror, reverse - length x2 for mirror, x2 again for grouping
[ 4, 4, X, 3, 3, X, 2, 2, X, 1, 1, X, 1, 1, X, 2, 2, X, 3, 3, X, 4, 4, X]

Finally "strip space" is mapped to actual LED addresses using ledmap.json.   (Not shown, but I mention for completeness.)

-----------------------------------------

Anyways it sounds like the right thing to copy is FX space.  I still feel strongly that we should not be editing the segment properties in the FX function -- the software expects those to be inputs, not outputs, and you'll have a bunch of weird side effects on the UI and config management as the software doesn't expect those to be writable by FX.

> there is no need to distinguish the code between 1D and 2D, a 1D strip can be treated as a nx1 matrix (unless that is really slow for large strips? need to test some more. For WLED_DISABLE_2D there can be an ifdef with different code.

Yeah I'd had a feeling we could get away with something there, but I hadn't worked through all the implications yet.   There's also the 1D->2D conversion to consider; `setPixelColor(index)` knows how to do the conversion using the standard 2D expansions.

DedeHai:
Ah get it, if you access outside of 'virtual space' it does not return a valid color, hence getting black.
I agree that operating in 'FX space' is the safe way to do it and all the safety checks are already in place. 
Not modifying the properties is probably also a valid point, especially if it does not get properly mapped back to the UI (which is what I saw, sometimes it would, sometimes it would not) it will result in incorrect presets.
a workaround would be to have that checkmark I mentioned and setting the properties just for rendering, then reverting them back. would that work or is data to the UI being processed in parallel (i.e. not just between frames)?

wled00/FX.cpp
softhack007: if the `shift_hue` function is on a critical path (like pixel copy), then it would be smarter to put the static function into fx.cpp so the compiler can inline. Inlined functions are faster (no overhead due to call-and-return). 

But the compiler will not inline things that are defined in a different .cpp file.

softhack007:
> did some testing on the latest commit. it works fine with one exception: setting spacing other than 0 seems to screw things up. the target segment does weird things and I have no idea how to fix it.

🤔 I think @willmmiles has described it quite accurate, maybe just a bit over-compilcated.

* segments are the "drawing canvas" --> reverse, transpose, mirror, grouping and spacing are only applied at the moment when going from `segment.setPixelColorXY` to `strip.setPixelColorXY`
* strip is in intermediate layer, which basicially performs mapping (ledmaps or gapmaps) when going to down to `busses.setPixelColor` (no XY any more)
* busses manages double buffering and forwards pixels to the proper bus driver instance

As you want to copy effects ( = segments), I thinks its enough to copy what `segment.getPixelColorXY` provides - the raw pixels without grouping and spacing.

The target segment might not be the same size as the source segment -> so maybe implement your own version of "mirror" and "mirrorY", to preserve the overall look. But don't try to duplicate segments settings because the result will not look the same any way. Dimensions might be completely different, so the mirror axis will usually not match with the source segment.

softhack007:
PS: if you need a few nasty testcase 
- copy from a 2D segment running "Juggle" with "pinwheel" mapping, into another 2D segment with a different layout
- copy from a 2D segment running "Pacifica" with pixels mapping, into a small 1D segment with grouping=2

softhack007:
> edit:
> addendum: I tested to use 1D and 2D (i.e the XY versions) of getPixelColor and setPixelColor on a 2D strip and see no difference in speed (looked at FPS). There may be a tiny advantage on the 1D version but it is negligible

I'd say its better to always use the "right" versions - 1D or 2D - if you are on `segment` level. The "not XY" getPixelColor and setPixelColor versions will perform 1D->2D mapping, which is maybe not what you want.

Actually this function is a good example how to handle 1D and 2D
https://github.com/Aircoookie/WLED/blob/b9080f9479fc18bc1b3ba28ba55ec546ea256e87/wled00/FX_fcn.cpp#L1101-L1111

(just except for the `if (is2D())` inside the for loop, which costs performance)

willmmiles:
I think a lot of the challenge here comes from some of the corner cases.

If I have a segment 1 set as1x8, with mirror enabled, I get virtualLength() == 4, values [1 2 3 4], and what is drawn on segment 1 is [1 2 3 4 4 3 2 1].  If I then set "Copy" on Segment 2 (also 1x8) - what should I get?

The simple implementation (434ba3ff) will give me [1 2 3 4 X X X X], where X is off.  I have to set 'mirror' on segment 2 to make it actually look like a copy of segment 1.  This seems counterintuitive, why don't I get a "whole" copy?

Supposing we accept that logic - if segment 1 is mirrored, we should render the mirrored output on segment 2.  Mirror, reverse, and transpose all seem reasonable to copy from -- though as @softhack007 notes, copying the flags won't do what we're expecting.   Grouping?  Maybe - this is somewhat less clear.   Spacing?  Definitely not, weird stuff will happen.

What if segment 2 is 1x16, and has the mirror flag set on it?  With the naiive code, I'd get [ 1 2 3 4 X X X X X X X X 4 3 2 1]?   Or should I expect [1 2 3 4 4 3 2 1 1 2 3 4 4 3 2 1]?

This is why I added that idea of "render space" above - to capture what I think we want as the "copy source".

Anyhow, I suggest https://github.com/willmmiles/WLED/tree/pr/DedeHai/4124 as an example implementation - it's lame but I think it should copy the "obvious" way without disrespecting the target segment settings.  Note I've only tested on 1D, I don't have a 2D system on my bench right now. :(   Note that we don't need to check the source segment sizes - getRenderedPixelXY checks the coordinate validity and returns black if it's out of range - so we can also omit the fadetoblack, too.

DedeHai:
I tested temporarily setting the target segment settings to the sorce settings (which can be enabled in the FX by a check). This works but is not perfect (does not solve the things mentioned about non-equal sizes, mirroring will leave a gap in the center but that is a separate issue).
When settings in the target are changed, sometimes (with unlucky timing) they get reverted but that shows in the UI. There may be cases where it does not show in the UI and when changing the FX to some other, it may show wrong (unless the UI also sends out a segment setting update when FX is changed). 
while not perfect, this is a 'workaround' that would enable users to change the source segment settings and have all targets do the same (I imagin someone having many slave segments, it would be really annoying to have to manually change each one every time). 
Making it 'perfect' would require a semaphore or some other rework of code, which I personally think is way overkill for the edge cases. 
As for copying the source including the mirror part, that could be done by a new 'getRenderedPixelXY' function or it could be directly handled in the FX (I am not sure which one is the better approach).

willmmiles:
> I thought about this on my way to work and realized that if a getRenderedPixelXY() function is implemented, there is no need to copy the source settings (except maybe for grouping and spacing, those could still be copied the way I mentioned above or completely left to the user. So I am now in favour of using a getRenderedPixelXY(), question is just how robust does it have to be (and how much code we want to spend on edge cases).

I dug out a 2D panel and tested the implementation I sketched last night: https://github.com/willmmiles/WLED/tree/pr/DedeHai/4124  -- it seems like it handled every case I tried reasonably well, even 1D->2D conversion; though it's still missing the 2D->1D conversion case.   Give it a go and let me know what you think?

DedeHai:
I tested your version, there is a few bugs which I fixed and now it works amazingly well and does exactly what we discussed. There is still some cases I want to run but generally this is the way to go.
One thing that took me quite a while to pinpoint: 2D was all messed up until I figured out that `seg.offset` is not zero in my 2D config although that is not even an option to set in the UI (it is 6 for some reason). Before I commit my fixes: how is this to be handled? could do an if case (if there is any reason it is not zero) or find out why it is non zero and enforce it there?

willmmiles:
Apparently I don't do enough with 2D. ;)   Possibly the offset field is carrying some historical value, as it's not used in the standard 2D calculations.

I can think of two ways to approach fixing that: either (a) separate the 1D and 2D `getRenderedPixelColor` functions, to mirror the 1D and 2D `getPixelColor`; or (b) make sure that `Segment::offset` is always set to zero for 2D segments, instead of carrying around an old value.  I lean towards the latter - I don't like unused values being copied around in the config - but the more experienced WLED devs might have a different opinion.

DedeHai:
latest commit now seems to handle all test cases (even the [nasty ones](https://github.com/Aircoookie/WLED/pull/4124#issuecomment-2332539933)). There are a few things that do not work properly though:
- 2D to 1D copies only one row. since there is not really any good way all users will like (or make sens) I left it at that
- when enabling source grouping / spacing + mirroring, the target is not a 1:1 copy, even if settings are adjusted
- when using spacing, 'in between' pixels are not always set to black (may be an underlying bug/feature as that also happens in the source segment if spacing is enabled and afterwards 'reverse' or 'mirror')

willmmiles:
> when using spacing, 'in between' pixels are not always set to black (may be an underlying bug/feature as that also happens in the source segment if spacing is enabled and afterwards 'reverse' or 'mirror')

The pixels omitted by spacing are not part of the segment at all -- they're expected to belong to some other segment.   Spacing+Offset lets you set up "interleaved" segments with different FX, eg with (spacing = 1, seg1: offset=0, seg2: offset=1) you can arrange [1 2 1 2 1 2 1 2] where you can set different FX on 1s and 2s.

willmmiles:
Following on by that: I intentionally omitted source grouping and spacing in the getRenderedPixel() sketch.  Grouping could perhaps be included, but spacing doesn't make sense -- the pixels skipped over by "spacing" are (meant to be) part of some other segment; including pixels from Segment 2 in a copy of Segment 1 seems wrong to me.

In the end it was a judgement call - mirror, reverse, transpose all feel to me like they're "part of" the segment FX rendering; while grouping and spacing feel more like they're part of the rendered FX->output strip mapping.  YMMV, though.

DedeHai:
I think I have it mostly done now (if there are no objections to latest changes).
What remains though: where should the `uint32_t getRenderedPixelXY()` be put? I think it should go in FX_fcn.cpp but should it be part of the `WS2812FX` class or just a standalone function?

willmmiles:
> I have it mostly done now (if there are no objections to latest changes). What remains though: where should the `uint32_t getRenderedPixelXY()` be put? I think it should go in FX_fcn.cpp but should it be part of the `WS2812FX` class or just a standalone function? Or just keep it in FX.h?

If we're committed to this approach, I think it should be defined as a member of `struct Segment`, and implemented in FX_fcn.cpp beside `getPixelColor` -- as any future changes to `getPixelColor` or `getPixelColorXY` might also need to be applied to `getRenderedPixelXY`.

wled00/FX.cpp
```diff
@@ -94,6 +95,50 @@ uint16_t mode_static(void) {
...
+ * Copy selected segment and perform (optional) color adjustments
+ */
+// TODO: what is the correct place to put getRenderedPixelXY()?
+static uint32_t getRenderedPixelXY(Segment& seg, unsigned x, unsigned y = 0) {
+  // We read pixels back following mirror/reverse/transpose but ignoring grouping
+  // For every group-length pixels, add spacing
+  x *= seg.groupLength(); // expand to physical pixels
+  y *= seg.groupLength(); // expand to physical pixels
+  if (x >= seg.width() || y >= seg.height()) return 0;  // fill out of range pixels with black
+  #warning this check can be removed once 2D offset is fixed
```
blazoncek, line 108: FYI `offset` field is *unused* AKA reserved if the segment is 2D.

wled00/FX.cpp
```diff
@@ -94,6 +95,50 @@ uint16_t mode_static(void) {
...
+  uint32_t offset = seg.is2D() ? 0 : seg.offset; // dirty fix, offset in 2D segments should be zero but is not TODO: fix the offset
+  return strip.getPixelColorXY(seg.start + offset + x, seg.startY + y);
```
blazoncek, line 110: This will get a pixel from another segment if you are overlapping segments and use spacing.

willmmiles: The groupLength() call above should include the spacing?

blazoncek: Correct. Needed to check sources.
Still it will not handle transposed segments correctly.

willmmiles: It's supposed to copy the data post-transposition.  But I think you're correct that there's a bug here - if transpose is set, we need to swap virtualWidth() and virtualHeight() in the copy fx function.

DedeHai: I tested transposed source segment (among many other things) it copies right (but maybe I missed a case)

wled00/FX.cpp
DedeHai: ok, can remove the #warning then and keep that if statement.

wled00/FX.cpp
willmmiles: I think it might run in to trouble if the source segment x/y sizes aren't multiples of each other.  Eg, source = 3x8 transpose.

DedeHai: So far I could not find a case where transpose does not work, tested various sizes (3x16, 5x16, 3x7) and its always correct. If segments are overlapping, the getRenderedPixelXY() will of course read whatever has been written to the buffer, so if a segment overwrites data, the copy function will still copy that. This can not be avoided and depends on order of segments (as does overlapping segments in general)

blazoncek:
@DedeHai please merge blending styles branch (to a new branch) to see how different blending affects your new effect.

DedeHai:
sorry for messing up the blending styles branch...
I was now able to test this and it generally works fine. There is one case that is 'incorrect': if using two segments in 2D and the source segment (the one that is copied) is set to 'transpose' the blending style is of course not transposed on the other segment. For example if I set blending to 'Swipe left' one segment is blended left-right, one up-down, so not an exact copy anymore. This can not be avoided unless special treatment of blending the copy FX is implemented.
Blending style in general does not play so nicely with segments: 
-if I have two segments, one running akemi, one crazy bees and I switch to a preset with one segment playing akemi, no blending is done (if the segment playing akemi is segment 0 in both cases). If I then switch back from 1 segment akemi to a different 2-segment preset,  the second segment blends from crazy-bees (i.e. the last FX that was set on this segment number)
-I also have had two occasions, where FX parameters were set to 'default' (not the default from the FX but all sliders centered) instead of what was saved in the preset, which with copy segment is obvious as it just goes to black with a wrong source selected (not sure this is related to blending styles but never saw it before)

wled00/FX.cpp
```diff
@@ -93,6 +94,37 @@ uint16_t mode_static(void) {
 }
 static const char _data_FX_MODE_STATIC[] PROGMEM = "Solid";
 
+/*
+ * Copy a segment and perform (optional) color adjustments
+ */
+uint16_t mode_copy_segment(void) {
+  uint32_t sourceid = SEGMENT.custom3;
+  if (sourceid >= strip._segments.size() || sourceid == strip.getCurrSegmentId()) { // invalid source
```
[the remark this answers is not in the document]
netmindz: The question I would ask is why are these variables visible then if there is an API that should be used for access? We should fix the visibility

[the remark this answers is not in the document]
blazoncek: > why are these variables visible

Code size.

My first iterations used `private` variables but the code size reduction was tremendous (about 12kB so roughly 10%) when using "hidden" `public` variables instead.
See `SEGMENT` macro at the top of `FX.h`.

DedeHai:
> As the underlying issue is "synchronised" segments (i.e. segments that mimic each other) a correct approach would be to execute effect function original segment uses on the destination segment with parameters (and timers and PRNG) of the original segment.

that would indeed be the best approach BUT: how do you control it? are FX with the same parameters always in sync (if started with the same timebase)? This whole endevour is mostly to sync particle effects, which are frame based (and I have no intention to change that as it makes collisions and speeds way more complex to calculate and therefore a lot slower, the only workaround would be to have fixed frame rates but that poses new issues). 
so with each segment running a particle FX, you need to track the state of the PRNG so it can be repeatedly called by each segment with the same outcome (did not think this through all the way, there may be hidden issues with this approach). While this sounds doable, I don't really see the advantage over a 'copy segment' function. And I also do not see why this would lead to frustration on the user side.

blazoncek:
Consider the complexity of overlapping segments.

DedeHai:
overlapping segments work fine, but you are correct that it is a bit complex i.e. the segment layering has to be done correctly. This is already the case for overlapping segments, adding copy FX on top does not add much to complexity. 
A short description in the KB on how to use the copy FX should be sufficient. 
IMHO having a feature that is useful for most users but not perfect for everyone is better than not having it at all.

wled00/FX.cpp
```diff
@@ -93,6 +94,37 @@ uint16_t mode_static(void) {
...
+      uint32_t cl; // length to copy
+      for (unsigned i = 0; i < SEGMENT.virtualLength(); i++) {              
+        sourcecolor = SEGMENT.getRenderedPixelXY(strip._segments[sourceid], i);           
+        SEGMENT.setPixelColor(i, adjust_color(sourcecolor, SEGMENT.intensity, SEGMENT.custom1, SEGMENT.custom2));
+      }
+    } else { // 2D source, note: 2D to 1D just copies the first row (or first column if 'Switch axis' is checked in FX)    
+      for (unsigned y = 0; y <  SEGMENT.virtualHeight(); y++) {
+        for (unsigned x = 0; x < SEGMENT.virtualWidth(); x++) {  
+          if(SEGMENT.check2) sourcecolor = SEGMENT.getRenderedPixelXY(strip._segments[sourceid], y, x); // flip axis (for 2D -> 1D, in 2D Segments this does the same as 'Transpose')
+          else sourcecolor = SEGMENT.getRenderedPixelXY(strip._segments[sourceid], x, y);
```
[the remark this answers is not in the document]
DedeHai: these calls are inside a loop, using `strip.getSegment(sourceid)` adds overhead, direct access would be faster. 
How about saving it once, then using it multiple times:
` Segment sourcesegment = strip.getSegment(sourceid);`

blazoncek:
I disagree.

Consider the following valid example:
Segment 0: spans entire matrix and displays Frizzles (or any asymmetric effect)
Segment 1: spans left half of the matrix and displays Black Hole (or whatever that does not paint every pixel)
Segment 2: spans right half of matrix and copies Segment 1

Segment 2 will show artifacts from underlying Segment 0 which is unwanted. I would consider this a flaw and would report a bug as I only want to copy Segment 1 and not the underlying segment.

If it is to be done, it has to be done correctly or not at all.

willmmiles:
I think the implementation of "copy the rendered buffer including all underlays at that segment layer" is probably the correct behavior.  It's true there might be some counterintuitive cases, but it offers a lot of opportunity for clever compositions.  With some easy extensions, there are some really cool tricks we could use it for.

- We could add a 'chroma key' feature, so there's a particular color (eg. black) that doesn't get copied
- We could also add an alpha blended form for specifically for overlaying
- Segments backed by virtual LEDs can be used as a render canvas that can then be cheaply copied/overlaid/transformed for multiple output segments for expensive fx

As with all powerful features -- like segment overlays themselves! -- there's going to be some sharp edges and cases where it takes a little planning to compose the FX you want to see.  Still, I think this feature offers a surprisingly large amount of new capability for a minimum of code.

My $0.02, though!

blazoncek:
> it offers a lot of opportunity for clever compositions

Yes if that was optional. Now it is not. The solution is to optionally enable individual segment buffer for rendered pixels. So that user has an option if he/she wants underlying segments copied or not.

> there are some really cool tricks we could use it for

I do not disagree with that.

To answer @DedeHai regarding "effect blending styles": Yes, not all styles will look great with every effect. I particularly dislike "push" variants as the outcome may be *weird* to say the least. Yet, they are there to be in line with #3669 . Still, the user has an option to select them or not. The code is also largely untested and there may be bugs still present.

DedeHai:
I fully agree with @willmmiles: while there are some limits in the current state of things, this PR adds a lot of value for the average use case. 

> Yes if that was optional. Now it is not. The solution is to optionally enable individual segment buffer for rendered pixels. So that user has an option if he/she wants underlying segments copied or not.

I see problem in adding this later when (or if) individual segment buffers are implemented

> Still, the user has an option to select them or not.

same goes for the copy FX :)

DedeHai:
just FYI: with the latest commit the code size shrinks by 1.2kB, making it 200bytes smaller in total than vanilla 0.15 :)

blazoncek:
You have started to mix two different things into this PR.
Please reserve optimisations for your other PR.

DedeHai:
that is a good point. the changes are kind of 'required' to improve this PR, but keeping optimizations separated is cleaner. Will move it over to the optimization PR and revert the changes here.

DedeHai:
ready to merge from my side. It will need adjustments once segment layering is added.

willmmiles:
Might be worth waiting for layering.  Once that's in, "copy" becomes truly trival - it could point right at the source FX segment buffer.

wled00/colors.cpp
```diff
@@ -83,6 +83,23 @@ uint32_t color_fade(uint32_t c1, uint8_t amount, bool video)
...
 }
 
+/*
+ * color adjustment in HSV color space (converts RGB to HSV and back), color conversions are not 100% accurate!
+   shifts hue, increase brightness, decreases saturation (if not black)
+   note: inputs are 32bit to speed up the function, useful input value ranges are 0-255
+ */
+uint32_t adjust_color(uint32_t rgb, uint32_t hueShift, uint32_t lighten, uint32_t brighten) {
+    if(rgb == 0 | hueShift + lighten + brighten == 0) return rgb; // black or no change
+    CHSV32 hsv;
```
blazoncek, line 93: Since CHSV32 is custom HSV class, I would suggest to incorporate constructor from RGBW32 and `uint32_t` which should hide explicit calls to HSV conversion.

DedeHai:
Updated to work with latest changes. It now copies the source segment buffer, truly copying a single segment.

In addition I added the option to copy "full stack" by reading the last rendered frame: this should not be used with an overlapping "copy segment" as it copies itself (unless the user want that effect to happen)

the idea for full stack copy is:
- create a virtual strip/panel besides the physical LEDs (I used DDP with IP 0.0.0.0 as a test)
- Create a layered segment stack on the virtual strip/panel
- On the physical LED segment: copy FX of the virtual segment with "full stack" checked

The full-stack copy is not the cleanest most user-friendly solution. Its a bit convoluted to use and also lags one frame behind as it copies from the global frame buffer which is updated after all segments were rendered. Although not a perfect solution for a full-stack segment copy, I still think the option should be available: on larger setups with say 3-overlapping layers, a copy would require another 3 layers for each segment that the original is being copied to, eating up way more ram. AND: the option of just copying each segment is still available.

DedeHai:
I did run some tests and everything seems to be working as intended.
Ready to merge.

wled00/colors.cpp
DedeHai: constructor is implemented in https://github.com/wled/WLED/pull/4615 but it uses the fastled equivalent "rainbow" conversion for nicer colors, however, here we want the technically correct spectrum conversion to avoid color shifts.

PR #4126, opened 2024-09-02
fixed offsets and inverted signal plus dead time

DedeHai:


blazoncek:
I think this does it. I tried to optimise a bit (inverted signal is inverted on all channels so needs no special treatment).
Also named variables to be more logical.
When you have time please verify if it still works as intended. 😇

DedeHai:
it does not ;)
- dead time must be added regardless of hpoint
- dead time must be added BEFORE inverting
- hpoint shift must be reversed if inverted (inverted means on if signal is low, so low pulses must not overlap. when using an H-bridge inverting is only required if not using any driver i.e. built with transistors. there the signals can be inverted (low is on).
I will fix these points, test and commit.

blazoncek:
Turns out this is a nice brain gym! 😄

I think there is no distinction between normal and inverted signal. It is just the logical interpretation of the brightness slider orientation. If it is set to 32 or 224, dead time follows the pulse regardless of its width.

And hPoint is just the amount of time (in single bit pulses) that need to pass before the output goes high.

So if our duty is 32 with a dead time of 2 our hPoint will be 34. Similar when 224, it will be 226. So the second output will be delayed appropriately. Remember that the second output is also inverted and the sum _data[0]+_data[1] is always less or equal to 255.

DedeHai:
you forgot two things:
a) if you invert before adding dead time, negative pulses are enlarged not shortened
b) on inverted signals, the low pulse must be shifted 'to the left'
this is all very easy to see on a scope ;)

blazoncek:
I will trust you on this. 😄 

There is a small error though. `deadTime` is `unsigned`.

blazoncek:
@DedeHai are you comfortable if I merge this? I will then PR bus config branch.

Issue #4117, opened 2024-08-26
Dynamic UI for LED settings 


netmindz:
The BusManager supplies the list of LED Types

Once a type has been selected then config requirements are returned for this type. 

This needs to cater for this like which pins are needed, colour order etc

For each required value the following are required

- Label - e.g Data Pin
- Type - pin / numeric / option 
- Type info - min/max for numeric, key-value for options 
- config key - where to save this value

blazoncek:
Note: This enhancement may be part of (or prequel to) Settings rewrite which "is supposed to read all information from cfg.json file".
As cfg.json does not contain any field descriptions (apart from key names) or limits or "lists" this may need to be delegated to the firmware in the form of current xml.cpp constructs (injected JavaScript) or similar.

To allow educated UI the following must also be provided for particular LED type:
- has RGB
- has White
- has CCT
- bit depth and buffer use (for memory calculations)
- name and number of "pin" fields with their min/max values or possible "lists" of key-value pairs
- uses GPIO or not
- what type of interface/driver does it need (RMT, I2S, UDP, I2C, LEDC, etc) (limit selection when HW is exhausted)
- etc/tbd

netmindz:
Should the power and memory being calculated by the UI or by the bus?

blazoncek:
If it is to be done by UI then UI needs to be informed of HW limitations and capabilities. Difficult.
If it is to be performed by firmware then there should be a HTTP endpoint within firmware to provide those calculations. UI would need to "consult" firmware (most likely using XHR) on each change and not do that too quickly to not overwhelm the CPU.
I do not know which is better. The second would be more accurate IMO except if the first would get *every detail*.

netmindz:
If the goal is that we can change/add what output options are added without needing to add LED type specific code to the UI, then it has to be on the backend that these details are calculated.

That might be taking that requirement too strongly/literally however?

netmindz:
The request payload and response should be very small so really shouldn't be an issue. If we do see issues than that's really an indication of bigger issues with the webserver that needs attention

blazoncek:
> The request payload and response should be very small so really shouldn't be an issue. If we do see issues than that's really an indication of bigger issues with the webserver that needs attention

You are welcome to join @willmmiles 's efforts in making AWS better. 😄 

Back to topic, I am starting to see issues with compile-time selections of possible bus types and their capabilities.

[...]

PR #4129, opened 2024-09-05
Fetch LED types from Bus classes (dynamic UI)

### LINKED ISSUE #4112: two wire cc cw with swapped polarity
i have a led strip that contains cold and warm leds, but it only has two wires.
image

I want two pwm channels (cc, cw) that are never high at the same time. so that i have a option to configure this type of led strip.

Use the led strip with only warm or cold light.

**Additional context**
I know a lot of chips cant handle phase shifting, but if you support it only for the easy ones (esp32 for example) i would be verry happy! (because i already solderd one..., would be awsome to add this functionality)

blazoncek:
The idea was sparked by @netmindz (#4056) to allow dynamic creation of enabled (compiled) LED types in LED settings UI instead of relying on editing HTML file to represent them.

This PR implements @netmindz principles while also improves JavaScript handling of implemented bus types in UI (no longer relies on magic numbers).

As a bonus BusPwm is enhanced to support "dithering" and "phase shifting" on ESP32. LEDC supports 4 fractional bits that enhance resolution by 1 bit using temporal statistical distribution (see ESP32 technical reference manual, LEDC chapter).
Phase shifting can be used to distribute load on PSU and, more importantly, drive H-bridge used for reverse-polarity PWM CCT LED strips (thanks goes to @dedehai for debugging signals on the oscilloscope and adjusting calculations).

The downside of this PR is increased stack buffer size for `oappend()` function as LED types are pushed into it. When the number of LED types reaches the limit of usable stack that information will have to be split into a separate HTTP/XHR request.

What remains to be implemented is:
- passing RAM and power requirements for each LED type (partially implemented by `is16bit()` function)
- detection of parallel I2S vs. RMT buffers on ESP32 (RAM and available outputs)
- better handling of available HW (avoid const.h and use HAL)
- possible additional true "virtual" or usermod buses (see #4123)
- pass information if LED type requires GPIO and how many (currently length of "t" JSON key determines it)
- there is only one "virtual" type (currently used for network DDP/ArtNet)
- if color orders are supported, off refresh is required, etc (could be stored in capabilities "c" JSON key)

Closes #4112 #4117


willmmiles:
Re 43cc4ec, a quick thought offhand regarding oappend: I think there's really little value in using a stack buffer at all, since the two places where it's used immediately turn around and copy the contents to a heap allocated buffer anyways.

With that in mind and since this limitation keeps coming up, [I sketched a proof-of-concept "remove oappend" patch last week](https://github.com/willmmiles/WLED/tree/end-oappend) -- there's some more cleanup to be done, the API change would need to be applied to all usermods, and it probably wants to be extended to use a rope-like data structure like DynamicBufferList, but it should give the general idea for a possible future direction.

wled00/bus_manager.h
```diff
@@ -6,6 +6,8 @@
  */
 
 #include "const.h"
+#include <array>
```
willmmiles, line 9: I don't think we ended up using `<array>` in the end.

wled00/bus_manager.h
```diff
@@ -21,89 +23,44 @@ uint16_t approximateKelvinFromRGB(uint32_t rgb);
...
+struct BusConfig; // forward declaration
 
 // Defines an LED Strip and its color ordering.
-struct ColorOrderMapEntry {
+typedef struct {
   uint16_t start;
   uint16_t len;
   uint8_t colorOrder;
-};
+} ColorOrderMapEntry;
```
willmmiles, line 33: This change isn't necessary - we don't need to typedef structs in C++?

wled00/bus_manager.h
```diff
@@ -122,59 +79,76 @@ class Bus {
...
+    inline  uint8_t  getAutoWhiteMode(void) const                 { return _autoWhiteMode; }
+    inline  uint8_t  getNumberOfChannels(void) const              { return hasWhite() + 3*hasRGB() + hasCCT(); }
+    inline  uint16_t getStart(void) const                         { return _start; }
+    inline  uint8_t  getType(void) const                          { return _type; }
+    inline  bool     isOk(void) const                             { return _valid; }
+    inline  bool     isReversed(void) const                       { return _reversed; }
+    inline  bool     isOffRefreshRequired(void) const             { return _needsRefresh; }
+    inline  bool     containsPixel(uint16_t pix) const            { return pix >= _start && pix < _start + _len; }
+
+    static std::vector<LEDType> getLEDTypes(void)                 { return {{TYPE_NONE, "", PSTR("None")}}; }
```
willmmiles, line 119: Do we need this?   It's not called anywhere - maybe a comment indicating that this is a mandatory call on all Bus subclasses would be sufficient.

wled00/bus_manager.cpp
```diff
@@ -499,46 +539,92 @@ uint32_t BusPwm::getPixelColor(uint16_t pix) {
...
+  // Phase shifting requires that LEDC timers are synchronised (see setup()). For PWM CCT (and H-bridge) it is
+  // also mandatory that both channels use the same timer (pinManager takes care of that).
   for (unsigned i = 0; i < numPins; i++) {
-    unsigned scaled = (_data[i] * pwmBri) / 255;
-    if (_reversed) scaled = maxBri - scaled;
+    unsigned duty = (_data[i] * pwmBri) / 255;    
     #ifdef ESP8266
-    analogWrite(_pins[i], scaled);
+    if (_reversed) duty = maxBri - duty;
+    analogWrite(_pins[i], duty);
```
willmmiles, line 575: Can we open an issue to track ESP8266 phase support?  I'll take that one on once this merge lands.

willmmiles (commented):
High level thoughts:
- This is an excellent step in the right direction
- Big thumbs up for moving from macros to constexpr
- It really makes me want to take a stab at modularizing Bus type handling/config management/construction

wled00/bus_manager.h
blazoncek: Agreed, though I thought to keep that in-line with other declarations.

wled00/bus_manager.h
blazoncek: No, we don't. Added just in case someone might convert that into virtual in the future.
Does not seem likely, though. Can be made `inline` and it may not end up in the binary. Or can be commented out.

wled00/bus_manager.h
blazoncek: 😄 
There are a lot of relics like this.

blazoncek:
> a quick thought offhand regarding oappend

`oappend` is there since ever. If you intend to work on it it may be easiest just to replace `char[]` buffer with `String` so it can grow as needed.

wled00/bus_manager.cpp
blazoncek: By all means! 😄

wled00/bus_manager.h
willmmiles: There's no such thing as "static virtual", though.  To call a virtual function, you need an object to look up the function pointer from, so they can't be static.  For now I'd suggest commenting it out.

willmmiles:
> > a quick thought offhand regarding oappend
> 
> `oappend` is there since ever. If you intend to work on it it may be easiest just to replace `char[]` buffer with `String` so it can grow as needed.

Every time I see `String::operator+`, I weep for the allocator. ;) I went with a `Print` interface instead, since it's got a rich API, can be backed by different allocator strategies depending on the context, and AsyncWebServer already had a Stream response available.

wled00/data/settings_leds.htm
```diff
@@ -18,14 +19,17 @@
...
+		function isPWM(t)  { return gT(t).t.charAt(0) === "A"; }    // is PWM type
+		function isAna(t)  { return gT(t).t === "" || isPWM(t); }   // is analog type
+		function isDig(t)  { return gT(t).t === "D" || isD2P(t); }  // is digital type
+		function isD2P(t)  { return gT(t).t === "2P"; }             // is digital 2 pin type
+		function isNet(t)  { return gT(t).t === "N"; }              // is network type
+		function isVir(t)  { return gT(t).t === "V" || isNet(t); }  // is virtual type
+		function hasRGB(t) { return !!(gT(t).c & 0x01); }           // has RGB
+		function hasW(t)   { return !!(gT(t).c & 0x02); }           // has white channel
+		function hasCCT(t) { return !!(gT(t).c & 0x04); }           // is white CCT enabled
+		function is16b(t)  { return !!(gT(t).c & 0x10); }           // is digital 16 bit type
```
netmindz, line 32: No more magic numbers makes me happy :smile:

wled00/data/settings_leds.htm
```diff
@@ -224,6 +225,41 @@
...
+						if (gT(t).t.length > 1) p0d = "GPIOs:";
+						off = "Dithering";
+						break;
+					case 'N': // network
+						p0d = "IP address:";
+						break;
+					case 'V': // virtual/non-GPIO based
+						p0d = "Config:"
+						break;
+				}
```
netmindz, line 249: This is definitely a good step in the right direction, which I hope to improve in the next PR where this is returned by the Bus type rather than hard coded in the HTML - e.g hub75 having "chain length", panel size etc

wled00/const.h
```diff
@@ -298,26 +300,36 @@
...
+#define TYPE_VIRTUAL_MIN         80
 #define TYPE_NET_DDP_RGB         80            //network DDP RGB bus (master broadcast bus)
 #define TYPE_NET_E131_RGB        81            //network E131 RGB bus (master broadcast bus, unused)
 #define TYPE_NET_ARTNET_RGB      82            //network ArtNet RGB bus (master broadcast bus, unused)
 #define TYPE_NET_DDP_RGBW        88            //network DDP RGBW bus (master broadcast bus)
 #define TYPE_NET_ARTNET_RGBW     89            //network ArtNet RGB bus (master broadcast bus, unused)
+#define TYPE_VIRTUAL_MAX         95
 
+/*
+// old macros that have been moved to Bus class
```
netmindz, line 332: Can we delete rather than comment out ?

wled00/data/settings_leds.htm
blazoncek: There may be remnants somewhere still.
Would be happy if another pair of eyes could have a look.

wled00/const.h
blazoncek: Yes, once a few review the code.

blazoncek:
@netmindz & @willmmiles would you think this is good to be merged? @softhack007 do you have any second thoughts?
This (principle), along with #4134, may make expanding settings (including and where most beneficial - audioreactive settings) much easier.
@Aircoookie please review (or at least look at) this PR and #4134 if they clash somehow with what you planned for settings rewrite.

netmindz (approved):

wled00/bus_manager.h
```diff
@@ -21,89 +22,44 @@ uint16_t approximateKelvinFromRGB(uint32_t rgb);
...
+      if (n >= count()) return nullptr;
       return &(_mappings[n]);
     }
 
     uint8_t getPixelColorOrder(uint16_t pix, uint8_t defaultColorOrder) const;
 
   private:
-    uint8_t _count;
-    ColorOrderMapEntry _mappings[WLED_MAX_COLOR_ORDER_MAPPINGS];
+    std::vector<ColorOrderMapEntry> _mappings;
```
willmmiles, line 52: Is the switch to `std::vector` intended to reduce memory usage?

One of the challenges is that `std::vector` has an exponential allocation strategy.  Depending on the relationship between WLED_MAX_COLOR_ORDER_MAPPINGS and the number of entries actually used, it may end up using *more* memory.  I checked the code for std::vector for ESP32 -- this implementation of `push_back` uses the classic "double the array length if it needs more memory" strategy.  So for `WLED_MAX_COLOR_ORDER_MAPPINGS == 10` (esp32), when you push the 9th element on to the list, the buffer will expand to 16 -- leaving 6 elements you pay for in RAM, but will never use.

What you can do to mitigate this is any of:
- You could ensure that the limits are powers of 2
- When loading the vector, figure out how many elements you need, then call `reserve()` up front to allocate exactly the right amount of memory
- Alternately, you can preallocate the maximum memory you plan to use with `reserve()` at construct time; won't save any RAM but also guarantees you'll never allocate more than you need.
- You could call `shrink_to_fit()` after filling the vector, or make a copy and discard the original std::vector object.  (Cost is the same, a complete copy and temporarily holding more RAM than it needs)

blazoncek: > Is the switch to `std::vector` intended to reduce memory usage?

No. It is to allow variable size (in the future) without the expense of pre-allocated memory. The question was "why 10 and why only 10?".
`reserve()` could be done, yes. That'll go into `cfg.cpp` and `set.cpp`. Thanks.

blazoncek:
@jw2013 please rebase your branch(es). I'm going to delete `bus-config` soon.

PR #4138, opened 2024-09-12
Speed improvements for discussion

DedeHai:
Added a lot of improvements to core functions making them slightly faster and saving some flash.
I tested all functions to work the same as before. I tested speed by looking at FPS and it's hard to say how much faster it actually is but I see some improvement, maybe 1%.
The inline of getMappedPixelIndex() is just an idea, not sure that is correct nor do I know if it improves anything.
Did not test on ESP8266.


wled00/FX_2Dfcn.cpp
```diff
@@ -173,47 +173,44 @@ void IRAM_ATTR_YN Segment::setPixelColorXY(int x, int y, uint32_t col)
...
-
   int W = width();
   int H = height();
-  if (x >= W || y >= H) return;  // if pixel would fall out of segment just exit
-
-  uint32_t tmpCol = col;
+ 
+  int yY = y;
   for (int j = 0; j < grouping; j++) {   // groupping vertically
+    if(yY >= H) continue;
```
blazoncek, line 191: `break` would do just as well as `yY` is only ever increased.

wled00/FX_fcn.cpp
```diff
@@ -912,7 +916,7 @@ uint32_t IRAM_ATTR_YN Segment::getPixelColor(int i) const
         return getPixelColorXY(i % vW, i / vW);
         break;
       case M12_pBar:
-        if (vStrip>0) return getPixelColorXY(vStrip - 1, vH - i -1);
+        if (vStrip>0) { i &= 0xFFFF; return getPixelColorXY(vStrip - 1, vH - i -1); }
```
blazoncek, line 919: You could just go with `return getPixelColorXY(vStrip - 1, vH - (i&0xFFFF) -1);`

wled00/FX_fcn.cpp
```diff
@@ -1816,7 +1820,7 @@ bool WS2812FX::deserializeMap(uint8_t n) {
   return (customMappingSize > 0);
 }
 
-uint16_t IRAM_ATTR WS2812FX::getMappedPixelIndex(uint16_t index) const {
+__attribute__ ((always_inline)) inline uint16_t IRAM_ATTR WS2812FX::getMappedPixelIndex(uint16_t index) const {
```
blazoncek, line 1823: Non obligatory: I would prefer `__attribute__` at the end but `[[...]]` in front.

wled00/colors.cpp
```diff
@@ -39,21 +39,20 @@ uint32_t color_add(uint32_t c1, uint32_t c2, bool fast)
...
-  if (fast) {
-    uint8_t r = R(c1);
-    uint8_t g = G(c1);
-    uint8_t b = B(c1);
-    uint8_t w = W(c1);
-    r = qadd8(r, R(c2));
-    g = qadd8(g, G(c2));
-    b = qadd8(b, B(c2));
-    w = qadd8(w, W(c2));
+  uint32_t rb = (c1 & 0x00FF00FF) + (c2 & 0x00FF00FF);
```
blazoncek, line 42: Nice!

wled00/colors.cpp
```diff
@@ -39,21 +39,20 @@ uint32_t color_add(uint32_t c1, uint32_t c2, bool fast)
...
-    w = qadd8(w, W(c2));
+  uint32_t rb = (c1 & 0x00FF00FF) + (c2 & 0x00FF00FF); 
+  uint32_t r = rb >> 16;
+  uint32_t b = rb & 0xFFFF; 
+  uint32_t wg = ((c1>>8) & 0x00FF00FF) + ((c2>>8) & 0x00FF00FF); 
+  uint32_t w = wg >> 16;
+  uint32_t g = wg & 0xFFFF; 
+
+  if (fast) {    
+    r = r > 255 ? 255 : r;
```
blazoncek, line 50: Perhaps `r = std:min(255,r)` may be faster?

wled00/colors.cpp
```diff
@@ -70,27 +69,56 @@ uint32_t color_add(uint32_t c1, uint32_t c2, bool fast)
 
 uint32_t color_fade(uint32_t c1, uint8_t amount, bool video)
 {
-  if (c1 == BLACK || amount + video == 0) return BLACK;
+  if (c1 == BLACK || amount == 0) return BLACK;
+  else if (amount == 255) return c1;
```
blazoncek, line 73: `else` really does no difference here.

wled00/colors.cpp
```diff
@@ -70,27 +69,56 @@ uint32_t color_add(uint32_t c1, uint32_t c2, bool fast)
...
+    addRemains |= B(c1) ? 0x00000001 : 0;
+    addRemains |= W(c1) ? 0x01000000 : 0;
   }
+  uint32_t rb = (((c1 & 0x00FF00FF) * scale) >> 8) & 0x00FF00FF; // scale red and blue
+  uint32_t wg = (((c1 & 0xFF00FF00) >> 8) * scale) & 0xFF00FF00; // scale white and green
+  scaledcolor = (rb | wg) + addRemains;  
   return scaledcolor;
 }
 
+// 1:1 replacement of fastled function optimized for ESP, slightly faster, more accurate and uses less flash (~ -200bytes)
```
blazoncek, line 90: If you do that, why not integrate this into WLED native `Segment::color_from_palette()`? Just a thought.
The downside: Some effects would need to be rewritten.

And BTW I do not like how FastLED wraps palettes. IMO they should never be wrapped. If I want "round" palette I should create one that starts and stops with the same color.

blazoncek (commented):
Excellent!
There are some formatting issues (spaces) but that's minor.

wled00/FX_2Dfcn.cpp
DedeHai: good point.

wled00/FX_fcn.cpp
DedeHai: I honestly have no Idea what the precompiler instructions mean in detail, I copied this from what they use in fastled... 
the idea behind this is to get rid of the 'function entry' instructions that are added when a function is called. When I added the `inline` flash size increased by a few bytes, telling me that it is actually inlined. Since this short function is only called from two places and is called A LOT this may be faster. I have no way to check (would need a proper debugger that shows assembly instructions being executed line by line).

wled00/colors.cpp
DedeHai: I thought about this long and hard and then came up with the idea to add (or multiply) two 8bit values in one go on the 32bit system. Also masking two values in one go is faster than shifting twice and then casting to `byte` twice. In "raw" instructions, these changes cut the instructions in half, the compiler may optimize though. Code is still smaller (I checked each of the improvements agains original code in size and they all make code smaller, meaning its also less instructions to execute)

wled00/colors.cpp
DedeHai: maybe, I have no way of checking. but the compiler is quite clever I found and does optimize.

wled00/colors.cpp
DedeHai: true, I had this removed but I shifted code around so much, that probably got lost...

wled00/colors.cpp
DedeHai: that is a valid point. I just wanted to see if I can make the function faster as a test and found that a 32bit native version is not only faster, it also is more compact and more accurate (fastled uses dirty tricks).
I think wrapping is a valid feature. many of the palettes are not "round". The palette FX is a good example where it is useful: it give a "hard color cut" if palettes are not wrapped.

wled00/FX_fcn.cpp
blazoncek: As I've recently learned these are compiler attributes.

wled00/fcn_declare.h
```diff
@@ -81,6 +82,7 @@ class NeoGammaWLEDMethod {
 [[gnu::hot]] uint32_t color_blend(uint32_t,uint32_t,uint16_t,bool b16=false);
 [[gnu::hot]] uint32_t color_add(uint32_t,uint32_t, bool fast=false);
 [[gnu::hot]] uint32_t color_fade(uint32_t c1, uint8_t amount, bool video=false);
+CRGB ColorFromPaletteWLED(const CRGBPalette16 &pal, unsigned index, uint8_t brightness = (uint8_t)255U, TBlendType blendType = LINEARBLEND);
```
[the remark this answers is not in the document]
DedeHai: I did not know what that means. now I do. makes sense.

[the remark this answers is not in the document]
blazoncek: Tells compiler that this code is used frequently and it should be optimised as such (for speed).

wled00/colors.cpp
DedeHai: I also thought about implementing a new struct that is like CRGB from fastlet, but CRGBW32 colors in 32bit, with `union` access as a native 32bit (would enforce compiler to properly align it in memory too) as well as some casting functions but my C++ knowledge is not at that level to make it good. Doing it like that _may_ improve some passing around of colors, especially with palettes being CRGB (or rewrite the whole fastled stuff to be 32bit native and do away with all the conversions for good but that sounds like a lot of work). The compiler does a good job in making it fast though so not much may be gained in speed.

wled00/colors.cpp
```diff
@@ -70,27 +69,56 @@ uint32_t color_add(uint32_t c1, uint32_t c2, bool fast)
...
-    scaledcolor  = (((r * scale) >> 8) + ((r && scale) ? 1 : 0)) << 16;
-    scaledcolor |= (((g * scale) >> 8) + ((g && scale) ? 1 : 0)) << 8;
-    scaledcolor |=  ((b * scale) >> 8) + ((b && scale) ? 1 : 0);
-    scaledcolor |= (((w * scale) >> 8) + ((w && scale) ? 1 : 0)) << 24;
-  } else {
-    scaledcolor  = ((r * scale) >> 8) << 16;
-    scaledcolor |= ((g * scale) >> 8) << 8;
-    scaledcolor |=  (b * scale) >> 8;
-    scaledcolor |= ((w * scale) >> 8) << 24;
+  uint32_t addRemains = 0;
```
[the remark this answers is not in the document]
DedeHai: it is quite simple once you understand the principle:
- `((r>>16) + 1)<<16` (which is about what the original code does) ist the same as (r + (1<<16))
- `(byte(((r>>16)*scale)>>8))<<16` is the same as (((r&0xFF0000)*scale)>>8)&0xFF0000 
its simple math rules but looks more complicated in binary/hex. just replace shifts with divisions and it becomes more ovbious.

DedeHai:
one more thing I was thinking about:
the way I do color handling in the particle system is very efficient by creating a buffer of the segment size, rendering to that and then passing it to WLED. it would be even more efficient if I would drop rendering 'mirrored' pixels (may implement that at one point). So my idea was this: create a buffer (RGB, I don't think white is needed for FX right?) that is the size of the largest segment (and stays in RAM so no fragmentation unless segments are changed a lot, not sure how buffers work currently). Render one segment to that buffer (clear to zero whatever the segment needs), the buffer is accessed as 'in order' so no mapping (except mirror and transpose) and once it is rendered (including blurring and stuff) transfer the buffer to the 'strip' using mapping. Then clear the buffer, do the next segment and so on. This would also allow to use 'add pixel color' when a buffer is transferred, making overlays play more nicely, independent of segments supporting it. If a segment FX calls 'get pixel color' it would still have to be fetched from the strip buffer though. but blurring would be a lot faster and overlapping segments could be 'transparent'. 
This is only a half-baked idea as I do not yet fully grasp how the whole 'color chain' works from segment to strip (and back) but the whole 'mapping' checks could be cut down significantly if it is not done for each 'setPixelColor' (at least I think so)

willmmiles:
>  the buffer is accessed as 'in order' so no mapping (except mirror and transpose) and once it is rendered (including blurring and stuff) 

I had the same idea!  I expect that using flat temporary render buffer of the virtual() sizes, then applying all mapping operations as a post-processing step, would yield a substantial average-case performance improvement.  That said, if you try to do mirror and transpose during the render, I think you'll give up most of the benefits as every write still ends up being a pile of variable lookups and conditionals instead of just straight local pointer arithmetic -- I think it'd be better to perform *all* mapping operations after the render.   We'd also get some advantages with code layout and IRAM caching, as each FX function wouldn't need to make so many external function calls.

The big tradeoff is RAM usage.  The catch is that the temporary buffer needs contiguous RAM.  If it's held allocated, we lose the ability to share that memory (in highly constrained systems) with other tasks like the web server... but if it's not, we risk not being able to get a big enough buffer when it's needed for a render.   There may also be systems where there just isn't enough RAM to hold two copies of every pixel state at once.

That said, I'd still love to see a PR that implements a temporary render buffer and pulls the mapping operation out to a separate pass after the FX function call.  I think the performance and code size improvements are likely to be compelling, even if it reduces the maximum LED count that can be supported on memory constrained hardware.

DedeHai:
How are buffers currently? What is the global buffer vs no global buffer?

blazoncek:
Let's start at the bottom and make our way up.

NeoPixelBus (in 2.7.9) uses 3 buffer: 2 render buffers and 1 transmit buffer. Render buffers are used in "double buffer" fashion. Transmit buffer may be small, but render buffers are of size _LED count * pixel size_ where _pixel size_ can be 24bits up to 64bits. (2.8.0 or later may change that as @makuna is working on optimisations). Different hardware (RMT, I2S, ...) may have transmit buffers of different size.

WLED creates its own global buffer of size _LED count * 32bit_ (if so selected in settings) for its `strip`. This is due to the fact that WLED uses `NeoPixelBusLg` which does luminance adjustments after each `SetPixelColor()`. This means that `GetPixelColor()` will not return original color that `SetPixelColor()` wrote if brightness is less than 255. If you disable global buffer, WLED tries to reconstruct original color but that is not always possible without loss.

Segment does not create any buffers of its own (though in the past we had `setUpLeds()` that did that). It will utilize `strip`'s _buffering_ when `getPixelColor()` is called.

This (lossy reconstruction) is why mirroring, reversing and transposing have to be done in the `setPixelColor()` instead of separately when FX is done writing (BTW each pixel is written only once in FX if the FX function is written correctly).

Newer versions of NPB are going to do brightness scaling only during transmission of data to LEDs so `GetPixelColor()` will return originally set value. Until that happens we need to deal with it ourselves.
On top of that NPB uses the same bit depth for transmit and render buffers but a future version is scheduled to have transmit buffer of the same bit depth as required by LEDs while render buffers may be of bit depth required by user (in our case 32bits).

And then, there is a CCT support.

NOTE: Note the difference between 'GetPixelColor()` and `getPixelColor()`. One is from NBP the other from WLED.

DedeHai:
Thank you for detailing this, much clearer now.

I was just looking at 'soap' FX as an example that uses `setPixelColorXY()` as well as `getPixelColorXY()` but the following points apply to most FX. The reason I see this is quite "slow":
- colors are handled in CRGB, each `getPixelColorXY()` is converted from 32bit color, manipulated, converted back in `setPixelColorX()Y`   -> native 32bit handling would improve that i.e. ditch fastled and implement the used functions in CRGBW32, requires a lot of code refactoring though. Speed improvement is probably not going to be huge as conversion is fast (but in principle mostly unnessecary).
- the global buffer is in 'strip space mapping' so each call to set or get is mapped to and from there, including mirror, reverse, grouping etc. plus brightness adustment in `setPixelColorXY()` plus the LED map. This is what mostly makes it slow. 

A solution could be to replace the global buffer with individual segment buffers that are not mapped nor brightness adjusted but do that in one go when it is transferred to NeoPixelBus buffer (saving a lot of if statements and back and forth mapping). 
The main draw-back would be a lot more memory is used if segments are overlapping. This could be mitigated by setting a memory limit (like it is done with `SEGMENT.data`) and if that limit is exceeded, fall back to not using the buffers (this is how I do it in the ParticleSystem, btw: if I disable local buffers, PS Fire for example drops from ~85FPS to ~55FPS).

Would that be a good way to try and solve it or is there something fundamentally flawed with this approach?

blazoncek:
>     * colors are handled in CRGB, each `getPixelColorXY()` is converted from 32bit color, manipulated, converted back in `setPixelColorX()Y`   -> native 32bit handling would improve that i.e. ditch fastled and implement the used functions in CRGBW32, requires a lot of code refactoring though. Speed improvement is probably not going to be huge as conversion is fast (but in principle mostly unnessecary).

I am not going to rewrite all of the effects. 😄 

>     * the global buffer is in 'strip space mapping' so each call to set or get is mapped to and from there, including mirror, reverse, grouping etc. plus brightness adustment in `setPixelColorXY()` plus the LED map. This is what mostly makes it slow.

`strip` only does mapping via ledmaps. Swapping index if necessary. No real slowness there, at least there was none until someone wanted to exclude ledmaps while realtime data was received.
`BusDigital::getPixelColor()` uses global buffer instead of querying NeoPixelBus.

> A solution could be to replace the global buffer with individual segment buffers that are not mapped nor brightness adjusted but do that in one go when it is transferred to NeoPixelBus buffer (saving a lot of if statements and back and forth mapping). The main draw-back would be a lot more memory is used if segments are overlapping. This could be mitigated by setting a memory limit (like it is done with `SEGMENT.data`) and if that limit is exceeded, fall back to not using the buffers (this is how I do it in the ParticleSystem, btw: if I disable local buffers, PS Fire for example drops from ~85FPS to ~55FPS).

This was used in first iteration of 0.14 while we used `setUpLeds()` (MM still does). It presented a whole lot of problems which were mitigated using global buffer (which presented a new set of issues, though).

DedeHai:
> I am not going to rewrite all of the effects. 😄
no one is asking you to. I am imagining to clone most of the CRGB struct/methods so it CRGBW32 could be used in the exact same way, if done right, it could be a 1:1 replacement, needing update on all FX but it should not be too complex (mostly a direct find and replace, maybe even a #define to override it for starts).
> `strip` only does mapping via ledmaps. Swapping index if necessary. No real slowness there, at least there was none until someone wanted to exclude ledmaps while realtime data was received.  
true, lookup is fast.
>`BusDigital::getPixelColor()` uses global buffer instead of querying NeoPixelBus.
yes, the main difference being the color correction, right? If I disable global buffer, colors get worse but speed stays (almost) the same.

> This was used in first iteration of 0.14 while we used `setUpLeds()` (MM still does). It presented a whole lot of problems which were mitigated using global buffer
Do you mind to quickly elaborate what problems these were?

blazoncek:
> Do you mind to quickly elaborate what problems these were?

Overlapping segments.

wled00/FX_fcn.cpp
```diff
@@ -1181,18 +1181,21 @@ uint32_t Segment::color_wheel(uint8_t pos) const {
...
+  uint32_t color = currentColor(mcol);
   // default palette or no RGB support on segment
-  if ((palette == 0 && mcol < NUM_COLORS) || !_isRGB) return (pbri == 255) ? color : color_fade(color, pbri, true);
+  if ((palette == 0 && mcol < NUM_COLORS) || !_isRGB) {
+    color = gamma32(color);
+    return (pbri == 255) ? color : color_fade(color, pbri, true);
+  }
 
-  unsigned paletteIndex = i;
+  uint8_t paletteIndex = i;
```
blazoncek, line 1192: why?

DedeHai: it is only used as an uint8_t down the line and not involved in any calculation.

blazoncek: yes. ATM. what about in the future?
if it is unsigned, compiler will trim it down when passing to a function.

DedeHai: ok reverted to unsigned locally.

willmmiles: XTensa doesn't have non-word arithmetic instructions -- any operations done on a 8-bit or 16-bit types have to be performed by upcasting, operating, and downcasting afterwards to restrict the output range.  At least here, `int` or `unsigned` is almost always less code (and slightly faster).

DedeHai: > XTensa doesn't have non-word arithmetic instructions -- any operations done on a 8-bit or 16-bit types have to be performed by upcasting, operating, and downcasting afterwards to restrict the output range. At least here, `int` or `unsigned` is almost always less code (and slightly faster).

in general: absolutely yes. But if a function is performed purely in 8bit (like most of the fastled stuff) sometimes the compiler optimizes in a way that makes it actually faster (smaller code) when using `uint8_t`, don't ask me why exactly, I have been playing with types to improve code quite a bit and sometimes its just not really clear what is happening.

blazoncek:
> but glad it is 'only that'

It is not. wait for "blending styles"

DedeHai:
future improvements to buffers aside: should i squash this draft and make a PR? any changes required before I do that?

blazoncek:
Please, and do address points discussed.

blazoncek:
One thing to keep in mind regarding existing effects is that most were ported from FastLED type effects. Some taken directly from FastLED, some from places like soulmate.com.

FastLED operated on its CRGB (or CHSV) and so most effects disregarded the W channel. I guess this was the reason to introduce "Automatic White Calculation" for strips like SK6812 where white channel could be used to reduce power draw.

Any new effect should, if possible, take into account white channel as well IMO though most will really not benefit at all. It all comes down to the "artist" how he/she envisions an effect on RGB and RGBW strips.

I do not know if stripping W from calculations is wise or not without prior notice to users.

DedeHai:
> I do not know if stripping W from calculations is wise or not without prior notice to users.

Agreed. These are just ideas.
I don't want to get too deep in this topic but RGB already contains white, hence my question how and where it is extracted to be a separate channel. To me it makes little sense to treat is seperately in effects: yes there may be some artistic use on RGBW strips for that but it could also be done in RGB. Or how do RGB strips treat the seperated white if I would set it non zero in an effect, expecting it to turn out white?
Also: with the CRGBW struct (or is it a class?) it may be possible to do both with minimal overhead. If a strip is RGBW, the buffer gets allocated as CRGBW, if no seperate white channel: RGB. Not sure this is doable in a easy way but that would get the best of both worlds.

blazoncek:
I would avoid CRGB. Why? Because WLED uses NeoPixelBus and not FastLED (well, palettes are exception). NeoPixelBus has its own classes comparable to CRGB but also extended to include CCT (or WW & CW) information which does not exist in FastLED.
So, to avoid confusion work with `uint32_t` if possible for now.

Unfortunately the world is not as simple as that. CRGB is very common and a lot of people know what it represents. To top that some people want to control white channel independently of effects.

DedeHai:
hmm, a complex topic with no 'one fits all' solution. All I can say: if we want speedy effects, dragging (unused) white channel along takes a toll on speed.
I will take a look at NeoPixelBus and what it can do, sounds interesting.

blazoncek:
I think the biggest problem may be unnecessary conversions (a lot of them) as there is no internal standard how to work with colors.
On its own W channel is not that of a problem.

Makuna:
Some thoughts on RGB+W (including WW and even WWW) from the NPB perspective:
The white output on the chips varies by the wavelength and power.  Some are cold, some neutral, some warm, and even some that are way brighter than the RGB at full bright.  There are requests to support having a white color component (temperature) that will automatically blend through the white colors.  But this is a LED by LED specific process.  Not sure how an effect would do it consistently.  My current thinking on this detail is conversion is done right as gamma and luminance are applied.  In my library this is part of the "shader" for the NeoPixelBusLg and should be done only once.
I have not ratified this into a new color object RGBWT (RGB + White + temperature) yet as this still doesn't define what the brightness comparison is between the RGB and the W channel(s).  
The nice thing about using the uint32_t (32 bit object) is that it is the word size for ESP32, so optimized.   While having white in an effect is interesting, having an alpha channel will allow some improved blending.  Just thoughts.

blazoncek:
Thank you @Makuna for your thoughts. 

> Not sure how an effect would do it consistently.

Most likely it can't. Not in the way currently things are implemented.

> as this still doesn't define what the brightness comparison is between the RGB and the W channel(s)

I would recommend you leave initial W channel scaling to the user. Just do L and G as you do for RGB.

> having an alpha channel will allow some improved blending.

Now this is indeed a good idea. Some other folks will like it if we move `SEGMENT.opacity` into W channel and then create `Segment::setPixelColorWithAlpha()` or use something like `SEGMENT.setPixelColor(i, color_blend(c, SEGMENT.getPixelColor(i), W(c)))`. Currently `SEGMENT.opacity` just reduces brightness of all channels.
Unfortunately that will mean we'll need to come up with something if user selects manual (or dual) white management as that separates RGB and W channels logically (effect function needs to set both RGB and W information).

wled00/colors.cpp
DedeHai: If it's ok with you, I would leave it as-is for now. Integrating it into `Segment::color_from_palette()` would require a lot of code changes all over, even in usermods.

softhack007:
> This was used in first iteration of 0.14 while we used `setUpLeds()` (MM still does). It presented a whole lot of problems which were mitigated using global buffer (which presented a new set of issues, though).

@blazoncek I fully agree. We still have segment level buffers in the MM fork, but it really adds complications.
Actually, the new upstream solution with buffering on `strip` or `busses` level makes more sense. No matter where the buffer is placed, it can guarantee lossless `getPixelColorXY()`. 

From my own experiments, buffering on `segment` level is fastest, but that's due to the complicated logic when going down the chain from segment -> strip -> busses -> neoPixelBus etc. Once the licensing issues (MIT vs. GPLv3) are solved between our forks, I can start bringing some optimizations to upstream that were developed by @troyhacks and myself, to speed up pixel drawing. In MM, we can achieve 96 FPS on a 64x64 pixel fixture (HUB75).

softhack007:
> overlapping segments is a problem but mostly due to additional RAM usage, it solves other problems overlapping segments have in a global buffer, namely overlapping being FX dependent.

Well, overlapping segments has been a challenge for years. If you have segment level (instead of global) buffering, effects do not influence each other, but you can never be sure which segment will "win" the race for a physical pixel. The "overlay" option added in some effects solves flickering problems, but its still a half-hearted thing.
Sometimes - for creativity reasons - you even want a global buffer. It allows for some neat creations like "akemi of hell", where the "waverly" effect is drawn on top of akemi as a background.

I have thought about solutions, but still could not imagine (yet) how to have "real" overlays in WLED. You would need a few things like:
* segment layers, or at least a notion of priority for effects. So that the overlay effect is always painted as the last one.
* a clear definition of "background color". Because when drawing overlays, the background color must be considered as transparent.
* ....

wled00/colors.cpp
blazoncek: Yes, we can do that.

blazoncek:
A thought.

If we remove W channel from FX functions (i.e. always render all effects in RGB only) and use W channel as a last step of painting pixels (if user has "manual" or "dual" auto white calculation), that could work but may need some tweaking in `setPixelColor()` or `show()`.
In such case W part of color can be used as alpha channel and will enable effective blending of segments (or layers if you wish). That would of course mean FX function needs to be aware of this (which is not ATM) so may need a rewrite.

That is something I consider worth exploring.

P.S. I know @makuna is working on "unified" buffers where only the transmit buffers are device dependant, but render buffers are not. That change (IIRC) will also bring lossless `GetPixelColor()` which will simplify (and speed up) our `getPixelColor()`. So working with opacity/alpha may become easy & fast.

wled00/colors.cpp
```diff
@@ -33,34 +33,32 @@ uint32_t color_blend(uint32_t color1, uint32_t color2, uint16_t blend, bool b16)
...
-    if (max < 256) return RGBW32(r, g, b, w);
-    else           return RGBW32(r * 255 / max, g * 255 / max, b * 255 / max, w * 255 / max);
+  uint32_t rb = (c1 & 0x00FF00FF) + (c2 & 0x00FF00FF); // mask and add two colors at once
+  uint32_t wg = ((c1>>8) & 0x00FF00FF) + ((c2>>8) & 0x00FF00FF);
+  uint32_t r = rb >> 16; // extract single color values
+  uint32_t b = rb & 0xFFFF;
+  uint32_t w = wg >> 16;
+  uint32_t g = wg & 0xFFFF;
+
+  unsigned max = r; // check for overflow note: not checking and just topping out at 255 (formerly 'fast') is not any faster (but even slower if not overflowing)
```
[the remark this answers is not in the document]
DedeHai: that is true, went a bit too aggressive :)
will add it back.

wled00/FX_fcn.cpp
softhack007: 🤔 not sure if always_inline plays well with IRAM_ATTR .... the first tell the compiler to always inline the function, the latter says "put the function into IRAM" which means that a real function is needed.

wled00/colors.cpp
```diff
@@ -453,23 +486,7 @@ uint16_t approximateKelvinFromRGB(uint32_t rgb) {
...
-   25, 26, 27, 27, 28, 29, 29, 30, 31, 32, 32, 33, 34, 35, 35, 36,
-   37, 38, 39, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 50,
-   51, 52, 54, 55, 56, 57, 58, 59, 60, 61, 62, 63, 64, 66, 67, 68,
-   69, 70, 72, 73, 74, 75, 77, 78, 79, 81, 82, 83, 85, 86, 87, 89,
-   90, 92, 93, 95, 96, 98, 99,101,102,104,105,107,109,110,112,114,
-  115,117,119,120,122,124,126,127,129,131,133,135,137,138,140,142,
-  144,146,148,150,152,154,156,158,160,162,164,167,169,171,173,175,
-  177,180,182,184,186,189,191,193,196,198,200,203,205,208,210,213,
-  215,218,220,223,225,228,231,233,236,239,241,244,247,249,252,255 };
+uint8_t NeoGammaWLEDMethod::gammaT[256];
```
softhack007, line 489: What's the reason to keep this table uninitialized?

wled00/FX_fcn.cpp
DedeHai: that is what I was wondering too, this is just a suggestion, i.e. to inline this for speed but how to tell the compiler to inline it to the functions that are in ram... not sure how it will do it.

wled00/colors.cpp
DedeHai: it is filled at boot.

wled00/FX_fcn.cpp
```diff
@@ -1184,18 +1188,21 @@ uint32_t Segment::color_wheel(uint8_t pos) const {
...
+  }
 
   unsigned paletteIndex = i;
   if (mapping && virtualLength() > 1) paletteIndex = (i*255)/(virtualLength() -1);
   // paletteBlend: 0 - wrap when moving, 1 - always wrap, 2 - never wrap, 3 - none (undefined)
   if (!wrap && strip.paletteBlend != 3) paletteIndex = scale8(paletteIndex, 240); //cut off blend at palette "end"
   CRGB fastled_col = ColorFromPalette(_currentPalette, paletteIndex, pbri, (strip.paletteBlend == 3)? NOBLEND:LINEARBLEND); // NOTE: paletteBlend should be global
 
-  return RGBW32(fastled_col.r, fastled_col.g, fastled_col.b, W(color));
+  return RGBW32(fastled_col.r, fastled_col.g, fastled_col.b, gamma8(W(color)));
```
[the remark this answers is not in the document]
DedeHai: it is as it was before, compare to full function before the change. I just removed applying gamme to all of the `color` channels if that color is not used.

wled00/FX_fcn.cpp
```diff
@@ -712,11 +712,16 @@ void IRAM_ATTR_YN Segment::setPixelColor(int i, uint32_t col)
 {
   if (!isActive()) return; // not active
 #ifndef WLED_DISABLE_2D
-  int vStrip = i>>16; // hack to allow running on virtual strips (2D segment columns/rows)
+  int vStrip;
```
softhack007, line 715: I'm not a big fan of leaving locals uninitialized... are you sure that vStrip is never used without first assigning a value?

wled00/FX_2Dfcn.cpp
```diff
@@ -378,14 +375,14 @@ void Segment::blur2D(uint8_t blur_amount, bool smear) {
       uint32_t part = color_fade(cur, seep);
       curnew = color_fade(cur, keep);
       if (y > 0) {
-        if (carryover) curnew = color_add(curnew, carryover, true);
-        uint32_t prev = color_add(lastnew, part, true);      
+        if (carryover) curnew = color_add(curnew, carryover);
+        uint32_t prev = color_add(lastnew, part);
```
softhack007, line 379: Maybe I overlooked something in the previous discussions - I guess that setting the last parameter to true was done on purpose ... are you sure it does the same thing?

wled00/FX_fcn.cpp
DedeHai: good call.

wled00/FX_fcn.cpp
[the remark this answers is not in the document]
softhack007: 👍

wled00/colors.cpp
softhack007: Ok, makes sense.

wled00/FX_2Dfcn.cpp
DedeHai: this change is to be approved. It does not do 100% the same thing but it does it faster. I tested this with fire2012 at full blur and the 'overflow' happens only once (i.e. for one pixel) every few frames. There may be situations where this change actually is visible but I could not find one (except the ones I updated in the other PR, smearing looks better without saturation)

wled00/colors.cpp
willmmiles: Does this array need to be changed at runtime?  Could it be filled with a `constexpr` function and save some RAM?

DedeHai: yes, this is the gamma correction, can be changed by user.

wled00/FX_fcn.cpp
softhack007: To my understanding:
* `inline` is a hint/suggestion to the compiler. So it might get inlined, or not.
* `__attribute__((always_inline))` is a directive. So the compiler _must_ inline this function, no matter if its efficient or not.

If you want to optimize function calls, its sometime useful to add `__attribute__((pure))` or `__attribute__((const))` to the function declaration. But only do this after double-checking that the code is actually "pure" (no side-effects) or "const" (solely depends on arguments). I did this in the MoonModules fork, but honestly it does not give you more than 1 or 2 fps even if you apply it to lots of functions.

https://github.com/MoonModules/WLED/commit/7f9da309c9d37a27226b0a2e247e47494fafa50b

DedeHai:
enabled now.

wled00/FX_2Dfcn.cpp
```diff
@@ -461,37 +461,37 @@ void Segment::box_blur(unsigned radius, bool smear) {
 
 void Segment::moveX(int8_t delta, bool wrap) {
```
DedeHai, line 462: any reason delta cann not be an `int`?

wled00/FX_2Dfcn.cpp
```diff
@@ -461,37 +461,37 @@ void Segment::box_blur(unsigned radius, bool smear) {
...
-      for (int x = -delta-1; x >= 0; x--)    newPxCol[x] = getPixelColorXY(wrap ? (x + delta) + cols : x, y);
+      for (int x = vW-1; x >= -delta; x--) newPxCol[x] = getPixelColorXY((x + delta), y);
+      for (int x = -delta-1; x >= 0; x--)  newPxCol[x] = getPixelColorXY(wrap ? (x + delta) + vW : x, y);
     }
-    for (int x = 0; x < cols; x++) setPixelColorXY(x, y, newPxCol[x]);
+    for (int x = 0; x < vW; x++) setPixelColorXY(x, y, newPxCol[x]);
   }
 }
 
 void Segment::moveY(int8_t delta, bool wrap) {
```
DedeHai, line 480: dito

wled00/FX_2Dfcn.cpp
blazoncek: It could.

DedeHai:
Latest changes from @blazoncek have a huge positive impact on FPS however the addition of `bool unScaled` has no measurable benefit and adds to code size. Here are my test results:
Vanilla 0_15:
52FPS
RAM:   [==        ]  15.3% (used 50020 bytes from 327680 bytes)
Flash: [========= ]  89.7% (used 1410392 bytes from 1572864 bytes)

My changes including https://github.com/Aircoookie/WLED/pull/4138/commits/202901b09f2a8c6da3a453b0d3ed8c8140ea5557:
54FPS
RAM:   [==        ]  15.3% (used 50020 bytes from 327680 bytes)
Flash: [========= ]  89.5% (used 1407852 bytes from 1572864 bytes) -> saves 2540bytes 

Latest addition from blazoncek https://github.com/Aircoookie/WLED/pull/4138/commits/9114867578477e4cf5fa652769ea0690503c5f10: 
**65FPS (!)**
RAM:   [==        ]  15.3% (used 50036 bytes from 327680 bytes)
Flash: [========= ]  89.6% (used 1409002 bytes from 1572864 bytes) ->  adds 1150bytes

I have a commit ready removing the `bool unScaled` again if there are no objections (also including some fixes and improvements)

softhack007:
> I have a commit ready removing the `bool unScaled` again if there are no objections (also including some fixes and improvements)

@DedeHai the "unscaled" optimization only affects Segment::drawCircle and Segment::drawLine with anti-aliased  enabled. Did you test with an effect that uses these functions?

DedeHai:
Yes, I tested it using "Ripple" and "Blobs" FX (4 overlapping segments to intensify the calculation) and saw no FPS change in either case.

softhack007:
> Yes, I tested it using "Ripple" and "Blobs" FX (4 overlapping segments to intensify the calculation) and saw no FPS change in either case.

Strange ... actually "unscaled=false" removes `color_fade(c, currentBri())` from the inner loop, this should have a noticeable effect. Can you test again with a bigger matrix setup? I would expect that some speedup get visible. 

Also keep in mind that our "fps" calc is quite inaccurate, like +- 2 fps of "uncertainty" and almost meaningless above 80fps.

DedeHai:
I tested various versions, with scaling and without scaling (completely commented out) and saw no real impact. Maybe my optimized scaling function is now so fast that it takes a lot of pixels to actually make a difference (I am sure it will on 1000+).
I just discussed with blazoncek to commit my changes. A different approach would be to use a static variable instead of passing a parameter (which is the actual problem eating up flash on every call throughout the code).

softhack007:
> A different approach would be to use a static variable instead of passing a parameter (which is the actual problem eating up flash on every call throughout the code).

just saw the commit, looks ok for me, too 👍- it achieves the same thing, just making "_colorScaled" a private variable (aka sideeffect) instead of using an explicit function parameter.

wled00/FX_fcn.cpp
```diff
@@ -437,7 +440,21 @@ uint32_t IRAM_ATTR_YN Segment::currentColor(uint8_t slot) const {
 #endif
 }
 
-void Segment::setCurrentPalette() {
+// pre-calculate drawing parameters for faster access
```
softhack007, line 445: @blazoncek I have the feeling you took an idea that I implemented previously in the MM fork - pre-calculate drawing parameters for faster access. It is indeed a big speedup (my test in MM were up to 10% faster).

I don't want to go pricky about this, but it would be nice if you state where you took inspiration from. Something like "based on an idea from @softhack007 in the MM fork". Please.

softhack007: @blazoncek thanks

blazoncek: > I don't want to go pricky about this, but it would be nice if you state where you took inspiration from.

None of this would be needed if you'd provide similar optimisations upstream or not adopt GPL in MM. So I had to come up with something similar. You have the liberty (due to MIT) to pick anything from upstream while I need to beg people or figure out something.
Just my 2¢.

softhack007: GPLv3 is not the problem. The problem is MIT licensing in the AC repo, which has become a real no-go for me (and a few others) after some very frustrating experiences. I an still the autor of the code, so I may "provide" it to a repo licensed with MIT, but you're not allowed to "take" without explicit permission. I know this is not optimal and we must find a better approach soon.


@Aircoookie You have proposed a while ago to harmonize the licensing between both repo's, to remove this "rift", by using EUPL-1-2. I'm not sure if there was further discussion on discord. 
The MM team is willing to go this way. If you also agree to change to EUPL-1.2, we (MM team) will start collecting the necessary permissions from contributor since our change to GPLv3. Please give me a sign :-)

blazoncek:
IMO this is ready for broader testing.
@netmindz suggested to wait with merging until 0.15.0 is released so IMO it may be good to concentrate on releasing 0.15.0 ASAP ant then release 0.15.1 in "short" period of time.

wled00/FX_fcn.cpp
blazoncek: > The problem is MIT licensing in the AC repo, which has become a real no-go for me (and a few others) _**after some very frustrating experiences**_.

NOTE: Emphasis mine.

Can you explain? Can others mentioned explain as well? I've added about 48k lines of code to WLED (not counting 0_15 branch) but have yet to encounter "frustrating experience". What does "frustrating experience" even mean?
Is that a stolen or "borrowed" code? It inevitably happens with GPL as well.

I am not against license change, I just want to know/understand true reasons for it. Is MIT too permissive? Everyone knew/knows WLED uses MIT so there should be no "frustrating experiences" as a consequence of that IMO. I would really love to hear everyone's perspective, not for WLED project but my personal insights.

netmindz: I can understand you wanting to know more, but I'm not sure that a public discussion as a side topic on a PR is the appropriate place. 

Can we arrange a call/meeting with relevant parties?

blazoncek:
I have some additional modifications that are not strictly speed related but do decrease binary size for another kB or two.
Do I push them into this PR?

DedeHai:
as long as we keep the commit history, this PR could be a collection of code improvements. 
Initially we said to make this a squashed commit, but with all the changes that could become a nightmare to debug as by now most core (rendering) functions were touched up.

wled00/FX_2Dfcn.cpp
```diff
@@ -168,50 +168,63 @@ uint16_t IRAM_ATTR_YN Segment::XY(int x, int y)
...
-      // if blending modes, blend with underlying pixel
-      if (_modeBlend) col = color_blend(strip.getPixelColorXY(start + xX, startY + yY), col, 0xFFFFU - progress(), true);
-#endif
-      strip.setPixelColorXY(start + xX, startY + yY, col);
-      if (mirror) { //set the corresponding horizontally mirrored pixel
-        if (transpose) strip.setPixelColorXY(start + xX, startY + height() - yY - 1, col);
-        else           strip.setPixelColorXY(start + width() - xX - 1, startY + yY, col);
+  unsigned groupLen = groupLength();
+
+  if(groupLen > 1)
```
blazoncek, line 208: This reduces the processing if group length is 1 by removing:
- four assignments (`yY` && `xX`, `W` & `H`)
  - hidden two subtractions and a comparison (`W` & `H`)
- two multiplications
- two comparisons (`j` & `g`, `uint8_t`!)
- two comparisons (`yY>=H` && `xX>=W`)
- two increments (`xX` & `yY`)

For each `setPixelColorXY()` call.

What may seem irrelevant with low number of pixels actually produces a noticeable slowdown on large pixel count.

wled00/FX_2Dfcn.cpp
```diff
@@ -273,119 +286,62 @@ uint32_t IRAM_ATTR_YN Segment::getPixelColorXY(int x, int y) const {
...
+        last = cur; //save original value for comparison on next iteration
+        carryover = part;
+      }
+      setPixelColorXY(col, rows - 1, curnew);
     }
-    setPixelColorXY(col, rows - 1, curnew);
   }
 }
 
+/*
```
blazoncek, line 344: I know it is unused ATM but it produces quite different blur and I put this function in intentionally.
I would not like it removed.

wled00/FX_2Dfcn.cpp
```diff
@@ -458,43 +414,69 @@ void Segment::box_blur(unsigned radius, bool smear) {
...
 void Segment::moveX(int delta, bool wrap) {
-  if (!isActive()) return; // not active
+  if (!isActive() || !delta) return; // not active
   const int vW = vWidth();   // segment width in logical pixels (can be 0 if segment is inactive)
   const int vH = vHeight();  // segment height in logical pixels (is always >= 1)
-  if (!delta || abs(delta) >= vW) return;
+  int absDelta = abs(delta);
+  if (absDelta >= vW) return;
   uint32_t newPxCol[vW];
+  int newDelta;
```
blazoncek, line 425: I do not see the point in modifying `move()` functions, especially with convoluted math that is hard to follow.
How much speed does it gain?

blazoncek (commented):
Just a few thoughts.

wled00/FX_2Dfcn.cpp
```diff
@@ -676,7 +652,10 @@ void Segment::drawCharacter(unsigned char chr, int16_t x, int16_t y, uint8_t w,
       case 60: bits = pgm_read_byte_near(&console_font_5x12[(chr * h) + i]); break; // 5x12 font
       default: return;
     }
-    col = ColorFromPalette(grad, (i+1)*255/h, 255, NOBLEND);
+    uint32_t col = ColorFromPaletteWLED(grad, (i+1)*255/h, 255, NOBLEND);
```
blazoncek, line 655: This is a "shadow" of the `col` from a few linea above.

blazoncek:
FYI I am now running these improvements on my test set up (8266,32,S2,C3) for about a week and see no pitfalls.

Code is smaller and faster.

blazoncek:
As I am moving away from WLED (at least for a while if not permanently) I would urge you to implement this PR ASAP as it benefits WLED tremendously.

DedeHai:
I can rebase and merge any time. It would require another beta release as there are many changes in this PR that are not well tested. If we do that https://github.com/Aircoookie/WLED/pull/4181 should go along with it though.

wled00/FX_fcn.cpp
```diff
@@ -489,20 +489,30 @@ void Segment::handleRandomPalette() {
...
       && (!grp || (grouping == grp && spacing == spc))
-      && (ofs == UINT16_MAX || ofs == offset)) return;
+      && (ofs == UINT16_MAX || ofs == offset)
+      && (m12 == map1D2D)
+     ) return;
 
   stateChanged = true; // send UDP/WS broadcast
 
-  if (stop) fill(BLACK); // turn old segment range off (clears pixels if changing spacing)
+  if (stop || spc != spacing || m12 != map1D2D) {
```
softhack007, line 495: Does "if (stop)" do the same as `isActive()`, or is there a special reason to avoid isActive() here?

wled00/FX_fcn.cpp
```diff
@@ -489,20 +489,30 @@ void Segment::handleRandomPalette() {
...
+    _vWidth  = virtualWidth();
+    _vHeight = virtualHeight();
+    _vLength = virtualLength();
+    _segBri  = currentBri();
+    fill(BLACK); // turn old segment range off or clears pixels if changing spacing (requires _vWidth/_vHeight/_vLength/_segBri)
```
softhack007, line 500: Maybe use `beginDraw()`, instead of updating all cached values explicitly? It would make the code easier to read.

wled00/FX_fcn.cpp
blazoncek: IDK, it is an old code. Perhaps if you shorten a segment. Most likely on second thought.

wled00/FX_fcn.cpp
blazoncek: `beginDraw()` does a lot of other things too which are unnecessary.
I did consider calling it, but decided against and added comment.

DedeHai:
FYI: I just ran a final test before merging this. Overall speed improvement is around 20%!

PR #4142, opened 2024-09-14
Fix AsyncWebServer version pin

willmmiles:
We want to always use the tagged version, not complain if the current main is a mismatch.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -140,7 +140,7 @@ lib_deps =
     IRremoteESP8266 @ 2.8.2
     makuna/NeoPixelBus @ 2.8.0
     #https://github.com/makuna/NeoPixelBus.git#CoreShaderBeta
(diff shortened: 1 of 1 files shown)


blazoncek:
Does not work on my system.
```
warning: Could not find remote branch 2.2.1 to clone.
fatal: Remote branch 2.2.1 not found in upstream origin
```

blazoncek:
I guess we'll need to go with 2.3.0 in main.

willmmiles:
I'm not sure h

> Does not work on my system.
> 
> ```
> warning: Could not find remote branch 2.2.1 to clone.
> fatal: Remote branch 2.2.1 not found in upstream origin
> ```

That's definitely not right - this PR sets the target to the version tag ("v2.2.1"), not "2.2.1".  The CI seems to build correctly.

PR #4145, opened 2024-09-15
FX improvements and cleanup

DedeHai:
This has been on my todo list for a long time, finally done.
It adds some new functionality to 2D FX (blurring / smearing / palette support) and saves on code size (merged 1D FX)
Note: Polar Lights is now slightly different as no palette matches the original effect palette

* Waving Cell: Improved with higher temporal resolution (smoother at lower speeds) and added additional mode setting and optional blurring
* Merged puddles and puddlepeak
* Merged Gravcenter, Gravcentric, Gravfreq and Gravimeter (saves 1.2k of flash)
* Merged meteor and meteor smooth
* Renamed police_base into mode_two_dots as that was just an alias
* Changed default of Palette effect: Animate rotation is now unchecked by default (it looks weird and confusing in 1D as a default)

* Added or enhanced blurring of
    * DNA
    * DNA Spiral
    * Drift
    * Drift Rose
    * Crazy Bees
    * Ripple
    * Colored Bursts
    * Frizzles
    * Lissajous
    * Sindots
    * Spaceships

* Added palette support to
    * Crazy Bees
    * Polar Lights
    * Drift Rose

* Some code cleanup (removed unused / commented stuff)
* Moved dev info for AR to the top so ist easier to find as a reference, also added link to KB there


blazoncek:
@ewoudwijma @netmindz would you care to take a look as some effects were created by you.

> * Changed default of Palette effect: Animate rotation is now unchecked by default (it looks weird and confusing in 1D as a default)

This will need to be checked with @triplewhy

wled00/FX.cpp
```diff
@@ -2042,7 +2071,7 @@ uint16_t mode_palette() {
   }
   return FRAMETIME;
 }
-static const char _data_FX_MODE_PALETTE[] PROGMEM = "Palette@Shift,Size,Rotation,,,Animate Shift,Animate Rotation,Anamorphic;;!;12;c1=128,c2=128,c3=128,o1=1,o2=1,o3=0";
```
blazoncek, line 2045: The rotation was put there intentionally by last modifier (@TripleWhy) and I agreed to that.
I do understand that it makes less sense for a single, undivided strip but I would like to avoid yet another breaking change.

wled00/FX.cpp
```diff
@@ -4954,11 +4957,11 @@ uint16_t mode_2DColoredBursts() {              // By: ldirko   https://editor.so
       SEGMENT.setPixelColorXY(y1, y2, DARKSLATEGRAY);
     }
   }
-  if (SEGMENT.custom3) SEGMENT.blur(SEGMENT.custom3/2);
+  SEGMENT.blur(SEGMENT.custom3<<1, SEGMENT.check2);
```
blazoncek, line 4960: Contrary to the original this now doubles blurring instead of halving it.
This will cause *different looking* effect on existing presets. Please consider backwards compatibility.

wled00/FX.cpp
```diff
@@ -5021,10 +5024,11 @@ uint16_t mode_2DDNASpiral() {               // By: ldirko  https://editor.soulma
       SEGMENT.setPixelColorXY(x1, i, WHITE);
     }
   }
+  SEGMENT.blur(SEGMENT.custom1 >> (!SEGMENT.check1), SEGMENT.check1);
```
blazoncek, line 5027: This will cause *different looking* effect on existing presets. Please consider backwards compatibility.

wled00/FX.cpp
```diff
@@ -5048,13 +5052,13 @@ uint16_t mode_2DDrift() {              // By: Stepko   https://editor.soulmateli
     int mySin = sin_t(angle) * i;
     int myCos = cos_t(angle) * i;
     SEGMENT.setPixelColorXY(colsCenter + mySin, rowsCenter + myCos, ColorFromPalette(SEGPALETTE, (i * 20) + t_20, 255, LINEARBLEND));
-    if (SEGMENT.check1) SEGMENT.setPixelColorXY(colsCenter + myCos, rowsCenter + mySin, ColorFromPalette(SEGPALETTE, (i * 20) + t_20, 255, LINEARBLEND));
+    if (SEGMENT.check2) SEGMENT.setPixelColorXY(colsCenter + myCos, rowsCenter + mySin, ColorFromPalette(SEGPALETTE, (i * 20) + t_20, 255, LINEARBLEND));
```
blazoncek, line 5055: This will cause *different looking* effect on existing presets. Please consider backwards compatibility.

wled00/FX.cpp
```diff
@@ -5100,17 +5104,17 @@ uint16_t mode_2DFrizzles(void) {                 // By: Stepko https://editor.so
   const int cols = SEGMENT.virtualWidth();
   const int rows = SEGMENT.virtualHeight();
 
-  SEGMENT.fadeToBlackBy(16);
+  SEGMENT.fadeToBlackBy(18);
```
blazoncek, line 5107: This will cause *different looking* effectand may case playlists to no longer function as intended. Please consider backwards compatibility.

wled00/FX.cpp
```diff
@@ -5582,16 +5587,14 @@ static const char _data_FX_MODE_2DPLASMABALL[] PROGMEM = "Plasma Ball@Speed,,Fad
...
-uint16_t mode_2DPolarLights(void) {        // By: Kostyantyn Matviyevskyy  https://editor.soulmatelights.com/gallery/762-polar-lights , Modified by: Andrew Tuline
+
+uint16_t mode_2DPolarLights(void) {        // By: Kostyantyn Matviyevskyy  https://editor.soulmatelights.com/gallery/762-polar-lights , Modified by: Andrew Tuline & @dedehai (palette support)
   if (!strip.isMatrix || !SEGMENT.is2D()) return mode_static(); // not a 2D set-up
 
   const int cols = SEGMENT.virtualWidth();
   const int rows = SEGMENT.virtualHeight();
 
-  CRGBPalette16 auroraPalette  = {0x000000, 0x003300, 0x006600, 0x009900, 0x00cc00, 0x00ff00, 0x33ff00, 0x66ff00, 0x99ff00, 0xccff00, 0xffff00, 0xffcc00, 0xff9900, 0xff6600, 0xff3300, 0xff0000};
+ // CRGBPalette16 auroraPalette  = {0x000000, 0x003300, 0x006600, 0x009900, 0x00cc00, 0x00ff00, 0x33ff00, 0x66ff00, 0x99ff00, 0xccff00, 0xffff00, 0xffcc00, 0xff9900, 0xff6600, 0xff3300, 0xff0000};
```
blazoncek, line 5597: This will cause *different looking* effect. Please consider backwards compatibility.

You should keep this if the `SEGMENT.palette == 0`, a default value for palette.

wled00/FX.cpp
```diff
@@ -5890,14 +5878,14 @@ uint16_t mode_2Dcrazybees(void) {
...
-    SEGMENT.fadeToBlackBy(32);
-
+    SEGMENT.fadeToBlackBy(32 + ((SEGMENT.check1*SEGMENT.intensity) >> 2));
+    SEGMENT.blur(SEGMENT.intensity >> 1, SEGMENT.check1);
     for (size_t i = 0; i < n; i++) {
-      SEGMENT.addPixelColorXY(bee[i].aimX + 1, bee[i].aimY, CHSV(bee[i].hue, 255, 255));
-      SEGMENT.addPixelColorXY(bee[i].aimX, bee[i].aimY + 1, CHSV(bee[i].hue, 255, 255));
-      SEGMENT.addPixelColorXY(bee[i].aimX - 1, bee[i].aimY, CHSV(bee[i].hue, 255, 255));
-      SEGMENT.addPixelColorXY(bee[i].aimX, bee[i].aimY - 1, CHSV(bee[i].hue, 255, 255));
+      CRGB flowerCcolor = ColorFromPalette(SEGPALETTE,bee[i].hue);
```
blazoncek, line 5884: Why not use `SEGMENT.color_from_palette()` to keep using WLED native pixels and behaviour?

wled00/FX.cpp
```diff
@@ -6229,13 +6217,13 @@ uint16_t mode_2Ddriftrose(void) {
     float angle = radians(i * 10);
     uint32_t x = (CX + (sin_t(angle) * (beatsin8(i, 0, L*2)-L))) * 255.f;
     uint32_t y = (CY + (cos_t(angle) * (beatsin8(i, 0, L*2)-L))) * 255.f;
-    SEGMENT.wu_pixel(x, y, CHSV(i * 10, 255, 255));
+    SEGMENT.wu_pixel(x, y, ColorFromPalette(SEGPALETTE, i * 10));
```
blazoncek, line 6220: Use palette only if `SEGMENT.palette != 0` otherwise use existing behaviour.

blazoncek (changes requested):
I *do welcome* the changes and see great value in optimisations, but...

Please avoid breaking existing default behaviour of effects unless strictly necessary. It is not all about speed when you have 10.000+ users.
There are also some trailing spaces and formatting glitches to be fixed.

wled00/FX.cpp
blazoncek: Regarding `ColorFromPalette()` and `SEGMENT.color_from_palette()`.

The former was used in FastLED effects ported to WLED with minimal effort (or to keep changes simple). Unfortunately this does not adhere to WLED's "default" behaviour for palettes and color, unless specifically crafted within effect. So I would prefer to use `SEGMENT.color_from_palette()` wherever possible if there are changes to an effect function.

DedeHai:
I did not consider backward compatibility to be so important but you are right, will revert default behaviour back to what it was before.

blazoncek:
> should meteor_smooth be kept as an alias or is removing it ok

As we removed some effects in 0.13 and then again in 0.14 I think we can do that again in favour of the possibility to finally merge in PS effects. I would really like to see those (at least 2D) in.

wled00/FX.cpp
```diff
@@ -7913,7 +7779,7 @@ void WS2812FX::setupEffectData() {
   addEffect(FX_MODE_COLORTWINKLE, &mode_colortwinkle, _data_FX_MODE_COLORTWINKLE);
   addEffect(FX_MODE_LAKE, &mode_lake, _data_FX_MODE_LAKE);
   addEffect(FX_MODE_METEOR, &mode_meteor, _data_FX_MODE_METEOR);
-  addEffect(FX_MODE_METEOR_SMOOTH, &mode_meteor_smooth, _data_FX_MODE_METEOR_SMOOTH);
+  //addEffect(FX_MODE_METEOR_SMOOTH, &mode_meteor_smooth, _data_FX_MODE_METEOR_SMOOTH); // merged with mode_meteor
```
TripleWhy, line 7782: remove?

wled00/FX.cpp
```diff
@@ -1375,15 +1419,17 @@ uint16_t mode_loading(void) {
...
+ if (SEGLEN == 1) return mode_static();
   unsigned delay = 1 + (FRAMETIME<<3) / SEGLEN;  // longer segments should change faster
   uint32_t it = strip.now / map(SEGMENT.speed, 0, 255, delay<<4, delay);
   unsigned offset = it % SEGLEN;
-
   unsigned width = ((SEGLEN*(SEGMENT.intensity+1))>>9); //max width is half the strip
   if (!width) width = 1;
+  if (!SEGMENT.check2) SEGMENT.fill(SEGCOLOR(2));
+  uint32_t color1 = SEGCOLOR(0);
+  uint32_t color2 = (SEGCOLOR(1) == SEGCOLOR(2)) ? color1 : SEGCOLOR(1);
```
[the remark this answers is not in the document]
blazoncek: Legacy.
If colors 2 & 3 match, use color 1 for both "dots". Color 3 is background. Color 3 (AKA background) can be omitted if "Overlay" is selected.
Unfortunately a poor choice of variable names if that is what you are referring to. 😁

wled00/FX.cpp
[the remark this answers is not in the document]
blazoncek: My recommendation is to use this palette (as is) if user has selected "Default" palette (with ID==0).
Check can be made as `if (SEGMENT.palette == 0) auroraPalette = CRGBPalette16({...}); else SEGMENT.loadPalette();` or some similar construct.

wled00/FX.cpp
blazoncek: There are plenty of complaints I receive that say, in one way or the other, this effect changed when users click on UI elements. This is one of those situations.

But yes, you are correct, 0_15 is not yet released so anything can happen. I was just worried that you might get frustrated when your change was modified. If you are not, I am ok with the change as I normally don't use rotation on 1D strips.

DedeHai: I am ok with either, it was just that I have seen user reports about this FX being 'broken' on 1D as they did not notice the rotation button.

wled00/FX.cpp
DedeHai: I fixed this FX to be useable with any palette, there is no longer the requirement of 'black edges' as it was before. But I agree that adding the palette to the list of palettes makes more sense. I will see what I can come up with.

wled00/FX.cpp
DedeHai: generally agreed. 
`color_from_palette` does some additional checks and calculations, plus converts to 32bits, then FX convert it back to CRGB, then setPixelColor converts it again to 32bit. That makes FX slow that call this for every pixel so I don't see it as a good solution. I have some basic ideas on how to fix this with CRGBW (basically overload some functions to accept RGB and RGBW and calculate the fastest way possible)

wled00/FX.cpp
```diff
@@ -24,12 +24,56 @@
...
+#define MAX_FREQ_LOG10  4.04238f // log10(MAX_FREQUENCY)
+
+// for 20Khz sampling
+//#define MAX_FREQUENCY   10240
+//#define MAX_FREQ_LOG10  4.0103f
+
+// for 10Khz sampling
+//#define MAX_FREQUENCY   5120
+//#define MAX_FREQ_LOG10  3.71f
+
```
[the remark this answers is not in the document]
DedeHai: not familiar with constexpr (plus I like the more nostalgic C style ;) )

wled00/FX.cpp
```diff
@@ -1375,15 +1419,17 @@ uint16_t mode_loading(void) {
 static const char _data_FX_MODE_LOADING[] PROGMEM = "Loading@!,Fade;!,!;!;;ix=16";
 
 
-//American Police Light with all LEDs Red and Blue
-uint16_t police_base(uint32_t color1, uint32_t color2) {
-  if (SEGLEN == 1) return mode_static();
+//Police Lights with custom colors
```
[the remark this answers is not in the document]
DedeHai: agreed, not removing that line was an oversight, will adjust.

wled00/FX.cpp
```diff
@@ -1375,15 +1419,17 @@ uint16_t mode_loading(void) {
...
+uint16_t mode_two_dots() {
+ if (SEGLEN == 1) return mode_static();
   unsigned delay = 1 + (FRAMETIME<<3) / SEGLEN;  // longer segments should change faster
   uint32_t it = strip.now / map(SEGMENT.speed, 0, 255, delay<<4, delay);
   unsigned offset = it % SEGLEN;
-
   unsigned width = ((SEGLEN*(SEGMENT.intensity+1))>>9); //max width is half the strip
   if (!width) width = 1;
+  if (!SEGMENT.check2) SEGMENT.fill(SEGCOLOR(2));
+  uint32_t color1 = SEGCOLOR(0);
```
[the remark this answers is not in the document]
DedeHai: for me it's less about caring and more a lack of knowledge when and how to use it properly (plus in short, contained functions: does it matter apart from readability?)

wled00/FX.cpp
```diff
@@ -2360,85 +2389,59 @@ uint16_t mode_meteor() {
...
-// send a meteor from begining to to the end of the strip with a trail that randomly decays.
-// adapted from https://www.tweaking4all.com/hardware/arduino/adruino-led-strip-effects/#LEDStripEffectMeteorRain
-uint16_t mode_meteor_smooth() {
-  if (SEGLEN == 1) return mode_static();
-  if (!SEGENV.allocateData(SEGLEN)) return mode_static(); //allocation failed
-
-  byte* trail = SEGENV.data;
-
   const unsigned meteorSize = 1+ SEGLEN / 20; // 5%
   uint16_t in = map((SEGENV.step >> 6 & 0xFF), 0, 255, 0, SEGLEN -1);
```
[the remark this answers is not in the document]
DedeHai: I will check. I put the originals side-by-side and copied the lines over, then merged. Could be an error.

wled00/FX.cpp
```diff
@@ -2360,85 +2389,59 @@ uint16_t mode_meteor() {
...
-      index -= SEGLEN;
+    unsigned index = (in + j) % SEGLEN;      
+    if(SEGMENT.check3) { //smooth
+        trail[index] = max;
+        uint32_t col = SEGMENT.check1 ? SEGMENT.color_from_palette(index, true, false, 0, trail[index]) : SEGMENT.color_from_palette(trail[index], false, true, 255);
+        SEGMENT.setPixelColor(index, col);
+    }
+    else{
+      int idx = 255;
+      int i = trail[index] = max;
```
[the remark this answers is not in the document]
DedeHai: I could not make heads or tails of some of the code, but not my intention to improve it, just wanted to get rid of all the double-code to save flash space.

wled00/FX.cpp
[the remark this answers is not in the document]
DedeHai: that was unintentional, should be a right shift. something that slipped my testing. I did test a few versions on each of the FX I updated, what you see is just the final outcome (I think the version beofre this mistake was something like `custom3/(check2<<1)`

wled00/FX.cpp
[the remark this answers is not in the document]
DedeHai: artistic freedom ;)

DedeHai: in FX where I added blur, the default slider position is set to 0, checks are set to 0.
that will change behaviour on users that have 'default fx parameters' unused. if that is a restriction, none of the FX can be extended (ever).

wled00/FX.cpp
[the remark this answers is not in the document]
TripleWhy: `const` is easy: just stick it to every variable, then remove it again if it doesn't compile, and the reason for that is that you actually mean to modify it multiple times :)

Does it matter? The answer to that is really the same for most programming good practices:

* It *probably* doesn't matter for performacne.
* It may or may not increase readability, depending on who you ask:
  * It does state intent, and that does simplify a readers process to understand the code.
  * It gives the reader some more information that he doesn't have to look up, reducing the mental load.
  * It clutters the code, and too much noise can hinder fast reading.
* It however protects you and future maintainers from unintentional actions. In the case of `const` from modifying a variable that is supposed to be constant. You can obviously remove the `const` later if necessary, but you force that decision to be made consciously.

The length of the context is a consideration that I have only ever heard about variable naming before. I think the thing about good practices is that you have to make them a habit, and therefore the scope doesn't matter. If you do it everywhere, it also improves consistency, which is always good.

wled00/FX.cpp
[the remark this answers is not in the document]
TripleWhy: constexpr should result in the same binary code, but has all kinds of guarantees that define doesn't provide. As long as you always know what you are doing, never make mistakes, and the same applies to every future maintainer, defines are fine :)

wled00/FX.cpp
```diff
@@ -6497,7 +6445,10 @@ typedef struct Gravity {
 ///////////////////////
 //   * GRAVCENTER    //
 ///////////////////////
-uint16_t mode_gravcenter(void) {                // Gravcenter. By Andrew Tuline.
+// Gravcenter effects By Andrew Tuline.
+// Gravcenter base function for Gravcenter (0), Gravcentric (1), Gravimeter (2), Gravfreq (3) (merged by @dedehai)
+
+uint16_t mode_gravcenter_base(unsigned mode) {
```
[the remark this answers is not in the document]
DedeHai: well, merging them like this saves 1.2k of flash so actually it is quite some shared code. 
while I agree this could be made prettier I don't think seperating it in a different way would improve the readability.
the runtime checks are negligible, that is only done once per frame.

wled00/FX.cpp
DedeHai: I would leave it there for now in case people look for it in the source code.

wled00/FX.cpp
blazoncek: > then FX _may_ convert it back to CRGB,

This (and other manipulations like that) will need to be removed or converted. `color_from_palette()` is there for a reason, to ensure consistent behaviour of "Default" options.

wled00/FX.cpp
blazoncek: > I disagree, if this palette is so revolutionary, it should be available to every effect, not just this one.

Do not forget: Most effects were ported from FastLED effects. Not everyone wanted (or knew how) to make them WLED native. Also, there wasn't a generic way to specify default values for an effect (metadata). Things have changed since.

netmindz:
> @ewoudwijma @netmindz would you care to take a look as some effects were created by you.
> 
> > * Changed default of Palette effect: Animate rotation is now unchecked by default (it looks weird and confusing in 1D as a default)
> 
> This will need to be checked with @TripleWhy

I don't think any of those were added by me, but I can try and take a look at this PR in the next few days

wled00/FX.cpp
TripleWhy: I wasn't baling anyone for anything, just saying that we should either remove or move it, but not keep it here

DedeHai: I moved it to a palette in latest commit

wled00/FX.cpp
```diff
@@ -2360,85 +2389,59 @@ uint16_t mode_meteor() {
...
-      trail[i] = constrain(change, 0, max);
-      uint32_t col = SEGMENT.check1 ? SEGMENT.color_from_palette(i, true, false, 0, trail[i]) : SEGMENT.color_from_palette(trail[i], false, true, 255);
+    uint32_t col;
+    if (random8() <= 255 - SEGMENT.intensity) {
+      if(SEGMENT.check3) {
+        int change = trail[i] + 4 - random8(24); //change each time between -20 and +4
+        trail[i] = constrain(change, 0, max);;
+        col = SEGMENT.check1 ? SEGMENT.color_from_palette(i, true, false, 0,  trail[i]) : SEGMENT.color_from_palette(trail[i], false, true, 255);
+      }
+        else {
```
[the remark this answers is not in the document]
TripleWhy: else is still not aligned with if

wled00/FX.cpp
```diff
@@ -5665,11 +5664,11 @@ uint16_t mode_2DSindots(void) {                             // By: ldirko   http
     int y = sin8(t2 + i * SEGMENT.intensity/8)*(rows-1)/255;  // max index now 255x15/255=15!
     SEGMENT.setPixelColorXY(x, y, ColorFromPalette(SEGPALETTE, i * 255 / 13, 255, LINEARBLEND));
   }
-  SEGMENT.blur(SEGMENT.custom2>>1);
+  SEGMENT.blur(SEGMENT.custom2 >> (3 - (SEGMENT.check1<<1)));
 
   return FRAMETIME;
 } // mode_2DSindots()
-static const char _data_FX_MODE_2DSINDOTS[] PROGMEM = "Sindots@!,Dot distance,Fade rate,Blur;;!;2;c2=64";
+static const char _data_FX_MODE_2DSINDOTS[] PROGMEM = "Sindots@!,Dot distance,Fade rate,Blur,,Extra Blur;;!;2;";
```
TripleWhy, line 5671: call it smear maybe? that would be the same as in other effects where you added it

DedeHai: I am not happy with that name either... but smear is a different behaviour (it blurs without fading underlaying colors, causing a motion blurring or 'smear'). Well screw it, I will just update it to use smear (did not succeed in that before but I have an Idea)

DedeHai:
last two commits (pal0 override and palette gradient for scrolling test) are just suggestions for review/comment.

wled00/FX_fcn.cpp
```diff
@@ -213,24 +213,12 @@ CRGBPalette16 &Segment::loadPalette(CRGBPalette16 &targetPalette, uint8_t pal) {
   if (pal < 245 && pal > GRADIENT_PALETTE_COUNT+13) pal = 0;
   if (pal > 245 && (strip.customPalettes.size() == 0 || 255U-pal > strip.customPalettes.size()-1)) pal = 0; // TODO remove strip dependency by moving customPalettes out of strip
   //default palette. Differs depending on effect
-  if (pal == 0) switch (mode) {
```
blazoncek, line 216: The unfortunate consequence of this removal is that when used from existing preset, there will be no "default" palette to choose from. So, unless user chooses the effect in UI (with defined "default" palette from metadata) it will not have "default" palette defined. Once user selects it from UI the palette will be defined.

This is then inconsistent behaviour.

DedeHai: the default palette gets set when FX is loaded.
I don't fully understand in what scenario the behaviour would differ (with the latest bugfix). Can you check latest commit changes and give me a description of a test case I can try?

blazoncek: Take, for example, effect "Fire 2012". Its default platte is "Fire" (35) even if user selects palette "Default" (0).

If I have an existing preset with Fire 2012 it will most likely have `”pal”:0` as its selected palette. But after removal of this code this will mean "Party" palette until user manually selects Fire 2012 from UI which will populate its default palette to 35.

Steps to reproduce:
- use 0.14 to create preset of Fire 2012 with default values (i.e. palette chosen is Default)
- update to this PR
- restart ESP
- select preset with Fire 2012 effect
- it will display with Party colors instead of Fire
- select another preset or effect
- now select Fire 2012 in Effects tab, palette will change to Fire
- select another preset or effect
- select preset with Fire 2012
- it will display with Fire palette.

DedeHai: Got it, I will try to fix this.

DedeHai: ok, fixed it. can confirm your test scenario now works.

wled00/FX_fcn.cpp
```diff
@@ -577,27 +577,23 @@ void Segment::setMode(uint8_t fx, bool loadDefaults) {
       sOpt = extractModeDefaults(fx, "mi");  if (sOpt >= 0) mirror    = (bool)sOpt; // NOTE: setting this option is a risky business
       sOpt = extractModeDefaults(fx, "rY");  if (sOpt >= 0) reverse_y = (bool)sOpt;
       sOpt = extractModeDefaults(fx, "mY");  if (sOpt >= 0) mirror_y  = (bool)sOpt; // NOTE: setting this option is a risky business
-      sOpt = extractModeDefaults(fx, "pal"); if (sOpt < 0) sOpt = 0; setPalette(sOpt, true); //else setPalette(0);
+      sOpt = extractModeDefaults(fx, "pal"); if (sOpt >= 0) setPalette(sOpt); //else setPalette(0);
```
blazoncek, line 580: You can remove this one as it will be set below.

wled00/FX_fcn.cpp
```diff
@@ -577,27 +577,23 @@ void Segment::setMode(uint8_t fx, bool loadDefaults) {
...
     }
+    sOpt = extractModeDefaults(fx, "pal"); // always extract 'pal' to set _default_palette
+    if(sOpt == 0 || sOpt == 255) sOpt = 6; // partycolors if zero or not set (=255)
```
blazoncek, line 583: 255 is not "undefined" but is used for custom palettes. Custom palettes start at 255 and grow downwards.

blazoncek (changes requested):
This is now a little better but still does not cover HTTP API.
FYI palette with ID 255 is first custom palette.

wled00/FX_fcn.cpp
DedeHai: correct, it should be `-1` , this is a leftover from when I had this call inside setPalette() where it gets cast into an `uint8_t` resulting in 255

DedeHai:
> This is now a little better but still does not cover HTTP API.

a little hint which function handles this?

blazoncek:
> a little hint which function handles this?

set.cpp ; `handleSet()`

DedeHai:
`handleSet()` calls `seg.setMode()` which takes care of setting the default palette.

blazoncek:
> what am I missing?

Then you are all set. I do not know every line of code from the top of my head, but I do now a few relations that may be hidden.

blazoncek:
I have some additional modifications that are not strictly speed related but do decrease binary size for another kB or two.
Do I push them into this PR?

blazoncek (approved):

DedeHai:
maybe cherry-pick the latest commit before 0.15 release?(https://github.com/Aircoookie/WLED/pull/4145/commits/6a269447f6438e53835545b51d4ce2e3ebee5b15 )

blazoncek:
> maybe cherry-pick the latest commit before 0.15 release?([6a26944](https://github.com/Aircoookie/WLED/commit/6a269447f6438e53835545b51d4ce2e3ebee5b15) )

As this is a bugfix you can edit and push 0_15 branch directly. Just don't do it like you did for #3877 😄 use AC repo.

wled00/FX.cpp
```diff
@@ -2337,90 +2365,73 @@ uint16_t mode_lake() {
...
-      trail[i] = scale8(trail[i], meteorTrailDecay);
-      int index = trail[i];
-      int idx = 255;
-      int bri = SEGMENT.palette==35 || SEGMENT.palette==36 ? 255 : trail[i];
-      if (!SEGMENT.check1) {
-        idx = 0;
-        index = map(i,0,SEGLEN,0,max);
-        bri = trail[i];
+      if(meteorSmooth) { 
+        int change = trail[i] + 4 - random8(24); //change each time between -20 and +4
```
blazoncek, line 2391: This produces visible artefacts on already faded trail.
The correct way is to modify `trail[i]` only when it is > 0.

DedeHai: is this the only change required to fix it to pre 0.14 state?

blazoncek: No. Also belndPixelColor() is used when drawing meteor but I did not notice huge difference.

DedeHai: I see nothing wrong with the meteor effect in smooth setting. what exactly is your setup and what is the effect I should see?

blazoncek: When trail disappears, random pixels will briefly reappear. This is due to `change` being positive occasionally.

PR #4152, opened 2024-09-19
End oappend v2

willmmiles:
Replacement for #4134 without any usermod changes, which can be considered as a separate PR.

Replace the usage of oppend with Arduino's Print abstraction. This permits efficient use of dynamically allocated memory for returning settings page data, without requiring a single large contiguous allocation up front, or risking smashing the stack.

This change also permits the output generation functions to easily use more sophisticated serialization functions such as printf, reducing the number of required function calls. With this optimization, the overall result is approximate parity in code size to the original implementation.


softhack007:
👍

blazoncek:
I am ok to merge this PR.

netmindz:
Great work

FreakyJ:
Is this a bug or by purpose in xml.cpp? SET_F changed to F
grafik

willmmiles:
> Is this a bug or by purpose in xml.cpp? SET_F changed to F

It's intentional.  `F()` informs `print()` (via overloading) that the string is stored in PROGMEM on PROGMEM-distinct boards, so it can use an optimized/safer algorithm.  `SET_F()` casts away the distinction, which can result in poor performance or crashes.

PR #4160, opened 2024-09-25
Added ETH support for the cod.m WLED LAN Controller

UtechtDustin:
Added ethernet configuration for the upcoming cod.m WLED LAN Controller.

### DIFF
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -367,7 +367,7 @@
 #define BTN_TYPE_TOUCH_SWITCH     9
 
 //Ethernet board types
-#define WLED_NUM_ETH_TYPES        13
+#define WLED_NUM_ETH_TYPES        14
 
 #define WLED_ETH_NONE              0
 #define WLED_ETH_WT32_ETH01        1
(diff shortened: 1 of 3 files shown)


UtechtDustin:
If a customer of us will reset the controller it's easier to find the correct config.

softhack007:
> If a customer of us will reset the controller it's easier to find the correct config. Same thing like QuinLed-ESP32-Ethernet <-> TwilightLord-ESP32 Ethernet Shield.

Instead of adding a duplicate config, it's better to update the existing UI dropdown entry. 
-->  "ESP32-POE Ethernet or cod.m controler".

Case solved 😉

blazoncek:
> Same thing like QuinLed-ESP32-Ethernet <-> TwilightLord-ESP32 Ethernet Shield

These were added at the time when space was not yet at premium (more than 3 years ago). Unfortunately removing them is a breaking change that may render devices unreachable.

blazoncek (changes requested):
Adding a new type is only justified if it differs in HW configuration.

UtechtDustin:
@blazoncek so can we rename the option as @softhack007 suggested ?   
That was also our first idea, but that could end up in something like that `ESP32-POE Ethernet or cod.m controller or xyz controller or foobar ETH controller`.

PR #4163, opened 2024-09-28
Fix oappend merge

willmmiles:
A merge issue with end-oappend-v2: some strings did not get correctly converted from SET_F() to F(), which can cause crashes on ESP8266.


softhack007:
Thanks, merged 👍

PR #4174, opened 2024-10-03
Fixed Improv rejecting all properly formatted packets

### LINKED ISSUE #4173: Improv broken in 0.15.0-b5
### What happened?

Improv rejects all incoming packets.
Bug happened in commit efa32ed4f6b350512f306cae7e4ee47528416dab.

Cause is that the checksum (l.48) that used to be an "uint8_t" has been changed to a "unsigned" without adding code to keep the value in the same range as before. It thus is not equal to the checksum at the end of the packet.

### To Reproduce Bug

Try to use Improv, it rejects all communication, answering INVALID_RPC_PACKET.

### Expected Behavior

Improv correctly parses incoming packets.

### Install Method

Self-Compiled

### What version of WLED?

0.15.0-b5

ESP32-S3

### Relevant log/trace output

_No response_

### Anything else?

Either restore type to uint8_t, or for example add l.126:
    checksum &= 0xFF;

### Code of Conduct

Xevel:
Fixes #4173

### DIFF
--- a/wled00/improv.cpp
+++ b/wled00/improv.cpp
@@ -123,6 +123,7 @@ void handleImprovPacket() {
     }
 
     checksum += next;
+	checksum &= 0xFF;
     packetByte++;
   }
 }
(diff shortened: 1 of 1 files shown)


blazoncek:
Thank you.

PR #4175, opened 2024-10-03
fix Polybus canShow

Xevel:
Polybus canShow is broken on ESP32 since f4475b9d2a5c66f536568b84374588281dabaae5  (always returns true).
This PR fixes that.


blazoncek:
Ahhh! C/P should be forbidden.
Thank you.

PR #4181, opened 2024-10-05
Added integer based `sin()/cos()` functions, changed all trig functions to wled_math

DedeHai:
I wrote a new integer math based sine/cosine approximation function implementing Bhaskara I's approximation formula. It is blazing fast and quite accurate (almost 3 significant digits when converting to float). 

- new `sin16_t() / cos16_t()` are faster and more accurate than fastled versions
- `sin_approx() / cos_approx()` are float wrappers for `sin16_t() / cos16_t()` and are accurate enough to replace `sinf()/cosf()` for all current WLED purposes
- `atan2()` is used only in octopus to calculate center offset, new approximated version saves flash
- `tan(), atan(), asin(), acos(), floor(), fmod()` are used only for sunrise/sunset calculation, using existing wled_math version saves flash
- `beatsinx()` replacements are to make use of new `sin16_t()/sin8_t()` functions to reduce flash size
- Extensively tested surnise/sunset calculation: deviation is 1min. max
- Tested some of the relevant FX and found no visual difference: Julia, 2D Drift, Drift Rose, Ghost rider, Rotozoomer, Palette, Arc 1D expansion
- total flash savings: ~7.4k

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -7749,8 +7749,10 @@ uint16_t mode_2Doctopus() {
     const int C_Y = (rows / 2) + ((SEGMENT.custom2 - 128)*rows)/255;
     for (int x = 0; x < cols; x++) {
       for (int y = 0; y < rows; y++) {
-        rMap[XY(x, y)].angle  = 40.7436f * atan2_t((y - C_Y), (x - C_X));  // avoid 128*atan2()/PI
-        rMap[XY(x, y)].radius = hypotf((x - C_X), (y - C_Y)) * mapp;      //thanks Sutaburosu
+        rMap[XY(x, y)].angle  = int(40.7436f * atan2_t((y - C_Y), (x - C_X)));  // avoid 128*atan2()/PI
+        int dx = (x - C_X);
+        int dy = (y - C_Y);
+        rMap[XY(x, y)].radius = sqrtf(dx * dx + dy * dy) * mapp; //thanks Sutaburosu
       }
     }
   }
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -39,7 +39,7 @@
 // effect utility functions
 uint8_t sin_gap(uint16_t in) {
   if (in & 0x100) return 0;
-  return sin8(in + 192); // correct phase shift of sine so that it starts and stops at 0
+  return sin8_t(in + 192); // correct phase shift of sine so that it starts and stops at 0
 }
 
 uint16_t triwave16(uint16_t in) {
@@ -349,7 +349,7 @@ uint16_t mode_breath(void) {
   counter = (counter >> 2) + (counter >> 4); //0-16384 + 0-2048
   if (counter < 16384) {
     if (counter > 8192) counter = 8192 - (counter - 8192);
-    var = sin16(counter) / 103; //close to parabolic in range 0-8192, max val. 23170
+    var = sin16_t(counter) / 103; //close to parabolic in range 0-8192, max val. 23170
   }
 
   unsigned lum = 30 + var;
@@ -531,7 +531,7 @@ static uint16_t running_base(bool saw, bool dual=false) {
       }
       a = 255 - a;
     }
-    uint8_t s = dual ? sin_gap(a) : sin8(a);
+    uint8_t s = dual ? sin_gap(a) : sin8_t(a);
     uint32_t ca = color_blend(SEGCOLOR(1), SEGMENT.color_from_palette(i, true, PALETTE_SOLID_WRAP, 0), s);
     if (dual) {
       unsigned b = (SEGLEN-1-i)*x_scale - counter;
@@ -1893,16 +1893,16 @@ uint16_t mode_pride_2015(void) {
   unsigned sPseudotime = SEGENV.step;
   unsigned sHue16 = SEGENV.aux0;
 
-  uint8_t sat8 = beatsin88( 87, 220, 250);
-  uint8_t brightdepth = beatsin88( 341, 96, 224);
-  unsigned brightnessthetainc16 = beatsin88( 203, (25 * 256), (40 * 256));
-  unsigned msmultiplier = beatsin88(147, 23, 60);
+  uint8_t sat8 = beatsin88_t( 87, 220, 250);
+  uint8_t brightdepth = beatsin88_t( 341, 96, 224);
+  unsigned brightnessthetainc16 = beatsin88_t( 203, (25 * 256), (40 * 256));
+  unsigned msmultiplier = beatsin88_t(147, 23, 60);
 
   unsigned hue16 = sHue16;//gHue * 256;
-  unsigned hueinc16 = beatsin88(113, 1, 3000);
+  unsigned hueinc16 = beatsin88_t(113, 1, 3000);
 
   sPseudotime += duration * msmultiplier;
-  sHue16 += duration * beatsin88( 400, 5,9);
+  sHue16 += duration * beatsin88_t( 400, 5,9);
   unsigned brightnesstheta16 = sPseudotime;
 
   for (unsigned i = 0 ; i < SEGLEN; i++) {
@@ -1910,7 +1910,7 @@ uint16_t mode_pride_2015(void) {
     uint8_t hue8 = hue16 >> 8;
 
     brightnesstheta16  += brightnessthetainc16;
-    unsigned b16 = sin16( brightnesstheta16  ) + 32768;
+    unsigned b16 = sin16_t( brightnesstheta16  ) + 32768;
 
     unsigned bri16 = (uint32_t)((uint32_t)b16 * (uint32_t)b16) / 65536;
     uint8_t bri8 = (uint32_t)(((uint32_t)bri16) * brightdepth) / 65536;
@@ -1935,7 +1935,7 @@ uint16_t mode_juggle(void) {
   CRGB fastled_col;
   byte dothue = 0;
   for (int i = 0; i < 8; i++) {
-    int index = 0 + beatsin88((16 + SEGMENT.speed)*(i + 7), 0, SEGLEN -1);
+    int index = 0 + beatsin88_t((16 + SEGMENT.speed)*(i + 7), 0, SEGLEN -1);
     fastled_col = CRGB(SEGMENT.getPixelColor(index));
     fastled_col |= (SEGMENT.palette==0)?CHSV(dothue, 220, 255):ColorFromPalette(SEGPALETTE, dothue, 255);
     SEGMENT.setPixelColor(index, fastled_col);
@@ -1956,8 +1956,8 @@ uint16_t mode_palette() {
   constexpr mathType maxAngle                = 0x8000;
   constexpr mathType staticRotationScale     = 256;
   constexpr mathType animatedRotationScale   = 1;
-  constexpr int16_t (*sinFunction)(uint16_t) = &sin16;
-  constexpr int16_t (*cosFunction)(uint16_t) = &cos16;
+  constexpr int16_t (*sinFunction)(uint16_t) = &sin16_t;
+  constexpr int16_t (*cosFunction)(uint16_t) = &cos16_t;
 #else
   using mathType = float;
   using wideMathType = float;
@@ -2141,15 +2141,15 @@ uint16_t mode_colorwaves() {
   unsigned sPseudotime = SEGENV.step;
   unsigned sHue16 = SEGENV.aux0;
 
-  unsigned brightdepth = beatsin88(341, 96, 224);
-  unsigned brightnessthetainc16 = beatsin88( 203, (25 * 256), (40 * 256));
-  unsigned msmultiplier = beatsin88(147, 23, 60);
+  unsigned brightdepth = beatsin88_t(341, 96, 224);
+  unsigned brightnessthetainc16 = beatsin88_t( 203, (25 * 256), (40 * 256));
+  unsigned msmultiplier = beatsin88_t(147, 23, 60);
 
   unsigned hue16 = sHue16;//gHue * 256;
-  unsigned hueinc16 = beatsin88(113, 60, 300)*SEGMENT.intensity*10/255;  // Use the Intensity Slider for the hues
+  unsigned hueinc16 = beatsin88_t(113, 60, 300)*SEGMENT.intensity*10/255;  // Use the Intensity Slider for the hues
 
   sPseudotime += duration * msmultiplier;
-  sHue16 += duration * beatsin88(400, 5, 9);
+  sHue16 += duration * beatsin88_t(400, 5, 9);
   unsigned brightnesstheta16 = sPseudotime;
 
   for (int i = 0 ; i < SEGLEN; i++) {
@@ -2163,7 +2163,7 @@ uint16_t mode_colorwaves() {
     }
 
     brightnesstheta16  += brightnessthetainc16;
-    unsigned b16 = sin16(brightnesstheta16) + 32768;
+    unsigned b16 = sin16_t(brightnesstheta16) + 32768;
 
     unsigned bri16 = (uint32_t)((uint32_t)b16 * (uint32_t)b16) / 65536;
     uint8_t bri8 = (uint32_t)(((uint32_t)bri16) * brightdepth) / 65536;
@@ -2182,7 +2182,7 @@ static const char _data_FX_MODE_COLORWAVES[] PROGMEM = "Colorwaves@!,Hue;!;!";
 // colored stripes pulsing at a defined Beats-Per-Minute (BPM)
 uint16_t mode_bpm() {
   uint32_t stp = (strip.now / 20) & 0xFF;
-  uint8_t beat = beatsin8(SEGMENT.speed, 64, 255);
+  uint8_t beat = beatsin8_t(SEGMENT.speed, 64, 255);
   for (int i = 0; i < SEGLEN; i++) {
     SEGMENT.setPixelColor(i, SEGMENT.color_from_palette(stp + (i * 2), false, PALETTE_SOLID_WRAP, 0, beat - stp + (i * 10)));
   }
@@ -2198,7 +2198,7 @@ uint16_t mode_fillnoise8() {
     unsigned index = inoise8(i * SEGLEN, SEGENV.step + i * SEGLEN);
     SEGMENT.setPixelColor(i, SEGMENT.color_from_palette(index, false, PALETTE_SOLID_WRAP, 0));
   }
-  SEGENV.step += beatsin8(SEGMENT.speed, 1, 6); //10,1,4
+  SEGENV.step += beatsin8_t(SEGMENT.speed, 1, 6); //10,1,4
 
   return FRAMETIME;
 }
@@ -2210,13 +2210,13 @@ uint16_t mode_noise16_1() {
   SEGENV.step += (1 + SEGMENT.speed/16);
 
   for (int i = 0; i < SEGLEN; i++) {
-    unsigned shift_x = beatsin8(11);                          // the x position of the noise field swings @ 17 bpm
+    unsigned shift_x = beatsin8_t(11);                          // the x position of the noise field swings @ 17 bpm
     unsigned shift_y = SEGENV.step/42;                        // the y position becomes slowly incremented
     unsigned real_x = (i + shift_x) * scale;                  // the x position of the noise field swings @ 17 bpm
     unsigned real_y = (i + shift_y) * scale;                  // the y position becomes slowly incremented
     uint32_t real_z = SEGENV.step;                            // the z position becomes quickly incremented
     unsigned noise = inoise16(real_x, real_y, real_z) >> 8;   // get the noise data and scale it down
-    unsigned index = sin8(noise * 3);                         // map LED color based on noise data
+    unsigned index = sin8_t(noise * 3);                         // map LED color based on noise data
 
     SEGMENT.setPixelColor(i, SEGMENT.color_from_palette(index, false, PALETTE_SOLID_WRAP, 0));
   }
@@ -2234,7 +2234,7 @@ uint16_t mode_noise16_2() {
     unsigned shift_x = SEGENV.step >> 6;                        // x as a function of time
     uint32_t real_x = (i + shift_x) * scale;                    // calculate the coordinates within the noise field
     unsigned noise = inoise16(real_x, 0, 4223) >> 8;            // get the noise data and scale it down
-    unsigned index = sin8(noise * 3);                           // map led color based on noise data
+    unsigned index = sin8_t(noise * 3);                           // map led color based on noise data
 
     SEGMENT.setPixelColor(i, SEGMENT.color_from_palette(index, false, PALETTE_SOLID_WRAP, 0, noise));
   }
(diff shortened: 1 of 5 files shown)


softhack007:
> * total flash savings: ~7.4k

Is this with or without audioreactive usermod? Because ArduinoFFT pulls in some trig functions (sin,cos,atan, sqrt) -so possibly this finally increases flash size because two variants of the function need to be linked into the final binary.

softhack007:
> * `sin_approx() / cos_approx()` are float wrappers for `sin16_t() / cos16_t()` and are accurate enough to replace `sinf()/cosf()` for all current WLED purposes

One question: is the approximated function also faster than sinf() and cosf()? In earlier tests I was suprised to find that the libm functions are really fast, at least on esp32 and ESP32-S3 with FPU.

softhack007:
I'll give it a try with pinwheel mapping in the next days. 

Pinwheel is very sensitive to inaccuracies in sinf() and cosf(), and possibly leaves holes as a consequence.

wled00/wled_math.cpp
```diff
@@ -12,6 +12,12 @@
 
 #define modd(x, y) ((x) - (int)((x) / (y)) * (y))
 
+// Note: cos_t, sin_t and tan_t are very accurate but may be slow
+// the math.h functions use several kB of flash and are to be avoided if possible
+// sin16_t / cos16_t are faster and much more accurate than the fastled variants
+// sin_approx and cos_approx are float wrappers for sin16_t/cos16_t and have an accuracy of +/-0.0015 compared to sinf()
+// sin8_t / cos8_t are fastled replacements and use sin16_t / cos16_t. Slightly slower than fastled version but very accurate
```
softhack007, line 19: 👍 very good that we have a replacement for sin8 and cos8. 
These FastLed functions were so inaccurate that you could not even draw a circle with them.

DedeHai: Here is the comparison between the new version and fastled (compared to `sinf()` )
sin8():
image

sin16():
image

the new sin8_t() is a bit slower than the fastled variant due to overhad in converstion to 16bit range. The approximation can be re-written to any scale so it could be done for 8bit inputs/outputs natively but since sin16_t() is now faster than sin8() it would make more sense to use the sin16_t() instead wherever speed is of utmost importance.

DedeHai:
> Is this with or without audioreactive usermod?
Goot point. I checked and it is the same: ArduinoFFT uses cos(x) which is the double float implementation and sqrt() was not touched in this PR. 
btw: did you ever check if the Espressif DSP library functions can be used for FFT and if there is any speed/memory/precision advantage? According to the doc it has speed optimized FFT for ESP32 and ESP32 S3.

DedeHai:
> One question: is the approximated function also faster than sinf() and cosf()? In earlier tests I was suprised to find that the libm functions are really fast (faster than fastled sin16/cos16), at least on esp32 and ESP32-S3 with FPU.

Another good point. I did not actually measure that. But now I checked with rough test. Here are my findings:
numbers are in function calls per second executed:

| MCU             | `sinf()` | `sin_approx()` | 
| :--------------- | :------: | :----: |
| ESP32-C3        |   9k |  75k | 
| ESP32-S3        |   130k   | 184k |

I am not 100% sure about the method I used so no warranty about correctness of these numbers. Using the same method for sin16() functions gave like a 50x speed increase, which seems impossible without the compiler playing some tricks.

So on an S3 the difference is not huge thanks to the FPU but still an improvement. On S2, C3 (and ESP8266) the speed-gain is huge.

wled00/wled_math.cpp
softhack007: Thanks, I like the graphs 👍 very useful to understand accuracy and limitations.

wled00/wled_math.cpp
```diff
@@ -49,6 +55,85 @@ float tan_t(float x) {
...
+}
+
+int16_t cos16_t(uint16_t theta) {
+  return sin16_t(theta + 16384); //cos(x) = sin(x+pi/2)
+}
+
+uint8_t sin8_t(uint8_t theta) {
+  int32_t sin16 = sin16_t((uint16_t)theta * 257); // 255 * 257 = 0xFFFF
+  sin16 += 0x7FFF; //shift result to range 0-0xFFFF
+  return sin16 >> 8;
```
softhack007, line 81: I think a bit of rounding could improve the result - the plot shows that all errors are to the negative, while ideally they should be distributed evenly around 0.

Maybe add something like
``` c++
uint32_t usin16 = sin16; // make sure we're working unsigned
usin16 += 127;                  // prepare for proper rounding
return usin16 >> 8;           // result rounded to nearest positive integer
```

wled00/wled_math.cpp
softhack007: > but since sin16_t() is now faster than sin8() it would make more sense to use the sin16_t() instead wherever speed is of utmost importance.

I agree - I don't think we need a specialized sin8, as most "sine wave" effects are using beatsin88() and not sin8.

softhack007: Another idea about sin8_t and cos8_t:

You could move these two functions into fx.cpp and make them "static inline". This will allow the compiler to optimize out function calls (->faster). I think there is not other place where these 8bit version would be needed.

wled00/wled_math.cpp
DedeHai: good idea. I already tried that and the problem is: overflow.
the returned values are -0x7FFF to + 0x7FFF, sin8 does use the output range 0-255, so the 16bit int needs to be shifted to uint values by addint 0x7FFF. This makes the min value 0 and the max value 0xFFFF. 0xFFFF >> 8 is 255, add 127 and it overflows to 0. To prevent that, an if statement is required, making it slower. Unless there is a clever workaround I am not seeing.

wled00/wled_math.cpp
DedeHai: > I think there is not other place where these 8bit version would be needed.

It is used in beatsin8, which is also used in sound simulation in util.cpp.
While making that an inline function could boost speed, I don't think it those extra few instructions will do any harm. If we really want it faster, it should have its own dedicated function (which is doable, it uses roughly 100 bytes)

DedeHai:
> I'll give it a try with pinwheel mapping in the next days.
> 
> Pinwheel is very sensitive to inaccuracies in sinf() and cosf(), and possibly leaves holes as a consequence.

I am looking at pinwheel RN. The issue is not the precision of sin() cos() but the number of steps:
I just tried with 128x16 and even with floats there are missing pixels. If I increase steps for XL, they are gone, even when using the integer version of sin() cos() (more testing required though)

wled00/wled_math.cpp
softhack007: Good point 👍 yes 8 did not think about overflows.
I see o e workarounds, not sure which variant would be faster
* `if (usin16 < 65408) usin16 += 127;
* `usin16 = max(usin16, usin16 + 127);`

softhack007:
@dedehai pinwheel has a generic problem with holes - we tested it until 56x56, so below 56x56 there should be no holes  (using sinf()/cosf(), not the _t approximations from AC).

softhack007:
> any reason it is 512? it could be

51w was just a "shot in the dark". You're right we could increase the value to 2^15.

DedeHai:
> @DedeHai pinwheel has a generic problem with holes - we tested it until 56x56, so below 56x56 there should be no holes (using sinf()/cosf(), not the _t approximations from AC).

I saw that and I am trying to work out the math why that is.
Here is what I have so far:

`(maxdim/2)*sin(2pi/steps) <= 1`  -> requirement for no holes, i.e. maximum angle such that at the furthest distance its less than 1 pixel
therefore:
`steps=(2*pi)/asin(1/(maxdim/2))`
since asin(x) = x for x << 1  this resolves into:
`steps >= pi * maxdim`
now what comes  into play is the integer math applied and the rounding errors adding up as it is done incrementally and not by multiplication. So steps needs to be increased to overcome that. That is what I am trying to figure out.

wled00/wled_math.cpp
softhack007: @DedeHai can you test this variant? 

```c++
uint8_t sin8_t(uint8_t theta) {
  int32_t sin16 = sin16_t((uint16_t)theta * 257); // 255 * 257 = 0xFFFF
  sin16 += 0x7FFF + 128; //shift result to range 0-0xFFFF, +128 for rounding
  return min(sin16, int32_t(0xFFFF)) >> 8; // min performs saturation, and prevents overflow
}
```

 It just adds one or two CPU instructions for "max", and should keep the error below +-0.5

DedeHai:
@softhack007 looking at the pinwheel some more. I have some fixes but also: why is the center calculated in such a complicated way (with shifting odd rays)? I removed all that code and to me it still looks the same. Can you explain the intentions?

wled00/wled_math.cpp
softhack007: @DedeHai can you test this variant? 

```c++
uint8_t sin8_t(uint8_t theta) {
  int32_t sin16 = sin16_t((uint16_t)theta * 257); // 255 * 257 = 0xFFFF
  sin16 += 0x7FFF + 128; //shift result to range 0-0xFFFF, +128 for rounding
  return min(sin16, int32_t(0xFFFF)) >> 8; // min performs saturation, and prevents overflow
}
```

 It just adds one or two CPU instructions for "min", and should keep the error below +-0.5

wled00/wled_math.cpp
DedeHai: that works, maybe a good tradoff between accuracy and speed.
image

softhack007:
> @softhack007 looking at the pinwheel some more. I have some fixes but also: why is the center calculated in such a complicated way (with shifting odd rays)? I removed all that code and to me it still looks the same. Can you explain the intentions?

@dedehai the "odd rays" part was an invention from @Brandon502, to reduce the number of "full rays" by only drawing outer parts of odd-numbered Ray's. That was around 30% faster in our tests.

Its correct that omitting this optimization should not influence the look.

wled00/wled_math.cpp
softhack007: 👍 👍

wled00/wled_math.cpp
softhack007: @DedeHai ok if I push my sin8_t to your branch?

wled00/wled_math.cpp
softhack007: > that works, maybe a good tradoff between accuracy and speed. 

I've just compared your new trig functions against the old one, using the "Lissajous" effect on 128x64. It's really a major difference. The "fastled" lissajous curve looks very edgy and somehow _wrong_, more like a twisted hexagon. With your new functions, I see a perfect wave.

 Good job👍 thanks.

softhack007:
> @softhack007 looking at the pinwheel some more. I have some fixes but

@DedeHai would be nice if we find an improvement, because each time I get a larger setup, there ar new holes in pinwheel. Looks like a kind of Moire effect. I've found the "magic parameters" up to 128x128: see https://github.com/MoonModules/WLED/blob/a84216947bfb28488e58837d5a6732836be93d6c/wled00/FX_fcn.cpp#L833-L854

'm really curious to know if there is a better way to avoid these inaccuracies.

Brandon502:
> > @softhack007 looking at the pinwheel some more. I have some fixes
> 
> @DedeHai would be nice if we find an improvement, because each time I get a larger setup, there are new holes in pinwheel. Looks like a kind of Moire effect.
> 
> I've found the "magic parameters" up to 128x128: see https://github.com/MoonModules/WLED/blob/a84216947bfb28488e58837d5a6732836be93d6c/wled00/FX_fcn.cpp#L833-L854
> 
> I'm really curious to know if there is a better way to avoid these inaccuracies, while preserving speed.

How much memory can a expand1D type use? If it could work similar to jMap you could create a lookup map on the first call then never have to do the math again. Would never produce holes and likely be much faster, but the memory footprint might be too big for 128x128+

DedeHai:
@Brandon502  I have _sort of_ working code improvements: it may take a slight hit on speed but it work for larger sizes and reduces steps for smaller ones by dynamically calculating the required steps. Also "pixel holes" are controlled better by reducing the step size, running more 'blind' loops (but setting less pixels):
https://github.com/DedeHai/WLED/tree/0_15_pinwheel_improvement
I have a basic concept on how to improve speed by also dynamically calculating required steps vs. current "radius" being drawn. While it is not perfect and a compromise between code-complexity and calculation speed (it needs to run on ESP32 C3 and ESP8266 as well). For the ESP32 and ESP32 S3 more calculations could be done in float to squeeze out more speed (by more accurately calculating if a pixel needs to be drawn or not). 
I will update the branch later today and make a PR, we can continue the discussion there as this PR has a different purpose.

softhack007 (approved):
Tested and works well for me.

wled00/wled_math.cpp
```diff
@@ -88,8 +88,8 @@ uint8_t cos8_t(uint8_t theta) {
 float sin_approx(float theta)
 {
   theta = modd(theta, TWO_PI); // modulo: bring to -2pi to 2pi range
-  if(theta < 0) theta += M_TWOPI; // 0-2pi range
-  uint16_t scaled_theta = (uint16_t)(theta * (0xFFFF / M_TWOPI));
+  if(theta < 0) theta += TWO_PI; // 0-2pi range
```
blazoncek, line 91: Why? Why not change `M_PI` below?

DedeHai: good question. I think this was auto-complete suggestion from co-pilot. 
I will make it consistant to M_PI variants.
Also: the conversion of the angle using float is really not needed, I added that to fix a bug. but the actual fix is: cast a negative float to a `int` and NEVER to a `unsigned` type, as that behaviour is undefined and depends on the CPU/Compiler, hence the issues on the C3 which does `(uint)(float) = 0`

softhack007:
I was wondering if this should be included in the upcoming 0.15.0 release, maybe together with the new WLEDgetColorFromPalette function extracted from #4138 ?

We have "feature freeze" for 0.15.0 already, but I think this PR might be worth making an exception to the rule
* it's well tested and well analysed already
* it really  improves visual quality of some effects, like lissajous
* the new WLEDgetColorFromPalette function makes some effects (like octopus) look much smoother on larger setup.

@DedeHai @willmmiles @netmindz @blazoncek @Aircoookie what do you think?

DedeHai:
I am not sure about this. These changes were not in any beta release and while I tested the changes and found no problem so far, I think this should first be run through beta testing, there may be hidden surprises on ESP8266 or large 1D setups.

blazoncek:
IMO, no but there is no reason to take my answer into account.

softhack007:
> I am not sure about this. These changes were not in any beta release

@DedeHai thanks I think your judgement as the original author is very important. I agree with with, better to get some confidence (beta testing) before releasing this (and #4138) to all users. 👍 

So we'll merge all the PR tagged as "optimization" after 0.15.0 is released, right?

DedeHai:
I think the decision is up to @Aircoookie whether to push 0.15 release ASAP or add in these improvements.

If it were up to me I would add all pending improvements I and @blazoncek made as they improve FX speed and FPS  and FX quality quite a bit but only if these are extensively tested as there are chances of introducing new bugs. By all I mean this PR as well as https://github.com/Aircoookie/WLED/pull/4138 https://github.com/Aircoookie/WLED/pull/4145 https://github.com/Aircoookie/WLED/pull/4245

TripleWhy:
FYI I just stumbled upon this PR, and decided to run some quick execution time benchmarks on the ESP32 (lower is better):
image
[benchmark.log](https://github.com/user-attachments/files/17904088/benchmark.log)

Btw. I believe the `_t` postfix used to indicate that the implementation was a taylor series, so adding new functions with that postfix but without taylor seems strange

blazoncek:
@TripleWhy there's also code size aspect to sin_t.

This was the primary reason @Aircoookie introduced these functions.

DedeHai:
@TripleWhy interesting results, do you have a link to the code you used?  Would be interesting to compare other MCUs. 
the _t I just chose to match the old ones, if that is confusing, I do not mind chanigng it, maybe to _b for Bhaskara? 

One thing that is really strange in your graph is that `sin_approx` and `sin16_t` are so far apart in speed: approx is just one multiplication on top plus the overhead of the function call. Maybe it should be inlined? Even stranger is the same for `cos_approx` which is just a call to `sin_approx` with one addition, so there is zero reason it being not almost identical to `sin_approx` but double. how sure are you about the results (and the naming)?

softhack007:
Thanks @TripleWhy for the benchmark 🥇 
Actually from my POV the main reason to use the new functions is accuracy when compared to fastLed sin8 or sin16 - especially fastled sin8 is so bad that you can't even draw a circle with it.


> Btw. I believe the `_t` postfix used to indicate that the implementation was a taylor series, so adding new functions with that postfix but without taylor seems strange

Finally someone has an explanation for the "_t" - it's not written in the code, and my best guess was it stands for "trick" 😆 

As nobody except the three of us knows what "_t" stands for, 8t does not matter much which letter we pick. But for minimal code change, imho it would be good to keep the "_t".

TripleWhy:
> there's also code size aspect to sin_t.

I didn't mean to give any recommendation here, just information.

> interesting results, do you have a link to the code you used?

Just pushed it here: https://github.com/TripleWhy/WLED/tree/benchTrig  
I believe there are some other changes, but the measurements should be representative.

> One thing that is really strange in your graph is that sin_approx and sin16_t are so far apart in speed: approx is just one multiplication on top plus the overhead of the function call.

And some float<->int conversions.

> Even stranger is the same for cos_approx which is just a call to sin_approx with one addition

I guess float additions and calls are expensive?  
If you want to do a relative comparison, you also need to take into account that the benchmark code itself also consumes a bit of time. So, here is an updated graph with float additions, and all times adjusted for benchmark code overhead:

image

Well... According to this diagram, I have no idea what's going on. I guess more fiddling is required here.

> how sure are you about the results (and the naming)?

Very. Look at the raw log of sin_t vs sin_approx and cos_t vs cos_approx. They are aliases, I just left them in for sanity checking. The numbers are identical. There is also (almost) no run-to run variation, these are fixed frequency single threaded risc chips without task scheduling after all. You are free to check the code and run it for yourself.

> main reason to use the new functions is accuracy when compared to fastLed sin8 or sin16 - especially fastled sin8 is so bad that you can't even draw a circle with it.

I don't disagree. As I said, I was just publishing information that I generated locally.

TripleWhy:
> there's also code size aspect to sin_t.

Oh right, I also did a size comparison:

```
with both *_t and *_approx:
RAM:   [==        ]  16.0% (used 52592 bytes from 327680 bytes)
Flash: [========= ]  86.4% (used 1358941 bytes from 1572864 bytes)


with std::*
RAM:   [==        ]  16.0% (used 52592 bytes from 327680 bytes)
Flash: [========= ]  86.8% (used 1364649 bytes from 1572864 bytes)


with *f:
RAM:   [==        ]  16.0% (used 52592 bytes from 327680 bytes)
Flash: [========= ]  86.7% (used 1363177 bytes from 1572864 bytes)
```

TripleWhy:
> sin_t vs sin_approx and cos_t vs cos_approx. They are aliases

Uh... looking at the code a bit more, it appears that this aliasing isn't intentional?

DedeHai:
> Uh... looking at the code a bit more, it appears that this aliasing isn't intentional?

it is intentional, sin_t / cos_t are the original functions and are replaced with the new, faster implementation.

I ran my own benchmark code and it comes out to this:
image
this is the time in ms for 1 million executions (+ a little overhead), in my test, sin_approx and cos_approx are closer to what I would expect. still slower than what I hoped, given that the underlaying sin16_t is so much faster and ESP32 has hardware float.

TripleWhy:
> it is intentional, sin_t / cos_t are the original functions and are replaced with the new, faster implementation.

In that case the original implementations should be removed.

> this is the time in ms for 1 million executions (+ a little overhead), in my test, sin_approx and cos_approx are closer to what I would expect.

Interesting. What are the differences to my runs?  
Were those executed on a regular ESP32 (not a risc-v version) as well? What numbers do you call the functions with? I took samples from all normal floats greater than 0 and lte TWO_PI. How sure are you that the compiler didn't optimize the repetition? Is your benchmark code available somewhere?

DedeHai:
> In that case the original implementations should be removed.

I wanted to keep the original functions as they do take a completely different approach. I could comment them out for readability though.

> Interesting. What are the differences to my runs? Were those executed on a regular ESP32 (not a risc-v version) as well? What numbers do you call the functions with? I took samples from all normal floats greater than 0 and lte TWO_PI. How sure are you that the compiler didn't optimize the repetition? Is your benchmark code available somewhere?

this is the code I used:
[trig_benchmark.zip](https://github.com/user-attachments/files/17908311/trig_benchmark.zip)

blazoncek:
I want to clarify that anything I introduced into WLED was `libc` based. I think those functions are well tested and optimised for speed *and* accuracy. And I never bothered doing size or speed comparisons of alternatives.

But that does not matter. There may have been time in WLED development that needed those functions and that time may have passed. It may be time to move forward, there was a lot of development and progress in every way since that introduction. There will always be a reason to optimise either for speed or accuracy or code size. It is difficult to have all at once.

As for this PR: Since @DedeHai started to replace FastLED core functions (here and elsewhere) it might be beneficial to remove FastLED entirely from dependencies if those replacements are not copy/paste and perform better.

TripleWhy:
The most obvious difference between your benchmark and mine is that you only test for a single input: 0.5.

Here is my esp32 result again (this time with the same dimensions as the following images for better comparison):
esp32

And here is one where I replaced the call argument with 0.5:
esp32 0 5 only

For completeness, here is my run on a c3:
c3

DedeHai:
> For completeness, here is my run on a c3:

this is really weird, I do get completely different results, especially on sinf/cosf: they get much worse if I use an incremented input.
image

I also found room to optimize the approx versions, turns out `M_PI` and its variations are treated as a double float by default.

softhack007:
> Are you happy for this to go into the 0.15.0.rc1 @softhack007 or wait for 0.1.5.1 ?

I think it could even go into 0.15.0.rc1 - just that @DedeHai previously said he would like to see some more confidence testing on 8266. If @DedeHai is happy with the benchmarks and tests we have, then OK for me to put this into 0.15.0.rc1.

DedeHai:
I think it can go into  0.15.0.rc1, @TripleWhy an me were running some more tests and it is looking good, I will post results in a bit, once I confirmed it works on S2, S3 and ESP8266 without any issues.

netmindz:
Ok, if you can fix the merge conflict @DedeHai fixed, we can merge

DedeHai:
my tests are done, its all looking good but there may be a small pitfall @TripleWhy found: calling `sin_approx()` and `cos_approx` with large floats leads to increasing errors. `mode_2DJulia()` for example calls the `sin()` function with a parameter of `strip.now/305.f` which can grow large. There may be other places in the code where this happens. These function calls should be replaced with the integer version as that one truncates to a uint16_t and always works. 
I will check if I find other places in the code where this could be problematic. 
Apart from that sure, I can fix the merge errors and merge.

As for my final results: here are some graphs. Input to float versions was approximately 0-2pi, integer versions just use a increment of 1 for each loop.
image
image
image
image

wled00/FX.cpp
```diff
@@ -7737,9 +7737,9 @@ uint16_t mode_2Doctopus() {
     const int C_Y = (rows / 2) + ((SEGMENT.custom2 - 128)*rows)/255;
     for (int x = 0; x < cols; x++) {
       for (int y = 0; y < rows; y++) {
-        rMap[XY(x, y)].angle  = int(40.7436f * atan2_t((y - C_Y), (x - C_X)));  // avoid 128*atan2()/PI
         int dx = (x - C_X);
         int dy = (y - C_Y);
+        rMap[XY(x, y)].angle  = int(40.7436f * atan2_t(dy, dx));  // avoid 128*atan2()/PI
```
blazoncek, line 7742: This is totally unrelated change.

PR #4183, opened 2024-10-06
Add WLED_AUTOSEGMENTS compile flag

PaoloTK:
Add flag to enable setting "Make a segment for each output" at compile.

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -720,7 +720,7 @@ class WS2812FX {  // 96 bytes
 #ifndef WLED_DISABLE_2D
       panels(1),
 #endif
-#ifdef WLED_AUTOSEGMENT_OUTPUTS
+#ifdef WLED_AUTOSEGMENTS
       autoSegments(true),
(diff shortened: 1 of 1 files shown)


blazoncek:
If you don't mind I think WLED_OUTPUT_SEGMENTS or WLED_OUTPUT_AUTOSEGMENTS or WLED_CREATE_AUTOSEGMENTS or perhaps just WLED_AUTOSEGMENTS would be more appropriate.

PaoloTK:
> If you don't mind I think WLED_OUTPUT_SEGMENTS or WLED_OUTPUT_AUTOSEGMENTS or WLED_CREATE_AUTOSEGMENTS or perhaps just WLED_AUTOSEGMENTS would be more appropriate.

Sure, I like 2nd and 4th the most but your call.

blazoncek:
I like the simplicity of the last one too.

blazoncek (approved):
Approved from my side.

PR #4188, opened 2024-10-12
Added BRT timezone

LuisFadini:
Added the Brasília timezone, which is widely used across most regions of Brazil.


Aircoookie (approved):
I approve of this change.

In general and thinking long term,  I think we could consider storing time zones in JS as json (or potentially even loading them in externally so that every actual timezone is also supported) and only passing UTC offset + DST change dates to the ESP as settings. Not pressing though as more timezone entries do not require extra RAM at least.

blazoncek (approved):
Good for me.

PR #4200, opened 2024-10-17
compile warning fix

ingDIY:
added check for custom compile user defines, otherwise warning message are generated for custom build with DEFAULT_LED_TYPE or DEFAULT_LED_COUNT defined

### DIFF
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -584,7 +584,7 @@
 #endif
 
 #ifndef DEFAULT_LED_COUNT
-#define DEFAULT_LED_COUNT 30
+  #define DEFAULT_LED_COUNT 30
 #endif
 
 #define INTERFACE_UPDATE_COOLDOWN 1000 // time in ms to wait between websockets, alexa, and MQTT updates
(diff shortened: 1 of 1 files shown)


blazoncek:
Defining LED outputs was redesigned in 0.15.0-b6.
This PR is irrelevant.

softhack007:
@ingDIY you can already use these defines for custom-compile overrides of pixel counts, LED type, and LED pins:

https://github.com/Aircoookie/WLED/blob/e9d2182390d43d7dd25492f6555d082280e79b3b/wled00/FX_fcn.cpp#L32-L46

softhack007:
From what I see in the code, it shouldn't be necessary to override DEFAULT_LED_TYPE or DEFAULT_LED_COUNT.

Closing PR without merge.

ingDIY:
> @ingDIY you can already use these defines for custom-compile overrides of pixel counts, LED type, and LED pins:
> 
> https://github.com/Aircoookie/WLED/blob/e9d2182390d43d7dd25492f6555d082280e79b3b/wled00/FX_fcn.cpp#L32-L46

OK thanks. Sorry but it's quite hard to track all these code changes. So the #defines i fixed are old and will be removed?

I based my modification following [WLED compile helper YAML](https://wled-compile.github.io)
and in the generated YAML it is still using
-D DEFAULT_LED_COUNT=50
-D DEFAULT_LED_TYPE=TYPE_WS2812_RGB

so please could you help me to better understand what custom define should I use?

PR #4209, opened 2024-10-18
Remove simulateSound() in favour of getAudioData()

blazoncek:
This is supposed to simplify getting either simulated sound or actual audio data for effects.
Builds on the work by @netmindz .

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -59,15 +59,6 @@ int8_t tristate_square8(uint8_t x, uint8_t pulsewidth, uint8_t attdec) {
   return 0;
 }
 
-static um_data_t* getAudioData() {
-  um_data_t *um_data;
-  if (!UsermodManager::getUMData(&um_data, USERMOD_ID_AUDIOREACTIVE)) {
-    // add support for no audio
-    um_data = simulateSound(SEGMENT.soundSim);
-  }
-  return um_data;
-}
-
 // effect functions
 
 /*
@@ -6330,13 +6321,13 @@ uint16_t mode_ripplepeak(void) {                // * Ripple peak. By Andrew Tuli
   if (!SEGENV.allocateData(dataSize)) return mode_static(); //allocation failed
   Ripple* ripples = reinterpret_cast<Ripple*>(SEGENV.data);
 
-  um_data_t *um_data = getAudioData();
-  uint8_t samplePeak    = *(uint8_t*)um_data->u_data[3];
+  um_data_t *um_data = getAudioData(SEGMENT.soundSim);
   #ifdef ESP32
-  float   FFT_MajorPeak = *(float*)  um_data->u_data[4];
+  float   FFT_MajorPeak = *(float*)um_data->u_data[4];
   #endif
-  uint8_t *maxVol       =  (uint8_t*)um_data->u_data[6];
-  uint8_t *binNum       =  (uint8_t*)um_data->u_data[7];
+  uint8_t samplePeak = *(uint8_t*)um_data->u_data[3];
+  uint8_t *maxVol    =  (uint8_t*)um_data->u_data[6];
+  uint8_t *binNum    =  (uint8_t*)um_data->u_data[7];
 
   // printUmData();
 
@@ -6417,9 +6408,9 @@ uint16_t mode_2DSwirl(void) {
   int ni = (cols - 1) - i;
   int nj = (cols - 1) - j;
 
-  um_data_t *um_data = getAudioData();
-  float volumeSmth  = *(float*)   um_data->u_data[0]; //ewowi: use instead of sampleAvg???
-  int   volumeRaw   = *(int16_t*) um_data->u_data[1];
+  um_data_t *um_data = getAudioData(SEGMENT.soundSim);
+  float volumeSmth = *(float*)  um_data->u_data[0]; //ewowi: use instead of sampleAvg???
+  int   volumeRaw  = *(int16_t*)um_data->u_data[1];
 
   SEGMENT.addPixelColorXY( i, j, ColorFromPalette(SEGPALETTE, (strip.now / 11 + volumeSmth*4), volumeRaw * SEGMENT.intensity / 64, LINEARBLEND)); //CHSV( ms / 11, 200, 255);
   SEGMENT.addPixelColorXY( j, i, ColorFromPalette(SEGPALETTE, (strip.now / 13 + volumeSmth*4), volumeRaw * SEGMENT.intensity / 64, LINEARBLEND)); //CHSV( ms / 13, 200, 255);
@@ -6443,8 +6434,8 @@ uint16_t mode_2DWaverly(void) {
   const int cols = SEGMENT.virtualWidth();
   const int rows = SEGMENT.virtualHeight();
 
-  um_data_t *um_data = getAudioData();
-  float   volumeSmth  = *(float*)   um_data->u_data[0];
+  um_data_t *um_data = getAudioData(SEGMENT.soundSim);
+  float volumeSmth = *(float*)um_data->u_data[0];
 
   SEGMENT.fadeToBlackBy(SEGMENT.speed);
 
@@ -6487,8 +6478,8 @@ uint16_t mode_gravcenter(void) {                // Gravcenter. By Andrew Tuline.
   if (!SEGENV.allocateData(dataSize)) return mode_static(); //allocation failed
   Gravity* gravcen = reinterpret_cast<Gravity*>(SEGENV.data);
 
-  um_data_t *um_data = getAudioData();
-  float   volumeSmth  = *(float*)  um_data->u_data[0];
+  um_data_t *um_data = getAudioData(SEGMENT.soundSim);
+  float volumeSmth = *(float*)um_data->u_data[0];
 
   //SEGMENT.fade_out(240);
   SEGMENT.fade_out(251);  // 30%
@@ -6532,8 +6523,8 @@ uint16_t mode_gravcentric(void) {                     // Gravcentric. By Andrew 
   if (!SEGENV.allocateData(dataSize)) return mode_static();     //allocation failed
   Gravity* gravcen = reinterpret_cast<Gravity*>(SEGENV.data);
 
-  um_data_t *um_data = getAudioData();
-  float   volumeSmth  = *(float*)  um_data->u_data[0];
+  um_data_t *um_data = getAudioData(SEGMENT.soundSim);
+  float volumeSmth = *(float*)um_data->u_data[0];
 
   // printUmData();
 
@@ -6580,8 +6571,8 @@ uint16_t mode_gravimeter(void) {                // Gravmeter. By Andrew Tuline.
   if (!SEGENV.allocateData(dataSize)) return mode_static(); //allocation failed
   Gravity* gravcen = reinterpret_cast<Gravity*>(SEGENV.data);
 
-  um_data_t *um_data = getAudioData();
-  float   volumeSmth  = *(float*)  um_data->u_data[0];
+  um_data_t *um_data = getAudioData(SEGMENT.soundSim);
+  float volumeSmth = *(float*)um_data->u_data[0];
 
   //SEGMENT.fade_out(240);
   SEGMENT.fade_out(249);  // 25%
@@ -6617,8 +6608,8 @@ static const char _data_FX_MODE_GRAVIMETER[] PROGMEM = "Gravimeter@Rate of fall,
 //   * JUGGLES      //
 //////////////////////
 uint16_t mode_juggles(void) {                   // Juggles. By Andrew Tuline.
-  um_data_t *um_data = getAudioData();
-  float   volumeSmth   = *(float*)  um_data->u_data[0];
+  um_data_t *um_data = getAudioData(SEGMENT.soundSim);
+  float volumeSmth = *(float*)um_data->u_data[0];
 
   SEGMENT.fade_out(224); // 6.25%
   unsigned my_sampleAgc = fmax(fmin(volumeSmth, 255.0), 0);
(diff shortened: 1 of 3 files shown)


wled00/FX.cpp
```diff
@@ -6443,8 +6434,8 @@ uint16_t mode_2DWaverly(void) {
   const int cols = SEGMENT.virtualWidth();
   const int rows = SEGMENT.virtualHeight();
 
-  um_data_t *um_data = getAudioData();
-  float   volumeSmth  = *(float*)   um_data->u_data[0];
+  um_data_t *um_data = getAudioData(SEGMENT.soundSim);
```
netmindz, line 6437: Why pass a variable that was already visible inside the method?

blazoncek: I was thinking about visibility change in the future.

blazoncek:
> Please can you write a quick summary of what you were trying to do so I'm not having to guess that from the code

Just condense 2 functions into one.

wled00/FX.cpp
DedeHai: Passing by reference may save on code size, just a guess.

blazoncek: There are 29 calls to `getAudioData()`. While it is true that `SEGMENT.soundSim` is globally accessible, it is only due to the fact that, if hidden behind function calls, the code would increase tremendously (done that).

> Passing by reference may save on code size

`soundSim` is a bitfield so passing by reference is not possible.

willmmiles:
> Just condense 2 functions into one.

Heh, I was going to suggest keeping the two parts as separate functions in the source!  It makes more sense to me to keep the code for "selecting the sound source" isolated from the code for "computing the audio data simulation".  (That said: moving `getAudioData()` to utils.cpp, and marking `simulateSound()` static, should allow the compiler to inline `simulateSound` and reduce the total number of function calls.)

DedeHai:
From an FX view I like the consolidation approach better, it should not be up to an effect to choose a 'system setting'

willmmiles:
> From an FX view I like the consolidation approach better, it should not be up to an effect to choose a 'system setting'

Was this meant to be a reply to me?

To clarify what I mean, I think that this code:
```c++
// utils.cpp
static um_data_t* simulateSound(uint8_t simulationId) {
  // .. body of simulateSound
}

um_data_t* getAudioData(uint8_t simulationId) {
  um_data_t *um_data;
  if (UsermodManager::getUMData(&um_data, USERMOD_ID_AUDIOREACTIVE)) {
    return um_data;
  }
  return simulateSound(simulationId);
}
```
... is a better organization than this code:

```c++
// utils.cpp
um_data_t* getAudioData(uint8_t simulationId) {
  um_data_t *um_data;
  if (UsermodManager::getUMData(&um_data, USERMOD_ID_AUDIOREACTIVE)) {
    return um_data;
  }

  // .. body of simulateSound
}
```

While the API to the FX layer and the compiled results should be identical, I think there's some practical value in arranging the source code to keep individual tasks ("how do I decide where to get some audio data?" and "how do I compute audio simulation?") as separate functions.  IMO it can make things easier to follow and maintain, and clearly scopes which local variables are relevant to what part of the overall task.

That's all I was trying to suggest!   Bikeshedding at its finest. ;)

DedeHai:
That makes a lot of sense actually, I misunderstood.

wled00/util.cpp
```diff
@@ -422,74 +424,60 @@ um_data_t* simulateSound(uint8_t simulationId)
...
       break;
     case UMS_WeWillRockYou:
       if (ms%2000 < 200) {
         volumeSmth = random8(255);
-        for (int i = 0; i<5; i++)
-          fftResult[i] = random8(255);
-      }
-      else if (ms%2000 < 400) {
+        for (int i = 0; i < 5; i++) fftResult[i] = random8(255);
+      } else if (ms%2000 < 400) {
```
DedeHai, line 442: not related to this PR directly but something I recently learned:
since modulo is slow on ESP8266, ESP32 C3 and maybe even S2 it can be replaced with a simple `&` if done with a power of 2. If the 2000ms is not so important (which I am guessing it is not) all the `ms%2000`could be replaced by `ms&2047` or `ms&0x7F`

wled00/FX.cpp
```diff
@@ -7305,9 +7293,12 @@ uint16_t mode_waterfall(void) {                   // Waterfall. By: Andrew Tulin
...
 
-    //uint8_t pixCol = (log10f((float)FFT_MajorPeak) - 2.26f) * 177;  // 10Khz sampling - log10 frequency range is from 2.26 (182hz) to 3.7 (5012hz). Let's scale accordingly.
-    uint8_t pixCol = (log10f(FFT_MajorPeak) - 2.26f) * 150;           // 22Khz sampling - log10 frequency range is from 2.26 (182hz) to 3.967 (9260hz). Let's scale accordingly.
-    if (FFT_MajorPeak < 182.0f) pixCol = 0;                           // handle underflow
+#if defined(CONFIG_IDF_TARGET_ESP32C3) || defined(CONFIG_IDF_TARGET_ESP32S2)
+    uint8_t pixCol = (log10f(FFT_MajorPeak) - 2.26f) * 177;   // 16Khz sampling - log10 frequency range is from 2.26 (182hz) to 3.85 (7079hz). Let's scale accordingly.
+#else
+    uint8_t pixCol = (log10f(FFT_MajorPeak) - 2.26f) * 150;   // 22Khz sampling - log10 frequency range is from 2.26 (182hz) to 3.967 (9260hz). Let's scale accordingly.
+#endif
+    if (FFT_MajorPeak < 182.0f) pixCol = 0;                   // handle underflow
```
netmindz, line 7301: what is this about - it seems unrelated to the merge of simulateSound and getAudioData?

wled00/util.cpp
```diff
@@ -401,6 +400,9 @@ um_data_t* simulateSound(uint8_t simulationId)
 
   static um_data_t* um_data = nullptr;
 
+  um_data_t* um_data_tmp;
+  if (UsermodManager::getUMData(&um_data_tmp, USERMOD_ID_AUDIOREACTIVE)) return um_data_tmp;
```
netmindz, line 404: Easier to read  is return is on it's own line inside {}

netmindz:
I'm happy with the basic principle of the change,  a single method with "fallback" of simulation where there is no data, I don't have an objections of such a method per say.

What I'm not really clear on is **_why_** the change - what are we expecting to achieve by it? My original change was to remove a load of copy and paste code into a common method. Given we are still calling a single method from the effects and the implementation is basically the same.

If there is some subtle difference that means passing the simulationId is more performant or reduces code size, then that should be a change in its own right really, explained as the reasoning in the PR with evidence of the improvement. Without the context for the change I'm a little lost to understand how this change improve the code

blazoncek:
> a single method with "fallback" of simulation

That was the whole idea from the start as there is really no reason to have 2 functions performing the same.
It is ok with me to abandon this PR if there is no visible/perceivable benefit from it.

netmindz:
I don't feel it really helps improve the code readability and if it's not performance related then it's just decreasing code quality by blending expected operations with graceful fallback

wled00/FX.cpp
softhack007: Confirmed-its unrelated. 
The code tried to adjust this effect's frequency range for -S2 and -C3.  Which is not so bad, if -S2/-C3 use a different sampling rate in audioreactive. 

Reminds me that I want to clean up this log10() chaos in fx.cpp: 1. use logf() instead of log10f() 2. Use constants instead of magic numbers.
But that's a different PR.

softhack007:
I agree with @netmindz - the code is not time critical because  it's running _once_ per effect.

DedeHai:
@softhack007 
regarding AR on C3 & S2 (abusing this PR for an AR question):
**I see there is plans to reduce sampling rate for C3/S2, is there work in progress somewhere?**

I just ran a few tests on a C3 with AR and there is great potential:
- reducing to 16kHz AND reducing sample size to 256 brings the FFTtime down to 19ms
- FFT_SPEED_OVER_PRECISION and FFT_SQRT_APPROXIMATION make it slightly slower (maybe 0.5ms) so there is no point in using those
- Frequency range with 16kHz and 256 samples is 62.5Hz - 8kHz (vs. 43Hz - 11 kHz with 22kHz sampling and 512 samples)

I think that is a reasonable trade-off and works quite well. S2 should use the same as it also does not have an FPU.

**Are there any plans to support analog mic on S3, S2, C3?**
Especially the C3 would give users on a tight budget some options, or for projects where you want a lot of distributed cheap AR lights. I did not look at the code in detail yet but ADC sampling is quite easy to implement using DMA (in case you did not look into that already: here is some example code https://github.com/ClaudeMarais/ContinousAnalogRead_ESP32-C3/blob/main/ContinousAnalogRead_ESP32-C3.ino )
I am not (yet) familiar with the ESP DMA controller, but apparently it is quite flexible, I am quite sure it can do ring-buffer or dual buffer or one-shot, whichever works best for AR.  I did use DMA on a STM32: I assume ESP has the same basic functionalities.

softhack007:
> @softhack007
> regarding AR on C3 & S2 (abusing this PR for an AR question):
> I see there is plans to reduce sampling rate for C3/S2, is there work in progress somewhere?

Hi @DedeHai yes there is a lot in progress, [over at the MoonModules fork](). The code is somewhat cluttered, but now that license topics are solved, I'm planning to bring improvements back to upstream.

Your proposal to reduce sampling rates for C3 (and maybe for S2, too) makes sense. In fact S2 is still two times faster on FFT than C3, so its "edgy" but deducing sampling rate to 18khz should be enough for the chip.

Reducing FFT size is something I was also considering - but this reduces frequency resolution a lot, especially in the lower range. For -C3, it looks like another trick works well, while keeping the frequency resolution as it is: in MM we are skipping every second FFT run on C3. This has only minor impact on the result, but it allows C3 to take about 30-40ms for one FFT. Works well.

For the future, the plan is to replace ArduinoFFT with [esp-dsp](https://docs.espressif.com/projects/esp-dsp/en/latest/esp32/esp-dsp-benchmarks.html). This requires to move to arduino-esp32 version 2.0.x, but it will be a major step forward. It will also allow us to replace my hand-crafted IIR filters with standard biquad filters from esp-dsp. @troyhacks has made some experiments and it looks very promising. FFT on S2 and C3 will also be faster, even when using float with FPU.

softhack007:
About analog mic support for -S3/-S2/-C3:
For me this is very low priority, because we currently use the I2S driver and  making an "alternate" driver that's using continuous analog sampling mode means to dig into a completely different tech. I have no experience with the continuous ADC sampling mode.

Also I don't expect good results from esp ADC sampling, because the ADC unit is well-known for low accuracy. So the FFT spectrum will contain lots of artefacts and "ghost frequencies". Plus most analog mics are very sensitive to noise on the Vcc line -basicially any wifi access can cause a noise burst on analog reading. We have been through this in the old Atuline soundreactive - and honestly speaking, most user should really invest into a cheap I2S microphone (like INMP441) or even a cheaper PDM microphone - the headaches and troubles with analog in are really not worth it.

There are also ◇good line-in to I2S boards](https://mm.kno.wled.ge/soundreactive/Line-Input/) that we support - availeable for cheap. So line-in audio is no reason for analog imho.

Actually if you want to create a continuous ADC input driver, be my guest ;-)

DedeHai:
thanks, you make valid points.
how does the "skipping every other frame" on C3 affect FPS? I see you re-did the binning in MM, in AC even the lowest bins are summed, so switching to 256 samples and changing binning would not result in much resolution loss (which is what I was thinking). I am not sure which option is better though, skipping audio frames or losing resolution. I would not gear the C3 towards "highest audio quality" because if users want that, ESP32 or S3 are the only good options here. I do not have the gear nor the experience you guys have to test any difference. I only bring some theoretical background to the table (acoustics at university but that was quite some time ago).

Regarding analog mic: I am well aware of the down sides, I do precision ultra low noise analog electronics for a living, not audio though ;)
You also need to take into account that we have users where a 0.3$ analog mic vs a 1.5$ digital one makes a difference (had this exact example on discord), so yes, digital mic over analog anytime but I do not see that as a reason to drop support for analog. I may run a test when I get around to it to see if I can make the ADC-DMA sampling work on a C3 but I have a lot of other things on my list right now

softhack007:
Hi @DedeHai thanks for sharing your thoughts

> how does the "skipping every other frame" on C3 affect FPS?

In fact it does not affects FPS directly, because FFT is running in a separate task and effects do not wait for a new batch of results. Indirectly the CPU load for FFT is actually impacting effects, but I've never measured it in detail. I was more concerned to keep FFT within the time limits of 1-2 batches of samples, because lagging behind (and possible buffer saturation) is something that's very noticeable for user.

The FFT task is doing some post-processing and smoothing, so skipping every other FFTdid not lead to any jumping in effects.


>  I see you re-did the binning in MM, in AC even the lowest bins are summed, so switching to 256 samples and changing binning would not result in much resolution loss (which is what I was thinking)

The problem comes in the low frequencies - with 256 samples  you lose a lot of detail in the low "GEQ bands", so I tried to avoid that.

> I only bring some theoretical background to the table (acoustics at university but that was quite some time ago).

Me-too 😉 I've had acoustics as a secondary subject (3 semesters) when I studied computer science (called "Informatik" or "Kybernetik" in the 1990's). I think that's a great basis for understanding what's needed.

softhack007:
> * Frequency range with 16kHz and 256 samples is 62.5Hz - 8kHz (vs. 43Hz - 11 kHz with 22kHz sampling and 512 samples)

@DedeHai about this side-topic on making audioreactive faster on -C3 .... 

I had some time to let your idea "sink in" - you're right it's a good "quick win" improvement for -C3. The final solution of replacing ArduinoFFT  is still some time away. But changing FFT width from 512 to 256 could be done easily - just a few constants plus a different  mapping from FFT bins to fftResult[] channels. Maybe this "low quality" mode could even be an option (build flag) for -S2.

Its seems that people tend to ignore any warnings about -C3 being slow (too slow for audioreactive) - they just attach a mic and recognise their board is running hot while effects get slow. So having low resolution FFT might be an interesting way to make -C3 a bit faster on running effects.

If I find time this week, i can make a PR and would be happy to have your review and testing on -C3.

DedeHai:
I would be happy to test it. AR on C3 has it's right to exist IMHO: not every one uses it for party-displays where ms timing and best resolution matters. Some just want it to react to loudness, there are probably hundreds of examples where the quality of the signal is really not that important. 
What I would also like to check: is it better to have more base resolution (i.e. using every second sample at 512 FFT width) or having better reaction time but loose base resolution (i.e. reduce to 256 samples but process every frame). Or could even be ok to do both, reduce and only do every second frame. What I would like to test here is record some music and compare the FFT bins generated for a short sequence.

PR #4214, opened 2024-10-21
Fix alexa devices invisible/uncontrollable

Svennte:
Fix for LED and Scenes uncontrollable using Alexa. 
Weird behavior regarding to the device names and shared scenes fixed with this. 

Seen in issue https://github.com/Aircoookie/Espalexa/issues/228 and fixed from @ams-hh
Tested by myself and works just fine. Created second pull request here because the library seems to be a bit different from the official Espalexa repo.


DedeHai:
Thank you for this fix. How well tested is it?
Looks good to me and compiles to the same file sizes. 
Just one change request: remove the now unused code.

Svennte:
Just tested on Esp32 devboard. When I have time today I check with a wemos mini and remove the unused code then.
In the issue I mentioned the people also got it to work with this fix, but I dont now which board they used.

ams-hh:
As mentioned, the fix is not from me, but from PakoCosmos
[https://github.com/vintlabs/fauxmoESP/issues/259#issue-2576866208](url)
I only transplanted it to get a solution for my issue...
And as he mentioned (and me too) it is NOT well tested. Only confirmed working in
the own environment with the own generation of Alexa device.
(I do not use a comercial board. Only naked ESP32S3 on selfmade PCB)

softhack007:
> Seems like I did bullshit here, sorry for that.
> Will look at this this evening.

@Svennte please undo unrelated changes (newlines in platformIO.ini and FX_2Dfcn.ccp). After that, I think we can merge your PR.

softhack007:
@Svennte I've fixed unrelated changes by maintainer edit. Ready to merge.

Svennte:
@softhack007 Thank you, much appreciated!

PR #4217, opened 2024-10-22
Textarea fields in usermods

obar:
A textarea with a label ending in `>` will be presented as a textarea, allowing for multi-line text.

It's a small change, I think it opens up possibilities for usermods that need more than just a single line but don't need their own custom settings page.

### DIFF
--- a/wled00/data/settings_um.htm
+++ b/wled00/data/settings_um.htm
@@ -153,14 +153,19 @@
 					}
 					break;
 				default:
-					t = "text"; c = `value="${o}" style="width:250px;"`;
+				    if (f.substr(-1)===">") {
+						t = "textarea"; c = `style="width:300px;height:150px;"`;
+					} else {
+					    t = "text"; c = `value="${o}" style="width:250px;"`;
+					}
 					break;
 			}
 			urows += ` ${initCap(f)} `; //only show field (key is shown in grouping)
 			// https://stackoverflow.com/questions/11657123/posting-both-checked-and-unchecked-checkboxes
 			if (t=="checkbox") urows += `<input type="hidden" name="${k}:${f}${a?"[]":""}" value="false">`;
 			else if (!a)       urows += `<input type="hidden" name="${k}:${f}${a?"[]":""}" value="${t}">`;
-			urows += `<input type="${t==="int"?"number":t}" name="${k}:${f}${a?"[]":""}" ${c} oninput="check(this,'${k.substr(k.indexOf(":")+1)}')"><br>`;
+			if (t=="textarea") urows += `<br><textarea name="${k}:${f}${a?"[]":""}" ${c}>${o}</textarea><br>`;
+			else    		   urows += `<input type="${t==="int"?"number":t}" name="${k}:${f}${a?"[]":""}" ${c} oninput="check(this,'${k.substr(k.indexOf(":")+1)}')"><br>`;
 		}
 	}
 	function pinDropdowns() {
--- a/wled00/data/style.css
+++ b/wled00/data/style.css
@@ -67,7 +67,7 @@ button.sml {
 .warn {
   color: #fa0;
 }
-input {
+input, textarea {
   background: #333;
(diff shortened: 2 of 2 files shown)


blazoncek (changes requested):
Please respect indentation of existing code.

Other consideration is that `textarea` fields allow huge amounts of data to be entered by user. ESP is not well suited for such task and content may either be trimmed or rejected entirely. This may be baffling for users as it will work for some texts but not others.

I do not approve.

DedeHai:
I agree, this needs to be done properly i.e. limit to a size that does not overflow any buffers.
Other than that: am I right in assuming that this tries to solve a currently non-existing use case? If so adding code 'just for convenience' is not a good idea due to the stringent flash memory restrictions.

softhack007:
I think the use case question from @DedeHai is a very valid point.
We already have a type for "text" which is good for one-line text (e.g. MQTT topic name). With multi-line fields, we could multi-line input like you mom's postal address including phone number and birthday letter. 

a) it may create new issues, as the textblock must put into a buffer (stack?) for processing, it must be added to the global JSON object (heap), and written out to cfg.json while preserving line breaks and special characters including '"', '[' and ':'.

b) how would I create such a text area?

Today we typically do this in a ausermod:
```C++
    uint8_t soundSquelch = 8;
    void addToConfig(JsonObject& root) override
    {
      JsonObject top = root.createNestedObject(FPSTR(_name));
      JsonObject cfg = top.createNestedObject(FPSTR(_config));
      cfg[F("squelch")] = soundSquelch;
   }
   
      bool readFromConfig(JsonObject& root) override
    {
      JsonObject top = root[FPSTR(_name)];
      bool configComplete = !top.isNull();
      configComplete &= getJsonValue(top[FPSTR(_config)][F("squelch")], soundSquelch);
      return configComplete;
```

What's necessary to create a textarea? `String` is not enough, I assume?

obar:
Good point about indentation. There's a mix of tabs and spaces in that chunk of the text file and I should take a closer look at how it's done here, I just trusted my text editor to figure it out.

It's true that a lot of text could be entered in a textarea, but the same is true for `input type="text"` as well, both have a `maxlength` that can be specified but that's up to the browser to enforce. It's still a String, just one with `\n` included. There are already textareas in the stock UI for power users (like people activating usermods), when configuring presets. If a usermod would make use of this feature in a way that's potentially risky, I agree with the three of you: there should be a warning and maybe even guard rails.

My use case: I've got a [usermod I'd like to contribute](https://github.com/obar/WLED/tree/usermod-preset-array/usermods/preset_array) that uses a textarea for a reasonable string length, where the use of newlines to delimit makes it legible. You can see the usermod in action in #showcase on Discord, I just posted a clip a few minutes ago!

I messed up this pull request in other ways than the tabs regardless (ended up making a new branch on main and not 0_15), and I should also include the npm-generated files, as well as an addition to the Example usermod... maybe a new pull request is in order! Will resubmit soonish and I hope I can win you over on the UI benefit of a tastefully used textarea ;)

obar:
Didn't need to open a new pull request in the end, amended to fix the tabs/spaces issue @blazoncek pointed out, and put in an example as @softhack007 suggested. Looks like npm doesn't touch the stylesheet and the different branch doesn't matter in this case. Thanks all for the help and suggestions :)

DedeHai:
I do not see the benefit. While I like what you did in your usermod, would this be an acceptable workaround?

- add a config field to your UM with "number of groups" (or get the number of groups some other way)
- add a one line input field for each group (dynamically)

obar:
I believe the major concerns were addressed, and of course a usermod using this feature that encourages a ton of text should come with a warning. Anything blocking a merge? Here is my example usage, in a usermod I'd like to submit once this feature is in:

Screenshot 2024-12-03 at 17-47-17 Usermod Settings

rtist2g (approved):

rtist2g:
Sorry, I did this on accident.

PR #4221, opened 2024-10-23
Add initial support for SM16714P chip based LEDs (RGBW)

pmikka:
Add initial support for SM16714P chip based. See https://wled.discourse.group/t/sm16714-pixel-ic/9794 for details. This does not support additional 20-bit of current setting, for this most probably NeoPixelLibrary needs also to be adapted.

### DIFF
--- a/wled00/bus_manager.cpp
+++ b/wled00/bus_manager.cpp
@@ -399,6 +399,7 @@ std::vector<LEDType> BusDigital::getLEDTypes() {
     {TYPE_UCS8904,       "D",  PSTR("UCS8904 RGBW")},
     {TYPE_WS2805,        "D",  PSTR("WS2805 RGBCW")},
     {TYPE_SM16825,       "D",  PSTR("SM16825 RGBCW")},
+    {TYPE_SM16714,       "D",  PSTR("SM16714P or Inverted WS2814 RGBW")},
     {TYPE_WS2812_1CH_X3, "D",  PSTR("WS2811 White")},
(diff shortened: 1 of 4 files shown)


jayceegee87:
Hi all,

Just chiming in that these are increasingly being utilized in outdoor LED applications (in my case LIFX string lights) given the ability to calibrate all of the individual color settings. It would be great to add compatibility and I happen to have one extra string available to fiddle around with if you need someone to check/test something.

I also experienced the exact same artifacts as everyone else who has requested SM16714 support, but did identify an odd partial workaround at the expense of the overall brightness: https://www.reddit.com/r/WLED/s/krDfmbn2co.

bonfireva:
Govee clear string lights use this chip as well. Would be great to have WLED support.
IMG_0736

DedeHai:
@pmikka do you have the full datasheet? I could add support to my new WLEDpixelBus, it also supports suffix and automatic calculation of brightness current settings, but it has to be adapted for this chip.

jayceegee87:
Mikkas original 

> @pmikka do you have the full datasheet? I could add support to my new WLEDpixelBus, it also supports suffix and automatic calculation of brightness current settings, but it has to be adapted for this chip.

His original post in the wled discourse included screenshots from the full technical specs, though I'm not sure if it includes everything you need. @pmikka indicated that he was unable to attach the full pdf, but I'm not sure if that's a limitation of the forum or something requested by SunMoon. I think it's the former. 

The technical summary page is here. I'll email the contact there to see if they will send me the full expanded data sheet. 

https://www.chinaasic.com/websiteManage/product/20230523102223341.pdf

PR #4222, opened 2024-10-24
Esp8266 oappend fix

### LINKED ISSUE #4219: 8266 160MHz /settings/um crash/reboot loop
### What happened?

An otherwise stable 8266 running 0.15b6 at 160Mhz crashes and reboots while a web UI is displaying /settings/um . Confirmed by ALDIY @ Discord

Demo https://youtube.com/shorts/lTwzndDkhG0

### To Reproduce Bug

Install 0.15b6 160Mhz build on 8266
Enable AR in Usermods (not sure if this is required to reproduce)
Device goes into a reboot loop

### Expected Behavior

It shouldn’t crash :-)

### Install Method

Binary from WLED.me

### What version of WLED?

0.15.0-b6 (build 2410140)

ESP8266

### Relevant log/trace output

_No response_

### Anything else?

_No response_

### Code of Conduct

willmmiles:
The `oappend` shim logic introduced in #4152 also has issues with `SET_F()`, as it casts away the helper type needed for `print()` to identify PROGMEM strings.  Improve the shim logic to safely handle this case in two different ways:

1. Detect PROGMEM addresses and re-apply the helper type, allowing the correct overload of `print()` to be called;
2. Ask the platform to forgive us if we make a mistake, and handle it gracefully (if very slowly).

Fixes #4219.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -197,6 +197,7 @@ build_flags =
   ; decrease code cache size and increase IRAM to fit all pixel functions
   -D PIO_FRAMEWORK_ARDUINO_MMU_CACHE16_IRAM48 ;; in case of linker errors like "section `.text1' will not fit in region `iram1_0_seg'"
   ; -D PIO_FRAMEWORK_ARDUINO_MMU_CACHE16_IRAM48_SECHEAP_SHARED ;; (experimental) adds some extra heap, but may cause slowdown
+  -D NON32XFER_HANDLER ;; ask forgiveness for PROGMEM misuse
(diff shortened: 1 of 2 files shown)


blazoncek:
@dosipod can you test this as well?
IMO it is good.

dosipod:
We tested both nodemcuv2_160  and esp8266_2m_160  .In both AR worked as expected and  no more error or crash .

softhack007:
Good to merge 👍

PR #4223, opened 2024-10-24
Usermod F-string hygiene

willmmiles:
Another missing remnant of the end-oappend work: migrating usermods to use F() instead of F_STR() so as to pass the FlashStringHelper type along to the print() layer.  Usermods calling `oappend(F_STR("")` on ESP8266 will crash without the shim in #4222.  This could be considered an alternative fix: correct the source, rather than provide a workaround at a runtime cost.


blazoncek:
IMO both are good. The other one will prevent misuse in the future.

blazoncek (approved):
Has a go from my side.

PR #4229, opened 2024-10-26
specified required nodejs version

maxi4329:
Set the Node.js version in `package.json` to 20 or higher, according to the [wiki](https://kno.wled.ge/advanced/compiling-wled/#:~:text=Also%20make%20sure%20that%20you%20have%20Node.js%2020%20or%20higher%20installed.)


package.json
```diff
@@ -27,5 +27,8 @@
     "html-minifier-terser": "^7.2.0",
     "inliner": "^1.13.1",
     "nodemon": "^3.1.7"
+  },
+  "engines": {
+    "node": ">=21.0.0"
```
netmindz, line 32: Should this not be 20 then not 21?

netmindz (changes requested):

package.json
maxi4329: I was working with different folders and got a bit confused. I changed it to 20.

netmindz (approved):

w00000dy (approved):
Good idea

PR #4234, opened 2024-10-29
RF433 json usermod

kilrah:
Inspired by https://github.com/athom-tech/Sound-Reactive-WLED/tree/master/usermods/RF433
Modified to handle similar cases as other remote methods through json

Usermod for controlling WLED using a generic 433 / 315MHz remote and simple 3-pin receiver
See https://github.com/sui77/rc-switch/ for details

Build:
    - Enable by defining USERMOD_RF433 in my_config.h
    - Uncomment "sui77/rc-switch @ 2.6.4" in platformio.ini default env lib_deps section

Usage:
    - Connect receiver to an interrupt-capable pin
    - Set pin in Config->Usermods
    - Info pane will show the last received button code
    - See https://kno.wled.ge/interfaces/json-ir/json_infrared/ as reference to create a "remote433.json" file
      and upload it to the ESP in http://ip/edit, the key for the actions is the button number retrieved from above.


DedeHai:
This looks like a great addition. Did you test it to work on all ESP variants?

A few suggestions:
- provide a list of compatible pins for all ESPs in the readme
- even better: restrict pin selection to compatible pins
- provide an example json file

kilrah:
> Did you test it to work on all ESP variants?

Tested on esp32, c3 and esp8266 

> even better: restrict pin selection to compatible pins

How can that be done? Believe it's all of them anyway

> provide an example json file

No problem to just put one in the folder?

DedeHai:
not sure how to restrict pins but I think AR usermod does it, maybe look at others for examples but I have seen UMs where unusable pins are red. In the deep-sleep usermod I just create a list but there I do not use 'pin' because already used pins can be repurposed, so not going through pinmanager. 

yea, I would just put the example json in the usermod folder, easier to find than going through KB and download an IR remote file.

blazoncek:
FYI you cannot restrict GPIO in usermods. Not yet.
Existing restrictions come from used pins.

kilrah:
I looked around and at least for ESP32, S3 and 8266 all pins can be used as interrupts so there should not be any restriction.

kilrah:
There, hopefully that's fine... I changed the default_env in `platformio_override.sample.ini` because the existing one doesn't exist anymore and so just copying the whole sample would break.

usermods/usermod_v2_RF433/usermod_v2_RF433.h
```diff
@@ -0,0 +1,178 @@
...
+  {
+    char objKey[14];
+    bool parsed = false;
+
+    if (!requestJSONBufferLock(22)) return false;
+
+    sprintf_P(objKey, PSTR("\"%d\":"), button);
+
+    // attempt to read command from remote.json
+    readObjectFromFile(PSTR("/remote433.json"), objKey, pDoc);
```
netmindz, line 132: Would it make more sense to read the JSON file during your setup function rather than for every button press?

netmindz (changes requested):

usermods/usermod_v2_RF433/usermod_v2_RF433.h
kilrah: This is done the same as for IR and wizmote. If it wasn't there if you went to /edit and changed the file changes wouldn't be taken into account until a reboot? That wouldn't be user-friendly.

platformio_override.sample.ini
```diff
@@ -5,7 +5,7 @@
 # Please visit documentation: https://docs.platformio.org/page/projectconf.html
 
 [platformio]
-default_envs = WLED_tasmota_1M  # define as many as you need
+default_envs = WLED_generic8266_1M  # define as many as you need
```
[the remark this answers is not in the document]
kilrah: Why?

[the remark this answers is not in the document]
DedeHai: then it needs to be changed in a separate PR as it is not related to your UM.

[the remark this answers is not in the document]
kilrah: Done

usermods/usermod_v2_RF433/usermod_v2_RF433.h
DedeHai: I agree, this (reading the file upon receive) is the way it is done in all other json remote code. Also there is no restriction on creating say 50 different commands: that would hog a lot of RAM permanently.

DedeHai:
@kilrah can you update to latest (probably just need to up the UM number)? 
I would then merge this. 
Oh, and you may want to add the "wait for bus" glitch-fix (see https://github.com/Aircoookie/WLED/pull/4386)

kilrah:
will do

usermods/usermod_v2_RF433/usermod_v2_RF433.h
```diff
@@ -0,0 +1,183 @@
...
+  bool initDone = false;
+
+public:
+
+  void setup()
+  {
+    mySwitch.disableReceive();
+    if (modEnabled)
+    {
+      mySwitch.enableReceive(receivePin);
```
DedeHai, line 32: the receivePin is not allocated through the pin manager, meaning it can also be assigned to other functions.
@kilrah I don't think that is intentional?

kilrah: The pin selection uses it so it is blocked in the GUI, are we supposed to expect that not to work?

image

image

PR #4238, opened 2024-10-31
Add clang-format file

netmindz:
Rather than just describing what code style we look for,  it would be good to provide a clang-format file

To get us started I have created one based on a style detection tool and one of the current files

In order to verify if the file is right, we can run  clang-format -i wled00/*.cpp and if we have the right values then all the changes should match our expectations

We might not choose to mass reformat all existing code, but it serves as useful test to check that new code being written will indeed conform to our style

I very much welcome input from others, especially those with knowledge of how to write a clang-format file

## Summary by CodeRabbit

- **Chores**
  - Added a project-wide code-formatting configuration based on Google style guidelines.
  - Enabled consistent formatting while preserving comments, selected compact code constructs, and modern constructor syntax.
  - Updated repository settings so the formatting configuration is tracked and applied.
  - Added safeguards to exclude third-party and generated source code from automatic formatting.
  - Documented compatibility with clang-format 16 and newer.


Moustachauve (approved):

w00000dy:
I think that's a very good idea.
The question is, do we want a .clang-format file that is as close as possible to the current WLED code, or do we want a .clang-format file that also fixes some of the unattractive things in the WLED code?

netmindz:
That's where I've stumbled a bit to be honest @w00000dy 
If you use on of the automated tools to help build a clang file against current code then you get slightly different rules depending on which file you use as your reference

.clang-format
```diff
@@ -0,0 +1,23 @@
...
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: Always
+AlwaysBreakBeforeMultilineStrings: false
+BreakBeforeTernaryOperators: false
+BreakConstructorInitializersBeforeComma: true
+BreakStringLiterals: false
+ColumnLimit: 240
+ContinuationIndentWidth: 2
+IndentPPDirectives: BeforeHash
+KeepEmptyLinesAtTheStartOfBlocks: true
```
w00000dy, line 19: This option is **deprecated**. See: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#keepemptylinesatthestartofblocks

.clang-format
```diff
@@ -0,0 +1,23 @@
...
+MaxEmptyLinesToKeep: 2
```
w00000dy, line 20: Maybe we should change this so that the code doesn't take up so many lines?

.clang-format
```diff
@@ -0,0 +1,23 @@
...
+ReflowComments: false
```
w00000dy, line 21: Maybe we should at least set this to `IndentOnly` so that misaligned comments are formatted.

.clang-format
```diff
@@ -0,0 +1,23 @@
...
+SortIncludes: Never
```
w00000dy, line 22: Why not?

w00000dy (commented):
Since you wanted some feedback, I had a look at it. I've written what I think about a few options. The following options are suggestions for additions to the `.clang-format` file from my side:

- `InsertNewlineAtEOF: true`
- `AllowShortBlocksOnASingleLine: true`

`AllowShortBlocksOnASingleLine` allows this:
```c++
while (true) {}
```
and
```c++
while (true) { continue; }
```

netmindz:
Thank you @w00000dy for your comments,  this is my first time using clang and I'm not a C++ developer I am very grateful for input from those with experience.

I'll have a look through your individual comments when I get time to do so

w00000dy:
Please don't close this PR. I would like the idea of this PR

softhack007:
@coderabbit please re-review.

Additionally, summarize the style changes suggested in the review comments, together with  a short description of what each change achieves.

.clang-format
```diff
@@ -0,0 +1,23 @@
+---
+BasedOnStyle: Google
+AlignAfterOpenBracket: DontAlign
+AlignConsecutiveDeclarations: Consecutive
```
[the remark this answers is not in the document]
softhack007: I agree - the extra spaces in `int    tquot` - to visually align it with the definition of `float fmod_t` - are confusing.

netmindz:
Pushed an update that trims this down to a minimal, empirically-tuned rule set — 24 keys → 12 — based on measuring the formatted diff against the real `wled00` source rather than debating options in the abstract. Summary for reviewers:

## Method

For every contested option (from @w00000dy's review and CodeRabbit's passes), I checked out upstream `main`'s `wled00/*.cpp`/`*.h` (68 files) and measured how many lines `clang-format` would change for each candidate value. A lot of the original file's overrides turned out to be **actively worse than just leaving Google's default alone** — e.g. `IndentPPDirectives: BeforeHash` cost ~650 extra changed lines vs. Google's default `None`; `SpacesBeforeTrailingComments: 1` cost ~250 vs. the default `2`. Those are simply dropped now rather than replaced — the fix was mostly subtraction, not addition.

Since this PR isn't proposing a mass reformat (existing style stays as-is; new/touched code converges going forward), I also spot-checked the finalists with `clang-format -lines=` range formatting (the actual usage mode) rather than trusting whole-file diff counts alone.

## The standout keep: `AlignTrailingComments: Leave`

This one isn't about aggregate line count — it's about what happens when a contributor edits *one* line inside an existing comment-aligned block:

```c++
// before, only the "second" line gets touched:
int a = 1;      // first
int bb = 2;     // second - EDIT THIS LINE
int ccc = 3;    // third

// AlignTrailingComments: Never (or false) collapses just the touched line's spacing:
int a = 1;      // first
int bb = 2;  // second - EDIT THIS LINE      <- now misaligned with its neighbours
int ccc = 3;    // third

// AlignTrailingComments: Leave preserves it untouched:
int a = 1;      // first
int bb = 2;     // second - EDIT THIS LINE   <- unchanged
int ccc = 3;    // third
```

`Leave` requires clang-format 16+ (its `Kind`-based syntax). Given that trade-off, this needs clang-format 16+ as the floor for the whole file — flagging this explicitly as a compatibility decision for maintainers to weigh in on, since anyone on an older clang-format gets a hard parse error (no formatting at all) rather than a silent difference.

## Correcting a CodeRabbit suggestion

CodeRabbit's fix for the `KeepEmptyLinesAtTheStartOfBlocks` deprecation warning was `KeepEmptyLinesAtTopOfFunctions`. I tested it directly against clang-format 14, 16, 18, and 22 — it's an **unknown key that hard-errors on all four**. The modern nested replacement (`KeepEmptyLines: {AtStartOfBlock: true}`) also errors on 16 and 18 (only became valid later). The old deprecated scalar spelling parses cleanly across the entire 14–22 range, so it's kept intentionally rather than "fixed" — noting this so it doesn't get re-flagged.

`BreakConstructorInitializersBeforeComma` → `BreakConstructorInitializers: BeforeComma` *is* a safe rename (verified on all four versions), so that one's updated.

## `AllowShortIfStatementsOnASingleLine: AllIfsAndElse`

Un-deprecates the old `Always` value (no longer a valid enum member) and picks the modern spelling closest to what it aliased to. Verified with range-formatting that it can't reach into an untouched sibling `if`/`else` branch — clang-format's line-range restriction is a hard boundary, so this only affects the branch actually being edited.

## One caveat worth knowing: `IndentPPDirectives: None`

Existing code is mostly flush-left on nested `#if` directives, so `None` (Google's default) matches more often than `BeforeHash` did. But it's not free: touching a line inside one of the rarer *indented* nested blocks (e.g. the `FXparticleSystem.h` guard in `FX.cpp`) will flush just that touched directive left against its still-indented siblings. The reverse problem exists with `BeforeHash` against the (much more common) flush-left blocks, so this is a net improvement, not a strict win — just flagging it so it's not a surprise later.

## New file: `wled00/src/.clang-format`

`wled00/src/` holds vendored third-party code (`dependencies/`: json, fastled_slim, espalexa, e131, dmx, ws2812fx, etc.) and generated font headers (`font/`), none of which follows WLED's style. Child `.clang-format` files fully replace the parent's config for their subtree (no merging), so a `DisableFormat: true` there stops the root config from reaching in and reformatting upstream/generated code.

## Net result

14,036 lines would change if the whole `wled00` were reformatted with the final config, vs. 17,996 with the file as it stood before this update, and 22,860 for plain `BasedOnStyle: Google` — but since this PR doesn't reformat anything, the more relevant framing is that the config now sits measurably closer to existing convention with fewer, better-justified overrides.

PR #4244, opened 2024-11-01
Improved framerate control code - strip.show(), strip.service()

softhack007:
Reworked framerate control loop, to allow more fluent animations and support higher framerates

#### the problem (in a nutshell) 
image

#### solution
* separated fps calculation (strip.show) from framerate control (strip.service)
* improved condition for early exit in strip.service
* made MIN_SHOW_DELAY depend on target fps
* strip.show now considers the complete time of effect calculation + show; old code wrongly used the time between completion of last show and start of next effect drawing, causing unexpected slowdown and sometimes leading to stuttering effects.
* added "unlimited FPS mode" for testing (use Target FPS = 120)
* increased warning limits for "slow strip" and "slow effects" (WLED_DEBUG)

As 8266 and esp32-c3 might suffer from higher CPU load, these chips use the previous MIN_SHOW_DELAY logic that causes more idle loops.

#### Comparison, Using a 64x64 "network DDP" output (esp32_wrover)
* Akemi : old code 38 fps, new code 53 fps
* Black Hole: old code 40 fps, new code 54 fps
* Lissajous: old code 37 fps, new code 51 fps

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -48,8 +48,8 @@
 #define FRAMETIME        strip.getFrameTime()
 #if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32C3) && !defined(CONFIG_IDF_TARGET_ESP32S2)
   #define MIN_FRAME_DELAY  2                                              // minimum wait between repaints, to keep other functions like WiFi alive 
-#elif defined(CONFIG_IDF_TARGET_ESP32S2)
-  #define MIN_FRAME_DELAY  4                                              // S2 is slower than normal esp32, and only has one core
+#elif defined(CONFIG_IDF_TARGET_ESP32S2) || defined(CONFIG_IDF_TARGET_ESP32C3)
+  #define MIN_FRAME_DELAY  3                                              // S2/C3 are slower than normal esp32, and only have one core
 #else
   #define MIN_FRAME_DELAY  8                                              // 8266 legacy MIN_SHOW_DELAY
 #endif
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -44,20 +44,15 @@
 
 /* Not used in all effects yet */
 #define WLED_FPS         42
+#define FRAMETIME_FIXED  (1000/WLED_FPS)
 #define FRAMETIME        strip.getFrameTime()
 #if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32C3)    // all ESP32 except -C3(very slow, untested)
-  #define FRAMETIME_FIXED  (strip.getFrameTime() < 10 ? 12 : 24)            // allow faster FRAMETIME_FIXED when target FPS >= 100  
   #define MIN_SHOW_DELAY   (max(2, (_frametime*5)/8))                       // supports higher framerates and better animation control -- 5/8 = 62%
   // used to initialize for strip attributes:
-  #define WLED_FPS_SLOW    42
-  #define FRAMETIME_FIXED_SLOW  (24)                                        // 1000/42 = 24ms
 #else
-  #define FRAMETIME_FIXED  (1000/WLED_FPS)
   #define MIN_SHOW_DELAY   (_frametime < 16 ? 8 : 15)                       // legacy MIN_SHOW_DELAY - creates more idle loops, but reduces framerates
-  #define WLED_FPS_SLOW     WLED_FPS
-  #define FRAMETIME_FIXED_SLOW FRAMETIME_FIXED
 #endif
-#define FPS_UNLIMITED    119
+#define FPS_UNLIMITED    120
 
 /* each segment uses 82 bytes of SRAM memory, so if you're application fails because of
   insufficient memory, decreasing MAX_NUM_SEGMENTS may help */
@@ -737,8 +732,8 @@ class WS2812FX {  // 96 bytes
       _length(DEFAULT_LED_COUNT),
       _brightness(DEFAULT_BRIGHTNESS),
       _transitionDur(750),
-      _targetFps(WLED_FPS_SLOW),
-      _frametime(FRAMETIME_FIXED_SLOW),
+      _targetFps(WLED_FPS),
+      _frametime(FRAMETIME_FIXED),
       _cumulativeFps(2),
       _isServicing(false),
       _isOffRefreshRequired(false),
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -44,8 +44,20 @@
 
 /* Not used in all effects yet */
 #define WLED_FPS         42
-#define FRAMETIME_FIXED  (1000/WLED_FPS)
 #define FRAMETIME        strip.getFrameTime()
+#if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32C3)    // all ESP32 except -C3(very slow, untested)
+  #define FRAMETIME_FIXED  (strip.getFrameTime() < 10 ? 12 : 24)            // allow faster FRAMETIME_FIXED when target FPS >= 100  
+  #define MIN_SHOW_DELAY   (max(2, (_frametime*5)/8))                       // supports higher framerates and better animation control -- 5/8 = 62%
+  // used to initialize for strip attributes:
+  #define WLED_FPS_SLOW    42
+  #define FRAMETIME_FIXED_SLOW  (24)                                        // 1000/42 = 24ms
+#else
+  #define FRAMETIME_FIXED  (1000/WLED_FPS)
+  #define MIN_SHOW_DELAY   (_frametime < 16 ? 8 : 15)                       // legacy MIN_SHOW_DELAY - creates more idle loops, but reduces framerates
+  #define WLED_FPS_SLOW     WLED_FPS
+  #define FRAMETIME_FIXED_SLOW FRAMETIME_FIXED
+#endif
+#define FPS_UNLIMITED    119
 
 /* each segment uses 82 bytes of SRAM memory, so if you're application fails because of
   insufficient memory, decreasing MAX_NUM_SEGMENTS may help */
@@ -67,8 +79,6 @@
 /* How much data bytes each segment should max allocate to leave enough space for other segments,
   assuming each segment uses the same amount of data. 256 for ESP8266, 640 for ESP32. */
 #define FAIR_DATA_PER_SEG (MAX_SEGMENT_DATA / strip.getMaxSegments())
-
-#define MIN_SHOW_DELAY   (_frametime < 16 ? 8 : 15)
 
 #define NUM_COLORS       3 /* number of colors per segment */
 #define SEGMENT          strip._segments[strip.getCurrSegmentId()]
@@ -727,8 +737,8 @@ class WS2812FX {  // 96 bytes
       _length(DEFAULT_LED_COUNT),
       _brightness(DEFAULT_BRIGHTNESS),
       _transitionDur(750),
-      _targetFps(WLED_FPS),
-      _frametime(FRAMETIME_FIXED),
+      _targetFps(WLED_FPS_SLOW),
+      _frametime(FRAMETIME_FIXED_SLOW),
       _cumulativeFps(2),
       _isServicing(false),
       _isOffRefreshRequired(false),
@@ -739,6 +749,7 @@ class WS2812FX {  // 96 bytes
       customMappingTable(nullptr),
       customMappingSize(0),
       _lastShow(0),
+      _lastServiceShow(0),
       _segment_index(0),
       _mainSegment(0)
     {
@@ -949,6 +960,7 @@ class WS2812FX {  // 96 bytes
     uint16_t  customMappingSize;
 
     unsigned long _lastShow;
+    unsigned long _lastServiceShow;
 
     uint8_t _segment_index;
     uint8_t _mainSegment;
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -47,8 +47,7 @@
 #define FRAMETIME_FIXED  (1000/WLED_FPS)
 #define FRAMETIME        strip.getFrameTime()
 #if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32C3)    // all ESP32 except -C3(very slow, untested)
-  #define MIN_SHOW_DELAY   (max(2, (_frametime*5)/8))                       // supports higher framerates and better animation control -- 5/8 = 62%
-  // used to initialize for strip attributes:
+  #define MIN_SHOW_DELAY   3                                              // supports higher framerates and better animation control
 #else
   #define MIN_SHOW_DELAY   (_frametime < 16 ? 8 : 15)                       // legacy MIN_SHOW_DELAY - creates more idle loops, but reduces framerates
 #endif
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -46,10 +46,12 @@
 #define WLED_FPS         42
 #define FRAMETIME_FIXED  (1000/WLED_FPS)
 #define FRAMETIME        strip.getFrameTime()
-#if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32C3)    // all ESP32 except -C3(very slow, untested)
-  #define MIN_SHOW_DELAY   3                                              // supports higher framerates and better animation control
+#if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32C3) && !defined(CONFIG_IDF_TARGET_ESP32S2)
+  #define MIN_FRAME_DELAY  2                                              // minimum wait between repaints, to keep other functions like WiFi alive 
+#elif defined(CONFIG_IDF_TARGET_ESP32S2)
+  #define MIN_FRAME_DELAY  4                                              // S2 is slower than normal esp32, and only has one core
 #else
-  #define MIN_SHOW_DELAY   (_frametime < 16 ? 8 : 15)                       // legacy MIN_SHOW_DELAY - creates more idle loops, but reduces framerates
+  #define MIN_FRAME_DELAY  8                                              // 8266 legacy MIN_SHOW_DELAY
 #endif
 #define FPS_UNLIMITED    0
 
@@ -842,7 +844,7 @@ class WS2812FX {  // 96 bytes
       getMappedPixelIndex(uint16_t index) const;
 
     inline uint16_t getFrameTime() const    { return _frametime; }        // returns amount of time a frame should take (in ms)
-    inline uint16_t getMinShowDelay() const { return MIN_SHOW_DELAY; }    // returns minimum amount of time strip.service() can be delayed (constant)
+    inline uint16_t getMinShowDelay() const { return MIN_FRAME_DELAY; }   // returns minimum amount of time strip.service() can be delayed (constant)
     inline uint16_t getLength() const       { return _length; }           // returns actual amount of LEDs on a strip (2D matrix may have less LEDs than W*H)
     inline uint16_t getTransition() const   { return _transitionDur; }    // returns currently set transition time (in ms)
 
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -52,7 +52,7 @@
 #else
   #define MIN_SHOW_DELAY   (_frametime < 16 ? 8 : 15)                       // legacy MIN_SHOW_DELAY - creates more idle loops, but reduces framerates
 #endif
-#define FPS_UNLIMITED    120
+#define FPS_UNLIMITED    0
 
 /* each segment uses 82 bytes of SRAM memory, so if you're application fails because of
   insufficient memory, decreasing MAX_NUM_SEGMENTS may help */
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1312,20 +1312,15 @@ void WS2812FX::service() {
   if (_suspend) return;
   unsigned long elapsed = nowUp - _lastServiceShow;
 
-  #if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32C3)
-  if (elapsed < 2) return;                                                       // keep wifi alive
-  if ( !_triggered && (_targetFps != FPS_UNLIMITED)) {
-    if (elapsed < MIN_SHOW_DELAY) return;                                        // WLEDMM too early for service
+  if (elapsed < 2) return;                                                       // keep wifi alive - no matter if triggered or unlimited
+  if ( !_triggered && (_targetFps != FPS_UNLIMITED)) {                           // unlimited mode = no frametime
+    if (elapsed < _frametime) return;                                            // too early for service
   }
-  #else  // legacy
-  if (nowUp - _lastShow < MIN_SHOW_DELAY) return;
-  #endif
 
   bool doShow = false;
 
   _isServicing = true;
   _segment_index = 0;
-  unsigned speedLimit = (_targetFps != FPS_UNLIMITED) ? (0.85f * FRAMETIME) : 1;      // lower limit for effect frametime
 
   for (segment &seg : _segments) {
     if (_suspend) return; // immediately stop processing segments if suspend requested during service()
@@ -1362,7 +1357,6 @@ void WS2812FX::service() {
         // would need to be allocated for each effect and then blended together for each pixel.
         [[maybe_unused]] uint8_t tmpMode = seg.currentMode();  // this will return old mode while in transition
         frameDelay = (*_mode[seg.mode])();         // run new/current mode
-        if (frameDelay < speedLimit) frameDelay = FRAMETIME;                    // limit effects that want to go faster than target FPS
 #ifndef WLED_DISABLE_MODE_BLEND
         if (modeBlending && seg.mode != tmpMode) {
           Segment::tmpsegd_t _tmpSegData;
@@ -1375,6 +1369,7 @@ void WS2812FX::service() {
           Segment::modeBlend(false);          // unset semaphore
         }
 #endif
+        frameDelay = max(frameDelay, unsigned(MIN_SHOW_DELAY));                     // limit effects that want to go faster than target FPS
         seg.call++;
         if (seg.isInTransition() && frameDelay > FRAMETIME) frameDelay = FRAMETIME; // force faster updates during transition
         BusManager::setSegmentCCT(oldCCT); // restore old CCT for ABL adjustments
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1312,7 +1312,7 @@ void WS2812FX::service() {
   if (_suspend) return;
   unsigned long elapsed = nowUp - _lastServiceShow;
 
-  if (elapsed < 2) return;                                                       // keep wifi alive - no matter if triggered or unlimited
+  if (elapsed <= MIN_FRAME_DELAY) return;                                        // keep wifi alive - no matter if triggered or unlimited
   if ( !_triggered && (_targetFps != FPS_UNLIMITED)) {                           // unlimited mode = no frametime
     if (elapsed < _frametime) return;                                            // too early for service
   }
@@ -1369,7 +1369,6 @@ void WS2812FX::service() {
           Segment::modeBlend(false);          // unset semaphore
         }
 #endif
-        frameDelay = max(frameDelay, unsigned(MIN_SHOW_DELAY));                     // limit effects that want to go faster than target FPS
         seg.call++;
         if (seg.isInTransition() && frameDelay > FRAMETIME) frameDelay = FRAMETIME; // force faster updates during transition
         BusManager::setSegmentCCT(oldCCT); // restore old CCT for ABL adjustments
@@ -1390,7 +1389,7 @@ void WS2812FX::service() {
     yield();
     Segment::handleRandomPalette(); // slowly transition random palette; move it into for loop when each segment has individual random palette
     show();
-    _lastServiceShow = nowUp; // correct timestamp, for better FPS control
+    _lastServiceShow = nowUp; // update timestamp, for precise FPS control
   }
   #ifdef WLED_DEBUG
   if ((_targetFps != FPS_UNLIMITED) && (millis() - nowUp > _frametime)) DEBUG_PRINTF_P(PSTR("Slow strip %u/%d.\n"), (unsigned)(millis()-nowUp), (int)_frametime);
@@ -1445,9 +1444,9 @@ uint16_t WS2812FX::getFps() const {
 }
 
 void WS2812FX::setTargetFps(uint8_t fps) {
-  if (fps <= 120) _targetFps = fps;
+  if (fps <= 250) _targetFps = fps;
   if (_targetFps > 0) _frametime = 1000 / _targetFps;
-  else _frametime = 3;     // unlimited mode
+  else _frametime = MIN_FRAME_DELAY;     // unlimited mode
 }
 
 void WS2812FX::setMode(uint8_t segid, uint8_t m) {
@@ -1495,7 +1494,7 @@ void WS2812FX::setBrightness(uint8_t b, bool direct) {
   BusManager::setBrightness(b);
   if (!direct) {
     unsigned long t = millis();
-    if (_segments[0].next_time > t + 22 && t - _lastShow > MIN_SHOW_DELAY) trigger(); //apply brightness change immediately if no refresh soon
+    if (_segments[0].next_time > t + 22 && t - _lastShow > MIN_FRAME_DELAY) trigger(); //apply brightness change immediately if no refresh soon
   }
 }
 
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1309,11 +1309,23 @@ void WS2812FX::finalizeInit() {
 void WS2812FX::service() {
   unsigned long nowUp = millis(); // Be aware, millis() rolls over every 49 days
   now = nowUp + timebase;
-  if (nowUp - _lastShow < MIN_SHOW_DELAY || _suspend) return;
+  if (_suspend) return;
+  unsigned long elapsed = nowUp - _lastServiceShow;
+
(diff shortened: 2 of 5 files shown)


dosipod:
The below info is from a 32x32 matrix on 4 output  with target FPS set to 120, I only tested  2d effects and  I used ws but also verified some from the  info page (give a small margin for errors  due to ws which I kept unedited )   . Visually all effects seems faster with the new code ( too fast on smaller 32x16) . Some of the effects are very hard to notice a difference and might be more apparent on larger matrix then 32x32 but for sure it is better now . 
[Nov_2_2024_FPS_test.csv](https://github.com/user-attachments/files/17606251/Nov_2_2024_FPS_test.csv)
FPS_TEST

wled00/FX_fcn.cpp
```diff
@@ -1438,6 +1453,8 @@ uint16_t WS2812FX::getFps() const {
 void WS2812FX::setTargetFps(uint8_t fps) {
   if (fps > 0 && fps <= 120) _targetFps = fps;
   _frametime = 1000 / _targetFps;
+  if (_frametime < 1) _frametime = 1;           // better safe than sorry
```
blazoncek, line 1456: These two lines are completely unnecessary. `_targetFps` can never be less than 1 or more than 120. The first line makes sure of that.
Hence `_frametime` can only be between 1000 and 8 respectively (usually 23 (1000/42)), which is wrong as it is *more* than that due to integer math. Perhaps adding 1 to `_frametime` is a better option.

wled00/FX.h
```diff
@@ -44,8 +44,20 @@
 
 /* Not used in all effects yet */
 #define WLED_FPS         42
-#define FRAMETIME_FIXED  (1000/WLED_FPS)
 #define FRAMETIME        strip.getFrameTime()
+#if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32C3)    // all ESP32 except -C3(very slow, untested)
+  #define FRAMETIME_FIXED  (strip.getFrameTime() < 10 ? 12 : 24)            // allow faster FRAMETIME_FIXED when target FPS >= 100
```
blazoncek, line 49: This counters the **fixed** logic. If it is fixed than it should not vary.

wled00/FX.h
```diff
@@ -44,8 +44,20 @@
 
 /* Not used in all effects yet */
 #define WLED_FPS         42
-#define FRAMETIME_FIXED  (1000/WLED_FPS)
 #define FRAMETIME        strip.getFrameTime()
+#if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32C3)    // all ESP32 except -C3(very slow, untested)
+  #define FRAMETIME_FIXED  (strip.getFrameTime() < 10 ? 12 : 24)            // allow faster FRAMETIME_FIXED when target FPS >= 100  
+  #define MIN_SHOW_DELAY   (max(2, (_frametime*5)/8))                       // supports higher framerates and better animation control -- 5/8 = 62%
```
blazoncek, line 50: I do not know why @Aircoookie introduced `MIN_SHOW_DELAY` but by examining the code IMO it is intended to postpone `strip.service()` if minimum time has not yet passed.
I'm not sure if it is ok to be variable. It may not give other tasks enough time.

wled00/FX.h
```diff
@@ -727,8 +737,8 @@ class WS2812FX {  // 96 bytes
       _length(DEFAULT_LED_COUNT),
       _brightness(DEFAULT_BRIGHTNESS),
       _transitionDur(750),
-      _targetFps(WLED_FPS),
-      _frametime(FRAMETIME_FIXED),
+      _targetFps(WLED_FPS_SLOW),
```
blazoncek, line 740: What's wrong with original values?

wled00/FX_fcn.cpp
```diff
@@ -1326,10 +1338,10 @@ void WS2812FX::service() {
     if (!seg.isActive()) continue;
 
     // last condition ensures all solid segments are updated at the same time
-    if (nowUp > seg.next_time || _triggered || (doShow && seg.mode == FX_MODE_STATIC))
+    if (nowUp >= seg.next_time || _triggered || (doShow && seg.mode == FX_MODE_STATIC))
     {
       doShow = true;
-      unsigned delay = FRAMETIME;
+      unsigned frameDelay = FRAMETIME;
```
blazoncek, line 1344: Changing variable name just makes change harder to follow.

wled00/FX_fcn.cpp
```diff
@@ -1376,15 +1389,16 @@ void WS2812FX::service() {
   _triggered = false;
 
   #ifdef WLED_DEBUG
-  if (millis() - nowUp > _frametime) DEBUG_PRINTF_P(PSTR("Slow effects %u/%d.\n"), (unsigned)(millis()-nowUp), (int)_frametime);
+  if (millis() - nowUp > _frametime*2) DEBUG_PRINTF_P(PSTR("Slow effects %u/%d.\n"), (unsigned)(millis()-nowUp), (int)_frametime);
```
blazoncek, line 1392: I do not agree with this.
Reasoning: If effect calculations took more than a single frame, the effects are slow. Same below.

wled00/FX_fcn.cpp
```diff
@@ -1404,18 +1418,19 @@ void WS2812FX::show() {
...
   // See https://github.com/Makuna/NeoPixelBus/wiki/ESP32-NeoMethods#neoesp32rmt-methods
   BusManager::show();
 
-  unsigned long showNow = millis();
   size_t diff = showNow - _lastShow;
   size_t fpsCurr = 200;
   if (diff > 0) fpsCurr = 1000 / diff;
   _cumulativeFps = (3 * _cumulativeFps + fpsCurr +2) >> 2;   // "+2" for proper rounding (2/4 = 0.5)
   _lastShow = showNow;
+  _lastServiceShow = showNow;
```
blazoncek, line 1433: Unnecessary as it is overwritten in `service()`.

wled00/FX_fcn.cpp
```diff
@@ -1376,15 +1389,16 @@ void WS2812FX::service() {
...
 
   #ifdef WLED_DEBUG
-  if (millis() - nowUp > _frametime) DEBUG_PRINTF_P(PSTR("Slow effects %u/%d.\n"), (unsigned)(millis()-nowUp), (int)_frametime);
+  if (millis() - nowUp > _frametime*2) DEBUG_PRINTF_P(PSTR("Slow effects %u/%d.\n"), (unsigned)(millis()-nowUp), (int)_frametime);
   #endif
   if (doShow) {
     yield();
     Segment::handleRandomPalette(); // slowly transition random palette; move it into for loop when each segment has individual random palette
     show();
+    _lastServiceShow = nowUp; // correct timestamp, for better FPS control
```
blazoncek, line 1398: This overwrites variable set in `show()`.

wled00/FX_fcn.cpp
```diff
@@ -1349,7 +1361,8 @@ void WS2812FX::service() {
         // overwritten by later effect. To enable seamless blending for every effect, additional LED buffer
         // would need to be allocated for each effect and then blended together for each pixel.
         [[maybe_unused]] uint8_t tmpMode = seg.currentMode();  // this will return old mode while in transition
-        delay = (*_mode[seg.mode])();         // run new/current mode
+        frameDelay = (*_mode[seg.mode])();         // run new/current mode
+        if (frameDelay < speedLimit) frameDelay = FRAMETIME;                    // limit effects that want to go faster than target FPS
```
blazoncek, line 1365: This should involve `_frametime` not _speedLimit_.
`_frametime` reflects desired FPS (rounded).

blazoncek (changes requested):
I did not intend to review but changed my mind as I think there are flaws that need addressing.
I do not know the reasons for `MIN_SHOW_DELAY` so that must be explored separately.

wled00/FX_fcn.cpp
softhack007: This was intentionally - the name "delay" is a function, and I found it's better to avoid having a local with the same name.

wled00/FX_fcn.cpp
softhack007: Not always. With network "override" strip.show() can get called directly, calling "service()" much later due to strip.trigger(). So this was a backup solution when strip.service is not involved.

wled00/FX_fcn.cpp
softhack007: This is intentionally.

wled00/FX_fcn.cpp
softhack007: speed limit is actually 85% of _frametime. I've added it to prevent that effects that return a custom FRAMETIME exceed user defined target FPS. 85% to have some margin for "acdeptable" hoverspeed that can be handled.

softhack007:
@blazoncek thanks for your review 👍 
I'll think about your comments and will consider siwhat can be improved.

The whole logic was quite hard to "crack", so happy to get your thoughts 😃

wled00/FX.h
softhack007: That's the reason I kept the old MIN_SHOW_DELAY for 8266 and for -C3.

Making the delay variable (depending on target FPS) works on esp32 - its tested in the MoonModules  forks and we have a similar variable delay for more than a year already.

wled00/FX.h
softhack007: Good point, will double check

wled00/FX_fcn.cpp
blazoncek: Having network data override built in effect calculation should have no effect on FPS. As in such case FPS is determined by the caller.

If on the other hand you want to get actual FPS LEDs are updating, then only use one in `show()`.

softhack007: > If on the other hand you want to get actual FPS LEDs are updating, then only use one in `show()`.

That's exactly what I didn't want 😉 
I'll remove this assignment so that only service() will update `_lastServiceShow`.

wled00/FX_fcn.cpp
softhack007: The problem I wanted to solve is - when (example) you've updated drawLine and want to test performance, i'd say use "unlimited fps mode". But with debug enabled you'll get spammed on serial with "slow strip/slow effect" messages, so measurement get meaningless.

Another possibility would be to suppress the debug message by `if (getTagetFps() < FPS_UNLIMTED) ...`

wled00/FX_fcn.cpp
softhack007: I'll remove line 1456. The line below that is actually necessary for FPS_UNLIMITED test mode.

wled00/FX_fcn.cpp
willmmiles: Avoiding name collisions is a good idea!  I wish it was separate commit, though - it makes it easier on us reviewers and you can put the rationale in the commit message so we know what it's for.

wled00/FX.h
willmmiles: I'm somewhat leery of this change, too - there are a couple of FX that use this define, changing it will change their timing.  (Though probably the FX should be updated to call the strip method instead).

I'd also note that you do still have a fixed "minimum show delay" of 2 in the new ESP32 logic, expressly tested for on https://github.com/Aircoookie/WLED/blob/451cd4c74aff2faff9ab11f746f9e454ab10bd2e/wled00/FX_fcn.cpp#L1316 .

wled00/FX_fcn.cpp
willmmiles: +1 for the explicit test.

wled00/FX.h
softhack007: > This counters the **fixed** logic. If it is fixed than it should not vary.

I agree - it's only used 3 times in fx.cpp, and it seems to be more like a "reference time" for speedup/slowdown. I'll change it and keep the original fixed value.

wled00/FX.h
```diff
@@ -44,8 +44,20 @@
...
+  // used to initialize for strip attributes:
+  #define WLED_FPS_SLOW    42
+  #define FRAMETIME_FIXED_SLOW  (24)                                        // 1000/42 = 24ms
+#else
+  #define FRAMETIME_FIXED  (1000/WLED_FPS)
+  #define MIN_SHOW_DELAY   (_frametime < 16 ? 8 : 15)                       // legacy MIN_SHOW_DELAY - creates more idle loops, but reduces framerates
+  #define WLED_FPS_SLOW     WLED_FPS
+  #define FRAMETIME_FIXED_SLOW FRAMETIME_FIXED
+#endif
+#define FPS_UNLIMITED    119
```
willmmiles, line 60: I'd recommend using '0' for the "unlimited" mode.  0 happens to be a magic number that can be tested for with one fewer instruction everywhere we need to, and is an otherwise "meaningless" value for an FPS limit.

softhack007: Good point - actually it would need a small change in the UI, but I like the idea of saying "0 fps = unlimited".

softhack007: I'll double-check if `getTargetFPS() == 0` can cause problems like div/0 anywhere in the code.

wled00/FX_fcn.cpp
softhack007: Yeah, I'll make a separate commit in 0_15 only for the renaming. Then it should automatically  disappear from changes listed in this PR .

willmmiles:
Overall comment is that I think there's a number of otherwise independent changes all glommed together here, and it does make it a little tough to review all at once.
- delay -> frameDelay
- Fixing the FPS calculation with _lastServiceShow
- Adding an "unlimited FPS" feature
- Changes to the timing parameters for fast chips

I'm not super thrilled with the conditional logic for "fast enough" chips -- it feels to me like there's a generally safe algorithm somewhere in here, struggling to get out.  In some ways, high FPS is perfectly fine for fast FX even on slow chips, since they'll be servicing the wifi stack in between every one of those fast frames.  The tougher challenge is delivering *stable* FPS, since servicing the wifi can itself be slow.  I think the trick is to balance time spent computing and publishing frames with slack time left for other tasks, so that when one of those tasks wakes up, we don't have to see an FPS glitch.  I think a good algorithm for that estimation will be applicable in all environments.

Finally, maybe we want to go past `millis()` and use `ESP.getCycleCount()` for improved precision?  I do get the feeling that rounding off to millisecond precision is making some of these calculations trickier than they need to be.

DedeHai:
> Finally, maybe we want to go past `millis()` and use `ESP.getCycleCount()` for improved precision? I do get the feeling that rounding off to millisecond precision is making some of these calculations trickier than they need to be.

Is more precision required? 1/10th of a frame at 100FPS seems plenty accurate. or which part are you referring to that needs better precision?  `ESP.getCycleCount()` roll over happens much more frequently and may need handling.

wled00/FX_fcn.cpp
softhack007: done.

wled00/FX_fcn.cpp
softhack007: explicit test added.

wled00/FX_fcn.cpp
softhack007: removed.

wled00/FX.h
softhack007: old code restored.

wled00/FX_fcn.cpp
softhack007: check for (_frametime < 1) removed.

wled00/FX.h
softhack007: done.

softhack007:
> > Finally, maybe we want to go past `millis()` and use `ESP.getCycleCount()` for improved precision? I do get the feeling that rounding off to millisecond precision is making some of these calculations trickier than they need to be.
> 
> Is more precision required? 1/10th of a frame at 100FPS seems plenty accurate. or which part are you referring to that needs better precision? `ESP.getCycleCount()` roll over happens much more frequently and may need handling.

I'm not sure either. 

In fact as we approach the higher framerates, the "air gets thin" with milliseconds time
* 12 ms => 83 fps
* 11 ms => 90 fps
* 10 ms => 100 fps
*  9 ms => 111 fps
*  8 ms => 120 fps
*  7 ms => 142 fps
*  6 ms => 166 fps
*  5 ms => 200 fps
*  4 ms => 250 fps
*  3 ms => 333 fps

willmmiles:
> > * Fixing the FPS calculation with _lastServiceShow
> 
> Maybe I should have explained better - `_lastServiceShow` is not just to fix the FPS calculation. Its actually needed to have a robust scheduling, that can deliver the requested framerates even when there are other things that add unknown delays - like NPB.show() taking longer every second frame, or unexpected delays in the main loop.

True!  While the metrics fix could be done as its own commit, having those accurate metrics is a prerequisite for scheduling improvements.

wled00/FX_fcn.cpp
```diff
@@ -1398,7 +1398,7 @@ void WS2812FX::service() {
     _lastServiceShow = nowUp; // correct timestamp, for better FPS control
   }
   #ifdef WLED_DEBUG
-  if (millis() - nowUp > _frametime*2) DEBUG_PRINTF_P(PSTR("Slow strip %u/%d.\n"), (unsigned)(millis()-nowUp), (int)_frametime);
+  if ((_targetFps != FPS_UNLIMITED) && (millis() - nowUp > _frametime)) DEBUG_PRINTF_P(PSTR("Slow strip %u/%d.\n"), (unsigned)(millis()-nowUp), (int)_frametime);
```
blazoncek, line 1401: This will end up in non-debug builds like `if (...) ;` which is a waste of CPU cycles.
Wrap `#ifdef` around it.

softhack007: aehm ... not sure what you refer to ... its already wrapped
```C++
#ifdef WLED_DEBUG
    if ((_targetFps != FPS_UNLIMITED) && (millis() - nowUp > _frametime)) DEBUG_PRINTF_P(PSTR("Slow strip %u/%d.\n"), (unsigned)(millis()-nowUp), (int)_frametime);
#endif
```

blazoncek: Uh. I'm sorry. I've become sloppy.

softhack007:
> In fact as we approach the higher framerates, the "air gets thin" with milliseconds time

@willmmiles @DedeHai The fps  calculation in strip.show() could be changed to use `esp_timer_get_time()`. Its a system call returning elapsed time in micros as uint64_t. It's only availeable on esp32, but we could keep the old code for 8266.
Accurate fps calculation will be a win for performance measurements after code changes.

The effect timing code would stay as-is. For measuring FPS, rollover would not be a problem because we can simply skip measurements where last_timestamp > current_timestamp.

The MoonModules fork is using `esp_timer_get_time()` already, so I could clean up the code a bit and make a separate  PR.

https://github.com/MoonModules/WLED/blob/ffa5445aec9a14b39fd13f8c00d6fc9489ac92bb/wled00/FX_fcn.cpp#L2073-L2083

What do you think?

willmmiles:
> > In fact as we approach the higher framerates, the "air gets thin" with milliseconds time
> 
> @willmmiles @DedeHai The fps calculation in strip.show() could be changed to use `esp_timer_get_time()`. Its a system call returning elapsed time in micros as uint64_t. It's only availeable on esp32, but we could keep the old code for 8266. Accurate fps calculation will be a win for performance measurements after code changes.
> 
> The effect timing code would stay as-is. For measuring FPS, rollover would not be a problem because we can simply skip measurements where last_timestamp > current_timestamp.
> 
> The MoonModules fork is using `esp_timer_get_time()` already, so I could clean up the code a bit, remove float calculations, and make a separate PR.
> 
> https://github.com/MoonModules/WLED/blob/ffa5445aec9a14b39fd13f8c00d6fc9489ac92bb/wled00/FX_fcn.cpp#L2073-L2083
> 
> What do you think?

Personally I strongly prefer solutions that work on all supported platforms, hence the suggestion of `getCycleCount()`.  Rollovers are not an issue for relative measurements with unsigned types; `new_timestamp - old_timestamp` is always correct unless you've "lapped" the clock (for a 240MHz clock with a 32-bit cycle counter, that's a delay of 17.896 seconds -- not likely for an FPS counter!).  `ESP.getCpuFreqMHz()` returns the factor needed to convert back to microseconds (and by proxy, milliseconds).

DedeHai:
I like Will's suggestion, if only relative comprisons are needed, than using getCycleCount() is the better solution. It's very easy to write a `#define` or inline function to return the time in microseconds or even nanoseconds.

just FYI regarding performance measurements and FPS: I have modified code I use to increase the low-pass filter time constant of `getFps()`, I could add that with a define that can be set as a compile parameters, so debug builds can choose how much to average the value.

softhack007:
Yes @DedeHai @willmmiles what you say makes sense. I can try work out a PR during the week.

@DedeHai I'd love to see how you modified the FPS low-pass filtering. Actually the current FPS filter makes measurements very inaccurate, so any improvements is welcome 😃

softhack007:
> it feels to me like there's a generally safe algorithm somewhere in here, struggling to get out.

@willmmiles you might be right - I'm also not super happy with my implementation, however it solves some problems with the old code:
* broken metric for "time since last draw"
* too many idle loops, so testing FPS after code improvements produces misleading results
* the new code allows users to get the "Target FPS" they wanted, if effects and strip are fast enough. The old code was very unreliable in this respect.


The effect scheduling current works in a synchronous way, where all effects are updated at the same time (guaranteed by FRAMETIME). However effects that request a different frametime may be lucky (if requested time is a multiple of the scheduler timestep) or unlucky if they are completely of of sync. Luckily almost all effect have "return FRAMETIME" so they fit into the schedule.

If we want to faithfully schedule effects with different timing needs, it could not be done in an "all effects draw if needed, then strip.show updates all LEDs". 

With different timing, you would let some effects draw (if their time has passed), while skipping the ones which are not due yet (the current algo does this, too). But: the next scheduling cycle is dynamic. Its the minimum of all frame delays, i.e. min(segment.next_time - strip.now). 

This means additional draw cycles in the main loop, plus strip.show() would be called much more frequently.

Maybe there is a better way, so I'm open for any ideas and suggestions.

softhack007:
I've simplified the scheduler logic, and removed all estimates for "lower limits". The logic is much cleaner now
* _frametime is used to regulate the minimal time between two runs of strip.service()
* MIN_SHOW_DELAY is the minimum FRAMETIME that can be requested by effects

The logic is now very much in-line with the original function from WS2812FX
https://github.com/kitesurfer1404/WS2812FX/blob/ca19d320217afb970163c4771d62a183f761202a/src/WS2812FX.cpp#L81-L82

<br/>

side note
We'll need to take a look at the "Andriod" effects - it requests FRAMETIME between 3 and 25ms for a standard 1m 60leds strip. Actually the calculation breaks completely when SEGLEN > 128 when the effect will deliver a constant "3".

https://github.com/Aircoookie/WLED/blob/70323b947745e81644b955d3d16cc2bd24059636/wled00/FX.cpp#L811

blazoncek (commented):
I do not know what are the reasons for running (certain) effects faster than 120 FPS to grant the change in MIN_SHOW_DELAY or additional logic to prematurely exit `strip.service()` call.

If this is for purely *developer's assistance* then perhaps adding conditional code might be better option than "unlimited FPS" (which is false as it will still be limited to 333 FPS in best case).

IMO fixing MIN_SHOW_DELAY to 8 for ESP32 and 16 for ESP8266 would be a good option.

wled00/FX_fcn.cpp
```diff
@@ -1436,8 +1445,9 @@ uint16_t WS2812FX::getFps() const {
 }
 
 void WS2812FX::setTargetFps(uint8_t fps) {
-  if (fps > 0 && fps <= 120) _targetFps = fps;
-  _frametime = 1000 / _targetFps;
+  if (fps <= 120) _targetFps = fps;
+  if (_targetFps > 0) _frametime = 1000 / _targetFps;
+  else _frametime = 3;     // unlimited mode
```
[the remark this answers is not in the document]
softhack007: "3" is a safe value from testing. The problem  with "1" or "0" is I've had wifi stability problems. So I chose this value as a safe compromise.

wled00/FX_fcn.cpp
```diff
@@ -1309,7 +1309,14 @@ void WS2812FX::finalizeInit() {
 void WS2812FX::service() {
   unsigned long nowUp = millis(); // Be aware, millis() rolls over every 49 days
   now = nowUp + timebase;
-  if (nowUp - _lastShow < MIN_SHOW_DELAY || _suspend) return;
```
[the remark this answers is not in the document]
softhack007: No, using MIN_SHOW_DELAY here means that user can get up to 62 fps, or 125 fps, but nothing else. Why should we limit users this way? Do our users need a nanny? 

Technicially the new code is better, because it achieves any FPS chosen by the user, its efficient on CPU time, and it's safe enough on the slower chips to keep wifi running well.

wled00/data/settings_leds.htm
```diff
@@ -870,7 +874,9 @@ <h3>Advanced</h3>
 			<option value="2">Linear (never wrap)</option>
 			<option value="3">None (not recommended)</option>
 		</select><br>
-		Target refresh rate: <input type="number" class="s" min="1" max="120" name="FR" required> FPS
+		Target refresh rate: <input type="number" class="s" min="0" max="120" name="FR" oninput="UI()" required> FPS
+		<div id="fpsHelp" style="display: block;"><i>use 0 for unlimited</i><br></div>
+		<div id="fpsNone" style="color:orange; display: none;">Unlimited FPS Mode<br></div>
```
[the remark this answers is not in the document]
softhack007: Thanks, will change.

wled00/FX_fcn.cpp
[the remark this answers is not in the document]
blazoncek: > Why should we limit users this way? Do our users need a nanny?

IMO yes, you should limit because it makes code more/easily predictable. And yes, most users "need a nanny". Just look at the comments on Discord or forum. But that's just my opinion. Do as you please.

BTW I did not comment code efficiency. I was merely wondering why changing MIN_SHOW_DELAY and not using that in comparisons throughout the code (instead you use "magic" numbers as @netmindz once said to me).

wled00/FX_fcn.cpp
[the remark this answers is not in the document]
softhack007: About effect speed: especially audioreactive effect sometimes benefit from high framerates. For example DJLight, FreqMatrix, FreqWave create a nice "flow with the music" look at high framerates. GEQ also look very good when running very fast framerates. 

Actually I we can discuss if 120 fps as the "normal limit" still makes sense. Why not allow user to chose up to 250 if technicially possible?

wled00/FX_fcn.cpp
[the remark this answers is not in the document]
softhack007: > instead you use "magic" numbers as @netmindz once said to me).

In this case, I think the origin and purpose of "2" and "3" is well explained in the PR discussion - there might not be a solid proof of the need, but this is not always necessary imho. Also there are hints in the code comments about the "intend" where these numbers are used.  If you have a good idea for names of constants that contain "2" and "3", I'm all ears.

The previous "8" and "16" are legacy values, and afaik nobody here could explain any more where thy came from - we guessed that 8 and 16 might help slower CPUs like 8266. 

Let's wait until 0.15.1 with merging this PR, so we can get feedback from beta testers. I'll be there to take care of problems that might be related to the change. We can always fine-tune parameters. I think that measurements from @dosipod ([see above](https://github.com/Aircoookie/WLED/pull/4244#issuecomment-2452923650) ) clearly show that this PR has a value for users.

softhack007:
About the topic if users should get more guidance on the possible consequences of higher framerates: We have added a warning (see screenshot) in the MoonModules fork, that appears when setting FPS above 70.

image


@blazoncek @willmmiles @DedeHai @netmindz Should we add a similar warning in AC, too? 
Its just a few lines in led_settings.htm

softhack007:
@blazoncek to be honest, I'm tired of justifying my work against your vague concerns. Did you ever bother to compile and test this PR?

**To all maintainers**: I want to know if the changes I proposed in the PR are something you want in AC WLED (0.15.1) or not - as is or with improvements that we can still discuss.  

If nobody except myself supports this PR, I'll simply close it, and go on optimizing WLED in the MoonModules fork.

DedeHai:
just by looking at the FPS improvements, there clearly is something wrong with the current code, so I appreciate the efforts you are putting into this! I do however not understand the code or the issue enough to comment on the implementation of the improvements.

blazoncek:
> @blazoncek to be honest, I'm tired of justifying my work against your vague concerns

You requested my review, so I gave it. Feel free to do whatever you choose. I am no longer participating in WLED development.

DedeHai:
regarding the "unstable warning": while a warning is ok (keep it as short as possible to not waste flash memory) having a high FPS setting potentially corrupt files is not acceptable IMHO. Unstable as in 'it may occasionally crash' is ok, as long as it does not happen on simple setups. Users tend to ignore all warings and then go complain on discord / github.

DedeHai:
I run my particle system Tests on a C3 with FPS set to 99 all the time. I get eventual crashes when adding lots of segments or switching FX very fast but if I just use it 'normally' then it works. I assume the crashes are mostly due to heap exhaustion. It is a simple setup, no alexa or HA but sometimes with AR enabled.

softhack007:
> regarding the "unstable warning":

@DedeHai How about this one - more compact, and less scary. "backup" is a link to sec settings page.

image

softhack007:
> Edit: should I run the changes from this PR on a C3 for testing? What would be worst case scenarios for crashes?

yes please - behaviour on -C3 should be almost the same as before. Worst case might be that saving presets is not working, or UI reacts very slowly.

For a test - just to check if C3 can do it - change `#if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32C3)` to  `#if defined(ARDUINO_ARCH_ESP32)` in fx.h and fx_fcn.cpp.

wled00/FX_fcn.cpp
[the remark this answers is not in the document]
willmmiles: I think `_frametime = MIN_SHOW_DELAY` or `(MIN_SHOW_DELAY+1)` is probably the right value to use for the unlimited FPS case.

I've no objections to raising the FPS cap to `(1000/MIN_SHOW_DELAY)` -- I'm a big fan of "don't artificially limit what the hardware can do", and that dovetails with the idea that "MIN_SHOW_DELAY is what we believe is safe".  If we find it isn't safe enough, we can raise it further.

(Though long term I do want to aim for more flexible task management of the render/show, wifi, and API/usermod handling loops -- I think it's ok to throttle HA's incessant polling of preset.json to maintain a minimum FPS, too!)

willmmiles (changes requested):
This is shaping up well!  I think once we get the intent of `MIN_SHOW_DELAY` sorted, we'll be ready to go.

wled00/FX_fcn.cpp
```diff
@@ -1362,6 +1369,7 @@ void WS2812FX::service() {
           Segment::modeBlend(false);          // unset semaphore
         }
 #endif
+        frameDelay = max(frameDelay, unsigned(MIN_SHOW_DELAY));                     // limit effects that want to go faster than target FPS
```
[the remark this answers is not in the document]
softhack007: good point :-) yes this was left over from my previous code, I agree its not needed here any more.

willmmiles (approved):

softhack007:
@willmmiles The new logic would also cause very frequent to analog PWM outputs, if users set target FPS above 60.

I don't believe it will be a problem for esp32, however I'm not sure for 8266 (its more your area 😉). 
Maybe you could cross-check if 8266 PWM can take an update each 8ms?

I'm thinking about the code in BusPwm::show():
https://github.com/Aircoookie/WLED/blob/271a07a7d6a40026b184cb2c0f96c041f5165a42/wled00/bus_manager.cpp#L596

willmmiles:
> @willmmiles The new logic would also cause very frequent updates of analog PWM outputs, if users set target FPS above 60.
> 
> I don't believe it will be a problem for esp32, however I'm not sure for 8266 (its more your area 😉). Maybe you could cross-check if 8266 PWM can take an update each 8ms?
> 
> I'm thinking about the code in BusPwm::show():
> 
> https://github.com/Aircoookie/WLED/blob/271a07a7d6a40026b184cb2c0f96c041f5165a42/wled00/bus_manager.cpp#L596

Hmm, yes, I have an active PR in that area: https://github.com/Aircoookie/WLED/pull/4165

For non-phase-aligned channels, it's basically just writing the new duty cycle to some shared memory.   For phase aligned channels there's a short wait (at most 1 cycle at the analog frequency) per PWM channel updated.

So I expect it'll be fine.  I'll validate with a scope over the weekend.  Strictly speaking the old code could theoretically update at that rate too, if tickled the right way.

willmmiles:
Glad I ran the test!  The overall logic has no issues with high frame rates, but I found a case where the phase locking can break down (resulting in flickering) if the first channel is at fully on or fully off.

DedeHai:
Finally got around to test this PR. Got my upvote! 
I see a huge increase in FPS on the C3 and runs rock solid. I even lowered `MIN_FRAME_DELAY` to zero, polled FPS reading over websocket 10x per second or refreshing the webUI manually in quick succession, while running 2 overlapping segments on 16x16 with heavy computation, FPS set to unlimited, actual FPS was around 70, meaning FX calculation was the limiting factor. In this test, I could not get it to crash or to not save a prset or config. UI loading is very fast, on par with vanilla 0.15 if not faster, even in unlimited mode with `MIN_FRAME_DELAY` set to 1.
I would suggest to also use the `MIN_FRAME_DELAY` of 4 for the C3 like it is used for S2.

softhack007:
> I would suggest to also use the `MIN_FRAME_DELAY` of 4 for the C3 like it is used for S2.

@DedeHai thanks for the testing and feedback 😃 
I'll change as you suggested, so S2 and C3 will use the same `MIN_FRAME_DELAY` of 4 (or maybe reduce to 3 for both S2 and C3).

netmindz:
Are you happy for me to merge @DedeHai , now I've tagged 0.15.0.rc.1 ?

softhack007:
@netmindz I think this is safe to merge - we have similar code in MM since a long time.

The only "but" I can find is the android effect - it _might_ run a bit slower than before. That's actually due crazy coding of this one effect.

PR #4245, opened 2024-11-02
Optimization: color_blend() variable range now determined by overloading

DedeHai:
Removing the `bool` saves on code size and makes the function a tiny bit faster. Also this is a cleaner solution IMHO. 

Tested the affected FX, did not notice any change.

### DIFF
--- a/wled00/fcn_declare.h
+++ b/wled00/fcn_declare.h
@@ -78,7 +78,7 @@ class NeoGammaWLEDMethod {
 };
 #define gamma32(c) NeoGammaWLEDMethod::Correct32(c)
 #define gamma8(c)  NeoGammaWLEDMethod::rawGamma8(c)
-[[gnu::hot]] uint32_t color_blend(uint32_t,uint32_t,uint16_t,bool b16=false);
+[[gnu::hot]] uint32_t color_blend(uint32_t,uint32_t,uint16_t);
 inline uint32_t color_blend(uint32_t c1, uint32_t c2, uint8_t blend)  { return color_blend(c1, c2, uint16_t(((uint16_t)blend << 8))); };
 [[gnu::hot]] uint32_t color_add(uint32_t,uint32_t, bool fast=false);
 [[gnu::hot]] uint32_t color_fade(uint32_t c1, uint8_t amount, bool video=false);
--- a/wled00/fcn_declare.h
+++ b/wled00/fcn_declare.h
@@ -79,6 +79,7 @@ class NeoGammaWLEDMethod {
 #define gamma32(c) NeoGammaWLEDMethod::Correct32(c)
 #define gamma8(c)  NeoGammaWLEDMethod::rawGamma8(c)
 [[gnu::hot]] uint32_t color_blend(uint32_t,uint32_t,uint16_t,bool b16=false);
+inline uint32_t color_blend(uint32_t c1, uint32_t c2, uint8_t blend)  { return color_blend(c1, c2, uint16_t(((uint16_t)blend << 8))); };
 [[gnu::hot]] uint32_t color_add(uint32_t,uint32_t, bool fast=false);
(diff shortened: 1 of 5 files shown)


wled00/fcn_declare.h
```diff
@@ -78,7 +78,8 @@ class NeoGammaWLEDMethod {
 };
 #define gamma32(c) NeoGammaWLEDMethod::Correct32(c)
 #define gamma8(c)  NeoGammaWLEDMethod::rawGamma8(c)
-[[gnu::hot]] uint32_t color_blend(uint32_t,uint32_t,uint16_t,bool b16=false);
+[[gnu::hot]] uint32_t color_blend(uint32_t,uint32_t,uint16_t);
+inline uint32_t color_blend(uint32_t c1, uint32_t c2, uint8_t blend)  { return color_blend(c1, c2, uint16_t(((uint16_t)blend << 8))); };
```
softhack007, line 82: The difference between uint8_t and uint16_t is very subtle - when writing new code it might be too "easy" to accidentally call the wrong function and get unexpected results.

As we already need to correct almost all occurrences - how about naming the 8bit version "color_blend" (legacy support), while using "color_blend16" for the 16bit one? It might prevent some mistakes, because we don't need to be careful about casting the last parameter to uint8_t or uint16_t.

DedeHai: that is a very good suggestion, this was something I did not like about the change.

softhack007:
@DedeHai I've playes around with the color_blend function, and I found some improvements

#### converting 8bit blend to 16bit: 
`(uint16_t)b << 8` : with the shit only, the max "blend16" we can achieve is 0xFF00 so `if (blend == blendmax) return color2` is never taken. 

To use the full range, we could use `(uint16_t)b << 8) | b`.

#### improved blend accuracy
The blend logic seems to be based on some very old FastLED code. FastLED has changed their `blend8()` function about 3 years ago, to avoid discontinuities and rounding errors. The new math is explained here and it seems to be numerically more stable:

https://github.com/FastLED/FastLED/blob/9176e0ff9a67365c7bd4393da52be76b6ce4a5b3/src/lib8tion/math8.h#L565 

For the 8bit case, we can use the same logic - i've tested it and indeed its more accurate.
```C++
   uint8_t r3 = (((r1 << 8) | r2) + (r2 * blend) - (r1*blend) ) >> 8;
``` 

For 16bit blend, it might be trickier, because    `| r2`   (solves rounding problems) should be either `| r2<<8` or `| r2<<8 | r2`

DedeHai:
> with the shift only, the max "blend16" we can achieve is 0xFF00

this was fixed in the latest commit by adding a check in the inline function.

I need to take a closer look at the optimization, maybe this new version can be done the way I did with `color_add()` in https://github.com/Aircoookie/WLED/pull/4138 (calculating two colors in one go)

DedeHai:
I made some further optimizations based on the new fastLED version you proposed. In my tests, your suggested variant was a tiny bit faster when testing it on the ripple FX, my updated version uses less code so I am guessing the compile did some inlining to speed up `drawCircle()` in some way. 
The new variant with ditching the 16bit blend and going to 8bit gives almost identical results, just the rounding is slightly different. When plotting the output of the functions, the 8bit variant is actually not any worse, at least in the cases I have taken a closer look at. When checking the ripple FX, which is the only one using the blend function apart from transitions, I could not see any difference but the new code is slightly faster.
Can you please check if this works as expected in your test cases?

netmindz:
Are we good to merge this once the conflict is resolved?

DedeHai:
there are some pending changes.
@blazoncek can you merge your changes into this PR rather than the speed_optimization PR?

blazoncek:
I'm not sure I need to do anything regarding this PR.
However, IMO the side effect of this PR is that it will always blend using 8 bit logic which may introduce posterization. If that is of any concern is up to the users.

PR #4250, opened 2024-11-04
Improved FPS calculation

DedeHai:
This improvement is for dev purposes mainly but it also gives a bit smoother FPS return values in normal builds.

Changes:
- fixed point calculation for improved accuracy, 9.7 bit, so max FPS can be 511
~~- dithering (in debug builds only)~~
- averaging and optional multiplier can be set as compile flags

Example for speed testing with long averaging and a 10x multiplier:
```
-D FPS_CALC_AVG=200
-D FPS_MULTIPLIER=10
```
Since the calculation resolution is limited (9.7bit fixed point) `FPS_CALC_AVG `values larger than 200 can hit resolution limit and get stuck before reaching the final value.

If `WLED_DEBUG` is defined, dithering is added to the returned value so sub-frame accuracy is possible in post-processing without enabling the multiplier.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1412,10 +1412,12 @@ void WS2812FX::show() {
 
   unsigned long showNow = millis();
   size_t diff = showNow - _lastShow;
-  size_t fpsCurr = 200;
-  if (diff > 0) fpsCurr = 1000 / diff;
-  _cumulativeFps = (3 * _cumulativeFps + fpsCurr +2) >> 2;   // "+2" for proper rounding (2/4 = 0.5)
-  _lastShow = showNow;
+
+  if (diff > 0) { // skip calculation if no time has passed
+    size_t fpsCurr = (1000<<7) / diff; // fixed point 9.7 bit
+    _cumulativeFps = (FPS_CALC_AVG * _cumulativeFps + fpsCurr + FPS_CALC_AVG / 2) / (FPS_CALC_AVG + 1);   // "+FPS_CALC_AVG/2" for proper rounding
+    _lastShow = showNow;
+  }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


softhack007:
@DedeHai looks very good for me, thanks. Two questions

* the max possible value of 200FPS is due to `uint16_t _cumulativeFps`, right? so in case i wanted more, changing `_cumulativeFps` to `uint32_t` would be enough?
* to use a higher resolution time (like ESP.getCycleCount()) is just a matter of scaling `diff` and `fpsCurr` appropriately? or would it be better to also increase fixedpoint resulution in that case?

DedeHai:
- the problem is resolution, I don't remember exactly which part causes the problem, but it is a standard problem for fixed point low pass filters. You do not want to crank up the number past 200 anyway, the time constant is roughly 6 times larger, so at 200 with 100FPS it takes over 10 seconds for the value to reach the targetvalue. 400 may still work, did not test. I slightly changed the code I used before for this PR, for me, I used 8.8 math but I thought 255FPS maximum may be exceeded in the future
- if you want more smoothing then yes, you need to increase it to 32bit AND (more importantly) increase the bit shift (and adjust what "0.5" is with the new shift). I really don't think its necessary though, if your FPS are so unstable that an averaging over like 10 seconds does not give you accurate numbers, you may have a different issue there...
- the time resolution is irrelevant, FPS are always in seconds ;) but yes, just need to adapt some values to match. The filter I implemented here (or I must say extended, it was the same before) is universal: value in = value out, as long as there are no overflows.

softhack007:
> edit:
> I realize now that you wrote 200FPS max. the limit is 511FPS.

Thanks for your explanations. I think 511 FPS is future-safe for a long time ;-)

DedeHai:
just for reference: I use this to check FPS in my tests https://pypi.org/project/wled2graph/

wled00/FX_fcn.cpp
```diff
@@ -1432,7 +1434,11 @@ bool WS2812FX::isUpdating() const {
  */
 uint16_t WS2812FX::getFps() const {
   if (millis() - _lastShow > 2000) return 0;
-  return _cumulativeFps +1;
+  #ifdef WLED_DEBUG
+  return (FPS_MULTIPLIER * (_cumulativeFps + (random16() & 63))) >> 7;  // + random("0.5") for dithering
+  #else
+  return (FPS_MULTIPLIER * _cumulativeFps) >> 7; // _cumulativeFps is stored in fixed point 9.7 bit
```
softhack007, line 1440: Maybe replace all places with "7" by a constant (#define) ?

 Its better for maintainability to leave a hint what the magic number "7" stands for - in this case the fixed point multiplier.

DedeHai: done.

softhack007 (approved):
Looks good 👍 I approve

Actually the old FPS code was always "broken" in a sense that it stops converging to real FPS when the delta is below 2. So we could see this as a bugfix that would fit into 0.15.

Issue #3019, opened 2023-01-12
[Feature Request] Manually Type in Slider Value


mermelmadness:
I would love to see a little text box next to the sliders that would allow for manually typed input (0-255).  The slider can still adjust the value, but being able to type it in manually would save time to get to where the user knows what they want.

tsr1tsr1:
Yes please

werkstrom:
I can see the appeal... In some cases it is indeed desirable to be very precise when setting these values. A few caveats though:

- IMHO the absolute majority of users and use cases will prefer the current UI and functionality most of the time, i.e. The "granular control" should be _available_ but not in the way/complicate of todays UI.
- Generally  adding such functionality, widely, has implications way out of the initial scope as it interacts with all other input logic. Collateral damage should be expected.

Given the above a solution that is minimally invasive on existing functionality would, IMHO, be preferred. A suggestion would be to add a "double click" function to the range sliders. Double click will not be unintentionally triggered by users but is easily available for any user wanting granular control.

I would initially propose something like this code:

```
function actDblC(){
    const rngSlds = document.querySelectorAll('input[type="range"]');

    // Loop through each range slider element
    rngSlds.forEach(sld => {
    sld.addEventListener("dblclick", () => {
        const nVal = prompt("Enter a new value:");
        if (nVal !== null) { // User clicked 'OK'
        sld.value = nVal;
        sld.dispatchEvent(new Event("change")); // Call the onchange event to update the value
        }
    });
    });
}
```

This adds a double click listener to all range slider objects on tha page and pops up a standard input prompt. When committing a new value the "onchange" event of the original control is called as to preserve functionality.

I'll leave this code here for testing purposes. To test, do as follows:

1. Navigate to a page in WLED you want to test on. Main page will work fine, with the brightness slider f.ex.
2. Open up the console in your browser (press F12 and possibly find the console tab)
3. Paste the whole function into the console and hit ENTER
4. Activate the code by pasting `actDblC();` into the console and hitting ENTER again.

Now if you double click on any range slider (like the brightness) you should be presented with a popup where you can enter a value.

blazoncek:
The problem really only exist on mobile devices where you have to use *imprecise finger*.
On computer you can always use arrow keys to set exact value.

mermelmadness:
Could it be a feature disabled by default that you enable through UI settings, like the HEX color input box?

PR #3170, opened 2023-04-09
(wip) added function for activating granual input

werkstrom:
Only the function is implemented, it is not called anywhere untill some testing is done. #3019

PR #4254, opened 2024-11-06
Convert Brightness, Speed, Intensity, and Custom Sliders from HTML output to number input

Svennte:
**Enhancement: Convert HTML Output Elements to Number Inputs for Improved Precision**

Sometimes, I struggled to input specific values using the range sliders for effects and brightness. To improve precision, I changed the HTML output elements for Brightness, Speed, Intensity, and Custom Sliders to number inputs. I kept changes minimal to fit within the current code structure.

**Tested with**  
- Branch: `0_15`
- Device: `esp32dev` 

---

**What Have I Changed**

**index.html**
- Deleted HTML output elements and inserted HTML number inputs.
- Added `this.value` as an argument for the Effects/Brightness setter functions.
- Added this.checkValidity() to ensure only valid values are accepted for number inputs.

**index.css**
- Renamed `.sliderbubble` class to `.sliderInput` and adjusted styles accordingly.
- Slightly reduced `.sliderwrap` class width to display the number input properly.
- Added styles to visually indicate invalid values in slider number inputs

**index.js**
- Added `val` as an argument for the Effects/Brightness setter functions, allowing both range and number inputs to function.
- Changed `updateTrail` to update the input instead of the output.
- Removed the now-unnecessary `slider bubble` event.
- Removed the `toggleBubble` function.

---

**Issues**

- The inputs function correctly across tested devices and browsers. However, in the WLED Android App, the on-screen keyboard doesn’t push the slider div to the top of the display (see screenshots below). This issue only affects the WLED Android App and does not occur in the WLED Native App or web clients.
- Just tested on ESP32dev, no other environments
- Not tested with apple devices/browser

**Questions**

I couldn't find functions that redefine min/max values for the range inputs. Is this correct, or is there something specific to handle min/max values that I may have overlooked?

Android_Firefox
Android_WLED-Native
Android_WLED-Standard
Windows_Chrome
Windows_Firefox
Windows_InvalidValue

### DIFF
--- a/wled00/data/index.css
+++ b/wled00/data/index.css
@@ -442,14 +442,14 @@ button {
 /* Slider wrapper div */
 .sliderwrap {
 	height: 30px;
-	width: 230px;
+	width: 222px;
 	max-width: 230px;
 	position: relative;
 	z-index: 0;
 }
 
 #sliders .slider {
-	padding-right: 64px; /* offset for bubble */
+	padding-right: 64px; /* offset for number input */
 }
 
 #sliders .slider, #info .slider {
@@ -694,26 +694,23 @@ img {
 	display: none;
 }
 
-.sliderbubble {
-	width: 24px;
-	position: absolute;
-	display: inline-block;
-	border-radius: 16px;
-	background: var(--c-3);
-	color: var(--c-f);
-	padding: 4px;
-	font-size: 14px;
-	right: 6px;
-	transition: visibility .25s ease,opacity .25s ease;
-	opacity: 0;
-	visibility: hidden;
-	/* left: 8px; */
-	top: 4px;
+input[type=number].sliderInput {
+    width: 24px;
+    position: absolute;
+    display: inline-block;
+    border-radius: 10px;
+    background: var(--c-3);
+    color: var(--c-f);
+    padding: 2px 4px;
+    font-size: 14px;
+    transition: visibility 0.25s ease, opacity 0.25s ease;
+    opacity: 1;
+    visibility: visible;
+    right: 4px;
+    bottom: 5px;
 }
-
-output.sliderbubbleshow {
-	visibility: visible;
-	opacity: 1;
+input[type=number].sliderInput:invalid {
+	color: #f00;
 }
 
 input[type=range] {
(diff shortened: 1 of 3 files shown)


softhack007:
Thanks for your proposal!

personally I'm not an expert on html and JS, but I can answer one question

> I couldn't find functions that redefine min/max values for the range inputs.

Ranges for all sliders (global brightness, effect sliders) are generally 0..255. The only exception is the "custom3" slider which has a reduced resolution of 0..31.

netmindz (approved):

Svennte:
> Ranges for all sliders (global brightness, effect sliders) are generally 0..255. The only exception is the "custom3" slider which has a reduced resolution of 0..31.

Okay, then the inputs are fine. 
Is there something I have to do now?

blazoncek:
Will fix #3019 but I would recommend restoring tabs which were replaced by spaces in JS file.

Svennte:
Thanks for your recommendations!

**> I would recommend restoring tabs which were replaced by spaces in JS file.**  
Tabs have been restored with the last commit. 

**> EDIT: I'd also try to reuse range from existing (hidden) input fields instead of introducing new ones.**  
While this is an interesting idea, I have some concerns about implementation. Here’s my reasoning:

- **Maintaining Readability and Manageability**: Reusing input ranges would require detaching and reattaching inputs on each tab change, such as when switching between segment value inputs.  
- **Complexity of Custom Logic**: This approach would demand custom logic to dynamically set min/max values, and bind `onchange`, `oninput`, and possibly other functions.  
- **Potential Future Issues**: To ensure that future UI changes wouldn’t introduce unexpected issues, we’d likely need an input manager. This manager would handle adding, removing, and modifying inputs based on a config file for each tab.  
- **Risk of Misuse Across Tabs**: Without the input manager, it might be challenging to track which input is in use on each tab, especially as this approach scales to more inputs.

If you have a simpler solution in mind that I may have overlooked, please let me know. Given the required work and the limited benefits, this approach seems impractical, especially since the inputs don’t share the same parent element.

DedeHai:
what is the status here?
https://github.com/wled/WLED/pull/3170 seems related: would using that be a better approach?

blazoncek:
I would much prefer #3170 as it is simpler and universal.

PR #4158, opened 2024-09-23
Effect blending styles (recreated PR from #3877)

blazoncek:
Implements 14 different transitions/blends between different effects/modes.

Alternative to https://github.com/Aircoookie/WLED/pull/3669 by @tkadauke with less memory fragmentation and utilisation.
Implemented using clipping rectangles which may not produce same results as original PR (https://github.com/Aircoookie/WLED/pull/3669).

New PR because of unintended push into #3877 (see relevant discussion there).

PR #4256, opened 2024-11-07
replaced repeated progress() calculation calls with a variable

DedeHai:
progress() is called in `setPixelColor()` re-calculating the transition progress for each pixel. 
Replaced that call with an inline function to get the new segment variable. The progress is updated in service() when `handleTransition()` is called. The new variable is in a spot where padding is added, so this should not use more RAM. 

Result: over 10% increase in FPS on 16x16 matrix during transitions

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -363,6 +363,7 @@ typedef struct Segment {
     };
     uint8_t startY;  // start Y coodrinate 2D (top); there should be no more than 255 rows
     uint8_t stopY;   // stop Y coordinate 2D (bottom); there should be no more than 255 rows
+    uint16_t transitionprogress; // current transition progress 0 - 0xFFFF
     char    *name;
 
     // runtime data
@@ -468,6 +469,7 @@ typedef struct Segment {
       check3(false),
       startY(0),
       stopY(1),
+      transitionprogress(0xFFFF),
       name(nullptr),
       next_time(0),
       step(0),
@@ -559,12 +561,13 @@ typedef struct Segment {
     // transition functions
     void     startTransition(uint16_t dur);     // transition has to start before actual segment values change
     void     stopTransition();                  // ends transition mode by destroying transition structure (does nothing if not in transition)
-    inline void handleTransition() { if (progress() == 0xFFFFU) stopTransition(); }
+    inline void handleTransition() { updateTransitionProgress(); if (progress() == 0xFFFFU) stopTransition(); }
     #ifndef WLED_DISABLE_MODE_BLEND
     void     swapSegenv(tmpsegd_t &tmpSegD);    // copies segment data into specifed buffer, if buffer is not a transition buffer, segment data is overwritten from transition buffer
     void     restoreSegenv(tmpsegd_t &tmpSegD); // restores segment data from buffer, if buffer is not transition buffer, changed values are copied to transition buffer
     #endif
-    [[gnu::hot]] uint16_t progress() const;                  // transition progression between 0-65535
+    [[gnu::hot]] void updateTransitionProgress();            // set current progression of transition
+    inline uint16_t progress() const { return transitionprogress; };  // transition progression between 0-65535
     [[gnu::hot]] uint8_t  currentBri(bool useCct = false) const; // current segment brightness/CCT (blended while in transition)
     uint8_t  currentMode() const;                            // currently active effect/mode (while in transition)
     [[gnu::hot]] uint32_t currentColor(uint8_t slot) const;  // currently active segment color (blended while in transition)
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -363,7 +363,7 @@ typedef struct Segment {
     };
     uint8_t startY;  // start Y coodrinate 2D (top); there should be no more than 255 rows
     uint8_t stopY;   // stop Y coordinate 2D (bottom); there should be no more than 255 rows
-    uint16_t transitionprogress; // current transition progress 0 - 0xFFFF
+    // note: two bytes of padding are added here
     char    *name;
 
     // runtime data
@@ -419,6 +419,7 @@ typedef struct Segment {
     static CRGBPalette16 _newRandomPalette;   // target random palette
     static uint16_t _lastPaletteChange;       // last random palette change time in millis()/1000
     static uint16_t _lastPaletteBlend;        // blend palette according to set Transition Delay in millis()%0xFFFF
+    static uint16_t _transitionprogress;      // current transition progress 0 - 0xFFFF
     #ifndef WLED_DISABLE_MODE_BLEND
     static bool          _modeBlend;          // mode/effect blending semaphore
     #endif
@@ -469,7 +470,6 @@ typedef struct Segment {
       check3(false),
       startY(0),
       stopY(1),
-      transitionprogress(0xFFFF),
       name(nullptr),
       next_time(0),
       step(0),
@@ -567,7 +567,7 @@ typedef struct Segment {
     void     restoreSegenv(tmpsegd_t &tmpSegD); // restores segment data from buffer, if buffer is not transition buffer, changed values are copied to transition buffer
     #endif
     [[gnu::hot]] void updateTransitionProgress();            // set current progression of transition
-    inline uint16_t progress() const { return transitionprogress; };  // transition progression between 0-65535
+    inline uint16_t progress() const { return _transitionprogress; };  // transition progression between 0-65535
     [[gnu::hot]] uint8_t  currentBri(bool useCct = false) const; // current segment brightness/CCT (blended while in transition)
     uint8_t  currentMode() const;                            // currently active effect/mode (while in transition)
     [[gnu::hot]] uint32_t currentColor(uint8_t slot) const;  // currently active segment color (blended while in transition)
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -317,12 +317,12 @@ void Segment::stopTransition() {
 }
 
 // transition progression between 0-65535
-uint16_t IRAM_ATTR Segment::progress() const {
+void IRAM_ATTR Segment::updateTransitionProgress() {
+  transitionprogress = 0xFFFFU;
   if (isInTransition()) {
     unsigned diff = millis() - _t->_start;
-    if (_t->_dur > 0 && diff < _t->_dur) return diff * 0xFFFFU / _t->_dur;
+    if (_t->_dur > 0 && diff < _t->_dur) transitionprogress = diff * 0xFFFFU / _t->_dur;
   }
-  return 0xFFFFU;
 }
 
 #ifndef WLED_DISABLE_MODE_BLEND
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -317,7 +317,7 @@ void Segment::stopTransition() {
 }
 
 // transition progression between 0-65535
-void IRAM_ATTR Segment::updateTransitionProgress() {
+void Segment::updateTransitionProgress() {
   transitionprogress = 0xFFFFU;
   if (isInTransition()) {
     unsigned diff = millis() - _t->_start;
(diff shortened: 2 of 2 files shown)


wled00/FX_fcn.cpp
```diff
@@ -317,12 +317,12 @@ void Segment::stopTransition() {
 }
 
 // transition progression between 0-65535
-uint16_t IRAM_ATTR Segment::progress() const {
+void IRAM_ATTR Segment::updateTransitionProgress() {
```
softhack007, line 320: I think you could use `IRAM_ATTR_YN` here - it means that esp32 puts the function into IRAM, while 8266 doesn't. We'll save some IRAM space especially on the "_compat" builds.

As the function is only called once per frame in WS2812FX::service() - via seg.handleTransition() - it might even be better to remove the IRAM_ATTR as this call is not performance critical any more.

wled00/FX_fcn.cpp
```diff
@@ -317,12 +317,12 @@ void Segment::stopTransition() {
...
+  transitionprogress = 0xFFFFU;
   if (isInTransition()) {
     unsigned diff = millis() - _t->_start;
-    if (_t->_dur > 0 && diff < _t->_dur) return diff * 0xFFFFU / _t->_dur;
+    if (_t->_dur > 0 && diff < _t->_dur) transitionprogress = diff * 0xFFFFU / _t->_dur;
```
softhack007, line 324: Just wondering - do we need an `else` branch here, so that transitionprogress always has a well-defined value?

softhack007: ah, forget my comment. did not see line 321.

wled00/FX_fcn.cpp
DedeHai: I thought about removing the attribute but left it as is since I have no way to check the difference. I once as a test removed all IRAM_ATTR and on my setup there was zero performance change. 
I think removing it here is safe, as you say, this is only called once per frame.

softhack007: esp8266 will thank you ;-)

DedeHai: Maybe `WLED_SAVE_IRAM` should also be defined on ESP32 C3: it is not as performant as the ESP32 any way and putting stuff in IRAM uses a lot more flash for some reason. If I enable `WLED_SAVE_IRAM` on the C3 that saves 1.6k of flash. On the ESP32 it only saves 68 bytes of flash.
Any suggestions performance tests that would show if this is a valid option?

wled00/FX.h
```diff
@@ -363,6 +363,7 @@ typedef struct Segment {
     };
     uint8_t startY;  // start Y coodrinate 2D (top); there should be no more than 255 rows
     uint8_t stopY;   // stop Y coordinate 2D (bottom); there should be no more than 255 rows
+    uint16_t transitionprogress; // current transition progress 0 - 0xFFFF
```
blazoncek, line 366: This could be a `static` variable as it will be modified in each `handleTransition()` call at the start of `service()` loop.

DedeHai: what is the current behaviour if for example the palette of one segment is changed with 10s transition, and after 5s palette of a second segment is changed? Will that stop transition of the first segment? i.e. are transition start/stop per segment or global?

blazoncek: The transition time is bound to `strip` ATM but each segment may start transition at its own (point in) time. So each segment should transition on its own (that was my goal when implementing current transitions, compared to previous, limited transitions) independent from others.

wled00/FX_fcn.cpp
blazoncek: Any function marked with IRAM_ATTR will always be kept in fast SRAM and will never be fetched from flash. The basic idea for IRAM_ATTR is to be used in ISR or functions that may access (write to) flash directly.
The benefit of using it elsewhere is to speed up access to such function as it will never go to cache hit/miss logic.

Contrary to what @softhack007 is saying I still think adding IRAM_ATTR to functions that are called very often is beneficial. I am telling this from experience with over 50 installed ESP8266's with various options and usermods installed.

softhack007: > Contrary to what @softhack007 is saying I still think adding IRAM_ATTR to functions that are called very often is beneficial.

It might be beneficial, however we talked about the new `progress()` that's only called a few hundred times per second (max) now. The function is not time critical any more with this PR, so why use IRAM_ATTR for it?

softhack007: (side-topic)

> I once as a test removed all IRAM_ATTR and on my setup there was zero performance change.

This aligns with my own experiments on -S3 and esp32 with 80mhz flash - no noticeable performance  impact, however sometimes IRAM_ATTR increases program size. This can be explained because the compiler cannot inline such a function, even when there would be a benefit for program size.

Maybe it's also depending a lot on the CPU caches. In fact a function that's called really often has a good chance to be cached by the CPU already. Also a board with fast flash (qio 80mhz) is like 4x faster on flash reading, compared to slow flash (dout 40mhz).

As most cheap 8266 still have 40mhz dout, plus smaller caches, so it makes sense that there is still some benefit of IRAM_ATTR on boards with slow flash.

wled00/FX.h
softhack007: > This could be a `static` variable as it will be modified in each `handleTransition()` call at the start of `service()` loop.

Rather than make it static, I'd say put it into `strip`. 

Static member attributes of a class are always a good source of confusion when trying to read code written by someone else - because a 'static' member is technicially not even part of the object you work with. It survives `delete segment`, and changing it in one object instances also changes the value in all other instances.

wled00/FX_fcn.cpp
blazoncek: I have no doubt that any ESP32 performs adequately without IRAM_ATTR.
However ESP8266 is another thing and while it may be old and lacking it is still used by many users (including me) who keep attaching plenty of peripherals to it while running WLED. Hence I strongly urge to keep IRAM_ATTR as many times as possible.

wled00/FX.h
blazoncek: > Rather than make it static, I'd say put it into `strip`.

That would be a worse choice IMO. It can be considered in the same way as _vLength, etc in speed improvements branch by @DedeHai . It is used as a speedup, pre-calculated value.
If you think this will cause confusion, keep it as an instance member rather than `strip` member.
But that is just my opinion, no need to take it into account.

wled00/FX_fcn.cpp
DedeHai: can you test if this PR with latest commit has any impact on ESP8266? i.e. removed `IRAM_ATTR`  from `updateTransitionProgress()`

The other question was to add `WLED_SAVE_IRAM`  to C3 builds as it would save on flash size but may have negative performance impacts in certain situations / setups. If there is a way to test that on the C3 I could check. After all, the C3 is more of an upgraded ESP8266 compared to other ESP32 variants.

wled00/FX.h
DedeHai: If individual segment transition is to be kept/extended in the future, this should not be static but per segment. Also there is no RAM impact currently as there are 3 bytes of padding added in that space anyway.

blazoncek: > this should not be static but per segment

Look at the `Segment::beginDraw()` in another PR (#4138). Uses the same principle for `virtualLength()` etc.
If this PR is embedded there, there is no need for different logic regarding calculation of `progress` variable. Still, if effect function runs for more than 1ms then the result may not be accurate (within that function) regardless of where this substitution/pre-calculation comes from. `beginDraw()` is called at the start of effect drawing and `progress()` is used within `setPixelColor()` when blending is enabled (and more so in #4158 ).

DedeHai: That is good point. I guess the question is whether progress should be per frame, which was my intention, or per frame per segment, which is what you are suggesting if I understand that right. 
Doing it per frame would keep segment progress aligned i.e. first and last segment rendered use the same progress. I am not sure what the best solution is with regards to sync (or frame based super sync) if ever implemented as I do not understand it yet in detail.

blazoncek: `progress()` determines the state where transition is. It should not be "frame" based. It only depends on time since transition (for that segment) started. BTW it is possible to have multiple segments in transition with differing `progress()` value.

DedeHai: what is the reason `progress()` should depend on the order of segment processing or when does this give an advantage aver calculating it once per frame? If for example there are lots of segments with heavy FX calculation (>1ms) later segments would already be further in the transition in the same frame, resulting in different palette or brightness for example. That does not appear to be the correct approach IMHO. It would only be right if startTransition() is called with the same time delay, which I don't think it is. Using the 'per segment' update is more consistent code though, as you said. I think your suggestion of following the `Segment::beginDraw()` is a good compromise.

blazoncek: > what is the reason `progress()` should depend on the order of segment processing

It should not be *dependant on the order of segment processing*. Each "segment" (if you want it) has its own "next time" for update in the form of effect function of return value. So some effects may run at independent frame rate. Look at Android and a few others (i.e. Game of life).

What matters is the time difference between the start of transition and current time (in regards to the transition duration). This is not frame or segment based. It is true that segment holds both start time and transition duration and these two are unique to each segment. This was the prime reason to have `progress()` as a function that calculates how far transition has progressed. If you impose a limit that progress is calculated at the start of frame (for optimisation for speed) then it needs to be pre-calculated for each segment separately. Hence the suggestion for `beginDraw()`.

Why do you think that having different progress value for is wrong? Each segment operates independently and displays effect and/or palette independently of other segments. Even if overlaid. It is totally acceptable for two segments to be at different progress level.

And it is quite easy to achieve that.

DedeHai: There are different scenarios, your reasoning seems sound to me. My scenario was this: have 10 segments, all displaying the same FX, change palette on all of them at the same time -> a transition starts on all segments. Palette change may now not be simulatanuous on all segments, the last ones may already have progressed further. Is my thinking here correct for current code? I am not arguing against your suggestion to keep it 'per segment', which I think is the way to go, just trying to understand where a 'once per frame' would be better and where it would be worse.

blazoncek: It is correct and in most cases this will be the case (same start for all segments). But not in all cases (and not all segments may have same effect and/or palette).

So, if you have an exception, treat it as a regular behaviour (not all segments may start transition at the same time).

blazoncek:
I've tested the implementation that uses static `transitionProgress` variable but otherwise uses exactly the same logic as this PR and everything seems to work as expected.

netmindz:
Did we reach an agreement on the final implementation? @DedeHai

DedeHai:
we did, as just (wrongly) mentioned in the other PR: @blazoncek has already implemented and tested the changes

netmindz:
Are these changes in another pr, already merged or just local on your machine at the moment @blazoncek ?

blazoncek:
All of the changes are in my fork, combined with other updates. Probably not cherry-pickable, unfortunately. Implementation used there uses static member.
IMO A better place for this introduction would be "speed improvements" PR as it is just that.

Speed improvements and `progress()` improvements are running on my set-ups for over a month now without issues or glitches. But with some of my additions on top.

DedeHai:
I am not sure it is the best idea to put all changes in the same "speed improvements" PR, if you feel confident about that, I am fine with closing this one. Just have to remember the speed improvements PR must not be squash-merged or debugging will be hard.

DedeHai:
@softhack007 I have most of my PRs running here already, merging them all is a bit of a pain as there are many conflicts to resolve. I merged them in this branch: https://github.com/DedeHai/WLED/tree/0_15_PS_potpourri
The open PRs in question besides this one are:
#4245 #4225 #4138
also #4145 (adds new FX features)

@blazoncek are your changes public? I can manually add them to this PR if not cherry-pickable, its just be a few lines that changed right?

blazoncek:
Anything I do is public when *stable* enough. Never hid anything. 😉

DedeHai:
@softhack007 I think merging into a new branch is not necessary, I can do the (non squashed) merge of the PRs in the correct order to have the least conflicts and resolve any that remain. What I probably will do is a test-run of the merges in my repo to figure out what the best order is: in that potpourri branch I had to start over twice...

DedeHai:
@blazoncek are the changes in the latest commit ok?

blazoncek:
I think yes, though I also moved `updateTransitionProgress()` into `inline` `handleTransition()` as it is pointless to force a function call since the function is only called once.

DedeHai:
pointless yes but I find it much more readable than jamming it all in that inline.

PR #4267, opened 2024-11-09
Fix problem with spaces

w00000dy:
Due to the minifying, we had some problems with spaces.

**Before:**
grafik
```
RAM:   [=         ]  14.8% (used 48472 bytes from 327680 bytes)
Flash: [========= ]  86.8% (used 1365765 bytes from 1572864 bytes)
```

**Now:**
grafik
```
RAM:   [=         ]  14.8% (used 48472 bytes from 327680 bytes)
Flash: [========= ]  86.9% (used 1366069 bytes from 1572864 bytes)
```

### DIFF
--- a/tools/cdata.js
+++ b/tools/cdata.js
@@ -101,6 +101,7 @@ function adoptVersionAndRepo(html) {
 async function minify(str, type = "plain") {
   const options = {
     collapseWhitespace: true,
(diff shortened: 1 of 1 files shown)


blazoncek:
Confirmed working.

blazoncek (approved):

PR #4270, opened 2024-11-10
Replaced single palette cases with an array to consolidate code

DedeHai:
- all palettes are now defined in palettes.h
- access to fastled palettes as an array to remove the switch cases
- palette creation in json.cpp in a loop instead of repeated calls to save some flash (about 300 bytes)


blazoncek (approved):
It might also be beneficial to completely remove FastLED dependency and include/duplicate its palettes in WLED.
IMO FastLED is loosing foothold in WLED (if you will implement your other FastLED replacements).

blazoncek:
`qadd8()`, `sin8/sin16()`, `random8/random16()` I think you touched all of them at some point.
Other that are used are overloaded operators (i.e. `+` etc) which operate on `struct CRGB` (which you also touched).

I am not saying that it is necessary to replace the library, I am just thinking out loud about possibility to replace library since you already ventured into those waters by replacing some of the functionality of FastLED.

netmindz (approved):

Issue #4269, opened 2024-11-10
Crashes when using HTTP API within MQTT 


dosipod:
### What happened?

Crashes when using HTTP API within MQTT  . I am raising this on behave of @mcsSolutions  as it is also confirmed from our side .Original post on discord 
https://discord.com/channels/473448917040758787/779395228624617512/1304888652161220729

( I never used nor will use HTTP API within MQTT but only helping log this )

### To Reproduce Bug

User PL=1 with mqtt 
![image](https://github.com/user-attachments/assets/1b27b500-a553-4122-b2d3-19f441f0c0d3)

### Expected Behavior

No crash

### Install Method

Self-Compiled

### What version of WLED?

0.15.0-b7 (build 2410270)

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell
[21:28:53]Connected! IP address: 10.0.6.18
[21:28:53]Init STA interfaces
[21:28:53]mDNS started
[21:28:53]Reconnecting MQTT
[21:28:53]Reconnecting MQTT
[21:28:53]Loop took 38ms.
[21:28:53]Usermods took 0ms.
[21:28:53]Strip took 0ms.
[21:28:53]MQTT ready
[21:28:53]Publish MQTT
[21:28:53]MQTT ready
[21:28:53]Publish MQTT
[21:28:56]MQTT msg: wled/WLED18/api
[21:28:56]PL=1
[21:28:56]JSON buffer locked. (15)
[21:28:56]API req: win&PL=1
[21:28:56]Playlist unloaded.
[21:28:56]Request to apply preset: 1
[21:28:56]Guru Meditation Error: Core  1 panic'ed (StoreProhibited). Exception was unhandled.
[21:28:56]Core 1 register dump:
[21:28:56]PC      : 0x401329d6  PS      : 0x00060230  A0      : 0x8010b1b8  A1      : 0x3ffdcf00  
[21:28:56]A2      : 0x00000000  A3      : 0x3ffe37c8  A4      : 0x3ffc3954  A5      : 0x0000003c  
[21:28:56]A6      : 0x00000000  A7      : 0x00000000  A8      : 0x80123518  A9      : 0x3ffdce10  
[21:28:56]A10     : 0x000000ad  A11     : 0x94cb258c  A12     : 0x94cb258c  A13     : 0x3ffc3954  
[...]

PR #4271, opened 2024-11-11
Fix for #4269

willmmiles:
The crash bug in #4269 was introduced by #4152; I inadvertently removed a null check on the 'request' parameter.  Restore the missing check so non-JSON API calls don't crash.


blazoncek:
Confirming the fix resolves the issue #4269

blazoncek (approved):

PR #4272, opened 2024-11-11
HTTP API cleanup

willmmiles:
Along with #4271, I took the opportunity to clean up the surrounding code a small amount.  This PR:

- Renames 'handleSet' to 'handleHttpApi', clarifying what this function does;
- Moves the check for leading 'win' to the one context where it matters (the web server), removing the need for all other callers to re-add this fixed string;
- Moves the response generation out to the web server, eliminating an otherwise unnecessary parameter;
- Removes the 'IN' tag, as that internal API was superseded by passing `nullptr` for the `request`.  The conditional was broken anyhow, and HTTP requests are required to generate a response.

### DIFF
--- a/wled00/set.cpp
+++ b/wled00/set.cpp
@@ -805,14 +805,12 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
 //HTTP API request parser
 bool handleHttpApi(AsyncWebServerRequest *request, const String& req, bool apply)
 {
-  if (!(req.indexOf("win") >= 0)) return false;
-
   int pos = 0;
   DEBUG_PRINTF_P(PSTR("API req: %s\n"), req.c_str());
 
   //segment select (sets main segment)
   pos = req.indexOf(F("SM="));
-  if (pos > 0 && !realtimeMode) {
+  if (pos >= 0 && !realtimeMode) {
     strip.setMainSegmentId(getNumVal(&req, pos));
   }
 
@@ -821,7 +819,7 @@ bool handleHttpApi(AsyncWebServerRequest *request, const String& req, bool apply
   bool singleSegment = false;
 
   pos = req.indexOf(F("SS="));
-  if (pos > 0) {
+  if (pos >= 0) {
     unsigned t = getNumVal(&req, pos);
     if (t < strip.getSegmentsNum()) {
       selectedSeg = t;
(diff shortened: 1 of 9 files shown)


wled00/set.cpp
```diff
@@ -831,7 +829,7 @@ bool handleSet(AsyncWebServerRequest *request, const String& req, bool apply)
 
   Segment& selseg = strip.getSegment(selectedSeg);
   pos = req.indexOf(F("SV=")); //segment selected
-  if (pos > 0) {
+  if (pos >= 0) {
```
netmindz, line 832: This change isn't described in the description, what does it relate to?

wled00/wled_server.cpp
```diff
@@ -482,7 +482,17 @@ void initServer()
       return;
     }
 
-    if(handleSet(request, request->url())) return;
+    if (request->url().indexOf("win") == 0) {
+      if (handleHttpApi(request->url().substring(3))) {
```
blazoncek, line 486: Is it 3 or is it 4?
I would think 4 as you do not compare `&` in `handleHttpApi()` or there is no need to add equality to comparisons there.

wled00/wled_server.cpp
```diff
@@ -482,7 +482,17 @@ void initServer()
...
+        if (request->url().indexOf(F("&FX=")) > 0) {
```
blazoncek, line 487: Wouldn't it be better to move this into `handleHttpApi()`?
IMO it's place is there.

blazoncek (approved):
Seems like a good improvement. Have yet to test it, though.

wled00/wled_server.cpp
willmmiles: The leading '&' is intentionally left on as some of the test strings in `handleHttpApi()` have a leading "&", eg. `pos = req.indexOf(F("&S="));`.  I'm not really sure why some do and some don't -- I should probably just clean that up too while I'm in there.

wled00/wled_server.cpp
willmmiles: In the original code, we call `unloadPlaylist()` only the FX was changed from an HTTP request, but not from any other context.  It made more sense to me to place that code only in the HTTP request context, rather than add a special parameter to `handleHttpApi()` *just* to switch this call on or off.  It does mean naming the tag twice, but I think that's the lesser evil -- https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction

An alternative might be to augment the return code to indicate if the FX value was changed, rather than parsing the tag multiple times.

wled00/wled_server.cpp
willmmiles: Actually, given that, I'm going to convert this to draft for now and do some more testing.  I think there's some corner cases where this code will differ from the original.

wled00/wled_server.cpp
willmmiles: I agree -- although this PR was aimed to be a refactor with no behavioural changes, and the original logic only unloaded the playlist in the context of an HTTP request.

I'll do a more thorough review of the other call contexts and see if I can draw a better map.  No reason not to further improve things as we go.

blazoncek:
@willmmiles shall we reopen this?

willmmiles:
Sure.  It's not a burning priority but I would like to get back to it at some point.

PR #4279, opened 2024-11-14
Add settings and OTA lock state to info

willmmiles:
As requested by @Moustachauve on Discord: add the settings and OTA lock states to the info struct, so UIs can direct users to an appropriate workflow for updating their devices.

I'm not 100% confident about the usage of `interfaceUpdateCallMode` in these contexts to induce websocket notifications.  The notification subsystem currently seems a bit tangled to me.  If this stands out as obviously wrong, please let me know.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced real-time synchronization of security and lock status across connected interfaces.
  * Security settings (PIN and OTA lock state) now properly notify all connected clients when changed or expired.
  * Added visibility of current lock status in device information responses.

### DIFF
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -643,6 +643,8 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
     CJSON(wifiLock, ota[F("lock-wifi")]);
     CJSON(aOtaEnabled, ota[F("aota")]);
     getStringFromJson(otaPass, pwd, 33); //normally not present due to security
+    interfaceUpdateCallMode = CALL_MODE_WS_SEND;
+    stateChanged = true;
   }
 
   #ifdef WLED_ENABLE_DMX
--- a/wled00/json.cpp
+++ b/wled00/json.cpp
@@ -802,6 +802,8 @@ void serializeInfo(JsonObject root)
   os += 0x40;
   #endif
 
+  root[F("settingsLocked")] = !correctPIN;
+
   //os += 0x20; // indicated now removed Blynk support, may be reused to indicate another build-time option
 
   #ifdef USERMOD_CRONIXIE
@@ -818,11 +820,10 @@ void serializeInfo(JsonObject root)
   #endif
   #ifndef WLED_DISABLE_OTA
   os += 0x01;
-  root[F("ota")] = !otaLock;
+  root[F("otaLocked")] = otaLock;
   #endif
   root[F("opt")] = os;
-  root[F("settings")] = correctPIN;
-
+  
   root[F("brand")] = F(WLED_BRAND);
   root[F("product")] = F(WLED_PRODUCT_NAME);
   root["mac"] = escapedMac;
--- a/wled00/json.cpp
+++ b/wled00/json.cpp
@@ -823,7 +823,7 @@ void serializeInfo(JsonObject root)
   root[F("otaLocked")] = otaLock;
   #endif
   root[F("opt")] = os;
-  
+
   root[F("brand")] = F(WLED_BRAND);
   root[F("product")] = F(WLED_PRODUCT_NAME);
   root["mac"] = escapedMac;
--- a/wled00/json.cpp
+++ b/wled00/json.cpp
@@ -818,8 +818,10 @@ void serializeInfo(JsonObject root)
   #endif
   #ifndef WLED_DISABLE_OTA
   os += 0x01;
+  root[F("ota")] = !otaLock;
   #endif
   root[F("opt")] = os;
+  root[F("settings")] = correctPIN;
 
   root[F("brand")] = F(WLED_BRAND);
   root[F("product")] = F(WLED_PRODUCT_NAME);
--- a/wled00/set.cpp
+++ b/wled00/set.cpp
@@ -587,6 +587,7 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
       wifiLock = request->hasArg(F("OW"));
       aOtaEnabled = request->hasArg(F("AO"));
       //createEditHandler(correctPIN && !otaLock);
+      interfaceUpdateCallMode = CALL_MODE_WS_SEND;
     }
   }
 
--- a/wled00/set.cpp
+++ b/wled00/set.cpp
@@ -588,6 +588,7 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
       aOtaEnabled = request->hasArg(F("AO"));
       //createEditHandler(correctPIN && !otaLock);
       interfaceUpdateCallMode = CALL_MODE_WS_SEND;
+      stateChanged = true;
     }
   }
 
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -155,6 +155,7 @@ void WLED::loop()
   if (strlen(settingsPIN)>0 && correctPIN && millis() - lastEditTime > PIN_TIMEOUT) {
     correctPIN = false;
     createEditHandler(false);
+    interfaceUpdateCallMode = CALL_MODE_WS_SEND; // schedule WS update
   }
 
   // reconnect WiFi to clear stale allocations if heap gets too low
--- a/wled00/util.cpp
+++ b/wled00/util.cpp
@@ -355,7 +355,11 @@ void checkSettingsPIN(const char* pin) {
   if (!correctPIN && millis() - lastEditTime < PIN_RETRY_COOLDOWN) return; // guard against PIN brute force
   bool correctBefore = correctPIN;
   correctPIN = (strlen(settingsPIN) == 0 || strncmp(settingsPIN, pin, 4) == 0);
-  if (correctBefore != correctPIN) createEditHandler(correctPIN);
+  if (correctBefore != correctPIN) {
+    createEditHandler(correctPIN);
+    interfaceUpdateCallMode = CALL_MODE_WS_SEND;
+    stateChanged = true;
+  }
   lastEditTime = millis();
 }
 
--- a/wled00/wled_server.cpp
+++ b/wled00/wled_server.cpp
@@ -323,9 +323,9 @@ void initServer()
     releaseJSONBufferLock();
 
     if (verboseResponse) {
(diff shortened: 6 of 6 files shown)


blazoncek:
In fact it is convoluted. It took me 2 years to decipher it. 😄 
If you want notifications to be broadcast, use: `stateUpdated = true;`

wled00/cfg.cpp
```diff
@@ -643,6 +643,8 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
     CJSON(wifiLock, ota[F("lock-wifi")]);
     CJSON(aOtaEnabled, ota[F("aota")]);
     getStringFromJson(otaPass, pwd, 33); //normally not present due to security
+    interfaceUpdateCallMode = CALL_MODE_WS_SEND;
```
blazoncek, line 646: I do not think it is necessary to add this here as cfg is only read at boot time.

wled00/json.cpp
```diff
@@ -802,6 +802,8 @@ void serializeInfo(JsonObject root)
   os += 0x40;
   #endif
 
+  root[F("settingsLocked")] = !correctPIN;
```
blazoncek, line 805: A shorter text would be preferrable to reduce flash usage.

wled00/json.cpp
```diff
@@ -818,6 +820,7 @@ void serializeInfo(JsonObject root)
   #endif
   #ifndef WLED_DISABLE_OTA
   os += 0x01;
+  root[F("otaLocked")] = otaLock;
```
blazoncek, line 823: IMO the if enclosing is enough (and its bitfield of options).
Adding another key seems wasteful.

wled00/set.cpp
```diff
@@ -587,6 +587,8 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
       wifiLock = request->hasArg(F("OW"));
       aOtaEnabled = request->hasArg(F("AO"));
       //createEditHandler(correctPIN && !otaLock);
+      interfaceUpdateCallMode = CALL_MODE_WS_SEND;
```
blazoncek, line 590: Do you think it is really necessart to broadcast state/info after settings are saved?
IMO not but I may be missing a usecase.

wled00/wled.cpp
```diff
@@ -155,6 +155,7 @@ void WLED::loop()
   if (strlen(settingsPIN)>0 && correctPIN && millis() - lastEditTime > PIN_TIMEOUT) {
     correctPIN = false;
     createEditHandler(false);
+    interfaceUpdateCallMode = CALL_MODE_WS_SEND; // schedule WS update
```
blazoncek, line 158: Missing `stateUpdated` if we want same behaviour as elsewhere.

blazoncek (commented):
I'm ok by adding information that update may be locked (either via PIN or OTA password) but the same may be achieved by the app by polling/requesting the update page (it will produce PIN entry or 503 IIRC). So I'm not really sure it is needed.

The more important thing is if OTA is not compiled. That information is already in info object hidden in options/capabilities bitfield.

I also have considerations about JSON buffer size as it may get too small on ESP8266 since we are adding new items to info and state objects all the time.
And when @DedeHai 's particle system gets merged it will definitely be too small.

IMO the most important thing for @Moustachauve that is missing is the source location for correct binary and what compilation options/flags were used. This being due to users unintentionally updating customised version with plain from Github.

Usermods can be determined by checking info object, but some do not advertise their presence there. Nor in state object.
Perhaps enhancing options/capabilities bitfield would be better or provide a hook for usermods to advertise their presence.

netmindz:
> I'm ok by adding information that update may be locked (either via PIN or OTA password) but the same may be achieved by the app by polling/requesting the update page (it will produce PIN entry or 503 IIRC). So I'm not really sure it is needed.

Parsing the page looking for the pin entry feels very hacky and fragile 

> The more important thing is if OTA is not compiled. That information is already in info object hidden in options/capabilities bitfield.

Please can you share a link as where this is documented

> I also have considerations about JSON buffer size as it may get too small on ESP8266 since we are adding new items to info and state objects all the time. And when @DedeHai 's particle system gets merged it will definitely be too small.

I'm not aware of why the PS would have this affect, please add info to the appropriate PR

> IMO the most important thing for @Moustachauve that is missing is the source location for correct binary and what compilation options/flags were used. This being due to users unintentionally updating customised version with plain from Github.

Agreed, but again a can we please keep comments on the appropriate issue/pr so that we keep them all together.

> Usermods can be determined by checking info object, but some do not advertise their presence there. Nor in state object. Perhaps enhancing options/capabilities bitfield would be better or provide a hook for usermods to advertise their presence.

This is where source repo and release name should be used, not expecting update apps to try and interpret. Please make any further comments on the appropriate PR/issue

wled00/set.cpp
netmindz: Yes it is. If this PR is that you know if the settings value, then you need to send the new value when you save

blazoncek:
> Parsing the page looking for the pin entry feels very hacky and fragile

It is not. What if I *know* the PIN and want to enter it while updating?

> Please can you share a link as where this is documented

[KB](https://kno.wled.ge/interfaces/json-api/) see `opt` field. It says "for debugging" but is available (and filled with options) nevertheless. That field is available and documented as such since pre 0.11. Does it help other developers? It does not as such but it only requires updated documentation.

> I'm not aware of why the PS would have this affect, please add info to the appropriate PR

https://discord.com/channels/473448917040758787/779395228624617512/1324429991101862049

I merely wanted to express my thoughts and shed a different light on the topic. Nothing more.

wled00/json.cpp
netmindz: As the docs do not explain how you read this info out of that field, can you at least explain here how to read that value

netmindz:
> > Parsing the page looking for the pin entry feels very hacky and fragile
> 
> It is not. What if I _know_ the PIN and want to enter it while updating?

I thought the question we are trying to answer is _if_ a pin is required. So either the API supplies that info, as I belive is the intention here - correctly me if I have misunderstood @willmmiles. If I am understanding your alternative proposal correctly @blazoncek you are suggesting that we do not provide that via the API but instead  @Moustachauve has to try and grab the HTML of the UI and see if there is an entry box for the PIN. That is both really nasty and also fragile

> > Please can you share a link as where this is documented
> 
> [KB](https://kno.wled.ge/interfaces/json-api/) see `opt` field. It says "for debugging" but is available (and filled with options) nevertheless. That field is available and documented as such since pre 0.11. Does it help other developers? It does not as such but it only requires updated documentation.

Are you able to provide the updates to the doc that actually explain what that field is and how to use it if we are now intending 3rd party developers to use this rather than it just being there for our own internal debugging?

> > I'm not aware of why the PS would have this affect, please add info to the appropriate PR
> 
> https://discord.com/channels/473448917040758787/779395228624617512/1324429991101862049 https://discord.com/channels/473448917040758787/779395228624617512/1324422249603928169
> 
> I merely wanted to express my thoughts and shed a different light on the topic. Nothing more.

Yes I understand that, my point is that we need to be better organised and ensure we write our comments against the correct issue rather than go off on a tangent. Not everyone will be reading every issue/PR and even if they do, it makes it harder to keep track of items if when you go back to look at the issue/PR where you expect to see the comment and then can't find it as it was placed elsewhere

wled00/set.cpp
Moustachauve: Indeed, it would allow an app using websockets to update its state right away

Moustachauve:
WLED Native ideally would need only the api (and in the future, websocket api) to keep the state up to date.

I would like to not offer OTA to people who can't have updates (including PIN for now). This means that without this API field, I would have to send a request to the update page every 15 seconds to know if it's enabled now.


I don't have a strong preference for this, as it's not the most important use case, but it would be nice if everything was doable through an API and if all the settings/status of the device could be read/written through a standard API. This felt like a useful addition

blazoncek:
1st off I did not create `opt` field, nor document it. I merely noticed its content and provided info here.
See: https://github.com/Aircoookie/WLED/blob/61b99471a14db706681a6a1680fb5018c30720a2/wled00/json.cpp#L782-L811

Regarding availability of OTA: The information is in `opt` field as `opt & 0x0001`. If that evaluates as `true` (or >0) then OTA is available. If not, OTA is unavailable and calling `/update` (POST or GET) will yield 501.

Regarding PIN and OTA lock: If you try to call `/update` (POST with data for binary) and device is OTA locked you'll get 401, but PIN entry (HTTP) when PIN is used and is locked. If you call `/update` via GET and PIN is locked you'll be prompted with PIN entry as well. When PIN is unlocked (or not used) `/update` will behave as expected.

I am merely describing behaviour, don't want to suggest anything.
However I do feel that cramming too much into JSON will inevitably lead to removing support for ESP8266. This is something I do not want.

netmindz:
Thank you for the further info @blazoncek

softhack007:
@willmmiles should we keep this PR "alive", or is it obsolete already?

softhack007:
@coderabbitai please review.

Create these summary lists in addition to your review:

willmmiles:
> @willmmiles should we keep this PR "alive", or is it obsolete already?

I put the work in, I might as well go back through it and address the review concerns.  The basic objective was to allow a UI to avoid round-tripping with the device to find out if the locks were active; I believe this is still an outstanding concern for integrators.  I'll rebase and patch it up.

Moustachauve:
This could kinda be useful for https://github.com/frenck/python-wled/issues/1900 making OTA possible in HA when a password is required. Could be useful for WLED app too, but I think we would need some more work to support providing the password.

willmmiles:
> This could kinda be useful for [frenck/python-wled#1900](https://github.com/frenck/python-wled/issues/1900) making OTA possible in HA when a password is required. Could be useful for WLED app too, but I think we would need some more work to support providing the password.

Yes.  One of my concerns about the approach in this PR is that it essentially broadcasts "hey, my security is disabled" to all clients.

As an alternative, I could look in to implementing some other way of passing the password on a request by request basis (basic auth maybe?).  It'd still all be in cleartext for anyone with a network sniffer, but that's no worse than the current approach.

PR #4284, opened 2024-11-15
Keep selected FX while searching

blazoncek:
Requested on Discord.
@w00000dy please review if you can.

### DIFF
--- a/wled00/data/index.js
+++ b/wled00/data/index.js
@@ -2850,7 +2850,7 @@ function search(field, listId = null) {
 		if (listId!=='pcont' && i===0) return;
 		const listItemName = listItem.querySelector('.lstIname').innerText.toUpperCase();
 		const searchIndex = listItemName.indexOf(field.value.toUpperCase());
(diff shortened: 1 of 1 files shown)


w00000dy (approved):
I agree with this in general, but I personally would like to see the selected option at the end of the search results instead of at the top. Because I think it is more important to see the search result at the top than the selected effect. If you agree with me, you can merge my PR blazoncek/WLED#46.

willmmiles:
> I agree with this in general, but I personally would like to see the selected option at the end of the search results instead of at the top. Because I think it is more important to see the search result at the top than the selected effect. If you agree with me, you can merge my PR [blazoncek#46](https://github.com/blazoncek/WLED/pull/46).

I concur with @w00000dy - keeping the selected FX at the bottom by the controls makes more sense to me.

blazoncek:
Adding blazoncek#46 was beyond my skill and understanding. Thanks @w00000dy .

PR #4286, opened 2024-11-15
xml comma bugfix + few string optimisations

blazoncek:
fix for incorrect JSON array creation.

### DIFF
--- a/wled00/xml.cpp
+++ b/wled00/xml.cpp
@@ -83,7 +83,7 @@ void appendGPIOinfo(Print& settingsScript) {
   // usermod pin reservations will become unnecessary when settings pages will read cfg.json directly
   if (requestJSONBufferLock(6)) {
     // if we can't allocate JSON buffer ignore usermod pins
-    JsonObject mods = pDoc->createNestedObject(F("um"));
+    JsonObject mods = pDoc->createNestedObject("um");
     UsermodManager::addToConfig(mods);
     if (!mods.isNull()) fillUMPins(settingsScript, mods);
     releaseJSONBufferLock();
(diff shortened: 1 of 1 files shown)


netmindz:
Can you please provide a little more info on what what the impact of this issue was so others know how to test your fix and also establish if it needs to be merged before the release candidate is tagged

blazoncek:
Sorry.
The JSON array for reserved pins (settings page) was incorrectly constructed (wrong order of commas) in some cases (Ethernet). It should be included in 0.15.0 release.
I also added a few string optimisations to reduce image size.

netmindz:
Thank you for that further information, but I must confess that I am still unclear on what was wrong or how to reproduce the error. I find examples really helpful

willmmiles:
> Thank you for that further information, but I must confess that I am still unclear on what was wrong or how to reproduce the error. I find examples really helpful

The bug is a code inconsistency on whether the commas in the reserved pin list ought to be added first or last.  The "disallowed" pins added commas after, but other cases (DMX, hardwareTx, wired Ethernet) would add a comma first, so the list would end up reading `d.rsvd=[1,2,,3,4,5]`  -- note extra comma in the middle.  To reproduce the issue, make a build with WLED_ENABLE_DMX, WLED_DEBUG, or WLED_USE_ETHERNET and inspect the 
settings output, eg. http://wled.local/settings/s.js?p=2

That said: while not technically correct, in my prior testing, the UI did seem to handle the empty element gracefully.

wled00/xml.cpp
```diff
@@ -637,7 +642,7 @@ void getSettingsJS(byte subPage, Print& settingsScript)
     #if defined(ARDUINO_ARCH_ESP32)
       ESP.getChipModel(),
     #else
-      F("esp8266"),
+      "esp8266",
```
willmmiles, line 645: IMO we should be going the other way around on ESP8266 -- better to waste ROM with multiple copies than even a little RAM.

...I should really look in to a better solution; I'm 75% confident that there's some linker trickery that could be done to make a version of `F("string")` that also coalesces duplicates.

wled00/xml.cpp
```diff
@@ -152,6 +159,7 @@ void getSettingsJS(byte subPage, Print& settingsScript)
   DEBUG_PRINTF_P(PSTR("settings resp %u\n"), (unsigned)subPage);
 
   if (subPage <0 || subPage >10) return;
+  char nS[32];
```
willmmiles, line 162: Did you find this reduced the overall function stack usage?  Otherwise my preference runs to the more narrowly scoped variables.

wled00/xml.cpp
```diff
@@ -142,7 +149,7 @@ void appendGPIOinfo(Print& settingsScript) {
   // add info about max. # of pins
   settingsScript.print(F("d.max_gpio="));
   settingsScript.print(WLED_NUM_PINS);
-  settingsScript.print(F(";"));
+  settingsScript.print(';');
```
willmmiles, line 152: A `printf_P()` call here would probably reduce code size.

willmmiles (commented):
Generally looks good to me and passes my functional tests.

wled00/xml.cpp
blazoncek: "esp8266" is actually used elsewhere in other libraries. As it is not stored in PROGMEM linker reuses it and only one instance is used throughout.

wled00/xml.cpp
blazoncek: just removes annoying "unused variable" warning

wled00/xml.cpp
blazoncek: indeed

wled00/xml.cpp
netmindz: Are you ok to approve as it stands or can you update so we can resolve this issue @willmmiles ?

willmmiles: Oh sorry, I thought I'd flagged this as resolved already!  I'm happy to accept as is and approach string RAM savings later in a global review or with better tools.

Issue #4280, opened 2024-11-14
Various usability problems in the LEDs settings page - Type not changeable, GPIO, Length and mA/LED settings sometimes lost upon save


Fonzi03:
### What happened?

After saving, only ports 1-3 applied properly. Ports 4-8 (in my instance) changed from the following settings:
mA/LED: 55mA (default setting)
Data GPIO: <GPIO numbers>

changed to:
mA/LED: Custom
Data GPIO: 0

### To Reproduce Bug

WT32-ETH01
 (installed ethernet version)

flash a new board, apply wifi settings. 
navigate to settings --> Usermods
Under Digitalmic:
     Change Pin I2S WS to unused from default
     Change Pin I2S SCK to unused from default
     Save
navigate to settings --> LED Preferences 
Add 8 strings ( I used sk6812 )
     Each string set to Length: 1

GPIO in order of ports 1-8 :
2, 4, 12, 14, 15, 17, 5, 33

Save

Go back to LED Preferences and the mA/LED, Length (sometimes only some ports are affect not all, never 1-3), and Data GPIO ports are not as they were set previously. 

Removing the strings and re-adding them saving again seems to have a similar issue where mA/LED clears but GPIO assignments stay. 

Change the incorrect mA/LED from custom back to the default 55mA
Save

Now settings remain

Start changing the length of strings
1: 145
Save

2: 283
Save

3: 55
Save
     Crash

Rebooting board does nothing. 
The only recovery option at this point is to re-flash

This is repeatable. (except sometimes the GPIO stays and sometimes does not, every time it loses the GPIO it only loses them after port 3)

I first started with putting the appropriate pixels and strings in order then save but to test for reporting the bug I did them individually and only 1 pixel each. 

This renders the software/ board unusable for me as I cannot get it set to the strings that I need it consistently locks up the board. 

This does not occur in 14.4 (however applying the above settings, then selecting the Ethernet type in 14.4 it goes into the same bricked state) 

### Expected Behavior

I can set all pixel settings without issues. 

### Install Method

Binary from WLED.me

### What version of WLED?

0.15.0-b7

### Which microcontroller/board are you seeing the problem on?

ESP32

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

k7bbr:
I compiled 0.15.0-b7 for Ethernet and tried it on a WT32 board. Same
result. When changing to other protocols like SK6812 or TM1814 the WT32
[...]

Issue #4277, opened 2024-11-13
Disabling brightness limiter requires additional changes before saving


Fonzi03:
### What happened?

Cannot save after disabling automatic brightness

### To Reproduce Bug

If you go to LED Preferences and de-select "Enable automatic brightness limiter" and try to save without making any other changes you cannot save. Change any value anywhere else and you can save. 

### Expected Behavior

you should be able to save after changing a setting. 

### Install Method

Binary from WLED.me

### What version of WLED?

2410270

### Which microcontroller/board are you seeing the problem on?

ESP32

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

softhack007:
As this affects the LEDs settings page, it would be good to understand what's going on before 0.15.0-RC1 https://github.com/Aircoookie/WLED/pull/4287

Fonzi03:
> As this affects the LEDs settings page, it would be good to understand what's going on before 0.15.0-RC1 https://github.com/Aircoookie/WLED/pull/4287

If you need me to test anything or provide more details I'm more than happy to do so.

DedeHai:
I just tried this and and saving works. 
@Fonzi03  did you try this on a fresh install or what is the history? Did you try clearing local browser storage?

Fonzi03:
> I just tried this and and saving works. 
> @Fonzi03  did you try this on a fresh install or what is the history? Did you try clearing local browser storage?

I don't remember if when this occured it was on a fresh install or not. I was having multiple different issues at the time and flashed a handful of times. If it can't be reproduced it's likely it was a secondary order issue. 

I can test again but probably not for another week or so.

dosipod:
If by "requires additional changes before saving" meaning it will not save unless you enter the "Maximum PSU Current" and next time also enter the "mA/LED" then yes we have reported that among others multiple times . Mostly in discord but I think it was seen as not important 

Uploading pref_no_save.mov…

Fonzi03:
> If by "requires additional changes before saving" meaning it will not save unless you enter the "Maximum PSU Current" and next time also enter the "mA/LED" then yes we have reported that among others multiple times . Mostly in discord but I think it was seen as not important 
> 
> Uploading pref_no_save.mov…
> 
> 

Precisely the issue, except it's not limited to mA/LED you just have to change literally any other value on the page for it to them allow saving.

dosipod:
[...]

Issue #4260, opened 2024-11-08
In 0.15.0-b7 and -b6, the "release" is always returned as `1`


Moustachauve:
### What happened?

Release is always `1` instead of `ESP32` or whatever should be needed.

### To Reproduce Bug

1) Download the release version from Github for 0.15.0-b7-ESP32.bin
2) flash it to an esp32
3) go to [ipAddress]/json/info
4) notice the `release` field is always `1`

### Expected Behavior

the `release` field should be `ESP32`.

### Install Method

Binary from WLED.me

### What version of WLED?

WLED 0.15.0-b7 1 (ESP32-D0WDQ5 build 2410270)

### Which microcontroller/board are you seeing the problem on?

ESP32

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Moustachauve:
I didn't try just manually building the release on my side, maybe they just need to be rebuilt with the correct flags in platformIO? Maybe the person that built them had an override or something.

softhack007:
Hi @Moustachauve,

I've just tried it, and indeed the firmware `WLED_0.15.0-b7_ESP32.bin` shows a wrong release string:

![image](https://github.com/user-attachments/assets/94fa423a-127f-47ba-9509-d0979ad4daa3)
```JSON
{
    "ver": "0.15.0-b7",
    "vid": 2410270,
    "cn": "Kōsen",
    "release": "1",
```

<br/>

While `WLED_0.15.0-b7_ESP32_WROVER.bin` has the correct info

![image](https://github.com/user-attachments/assets/de854d2a-c951-4b20-b833-05982821616c)


```JSON
{
    "ver": "0.15.0-b7",
    "vid": 2410270,
    "cn": "Kōsen",
    "release": "ESP32_WROVER",
```

netmindz:
It is based on the WLED_RELEASE_NAME define, which is also used to define the name of the output bin by reading this define with _get_cpp_define_value in output_bins.py

Therefore I think the value is being set correctly as the define, but is failing to then be properly read into releaseString in wled.h in some cases. 

I don't think it's anything to do with xml.cpp as all the version and release is single replacement

     `d.getElementsByClassName("sip")[0].innerHTML = "WLED 0.15.0-b7<br>1<br>(ESP32-D0WDQ6 build 2410270)";
`

netmindz:
I think I've worked out what is wrong, just confirming fix

netmindz:
[...]

Issue #4241, opened 2024-11-01
OTA Updating ESP32-S2 from  0.15.0-b4 fails - if Tasmota Web installer was used


espilioto:
### What happened?

I have 2 instances of wled running on a ESP32-S2 that updated fine up to 0.15.0-b4.
Updated them through Home Assistant actually, everything rosy.

The problem is that they can't be updated to any version after that through OTA, as both the web interface and HA fail.
I have a third wled running on a 8266 and it updated fine manually and through HA just now to 0.15.0-b7 actually.

Any ideas?
Is plugging 'em in and reflashing my only choice?

Thanks in advance <3

### To Reproduce Bug

Upload binary, WLED_0.15.0-b7_ESP32-S2.bin in my case and press the update button.
Alternatively through the update dialog in HA.

### Expected Behavior

The web interface times out.
HA fails with this message: Failed to perform the action update/install. Error communicating with WLED API

### Install Method

Binary from WLED.me

### What version of WLED?

 0.15.0-b4 

### Which microcontroller/board are you seeing the problem on?

ESP32-S2

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

DedeHai:
I just uploaded the 0.15.0-b7 OTA to an S2, works fine. 
can you backup your config, your presets, then do a factory reset and try to OTA again through web UI?

espilioto:
Hey, thanks for the reply. 

Nope, the reset happened but still can't OTA update.

DedeHai:
Did you try a fresh 0.14 install and then update it to 0.15 B7?

blazoncek:
S2, if used with MQTT or other functions may be low on free heap. This may prevent OTA.

espilioto:
> Did you try a fresh 0.14 install and then update it to 0.15 B7?

Tried downgrading to 0.14.4 via OTA but nothing. Obviously I'm trying to avoid manual flashing.

> S2, if used with MQTT or other functions may be low on free heap. This may prevent OTA.
Hmm, can I do something about that? Maybe disable something in the settings?

espilioto:
> Did you try a fresh 0.14 install and then update it to 0.15 B7?

Tried downgrading to 0.14.4 via OTA but nothing. Obviously I'm trying to avoid manual flashing.
 
> S2, if used with MQTT or other functions may be low on free heap. This may prevent OTA.

Hmm, can I do something about that? Maybe disable something in the settings?

[...]

PR #4287, opened 2024-11-15
0.15.0-rc1

netmindz:


willmmiles:
I recommend prioritizing a fix for #4260 before moving to RC, especially since it pertains to version identity data.

netmindz:
Thank you @willmmiles for bringing https://github.com/Aircoookie/WLED/issues/4260 to my attention, yes that will need to be fixed before the Release Candidate

blazoncek:
IMO #4260 was a build glitch. IDK how it happened (it shouldn't).
The easiest way to check is build, upload and verify.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## WLED changelog
 
+#### Build 2411150
+-   WLED 0.15.0-rc1 release
```
blazoncek, line 4: When making a release, update with all PRs and commits that happened after b7.

blazoncek (approved):
If there are any commits or PRs merged add them to change log.

softhack007:
I'm seeing a few new bug reports related to the LEDs setting UI. Maybe we should analyze them before going g to RC.
* https://github.com/Aircoookie/WLED/issues/4277
* https://github.com/Aircoookie/WLED/issues/4280

Plus this one that seems to affect all S2 users
* https://github.com/Aircoookie/WLED/issues/4241

blazoncek:
> [Disabling brightness limiter requires additional changes before saving #4277](https://github.com/Aircoookie/WLED/issues/4277)

I could not reproduce. Perhaps requesting browser logs is in order.

> [Ports >3 lose GPIO, Length and mA/LED settings upon save #4280](https://github.com/Aircoookie/WLED/issues/4280)

May be related to limiter. I could not reproduce. Same submitter as previous. Perhaps issue on his/her setup.

> [Crashes when leds length is 3 #4120](https://github.com/Aircoookie/WLED/issues/4120)

This one should be easy for anyone to reproduce if true. I could not.

> [OTA Updating ESP32-S2 from  0.15.0-b4 fails  #4241](https://github.com/Aircoookie/WLED/issues/4241)

I have 3 S2 units in production use (>400 pixels, multirelay, PIR sensor, temprature usermods, MQTT), all updated OTA regularly with no issues. Same with @DedeHai . @willmmiles did see issues though. Probably device FS issue.

I am sure that in all cases debug log and crash dumps with exception decoder are needed to diagnose. None were provided since submitting.

k7bbr:
> Disabling brightness limiter requires additional changes before saving
> #4277 <https://github.com/Aircoookie/WLED/issues/4277>
>
> I could not reproduce. Perhaps requesting browser logs is in order.
>
> Ports >3 lose GPIO, Length and mA/LED settings upon save #4280
> <https://github.com/Aircoookie/WLED/issues/4280>
>
> May be related to limiter. I could not reproduce. Same submitter as
> previous. Perhaps issue on his/her setup.
>
I easily reproduced this issue when flashing a new, out-of-the box WT32.
Unable to set pins or pixel type. The module was completely bricked for
0.15.x even after USB flashing to 0.14.4 and the flashing again to 0.15.x.
not just a single user.

willmmiles:
Re #4277 and #4280: I've seen something in this vein (LED settings not saving) on my test setup.  I was having a bunch of unrelated connectivity issues, so I just wrote it off at the time.  Trying to reproduce it now.

Re #4241: I'm 95% confident that what's happening is heap starvation of some kind.  S2s have practically as little RAM as 8266es, but all the code (including the platform!) treats them like the bigger ESP32s.  I could also reliably trigger a heap exhaustion crash by messing up the upload parameters so it didn't get handed to `Uploader` properly.  I'm going to probe to see if there's something I can do in AsyncWebServer to trap that case and not crash, at least.  (Doesn't answer why ArduinoOTA and other non-webserver uploads also fail for some users, though.)

netmindz:
Ok, it wasn't my original intention with this PR, but are we ok with adopting this as the ### issue where we track items that must be resolved before the RC?

blazoncek:
@k7bbr please provide crash dumps or debug logs in respective issues, not here.

@willmmiles try ArduinoOTA for S2. Uses different route.

softhack007:
I've just added some bug reports where i feel we should understand what's going on, before going forward to release RC1.
It seems like these bugs are "clustered" around
* Problems with the LEDs settings page
* Unexplained crashes: DMX-out; mixing RGBW and RGB leds
* Segment-related usability problems (maybe be non-blocking if we understand them better)

blazoncek:
@softhack007 I do agree with releasing as thoroughly tested code as possible but if you look in the past, b5 was officially released at the beginning of September and there was really sparse feedback on it (or prior to that, b6 and b7 included minor changes and received same amount of feedback).

IMO if you want to get enough feedback, release 0.15.0, gather feedback from real use and provide 0.15.1 ASAP based on the feedback. Otherwise you'll be stuck in *release limbo* in which WLED is for far too long already IMO.

No software, in all my 40+ years of experience, is released bugfree. Ever.

softhack007:
> IMO if you want to get enough feedback, release 0.15.0, gather feedback from real use and provide 0.15.1 ASAP based on the feedback. Otherwise you'll be stuck in _release limbo_ in which WLED is for far too long already IMO.
> 
> No software, in all my 40+ years of experience, is released bugfree. Ever.

@blazoncek I think we already have enough feedback to fix issue before "pushing the thing out". I agree with you that no software is bug-free. However from my also 35+ in software, and 20 years of this in software quality, I know that two things are very important if you don't want to lose trust of users:
* no egressions - what worked before should still work after installing a new release. Unless there are clear warnings in release notes.
* good first experience - so features like settings up LEDs or managing segments should not get you stuck (cannot save) or lead to crashes of loss of settings.

After that comes the "wow" about new stuff that comes in a new release. And yes, users will forgive errors and "glitches" in   newly added features. But on the other side, they may regret installing an update, if it means they cannot continue working on the Christmas Lights because missing a usb cable to flash the old software...


Just my "words of reason" - every maintainer should have an opinion and finally we will all decide together when to release. It doesn't mean what I'm saying is the only truth in the world 🤷

netmindz:
Fix for https://github.com/Aircoookie/WLED/issues/4260 ready for review

blazoncek:
Just a reminder:
There are 90 PRs open waiting to be merged (either for 0.15.1 or 0.16) and we are debating if we add quotes or not to WLED_RELEASE_NAME while the simplest solution is on hand.

And then there's this:
Screenshot 2024-11-25 at 22 07 46
Long overdue now.

Has anyone tried to reproduce issues @softhack007 pointed out? I was not able to do so.

#4260 just needs a replacement of ESP8266 and ESP32 to something not recursively expanding (see my suggestion and reasoning above and the following screenshot)
Screenshot 2024-11-25 at 22 37 44

#4120 has been resolved by @willmmiles 
#4241 has been most likely caused by Tasmota partition map (three of us were unable to reproduce)
what about #4277 & #4280? @netmindz pointed out (and my own observation) that that particular user opened a few issues, all at once, with no detailed feedback when requested

I am sorry if I sound and behave like a big PITA but time is so short now that if we do not release 0.15 before December it may be pointless to release it as-is and better go on developing and merging PRs.

softhack007:
> I am sorry if I sound and behave like a big PITA but time is so short now that if we do not release 0.15 before December it may be pointless to release it as-is and better go on developing and merging PRs.

Ahem... yes ... as you say - you sound like a manager who knows nothing but schedule.

It's good that some of the issues I pointed out are now clarified, and for sure we can decide to address some points later.

I'm not sure what you want to sag by quoting AC - so we should simply release because @Aircoookie would love to see a release in November ? C'mon give me a break - or are you just joking?

CHANGELOG.md
netmindz: Done

netmindz:
There is a LOT we need to do with regard to PR management and the actual release process, it's way too fragmented and labour-intensive as it stands, but we just need to get 0.15.0 out, get the new process and CI in place.

I've just updated the RC1 with what we have merged so far we still have a few open issues we need to decide one way or the other with.

A few items were labelled as being for the final release we haven't discussed and a few we haven't, some can perhaps wait for the GA release of 0.15.0

https://github.com/Aircoookie/WLED/milestone/12

blazoncek:
#4300 fixed and waiting in PR

PR #4290, opened 2024-11-17
Custom quick colors

colebot17:
This feature allows custom quick colors to be added from the User Interface settings page.

A string setting holds comma-separated hex values in local storage. When the UI is updated, each color is added as a button underneath the existing colors. With lots of colors, the number of colors per row matches the existing alternating row pattern.

Testing:
- Invalid hex codes are not added as buttons
- Extra spaces around the commas have no effect

Limitations:
- The user must edit the string to change the colors
- There is no color picker UI available


netmindz:
Please set the target branch to 0_15 on your merge request

PR #4291, opened 2024-11-17
Add a setting to create custom quick color buttons

colebot17:
This feature allows custom quick colors to be added from the User Interface settings page.

A string setting holds comma-separated hex values in local storage. When the UI is updated, each color is added as a button underneath the existing colors. With lots of colors, the number of colors per row matches the existing alternating row pattern.

Testing:
- Invalid hex codes are not added as buttons
- Extra spaces around the commas have no effect

Limitations:
- The user must edit the string to change the colors
- There is no color picker UI available


DedeHai:
Thank you for the contribution, this is a nice feature, however I do not see it fit for merging as a standard feature:
- colors are stored locally, meaning only available on the device you created them, correct?
- not using the color picker is not user friendly

colebot17:
I went ahead and updated this to fix the issues.

- Colors are now defined on the led settings page, and they can be added and removed much like led buses and color order mappings. They take the form of color inputs, leveraging the default browser color picker.
- Custom colors are now stored in cfg.json, making them common across browsers

Issue #4293, opened 2024-11-19
"magic numbers" for MQTT topic buffer length in mqtt.cpp and button.cpp (use already defined MQTT_MAX_TOPIC_LEN)


WouterGritter:
### What happened?

I was modifying the source code on a fork for my custom MQTT topic structure and noticed `MQTT_MAX_TOPIC_LEN` is not used in all places the associated variables `mqttDeviceTopic` and `mqttGroupTopic` are used.

When increasing this defined variable, buffer overflows will occur when trying to generate topics to publish on in `mqtt.cpp` and `button.cpp`. I personally organize all MQTT devices in a `<building>/<room>/<device>` topic structure, and this could potentially be more than the current 32 max topic length.

Currently, when trying to increase this variable, the following things need to be changed:
- MQTT_MAX_TOPIC_LEN
- lots of buffer allocation code in `mqtt.cpp` and `button.cpp`
- `maxlength` attribute on `<input>` elements in `settings_sync.html`

It's bug-prone to have to manually adjust the buffer sizes in both cpp files. However it might be nice to have to also automatically populate the `maxlength` attribute on the `<input>` element. Not sure if this easy though.

---

_Not applicable headers removed._

### What version of WLED?

0.15.0-b7

### Anything else?

~If no one else is one this after a while I'll try to submit a PR myself. Bare with me if I do that, because I've never contributed to open source before!~

I'm submitting a PR for the fix.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

WouterGritter:
Opened a PR for this.

WouterGritter:
!4295 uses `MQTT_MAX_TOPIC_LEN` to allocate topic string buffers.

> Currently, when trying to increase this variable, the following things need to be changed:
> - **`maxlength` attribute on `<input>` elements in settings_sync.html**

This was already implemented but the feature was broken. !4296 fixes this.

blazoncek:
You can close this now.

WouterGritter:
![4295](https://github.com/Aircoookie/WLED/pull/4295) (the one that fixes magic numbers) is still open. Two PRs were spawned from this issue.

PR #4295, opened 2024-11-20
FIX: "magic numbers" for MQTT topic buffer length in mqtt.cpp and button.cpp (use already defined MQTT_MAX_TOPIC_LEN)

WouterGritter:
See issue #4293.

As seen in `wled.h`, `MQTT_MAX_TOPIC_LEN` is currently defined to be `32`. All modified lines of code result in the same value by adding a number to `MQTT_MAX_TOPIC_LEN`, allowing this variable to be increased without the risk of buffer overflows at a later date or by someone compiling the project with the modification themselves.

### DIFF
--- a/wled00/button.cpp
+++ b/wled00/button.cpp
@@ -29,7 +29,7 @@ void shortPressAction(uint8_t b)
 #ifndef WLED_DISABLE_MQTT
   // publish MQTT message
   if (buttonPublishMqtt && WLED_MQTT_CONNECTED) {
-    char subuf[64];
+    char subuf[MQTT_MAX_TOPIC_LEN + 32];
     sprintf_P(subuf, _mqtt_topic_button, mqttDeviceTopic, (int)b);
     mqtt->publish(subuf, 0, false, "short");
   }
@@ -62,7 +62,7 @@ void longPressAction(uint8_t b)
 #ifndef WLED_DISABLE_MQTT
   // publish MQTT message
   if (buttonPublishMqtt && WLED_MQTT_CONNECTED) {
-    char subuf[64];
+    char subuf[MQTT_MAX_TOPIC_LEN + 32];
     sprintf_P(subuf, _mqtt_topic_button, mqttDeviceTopic, (int)b);
     mqtt->publish(subuf, 0, false, "long");
   }
@@ -83,7 +83,7 @@ void doublePressAction(uint8_t b)
 #ifndef WLED_DISABLE_MQTT
   // publish MQTT message
   if (buttonPublishMqtt && WLED_MQTT_CONNECTED) {
-    char subuf[64];
+    char subuf[MQTT_MAX_TOPIC_LEN + 32];
     sprintf_P(subuf, _mqtt_topic_button, mqttDeviceTopic, (int)b);
     mqtt->publish(subuf, 0, false, "double");
   }
@@ -151,7 +151,7 @@ void handleSwitch(uint8_t b)
 #ifndef WLED_DISABLE_MQTT
     // publish MQTT message
     if (buttonPublishMqtt && WLED_MQTT_CONNECTED) {
-      char subuf[64];
+      char subuf[MQTT_MAX_TOPIC_LEN + 32];
       if (buttonType[b] == BTN_TYPE_PIR_SENSOR) sprintf_P(subuf, PSTR("%s/motion/%d"), mqttDeviceTopic, (int)b);
       else sprintf_P(subuf, _mqtt_topic_button, mqttDeviceTopic, (int)b);
       mqtt->publish(subuf, 0, false, !buttonPressedBefore[b] ? "off" : "on");
--- a/wled00/mqtt.cpp
+++ b/wled00/mqtt.cpp
@@ -7,6 +7,10 @@
 #ifndef WLED_DISABLE_MQTT
 #define MQTT_KEEP_ALIVE_TIME 60    // contact the MQTT broker every 60 seconds
 
+#if MQTT_MAX_TOPIC_LEN > 32
+#warning "MQTT topics length > 32 is not recommended for compatibility with usermods!"
+#endif
+
 static void parseMQTTBriPayload(char* payload)
 {
   if      (strstr(payload, "ON") || strstr(payload, "on") || strstr(payload, "true")) {bri = briLast; stateUpdated(CALL_MODE_DIRECT_CHANGE);}
--- a/wled00/mqtt.cpp
+++ b/wled00/mqtt.cpp
@@ -23,24 +23,24 @@ static void parseMQTTBriPayload(char* payload)
 static void onMqttConnect(bool sessionPresent)
 {
   //(re)subscribe to required topics
-  char subuf[38];
+  char subuf[MQTT_MAX_TOPIC_LEN + 6];
 
   if (mqttDeviceTopic[0] != 0) {
-    strlcpy(subuf, mqttDeviceTopic, 33);
+    strlcpy(subuf, mqttDeviceTopic, MQTT_MAX_TOPIC_LEN + 1);
     mqtt->subscribe(subuf, 0);
     strcat_P(subuf, PSTR("/col"));
     mqtt->subscribe(subuf, 0);
-    strlcpy(subuf, mqttDeviceTopic, 33);
+    strlcpy(subuf, mqttDeviceTopic, MQTT_MAX_TOPIC_LEN + 1);
     strcat_P(subuf, PSTR("/api"));
     mqtt->subscribe(subuf, 0);
   }
 
   if (mqttGroupTopic[0] != 0) {
-    strlcpy(subuf, mqttGroupTopic, 33);
+    strlcpy(subuf, mqttGroupTopic, MQTT_MAX_TOPIC_LEN + 1);
     mqtt->subscribe(subuf, 0);
     strcat_P(subuf, PSTR("/col"));
     mqtt->subscribe(subuf, 0);
-    strlcpy(subuf, mqttGroupTopic, 33);
+    strlcpy(subuf, mqttGroupTopic, MQTT_MAX_TOPIC_LEN + 1);
     strcat_P(subuf, PSTR("/api"));
     mqtt->subscribe(subuf, 0);
   }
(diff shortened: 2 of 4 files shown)


wled00/button.cpp
```diff
@@ -29,7 +29,7 @@ void shortPressAction(uint8_t b)
 #ifndef WLED_DISABLE_MQTT
   // publish MQTT message
   if (buttonPublishMqtt && WLED_MQTT_CONNECTED) {
-    char subuf[64];
+    char subuf[MQTT_MAX_TOPIC_LEN + 32];
```
softhack007, line 32: maybe this should be `MQTT_MAX_TOPIC_LEN * 2`  ? because we have two "topic" inputs: _mqtt_topic_button, mqttDeviceTopic

wled00/button.cpp
```diff
@@ -62,7 +62,7 @@ void longPressAction(uint8_t b)
...
```
softhack007, line 65: same as above

wled00/button.cpp
```diff
@@ -83,7 +83,7 @@ void doublePressAction(uint8_t b)
...
```
softhack007, line 86: see above

wled00/button.cpp
```diff
@@ -151,7 +151,7 @@ void handleSwitch(uint8_t b)
 #ifndef WLED_DISABLE_MQTT
     // publish MQTT message
     if (buttonPublishMqtt && WLED_MQTT_CONNECTED) {
-      char subuf[64];
+      char subuf[MQTT_MAX_TOPIC_LEN + 32];
```
softhack007, line 154: same

WouterGritter:
@softhack007 no it should be + 32, see lines ~150+ in `button.cpp`:
```cpp
    if (buttonPublishMqtt && WLED_MQTT_CONNECTED) {
      char subuf[MQTT_MAX_TOPIC_LEN + 32];
      if (buttonType[b] == BTN_TYPE_PIR_SENSOR) sprintf_P(subuf, PSTR("%s/motion/%d"), mqttDeviceTopic, (int)b);
      else sprintf_P(subuf, _mqtt_topic_button, mqttDeviceTopic, (int)b);
      mqtt->publish(subuf, 0, false, !buttonPressedBefore[b] ? "off" : "on");
    }
```

Here, it puts `mqttDeviceTopic` at the beginning of `/motion/`, which is followed by a digit which has nothing to do with the topic length. The only variable part (apart from the digit which is always reasonably short in length, and accounted for in the +32) is the `mqttDeviceTopic` string.

Elsewhere, in the same file it doesn't use the literal string `"%s/motion/%d"` but rather a variable `_mqtt_topic_button`, which is essentially the same as using the literal `"%s/button/%d"`, only converted to a variable (stored in PROGMEM) to omtimize flash storage (line 15):

```cpp
static const char _mqtt_topic_button[] PROGMEM = "%s/button/%d";  // optimize flash usage
```

WouterGritter:
I am now noticing other places that use the `mqttGroupTopic` rather than `mqttDeviceTopic` which also use hard-coded char buffer lengths, even though `mqttGroupTopic` is also of length `MQTT_MAX_TOPIC_LEN`. Will need to modify these buffer sizes too.

WouterGritter:
In `wled_eeprom.cpp` the topic length is currently also hard-coded to a length of 32, something which this PR would also need to modify:
```cpp
  if (lastEEPROMversion > 8)
  {
    readStringFromEEPROM(2300, mqttServer, 32);
    readStringFromEEPROM(2333, mqttDeviceTopic, 32);
    readStringFromEEPROM(2366, mqttGroupTopic, 32);
  }
```

Modifying this would corrupt the EEPROM config when upgrading to a version of WLED which has the max topic length increased. It would also mean calculating every hard-coded EEPROM address with an offset that's dependent on `MQTT_MAX_TOPIC_LEN`... Suddenly this seems really impractical. Thoughts?

softhack007:
> In `wled_eeprom.cpp` the topic length is currently also hard-coded to a length of 32, something which this PR would also need to modify:
> Modifying this would corrupt the EEPROM config when upgrading to a version of WLED which has the max topic length increased. It would also mean calculating every hard-coded EEPROM address with an offset that's dependent on `MQTT_MAX_TOPIC_LEN`... Suddenly this seems really impractical. Thoughts?

I think you don't need to worry much about EEPROM - this feature is deprecated, and the code is not active  unless  `WLED_ADD_EEPROM_SUPPORT` is defined by the buildenv. Today WLED's configuration is stored in cfg.json on LittleFS.


I think it would be best to just keep the EEPROM code as it is - just add a warning like this in wled_EEPROM.cpp (after #include "wled.h")
```c++
#if defined(WLED_ENABLE_MQTT) && MQTT_MAX_TOPIC_LEN > 32
#error "MQTT topics length > 32 is not supported by the EEPROM module!"
#endif
```

WouterGritter:
@softhack007 great! I just saw that the wled_eeprom.cpp file is only used for reading older settings indeed. I don't think a warning is even needed, we can just read it as a hardcoded 32-byte length and put it in the larger `mqttDeviceTopic` string. This would cause issues when `MQTT_MAX_TOPIC_LEN` is set to less than 32 though, maybe we can set a warning or add a comment not to set it lower than 32.

I think I'll continue this fix then. Will update the other files that touch mqttDeviceTopic/mqttGroupTopic tomorrow, and I'll resolve the threads about the `+ 32` as per my comment above.

WouterGritter:
All other references to `mqttDeviceTopic` and `mqttGroupTopic` are outside the `wled00` source folder, mainly in the usermods. I think fixing those is outside of the concern of this PR and issue.

I've added a comment explaining this, that's also warning against setting the value lower of `MQTT_MAX_TOPIC_LEN` than `32`. If you'll approve the PR it's ready for merging now!

`#define MQTT_MAX_TOPIC_LEN 32  // should not be less than 32. might cause trouble when increased with usermods active that do not handle this correctly.`

softhack007:
> #define MQTT_MAX_TOPIC_LEN 32 // should not be less than 32. might cause trouble when increased with usermods active that do not handle this correctly.

I would actually prefer a compile-time #warning added in mqtt.cpp.

WouterGritter:
I've replaced the comment with a warning and an error:
- LEN > 32: warning in MQTT that this might cause problems when using usermods.
- LEN < 32: error in eeprom (if mqtt enabled) that this is not supported. len < 32 will cause buffer overflow when writing eeprom data to memory. len > 32 does not matter as it will only fill a max. of 32 chars.

softhack007 (approved):
Thanks, looks good to me 
code reviewed but not tested from my side, as I don't have a MQTT broker setup at home.

I'd suggest we include this PR into the next feature update after 0.15.0 was released.

netmindz:
Thank you for putting this together, magic numbers are horrible at great to remove them. However I still see some magic numbers within your change. If these could be eliminated as well please 

E.g MQTT_MAX_TOPIC_LEN + 32

That is better than just 64, but you still have the magic number of 32 so it not really resolved the magic number issue

WouterGritter:
> Thank you for putting this together, magic numbers are horrible so great to remove them. However I still see some magic numbers within your change. If these could be eliminated as well please
> 
> E.g MQTT_MAX_TOPIC_LEN + 32
> 
> That is better than just 64, but you still have the magic number of 32 so it not really resolved the magic number issue

I get what you're saying, but IMO this is way less worse than the situation was previously. With these "magic numbers", you see the code that uses it within 10 lines of the usage of the additional buffer length. You see a variable which is the mqtt prefix, and a static string that's added to it or used as a format.

This is in contrast to the use of `MQTT_MAX_TOPIC_LEN`, which is used across multiple files and lots of places.

I'm also not sure as how to get rid of these effectively, storing the cstring that's added to the topic prefix, and calculating the length before generating a buffer that holds the composite of both strings seems overkill and would add (imo unnecessary) CPU time.

WouterGritter:
> Thanks, looks good to me code reviewed but not tested from my side, as I don't have a MQTT broker setup at home.
> 
> I'd suggest we include this PR into the next feature update after 0.15.0 was released.

For what it's worth coming from a first-time collaborator, the code works fine on my end. I have been controlling some WLED lights with a build where the max length was increased to 64 using this PR without any weirdness.

netmindz:
> > Thank you for putting this together, magic numbers are horrible so great to remove them. However I still see some magic numbers within your change. If these could be eliminated as well please
> > E.g MQTT_MAX_TOPIC_LEN + 32
> > That is better than just 64, but you still have the magic number of 32 so it not really resolved the magic number issue
> 
> I get what you're saying, but IMO this is way less worse than the situation was previously. With these "magic numbers", you see the code that uses it within 10 lines of the addition to the buffer length. You see the mqtt prefix variable, a static string, and a strcpy call to copy them both to the buffer.
> 
> This is in contrast to the use of `MQTT_MAX_TOPIC_LEN`, which refers to the max length of the `mqttDeviceTopic` variable which was used across multiple files and lots of places with the assumption that `MQTT_MAX_TOPIC_LEN` was always 32, even though it could be changed.
> 
> I'm also not sure as how to get rid of these effectively, storing the cstring that's added to the topic prefix, and calculating the length before generating a buffer that holds the composite of both strings seems overkill and would add (imo unnecessary) CPU time.

It's definitely an improvement 
Can you not just define another constant for what 32 actually refers to?

> > Thank you for putting this together, magic numbers are horrible so great to remove them. However I still see some magic numbers within your change. If these could be eliminated as well please
> > E.g MQTT_MAX_TOPIC_LEN + 32
> > That is better than just 64, but you still have the magic number of 32 so it not really resolved the magic number issue
> 
> 
> This is in contrast to the use of `MQTT_MAX_TOPIC_LEN`, which refers to the max length of the `mqttDeviceTopic` variable which was used across multiple files and lots of places with the assumption that `MQTT_MAX_TOPIC_LEN` was always 32, even though it could be changed.
> 
> I'm also not sure as how to get rid of these effectively, storing the cstring that's added to the topic prefix, and calculating the length before generating a buffer that holds the composite of both strings seems overkill and would add (imo unnecessary) CPU time.

It's definitely an improvement, can you not just define a constant with a nice name and a value of 32 and the same for your magic 6 ?

PR #4296, opened 2024-11-21
Introduce printSetInputMaxlength to properly set an inputs maxlength …

WouterGritter:
Relates to [!4295](https://github.com/Aircoookie/WLED/pull/4295) by fixing MQTT related code that deals with the `MQTT_MAX_TOPIC_LEN` variable.

Previously, the `maxlength` attribute on the `<input>` fields in `settings_sync.html` were set with `inputElement.maxlength = <length>`, however this did not seem to actually modify the `maxlength` attribute as seen in the screenshot (also I couldn't type any more characters than the previous maxlength setting):
image

As seen, modifying `element.maxlength` does actually modify the variable, but it does not modify the `maxlength` attribute weirdly. Using `element.setAttribute("maxlength", X)` was the fix.

In this PR I've modified the code to implement a function to use this JS line instead of the one previously used.

### DIFF
--- a/wled00/xml.cpp
+++ b/wled00/xml.cpp
@@ -465,7 +465,7 @@ void getSettingsJS(byte subPage, Print& settingsScript)
     printSetFormValue(settingsScript,PSTR("MG"),mqttGroupTopic);
     printSetFormCheckbox(settingsScript,PSTR("BM"),buttonPublishMqtt);
     printSetFormCheckbox(settingsScript,PSTR("RT"),retainMqttMsg);
-    settingsScript.printf_P(PSTR("d.Sf.MD.maxlength=%d;d.Sf.MG.maxlength=%d;d.Sf.MS.maxlength=%d;"),
+    settingsScript.printf_P(PSTR("d.Sf.MD.maxLength=%d;d.Sf.MG.maxLength=%d;d.Sf.MS.maxLength=%d;"),
                   MQTT_MAX_TOPIC_LEN, MQTT_MAX_TOPIC_LEN, MQTT_MAX_SERVER_LEN);
     #else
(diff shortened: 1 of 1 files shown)


WouterGritter:
Tagging you @softhack007 as you're aware of the context of the previous PR.

softhack007:
> Tagging you @softhack007 as you're aware of the context of the previous PR.

Hi,

sorry HTML/JS is not my area.

blazoncek:
If I was doing something like general method for modifying any attribute, I would have done it something like this:
```js
function upAttr(o,a,v)
{
	d.Sf[o].setAttribute(a,v);
}
```
```c++
settingsScript.printf_P(PSTR("upAttr('MD','maxlength',%d);upAttr('MG','maxlength',%d);upAttr('MS','maxlength',%d);"),
   MQTT_MAX_TOPIC_LEN, MQTT_MAX_TOPIC_LEN, MQTT_MAX_SERVER_LEN);
```

But particular case is only the issue of missing uppercase letter. Instead of `maxlength` JavaScript access needs property called `maxLength`.

So, according to W3C Schools, the solution with least change is:
```c++
settingsScript.printf_P(PSTR("d.Sf.MD.maxLength=%d;d.Sf.MG.maxLength=%d;d.Sf.MS.maxLength=%d;"),
              MQTT_MAX_TOPIC_LEN, MQTT_MAX_TOPIC_LEN, MQTT_MAX_SERVER_LEN);
```

WouterGritter:
Damn, great catch. I agree with making the least amount of changes required. Will replace maxlength with **maxLength** like you proposed.

WouterGritter:
I think we can merge this now.

blazoncek (approved):

netmindz (approved):

PR #4297, opened 2024-11-21
Intermediate color updates while moving around the color wheel

WouterGritter:
Push color changes when moving your finger/mouse around the color wheel every 500 ms, instead of only when the mouse/finger gets released from the color wheel.

This real-time color changing is similar to how Philips Hue does it in their app, and it makes it feel really nice and responsive.

The implementation almost seems TOO easy, it makes me think someone has tried this before and it did not get merged for some reason. Would love to hear your feedback.

One problem I'm seeing currently is that the C code does not handle it nicely when it gets a new color while it is transition from another color. The transition is smooth for the first update, but for the second update it jumps to the second color. Then for the next update it's smooth again, and it jumps again. Could be fixed by increasing the interval that the JS code sends updates. However it might be better to update the C code.

### DIFF
--- a/wled00/data/index.js
+++ b/wled00/data/index.js
@@ -276,7 +276,7 @@ function onLoad()
 	handleLocationHash();
 	cpick.on("input:end", () => {setColor(1);});
 	cpick.on("color:change", () => {updatePSliders()});
-	cpick.on("color:change", () => {updateIntermediateColor();
+	cpick.on("color:change", () => {updateIntermediateColor()});
 	pmtLS = localStorage.getItem('wledPmt');
 
 	// Load initial data
--- a/wled00/data/index.js
+++ b/wled00/data/index.js
@@ -26,6 +26,7 @@ var pmt = 1, pmtLS = 0, pmtLast = 0;
 var lastinfo = {};
 var isM = false, mw = 0, mh=0;
 var ws, wsRpt=0;
+var lastIntermediateColorUpdate = new Date();
 var cfg = {
 	theme:{base:"dark", bg:{url:"", rnd: false, rndGrayscale: false, rndBlur: false}, alpha:{bg:0.6,tab:0.8}, color:{bg:""}},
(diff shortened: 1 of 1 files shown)


DedeHai:
thanks for contributing. While I like the idea, the implementation is, as you found yourself, flawed. 
try what happens if you increase transition-time to 2 seconds: it will just jump from color to color every 500ms, which I think will loook awful in a lighting situation. To solve it, the code handling transitions would need a complete make-over to be able to handle "partial" transitions.

blazoncek:
... now imagine having transition time of 10s. 😉 

Also, ESP8266, even with all the bells and whistles @willmmiles added to AWS, this rapid bombardment may bring it down.

WouterGritter:
Okay, might be worth it to have this be a feature that's off by default then, with warnings about sending too much updates to services like AWS in the settings menu.

However, this is similar to just tapping the color wheel once or twice per second manually, which ideally should not be a huge problem as this could be done by a user trying to fine-tune the perfect color.

It could be worth adding support for an "intermediate update" in the C code. This type of update would ONLY modify the GPIO pins, and would not make any changes to external services (like mqtt, aws, even udp broadcasting). Because it always expects a final "proper" update when the user has released their thumb off of the color wheel. If we want to go that far, we can implement this system using websockets to allow for even quicker updates without much overhead from separate HTTP requests.

softhack007:
@coderabbitai please review.

Create these summary lists in addition to your review:
* noteworthy good practices

softhack007:
> The change hooks into the iro.js `color:change` event (which fires on every pointer-move pixel) and throttles calls to the existing `setColor(1)` → `requestJson()` pipeline to at most once per 500 ms.

I think this is a dead end - increasing the number of WebUI JSON messages to the wled core is not a good idea, and 2 color updates per second are definitely not "smooth" . Also the PR is stale since almost two years ==> closing.

PR #4302, opened 2024-11-22
Optimzation: code consolidation in drawCircle()

DedeHai:
- replaced repeated function calls with loops
- saves about 600bytes of flash
- speed tradoff: drawing is now a tiny bit slower but insignificant in my tests

### DIFF
--- a/wled00/FX_2Dfcn.cpp
+++ b/wled00/FX_2Dfcn.cpp
@@ -507,7 +507,7 @@ void Segment::drawCircle(uint16_t cx, uint16_t cy, uint8_t radius, uint32_t col,
   if (!isActive() || radius == 0) return; // not active
   if (soft) {
     // Xiaolin Wu’s algorithm
-    int rsq = radius*radius;
+    const int rsq = radius*radius;
     int x = 0;
     int y = radius;
     unsigned oldFade = 0;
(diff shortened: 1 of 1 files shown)


WouterGritter:
> drawing is now a tiny bit slower but insignificant in my tests

What device have you tested it on? We need to make sure it doesn't pose an issue on the slowest supported devices.

DedeHai:
it won't pose any issue, as I mentioned, the slowdown is insignificant, tested on C3. The function is currently only used by ripple FX.

wled00/FX_2Dfcn.cpp
```diff
@@ -516,37 +516,37 @@ void Segment::drawCircle(uint16_t cx, uint16_t cy, uint8_t radius, uint32_t col,
...
     while (y >= x) {
-      setPixelColorXY(cx+x, cy+y, col);
-      setPixelColorXY(cx-x, cy+y, col);
-      setPixelColorXY(cx+x, cy-y, col);
-      setPixelColorXY(cx-x, cy-y, col);
-      setPixelColorXY(cx+y, cy+x, col);
-      setPixelColorXY(cx-y, cy+x, col);
-      setPixelColorXY(cx+y, cy-x, col);
-      setPixelColorXY(cx-y, cy-x, col);
+    for (int i = 0; i < 4; i++) {
```
softhack007, line 544: My suggestion would be to compact the "soft" Wu's algo (as you did), but keep Bresenham's code untouched.
Why: in some cases speed does matter, so I'd like to have the basic circle drawing as fast as possible. Also you did not save much code on Bresenham.

The "soft" algorithm is much slower any way, so I'd agree with you that making the code more compact here is better for readability.

blazoncek: > Also: there is no FX currently using the Bresenham's circle version.

There's an open PR that uses it.

netmindz:
Now I've tagged 0.15.rc.1 - feel free to merge this @softhack007

DedeHai:
I would hold off with merging the optimization PRs to not accidentally introduce bugs. This PR serves one purpose only: code consolidation and flash optmization, it adds no features and fixes no issues.

netmindz:
My plan for the GA release will be taken from the rc1 tag plus cherry pick of any fixes, not taking 0_15

DedeHai:
got it. I have quite a few PRs pending that need to go through beta testing. Shall I merge those into 0_15 or will there be a new branch?

netmindz:
You are good to go ahead and merge into 0_15 as this will be merged into main and then deleted as soon as the 0.15.0 GA release happens

PR #4307, opened 2024-11-23
Compile time lock PIN definition

blazoncek:
Adds ability to define default settings PIN at compile time (instead of no PIN defined).

Requested by a friend.

### DIFF
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -564,7 +564,7 @@ WLED_GLOBAL bool otaLock        _INIT(false);     // prevents OTA firmware updat
 WLED_GLOBAL bool wifiLock       _INIT(false);     // prevents access to WiFi settings when OTA lock is enabled
 WLED_GLOBAL bool aOtaEnabled    _INIT(true);      // ArduinoOTA allows easy updates directly from the IDE. Careful, it does not auto-disable when OTA lock is on
 WLED_GLOBAL char settingsPIN[5] _INIT(WLED_PIN);  // PIN for settings pages
-WLED_GLOBAL bool correctPIN     _INIT(true);
+WLED_GLOBAL bool correctPIN     _INIT(!strlen(settingsPIN));
 WLED_GLOBAL unsigned long lastEditTime _INIT(0);
 
 WLED_GLOBAL uint16_t userVar0 _INIT(0), userVar1 _INIT(0); //available for use in usermod
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -278,7 +278,11 @@ WLED_GLOBAL char releaseString[] _INIT(TOSTRING(WLED_RELEASE_NAME)); // somehow 
 
 // AP and OTA default passwords (for maximum security change them!)
 WLED_GLOBAL char apPass[65]  _INIT(WLED_AP_PASS);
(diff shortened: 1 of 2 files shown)


w00000dy:
How do I set this?

I tried `build_flags =... -D WLED_PIN=1234` and `-D WLED_PIN="1234"`. Both did not work.

blazoncek:
Either: `-D WLED_PIN=\"1234\"` or `-D WLED_PIN='"1234"'` as used elsewhere.

w00000dy (commented):
A few things I noticed:

- This only works if you have a fresh installation ( erased flash).
- On the first boot it does not ask for the pin. Only after the 2nd boot does it ask for the pin.

wled00/wled.h
```diff
@@ -556,10 +560,10 @@ WLED_GLOBAL byte macroLongPress[WLED_MAX_BUTTONS]     _INIT({0});
 WLED_GLOBAL byte macroDoublePress[WLED_MAX_BUTTONS]   _INIT({0});
 
 // Security CONFIG
-WLED_GLOBAL bool otaLock     _INIT(false);  // prevents OTA firmware updates without password. ALWAYS enable if system exposed to any public networks
-WLED_GLOBAL bool wifiLock    _INIT(false);  // prevents access to WiFi settings when OTA lock is enabled
-WLED_GLOBAL bool aOtaEnabled _INIT(true);   // ArduinoOTA allows easy updates directly from the IDE. Careful, it does not auto-disable when OTA lock is on
-WLED_GLOBAL char settingsPIN[5] _INIT("");  // PIN for settings pages
+WLED_GLOBAL bool otaLock        _INIT(false);     // prevents OTA firmware updates without password. ALWAYS enable if system exposed to any public networks
```
netmindz, line 563: Revert unrelated whitespace changes please

blazoncek:
> This only works if you have a fresh installation (erased flash).

Yes, that is expected. The code does not modify existing configuration.

> On the first boot it does not ask for the pin. Only after the 2nd boot does it ask for the pin.

Then a check where a bug is is in order. Though that may be good as it will allow to set WiFi credentials (unfortunately that will be available only once).

wled00/wled.h
blazoncek: I want to keep comments aligned. Personal preference.

netmindz: That's fine for when making changes, not for unrelated code

It's bad practice to apply formatting changes to code not related to the PR, it creates harder to review PRs and breaks the history of each line making debugging issues harder as git blame will make it appear the code has changed more recently that it has and with incorrect description

w00000dy (approved):

blazoncek:
There's one more thing that needs to be done: allow WiFi settings to be available until WiFi is configured.

w00000dy:
I would prefer that the WiFi settings be locked when I set a pin (as they are now). I think this would be the expected behavior when I set a pin. 
So if I run a device in AP-only mode, I can prevent anyone from changing that.

blazoncek:
Are there any objections to this PR? It is an "opt-in" feature for those that self-compile and does not affect regular official builds.
If no, I would merge it.

PR #4311, opened 2024-11-23
Delete Gitpod files and badge

w00000dy:
When I try to use Gitpod I get this error:
grafik

So I had the idea to remove it.

I don't see a reason to have this if you have to pay for it, and it seems to be the same as GitHub Codespaces. You could also use Github Codespaces and do not need any files to set it up. You can also use GitHub Codespaces for free with GitHub Pro.

I have already suggested the idea in the Discord and have received positive feedback.


Jason2866:
Gitpod has free minutes every month. Enough for compiling for most users of WLED

w00000dy:
Okay, but I still do not see an exact use case for Gitpod.
You can download compiled binaries from [Github Artifacts](https://github.com/Aircoookie/WLED/actions/runs/12048529197).
If you want to compile WLED, you can use the [WLED online compiler](https://wled-compile.github.io/).
So what would be a use case where I would need Gitpod?
Where do I get the free minutes? I tried to create a new account and I still can't use it for free.

grafik
grafik

Jason2866:
Exactly as when compiling on own local PC. Without the need to have a PC.

softhack007:
@w00000dy I see a use case for "intermediate beginners" who are scared of setting up VSCode locally (yes 😉 these people exist) - so these people can build their own firmware, then download the results and flash locally. Its a pity that gitPod only allows a few hours of free use per month.

As there is no clear consensus, maybe its better to keep the gitPod link.

Jason2866:
Did nothing. Only connected my github account. I see this when going in settings
image

netmindz:
This discussion is rather pointless, @w00000dy. Just because you would not want to use this platform, others might and there is no harm in having these two small files, so removing something that others might be using is not going to happen

If it created builds that were broken and caused support issues, only then would we remove

w00000dy:
I don't want to not use it, I just can't use it.

I just don't like having stuff in the repo that does not work.

netmindz:
You can't use it because you have used all your free credits, there is nothing wrong with the files. Speak to GitPod if you think you should still have free credits remaining

w00000dy:
I created an entirely new account.
If you compare the screenshot of @Jason2866 and mine, you can see that I never got any credits (monthly usage).
I don't want to bother with Gitpod when I can just use Github Codespaces.

w00000dy:
Then keep it if you feel that way. 🙂

PR #4319, opened 2024-11-25
Update settings_sec.htm for buttons placment and warning text  ( Cosmetic change ) 

dosipod:
Fix the  warning text and Backup presets button misalignment  in  settings_sec.htm .
Before and after view  below . PC browser tested are   firefox , chrome and edge . Phone tested are  Iphone7 and   Iphone10

image

This is  a cosmetic change but may lead to confusion  otherwise

### DIFF
--- a/wled00/data/settings_sec.htm
+++ b/wled00/data/settings_sec.htm
@@ -60,8 +60,8 @@
 		<hr>
 		<h3>Backup & Restore</h3>
 		<div class="warn">&#9888; Restoring presets/configuration will OVERWRITE your current presets/configuration.<br>
-		Incorrect upload or configuration may require a factory reset or re-flashing of your ESP.</div>
-		For security reasons, passwords are not backed up.
+		Incorrect upload or configuration may require a factory reset or re-flashing of your ESP.<br>
+		For security reasons, passwords are not backed up.</div>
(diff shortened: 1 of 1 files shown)


blazoncek:
Please remove changes to platformio.ini otherwise it is good to go.

dosipod:
Sure I will do that , i am just testing with someone for another thing  and will be done in a bit

softhack007 (approved):
I agree with @blazoncek - change looks good (and well spotted 👍 ).
Please undo changes to platformIO.ini, then the PR should be good to merge.

dosipod:
@blazoncek @softhack007  I have undone  changes to platformIO.ini . 
But now there is another thing i just noticed , what is the use of the upload buttons ? It is not clickable  but you can still hover on it and the hand will show  , I think that is another issue as it is now basically just a repeated label , we should remove them or fix it so it will upload and remove the other text above it as it will be redundant  . If yes then then I will cancel this PR and redo that .


https://github.com/user-attachments/assets/00548c75-ab91-4c4c-a1a2-613e5b40a7f4

willmmiles:
By "undo changes", I'd prefer those commits removed from the PR branch entirely.  Whatever unrelated things you were testing do not belong in the version history for this PR.  It's OK to force push the branch for this.

PR #4336, opened 2024-12-01
settings_leds: Initialize current limiter field

willmmiles:
When adding a new bus, the numeric current limit field was not being initialized; this was causing it to send an empty value (interpreted as 0) when saved if the current limiter drop-down was never touched.

This is a fix for one of the issues reported in #4280.

------------

While this fix technically works, I'm not really satisfied with it -- it replicates the initialization value, so it has to be kept in sync with `LAsel` should that ever change.  I couldn't see an easy alternative, though.  Suggestions welcome!

### DIFF
--- a/wled00/data/settings_leds.htm
+++ b/wled00/data/settings_leds.htm
@@ -6,7 +6,7 @@
 	<title>LED Settings</title>
 	<script src="common.js" async type="text/javascript"></script>
 	<script>
-		var laprev=55,maxB=1,maxD=1,maxA=1,maxV=0,maxM=4000,maxPB=2048,maxL=1664,maxCO=5,maxLbquot=0; //maximum bytes for LED allocation: 4kB for 8266, 32kB for 32
+		var maxB=1,maxD=1,maxA=1,maxV=0,maxM=4000,maxPB=2048,maxL=1664,maxCO=5; //maximum bytes for LED allocation: 4kB for 8266, 32kB for 32
 		var oMaxB=1;
 		var customStarts=false,startsDirty=[];
 		function off(n)    { gN(n).value = -1;}
--- a/wled00/data/settings_leds.htm
+++ b/wled00/data/settings_leds.htm
@@ -422,7 +422,7 @@ mA/LED: <select name="LAsel${s}" onchange="enLA(this,'${s}');UI();">
 <option value="15">15mA (seed/fairy pixels)</option>
 <option value="0">Custom</option>
 </select><br>
-<div id="LAdis${s}" style="display: none;">max. mA/LED: <input name="LA${s}" type="number" min="1" max="255" oninput="UI()" value="55"> mA<br></div>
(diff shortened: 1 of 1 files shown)


blazoncek (approved):
Another option, I think, is to set `min` value to 0 or call `enLA()` after new bus is added.
Otherwise looks good as in case limiter isn't enabled this field is not used.

softhack007:
@willmmiles unfortunately this is not working with my example cfg.json.
I'll do some more tests to find out what's happening.

softhack007:
@willmmiles i'm getting this error in MSEdge dev tools:

image

I think the div ID is called "psu" (lower) not "PSU" (upper)

softhack007:
> @willmmiles i'm getting this error in MSEdge dev tools:
> I think the div ID is called "psu" (lower) not "PSU" (upper)


nevermind ... was my error. I've accidentally deleted a quote when manually applying your patch 🤦

wled00/data/settings_leds.htm
```diff
@@ -422,7 +422,7 @@
 <option value="15">15mA (seed/fairy pixels)</option>
 <option value="0">Custom</option>
 </select><br>
-<div id="LAdis${s}" style="display: none;">max. mA/LED: <input name="LA${s}" type="number" min="1" max="255" oninput="UI()"> mA<br></div>
+<div id="LAdis${s}" style="display: none;">max. mA/LED: <input name="LA${s}" type="number" min="1" max="255" oninput="UI()" value=55> mA<br></div>
```
softhack007, line 425: might be better to put the value in quotes? like `value="55"`

it might also be possible to use the JS var `laprev`

https://github.com/Aircoookie/WLED/blob/99427c2ef734eb0bd04ec2c5f9d21655d0349e25/wled00/data/settings_leds.htm#L9

softhack007 (approved):
My testcase works now (after fixing my own stupid mistake when manually patching settings_leds.htm)

wled00/data/settings_leds.htm
willmmiles: > might be better to put the value in quotes? like value="55"

I should've looked that up; I'd assumed a numeric field required a numeric value, but you're right - it should be quoted.

> it might also be possible to use the JS var laprev

Oooh, good catch!  This variable is never used - perhaps the intent was to make the next new bus inherit the values from the previous one, which is a reasonable default.  I'm inclined to remove it for now - that's beyond my level of JS skill, especially given the upcoming release - but leave an issue entry for the duplication behaviour for the future.

blazoncek:
```js
enLA(d.Sf["LAsel"+s],s); // update LED mA
```
is needed just before `// disable inappropriate LED types` in `addLEDs()`.

This will fill whatever is selected as a default LED mA.

softhack007:
Last update works in my test, too. PR is ready to merge 👍

Issue #3605, opened 2023-12-21
Enhance API to choose a random filtered effect (1D, 2D, sound reactive)


aeronauth:
It would be great if HTTP/JSON API could start a random effect with filter flags (1D, 2D, sound reactive x2) to limit the random effect to these filters.

With that, one can use a push button to randomly go trough all matrix effects only, another button to randomly go trough all sound reactive effects only, ....

As a effect might be 2d and sound reactive at the same time, that flags might be implementend for combination use.

I know that it's possible to create alot of presets and use them in playlists, which can  be used by the buttons, but i think there should be a more convenience solution.

blazoncek:
Unfortunately that may be impossible to implement without severely impacting performance as this data (1D, 2D, ...) is embeded in string. Parsing it takes time.
Best approximation is to use `"fx":"X~Y~r"` where you select X and Y to encompass certain effect range.
See [KB](https://kno.wled.ge/interfaces/json-api/#contents-of-the-segment-object)

Unfortunately I was over-voted when I tried to group certain effects together and the SR effect numbering prevailed.

dosipod:
How about adding the ability to access FX numbers directly from a playlist as it is limited at the moment to preset numbers 
![image](https://github.com/Aircoookie/WLED/assets/87589371/48710785-95e1-4773-b300-53afb17f5a1c)

dosipod:
And I just tested your command listed about for "fx":"X~Y~r" as in {"fx":"124~180~r"} or {"seg":{"fx":"124~180~r"}}  if that is what you meant and it does not work so that

blazoncek:
I am sorry for the confusion. The `"X~Yr"` API only works for presets (notice the lack of second `~`).
Unfortunately effects have hardcoded limits for random selection.
I may add custom limits in the future.

czesiu20:
> Unfortunately that may be impossible to implement without severely impacting performance as this data (1D, 2D, ...) is embeded in string. Parsing it takes time. Best approximation is to use `"fx":"X~Y~r"` where you select X and Y to encompass certain effect range. See [KB](https://kno.wled.ge/interfaces/json-api/#contents-of-the-segment-object)
> 
> Unfortunately I was over-voted when I tried to group certain effects together and the SR effect numbering prevailed.

Limiting the range of effects would be very useful. I had to implement my own function to handle Music+/- buttons of my 38-key and 24-Key Music IR remotes.


[...]

PR #4346, opened 2024-12-05
Perceptual brightness usermod

obar:
This usermod lets you make changes to WLED global brightness in 'steps' that are evenly spaced to the human eye.

It lets the user define how many brightness steps to use (up to 32 by default) and what the maximum brightness should be. It also allows for skipping steps at the beginning and end of the range.

The usermod provides JSON API commands to step up and down in brightness, `{"pb":"+"}` and `{"pb":"-"}` respectively. It also provides commands to cycle brightness on a button hold, changing directions after a brief button release or optionally changing directions after reaching an extreme without a button release, `{"pb":"/"}` and `{"pb":"~"}` respectively.


DedeHai:
thanks for contributing!
question: how is this different to the setting "use gamma correction for brightness"? It uses the CIE brightness formula.

obar:
Oh that's interesting. Yeah the 'gamma for brightness' is computing brightness _basically_ the same way as this usermod. I'm struggling to recreate the button functionality I've got here without the usermod though. Is there a way to set lower/upper bound and step size in the API brightness command? Ideally I wouldn't want the gamma-adjusted brightness to go below 60 or above 160 on my setup. It seems with this gamma-adjusted brightness there are many low values which aren't really running.

It could be that my usermod's useful bit is press-and-hold button handling for brightness, with max/min/step size and the brightness cycling variations. Or does that make sense as an API extension and not a usermod?

DedeHai:
I am not familiar enough with JSON commands to say if it is possible. The press end hold for brightness already exists, you just assign a preset to increase or decrease brightness with a defined step size and assign that to "long press" of the button (which still may have a bug, as discussed on discord). So the only thing missing would be to set upper and lower limits, correct?

blazoncek:
`"bri":"w60~160~5"`
Buttons 1 and up, have repeating ability built in.

obar:
Hmm, I tried exactly that API command in a preset and it doesn't work, from 0_15 branch. It works for you @blazoncek?

blazoncek:
Looks like it is not working for "bri". You can file a bug report.

blazoncek:
Similar to #3605

blazoncek:
I have a generic fix. The API is `{"bri":"60~160w~5"}`.
The previous example is not valid any way.

obar:
I just applied your 52e08d2 commit and tried it out with and without wrapping for brightness. Works perfectly! Between that and the gamma correction for brightness feature, my usermod isn't adding much :grin: Will close the pull request. Thanks!

PR #4356, opened 2024-12-07
Proper fix for #3605 & #4346

blazoncek:
- allow incrementing/decrementing any numeric value as specified in API (universal)


netmindz:
Am I right that both issues listed are feature requests rather than bug fixes? If so, then will be merged after 0.15.0 GA

blazoncek:
Well, considering that API in KB mentioned such behaviour (prior to my partial and now proper fixes) I deem it a bugfix. But it is your choice.

PR #4358, opened 2024-12-07
Added wipe in and wipe out effects

Rinszky:
This is a wipe in and wipe out effect to turn on and off the led smoothly with slide.
It was very important for me, and maybe it'll be beneficial for others, I saw a lot of topic about this how to turn on and off like this...
Before this, it was possible only with playlists, but the timing was problematic.

[example1](https://wled.discourse.group/t/wipe-off-effect/5647)
[example2](https://www.reddit.com/r/WLED/comments/1eq2meu/wipe_then_stop/)
[example3](https://blog.dest-unreach.be/2023/11/22/WLED_wipe_on/)
[example4](https://github.com/Aircoookie/WLED/issues/1898)

So Maintainer, please decide if you need this effect or not!
Should I change something to be better?
If you know a better name for this effect, let me know and I'll rename it!

Thank you for this cool product!!!


DedeHai:
thanks for contributing!
however, this is a whole lot of code for something that can be done already, although a bit convoluted.
if you want to solve it in an elegant way:
how about adding two checkmarks to the existing wipe effect, one to 'wipe on' and one to 'wipe off'?

Rinszky:
> thanks for contributing! however, this is a whole lot of code for something that can be done already, although a bit convoluted. if you want to solve it in an elegant way: how about adding two checkmarks to the existing wipe effect, one to 'wipe on' and one to 'wipe off'?

Not a bad idea, I'm totally new in this world, and it was easier for me to create a new, instead of modifying and maybe breaking something... If you know how to do that, I would appreciate for your help!

blazoncek:
FYI Effect blending styles PR (#3877) adds universal ability to "wipe in" and "wipe out" any effect among other styles.

dosipod:
There seems to be a lot of other duplicate effects  and some of them are useless so I see why not add another wipe effect that is as @Rinszky  listed have been requested a lot and doing it with playlist is bad .

dosipod:
@Rinszky Now I see it I am also not sure why add 5 effects when only two can do .
Also something seems to be wrong with the wipe in ( or logic as you did not show your own  demo after your change) 
Once you select a wipe ( try with wipe in fast ) then it will light all the leds on peek and fixture but on the fixture 
the leds will go off after a while and will come back again .Peek will stay on . So seems like a bug 
 
The expected behavior we had in mind and done with playlist is  one effect to lights all the leds one by one and then all the leds should stay on . The 2nd effect will turn off the leds one by one and then keep all of them off 

image

DedeHai:
@dosipod can you confirm the new blending styles are suited to solve this? if so, there is not need to change the wipe effect as the blend-in is much more versatile.

dosipod:
@DedeHai I never tested that but if blaz said so then must be

dosipod:
I think I was traveling at the time but there is even video here of wipe in but not named so it was not clear to others what he was referencing https://discord.com/channels/473448917040758787/779395228624617512/1258428188527296614

wled00/FX.cpp
```diff
@@ -6466,6 +6466,113 @@ uint16_t mode_matripix(void) {                  // Matripix. By Andrew Tuline.
...
+  if (currentTime - SEGENV.aux0 > stepDuration) {
+    SEGENV.aux0 = currentTime;
+    if (SEGENV.aux1 < SEGLEN) {
+      SEGMENT.setPixelColor(SEGENV.aux1, BLACK);
+      SEGENV.aux1++;
+    } else {
+      // All LEDs are off, reset
+      SEGENV.aux1 = 0;
+      strip.setBrightness(0);
+      return 65535;
```
softhack007, line 6493: this will not cause a "forever black" - the effect will run again in 65535 millis = 65 seconds.  Also manipulation `strip` properties in an effect (at segment level) is a very fragile thing and should be avoided.

wled00/FX.cpp
```diff
@@ -6466,6 +6466,113 @@ uint16_t mode_matripix(void) {                  // Matripix. By Andrew Tuline.
...
 
+//////////////////////
+// *    WIPE OUT    //
+//////////////////////
+uint16_t mode_wipe_out(void) {
+  if (SEGLEN == 1) return mode_static();
+
+  uint16_t duration = 1000 - (SEGMENT.speed * 4); // Adjust speed (1-255) to duration (1000ms to 20ms)
+  uint16_t stepDuration = duration / SEGLEN;
+  uint32_t currentTime = millis();
```
softhack007, line 6477: use strip.now instead of millis()

wled00/FX.cpp
```diff
@@ -6466,6 +6466,113 @@ uint16_t mode_matripix(void) {                  // Matripix. By Andrew Tuline.
...
+
+//////////////////////
+// *    WIPE IN     //
+//////////////////////
+uint16_t mode_wipe_in(uint16_t segPerLoop) {
+  if (SEGLEN == 1) return mode_static();
+
+  uint16_t duration = 1000 - (SEGMENT.speed * 4); // Adjust speed (1-255) to duration (1000ms to 20ms)
+  uint16_t stepDuration = duration / SEGLEN;
+  uint32_t currentTime = millis();
```
softhack007, line 6513: see previous comment

blazoncek:
Everyone, please see blending styles before continuing discussion on this PR.

softhack007:
@Rinszky If I understand your code correctly, your implementation of "wipe" is always filling / clearing the segment with a single color, right? This means we still have no good solution for other animated effect.

To safe some space in the effect table, it might be better to add "wiping" options (checkbox) and wipe speed control (sliders) to the "Solid" effect. It would be possible to still have a "wipe-out" by using several presets for "wipe in" "normal light" and "wipe out"

dosipod:
This is blending styles demo , Something you may not want to hear is that wled is having too many effects already , if that good or bad  ,quantity vs quality is open to debate  , I personally in favor of having an option to disable some effects from ui better then having to hack fx.cpp to remove the ones some users  do not like or think they are redundant  and can  be easily replicated with others effects . 
blend

DedeHai:
superseeded by blending styles

PR #4359, opened 2024-12-07
WiFi reconnect bugfix

blazoncek:
- additional debug info

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -478,10 +478,7 @@ void WLED::setup()
   if (strcmp(multiWiFi[0].clientSSID, DEFAULT_CLIENT_SSID) == 0)
     showWelcomePage = true;
   WiFi.persistent(false);
-  #ifdef WLED_USE_ETHERNET
   WiFi.onEvent(WiFiEvent);
-  #endif
-
   WiFi.mode(WIFI_STA); // enable scanning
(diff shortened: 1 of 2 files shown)


wled00/wled.cpp
```diff
@@ -979,6 +974,7 @@ void WLED::handleConnection()
       initConnection();
       interfacesInited = false;
       scanDone = true;
+      return;
```
netmindz, line 977: Does this return actually fix the issue? I'm unclear if this PR is a fix or just adding debug info to help try and find the issue so we can then fix @blazoncek

blazoncek: Yes, this is the actual fix.
Debug changes were added to find the root cause.

netmindz: Great :)

Are you confident enough in this fix for this to go straight out into the GA release or do we need to get some feedback from the dev branch first?

blazoncek: I hunted this for a month. 😄 And the multiwifi implementation is mine.
The bug is odd. When connecting to WiFi after boot if 1st SSID is accessible it will connect to it but subsequent attempts (if disconnected) to it will never commence and only 2nd or 3rd SSID will be tried.
This early return insures that also 1st SSID is correctly retried.

netmindz (approved):

PR #4361, opened 2024-12-07
Usermod BME280:  Send "normalized" pressure (pressure at sealevel)

euphi:
Usermod BME280: Also send "normalized" pressure (equivalent pressure at sealevel) on MQTT.

This commit adds support to send pressure at sealevel on MQTT. Therefore a new config parameter "height" is added.


netmindz:
Please rebase your PR against 0_15, not 0_15_0 as this isn't a bugfix

PR #4367, opened 2024-12-08
Add UDP protocol for same-color range updates

comcloudway:
Allows easily changing a number of leds to the same color at once, without having to send individual colors.

Especially useful to easily change the color of the whole strip or large parts of it to the same color. i.e. custom fade in animations

---

I added this as I couldn't find a protocol implementation that could easily flash the entire strip / control custom whole-strip fade animations.
Update entire ranges to a specific color should save bandwidth as the amount of bytes sent is kept at a minimum.

|Byte|Description|
|---|---|
|0 |Protocol = 6 |
| 1 | Timeout |
| 2 | Strip starting index |
| 3 | Number of LEDs to update from the starting index |
| 4 | Red value |
| 5 | Green value |
| 6 | Blue value |
| 7 | White value |

*Note*: The white value exists, because I own a strip that supports the white channel, and I thought that whilst the idea of this protocol is keeping the byte count at a minimum, adding the single white byte would be OK.

Example transmissions:
- `{ 6, 255, 0, 30, 255, 0, 0, 0 }` - Changes the first 30 leds to red
- `{ 6, 255, 30, 10, 0, 255, 0, 0 }` - Changes leds 31 to 40 to green

### DIFF
--- a/wled00/udp.cpp
+++ b/wled00/udp.cpp
@@ -511,7 +511,7 @@ void handleNotifications()
   }
 
   //UDP realtime: 1 warls 2 drgb 3 drgbw
-  if (udpIn[0] > 0 && udpIn[0] < 5)
+  if (udpIn[0] > 0 && udpIn[0] < 7)
   {
     realtimeIP = (isSupp) ? notifier2Udp.remoteIP() : notifierUdp.remoteIP();
     DEBUG_PRINTLN(realtimeIP);
@@ -569,7 +569,17 @@ void handleNotifications()
         setRealtimePixel(id, udpIn[i], udpIn[i+1], udpIn[i+2], udpIn[i+3]);
         id++;
       }
+    } else if (udpIn[0] == 6) //scrgbw
+    {
+      uint8_t start = udpIn[2];
+      uint8_t count = udpIn[3];
+      for (size_t id = start; id<start+count; id++)
+      {
+        if (id >= totalLen) break;
+        setRealtimePixel(id, udpIn[4], udpIn[5], udpIn[6], udpIn[7]);
+      }
     }
+
     strip.show();
     return;
   }


netmindz:
Nice idea. Is there any software that exists yet that supports this format?
I wonder if this could also be extended so that you can send multiple ranges per frame, otherwise, you could get odd behaviour if you are trying to update different parts of the strip but get packet loss or out of order packets

netmindz:
You will also not want to limit yourself to only 255 leds, so will need more than a single byte

DedeHai:
> You will also not want to limit yourself to only 255 leds, so will need more than a single byte

I was wondering the same things:
a) is this a known protocol? `Timeout` is not used in the code provided
b) limiting the number of LEDs to 255 per call seems ok to me but why limit it to 8bit? 
c) The start-point must be 16bit, or how would you change LEDs starting at 500 for example?

comcloudway:
> Is there any software that exists yet that supports this format?

Not that I know of - it was just the first idea that came to mind when thinking about how to update the whole strip at once.

> `Timeout` is not used in the code provided

The first two bytes are only for reference purposes, [they are used like this by the other UDP protocols](https://github.com/Aircoookie/WLED/pull/4367/files#diff-c60badf90f24d60bdec37d9672fa83eec5009ba203d06e52d431d9709dd2ccf0R520).

> but why limit it to 8bit?

Because I `udpIn` is a `uint8` array, and [the basic *warls* protocol also only uses 8 bits](https://github.com/Aircoookie/WLED/pull/4367/files#diff-c60badf90f24d60bdec37d9672fa83eec5009ba203d06e52d431d9709dd2ccf0R534)

>  The start-point must be 16bit, or how would you change LEDs starting at 500 for example?

The *warls* protocol doesn't support them either, so I wasn't sure it was required :see_no_evil: 

---

> I wonder if this could also be extended so that you can send multiple ranges per frame

The following examples use the same syntax as [the official docs](https://kno.wled.ge/interfaces/udp-realtime/)

## Option 1

I'm not entirely sure, but I think using an offset notation might be the easiest way to address this, as it allows setting large strips in one package.

| Byte | Description |
|---|---|
|2| Segment count (may not be 0) |
|3+n*5| Offset |
|4+n*5| Led count |
|5+n*5| Red Value |
|6+n*5|Green Value|
|7+n*5|Blue Value|

This would allow updating a strip longer than 255 LEDs like this:
```c
{ 
6, // protcol
255,  // timeout
2, // two segment update blocks will follow

0, // starting point: led 0
255, // number of leds
255, // red value
0, // green value
0, // blue value

0, // offset 0 from last updated led (255 in this example)
255, // number of leds
255, // red value
0, // green value
0, // blue value
}
```

The idea being that we have something like this (I haven't checked to see if this works/is valid c code)
```c
uint16_t id = 0;
for (int b = 0; b<udpIn[2]; b++) {
  id += udpIn[3+b*5];
  for (int i = id; i < id+udpIn[4 + b*5]; i++) {
    // TODO: update colors using udpIn[{5,6,7} + b*5]
  }
  id += udpIn[4 + b*5];
}
```

NOTE: This would not easily allow changing the first and last block on strips with 400+ LEDs (unless one uses `offset=255` and `led_count=0`

## Option 2

This option basically addresses the problem of having to use full-offset-zero-length blocks as noted above, by allowing to update later segments or longer segments.

| Byte | Description |
|---|---|
|2| Segment count (may not be 0) |
|3+n*7| Start index high byte |
|3+n*7| Start index low byte |
|4+n*7| Led count high byte |
|4+n*7| Led count low byte |
|5+n*7| Red Value |
|6+n*7|Green Value|
|7+n*7|Blue Value|

It would also mean that we wouldn't have to update the `id` as much. Also, my existing code should be easily adaptable to fit this.

Example usage (I hope, that I didn't mix the low/high byte order):
```
{ 
6, // protcol
255,  // timeout
2, // two segment update blocks will follow

0, // starting point: led 0
0, 
0,  // number of leds
255,
255, // red value
0, // green value
0, // blue value

10, // starting point of segment two 
0, 
0,  // number of leds
255,
255, // red value
0, // green value
0, // blue value
}
```

netmindz:
Yeah it looks like we need to have the option of a single packet updating multiple ranges

blazoncek:
I am wondering why inventing something new for existing functionality?
Why don't you use JSON API over UDP which is already supported?

I.e. send `{"seg":[{"id":0,"i":[...]},...]}` as a packet content.

comcloudway:
> Why don't you use JSON API over UDP which is already supported?

I haven't tested it, but I'd assume that json deserialization requires more efford than simply reading bytes directly. Also the json data is probably longer (in terms of byte count).

blazoncek:
If you want fastest updates, you have Hyperion or E1.31/ArtNet protocols. Hyperion uses raw UDP data. The downside, no multipixel command. But that is not an issue IMO unless you want to submit several thousand pixels at once (which is a limiting factor in this PR as well).

Otherwise use what already exists and is well tested. JSON also supports segment use while this PR does not.
I see no real benefit in this PR.

BTW WARLS was written several years ago (and has not been updated since) when WLED only supported a single strip and was not optimised for large LED count or multiple segments. AR now features native sync between devices.

comcloudway:
OK, thanks for the feedback though

PR #4373, opened 2024-12-10
Update package.json

srg74:
Correct version

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wled",
-  "version": "0.15.0-dev",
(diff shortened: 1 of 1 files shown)


netmindz:
Builds taken from the development branch should be versioned as dev build not abusing the tag version

srg74:
When custom firmware built version is taken from this file. If disregarded names of generated will be wrong.

netmindz:
I'm very sorry but I do not understand the point you are trying to make @srg74

From now on,  main will be where all development happens

PR #4374, opened 2024-12-10
Update wled.h

srg74:
Corrected version and build

### DIFF
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -3,12 +3,12 @@
 /*
    Main sketch, global variable declarations
    @title WLED project sketch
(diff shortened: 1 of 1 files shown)


dosipod:
That will not work as the version is in package.json
image

netmindz:
I see no advantage of the @version in the source, especially if it's not automatically managed.

If it is there, then as already explained, it's value is correct. It should only be 0.15.0 on the tag, not on the ongoing development branch.

The build number also needs review as manual exiting as arbitration time period isn't sustainable and does not provide a reliable value to reference when users report issues

PR #4386, opened 2024-12-13
Fix for ESPNow remote causing output glitches

DedeHai:
Processing of received button command is no longer done in the interrup-callback, instead the value is saved to a variable and processed in the main loop.
The actual fix is to _not access the file system while data is being sent out_: even just trying to open a (non-existing) file causes glitches on the C3. Waiting for the bus to finish updating soves this BUT it can cause a frame-delay. A short delay however is the lesser evil and much less noticeable than random color flashes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "clean-css": "^5.3.3",
         "html-minifier-terser": "^7.2.0",
         "inliner": "^1.13.1",
-        "nodemon": "^3.0.2"
+        "nodemon": "^3.1.7"
       }
     },
     "node_modules/@jridgewell/gen-mapping": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wled",
-  "version": "0.15.0-b7",
+  "version": "0.15.0-dev",
   "description": "Tools for WLED project",
   "main": "tools/cdata.js",
   "directories": {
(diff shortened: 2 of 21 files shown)


DedeHai:
note: I did not realize I should base for main so this is still 0_15 based. shold I rebase or merge to 0_15 and then cherry pick?

netmindz:
I should be able to just change the base of the PR, no need to cherry-pick

DedeHai:
I had lots of conflicts initially, that is why I switched from main to 0_15. 
@netmindz  Should all still pending PRs that have 0_15 as a target also be switched to main? I did merge a PR into 0_15 yesterday... how long is that still a valid process?

netmindz:
Yeah if you opened before I merged 0_15 into main you would have seen all sorts of issues.

At the moment they are basically the same, but if you could please use main for anything new. 

I'm going through cleaning up anything that is still 0_15 and will make sure all changes to 0_15 have been merged before I finally delete 0_15

DedeHai:
if there are no objections to the updated function names @softhack007 @netmindz this is ready to merge. tested and working.

softhack007:
@DedeHai I possibly have time tomorrow to take a look at your changes. In case I can't review until Saturday, please feel free to merge. Tho overall idea looks sound to me.

DedeHai:
this is not urgent, no worries.
please also consider the alternative implementation of this by @blazoncek  https://github.com/blazoncek/WLED/commit/43a9663eb923e2b53c9e81e7d6af643fd84cbe15

the outcome is the same, code structure is different. as mentioned in that commit, I'd rather keep the code for the remote contained in remote.cpp as much as possible for clarity and maintainability but maybe my approach has flaws I have not considered.

DedeHai:
@softhack007 just a friendly reminder: still waiting on your feedback on this

DedeHai:
@softhack007 if you have no objections I would merge this PR

PR #4392, opened 2024-12-13
ESP32-S2-Mini constantly reboots

da-markson:
Soultion addresses problem where SOME of the ESP-S2-Mini's would constanly reboot itself, thus never allowing the Access Point to initialize

## Summary by CodeRabbit

- **Refactor**
	- Adjusted the sequence of setting WiFi transmit power for ESP32 devices during access point initialization. No changes to user-facing features or controls.

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -609,13 +609,13 @@ void WLED::initAP(bool resetAP)
     WLED_SET_AP_SSID();
     strcpy_P(apPass, PSTR(WLED_AP_PASS));
   }
+  #ifdef ARDUINO_ARCH_ESP32
+  WiFi.setTxPower(wifi_power_t(txPower));
+  #endif
   DEBUG_PRINT(F("Opening access point "));
   DEBUG_PRINTLN(apSSID);
   WiFi.softAPConfig(IPAddress(4, 3, 2, 1), IPAddress(4, 3, 2, 1), IPAddress(255, 255, 255, 0));
   WiFi.softAP(apSSID, apPass, apChannel, apHide);
-  #ifdef ARDUINO_ARCH_ESP32
-  WiFi.setTxPower(wifi_power_t(txPower));
-  #endif
 
   if (!apActive) // start captive portal if AP active
   {


wled00/wled.cpp
```diff
@@ -609,13 +609,13 @@ void WLED::initAP(bool resetAP)
...
   }
+  #ifdef ARDUINO_ARCH_ESP32
+  WiFi.setTxPower(wifi_power_t(txPower));
+  #endif
   DEBUG_PRINT(F("Opening access point "));
   DEBUG_PRINTLN(apSSID);
   WiFi.softAPConfig(IPAddress(4, 3, 2, 1), IPAddress(4, 3, 2, 1), IPAddress(255, 255, 255, 0));
   WiFi.softAP(apSSID, apPass, apChannel, apHide);
-  #ifdef ARDUINO_ARCH_ESP32
-  WiFi.setTxPower(wifi_power_t(txPower));
```
softhack007, line 617: 🤔 the "solution" basicially disables `WiFi.setTxPower(wifi_power_t(txPower))`, because trying to set TX power before AP is started will be ignored. I don't think this can be a general solution, because some boards require setting a reduced tx power otherwise wifi will not work.

Its interesting that this solves any problem ... maybe the default TX power is too high.

softhack007 (changes requested):
As explained (see above) I don't agree with the "solution" because it's just making the TX power feature useless in AP mode. 

I think we need a different solution, maybe by only setting TX power when it's different from the default value. Or an extra option in wifi settings, like "default" or "no change".

da-markson:
I understand that it makes TX power useless, but nonetheless, this solution
worked for me. I have nine of the ESP32-S2-mini's and seven of them would
constantly reboot when executing "WiFi.softAPConfig(IPAddress(4, 3, 2, 1),
IPAddress(4, 3, 2, 1), IPAddress(255, 255, 255, 0));". I didn't debug
further because I was working issues outside of the WLED library, but I
surmise that maybe there is a bug within the WiFi library for the mini's?

I do understand if you don't want to merge this request. I'll just keep a
one-off version for my own use.  Great product though!

Thanks,

David Markson
*Cell*:        (817) 584-4755
*E-mail*:    ***@***.***

da-markson:
Unfortunately I don't have an external JTAG debugger, so I can't produce a
crash dump on the ESP32-S2-mini.

Thanks,

David Markson
*Cell*:        (817) 584-4755
*E-mail*:    ***@***.***

DedeHai:
the IDF has an option to "reduce TX power on brown out", but I am not sure this is enabled in WLED and also I did not check on how that is implemented and setting TX power manually will probably overrule that anyway. 
what could be a solution to this: enable a timer that persists between reboots (RTC peripheral has something like that, at least I saw it in an IDF example). It could be used to detect boot-loops and reduce TX power level-by-level to a defined minimum. It is more of a hack though than a real solution, as the real solution would be to fix the hardware with a capacitor to avoid brownout.

softhack007:
> Unfortunately I don't have an external JTAG debugger

@da-markson could you  check if your S2 board still works when you just comment out the line with WiFi.setTxPower(wifi_power_t(txPower)) ? 
This wag we could verify that setTxPower has no other side-effects that we did not consider yet.

blazoncek:
@da-markson you do not need JTAG to get crash dumps. connect USB serial, add -D WLED_DEBUG, add `monitor_filters = esp32_exception_decoder` and monior serial output using PIO

blazoncek:
BTW I have several Lolin S2 Mini and all of them work as expected. Perhaps there is a HW issue with your boards.

da-markson:
I will complete all the requests after the holidays as all my ESP32's are
currently running in my light display.

Thanks,

David Markson
*Cell*:        (817) 584-4755
*E-mail*:    ***@***.***

PR #4393, opened 2024-12-14
Fixes "preset not found" issue when using internal temp UM 

### LINKED ISSUE #4390: Internal temperature causes 'error 12: preset not found' even without any presets
### What happened?

every time i open one of my two wled instances, I get 'error 12: preset not found'

### To Reproduce Bug

I have updated to version 15 recently and don't remember the error showing there, but I can isolate it to this mod because I emptied all my presets and the error still showed, and only stops showing when i disable the mod, and starts showing again when i re-enable the mod

### Expected Behavior

stop showing the error 

### Install Method

Self-Compiled. Completely vanilla to the 'main' branch except for the config in platformio.ini changes to default_envs, uncommented line to the onewire lib, 

and added: 

```
-D USERMOD_DALLASTEMPERATURE
-D USERMOD_INTERNAL_TEMPERATURE
-D USERMOD_PWM_FAN
```
under [env:esp32c3dev]
### What version of WLED?

0.15.0

ESP32-C3

### Relevant log/trace output

_No response_

### Anything else?

_No response_

### Code of Conduct

lost-hope:
fixes #4390

### DIFF
--- a/usermods/Internal_Temperature_v2/usermod_internal_temperature.h
+++ b/usermods/Internal_Temperature_v2/usermod_internal_temperature.h
@@ -50,7 +50,7 @@ public:
 #else                                    // ESP32 ESP32S3 and ESP32C3
     temperature = roundf(temperatureRead() * 10) / 10;
 #endif
-
+ if(presetToActivate != 0){
     // Check if temperature has exceeded the activation threshold
     if (temperature >= activationThreshold) {
       // Update the state flag if not already set
@@ -58,7 +58,7 @@ public:
         isAboveThreshold = true;
         }
       // Check if a 'high temperature' preset is configured and it's not already active
-      if (presetToActivate != 0 && currentPreset != presetToActivate) {
+      if (currentPreset != presetToActivate) {
         // If a playlist is active, store it for reactivation later
         if (currentPlaylist > 0) {
           previousPlaylist = currentPlaylist;
@@ -101,6 +101,7 @@ public:
           }
         }
       }
+ }
 
 #ifndef WLED_DISABLE_MQTT
     if (WLED_MQTT_CONNECTED)


softhack007 (approved):
Changes look good for me 👍 

🤔 Actually I'm still wondering why someone would use a temperature sensor that's so inaccurate for anything like a safety function - but that's another matter.

Just to explain "inaccurate" - I have two esp32 sitting next to each other, doing the same thing. Internal temperature for the first is 50°C, the second shows 78°C. And in reality, both are note even "Luke warm" when  I touch then with my finger. This tells me that the internal  temp sensor of esp32 is really a useless toy nothing more.

orbitinstasis:
> Changes look good for me 👍
> 
> 🤔 Actually I'm still wondering why someone would use a temperature sensor that's so inaccurate for anything like a safety function - but that's another matter.
> 
> Just to explain "inaccurate" - I have two esp32 sitting next to each other, doing the same thing. Internal temperature for the first is 50°C, the second shows 78°C. And in reality, both are note even "Luke warm" when I touch the esp32 with my finger. This tells me that the internal temp sensor of esp32 is really a useless toy nothing more.

I found the bug - the SoC dye is a heck of a lot hotter than the actual carrier board, and I wouldn't really rely on it for a safety feature (for that I use a proper sensor with the dallas usermod), but this mod gives  a nice idea of how the ambient temp is affecting the SoC for certain environments say where the electronics is enclosed, you'd want a dye temp and an ambient temp.

DedeHai:
just FYI, from the docs:
image

the default range is -10°C - 80°C so it should be fairly accurate, if all you need is the die temperature.

lost-hope:
Closing this after talking to @netmindz to make a new PR against main branch. 
#4397 is the new one

PR #4402, opened 2024-12-16
Idle current bugfix

### LINKED ISSUE #4383: Estimated Current Isn't Reported When Off Since 0.15.0
### What happened?

Estimated Current Isn't Reported When off Since 0.15.0:
Screenshot 2024-12-12 at 16 26 40

0.14.0:
Screenshot 2024-12-12 at 16 28 17

Brighness limiter was on in both cases
After reverting back to 0.14.0 brightness limiter got disbaled so I had to re-enable it manually

Also on Athom bulbs 0.15.0 doesn't report estimated current at all!

### To Reproduce Bug

 - update to 0.15.0
 - look at "estimated current"

### Expected Behavior

Estimated Current should be prperly reported

### Install Method

Binary from WLED.me

### What version of WLED?

0.15.0

ESP32

### Relevant log/trace output

_No response_

### Anything else?

_No response_

### Code of Conduct

blazoncek:
Fixes #4383


DedeHai (approved):
I have no means to actually test this but code looks good. approved.

xenjke:
Thank you for this 👍

PR #4406, opened 2024-12-18
Prevent UI from cutting off on mobile devices

kemp:
When the web UI is added to the iOS home screen (via the "Add to Home Screen" option) the navigation tabs are cut off.

This PR adds some padding so they aren't cut off anymore.

| Before | After |
|-----------|----|
| image | image |

Notice the buttons on the bottom are no longer cut off. *Please ignore the error that appears, it is unrelated.*

Side note: if there is interest in adding more "PWA" functionality to the UI (e.g. change the icon on the home screen, add a message when it's disconnected, let me know and I will open another PR :smile:  ).

### DIFF
--- a/wled00/data/index.css
+++ b/wled00/data/index.css
@@ -277,6 +277,7 @@ button {
 .tab {
 	background-color: transparent;
 	color: var(--c-d);
+	padding: env(safe-area-inset-top) env(safe-area-inset-right) env(safe-area-inset-bottom) env(safe-area-inset-left);
 }
 
 .bot {
--- a/wled00/data/index.htm
+++ b/wled00/data/index.htm
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-	<meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1">
+	<meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1, viewport-fit=cover">
 	<meta charset="utf-8">
 	<meta name="theme-color" content="#222222">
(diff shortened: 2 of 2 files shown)


dosipod:
@kemp Looks fine , apparently the issue is with newer iOS devices and sure you are free to do more  improvements , if you do then please make sure you have github actions enabled on your repo so the changes can be quickly tested .
image

Jacobnordvall:
Why add the extra padding on the header/wrapper? 
The middle space is already a bit too tiny.

kemp:
The proposed change adds only the minimum necessary padding defined by the user agent (iOS, in this case).

We could put in a pixel number so that it's manual, but this would be prone to breakage as it would be specific per device and not a general solution.

Jacobnordvall:
Yeah if its tested properly and works in all environments its all good. but from the screenshots youre adding the safe space wrong at top. it adds space for the big notch underr the top toggles instead of above. meaning the notch will still overlay those but youll get empy padding bellow the notch.

blazoncek:
@kemp thanks and it may be ok for this particular PR but please follow [this](https://github.com/Aircoookie/WLED/wiki/How-to-properly-submit-a-PR) for any future PR.

kemp:
Is there anything blocking this from being merged?

blazoncek:
Nothing but patience. 😉

blazoncek:
I do not like padding and the result is visually disappointing.
Also changing `.tab` styles is inappropriate in this case IMO. Rather use `.bot` or better still `#bot`.
I would use `@media` appropriate for mobile display to enforce this (as it only makes sense there).

PR #4422, opened 2024-12-24
Updated TTGO T-Display usermod to fix compilation errors

spiff72:
Revised platformio_override.ini to include all necessary changes for this usermod (no more changes to platformio.ini needed).

Forced older version of the TFT_eSPI library.


spiff72:
found another problem that needs a fix

PR #4424, opened 2024-12-24
Use consistent node version for local, CI, and devcontainer setup

inverse:
- Added `.nvmrc` to define node version for the project (`20.18`)
- Wire up this version in pipelines
- Update dev container setup
  - Use maintained image - previous version archived repo - see: https://github.com/microsoft/vscode-dev-containers/blob/main/containers/python-3/README.md
  - Install node version based on `.nvmrc` file 
  - Drop optional env vars around this since node is required for setup


netmindz:
Do we need to be so specific right down to the patch version?

inverse:
Can be modified to not include the patch version but would mean more consistent builds.

Leveraging tools such as [renovate](https://docs.renovatebot.com/) can automate bumping the patch version too.

netmindz:
Yeah if you can drop the patch version please. We don't have the resources to keep on top of patch updates

.devcontainer/Dockerfile
```diff
@@ -2,12 +2,7 @@
 
 # [Choice] Python version: 3, 3.9, 3.8, 3.7, 3.6
 ARG VARIANT="3"
-FROM mcr.microsoft.com/vscode/devcontainers/python:0-${VARIANT}
```
inverse, line 5: This container has been deprecated since last year so migrated to the new version

inverse:
@netmindz any further feedback on this one?

PR #4426, opened 2024-12-24
replacement for fastled sqrt16()

DedeHai:
replacing yet another fastled function: it is about 10% slower for numbers smaller 128 but faster for larger numbers. Speed difference is irrelevant to WLED (we are talking 30ns slower per call on a C3) but it saves some flash and takes us one step closer to ditching fastled.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -678,7 +678,7 @@ uint16_t Segment::virtualLength() const {
         vLen = max(vW,vH); // get the longest dimension
         break;
       case M12_pArc:
-        vLen = sqrt16_bw(vH*vH + vW*vW); // use diagonal
+        vLen = sqrt32_bw(vH*vH + vW*vW); // use diagonal
         break;
       case M12_sPinwheel:
         vLen = getPinwheelLength(vW, vH);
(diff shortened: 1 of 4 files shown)


wled00/FX_fcn.cpp
```diff
@@ -678,7 +678,7 @@ uint16_t Segment::virtualLength() const {
         vLen = max(vW,vH); // get the longest dimension
         break;
       case M12_pArc:
-        vLen = sqrt16(vH*vH + vW*vW); // use diagonal
+        vLen = sqrt16_bw(vH*vH + vW*vW); // use diagonal
```
softhack007, line 681: @DedeHai would it be possible to extend the function to handle 18 bit numbers? Reason is that the above calculation breaks down already at 181 x 181 pixels, because the argument to sqrt16 does not fit into 16bit any more. 

This could be avoided by simply making the function "sqrt18(uint32_t)".

DedeHai: the function/algorithm works with any size input, could just make the argument 32bit and call it sqrt32, it would work the same except it would be a bit slower as `bit = 1 << 14` would need to be shifted to 30, resulting in a few more init-loops. 
I am not sure if versatility or speed should be prioritized here. in the 1D2D expansion, `sqrtf()`could be used (as long as AR also requires it) on FPU MCUs, which are the only ones in question for such large displays.

DedeHai: thought about it and updated to 32bit as that makes a lot more sense on 32bit system.

PR #4427, opened 2024-12-24
fix POV Display usermod

Liliputech:
There was a compile issue when trying to enable POV Image effect.
This small PR fixes the problem.

### DIFF
--- a/usermods/pov_display/usermod_pov_display.h
+++ b/usermods/pov_display/usermod_pov_display.h
@@ -2,6 +2,10 @@
 #include "wled.h"
 #include <PNGdec.h>
 
+PNG png;
+File f;
+static const char _data_FX_mode_pov_image[] PROGMEM = "POV Image@!;;;1";
+
 void * openFile(const char *filename, int32_t *size) {
     f = WLED_FS.open(filename);
     *size = f.size();
@@ -61,16 +65,22 @@ uint16_t mode_pov_image(void) {
     return FRAMETIME;
 }
 
-class PovDisplayUsermod : public Usermod
-{
-  public:
-    static const char _data_FX_MODE_POV_IMAGE[] PROGMEM = "POV Image@!;;;1";
+class PovDisplayUsermod : public Usermod {
+  protected:
+        bool enabled = false; //WLEDMM
+        const char *_name; //WLEDMM
+        bool initDone = false; //WLEDMM
+        unsigned long lastTime = 0; //WLEDMM
 
-    PNG png;
-    File f;
+  public:
+    PovDisplayUsermod(const char *name, bool enabled) {
+	this->_name = name;
+	this->enabled = enabled;
+    } //WLEDMM
 
     void setup() {
-	strip.addEffect(255, &mode_pov_image, _data_FX_MODE_POV_IMAGE);
+	strip.addEffect(255, &mode_pov_image, _data_FX_mode_pov_image);
+	initDone = true;
     }
 
     void loop() {
--- a/wled00/usermods_list.cpp
+++ b/wled00/usermods_list.cpp
@@ -468,6 +468,6 @@ void registerUsermods()
   #endif
 
   #ifdef USERMOD_POV_DISPLAY
-  UsermodManager::add(new PovDisplayUsermod());
+  UsermodManager::add(new PovDisplayUsermod("Pov Display", false));
   #endif
 }


usermods/pov_display/usermod_pov_display.h
```diff
@@ -2,6 +2,10 @@
 #include "wled.h"
 #include <PNGdec.h>
 
+PNG png;
+File f;
+static const char _data_FX_mode_pov_image[] PROGMEM = "POV Image@!;;;1";
```
DedeHai, line 7: all other FX use uppercase, as it was.

DedeHai:
did not compile, but generally looks ok to me.

softhack007:
@Liliputech please _do not force-push_ while you have a PR open for the the main WLED repo. It causes lots of crazy side-effects like lost review comments, auto-closed bug reports and broken links to code authors.

> [Liliputech](https://github.com/Liliputech) [force-pushed](https://github.com/Aircoookie/WLED/compare/3402a052c0b60266b742ce99d9b41a412f286dd0..001b882e11e6d70f104c6984f94b35ff53c93659) the pov_display branch from [3402a05](https://github.com/Aircoookie/WLED/commit/3402a052c0b60266b742ce99d9b41a412f286dd0) to [001b882](https://github.com/Aircoookie/WLED/commit/001b882e11e6d70f104c6984f94b35ff53c93659) [Compare](https://github.com/Aircoookie/WLED/compare/3402a052c0b60266b742ce99d9b41a412f286dd0..001b882e11e6d70f104c6984f94b35ff53c93659)[3 days ago](https://github.com/Aircoookie/WLED/pull/4427#event-15760460626)

softhack007:
@Liliputech I think the usermod needs some more fixing and improvements.

* I've tried to compile it with our default esp32dev environment, and it just fails inside the PNGdec library (bitbank2/PNGdec@^1.0.1). 
   Reverting to bitbank2/PNGdec@1.0.1 (i.e. pinned to version 1.0.1) allows me to successfully build.
* The usermod really needs a readme.md file to explain how to use it
* I like your effort to also make the usermod compatible with "MoonModules" style usermods :-)
* I'm wondering why "f" and "png" and some other functions are declared globally, with names that simply "invite trouble". 
   Normally everything should be defined inside the usermod class, unless there are very strong reasons to leave the class scope.

Liliputech:
> @Liliputech I think the usermod needs some more fixing and improvements.

Hi @softhack007 Thank you for your feedback :)
I'll make the requested changes as soon as I can (probably by the end of next week, to give a good start to 2025!)

>     * I've tried to compile it with our default esp32dev environment - even with your patches, compilation just fails inside the PNGdec library (bitbank2/PNGdec@^1.0.1).
>       Reverting to bitbank2/PNGdec@1.0.1 (i.e. pinned to version 1.0.1) allows me to successfully build.
Maybe I should try to fix that by updating my code to fit the new version of PNGDEC instead then.

>     * The usermod really needs a readme.md file to explain how to use it
Sure! And a couple pictures!

>     * I like your effort to also make the usermod compatible with "MoonModules" style usermods :-)
Glad you liked it :)

>     * I'm wondering why `f` and `png` and some other functions are declared globally, with names that simply "invite trouble".
>       Normally everything should be defined inside the usermod class, unless there are very strong reasons to leave the class scope.
Got to fix this indeed.
 
>     * Do you know why `setPixelColor` is used, and not `setPixelColorXY` ? Is this POV usermod intended for 1D strips, or for 2D setups?
It is intended for 1D strip, the idea is to have an imaged displayed on a rotating strip.

>     * the line buffer in the draw function (`uint16_t usPixels[SEGLEN]`) seems much too large, which is another case of "invites trouble"
Unfortunately there is not much to be done here if you want to properly decode PNG.
The GetLineAsRGB565 from PNGDEC need to store each pixel of a line in an int16, and the whole line in an array of the corresponding length (picture width).
The idea is to have a whole line of pixel displayed on the strip. then another, etc...
If the strip is moving fast enough and given the image width equals the strip length, you can get floating images in the air, like an hologram.

> Btw, do you know who is "Arthur Suzuki"? he appears as the author of initial commit, but does not have a github account. He has directly committed the usermod into the Aircoookie repo, so he must be a maintainer? or maybe just a hacker who abused a security flaw in github?

I am Arthur :) you can find the information elsewhere by searching for my nickname or profile picture ;)
AFAIK I'm not allowed to push straight or merge to Aircookie repo... Feel free to PM me if needed, I'd be happy to discuss with fellow WLED hackers! (sometime do when connected to discord, same nickname)
Have a nice 2025!
Arthur

dosipod:
@Liliputech Since you are fixing things might be consider adding a readme as it is strange the usermod even got merged without one

Liliputech:
Actually I'm thinking about re-using the routines provided by the "image_loader.cpp" file provided by the GIF decoder effect. That would make a lot more sense :)

Liliputech:
`* I'm wondering why `f` and `png` and some other functions are declared globally, with names that simply "invite trouble".
  Normally everything should be defined inside the usermod class, unless there are very strong reasons to leave the class scope.`

I can't reuse the routines from image_loader.cpp because the functions signatures are too different...
And unfortunately I cannot make most of the routine as method of a class because the compiler is complaining of non-static functions and variables : 
wled00/../usermods/pov_display/usermod_pov_display.h:11:120: error: invalid use of non-static member function 'void PovDisplay::draw(PNGDRAW*)

I will just rename my methods in a more explicit and non-conflicting way and provide a README.

Liliputech:
just to mention, the functions will only be compiled if the USERMOD_POV_DISPLAY flag is set, which for most wled user won't be the case. I don't think this usermod will have that much impact on other things in wled codebase.

DedeHai:
@Liliputech can you please update it to latest main?

Liliputech:
Hi sorry for the delay, I've reworked the usermod to make it better, and also rebased on current main.

DedeHai:
Thanks for the update. 
I have no way to test this. Can you extend the readme a little? like what setup is needed (LED type? how does the image relate to LEDs etc.). A video of the UM in action would be helpful too.

Liliputech:
> Thanks for the update. I have no way to test this. Can you extend the readme a little? like what setup is needed (LED type? how does the image relate to LEDs etc.). A video of the UM in action would be helpful too.

I've addressed this comment, added a small gif image to show it working in motion, added some information to the readme, fixed all suggestions from coderabbit :)

usermods/pov_display/README.md
```diff
@@ -2,16 +2,40 @@
...
+![the usermod at work](pov_display.gif?raw=true)
+
 To get it working:
 - Resize your image. The height must match the number of LEDs in your strip/segment.
 - Rotate your image 90° clockwise (height becomes width).
 - Upload a BMP image (24-bit, uncompressed) to the ESP filesystem using the “/edit” URL.
 - Select the “POV Image” effect.
 - Set the segment name to the absolute filesystem path of the image (e.g., “/myimage.bmp”).
+- The path is case-sensitive and must start with “/”.
 - Rotate the segment at approximately 20 RPM.
```
DedeHai: you mean rotate the LED Strip, not segment :)

DedeHai:
looks good. 
does this work on all platforms including ESP8266?
if ESP8266 does not have nough RAM or is too slow it should be excluded with an `#ifdef / #error`

willmmiles:
> looks good. does this work on all platforms including ESP8266? if ESP8266 does not have nough RAM or is too slow it should be excluded with an `#ifdef / #error`

I'd prefer instead to specify the supported platform(s) in `library.json`.  Avoid `#ifdef` wherever possible!

Liliputech:
> looks good. does this work on all platforms including ESP8266? if ESP8266 does not have nough RAM or is too slow it should be excluded with an `#ifdef / #error`

I've restrained to esp32, maybe it could work on esp8266 but I don't have the hardware at hand.

usermods/pov_display/README.md
Liliputech: sure, changed that :)

DedeHai:
a buffer of 64k is off limits for ESP8266, it has less than 20k of free heap, so restriction is ok.

Liliputech:
ok, applied all suggested changes, if everything is fine this is ready to merge for me!

Liliputech:
thank you very much for your approval!

PR #4428, opened 2024-12-25
FX: Waterfall, Matripix & Dissolve fix

### LINKED ISSUE #4401: Some LEDS stay on after turning off WLED with Dissolve effect
### What happened?

When using the dissolve effect on my esp8266 based Dig-Uno controller, turning off WLED leaves some of the leds running. I am unable to reproduce this on my other controllers, but they are all esp32 based, so either I have a bad esp8266, or a bad dig-uno, or this issue is something specific to the esp8266.

### To Reproduce Bug

1. Use an esp8266 based controller with WLED 0.15.0
2. Configure a WS281x LED output with 50 LEDs
3. Use the following preset JSON:
```json
{"on":true,"bri":32,"transition":7,"mainseg":0,"seg":[{"id":0,"start":0,"stop":50,"grp":1,"spc":0,"of":0,"on":true,"frz":false,"bri":255,"cct":127,"set":0,"n":"","col":[[255,0,0],[8,255,0],[0,0,0]],"fx":18,"sx":128,"ix":255,"pal":0,"c1":128,"c2":128,"c3":16,"sel":true,"rev":false,"mi":false,"o1":false,"o2":false,"o3":false,"si":0,"m12":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0}]}
```
4. Run the controller with that preset
5. Press the power button in the top left of the UI
6. Observe that some of the LEDs stay on (not sure if this happens every time, but it definitely seems to happen more often than not

### Expected Behavior

All LEDs should be powered off when the controller is turned off

### Install Method

Binary from WLED.me

### What version of WLED?

0.15.0

ESP8266

### Relevant log/trace output

_No response_

### Anything else?

One thing that I have observed is that on my esp32 based controllers, when turning the controller off with the dissolve effect active, the whole strip fades down to being off together.  On my esp8266 based controller the behavior is different, it looks like the effects "dissolves to off" if that makes sense.

Rather than the whole strip fading down and turning off, the LEDs turn off one at a time randomly, the same way the dissolve effect would transition from one color to the next, they are just dissolve transitioning to off.  It seems like while this is working, the controller just doesn't wait for the "dissolve to off" to complete, so a few LEDs get left on when the controller is fully powered down.

On my strip of 50 LEDs, it's typical for 2-5 LEDs to be left on if it is turned off while the dissolve transition is running.

Ideally, I would think the behavior should be the same between the esp32 and the esp8266, it should do a fade to off rather than a dissolve to off.

### Code of Conduct

### LINKED ISSUE #4416: Problem with Arc option (1d expand FX)
### What happened?

After the 0.15 b4 update to release I got a bug for the 2D matrix.
The arc modifier stopped working adequately for some
sound effects (waterfall,  matripix). The diodes light up only at the edge and there is no animation.
I tried a clean installation and installing the night version 0.16, but it did not help.
Can anyone confirm or deny the existence of the problem?

### To Reproduce Bug

Enable the Waterfall effect on the 2D matrix (you can use sine instead mic) and compare the animation with "1d expand FX" modifier "corner" and "arc". 

### Expected Behavior

Animation on the entire matrix area for both options

### Install Method

Binary from WLED.me

### What version of WLED?

WLED 0.15.0 (build 2412100)

ESP32

### Relevant log/trace output

_No response_

### Anything else?

_No response_

### Code of Conduct

blazoncek:
- for Arc expansion
- or gaps
- or ledmaps with missing pixels

Fixes #4416
Fixes #4401


softhack007:
I'm not sure that this is a good approach for fixing the problem.

Its basicially a way to implement setUpLeds() without implementing it ;-) many audioreactive effects are doing a "shift left" or "shift out", and I guess it's just a matter of time until the next user comes up with a combination  of matrix dimensions and 2D expand mode where the same fix would be needed in another effect.

* duplication of code
* the fix would need to be removed again once we implement segment level LEDs buffering
* 8266 might run into memory problems (yeah it would also happen with a real "setUpLeds()" implemented in the segment class)
* it's not strictly necessary to use the memory book keeping of `(!SEGENV.allocateData(dataSize)` for this purpose.

On the positive side, 
* this proposed fix shows a way to solve the problem in a generic manner. Maybe we should still separate out the buffer allocation/reallocation into a separate  function `segment::setUpLedsV2()` and add a user option to disabled segment buffers -i  case that memory gets tight. getpixelcolor/ setPixelColor could detect that a buffer was allocated, and handle buffer operations transparently -  so effect authors just need to call SEGENV.setUpLedsV2() nothing more.
* the fix will work with arbitrary ledMaps and even with gapMaps.

blazoncek:
Well, do as you please. 🤷‍♂️ I proposed a solution that works, has least impact and adds very little to code size.

If you call `pixels[]` code duplication, well, you have to store values somewhere. `gPC()` seems a waste of resources in such case, `setUpLeds()` or not. This is by far the fastest approach (as data is only allocated once ad will only grow if needed).

Tested on ESP8266 with up to 512 pixels (which is maximum recomended for it).

blazoncek:
Oh, one more thing. `allocateData()` is intended fot that particular purpose - persistent effect data - no matter in what form that data is. Pixels or otherwise.

`setUpLeds()` was used because audio effects were mostly ported FastLED effects that barely fit into WLED framework (and hence the issues we cope with now, remember Matrix?) and relied on `leds[]` array. `leds[]` did not know of ledmaps or gaps. But we got rid of `leds[]` since then (for good or bad).

IMO any effect that needs persistent data (pixels or other) should utilise `allocateData()` and use well known fallback if running out of memory (we are under a lot less pressure since @willmmiles optimised AWS and other stuff). And, yes I know it, that would require updating some effects.

There is a `getPixelColor()` on the NPB roadmap that will return unmodified value but that will still not solve gaps in LED layout.

DedeHai:
IMHO this is a good fix for the issue at hand. Implementing local buffers i.e. setUpLeds() (which I am not familar with) sounds like something that requires a lot of code-rework and out of scope for the moment.

blazoncek:
FYI there are several more effects that will not work correctly with gaps (and/or 2D expansion).

@DedeHai I invented `setUpLeds()` in one of the early prototypes of 0.14 when WLED still relied only on NPB `getPixelColor()` to retrieve previous frame's pixel values. `gPC` was not accurate and lead to color distortion at lower brightness. Audio effects (from SR fork) were written using FastLED and relied heavily on FastLED's `leds[]` array which did not exist in AC. With `setUpLeds()` a local approximation was created so code like `leds[x]=leds[x-1]` could work as expected. Unfortunately RAM was scarce and code still needed to be modified so in later versions (after MM split off) me and @Aircoookie modified that to use global buffer allocated on bus level. This change was never ported to MM.

Unfortunately bus knows nothing about gaps (as it is working on physical layer) and hence some effects can't behave properly if they rely on previous pixel values that need to be moved around.
Handling gaps needs to happen on the layer where they are introduced (`strip`) or higher up. Handling them in effect function works fastest but requires the most amount of RAM. It also introduces issues when you try to blend/overlap two segments. I have written elsewhere that effects have to be rewritten if they are to support overlapping or the change needs to go into `strip` logic to handle overlapping segments. The latter may prove challenging.

netmindz:
Can one of the reviewers please approve if this it ready to merge (and cherry pick to the release branch) ?

netmindz:
There were some conflicts merging back to 0_15_0 so it might be best for you to check the fix is still correct @blazoncek

blazoncek:
Looks ok.

netmindz:
> Looks ok.

Have you done the cherry-pick and fixed the conflicts @blazoncek ?

PR #4430, opened 2024-12-25
Use proper devcontainers schema for vscode customisations

inverse:
When working on #4424 I noticed the devcontainers definition was incorrect. It should be nested under the tooling for `vscode` See: https://containers.dev/supporting#visual-studio-code


netmindz:
Please never use force push on a PR

Issue #4417, opened 2024-12-22
Back button on Update.htm


Jacobnordvall:
### What happened?

Back button dosnt go back to index if the last page wast index. aka you clicked on git banner to get latest bin.

### To Reproduce Bug

Click the git banner. then use back gesture on android. then use the back botton. It now takes you back to the git page as it tracks the last page and dosnt actually go to index page.

### Expected Behavior

back goes back to last entry in the main page.

### Anything else?

Here is a fully working solution for reference:

```js
function goBackToMainPage() {
    let stepsBack = 0;
    const mainPages = ['/#Colors', '/#Effects', '/#Segments', '/#Presets'];

    function isMainPage() {
        const currentPage = window.location.href;
        return mainPages.some(page => currentPage.includes(page));
    }

    function navigateBack() {
        if (stepsBack < 60) { // Prevent infinite loop
            stepsBack++;
            window.history.back();

            setTimeout(() => {
                if (!isMainPage()) {
                    navigateBack();
                }
            }, 100); // Wait for navigation to take effect
        } else {
            // Fallback if no "main" page is found
            window.location = '/';
        }
    }

    if (!isMainPage()) {
        navigateBack();
    } else {
        // If already on a main page
        console.log("You're already on a main page!");
    }
}
```

and ofc the html:
```html
<button type="button" onclick="goBackToMainPage()">Back</button>
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Jacobnordvall:
that code (or a similar fix) should also be used for the settings back button preferably. as it goes back to the actuall last index page and not just to colors no matter what.

blazoncek:
Make a PR, please.

Jacobnordvall:
> Make a PR, please.

Want me to put that logic in common.js? 
As it can be used for proper back functionality for settings page too and others?

DedeHai:
keep in mind that flash memory is at a premium, keep the code as compact as possible even if it means cutting corners, IMHO this is solving an inconvenience which currently minimal code should be spent on.

blazoncek:
Also verify functionality behind reverse proxy.

Jacobnordvall:
> Also verify functionality behind reverse proxy.
im not entirely sure what that means tbh.
its a line swap fix. if its just to adress the bug.

[Pull request](https://github.com/Aircoookie/WLED/pull/4431)

blazoncek:
[...]

PR #4431, opened 2024-12-26
Fix back button in Update.htm

Jacobnordvall:
Current code falls appart if you click on the git banner etc as it fetches the last page in the window history instead of properly going back. I was asked to open a PR so here it is.. 

this is a fix for issue [#4417](url)

### DIFF
--- a/wled00/data/update.htm
+++ b/wled00/data/update.htm
@@ -4,7 +4,7 @@
 	<meta content='width=device-width' name='viewport'>
 	<title>WLED Update</title>
 	<script>
-		function B() { window.history.back(); }
+		function B() { window.location = '/'; }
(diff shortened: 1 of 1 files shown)


blazoncek (changes requested):
That's wrong.
The navigation should go back to previous page. Which is in history.
Previous page is either Security settings or main UI, depending on the entry point.

Jacobnordvall:
> That's wrong. The navigation should go back to previous page. Which is in history. Previous page is either Security settings or main UI, depending on the entry point.

Back should go back in wled and not back to another last visited non wled page (git in this case)

Its a bug on mobile as a new tab wont be opened.

DedeHai:
Tested original behaviour on android: a new tab IS opened. Can you describe in more detail how to reproduce the unexpected behaviour you are seeing?

PR #4435, opened 2024-12-27
palette effect overflow fix

### LINKED ISSUE #4425: Palette effect will not spread palette colors across entire strip
### What happened?

If the strip (or segment) pixel count is larger than 256 pixels, the effect will spread colors only across 256 pixels and then repeat, instead of spreading colors across entire strip (or segment) length.

### To Reproduce Bug

Create a strip (or segment) with more than 256 pixels (preferably more than 384 to pronounce the effect) and select Palette effect.
You will see a sharp color transition at pixel 256.

### Expected Behavior

Spread palette colors across entire strip (or segment) even if pixel count is larger than 256 pixels.

### Install Method

Self-Compiled

### What version of WLED?

WLED 0.15.0

ESP8266

### Relevant log/trace output

_No response_

### Anything else?

The bug was introduced with #3683 ~~when using `SEGMENT.color_wheel()` which only takes 256 values instead of `SEGEMENT.color_from_palette()`~~.
cc @TripleWhy  

### Code of Conduct

TripleWhy:
I belive this fixes #4425

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -2032,7 +2032,7 @@ uint16_t mode_palette() {
       const mathType sourceX = xtSinTheta + ytCosTheta + centerX;
       // The computation was scaled just right so that the result should always be in range [0, maxXOut], but enforce this anyway
       // to account for imprecision. Then scale it so that the range is [0, 255], which we can use with the palette.
-      int colorIndex = (std::min(std::max(sourceX, mathType(0)), maxXOut * sInt16Scale) * 255) / (sInt16Scale * maxXOut);
+      int colorIndex = (std::min(std::max(sourceX, mathType(0)), maxXOut * sInt16Scale) * wideMathType(255)) / (sInt16Scale * maxXOut);
       // inputSize determines by how much we want to scale the palette:
       // values < 128 display a fraction of a palette,
       // values > 128 display multiple palettes.


TripleWhy:
Note: this does replace some more 32 bit multiplications with slightly more expensive 64 bit multiplications.

blazoncek:
Tested and works.

wled00/FX.cpp
```diff
@@ -2032,7 +2032,7 @@ uint16_t mode_palette() {
       const mathType sourceX = xtSinTheta + ytCosTheta + centerX;
       // The computation was scaled just right so that the result should always be in range [0, maxXOut], but enforce this anyway
       // to account for imprecision. Then scale it so that the range is [0, 255], which we can use with the palette.
-      int colorIndex = (std::min(std::max(sourceX, mathType(0)), maxXOut * sInt16Scale) * 255) / (sInt16Scale * maxXOut);
+      int colorIndex = (std::min(std::max(sourceX, mathType(0)), maxXOut * sInt16Scale) * wideMathType(255)) / (sInt16Scale * maxXOut);
```
softhack007, line 2035: What does `wideMathType()` do? 

Essentially this PR changes "255" to "wideMathType(255)", I'm wondering is this the same as writing "255.0f" or maybe "255UL" ?

blazoncek: 255ULL for ESP8266 and 255.0f otherwise

softhack007: Looks like someone tried to avoid fixed point math?

DedeHai: the way it is implemented is quite an elegant way to support float AND fixed point math.

DedeHai (approved):
simple fix, approved.

PR #4442, opened 2024-12-30
Fix output glitches when playlist changes preset for S2 and C3

DedeHai:
same issue as with https://github.com/Aircoookie/WLED/pull/4386
waiting on bus to finish updating before file access fixes the glitches.
this issue is only present on S2 and C3, not on ESP8266 or dual-core ESPs, the fix is only applied for these two.

### DIFF
--- a/wled00/presets.cpp
+++ b/wled00/presets.cpp
@@ -164,6 +164,11 @@ void handlePresets()
 
   DEBUG_PRINTF_P(PSTR("Applying preset: %u\n"), (unsigned)tmpPreset);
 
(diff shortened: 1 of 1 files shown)


netmindz:
@DedeHai has not been merged to 0_15_x and has merge conflict. can you take a look please?

DedeHai:
@netmindz the follow-up commit was merged, so its all good: https://github.com/wled/WLED/commit/5fc2175dd4930f713eece93f1efa26e4f071da85

PR #4450, opened 2025-01-03
Add correct pin availability for ESP32 Mini modules

ladyada:
ESP32 Mini modules (https://www.espressif.com/en/module/esp32-mini-1-en) from Espressif have /slightly/ different pins available due to having internal SPI flash on different pins than WROVER/WROOMs. this code checks if its part of the "ESP32-U4WDH" family (or, due to https://github.com/espressif/arduino-esp32/issues/10683 may be called "ESP32-PICO-D2")

### DIFF
--- a/wled00/pin_manager.cpp
+++ b/wled00/pin_manager.cpp
@@ -214,7 +214,18 @@ bool PinManager::isPinOk(byte gpio, bool output)
     // JTAG: GPIO39-42 are usually used for inline debugging
     // GPIO46 is input only and pulled down
   #else
-    if (gpio > 5 && gpio < 12) return false;      //SPI flash pins
+
+    if (strncmp_P(PSTR("ESP32-U4WDH"), ESP.getChipModel(), 11) == 0) {
+      // this chip has 4 MB of internal Flash and different packaging, so available pins are different!
+      if ((gpio == 1) || (gpio == 3) || ((gpio >= 6) && (gpio =< 8)) ||
+          (gpio == 11) || (gpio == 16) || (gpio == 17) || (gpio == 20) || 
+          (gpio == 24) || ((gpio >= 28) && (gpio <= 31)))
+        return false;
+    } else {
+      // for classic ESP32 (non-mini) modules, these are the SPI flash pins
+      if (gpio > 5 && gpio < 12) return false;      //SPI flash pins
+    }
+
     if (strncmp_P(PSTR("ESP32-PICO"), ESP.getChipModel(), 10) == 0 && (gpio == 16 || gpio == 17)) return false; // PICO-D4: gpio16+17 are in use for onboard SPI FLASH
     if (gpio == 16 || gpio == 17) return !psramFound(); //PSRAM pins on ESP32 (these are IO)
   #endif
--- a/wled00/pin_manager.cpp
+++ b/wled00/pin_manager.cpp
@@ -215,7 +215,9 @@ bool PinManager::isPinOk(byte gpio, bool output)
     // GPIO46 is input only and pulled down
   #else
 
+    Serial.printf("GPIO TEST %d\n\r", gpio);
     if (strncmp_P(PSTR("ESP32-U4WDH"), ESP.getChipModel(), 11) == 0) {
+      Serial.println("U4WDH");
       // this chip has 4 MB of internal Flash and different packaging, so available pins are different!
       if ((gpio == 1) || (gpio == 3) || ((gpio >= 6) && (gpio <= 8)) ||
           (gpio == 11) || (gpio == 16) || (gpio == 17) || (gpio == 20) || 
--- a/wled00/pin_manager.cpp
+++ b/wled00/pin_manager.cpp
@@ -215,9 +215,8 @@ bool PinManager::isPinOk(byte gpio, bool output)
     // GPIO46 is input only and pulled down
   #else
 
-    Serial.printf("GPIO TEST %d\n\r", gpio);
-    if (strncmp_P(PSTR("ESP32-U4WDH"), ESP.getChipModel(), 11) == 0) {
-      Serial.println("U4WDH");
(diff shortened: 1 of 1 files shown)


wled00/pin_manager.cpp
```diff
@@ -214,7 +214,19 @@ bool PinManager::isPinOk(byte gpio, bool output)
     // JTAG: GPIO39-42 are usually used for inline debugging
     // GPIO46 is input only and pulled down
   #else
-    if (gpio > 5 && gpio < 12) return false;      //SPI flash pins
+
+    if ((strncmp_P(PSTR("ESP32-U4WDH"), ESP.getChipModel(), 11) == 0) ||    // this is the correct identifier, but....
+        (strncmp_P(PSTR("ESP32-PICO-D2"), ESP.getChipModel(), 13) == 0)) {  // https://github.com/espressif/arduino-esp32/issues/10683
+      // this chip has 4 MB of internal Flash and different packaging, so available pins are different!
+      if ((gpio == 1) || (gpio == 3) || ((gpio >= 6) && (gpio <= 8)) ||
```
softhack007, line 221: Hi @ladyada, thanks for your contribution 👍 

Is it necessary to also mark 1 and 3 (RX and TX) as unusable?

If I read the datasheet correctly, it looks like the board basicially has 16 and 17 connected to spi flash, while 9 and 10 are still availeable for use. 

There is a note in the datasheet rev 1.4 saying "GPIO6, GPIO7, GPIO8, GPIO11, GPIO16, and GPIO17 on the ESP32-U4WDH chip are connected to the SPI flash integrated in the chip and cannot be used for other purposes except for connecting external
PSRAM."

softhack007:
Do you know if the ESP32 Mini needs a newer bootloader (esp-idf 4.4.x), or would it also work with the older esp-idf 3.5.x / arduino-esp32 1.0.6 framework?

ladyada:
not sure about the bootloader, i tested it with the one here
https://github.com/Aircoookie/WLED/releases/download/v0.13.1/esp32_bootloader_v4.bin
and with whatever the default out of PlatformIO 'main' branch is. if you post a link to the older bootloader i can test it on hardware

wled00/pin_manager.cpp
ladyada: i can adjust the statement to just reject 16/17, allow 9/10 and let the defaults determine the rest.

softhack007: Yes, sounds good. Please do.

softhack007:
> i tested it with the one here
> https://github.com/Aircoookie/WLED/releases/download/v0.13.1/esp32_bootloader_v4.bin

This is the "old framework" bootloader. If the module works with it, the the mini1 should work with our standard bootloader (we're still on the old framework due to "200kb for nothing" flash increase when using a newer framework....).

wled00/pin_manager.cpp
ladyada: done!

wled00/pin_manager.cpp
```diff
@@ -214,8 +214,20 @@ bool PinManager::isPinOk(byte gpio, bool output)
...
+        (strncmp_P(PSTR("ESP32-PICO-D2"), ESP.getChipModel(), 13) == 0)) {  // https://github.com/espressif/arduino-esp32/issues/10683
+      // this chip has 4 MB of internal Flash and different packaging, so available pins are different!
+      if (((gpio > 5) && (gpio < 9)) || (gpio == 11))
+        return false;
+    } else {
+      // for classic ESP32 (non-mini) modules, these are the SPI flash pins
+      if (gpio > 5 && gpio < 12) return false;      //SPI flash pins
+    }
+
+    if (((strncmp_P(PSTR("ESP32-PICO"), ESP.getChipModel(), 10) == 0) ||
```
DedeHai, line 228: what is the difference between "ESP32-PICO" and "ESP32-PICO-D2" which both also correspond to "ESP32-U4WDH"? the double checking is a bit confusing to me, can you add comments on why this is needed?

blazoncek:
FYI using bootloader from older versions (PIO 3.5 or earlier) is prone to flash corruption. There is a post/comment from @Jason2866 regarding that. You must only use DOUT flash mode if using that bootloader.

My recommendation would be to finally switch to PIO platform 6.3.x for ESP32 and then update necesary GPIO reservations using chip names.

codeman256:
That's funny, I'm actually running in the same problem but for the ESP32-PICO-V3-02 and need pin 7 available. Can I piggyback off this and we add an additional check in place for the V3 and V3-02 somehow? Only gpio pins 6,9,10,11 should be blocked, not 7,8.

image
[https://www.espressif.com/sites/default/files/documentation/esp32-pico_series_datasheet_en.pdf](https://www.espressif.com/sites/default/files/documentation/esp32-pico_series_datasheet_en.pdf)

softhack007:
@codeman256 I'm sorry but your board is another chip, and I don't see a way to "mee-too" your board on this PR. Please research how to identify your special board with `ESP.getChipModel()` and open a separate PR.

@blazoncek I only wanted to explain why we still use an old framework, and for me a discussion of framework versions should be done separately. Its has nothing to do with the esp32 mini1 module.

@ladyada thanks for the changes, I'll test your code quickly in the next days and then we should be ready to merge.

softhack007 (approved):
OK from my side - tested against ESP32-PICO-D4 to ensure that changes for ESP32-PICO-D2  are not affecting pin availability on other PICO modules.

wled00/pin_manager.cpp
softhack007: @DedeHai I can explain partially: there are actually 3 PICO boards:
* ESP32-PICO-D4 (already supported)
* ESP32-PICO-V3-02 (our old framework does not support this type)
* ESP32-PICO-D2 (older frameworks wrongly identify the ESP32-U4WDH as PICO-D2)

All PICO boards, plus the U4WDH, have in common that gpio 16+17 are reserved for internal SPI flash. So blocking 16+17 for all these variants is correct.

DedeHai (approved):

PR #4455, opened 2025-01-03
Playlist item with infinite duration

joosthoi1:
Infinite duration for playlist presets, so playlists can be controlled by just calling the next playlist api call.

The ui isn't great.

### DIFF
--- a/wled00/data/index.js
+++ b/wled00/data/index.js
@@ -2090,12 +2090,12 @@ function makePlEntry(p,i)
 		<td class="c"><button class="btn btn-pl-add" onclick="addPl(${p},${i})"><i class="icons btn-icon">&#xe18a;</i></button></td>
 	</tr>
 	<tr>
-		<td class="c">Duration</td>
+		<td class="c">Duration (0 for infinite)</td>
 		<td class="c">Transition</td>
 		<td class="c">#${i+1}</td>
 	</tr>
 	<tr>
-		<td class="c" width="40%"><input class="segn" type="number" placeholder="Duration" max=6553.0 min=0.2 step=0.1 oninput="pleDur(${p},${i},this)" value="${plJson[p].dur[i]/10.0}">s</td>
+		<td class="c" width="40%"><input class="segn" type="number" placeholder="Duration" max=6553.0 min=0 step=0.1 oninput="pleDur(${p},${i},this)" value="${plJson[p].dur[i]/10.0}">s</td>
 		<td class="c" width="40%"><input class="segn" type="number" placeholder="Transition" max=65.0 min=0.0 step=0.1 oninput="pleTr(${p},${i},this)" value="${plJson[p].transition[i]/10.0}">s</td>
 		<td class="c"><button class="btn btn-pl-del" onclick="delPl(${p},${i})"><i class="icons btn-icon">&#xe037;</i></button></div></td>
 	</tr>
--- a/wled00/playlist.cpp
+++ b/wled00/playlist.cpp
@@ -6,7 +6,7 @@
 
 typedef struct PlaylistEntry {
   uint8_t preset; //ID of the preset to apply
-  uint16_t dur;   //Duration of the entry (in tenths of seconds)
+  uint16_t dur;   //Duration of the entry (in tenths of seconds) (0 = infinitely)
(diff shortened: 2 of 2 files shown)


softhack007:
@joosthoi1 the existing checkbox "Repeat indefinitely" did not work for you?

DedeHai:
@softhack007  this change would allow for infinite play per entry, not the whole list. its a "hack" to allow preset grouping. I did not have time yet to look at the nested playlist commit from @blazoncek, that may also do something similar from what I was able to pick up on discord.
I like the general approach taken here but need to check out nested playlists first.

blazoncek:
There isan issue with this (and the reason I haven't published my solution 2 days ago).
If you accidentally enter 0 for one of the entries playlist will stop at that point and may cause user frustration. My prefered method would be to add a selection checkbox stating that this is a *manual advance* playlist and hide duration entry field.

There's also a nonsense in this code. I think you will easily spot it. 😉

joosthoi1:
> If you accidentally enter 0 for one of the entries playlist will stop at that point and may cause user frustration. My prefered method would be to add a selection checkbox stating that this is a *manual advance* playlist and hide duration entry field.

Do you think it should still be implented so that duration = 0 is manual advance, or should there be a separate variable to stop automatic advancing?

I think duration 0 could still work and we'd just update the UI to abstract this away. I completely agree with you that this ui at the moment isn't great and a checkbox would be better.

> I haven't published my solution 2 days ago)
The only reason I uploaded this was because I couldn't find your commit and wanted to use it on my personal machine. Feel free to close this pr if your implementation is better. I already got what I need :P

> There's also a nonsense in this code. I think you will easily spot it. 😉

I'm not sure of the top of my head what you mean, if you mean dur > 1 or dur == 0, that can definitely be simplified to dur != 1 lol. I made this code late last night, so it's entirely stupid.

joosthoi1:
> If you accidentally enter 0 for one of the entries playlist will stop at that point and may cause user frustration. My prefered method would be to add a selection checkbox stating that this is a *manual advance* playlist and hide duration entry field.

Do you think it should still be implented so that duration = 0 is manual advance, or should there be a separate variable to stop automatic advancing?

I think duration 0 could still work and we'd just update the UI to abstract this away. I completely agree with you that this ui at the moment isn't great and a checkbox would be better.

> I haven't published my solution 2 days ago)
The only reason I uploaded this was because I couldn't find your commit and wanted to use it on my personal machine. Feel free to close this pr if your implementation is better. I already got what I need :P

> There's also a nonsense in this code. I think you will easily spot it. 😉

I'm not sure of the top of my head what you mean, if you mean dur > 1 or dur == 0, that can definitely be simplified to dur != 1 lol. I made this code late last night, so it's entirely stupid.

Edit: ah, I also saw that I didn't change how I checked if dur == 0 in handleplaylist, so that implementation is stupid aswell

blazoncek:
Screenshot 2025-01-04 at 11 18 43
That's better. Will be available shortly.

blazoncek:
67c4dbf2da5999fcd46075a75f7dcc36a57983fe
builds upon nested playlists

joosthoi1:
> [67c4dbf](https://github.com/Aircoookie/WLED/commit/67c4dbf2da5999fcd46075a75f7dcc36a57983fe) builds upon nested playlists

I haven't tested this yet, but doesn't this allow for duration = 1 (0,10 seconds)? which is specifically not allowed as explained in https://github.com/Aircoookie/WLED/issues/4164#issuecomment-2380632417 

Other than that, looks good

blazoncek:
100ms is about 4 frames, plenty of time for ESP8266 to switch preset.
Will it crash? Not because switching presets from a playlist. Using UI to switch presets as quickly is completely different story.
Why 200ms originally? Because you would have 100ms transition time and 100ms effect duration making effect never properly show.
Will this matter? Most likely not as additional code will make sure you cannot have transition longer than effect duration.

joosthoi1:
Okay, that's fine then

joosthoi1:
So how do we proceed from here? Do you make a seperate pull request, or add the changes to my branch?

blazoncek:
#4568 includes manual advance playlist as well as may other bugfixes and enhancements. It supersedes this PR, making it obsolete.

PR #4457, opened 2025-01-04
Implement WS2811 WW/CW

lmoe:
This should add support for WS2811 warm white / cold white strips. 
They work by using the WS2812 protocol (RGB) but using WW for R, CW for G and ignoring B.

I have seen support for WS2811 by using `WS2811 White` but this doesn't support CCT. 
You can use the original WS2812 type, but this makes it hard to properly tune CCT without losing brightness.

I have seen quite a few special if/else blocks for certain types, which made following the code a bit difficult for me. I've thought adding separate type handlers might make things easier to work on, even if it's more code in the end. 

Tested it with an ESP32.

Hope the PR is fine to submit like this :)

### DIFF
--- a/wled00/bus_wrapper.h
+++ b/wled00/bus_wrapper.h
@@ -89,6 +89,11 @@
 #define I_8266_U1_SM16825_5 104
 #define I_8266_DM_SM16825_5 105
 #define I_8266_BB_SM16825_5 106
+//WS2811 (WWCW)
+#define I_8266_U0_NEO_WS2811_WWCW 107
+#define I_8266_U1_NEO_WS2811_WWCW 108
+#define I_8266_DM_NEO_WS2811_WWCW 109
+#define I_8266_BB_NEO_WS2811_WWCW 110
 
 /*** ESP32 Neopixel methods ***/
 //RGB
@@ -139,6 +144,11 @@
 #define I_32_RN_SM16825_5 107
 #define I_32_I0_SM16825_5 108
 #define I_32_I1_SM16825_5 109
+//WS2811 (WWCW)
+#define I_32_RN_NEO_WS2811_WWCW 110
+#define I_32_I0_NEO_WS2811_WWCW 111
+#define I_32_I1_NEO_WS2811_WWCW 112
+#define I_32_I1_NEO_WS2811_WWCW_P 113
 
 //APA102
 #define I_HS_DOT_3 39 //hardware SPI
@@ -226,6 +236,11 @@
 #define B_8266_U1_SM16825_5 NeoPixelBusLg<NeoRgbwcSm16825eFeature, NeoEsp8266Uart1Ws2813Method, NeoGammaNullMethod>
 #define B_8266_DM_SM16825_5 NeoPixelBusLg<NeoRgbwcSm16825eFeature, NeoEsp8266Dma800KbpsMethod, NeoGammaNullMethod>
 #define B_8266_BB_SM16825_5 NeoPixelBusLg<NeoRgbwcSm16825eFeature, NeoEsp8266BitBangWs2813Method, NeoGammaNullMethod>
+// W2811 (WWCW)
+#define B_8266_U0_NEO_WS2811_WWCW NeoPixelBusLg<NeoGrbFeature, NeoEsp8266Uart0Ws2813Method, NeoGammaNullMethod> //3 chan, esp8266, gpio1
+#define B_8266_U1_NEO_WS2811_WWCW NeoPixelBusLg<NeoGrbFeature,   NeoEsp8266Uart1Ws2813Method, NeoGammaNullMethod> //3 chan, esp8266, gpio2
+#define B_8266_DM_NEO_WS2811_WWCW NeoPixelBusLg<NeoGrbFeature,   NeoEsp8266Dma800KbpsMethod, NeoGammaNullMethod> //3 chan, esp8266, gpio3
+#define B_8266_BB_NEO_WS2811_WWCW NeoPixelBusLg<NeoGrbFeature, NeoEsp8266BitBang800KbpsMethod, NeoGammaNullMethod> //3 chan, esp8266, bb (any pin but 16)
 #endif
 
 /*** ESP32 Neopixel methods ***/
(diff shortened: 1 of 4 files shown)


softhack007:
@lmoe please don't use force-push while you have a PR open in the main WLED repo. You can simply push changes to your branch and they will be added to your PR automatically.

> [lmoe](https://github.com/lmoe) [force-pushed](https://github.com/Aircoookie/WLED/compare/f8f3ed0235d601bc83382de2813a1f284d508bbb..3d3ce446c5a50b274f7764f15c9c615fbc978927) the add_ws2811_wwcw branch from [f8f3ed0](https://github.com/Aircoookie/WLED/commit/f8f3ed0235d601bc83382de2813a1f284d508bbb) to [3d3ce44](https://github.com/Aircoookie/WLED/commit/3d3ce446c5a50b274f7764f15c9c615fbc978927) [Compare](https://github.com/Aircoookie/WLED/compare/f8f3ed0235d601bc83382de2813a1f284d508bbb..3d3ce446c5a50b274f7764f15c9c615fbc978927)[4 hours ago](https://github.com/Aircoookie/WLED/pull/4457#event-15816033252)

lmoe:
Ah, sorry. Just wanted to condense a second commit I've made after I opened the PR into one, and this was muscle memory. 
If it helps I can create a new PR and close this one. :) @softhack007

softhack007:
@lmoe we can continue with this PR, no need to create a new one. 

You might get feedback requesting code changes, so just remember to push new commits but don't use force-push.

wled00/bus_wrapper.h
```diff
@@ -89,6 +89,11 @@
 #define I_8266_U1_SM16825_5 104
 #define I_8266_DM_SM16825_5 105
 #define I_8266_BB_SM16825_5 106
+//WS2811 (WWCW)
+#define I_8266_U0_NEO_WS2811_WWCW 107
```
blazoncek, line 93: These values overlap.

wled00/bus_wrapper.h
```diff
@@ -139,6 +144,11 @@
 #define I_32_RN_SM16825_5 107
 #define I_32_I0_SM16825_5 108
 #define I_32_I1_SM16825_5 109
+//WS2811 (WWCW)
+#define I_32_RN_NEO_WS2811_WWCW 110
```
blazoncek, line 148: So do these.

wled00/bus_wrapper.h
```diff
@@ -226,6 +236,11 @@
 #define B_8266_U1_SM16825_5 NeoPixelBusLg<NeoRgbwcSm16825eFeature, NeoEsp8266Uart1Ws2813Method, NeoGammaNullMethod>
 #define B_8266_DM_SM16825_5 NeoPixelBusLg<NeoRgbwcSm16825eFeature, NeoEsp8266Dma800KbpsMethod, NeoGammaNullMethod>
 #define B_8266_BB_SM16825_5 NeoPixelBusLg<NeoRgbwcSm16825eFeature, NeoEsp8266BitBangWs2813Method, NeoGammaNullMethod>
+// W2811 (WWCW)
+#define B_8266_U0_NEO_WS2811_WWCW NeoPixelBusLg<NeoGrbFeature, NeoEsp8266Uart0Ws2813Method, NeoGammaNullMethod> //3 chan, esp8266, gpio1
```
blazoncek, line 240: There is no difference from B_8266_U0_NEO_WS2812.

wled00/bus_wrapper.h
```diff
@@ -900,9 +948,14 @@ class PolyBus {
       case I_8266_U1_SM16825_5: (static_cast<B_8266_U1_SM16825_5*>(busPtr))->SetPixelColor(pix, Rgbww80Color(col.R*257, col.G*257, col.B*257, cctWW*257, cctCW*257)); break;
       case I_8266_DM_SM16825_5: (static_cast<B_8266_DM_SM16825_5*>(busPtr))->SetPixelColor(pix, Rgbww80Color(col.R*257, col.G*257, col.B*257, cctWW*257, cctCW*257)); break;
       case I_8266_BB_SM16825_5: (static_cast<B_8266_BB_SM16825_5*>(busPtr))->SetPixelColor(pix, Rgbww80Color(col.R*257, col.G*257, col.B*257, cctWW*257, cctCW*257)); break;
+      case I_8266_U0_NEO_WS2811_WWCW: (static_cast<B_8266_U0_NEO_WS2811_WWCW*>(busPtr))->SetPixelColor(pix, RgbColor(cctWW, cctCW, 0)); break;
```
blazoncek, line 951: This should be done in BusDigital instead.

blazoncek (changes requested):
I do not like the idea of adding new wrapper type as the only difference is in RGB value for color.

There is already a method that works for single white addressable strips in BusDigital. Modify that and leave wrapper alone.

I would also keep original const as it was specified since WS281X type of chips share similar signal specifications.

Another downside of this approach is that UI has no clue this is not a RGB strip and will show RGB controls instead of hiding them.

One thing that remains open is the use of A (amber) channel (since there are WWA strips). I think that using 0 is acceptable until WLED is modified to support such strip fully.

Still, there may be strips that have ww/cw/ww, cw/ww/cw, ... order of LEDs (maximizing chip utilisation). This should be added but I still have no idea how to implement with as little change as possible. And I loath duplicating code!

blazoncek:
For anyone interested my fork (in non-stable branch) has implementation where WWA strip is implemented in `BusDigital` instead of wrapper.

DedeHai:
@blazoncek does https://github.com/Aircoookie/WLED/pull/4484 implement this?

blazoncek:
> does #4484 implement this and supersede this PR?

It does, but differently.

netmindz:
So which would you recommend gets merged @blazoncek ?

blazoncek:
@netmindz I, personally, do not like enhancing BusWrapper when BusDigital can do the stuff. And do it better at that.

I've now had a second look at this PR and noticed that it implements the *other* WW/CW strip option than WWA, `TYPE_WS2812_2CH_X3` which is meant to implement digital CCT by using 2 WS281x chips to encode 3 CCT pixels. As such this PR (and its implementation) is incorrect.

I would prefer #4484 (as it also fixes other issues).
Unfortunately I do not own WWA strip of any kind so #4484 may need channel swap for WW/CW. A is not implemented.

matteocali:
Could we use this PR to implement the support for WS2811 RGBCCT strips?
They use two WS2811 chip per segment. The first one controls RGB channels the second one the WW and CW using the red and green channel.

blazoncek:
No. Digital RGB+CCT (sometimes +WWA) is already implemented for FW1906 and WS2805 and SM16825 chips.
If you need other chip support, adapt from these.

matteocali:
Ok thanks, I'm not an expert but I will try.
Since I see many strips online with this kind of controller setup

PR #4459, opened 2025-01-06
DO NOT MERGE - review of AR usermod differences between branches

netmindz:


netmindz:
Rather that trying to cherry-pick many years worth of work on this usermod from MM to AC, it's probably just easier to copy the MM version over the top of the out of date version I think @softhack007 - however before we do this I though it useful to at least do a rough PoC of this and also use this PR as a code review opportunity, to pick up bits like how my AR are in the usermod rather than core as in MM plus any other bits that perhaps we don't want to bring over like FASTPATH

DedeHai:
appreciate the effort. I only briefly looked at the code, compared to to AC it looks a bit bloated. I would suggest to bring only the very essential pieces: the full code from this PR eats 10k of flash! The additional I2S board drivers should be added with compile flags as few will use those.

softhack007:
> compared to to AC it looks a bit bloated

yeah, agreed - there are some experimental things in the new code that could be left out, and some parts would be easier to read after a clean-up.

PR #4462, opened 2025-01-06
Playlist output glitchfix update: found it also happens on S3

DedeHai:


softhack007:
@DedeHai thanks 👍  
I think that the RMT output driver of NPB is most sensitive to timing, so it might be best to enable the fix on any esp32 variant ( `#ifdef ARDUINO_ARCH_ESP32` ). Maybe you did not observe the flickering on esp32 yet, because its a bit tricky to _not_ get the I2S driver:
* more than 2 outputs needed (output 0 will use I2S#1)
* more than 300 leds needed on one or more outputs (so the parallel I2S#1 driver cannot be used)

Did you see any hickups or bad side-effects with the glitch fix?

DedeHai:
I initially wanted to remove the ifdef, but could not get ESP32 to glitch... will try again to force it to RMT and see if that glitches.
will also test if it has a side-effect on I2S outputs when the fix is applied

DedeHai:
ran some more tests:
I can not get any glitches on the ESP32, not in I2S nor RMT output mode (tested with 4 outputs, one of them set to 400 LEDs). I found no difference with or without adding the "wait for bus". So it could be also added for ESP32 but I am a bit torn on doing that without any reason behind it.

PR #4467, opened 2025-01-08
Add support for WS2811 on ESP32 and ESP8266 platforms

arikon:


blazoncek:
I own WS2811 and they work happily with WS281x selection. Why adding duplicate? Did you encounter any issues?

arikon:
@blazoncek I have not WS8211, but LEDs with close timing to is. The model is RF-W7SA50TS-A51-IC. It refused to work with WS2812x on ESP32-C3 in my case.

The timings of them are:

arikon:
@blazoncek I could double check, if they will work with WS2812x

PR #4473, opened 2025-01-09
Added new effect usermod - Diffusion Fire

mryndzionek:
Usermod adding new fire effect. Has more parameters than other fire effects and can be widely adjusted/tweaked to look good even on small matrices.
fire

### DIFF
--- a/usermods/diffusionfire/usermod_diffusionfire.h
+++ b/usermods/diffusionfire/usermod_diffusionfire.h
@@ -1,0 +1,118 @@
+static uint16_t mode_static(void)
+{
+  SEGMENT.fill(SEGCOLOR(0));
+  return strip.isOffRefreshRequired() ? FRAMETIME : 350;
+}
+
+static uint16_t mode_diffusionfire(void)
+{
+  static uint32_t call = 0;
+
+  if (!strip.isMatrix || !SEGMENT.is2D())
+    return mode_static(); // not a 2D set-up
+
+  const int cols = SEG_W;
+  const int rows = SEG_H;
+
+  const uint8_t refresh_hz = map(SEGMENT.speed, 0, 255, 20, 80);
+  const unsigned refresh_ms = 1000 / refresh_hz;
+  const int16_t diffusion = map(SEGMENT.custom1, 0, 255, 0, 100);
+  const uint8_t spark_rate = SEGMENT.intensity;
+  const uint8_t turbulence = SEGMENT.custom2;
+
+  unsigned dataSize = SEGMENT.length() + (cols * sizeof(uint16_t));
+  if (!SEGENV.allocateData(dataSize))
+    return mode_static(); // allocation failed
+
+  if (SEGENV.call == 0)
+  {
+    SEGMENT.fill(BLACK);
+    SEGENV.step = 0;
+    call = 0;
+  }
+
+  if ((strip.now - SEGENV.step) >= refresh_ms)
+  {
+    uint16_t *tmp_row = (uint16_t *)(SEGMENT.data + SEGMENT.length());
+    SEGENV.step = strip.now;
+    call++;
+
+    // scroll up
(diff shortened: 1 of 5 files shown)


DedeHai:
thanks for contributing!
it really looks a lot nicer on a small matrix. 
There is a pending PR for my particle system which opefully will be finished soon and it looks even better, almost independent of matrix size and it will most likely replace the current fire FX. 
I will keep this PR open for now. You are welcome to check out the particle effects (if you do, use my dev branch)

mryndzionek:
Okay, I managed to compile and flash the particle system branch and I think that my fire effect is different enough to warrant a merge, especially when it's a usermod, so conditionally compiled.

DedeHai:
The reason I did not yet approve this is the following:
the compile tools (discord bot, online compilers) need to be maintained, every usermod added increases the work. This FX may not be used by many (it only really looks good up to about 12x12 pixels IMHO), dedicating a usermod to simple effects complicates things. 
So here is what we thought: for effects like yours it would be good to have a collection of "user FX" instead of having a usermod for each one. Since you are the first to commit such an effect: would you mind to rename the Usermod to "user FX" (I am not too sure about the name) where others can also contribute FX to and they can all be compiled in with a single UM addition? 
I was even thinking of making a second UM with "legacy FX" for effects that were replaced or dropped out but users still may like.

mryndzionek:
> The reason I did not yet approve this is the following: the compile tools (discord bot, online compilers) need to be maintained, every usermod added increases the work. This FX may not be used by many (it only really looks good up to about 12x12 pixels IMHO), dedicating a usermod to simple effects complicates things. So here is what we thought: for effects like yours it would be good to have a collection of "user FX" instead of having a usermod for each one. Since you are the first to commit such an effect: would you mind to rename the Usermod to "user FX" (I am not too sure about the name) where others can also contribute FX to and they can all be compiled in with a single UM addition? I was even thinking of making a second UM with "legacy FX" for effects that were replaced or dropped out but users still may like.

Okay, I renamed to `user_fx`.

DedeHai:
thanks!
can you add a readme file with a short description of what this UM is for (just one or two sentences that user FX go into this UM) and what your FX does.
Also: can you update the formating to match the WLED standard (I think close to google standard) with opening curly brackets on the same line to make code more compact.
also: 
- use fixed point math instead of floats, it helps a lot on all controllers without floats
- the use of random8() is deprecated, use hw_random8() instead

mryndzionek:
> thanks! can you add a readme file with a short description of what this UM is for (just one or two sentences that user FX go into this UM) and what your FX does. Also:
> 
>     * can you update the formating to match the WLED standard (I think close to google standard) with opening curly brackets on the same line to make code more compact.
> 
>     * use fixed point math instead of floats, it helps a lot on all controllers without FPU (speed & code size)
> 
>     * the use of random8() is deprecated, use hw_random8() instead

Done.

DedeHai:
@mryndzionek if you want this merged, please update.

mryndzionek:
> @mryndzionek if you want this merged, please update.

No, I think we can close this. The new particle system subsumes this.

blazoncek:
@mryndzionek I would suggest you modify this PR to add this into usermod_example as an example for other people how to write custom effects.
If you wish of course.

mryndzionek:
> add this into usermod_example as an example for other people

What do you mean by that exactly? Something more than resolving the conflicts?

mryndzionek:
> changes usermods/EXAMPLE_v2/usermod_v2_example.cpp

Where is changes `usermods/EXAMPLE_v2/usermod_v2_example.cpp`? Doesn't exist on `main`.

mryndzionek:
I pushed #4667. Compiles on `main`. Let me know there if it's okay.

blazoncek:
usermods/EXAMPLE/usermod_v2_example.cpp

Issue #3955, opened 2024-05-03
Leds flickers to full brightness when powering on


Andro-Marian:
### What happened?

With the Athom Digital Miocrophone controller when powering on the Master the leds flicker to Full Brightness (And sometimes has wrong color) then going to lowest brightness and starting to increase the brightness to the selected brightness.

https://github.com/Aircoookie/WLED/assets/156662313/97a2163b-a20d-48bb-b278-a5608e6d741c

### To Reproduce Bug

Set the leds to a lower brightness and toggle the Master power from WLed.

### Expected Behavior

The flicker should not occur. The color should be black when turning on and after like 0.5 seconds to start to transition, maybe.

### Install Method

Binary from WLED.me

### What version of WLED?

0.14.3

### Which microcontroller/board are you seeing the problem on?

ESP32

### Anything else?

I see than if I set the black color before turning it on and selecting a color after the transition is good without any flicker.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

blazoncek:
All supported LED ICs work as expected. Consider checking [KB](https://kno.wled.ge) for supported LED ICs.
Please use WLED [forum](https://wled.discourse.group) or Discord for help and support questions.
Closing.

Andro-Marian:
This is not fixed. And it's a bug' On 0.13 works better

DedeHai:
does this happen to you on WS2812 as well?

Andro-Marian:
Yes. Also without a shifter and on Breadboard with WS2812b

[wled_cfg_Marian-Desk-2.json](https://github.com/user-attachments/files/18362931/wled_cfg_Marian-Desk-2.json)

softhack007:
@Andro-Marian it looks like the "white flash" is happening in the moment when the relay powers the strip ?
In this case, it could be that the strip is not in sync with the led datastream at power-on, so the leds "misunderstand" the data and you see a glitch. Unfortunately we cannot do much in software in this case, however @DedeHai has just prepared a PR that might help:
* https://github.com/Aircoookie/WLED/pull/4474

please test.

DedeHai:
forgot to mention: we discussed this at length in discord. code is tested.

PR #4474, opened 2025-01-09
Fix for #3955: added a delay after switching relay

DedeHai:
helps to stabilize power on the LEDs before sending data

### DIFF
--- a/wled00/button.cpp
+++ b/wled00/button.cpp
@@ -375,6 +375,7 @@ void handleIO()
       if (rlyPin>=0) {
         pinMode(rlyPin, rlyOpenDrain ? OUTPUT_OPEN_DRAIN : OUTPUT);
         digitalWrite(rlyPin, rlyMde);
+        delay(50); // wait for relay to switch and power to stabilize
       }
       offMode = false;
     }


softhack007 (approved):
Looks OK to me.

@DedeHai did you check that the delay is only happening once when the relay actually gets switched? 

The relay handling code is very old (=hard to read) - a repeated `digitalWrite` of the same value would not be noticeable, however a repeated delay might hurt ...

DedeHai:
the code is only executed once when exiting `offmode`: at the end of the function, `offmode` is disabled.

softhack007:
👍

softhack007:
@DedeHai on a second look - do you think that the same delay should also be added at startup when initially controlling the relay? If yes, then maybe this location should also be changed:

https://github.com/Aircoookie/WLED/blob/bb0c0af1893e774f8401515083097caf0429c381/wled00/wled.cpp#L590-L593

WLED::beginStrip() runs once at startup, long before handleIO() gets called for the first time.


Just a random after-merge thought. Could be I'm just not involved at discord so I missed something important 🤷‍♂️ ...

DedeHai:
I thought about that too, but I have noticed recently, that there may be a bug in that part of the code and I wanted to take a closer look at: I did notice in a test, that the relay did not always switch to on depending on config settings and preset settings. 
in addition (unrelated): I would like to change the "orange color" behavior at bootup: if there are more than 1 segments, it is impossible to get anything else than orange on the higher numbered segments, which actually puzzles me that there was never a request to change that. or it may just be me: I always shut down my lights with a hard switch rather than putting them in standby, so this behaviour of orange at bootup is annoying.

PR #4476, opened 2025-01-10
Swap AnimARTrix to be platformio library

netmindz:
Change the usermod to be proper platformIO library. 

It would be nicer if we could just reference the usermod by name in the dependencies, but I can't see how to treat a local directory like a registry where you only include what you need. It seems you either just pull in every file or ignored.

Using the symlink approach at least means we can still selectively include what we need and take advantage of defining the dependencies of the usermod in the library.json file rather than listing dependencies in the readme to be added to lib_deps


netmindz:
Superseded by https://github.com/Aircoookie/WLED/pull/4480

Issue #4446, opened 2024-12-31
Enable transitions unexpected behivior 


dosipod:
### What happened?

Enable transitions and you will see leds speed up when changing brightness or other things , testing with DDP 
![Enable_transitions2](https://github.com/user-attachments/assets/7a196e08-6c92-49bd-9ac9-d75785f10221)

### To Reproduce Bug

Enable transitions 

### Expected Behavior

No sudden speed up 

### Install Method

Self-Compiled

### What version of WLED?

WLED 0.16.0-dev (build 2412040)

### Which microcontroller/board are you seeing the problem on?

ESP32

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

DedeHai:
That is by design, during transitions, FPS limits are disabled. Not that I think its a good design but there appears to be some reason behind it. Maybe it is outdated. 
@softhack007 is this behaviour still needed after you updated FPS limit handling?

softhack007:
> @softhack007 is this behaviour still needed after you updated FPS limit handling?

🤔 good point. Actually it's not necessary. I think the reason for this speedup was the "solid" effect - it normally runs at 3fps so colors were flickering during brightness or color change. It might be better to change the effect so it does `return FRAMTIME` during transitions.

blazoncek:
FYI any effect that is not returning FRAMETIME will exhibit speedup/slowdown. That is bu design of the particular effect and the fact that during transition (if you want it to look smooth) strip has to update frequently.

blazoncek:
ICU is another example.

softhack007:
FRAMETIME was removed in 16.0.0-beta, a slight speedup might still be visible in some effects when setting "target fps" to a low value - this is necessary to prevent "jumping" transitions (cross-fade).

blazoncek:
Since 0.16's segment blending speedup is no longer necessary and could be avoided.
However removing `return FRAMETIME;` removed that possibility.

PR #4477, opened 2025-01-10
bugfix: avoid effect overspeed during transitions (solves #4446)

softhack007:
this still raises FPS during transitions, but only for effects that normally run very slow (solid).

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1388,7 +1388,7 @@ void WS2812FX::service() {
         }
 #endif
         seg.call++;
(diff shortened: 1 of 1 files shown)


DedeHai:
I still don't know why these FX do not run at full speed... I could fix them but have other priorities ATM.
if you want a fixed frame rate without fixed frame-time: I have one possible implementation done in PS 2D Fire FX that also solves transition flickering: the FX renders the frame but does not update any parameters so the frame is identical to the last. maybe some of these FX could do the same or something similar?
I agree that this PR is more of a patch-up than a fix.

PR #4480, opened 2025-01-11
Convert usermods to static libraries

willmmiles:
Redesign the usermod intergration so that usermods are implemented as PlatformIO libraries instead of headers.  This permits them to call for dependencies, and eliminates the compiler flags for enabling each one, allowing the build cache to operate consistently regardless of the selected modules.

The usermod list is built using some linker magic to construct a static list in ROM memory.  This eliminates the need for wasting SRAM on something well known at build time.

To use this integration:
- Each usermod `.h` should be moved to a `.cpp`, with a static instantiation of the module class at the end
- The module object is then passed to `REGISTER_USERMOD()` to add it to the compiled-in usermod list
- Each usermod must add a 'library.json':
   - The library name must match the folder name.  (This PR includes special case handling for removing the 'usermod_v2_' prefix)
   - The library.json must call for local build integration ("includeDir": "../../wled00", "libLDFMode": "chain+", and "libArchive": false)
   - The library.json can now list any library dependencies, or custom build flags *that apply to only that module*.
- PlatformIO target environments now support a new 'custom_usermods' property to enable usermods, space delimited
  - Example: `custom_usermods = audioreactive animartrix` builds in the "audioreactive" and "animartrix" usermods

The `custom_usermods` property is implemented via a platformio hook script that constructs the required symlink lines and injects them in to the build environment.

Currently I have converted 'auto_save', 'audioreactive', and 'animartrix' (based on #4476) to library-type as proof-of-concept.


willmmiles:
Annoyingly, PlatformIO requires locally-defined properties to be prefixed with 'custom_' ; I'd've liked it to just be 'usermods' but no such luck. :(

platformio.ini
```diff
@@ -261,11 +257,11 @@ lib_deps =
...
   https://github.com/pbolduc/AsyncTCP.git @ 1.2.0
   ${env.lib_deps}
-# additional build flags for audioreactive
-AR_build_flags = -D USERMOD_AUDIOREACTIVE 
-  -D sqrt_internal=sqrtf ;; -fsingle-precision-constant ;; forces ArduinoFFT to use float math (2x faster)
-AR_lib_deps = kosme/arduinoFFT @ 2.0.1
 board_build.partitions = ${esp32.default_partitions}   ;; default partioning for 4MB Flash - can be overridden in build envs
+# additional build flags for audioreactive - must be applied globally
+AR_build_flags = -D sqrt_internal=sqrtf ;; -fsingle-precision-constant ;; forces ArduinoFFT to use float math (2x faster)
+AR_lib_deps = kosme/arduinoFFT @ 2.0.1  ;; for pre-usermod-library platformio_override compatibility
```
netmindz, line 263: If we are updating to be a library, why do we still need the old `AR_lib_deps` info as well?

usermods/audioreactive/library.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "audioreactive",
+  "build": {
+    "srcDir": ".",
+    "includeDir": "../../wled00",
+    "libLDFMode": "chain+",
+    "libArchive": false
+  },
```
netmindz, line 8: Can we automatically generate/inject this bit with python to save lots of duplication and further issues if we ever needed to tweak these at all, e.g different libLDFMode value

wled00/usermods_list.cpp
```diff
@@ -1,482 +0,0 @@
-#include "wled.h"
```
netmindz, line 1: Is it better to delete this file or retain but with message about the new format  - What would be the best way to highlight the changes to developers of the may forks?

netmindz (commented):
I'll have a bit of a play around with this over the next few days

It looks like a great start, the key things in my mind however are what extra changes we can make both in terms of code, documentation and release management to help make the transition as painless as possible

One quick question  - how essential is the rename from .h to .cpp ?

platformio.ini
willmmiles: Yes, this would be a breaking change that will need to be staged appropriately.  Unfortunately if your platformio file is outright invalid, PlatformIO in VSCode doesn't tell you what's wrong, it just blankly refuses to operate with no error message.  I found it more useful to at least keep the tool operating while I was correcting my build definitions, but that's just my $0.02.

willmmiles:
> It looks like a great start, the key things in my mind however are what extra changes we can make both in terms of code, documentation and release management to help make the transition as painless as possible

Yes.  Unfortunately it looks like the docs aren't in the code proper (beyond the examples), which is a little frustrating when different versions might have different approaches. :(

> One quick question - how essential is the rename from .h to .cpp ?

Some .cpp file has to instantiate the module object; and this must be in the library, or you're back to all the problems with `usermod_list.cpp`.  You could add a .cpp file to each that just does `#include <my_module_header.h>; static MyModule my_module; REGISTER_USERMOD(my_module);` , but personally I'd rather see all the code together in the one file rather than a seemling bureaucratic split.   My $0.02 is that the rename is the lesser evil -- git handles merges across renames pretty well these days.

usermods/audioreactive/library.json
willmmiles: I don't think so -- PlaformIO doesn't let the project level code manipulate properties of libraries.

wled00/usermods_list.cpp
willmmiles: Good question!  I'd figured the deletion will create an obvious merge conflict, so it's clear that something must be done.  Personally I'd rather not leave a .cpp file that the build tooling will have to grovel over only for it to do nothing.

netmindz:
> > One quick question - how essential is the rename from .h to .cpp ?
> 
> Some .cpp file has to instantiate the module object; and this must be in the library, or you're back to all the problems with `usermod_list.cpp`. You could add a .cpp file to each that just does `#include <my_module_header.h>; static MyModule my_module; REGISTER_USERMOD(my_module);` , but personally I'd rather see all the code together in the one file rather than a seemingly bureaucratic split. My $0.02 is that the rename is the lesser evil -- git handles merges across renames pretty well these days.

How about a migration script that looks for an existing .h of a usermod, moves with git and then automatically added the extra static field and call to register?

netmindz:
Are you aware of the build failure when using the audioreactive usermod? @willmmiles 

```
In file included from usermods/audioreactive/audio_reactive.cpp:2:0:
wled00/wled.h:95:26: fatal error: LITTLEFS.h: No such file or directory
```

willmmiles:
> > > One quick question - how essential is the rename from .h to .cpp ?
> > 
> > 
> > Some .cpp file has to instantiate the module object; and this must be in the library, or you're back to all the problems with `usermod_list.cpp`. You could add a .cpp file to each that just does `#include <my_module_header.h>; static MyModule my_module; REGISTER_USERMOD(my_module);` , but personally I'd rather see all the code together in the one file rather than a seemingly bureaucratic split. My $0.02 is that the rename is the lesser evil -- git handles merges across renames pretty well these days.
> 
> How about a migration script that looks for an existing .h of a usermod, moves with git and then automatically added the extra static field and call to register?

Not impossible. but nontrivial - the script would have to pick out the class name and any constructor arguments.  We'd probably need to call on `clang` or something else that understands C++ syntax to be able to do a reliable job of it.

platformio.ini
```diff
@@ -424,23 +420,23 @@ build_flags = ${common.build_flags} ${esp8266.build_flags} -D WLED_RELEASE_NAME=
 board = esp32dev
 platform = ${esp32.platform}
 platform_packages = ${esp32.platform_packages}
+custom_usermods = audioreactive
 build_unflags = ${common.build_unflags}
 build_flags = ${common.build_flags} ${esp32.build_flags} -D WLED_RELEASE_NAME=\"ESP32\" #-D WLED_DISABLE_BROWNOUT_DET
   ${esp32.AR_build_flags}
```
netmindz, line 426: Is there any way to move the build_flags used for the usermod into it's library.json?

willmmiles:
> Are you aware of the build failure when using the audioreactive usermod? @willmmiles
> 
> ```
> In file included from usermods/audioreactive/audio_reactive.cpp:2:0:
> wled00/wled.h:95:26: fatal error: LITTLEFS.h: No such file or directory
> ```

Yup, looking in to it - seems like some sort of case sensitivity thing.

platformio.ini
willmmiles: In most cases, yes.

In the specific case of AR, the build flags aren't actually for the AR module, they're for the FFT library dependency.  Unfortunately, the only way to affect some dependent library's build is to do it at the project level.  PlaformIO doesn't let one library define flags for one of its dependencies.   https://community.platformio.org/t/setting-flags-defines-for-building-a-librarys-dependency/36744

willmmiles: Hm, maybe there is a way, using platformio hook scripts in the library.  Working on it.

netmindz:
> > How about a migration script that looks for an existing .h of a usermod, moves with git and then automatically added the extra static field and call to register?
> 
> Not impossible. but nontrivial - the script would have to pick out the class name and any constructor arguments. We'd probably need to call on `clang` or something else that understands C++ syntax to be able to do a reliable job of it.

I'm just knocking up a quick PoC of a script. If nothing else it will help show how many of the usermods in the official AC repo do not follow proper naming convention etc

willmmiles:
> I'm just knocking up a quick PoC of a script. If nothing else it will help show how many of the usermods in the official AC repo do not follow proper naming convention etc

Yup, a breaking API change like this is a good opportunity for cleanup.  Honestly I'd figured on just going through them by hand, there aren't so many it's a huge problem.

With a solution like this in hand, I was hoping to ultimately create a "core" modules folder with the modules that this team is directly supporting; and then work towards migrating anything with a "WLED_DISABLE_X" flag in to becoming a core module instead.  (Yes, I recognize that will likely require some expansion of the module API to handle specialized timing and polling requirements; I think that's probably a good thing, as if we have one use case for it, there may be others.)

Further, the linker section approach demonstrated here for the static usermod list could also be used later to make static library-based FX or bus type lists, too, allowing modules to include new FX or bus types without needing any RAM to build a runtime data structure.  One step at a time, though!

platformio.ini
netmindz: There are certainly usermods that use more typical build_flags that affect the usermod itself, but perhaps these still need to stay in the main platformio.ini if they are actual optional flags the user chooses to set

netmindz:
I didn't want to mess up your main PR with my attempts at automatic migration, so I've popped into it's own PR for now https://github.com/Aircoookie/WLED/pull/4481

These are the ones that don't follow any common naming 

WARNING: usermods/ADS1115_v2/ADS1115_v2.h missing
WARNING: usermods/AHT10_v2/AHT10_v2.h missing
WARNING: Artemis_reciever possibly still v1 usermod
WARNING: usermods/BH1750_v2/BH1750_v2.h missing
WARNING: usermods/BME280_v2/BME280_v2.h missing
WARNING: usermods/BME68X_v2/BME68X_v2.h missing
WARNING: usermods/EXAMPLE_v2/EXAMPLE_v2.h missing
WARNING: Enclosure_with_OLED_temp_ESP07 possibly still v1 usermod
WARNING: usermods/Fix_unreachable_netservices_v2/Fix_unreachable_netservices_v2.h missing
WARNING: usermods/INA226_v2/INA226_v2.h missing
WARNING: usermods/Internal_Temperature_v2/Internal_Temperature_v2.h missing
WARNING: usermods/JSON_IR_remote/JSON_IR_remote.h missing
WARNING: usermods/LD2410_v2/LD2410_v2.h missing
WARNING: usermods/MAX17048_v2/MAX17048_v2.h missing
WARNING: usermods/MY9291/MY9291.h missing
WARNING: RelayBlinds possibly still v1 usermod
WARNING: TTGO-T-Display possibly still v1 usermod
WARNING: usermods/TetrisAI_v2/TetrisAI_v2.h missing
WARNING: Wemos_D1_mini+Wemos32_mini_shield possibly still v1 usermod
WARNING: usermods/audioreactive/audioreactive.h missing
WARNING: usermods/battery_keypad_controller/battery_keypad_controller.h missing
WARNING: usermods/mqtt_switch_v2/mqtt_switch_v2.h missing
WARNING: photoresistor_sensor_mqtt_v1 possibly still v1 usermod
WARNING: usermods/project_cars_shiftlight/project_cars_shiftlight.h missing
WARNING: usermods/rotary_encoder_change_effect/rotary_encoder_change_effect.h missing
WARNING: usermods/sensors_to_mqtt/sensors_to_mqtt.h missing
WARNING: usermods/seven_segment_display_reloaded/seven_segment_display_reloaded.h missing
WARNING: usermods/stairway_wipe_basic/stairway_wipe_basic.h missing
WARNING: usermods/word-clock-matrix/word-clock-matrix.h missing

netmindz:
Fixed the naming of a few, so now just

WARNING: Artemis_reciever possibly still v1 usermod
WARNING: usermods/EXAMPLE_v2/EXAMPLE_v2.h missing
WARNING: Enclosure_with_OLED_temp_ESP07 possibly still v1 usermod
WARNING: usermods/Fix_unreachable_netservices_v2/Fix_unreachable_netservices_v2.h missing
WARNING: usermods/JSON_IR_remote/JSON_IR_remote.h missing
WARNING: RelayBlinds possibly still v1 usermod
WARNING: TTGO-T-Display possibly still v1 usermod
WARNING: Wemos_D1_mini+Wemos32_mini_shield possibly still v1 usermod
WARNING: usermods/battery_keypad_controller/battery_keypad_controller.h missing
WARNING: photoresistor_sensor_mqtt_v1 possibly still v1 usermod
WARNING: usermods/project_cars_shiftlight/project_cars_shiftlight.h missing
WARNING: usermods/rotary_encoder_change_effect/rotary_encoder_change_effect.h missing

netmindz:
> > Are you aware of the build failure when using the audioreactive usermod? @willmmiles
> > ```
> > In file included from usermods/audioreactive/audio_reactive.cpp:2:0:
> > wled00/wled.h:95:26: fatal error: LITTLEFS.h: No such file or directory
> > ```
> 
> Yup, looking in to it - seems like some sort of case sensitivity thing.

I think it might be because we are missing the `-D LOROL_LITTLEFS` that we normally apply to esp32 builds

netmindz:
> > > Are you aware of the build failure when using the audioreactive usermod? @willmmiles
> > > ```
> > > In file included from usermods/audioreactive/audio_reactive.cpp:2:0:
> > > wled00/wled.h:95:26: fatal error: LITTLEFS.h: No such file or directory
> > > ```
> > 
> > 
> > Yup, looking in to it - seems like some sort of case sensitivity thing.
> 
> I think it might be because we are missing the `-D LOROL_LITTLEFS` that we normally apply to esp32 builds

Confirmed. If I hack wled.h to define LOROL_LITTLEFS the problem goes away.  Do you have a wled_custom.h or something locally?

willmmiles:
> > > > Are you aware of the build failure when using the audioreactive usermod? @willmmiles
> > > > ```
> > > > In file included from usermods/audioreactive/audio_reactive.cpp:2:0:
> > > > wled00/wled.h:95:26: fatal error: LITTLEFS.h: No such file or directory
> > > > ```
> > > 
> > > 
> > > Yup, looking in to it - seems like some sort of case sensitivity thing.
> > 
> > 
> > I think it might be because we are missing the `-D LOROL_LITTLEFS` that we normally apply to esp32 builds
> 
> Confirmed. If I hack wled.h to define LOROL_LITTLEFS the problem goes away. Do you have a wled_custom.h or something locally?

No, I'm just building on a case-insensitive filesystem ;)

The problem is actually rooted in the PlatformIO dependency manager; it's getting confused about which library is providing LittleFS to the usermod library source file, and consequently it's not putting the correct include paths on the compile line. 
 It is supplying the correct flags (which is why it tries to get the all-caps header at all).  PlatformIO's dependency manager doesn't handle conditional #include very well, especially when the trigger for the include comes from `build_flags`; that's why putting it in the header explicitly resolves the problem.  It's also causing some kind of slowdown problem on my system, where every invocation of build takes a *long* time in "Scanning dependencies...".

I'll find a way to get it sorted out in the `load_usermods.py` script.

platformio.ini
willmmiles: Fixed! It turns out this is a way though platformio hook scripts.  8527d231

willmmiles:
> > > > > Are you aware of the build failure when using the audioreactive usermod? @willmmiles
> > > > > ```
> > > > > In file included from usermods/audioreactive/audio_reactive.cpp:2:0:
> > > > > wled00/wled.h:95:26: fatal error: LITTLEFS.h: No such file or directory
> > > > > ```
> > > > 
> > > > 
> > > > Yup, looking in to it - seems like some sort of case sensitivity thing.
> > > 
> > > 
> > > I think it might be because we are missing the `-D LOROL_LITTLEFS` that we normally apply to esp32 builds
> > 
> > 
> > Confirmed. If I hack wled.h to define LOROL_LITTLEFS the problem goes away. Do you have a wled_custom.h or something locally?
> 
> No, I'm just building on a case-insensitive filesystem ;)
> 
> The problem is actually rooted in the PlatformIO dependency manager; it's getting confused about which library is providing LittleFS to the usermod library source file, and consequently it's not putting the correct include paths on the compile line. It is supplying the correct flags (which is why it tries to get the all-caps header at all). PlatformIO's dependency manager doesn't handle conditional #include very well, especially when the trigger for the include comes from `build_flags`; that's why putting it in the header explicitly resolves the problem. It's also causing some kind of slowdown problem on my system, where every invocation of build takes a _long_ time in "Scanning dependencies...".
> 
> I'll find a way to get it sorted out in the `load_usermods.py` script.

...and done.  The new logic gets rid of the need for all the boilerplate in `library.json` too - all that's required now is a name.

blazoncek:
> but personally I'd rather see all the code together in the one file rather than a seemingly bureaucratic split

That will inevitably erase all commit history of such usermod. Hence you'll be unable to find original author if that matters.

Ona side not, several usermods are ripe for pruning.

willmmiles:
> > but personally I'd rather see all the code together in the one file rather than a seemingly bureaucratic split
> 
> That will inevitably erase all commit history of such usermod. Hence you'll be unable to find original author if that matters.

 git is not the VCSes of our youth.  `git blame` does the right thing, even across file moves.

blazoncek:
I've chewed through this and all I can say is - *awesome*.
Fully approved.

I would still remove several usermods, especially those that no longer compile and have not received complaints from endusers.

netmindz:
Yeah I'd certainly agree we delete any that are still v1

willmmiles:
> For some reason I can't set the PR to target back your branch @willmmiles , it doesn't show in the list for some reason, but do please edit the PR to point to your branch if github will let you

Probably because the PR is on the Aircoookie repo, while the `usermod-libs` branch is in my repo.  I'll merge your changes manually - thanks for putting that together!

willmmiles:
> ...and done. The new logic gets rid of the need for all the boilerplate in library.json too - all that's required now is a name.

Unfortunately it turns out that `"libArchive": false` is required for the static list builder to work.  I'm looking in to more PlatformIO scripting to solve it centrally rather than put it in every `library.json`, but unfortunately it looks like it's going to require hooking PlaformIO fairly deeply as there's no easy way to get in between "before library dependencies are parsed" and "after target actions are already planned".

Basically the alternatives seem to be:
- A: Monkey-patch PlatformIO to inject a library property centrally, after parsing but before action generation.  Doable but potentially fragile.
- B: Add a boilerplate `"libArchive": false` to all usermod library.json files
- C: Set "lib_archive: false" globally.  Risks including otherwise unneeded code from other libraries, bloating code size.  (I haven't tested this yet to get any metrics, though.)
- D: Have the hook script dynamically generate `library.json` for usermods in some temp folder, and symlink there instead.  Less invasive but adds a bunch of extra temp folder management.
- E: Instead of making usermods in to PlatformIO libraries via `library.json`, we could do our own thing: have the hook script parse a `module.json` instead and add the relevant folders, flags, scripts, dependencies, etc. to the base project construction environment.  Adds the most documentation and we'd have to roll the property list ourselves; offers a lot of control and compatibility at the cost of the most up front work.

I'm experimenting with A, and failing that, I'm leaning towards B as the alternative -- I'd prefer to keep to PlaformIO library.json format if possible, if only to save us from having to do our own documentation.  I could be convinced by good arguments for other solutions, though.

netmindz:
Try A but don't spent too much time on it. Option B is still way better than what we have currently

netmindz:
Any luck with the LOROL_LITTLEFS fix?

willmmiles:
> Any luck with the LOROL_LITTLEFS fix?

Yes, it's already integrated here (see the updated Action results!).  Pull updates from this branch, but keep `"libArchive": false` in the library.json files.

netmindz:
I have added `libArchive: false` back in to my branch

willmmiles:
> > ...and done. The new logic gets rid of the need for all the boilerplate in `library.json` too - all that's required now is a name.
> 
> I'm not sure it's working as you planned, usermods that were building for me now error with `AsyncTCP.h: No such file or directory`

Just pushed an initial implementation of option A -- but yes, I think the include path assembly still isn't quite right.  I'll pin that down next.

blazoncek:
> Any luck with the LOROL_LITTLEFS fix?

When you switch platform from 3.5.0 to 6.3.x, Lorol's LittleFS will not be needed any more.

pio-scripts/load_usermods.py
```diff
@@ -2,6 +2,11 @@
 from pathlib import Path   # For OS-agnostic path manipulation
 
 usermod_dir = Path(env["PROJECT_DIR"]) / "usermods"
+all_usermods = [f for f in usermod_dir.iterdir() if f.is_dir() and f.joinpath('library.json').exists()]
+
+if env['PIOENV'] == "usermods":
+   # Add all usermods
+   env.GetProjectConfig().set(f"env:usermods", 'custom_usermods', " ".join([f.name for f in all_usermods]))
```
netmindz, line 9: The downside to this change is that you have no control of what you are building

For testing I was setting my local platform_override.ini to only list the mods I was trying to build.

I guess I can still do that, but just need to name my env somehting like `usermods_local`

netmindz:
Any ideas why I'm now seeing errors for missing `wled.h` ?

netmindz:
> Any ideas why I'm now seeing errors for missing `wled.h` ?

Looks like the load_usermods.py doesn't have any reference to wled00, unlike the original libray.json. 

No idea if this is correct, but it makes the error for away @willmmiles 

```
diff --git a/pio-scripts/load_usermods.py b/pio-scripts/load_usermods.py
index 76359a7a..862f30e4 100644
--- a/pio-scripts/load_usermods.py
+++ b/pio-scripts/load_usermods.py
@@ -38,6 +39,7 @@ if usermods:
 cpl = env.ConfigureProjectLibBuilder
 # Our new wrapper
 def cpl_wrapper(env):
+  src_dir = proj.get("platformio", "src_dir") + "/wled00/"
   # Update usermod properties
   lib_builders = env.GetLibBuilders()  
   um_deps = [dep for dep in lib_builders if usermod_dir in Path(dep.src_dir).parents]
@@ -46,7 +48,9 @@ def cpl_wrapper(env):
     # Add include paths for all non-usermod dependencies
     for dep in other_deps:
         for dir in dep.get_include_dirs():
+            print(f"um = {um} dir = {dir}\n")
             um.env.PrependUnique(CPPPATH=dir)
+            um.env.PrependUnique(CPPPATH=src_dir)
     # Add the wled folder to the include path    

```

netmindz:
> > Any ideas why I'm now seeing errors for missing `wled.h` ?
> 
> Looks like the load_usermods.py doesn't have any reference to wled00, unlike the original libray.json.
> 
> No idea if this is correct, but it makes the error for away @willmmiles
> 
> ```
> diff --git a/pio-scripts/load_usermods.py b/pio-scripts/load_usermods.py
> index 76359a7a..862f30e4 100644
> --- a/pio-scripts/load_usermods.py
> +++ b/pio-scripts/load_usermods.py
> @@ -38,6 +39,7 @@ if usermods:
>  cpl = env.ConfigureProjectLibBuilder
>  # Our new wrapper
>  def cpl_wrapper(env):
> +  src_dir = proj.get("platformio", "src_dir") + "/wled00/"
>    # Update usermod properties
>    lib_builders = env.GetLibBuilders()  
>    um_deps = [dep for dep in lib_builders if usermod_dir in Path(dep.src_dir).parents]
> @@ -46,7 +48,9 @@ def cpl_wrapper(env):
>      # Add include paths for all non-usermod dependencies
>      for dep in other_deps:
>          for dir in dep.get_include_dirs():
> +            print(f"um = {um} dir = {dir}\n")
>              um.env.PrependUnique(CPPPATH=dir)
> +            um.env.PrependUnique(CPPPATH=src_dir)
>      # Add the wled folder to the include path    
> ```

Nope, I'm talking  rubbish. Every time you do a build after a full clean you get the missing wled.h, however just clicking build again fixes it

willmmiles:
> Nope, I'm talking rubbish. Every time you do a build after a full clean you get the missing wled.h, however just clicking build again fixes it

Yeah, there's something wrong with building the dependency list; it's sometimes not locating the usermods early enough, so the patch-up process isn't able to configure them.  I can't replicate it on my system, even in a fresh container, but the CI does it repeatably, so [I was debugging in a separate branch](https://github.com/willmmiles/WLED/tree/usermod-libs-wip).  I'll update when I've got it sorted out.

netmindz:
> > Nope, I'm talking rubbish. Every time you do a build after a full clean you get the missing wled.h, however just clicking build again fixes it
> 
> Yeah, there's something wrong with building the dependency list; it's sometimes not locating the usermods early enough, so the patch-up process isn't able to configure them. I can't replicate it on my system, even in a fresh container, but the CI does it repeatably, so [I was debugging in a separate branch](https://github.com/willmmiles/WLED/tree/usermod-libs-wip). I'll update when I've got it sorted out.

What OS and platformIO version are you on? I'm using linux and 6.1.4 and is is totally repeatable. Full clean, Build - error no wled.h, Build - OK

willmmiles:
> 
> What OS and platformIO version are you on? I'm using linux and 6.1.4 and is is totally repeatable. Full clean, Build - error no wled.h, Build - OK

Hm, maybe that's it and it was changed in a newer PlatformIO -- library management is definitely an area of active development.  I'm using whatever came down with the devcontainer; currently it's PlatformIO 6.1.17b2.

Once we've got a working solution, I'm going to see about opening a discussion on PlatformIO's own issue board about our "modules" use case and see what the devs there think about it.  [`lib_extra_dirs`](https://docs.platformio.org/en/latest/projectconf/sections/env/options/library/lib_extra_dirs.html) would make 85% of this unnecessary; all we'd need is the `"libArchive": false` magic.  With any luck they'll be suitably horrified by what we've had to do to work around that deprecation that they'll give some more thought on how best to handle this sort of thing.

pio-scripts/load_usermods.py
netmindz: Possibly less need for an "all in one" build if we just let the CI handle a matrix build for usermods, then just define the one we are working on locally in our own platform_override ?

https://github.com/willmmiles/WLED/pull/3

willmmiles: I stand by my opinion from above -- I think it's important that what the CI system does should be easily and directly reproducible by developers.  I don't think it's appropriate to have to hand hack custom files in your personal workspace to replicate what the CI does; that's just asking to get it wrong.

From a practical perspective, I'd also like it to be simple to validate my local changes before pushing to the CI.  If we depend on matrix builds, that becomes almost impossible.  Id be stuck waiting on the CI system any time I want to have some confidence I haven't inadvertently broken the build of some random usermod.

Finally, I don't know much about GitHub CI; I'm concerned that adding dozens of extra builds per run would be costly.

netmindz:
Good to see you are still chipping away at this tricky problem. We will definitely be in a much better situation once you complete the work. 

With regards to your question regarding inter module communication, it is true that they are are some issues with that design of UM data, which we may choose to address at some point, but as it's the best we have currently, then all usermods should comply with the interface and use UM data rather than any other unsupported mechanism

willmmiles:
> With regards to your question regarding inter module communication, it is true that they are are some issues with that design of UM data, which we may choose to address at some point, but as it's the best we have currently, then all usermods should comply with the interface and use UM data rather than any other unsupported mechanism

It wasn't so much a qustion as an observation of the current state of affairs.  If mod developers are finding the existing UM data approach too complex or unworkable for their use cases, I'm not going to judge for finding a local escape vs attempting an ecosystem-wide API upgrade.  (Which, if we want to take on, we should do soon-ish -- since this PR is already a breaking change, might as well get 'em all done at once!)

I consider such changes out of scope for this PR, though.  Better to support all the existing code as cleanly as possible, and update the APIs as something that can be reviewed separately.

blazoncek:
@willmmiles & @netmindz for usermods that interact (i.e. call each other) it would be best to include simple API to exchange a few variables or provide events. Perhaps best using JSON as it is flexible.
So far I am aware of interactions between:
- Rotary encoder and Four Line Display
- Auto-save and Four Line Display
- PWM_fan & Temperature or SHT
- Seven segment clock and SN_Photoresistor

No other usermods interact.

As (almost) all usermods employ `readFromJsonState()` and/or `addToJsonState()` which are also intended for sharing internal data they may be used for such purpose.
`UM_Exchange_Data` structure was invented to overcome cumbersome processing of JSON format used in `JsonState()` methods in effect function.

blazoncek:
One more thing regarding `UM_Exchange_Data`.
It was invented solely for Audioreactive usermod before I implemented effect array as a vector (dynamically added effects). If audio effects are moved from `FX.cpp` into Audioreactive sources then there is no need for it.

willmmiles:
> for usermods that interact (i.e. call each other) it would be best to include simple API to exchange a few variables or provide events.

I agree.  It also seems like there's a common pattern in "sensors to MQTT" type usermods too.  I think putting in a simpler framework for publishing and subscribing to sensor data might be a good plan, and maybe even some common code for routing things from the data API to MQTT.

..but *ultimately* where I'd like to go is a general facility for routing sensor data (and yes, even audio data) to FX parameters, maybe even with a little formula parser to set min/max/scale/combine/etc.  This is still a half-baked idea and would need a lot of fleshing out to handle corner cases, but I want to keep it in mind as a long term goal.

I don't think we want to round-trip through JSON for all sensor data, though.  That's enough of a hassle (and CPU cost) that I'd expect people to find shortcuts to avoid it, just like you did for the audio data.

> No other usermods interact.

I did put in the "gyro surge" partner for MPU6050; it basically routes the net angular rate to the global brightness.  They use the um_data API to communicate, though.

netmindz:
@coderabbitai review

willmmiles:
@coderabbitai review

willmmiles:
Looks like the last CI issue was a [platformio bug fixed in the latest release](https://github.com/platformio/platformio-core/pull/4941).  So here it is: usermods with real dependency lists, local build options, static instantiation, and the usermod list is assembled at build time.  Plus a "lamp test" CI build that validates that most of the bundled usermods at least compile.

blazoncek:
Awesome! 👏

netmindz:
Just noticed we still have a few usermods disabled

willmmiles:
> Just noticed we still have a few usermods disabled

Sorry I missed some!  IIRC there were some that required complex configuration (LCDs?) or just seemed to depend on broken upstreams.  There was at least one that exploded the RAM size with fonts or something that wouldn't fit; I'm sure there was something missing.

netmindz:
No worries, perhaps open a new issue so we can review those to see if we can fix or if perhaps we need to contact the author to fix

PR #4482, opened 2025-01-12
also sync segment name through udp

Liliputech:
This commit solves this issue : https://github.com/Aircoookie/WLED/issues/4468

## Summary by CodeRabbit

- **New Features**
  - Enhanced configuration options now support segment names.
  - The synchronization settings interface has an additional checkbox to enable or disable segment name reception.
  - UDP notifications are updated to include segment name details, improving segment management.
  - The system’s settings now offer robust handling of segment names for a more flexible and informative configuration experience.
  - New parameter added for segment name handling in synchronization settings.


netmindz:
Should this be an optional sync attribute, like the existing ones are?

Liliputech:
as of now its not optional but I can make it optional if needed.
I was not sure in which category to add this (effects, segment bounds, options?), or if I should create a new category.
The name sometime is used by the effect for the effect configuration (see GIF image branch).

blazoncek:
This is a breaking change. Old devices may crash or behave incorrectly due to message containing unexpected content.
Please rework so older devices will behave correctly.

Liliputech:
This is to be tested but since the "segment packet size" is sent in the UDP packet and read by the receiver it may be backward compatible.

netmindz:
I'd suggest a new checkbox called Segment Name

netmindz:
What testing have you done so far with regards to @blazoncek 's comments about compatibility testing between systems using your new packet format and without?

Liliputech:
Unfortunately didn't had the opportunity to run such tests yet. I will do ASAP
My idea of how to test this : flash one board with my branch and another board with main branch without packet format change.
then enable sync on both boards and see if everything is fine (color/segment options change).
Also see change the segment name on the board with my branch, trigger an update and see if the other board don't crash.

Liliputech:
Unfortunately didn't had the opportunity to run such tests yet. I will do ASAP
My idea of how to test this : flash one board with my branch and another board with main branch without packet format change.
then enable sync on both boards and see if everything is fine (color/segment options change).
Also try to change the segment name on the board with my branch, trigger an update and see if the other board don't crash.

Liliputech:
just push a new commit to fix the issue previously mentioned + rebased on latest main

netmindz:
Do not commit the html .h files as those are built on demand

netmindz:
Please do not use force push

Liliputech:
arf, too late. but commits didn't changed, only rebased on main

netmindz:
If you force push it means we have to start the code review process again. Are you able to take a look again @blazoncek ?

wled00/udp.cpp
```diff
@@ -4,17 +4,17 @@
...
+#define WLEDPACKETSIZE (SEG_OFFSET+(MAX_NUM_SEGMENTS*UDP_SEG_SIZE)+0)
 #define UDP_IN_MAXSIZE 1472
 #define PRESUMED_NETWORK_DELAY 3 //how many ms could it take on avg to reach the receiver? This will be added to transmitted times
 
 typedef struct PartialEspNowPacket {
   uint8_t magic;
   uint8_t packet;
   uint8_t noOfPackets;
-  uint8_t data[247];
+  uint8_t data[277+WLED_MAX_SEGNAME_LEN];
```
blazoncek, line 17: ESP-NOW packet size is limited to 250 bytes. It cannot be larger.

wled00/udp.cpp
```diff
@@ -338,6 +344,21 @@ static void parseNotifyPacket(const uint8_t *udpIn) {
           selseg.check1  = (udpIn[31+ofs]>>7) & 0x1;
         }
       }
+      if (receiveSegmentName) {
+	if (selseg.name) { //clear old name
+	  delete[] selseg.name;
```
blazoncek, line 349: In #4484 I implemented method for setting Segment name. I would recommend to use that once #4484 is merged as it moves memory management into Segment class (where it should be).

wled00/wled.h
```diff
@@ -647,18 +647,20 @@ typedef class Receive {
         bool    Color          : 1;
         bool    Effects        : 1;
         bool    SegmentOptions : 1;
+        bool    SegmentName    : 1;
```
blazoncek, line 650: That's why there is `reserved`. Use that if possible.

wled00/data/settings_sync.htm
```diff
@@ -101,7 +101,7 @@ <h3>Sync groups</h3>
 </table>
 <h3>Receive</h3>
 <nowrap><input type="checkbox" name="RB">Brightness,</nowrap> <nowrap><input type="checkbox" name="RC">Color,</nowrap> <nowrap><input type="checkbox" name="RX">Effects,</nowrap> <nowrap>and <input type="checkbox" name="RP">Palette</nowrap><br>
-<input type="checkbox" name="SO"> Segment options, <input type="checkbox" name="SG"> bounds
+<input type="checkbox" name="SO"> Segment options, <input type="checkbox" name="SN"> Segment name, <input type="checkbox" name="SG"> bounds
```
blazoncek, line 104: It could be "[x] Segment options, [x] name, [x] bounds"

wled00/udp.cpp
Liliputech: ok, i definitely agree on that. I will rebase my branch on top of yours to use the functions you defined

wled00/wled.h
Liliputech: I'm sorry but i don't understand how I'm supposed to use reserved... to me it looks like it is a variable to hold the sum of all boolean define above, which is then not to be changed by hand?

wled00/udp.cpp
Liliputech: There is a little issue in your PR, in the setName function the strlcpy is missing 1 char, resulting in truncated name.
I've fixed that locally and everything works fine.

netmindz:
Looks like the wrong branch was merged into this branch 

`Merge remote-tracking branch 'blazoncek/parallel-I2S' into udp_name_sync`

blazoncek:
@netmindz it will be ok once #4484 is merged.

Liliputech:
hi, just solved conflict after merge of #4484 :) hope this one can be merged!

wled00/udp.cpp
```diff
@@ -143,6 +143,11 @@ void notify(byte callMode, bool followUp)
     udpOut[33+ofs] = selseg.startY & 0xFF;
     udpOut[34+ofs] = selseg.stopY >> 8;     // ATM always 0 as Segment::stopY is 8-bit
     udpOut[35+ofs] = selseg.stopY & 0xFF;
+
+    if(selseg.name){
+	strcpy((char *)&udpOut[36+ofs], selseg.name);
+    }
```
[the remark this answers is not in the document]
blazoncek: This is one of the reasons why I am against syncing segment names across devices.
ESP8266 has a different maximum name length than ESP32. So the UX will not be the best.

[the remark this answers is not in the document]
Liliputech: this was actually handled in the "setName" function but good thing to double check

wled00/wled.h
blazoncek: You should have not "resolved" this as you did not address my comment.
`reserved`is a placeholder for future enhancement as this PR so replace it with `SegmentName`option.

wled00/udp.cpp
```diff
@@ -155,15 +160,11 @@ void notify(byte callMode, bool followUp)
...
-    size_t s0 = 0;
+    size_t packetSize = SEG_OFFSET;
     memcpy(buffer.data, udpOut, packetSize);
-    // stuff as many segments in first packet as possible (normally up to 5)
-    for (size_t i = 0; packetSize < bufferSize && i < s; i++) {
-      memcpy(buffer.data + packetSize, &udpOut[41+i*UDP_SEG_SIZE], UDP_SEG_SIZE);
-      packetSize += UDP_SEG_SIZE;
-      s0++;
-    }
+    // since we sync the name, only one segment can fit in the first packet
```
blazoncek, line 165: This is a waste of packet space. Do not forget that the same packet is re-used in ESP-NOW sync.
ESP-NOW packet can only contain 250 bytes so "compression" of data is a must.

Liliputech:
> I know there is a desire to sync segment names across devices (though I can't find justification) but this implementation is "wasteful" in terms of needed bytes to transmit a name (i.e. if I name segment "A" it will still transmit 32/64 bytes for name instead of only 2 - 'A' and '\0').

Well, it is much easier to copy a fixed length string, if you start to send data with unknown number of bytes, then I don't really see how to make it backward compatible given the "UDP_SEG_SIZE" would then be different from one segment to another. That would require a bit of redesign.

Liliputech:
I've turned the problem in many different ways,
I can't find a solution for both dynamic segment data size (to optimize on-air time and packet usage) and retrocompatibility.
in udp.cpp L272 the segment data size is a constant set in udp[40], and which defines an offset.
We could embed the segment data size as the first byte of the segment data and have this byte increment the offset, but then the parseNotify function wouldn't be compatible anymore across WLED version.

A workaround that I see is to define another usermod hook like the new one "onEspNowMessage" but for Udp. See udp.cpp L952.
Then I'd be happy to write a usermod to use this new hook to update the segment name, and also use "onStateChange" hook to send a custom "nameUpdate" specific message.
That way I get the functionality that I need (several effects needs to define segment name actually).
That would also permit more actions in usermod which is not a bad thing IMHO.

Another solution that I tried is to use DDP and one wled as the master and the other one as the slave but it takes a lot more bandwith and with very fast moving pixels it generates a lot of flickering which I want to avoid.

netmindz:
Possibly need for a new type of sync packet in addition to existing, especially given the size issue of ESP-NOW

Liliputech:
I've opened a new pull request to solve this problem with a different approach : #4859
Hope you'll like it better :)
I'm closing this one, but keeping the branch so we can make some comparisons.

Issue #4380, opened 2024-12-11
SK6812 RGBW colors broken in 0.15.0 on ESP32, not ESP8266, works in 0.14.4


trullock:
### What happened?

I have two strips of SK6812s RGBWs, one lot attached to an ESP32 and one to an ESP8266

Updating to 0.15 from 0.14.4 breaks the ESP32 colors but not the ESP8266 colors

I've reverted to 0.14.4 so cant debug further at this moment but can do later

### To Reproduce Bug

SK6812s RGBWs in GBR mode on an ESP32, the colors are the wrong. They are very dim, the hue isnt far off or might be right (havent fully debugged, had to quickly revert before the kids came home to see no xmas lights!)

### Expected Behavior

Colors to be correct ;)

### Install Method

Binary from WLED.me

### What version of WLED?

0.15.0

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell
can probive later, got to
```

### Anything else?

Can debug further later, though Id get this raised for everyones sake

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

willmmiles:
>  very dim

Try checking the brightness limiter settings.  Also, if you can, please post the config from your 0.14 system; it sounds like there's either (a) a bug in the upgrade path, or (b) a setting that was previously ignored is now being correctly applied; I'd like to poke the config and see if I can work out what it is.

listlesslife:
I was having a similar problem, but my esp8266s weren't changing colors and my esp32s were fine. Rolled back the 8266s to 0.14.4 and now color changing works. 0.15.0 also knocked a couple of my esp8266s offline and they had to be reflashed via serial, but that's a different issue.

trullock:
So, automatic brightness limiting had been turned on, turning that off solves it.

However the effects wrt segments is screwy, like its sped up

I have a vid but cant post it here. Can DM/email

nomonkeynodeal:
It looks like the blue channel just doesn't work for me with 15.0. I rolled back to 14.4 and it's fine.

trullock:
0.15 now seems completely broken for me, and I can't roll back to 0.14.4. I provide the binary and seems to apply the update, but it doesnt. 

Will have to get the ladders out and flash it directly I guess

c-b5r:
I updated (OTA) 5x ESP8266 (d1 mini) with SK6812 setup. Here is the result:

- 2x updated just fine and everything worked as before
[...]

PR #4484, opened 2025-01-12
WWA strip support & parallel I2S for S2/S3 (bumping outputs from 5/4 to 12)

### LINKED ISSUE #552: Support WWA channel type for SK6812-WWA type strips
Currently there is no support for WWA (white, white, amber) style strips. I plan on adding this and doing a PR, waiting for strips to test with first. This issue is just a placeholder.

Example strip: https://www.aliexpress.com/item/32430149013.html

blazoncek:
Bus wrapper modifications
- NeoPixelBus update 2.8.3
- automatic selection of appropriate single I2S bus (`X1xxxxxxMethod`)
- removed I2S0 on ESP32 (used by AudioReactive)
- renumbered internal bus numbers (iType)
- added buffer size reporting (prototype, requested @makuna to add appropriate methods to NPB)

Bus modifications
- WWA strip support
- bus initialisation rewrite (moved to `WS2812FX::finalizeInit()`)
- optional parallel I2S (ESP32, S2 & S3)
- ABL speed improvements (no float)

This modification will allow unrestricted use of AudioReactive usermod on ESP32 and ESP32-S3. S2 is still prone to conflict but since it is not best suited for large LED counts in combination with AudioReactive the compromise may be acceptable.

WWA strip support may need WW/CW channel swap as I do not have one for test. A(mber) channel is not supported.

Fixes #552


DedeHai:
I tried this PR on an S2 lolin: cannot get it to save correctly LED settings if more than 5 outputs are set. It saves but config gets screwed up a little, next time in LED settings, all outputs are gone except the first and it is reset (to a backup I assume)

blazoncek:
Hmm, that may be due to change in initialisation procedure. Will investigate.
In the meantime, if you can manually update `cfg.json` that should work.

blazoncek:
Latest update will also fix #4380 until resolved in NeoPixelBus.

wled00/cfg.cpp
```diff
@@ -220,24 +196,11 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
...
-        } else
-          mem += BusManager::memUsage(bc); // includes global buffer
-        if (mem <= MAX_LED_MEMORY) if (BusManager::add(bc) == -1) break;  // finalization will be done in WLED::beginStrip()
-      } else {
-        if (busConfigs[s] != nullptr) delete busConfigs[s];
-        busConfigs[s] = new BusConfig(ledType, pins, start, length, colorOrder, reversed, skipFirst, AWmode, freqkHz, useGlobalLedBuffer, maPerLed, maMax);
-        doInitBusses = true;  // finalization done in beginStrip()
-      }
+
+      busConfigs.push_back(std::move(BusConfig(ledType, pins, start, length, colorOrder, reversed, skipFirst, AWmode, freqkHz, useGlobalLedBuffer, maPerLed, maMax)));
```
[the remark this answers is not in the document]
TripleWhy: I (personally) would also have tried to simply emplace a default BusConfig, and then modify its values instead of creating a bunch of values, then copying them all over to a BusConfig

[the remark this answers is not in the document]
blazoncek: Good idea.
BusConfig is fixed in size and has simple constructor though.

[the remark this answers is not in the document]
TripleWhy: Well in that world I would have removed the explicit constructor. Better for performance anyway, and I imagine better for code size too?
I don't see how BusConfig's size would have any impact here.

ijspear:
> Latest update will also fix #4380 until resolved in NeoPixelBus.

Amazing! Are your artifacts using this latest code or should I compile myself?

blazoncek:
You can use artifacts from this PR. Click on ✅ of last commit.

wled00/FX_fcn.cpp
```diff
@@ -1220,6 +1233,34 @@ void WS2812FX::finalizeInit() {
...
+  for (const auto &bus : busConfigs) {
+    if (Bus::isDigital(bus.type) && !Bus::is2Pin(bus.type)) {
+      digitalCount++;
+      if (bus.count > maxLedsOnBus) maxLedsOnBus = bus.count;
+    }
+  }
+  DEBUG_PRINTF_P(PSTR("Maximum LEDs on a bus: %u\nDigital buses: %u\n"), maxLedsOnBus, digitalCount);
+  // we may remove 300 LEDs per bus limit when NeoPixelBus is updated beyond 2.9.0
+  if (maxLedsOnBus <= 300 && useParallelI2S) BusManager::useParallelOutput(); // must call before creating buses
+  else useParallelI2S = false; // enforce single I2S
```
willmmiles, line 1249: Changing a user parameter setting behind their back seems a bit dirty to me.  This'll cause weird behaviour like "I ticked the box, saved my settings, and it mysteriously unticked itself next time I visited the settings page."

If we want to give ourselves the option to disable parallel I2S when outside of the safe limits, I suggest calling the user adjustable parameter "allowParalleI2S" instead, and keep the parameter separate from the storage of when we're actually using parallel I2S.

wled00/wled.h
```diff
@@ -398,6 +398,9 @@ WLED_GLOBAL byte bootPreset   _INIT(0);                   // save preset to load
 WLED_GLOBAL bool useGlobalLedBuffer _INIT(false); // double buffering disabled on ESP8266
 #else
 WLED_GLOBAL bool useGlobalLedBuffer _INIT(true);  // double buffering enabled on ESP32
+  #ifndef CONFIG_IDF_TARGET_ESP32C3
```
willmmiles, line 401: Also 8266.

wled00/bus_manager.cpp
```diff
@@ -771,50 +776,74 @@ std::vector<LEDType> BusNetwork::getLEDTypes() {
...
 }
 
 int BusManager::add(const BusConfig &bc) {
+  DEBUGBUS_PRINTF_P(PSTR("Bus: Adding bus (%d - %d >= %d)\n"), getNumBusses(), getNumVirtualBusses(), WLED_MAX_BUSSES);
   if (getNumBusses() - getNumVirtualBusses() >= WLED_MAX_BUSSES) return -1;
+  unsigned numDigital = 0;
+  for (const auto &bus : busses) if (bus->isDigital() && !bus->is2Pin()) numDigital++;
   if (Bus::isVirtual(bc.type)) {
-    busses[numBusses] = new BusNetwork(bc);
+    //busses.push_back(std::make_unique<BusNetwork>(bc)); // when C++ >11
```
willmmiles, line 834: No need to wait - `make_unique` requires no new compiler features, it's simple enough we can bring a shim implementation:

```cpp
#if __cplusplus >= 201402L
using std::make_unique;
#else
// Really simple C++11 shim for non-array case; implementation from cppreference.com
template<class T, class... Args>
std::unique_ptr<T>
make_unique(Args&&... args)
{
    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
}
#endif
```

willmmiles (commented):
Overall thoughts: definitely a worthwhile change set.  I'm happy to see fewer and fewer C-style pointers.

Longer term, I'd like to squash down all the parallel I2S logic to be inside BusDigital; it bugs me a little bit that this needs to be exposed and handled elsewhere, as in some sense it's an "execution detail" of the digital output subsystem.  That said, it's also clear that there's value in having a staged config->plan->construction process which permits more "global" analysis and optimization over the plan, and it's not obvious to me how to devolve that to local knowledge (like how to manage digital output resource allocation).

wled00/FX_fcn.cpp
blazoncek: That's just a safety measure. The limit is also implemented in UI.

wled00/wled.h
blazoncek: Excluded 2 lines above.

wled00/bus_manager.cpp
blazoncek: Experimented by using `-std=gnu++14` as suggested by @TripleWhy and it worked.
Unfortunately code bloated by 1.4kB.

Wonder if this would make code any smaller.

wled00/FX_fcn.cpp
```diff
@@ -613,6 +613,19 @@ Segment &Segment::setPalette(uint8_t pal) {
   return *this;
 }
 
+Segment &Segment::setName(const char *newName) {
```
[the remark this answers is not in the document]
blazoncek: Internally?
I added a *from String* conversion function.

blazoncek:
What I also wanted to do is convert BusManager into namespace. But got linker errors that `_milliAmpsMax` was defined multiple times.

wled00/wled.h
willmmiles: Arg, outside the diff window!  Sorry about that, I should've checked. :)

wled00/bus_manager.cpp
blazoncek: Using this shim and removing `delete` increases code by 452B.
Compared to gnu++14 a huge win if it yields same safety.

willmmiles: I dug around online for flags to enable newer C++ standards, and turned up a bunch of forum posts from people recommending against it - something something ESP32 platform not tested, doesn't support it, strange crashes, etc.   I was surprised - I wouldn't've anticipated trouble from a whole-system build - but it didn't seem to be worth chasing down how applicable those old posts still were for this one case.

I did a quick test with the shim above, and the results were within few hundred bytes for most targets (though the older ESP32 platform really suffers - we need to drop that ASAP)

| Target | a98685d | 244e0c95 | Cost |
| ------ | ------- | -------- | ---- |
| nodemcuv2 | 879847 | 879879 | 32 |
| esp8266_2m | 878183 | 878215 | 32 |
| esp01_1m_full | 886715 | 886763 | 48 |
| nodemcuv2_160 | 879851 | 879883 | 32 |
| esp8266_2m_160 | 878187 | 878219 | 32 |
| esp01_1m_full_160 | 886719 | 886767 | 48 |
| nodemcuv2_compat | 866399 | 866447 | 48 |
| esp8266_2m_compat | 864719 | 864767 | 48 |
| esp01_1m_full_compat | 873347 | 873379 | 32 |
| esp32dev | 1347273 | 1347709 | 436 |
| esp32dev_V4 | 1250905 | 1251001 | 96 |
| esp32_eth | 1357633 | 1358029 | 396 |
| lolin_s2_mini | 1280009 | 1280137 | 128 |
| esp32c3dev | 1152868 | 1152938 | 70 |
| esp32s3dev_16MB_opi | 1150241 | 1150337 | 96 |
| esp32s3dev_8MB_opi | 1150241 | 1150337 | 96 |
| esp32s3_4M_qspi | 1148601 | 1148697 | 96 |
| esp32_wrover | 1171514 | 1171506 | -8 |

wled00/FX_fcn.cpp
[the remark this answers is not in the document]
willmmiles: Yes?  eg.  `String name`, instead of `char* name`.

wled00/FX_fcn.cpp
willmmiles: Hmm.  If we treat the UI parameter as 'allow' instead of 'force', we don't need to limit it in the UI, and there won't be any further UI changes necessary when/if a NPB fix is merged.  My $0.02, though - I'll defer to your judgement either way.

willmmiles:
> What I also wanted to do is convert BusManager into namespace. But got linker errors that `_milliAmpsMax` was defined multiple times.

A missing extern in the header maybe?  I think making BusManager in to a namespace is a good idea, too, but it could be a second PR.

wled00/FX_fcn.cpp
blazoncek: The issue of parallel I2S is that it uses 8x the memory needed for longest bus even if only one bus is used.
My own experimentation showed that it is unsafe to use parallel I2S if LED count on any bus exceeds 300 as free heap gets so low that AWS ceases to function properly, randomly locking ESP. For S2 that number gets lower still as it has way less RAM.

By using "force" instead of "allow" one can add safeguards which can be explained. The other issue with "allow" is when do you allow it? If user wants 15 outputs he/she/it could only do it by using less than 300 pixels per bus. So the moment user creates 10th bus the maximum length of any new bus (including 9th) must not exceed 300 (as first 8 use RMT in such case).

I could modify UI to to accommodate "allow" but I need to think about it.

blazoncek:
Got it working as a namespace. No real benefits that I can see, though.

wled00/bus_manager.cpp
blazoncek: There are only 4 calls in BusManager::add() function. Each for one of the classes: BusDigital, BusPwm, BusOnOff and BusNetwork.
There may be additional classes in the future (BusUsermod or BusVirtual).

TripleWhy: I was thinking more long term: What happens if make_unique starts to get used in more places

blazoncek:
@willmmiles @TripleWhy would [such](https://github.com/blazoncek/WLED/commit/7ec6f1e1190e9db2689b0b3182e1c23237e733b7) approach be ok?

I am not sure about `BusManager::getBus()` as it does no checking if vector is empty and assumes that `busses[0]` always exist (it should).

willmmiles:
> @willmmiles @TripleWhy would [such](https://github.com/blazoncek/WLED/commit/7ec6f1e1190e9db2689b0b3182e1c23237e733b7) approach be ok?
> 
> I am not sure about `BusManager::getBus()` as it does no checking if vector is empty and assumes that `busses[0]` always exist (it should).

Re conversion to namespace: Alas, exposing a couple internal variables is the price we pay for inlining.  Best we can do without link-time optimization.

Re getBus(): Hm, it's probably not ideal to return something strange given an invalid input.  It's not unreasonable to expect the callers of `getBus()` to have to explicitly handle any invalid inputs.  I think there are two real options: either (a) permit `getBus()` to indicate failure (for example, return a `Bus*` which can be `nullptr` if the index is bad -- this is one of those rare cases for bare pointers); or (b) decide that `getBus(index)` where `index>=getNumBusses()` is UB and the caller is responsible for making sure that doesn't happen.

TripleWhy:
Your question is just about line 473 in bus_manager.h in that commit, right? (I could say more about the changes of that commit if you want, but I'm not sure you'd like it :)

If it is ok depends on how sure you can be that `busses` is never empty when `getBus` is called. The safe approach would be to check for emptiness, and have an empty dummy bus somewhere that you can return (typically a static variable in the cpp file, but you'd need to move the implementation to the cpp file as well). That costs a bit of performance and space.
This is better suited for functions that return const references though. Another way would be to return an optional. (Or a pointer. You'd have to make it clear somehow that the caller doesn't own the pointer.)
  
You could also make it the caller's responsibility to never call `getBus` when busses is empty. In that case, you should make that clear with at least a comment.  
What I would usually do in such a case is add an assert as the first line in the function. That serves a way to formalize the precondition contract, and could be enforced by some contract framework if you want. But more importantly, the assert always runs on developer and tester machines, but never in production. I'm not sure how feasible such an approach is for something like WLED however.

Also note that typically you'd define two `getBus` functions, one that returns a mutable reference and one that returns a const reference. That works much better when you have non-static class methods, but can also be done with two separate function names. I guess that would also increase code size, however.

blazoncek:
> how sure you can be that `busses` is never empty

It is supposed to have 1 bus at least unless compilation overrides made impossible bus configuration (GPIO conflict).

> make it the caller's responsibility to never call `getBus` when busses is empty

That is already in place as all loops go: `for (i=0; i < BusManager::getNumBusses(); i++)` which is equal to `busses.size()`.

All in all it may be best to stick with returning a pointer and do explicit nullptr check everywhere.
Unfortunately it cannot be const as some calls to modifier functions are necessary.

> but I'm not sure you'd like it :)

Go ahead unless it is personal. 😉 Best done inline in the commit.

wled00/FX_fcn.cpp
willmmiles: First off: I'm now convinced the above behavior is OK for merging.

I guess the core idea with "allow" is "let the software decide what's the sensible thing for a given set of pins".  For example, the software could select the shortest 8 buses to give to parallel I2S; it doesn't necessarily have to be in order.  IIRC, the other challenge with parallel I2S is that you pay the RAM cost for 8 buses, even if you only use 2.  Either way - "force" or "allow" - describing the specific allocation strategy to users, especially those trying to work around bugs or manage RAM usage, can be challenging.

Long term, probably the right thing to do is to push bus hardware selection off to the configuration: let the user explicitly assign control resources, with an "auto" setting that makes a sensible guess.  The basic idea is that it gives the user control to avoid bugs or manage RAM usage without locking them in to an algorithmic decision; and auto gives us an out for the common use case of "don't care as long as it works".

blazoncek: That would be sensible but difficult.
The code would need to be modified heavily to determine order of creation (as they still need to be in the order user specified) and correctly decide between RMT and I2S. Current logic simply assumes either consecutive RMT or I2S as it is much easier to do that.

blazoncek:
After this is merged (if it is) I have another set of updates ready to allow unlimited virtual buses.

wled00/FX_fcn.cpp
willmmiles: Nothing wrong with starting simple and adding smarts later. :)

The big advantage of making the resource selection (RMT1, RMT2, I2S0, I2S1-parallel, etc.) in to explicit config parameters is that all the cleverness can be pushed to the UI; the C++ code only needs to confirm that there aren't any conflicts, it doesn't need to mess around with order or anything.  That said, I think at some point we'll want to make BusConfig an abstract class and let individual bus type classes (BusDigital, etc.) do their own JSON deserialization and store their own unique parameters in customized subclasses (BusDigitalConfig, etc.).  It'll make it easier to support new bus types as well.

wled00/FX_fcn.cpp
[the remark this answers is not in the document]
PaoloTK: > Yes? eg. `String name`, instead of `char* name`.

It's my understanding that String is 'evil' and should be avoided as much as possible, is this incorrect?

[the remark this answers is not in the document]
willmmiles: It's not my favourite string class, but IMHO writing your own manual memory allocation for heap allocated C strings is a greater evil.  Doubly so when your input is sometimes already Strings!

In a better world I'd advocate for `std::string` (because `std::string_view`, principally), but the benefits here don't outweigh the cost in consistency and/or code size.  (And, of course, `std` anything without exceptions offers no way to meaningfully handle OOM cases -- something we "pushing the hardware to its limits" folks have to contend with all too frequently...)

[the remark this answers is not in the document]
blazoncek: If you'll take a look into my fork you'll see that I experimented with modified malloc that prefers PSRAM if it exists. None of the built in classes use that.

As for evilness of manual memory management: In an ideal world where RAM is plenty and CPU clock is high you are better off using pre-made classes, but in a constrained world you *have to* know what are you doing.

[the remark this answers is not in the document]
TripleWhy: `std::string` supports allocators. I'm not in the loop what problems `std::string` has on esp though.

Also note that according to [this](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/external-ram.html), esp32 will by default prefer PSRAM when using `malloc`.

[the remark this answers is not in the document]
willmmiles: > `std::string` supports allocators. I'm not in the loop what problems `std::string` has on esp though.

`std` data structures do indeed support allocators - at the cost of doubling your code size if you end up mix-and-matching, and potentially making your API incompatible with anyone else's. :(

> 
> Also note that according to [this](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/external-ram.html#provide-external-ram-via-malloc), esp32 will by default prefer PSRAM when using `malloc`.

That's an ESP platform config setting which is set for us by the Arduino core.  We aren't building our own version of the platform yet, though I'm told some folks have experimented with it.

What we can do without changes is call [`heap_caps_malloc_extmem_enable()`](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/mem_alloc.html#_CPPv430heap_caps_malloc_extmem_enable6size_t) to explicitly enable sending malloc()s to external RAM.  We could even try calling `heap_caps_malloc_extmem_enable(0)` so all allocs go to external memory by default, and then explicitly `heap_caps_malloc_prefer(size, MALLOC_CAP_INTERNAL | MALLOC_CAP_DEFAULT, MALLOC_CAP_SPIRAM | MALLOC_CAP_DEFAULT)` for performance-sensitive allocations.

> As for evilness of manual memory management: In an ideal world where RAM is plenty and CPU clock is high you are better off using pre-made classes, but in a constrained world you _have to_ know what are you doing.

Absolutely true!  That's not an excuse for replicating code, though -- it means understanding the behaviour of the available classes/abstractions/tools so as to be able to choose the correct one for the job.  In fact I'd argue in this case that String's small-string optimization is almost certainly a net win, in that most segment names are short and so no extra heap allocation would be required at all.  If we're in to the micro-optimization game, it's better still; every heap allocation includes management overhead, so you get slammed twice on those small `malloc()`s (4 bytes for the pointer, 8+? bytes for the overhead).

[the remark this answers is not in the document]
blazoncek: > In fact I'd argue in this case that String's small-string optimization is almost certainly a net win

Can we agree to update the code once this is merged? So far this implementation is tested and works as expected (in fact does the same, albeit a bit more safely, than current `new`).

wled00/bus_manager.cpp
netmindz: Then we cross that bridge if and and when we come to it.  We should not keep PRs open longer than we need to while worrying about hyperthericals

netmindz:
I see we have quite a few comment threads open on this PR, do we need further changes or are some of them now resolved?

blazoncek:
There are just 2 questions really: Do we move make_unique into other header (and rely on built-in implementation when toolchain is updated) and do we switch to String for segment name.
Both can be resolved at later time IMO and the second is really unrelated to this PR and was just a collateral damage from modifying my own fork.

wled00/FX_fcn.cpp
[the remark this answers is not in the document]
willmmiles: Works for me.

wled00/bus_manager.cpp
willmmiles: I'm happy to defer migrating to `make_unique` to a follow-on PR.

wled00/FX_fcn.cpp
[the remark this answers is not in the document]
blazoncek: @willmmiles how do you use `String` in move assignemnt/constructor? I've tried:
```c++
Segment& Segment::operator= (Segment &&orig) noexcept {
  if (this != &orig) {
    //if (name) { free(name); name = nullptr; } // free old name
    name.clear();
    stopTransition();
    deallocateData(); // free old runtime data
    memcpy((void*)this, (void*)&orig, sizeof(Segment));
    //orig.name = nullptr;
    //what to do here?????
    orig.data = nullptr;
    orig._dataLen = 0;
    orig._t   = nullptr; // old segment cannot be in transition
  }
  return *this;
}
```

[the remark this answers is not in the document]
willmmiles: `String` is already movable.  The ideal solution is to make every member movable, so you don't have to implement a custom move constructor at all.   (For example, using `std::vector<byte>` or `DynamicBuffer` for `_data`, and something like `std::unique_ptr<Transition>` for `_t`).  If none of the standard classes have the right behaviour, it's ok to make up tiny little classes to describe the copy and move behaviour you want.

Barring that, declare the special typed members first in the struct, so you can block copy everything else without having to awkwardly work around them, eg:

```cpp
struct Segment { 
   String name;

   uint16_t start; 
   // other stuff
}

Segment& Segment::operator=(Segment&& orig) noexcept {
  if (this != &orig) {
    name = std::move(orig.name);
    memcpy(&start, &orig.start, sizeof(*this) - offsetof(decltype(this), start));
    // We've now stolen the other side's dynamic objects
    orig.data = nullptr;
    orig._datalen = 0;
    orig._t = nullptr;
};
```

wled00/FX_fcn.cpp
```diff
@@ -646,6 +646,19 @@ Segment &Segment::setPalette(uint8_t pal) {
...
+  if (newName) {
+    const int newLen = min(strlen(newName), (size_t)WLED_MAX_SEGNAME_LEN);
+    if (newLen) {
+      if (name) name = static_cast<char*>(realloc(name, newLen+1));
+      else      name = static_cast<char*>(malloc(newLen+1));
+      if (name) strlcpy(name, newName, newLen);
```
Liliputech, line 655: here it needs to be "newLen + 1" overwise the last character is truncated.

blazoncek:
If no one has any objections I think this is good to go in.

willmmiles (approved):

DedeHai:
@willmmiles this should also go into 0_15_0 even if its not a pure bugfix but it fixes the parallel I2S problem

blazoncek:
I agree. @netmindz what is your opinion?

willmmiles:
> @willmmiles this should also go into 0_15_0 even if its not a pure bugfix but it fixes the parallel I2S problem edit: sorry, I meant to tag netmindz :)

Lol I think it's probably fine.  :)  The line between "bugfix" and "feature" is often hazy - if backporting isn't too much work, might as well go for it.

netmindz:
I'm happy to go with the team consensus, provided suitable testing is done after the merge into 0-15-0 but before you push those changes

netmindz:
Multiple conflicts if you try and backport, this merge will need to be done by you @blazoncek

blazoncek:
Done and merged into 0_15_x

PR #4487, opened 2025-01-13
Swap to tasmota/platform-espressif32

netmindz:
Reduce the image size by swapping to tasmota's version of platform-espressif32

Motivation is that otherwise the move to IDF V4 means we run out of space

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -10,7 +10,7 @@
 # ------------------------------------------------------------------------------
 
 # CI/release binaries
-default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, nodemcuv2_compat, esp8266_2m_compat, esp01_1m_full_compat, esp32dev_V4, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_16MB_opi, esp32s3dev_8MB_opi, esp32s3_4M_qspi, esp32_wrover
+default_envs = nodemcuv2, esp8266_2m, esp8266_2m_tasmota, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, nodemcuv2_compat, esp8266_2m_compat, esp01_1m_full_compat, esp32dev_V4, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_16MB_opi, esp32s3dev_8MB_opi, esp32s3_4M_qspi, esp32_wrover
 
 src_dir  = ./wled00
 data_dir = ./wled00/data
@@ -379,6 +379,14 @@ platform_packages = ${common.platform_packages}
 board_build.ldscript = ${common.ldscript_2m512k}
 build_unflags = ${common.build_unflags}
 build_flags = ${common.build_flags} ${esp8266.build_flags} -D WLED_RELEASE_NAME=\"ESP02\"
+lib_deps = ${esp8266.lib_deps}
+
+[env:esp8266_2m_tasmota]
+board = esp_wroom_02
+platform = https://github.com/tasmota/platform-espressif8266#2024.09.00 ;; TODO: not sure this is the correct version
+board_build.ldscript = ${common.ldscript_2m512k}
+build_unflags = ${common.build_unflags}
+build_flags = ${common.build_flags} ${esp8266.build_flags} -D WLED_RELEASE_NAME=\"ESP02_Tasmota\"
 lib_deps = ${esp8266.lib_deps}
 
 [env:esp8266_2m_compat]
--- a/platformio.ini
+++ b/platformio.ini
@@ -10,7 +10,7 @@
 # ------------------------------------------------------------------------------
 
 # CI/release binaries
-default_envs = nodemcuv2, esp8266_2m, esp8266_2m_tasmota, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, nodemcuv2_compat, esp8266_2m_compat, esp01_1m_full_compat, esp32dev, esp32dev_V4, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_16MB_opi, esp32s3dev_8MB_opi, esp32s3_4M_qspi, esp32_wrover
+default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, nodemcuv2_compat, esp8266_2m_compat, esp01_1m_full_compat, esp32dev, esp32dev_V4, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_16MB_opi, esp32s3dev_8MB_opi, esp32s3_4M_qspi, esp32_wrover
 
 src_dir  = ./wled00
 data_dir = ./wled00/data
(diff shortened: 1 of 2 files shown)


platformio.ini
```diff
@@ -435,19 +433,18 @@ board_build.partitions = ${esp32.default_partitions}
...
 platform = ${esp32_idf_V4.platform}
-platform_packages = ${esp32_idf_V4.platform_packages}
 build_unflags = ${common.build_unflags}
 build_flags = ${common.build_flags} ${esp32_idf_V4.build_flags} -D WLED_RELEASE_NAME=\"ESP32_V4\" #-D WLED_DISABLE_BROWNOUT_DET
   ${esp32.AR_build_flags}
 lib_deps = ${esp32_idf_V4.lib_deps}
   ${esp32.AR_lib_deps}
 monitor_filters = esp32_exception_decoder
 board_build.partitions = ${esp32.default_partitions}
+board_build.flash_mode = qio ;; TODO: is this the correct flash mode?
```
netmindz, line 443: Is this the correct value? Adding fixed the missing sdkconfig.h error

Jason2866: I don't remember when the use of faulty bootloaders for mode QIO was fixed. Core 2.0.9 is really old.

platformio.ini
```diff
@@ -273,8 +273,9 @@ board_build.partitions = ${esp32.default_partitions}   ;; default partioning for
 ;;
 ;; please note that you can NOT update existing ESP32 installs with a "V4" build. Also updating by OTA will not work properly.
 ;; You need to completely erase your device (esptool erase_flash) first, then install the "V4" build from VSCode+platformio.
-platform = espressif32@ ~6.3.2
-platform_packages = platformio/framework-arduinoespressif32 @ 3.20009.0    ;; select arduino-esp32 v2.0.9 (arduino-esp32 2.0.10 thru 2.0.14 are buggy so avoid them)
+
+;; select arduino-esp32 v2.0.9 (arduino-esp32 2.0.10 thru 2.0.14 are buggy so avoid them)
+platform = https://github.com/tasmota/platform-espressif32/releases/download/2023.06.02/platform-espressif32.zip ;; Tasmota Arduino Core 2.0.9 with IPv6 support, based on IDF 4.4.4
```
netmindz, line 278: I was aiming to keep the same version of the Arduino Core as before, just swapping to Tasmota. 

Have I done this correctly? @Jason2866

Jason2866: The platform (with removing the `platform_packages` entry) change is correct.

Jason2866: But I wouldn't remove `platform_packages` completely. Just the entry.

netmindz: What should the correct value of `platform_packages` be then?

platformio.ini
```diff
@@ -383,6 +381,14 @@ build_unflags = ${common.build_unflags}
 build_flags = ${common.build_flags} ${esp8266.build_flags} -D WLED_RELEASE_NAME=\"ESP02\"
 lib_deps = ${esp8266.lib_deps}
 
+[env:esp8266_2m_tasmota]
+board = esp_wroom_02
+platform = https://github.com/tasmota/platform-espressif8266#2024.09.00 ;; TODO: not sure this is the correct version
```
netmindz, line 386: What is the correct version we need for 8266?

Were we really using Arduino 3.1.2 for 8266 while only 2.0.9 for esp32?!

blazoncek: Yes. IIRC @softhack007 had issues with newer Arduino cores.

softhack007: Confirmed - a few points are important:
* arduino for 8266 and arduino for esp32 are completely different - you can't compare their versions
* esp32 arduino 2.0.9 is "well proven" in use, while newer versions normally have new bugs. I've experimentally use 2.0.14 for some time and it seems to work, too. Beyond 2.0.14, there are new bugs again, like half-working serial, and some methods of the ESP class won't even compile for some boards.
* esp32 arduino v 3.x.y won't work for WLED. Too many API changes.

So esp32 arduino 2.0.14 looks most promising, and could be the right option for 0.16 development.

Arduino-esp32 v 3.0.x would be needed for supporting -C6. But in that case, I would suggest we switch to `pioarduino` which also comes from @Jason2866 and a few  others.

softhack007: PS2: arduino-esp32 v2.0.10 thru v2.0.13 are broken - I have stopped looking at them after experiencing some problems with audioreactive.

netmindz: Reverted back to 2.0.9 for esp32

platformio.ini
Jason2866: Just leave it empy. As following
`platform_packages =`

netmindz: What impact would that have?

platformio.ini
netmindz: I have updated flash_mode to dio now.

As for 2.0.9 we need to stay on 2.x due the breaking changes in 3.x that affect many of the libraries we use.

We have found various bugs with later releases of 2.0.x so hence why @softhack007 wants to stick with 2.0.9 for now

platformio.ini
netmindz: @Jason2866  If we want to stick with the 3.1.1 version of the Arduino Core for 8266 - is this possible with Tasmota?

netmindz:
non V4
```
RAM:   [=         ]  14.8% (used 48504 bytes from 327680 bytes)
Flash: [========= ]  86.0% (used 1351933 bytes from 1572864 bytes)
```
V4 - default
```
RAM:   [==        ]  17.1% (used 55972 bytes from 327680 bytes)
Flash: [==========]  96.6% (used 1519301 bytes from 1572864 bytes)
```

V4 - Tasmota
```
RAM:   [=         ]  14.7% (used 48216 bytes from 327680 bytes)
Flash: [========  ]  78.4% (used 1233457 bytes from 1572864 bytes) 

```

platformio.ini
Jason2866: Don't get the question. Arduino core for eap8266 and esp32x are two complete different things. So yes, for esp8266 everything you want can be used as platform/framework.

netmindz: Unlike the release page for the esp32 builds, it's not so clear exactly what is in each release.

If the versions listed follow the Arduino Core version, then the newest is 2.7.8 and we ideally want to stay on 3.1.1 as that was what we were using before and all the code is expecting

netmindz: Basically we were using this release https://github.com/platformio/platform-espressif8266/releases/tag/v4.1.0 and we want to stay with the same 3.1.1 of Arduino - but I am unsure if any of the release for tasmota's platform-espressif8266 match this

Jason2866: No Tasmota uses a modified Arduino esp8266 core 2.7.x version.
All 3.x versions uses more resources and Tasmota does not need features introduced with core 3.x. for esp8266. Furthermore we had stability issues with core 3.0.x when we tested with Tasmota. Probably this problems are fixed, but still there is no reason for Tasmota to use.

DedeHai:
`esp32c3dev` and `esp32s3_4M_qspi` are failing to compile

netmindz:
> `esp32c3dev`
fixed - set to qio

 >  `esp32s3_4M_qspi` 

Missing the board definition in Tasmota for some reason. Options are to either change the board or add a local copy of https://github.com/platformio/platform-espressif32/blob/master/boards/lolin_s3_mini.json to our repo. Thoughts?

DedeHai:
https://github.com/tasmota/platform-espressif32/blob/Arduino/IDF53/boards/adafruit_feather_esp32s3.json
has almost identical definitions but different `"extra_flags"` and an additional upload option `"use_1200bps_touch": true,`
I think it has little to no influence but I am not sure.

Jason2866:
The flag `use_1200bps_touch` has only an effect when TinyUsb is used in firmware.
For all boards not using TinyUsb strictly speaking this flag set is wrong.

netmindz:
I have just added a local copy of the board to our repo for now. We can remove later when we move to a platform version that contains `lolin_s3_mini`

netmindz:
All checks are now passing. Ok to approve? @blazoncek @softhack007

blazoncek:
If you are confident, go ahead since I do not have proper time to test this on my end.
Will be able to look into it on weekend.

PR #4497, opened 2025-01-16
Add IP address to info pane

kilrah:
When browsing from a computer by hostname there is no easy way to see the instance IP, could be useful to have it with the rest of the info


netmindz:
I don't see the need. If your network is properly setup and access by name works, then you should always use the name, in case you IP changes. 

If you need to know the ip of a device just ping the name. No point cluttering up the info pane

PR #4499, opened 2025-01-17
Fixed UDP Realtime streaming in DNRGBW Mode

Christanoid:
The udp.cpp class contained code to handle dnrgbw as a possible format, however a earlier if filtered out the possibility of actually using that mode due to not accepting the first protocol selection byte being 5 (4 is dnrgb, 5 was already added to be dnrgb**w**, however then filtered out again)

### DIFF
--- a/wled00/udp.cpp
+++ b/wled00/udp.cpp
@@ -605,8 +605,8 @@ void handleNotifications()
     return;
   }
 
-  //UDP realtime: 1 warls 2 drgb 3 drgbw
-  if (udpIn[0] > 0 && udpIn[0] < 5)
+  //UDP realtime: 1 warls 2 drgb 3 drgbw 4 dnrgb 5 dnrgbw
+  if (udpIn[0] > 0 && udpIn[0] < 6)
   {
     realtimeIP = (isSupp) ? notifier2Udp.remoteIP() : notifierUdp.remoteIP();
     DEBUG_PRINTLN(realtimeIP);


Christanoid:
Note: If this PR goes through smoothly I'll adapt the documentation next to ensure it is also documented on https://kno.wled.ge/interfaces/udp-realtime

softhack007:
@Christanoid thanks for your contribution.

As a general matter (applies to all PRs), please don't `force-push` while you have a PR open for our repo. We regularly lose review comments and have other weird behaviour due to force-pushers. You can simply add (push) commit to your branch, they will be added to the PR and visible to us. Thanks.

> [Christanoid](https://github.com/Christanoid) [force-pushed](https://github.com/Aircoookie/WLED/compare/b722f7ad922de840bef416ef0e7cccdefe460f3c..8ee2c44550677b1f03c71f3bce49d72a206d495b) the main branch from [b722f7a](https://github.com/Aircoookie/WLED/commit/b722f7ad922de840bef416ef0e7cccdefe460f3c) to [8ee2c44](https://github.com/Aircoookie/WLED/commit/8ee2c44550677b1f03c71f3bce49d72a206d495b)  [16 hours ago](https://github.com/Aircoookie/WLED/pull/4499#event-15983407329)

Christanoid:
I have read the rule on force pushing, sorry for going agianst it, I decided that since the PR was only a few minutes old and it was 1 AM (did not think about timezones, that one's on me), nobody will have seen it yet. Reason for the force push was a bit of local git chaos on my end forcing me to pull from the repo again before committing since my local checked out WLED folder was a bit older, so this PR's history was multiple commits, one of which was unnecessary. I therefore thought I might aswell clean the commit history. Won't do again in the future, once the PR is accepted it will likely get squashed anyways I assume.

Christanoid:
Is any further action from me required for the request to be merged? Since it has been stale with awaiting testing for a while now.

netmindz:
Sorry for the delay with testing 

To help with this could you just add some brief testing notes that explain setup needed, the behaviour of the bug Vs behaviour after the fix?

Christanoid:
No Problem.
To test this you only need an ESP32 with an RGBW stripe and a client to use the UDP Stream to live stream data. Previously simply nothing would happen, since the If condition changed in the PR would just filter out the mode for DNRGBW, however now if you set the protocol byte to 5 it will correctly let you send a stream of RGBW colors. 

I have been using it on my ambilight with a custom build for weeks now, seems to work fine for me. Only calculation still to be made is for the documentation, how many LED's can be controlled in one packet maximum. I calculated it in the past but have forgotten, would do so again once documentation needs changing. I do not see any possibility for a crash due to a packet being too big, since the handling of oversized packets should be identical to the one in the other udp streaming modes.

Christanoid:
I have been running this firmware on all my devices for months now. Any update on the review? Would appreciate going back to nightly / alpha again.

netmindz:
Sorry for the delay

PR #4506, opened 2025-01-19
Adding Particle System with many new FX

DedeHai:
this replaces the previous PR (https://github.com/Aircoookie/WLED/pull/3823) since I made too many changes and a rebase was needed.
**Some info on the particle system:**
- effects look best at 40-80FPS
- all FX can be overlaid: overlay is auto-detected, blurring may not work properly in multi-overlay
- PS and its buffers reside in segment data and is dynamically managed to keep a small RAM memory footprint (local render buffers are shared between segments)
- since most FX are random number based, segments do not run in sync
- 1M and 2M ESP8266 have the particle effects disabled for memory reasons

there are three compile time defines
- WLED_DISABLE_PARTICLESYSTEM2D -> disables 2D system and FX, automatically set if WLED_DISABLE_2D is set
- WLED_DISABLE_PARTICLESYSTEM1D -> disables 1D system and FX
- WLED_PS_DONT_REPLACE_FX -> re-enables the replaced FX (uses 12k more flash)

**Replaced FX**
to save on flash, effects that have a particle system based replacement are disabled. This saves 12k of flash. Users who want the originals back can custom compile using compile flag.

- Ghost Rider -> PS Ghost Rider (with new options and faster rendering)
- Blobs -> PS Blobs (with new options, optional AR functionality and nicer rendering)
- Fire 2012 -> PS Fire (1D and 2D versions)
- Fireworks 1D -> PS Fireworks (1D and 2D versions)
- Dancing Shadows -> PS Dancing Shadows
- Fireworks Starburst -> PS Starburst
- Glitter & Solid Glitter -> PS Sparkler (with overlay and settings tweked to users liking)
- Sparkle -> PS Sparkler
- Rolling Balls -> PS Pinball with checkmark "rolling"
- Multi Comet -> PS Pinball with gravity set to 0

**Known issues:**
- ESP8266 truncates the FX info JSON, the last added FX have missing sliders
- changing segment arrangements or changing FX during transitions can somettimes lead to crashes due memory race conditions/pointer issues
- replaced FX are not 100% identical

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -10,8 +10,6 @@
   Modified heavily for WLED
 */
 
-// information for custom FX metadata strings: https://kno.wled.ge/interfaces/json-api/#effect-metadata
-
 #include "wled.h"
 #include "FX.h"
 #include "fcn_declare.h"
@@ -2943,7 +2941,6 @@ typedef struct Ball {
 /*
 *  Bouncing Balls Effect
 */
-
 uint16_t mode_bouncing_balls(void) {
   if (SEGLEN <= 1) return mode_static();
   //allocate segment data
@@ -3023,7 +3020,6 @@ static const char _data_FX_MODE_BOUNCINGBALLS[] PROGMEM = "Bouncing Balls@Gravit
  *  Courtesy of pjhatch (https://github.com/pjhatch)
  *  https://github.com/Aircoookie/WLED/pull/1039
  */
-
 // modified for balltrack mode
 typedef struct RollingBall {
   unsigned long lastBounceUpdate;
@@ -3227,7 +3223,6 @@ typedef struct Spark {
 *  POPCORN
 *  modified from https://github.com/kitesurfer1404/WS2812FX/blob/master/src/custom/Popcorn.h
 */
-
 uint16_t mode_popcorn(void) {
   if (SEGLEN <= 1) return mode_static();
   //allocate segment data
@@ -4319,7 +4314,6 @@ typedef struct Spotlight {
  *
  * By Steve Pomeroy @xxv
  */
- 
 uint16_t mode_dancing_shadows(void)
 {
   if (SEGLEN <= 1) return mode_static();
@@ -5870,7 +5864,6 @@ static const char _data_FX_MODE_2DCRAZYBEES[] PROGMEM = "Crazy Bees@!,Blur,,,,Sm
 //     2D Ghost Rider  //
 /////////////////////////
 //// Ghost Rider by stepko (c)2021 [https://editor.soulmatelights.com/gallery/716-ghost-rider], adapted by Blaz Kristan (AKA blazoncek)
-
 #define LIGHTERS_AM 64  // max lighters (adequate for 32x32 matrix)
 uint16_t mode_2Dghostrider(void) {
   if (!strip.isMatrix || !SEGMENT.is2D()) return mode_static(); // not a 2D set-up
@@ -5956,12 +5949,10 @@ uint16_t mode_2Dghostrider(void) {
 static const char _data_FX_MODE_2DGHOSTRIDER[] PROGMEM = "Ghost Rider@Fade rate,Blur;;!;2";
 #undef LIGHTERS_AM
 
-
 ////////////////////////////
 //     2D Floating Blobs  //
 ////////////////////////////
 //// Floating Blobs by stepko (c)2021 [https://editor.soulmatelights.com/gallery/573-blobs], adapted by Blaz Kristan (AKA blazoncek)
-
 #define MAX_BLOBS 8
 uint16_t mode_2Dfloatingblobs(void) {
   if (!strip.isMatrix || !SEGMENT.is2D()) return mode_static(); // not a 2D set-up
@@ -7486,9 +7477,9 @@ uint16_t mode_2Dsoap() {
     }
   }
   // init also if dimensions changed
-  if (SEGENV.call == 0 || SEGENV.aux0 != cols || SEGENV.aux1 != rows) {
-    SEGENV.aux0 = cols;
-    SEGENV.aux1 = rows;
+  if (SEGENV.call == 0 || SEGMENT.aux0 != cols || SEGMENT.aux1 != rows) {
+    SEGMENT.aux0 = cols;
+    SEGMENT.aux1 = rows;
     for (int i = 0; i < cols; i++) {
       for (int j = 0; j < rows; j++) {
         SEGMENT.setPixelColorXY(i, j, ColorFromPalette(SEGPALETTE,~noise3d[XY(i,j)]*3));
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -10,10 +10,17 @@
   Modified heavily for WLED
 */
 
+// information for custom FX metadata strings: https://kno.wled.ge/interfaces/json-api/#effect-metadata
+
 #include "wled.h"
 #include "FX.h"
 #include "fcn_declare.h"
 
+#if !(defined(WLED_DISABLE_PARTICLESYSTEM2D) && defined(WLED_DISABLE_PARTICLESYSTEM1D))
+#include "FXparticleSystem.h"
+#else
+#define WLED_PS_DONT_REPLACE_FX
+#endif
 
  //////////////
  // DEV INFO //
@@ -29,7 +36,7 @@
   uint8_t  *fftResult = nullptr;
   float    *fftBin = nullptr;
   um_data_t *um_data;
-  if (usermods.getUMData(&um_data, USERMOD_ID_AUDIOREACTIVE)) {
+  if (UsermodManager::getUMData(&um_data, USERMOD_ID_AUDIOREACTIVE)) {
     volumeSmth    = *(float*)   um_data->u_data[0];
     volumeRaw     = *(float*)   um_data->u_data[1];
     fftResult     =  (uint8_t*) um_data->u_data[2];
@@ -1215,7 +1222,7 @@ uint16_t mode_dual_larson_scanner(void){
 }
 static const char _data_FX_MODE_DUAL_LARSON_SCANNER[] PROGMEM = "Scanner Dual@!,Trail,Delay,,,Dual,Bi-delay;!,!,!;!;;m12=0,c1=0";
 
-
+#ifdef WLED_PS_DONT_REPLACE_FX
 /*
  * Firing comets from one end. "Lighthouse"
  */
@@ -1242,7 +1249,7 @@ uint16_t mode_comet(void) {
   return FRAMETIME;
 }
 static const char _data_FX_MODE_COMET[] PROGMEM = "Lighthouse@!,Fade rate;!,!;!";
-
+#endif // WLED_PS_DONT_REPLACE_FX
 
 /*
  * Fireworks function.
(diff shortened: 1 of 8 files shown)


DedeHai:
the issue with truncated FXdata on ESP8266 can only be solved by increasing the `JSON_BUFFER_SIZE`. I found it requires at least 500 additional bytes but the problem will re-appear as soon as more FX are added.
@blazoncek and @willmmiles is it safe to increase the `JSON_BUFFER_SIZE` on the ESP8266 by 512 bytes? Is there a better solution?

DedeHai:
for reviewers:
- are changes in platformio.ini ok like that? the idea is that overrides can disable the system or re-enable replaced FX
- replaced FX are up for discussion, it definitely will lead to some annoyed users but having two FX doing almost the same would be a waste of flash (the particle FX do it more nicely and more versatile)

blazoncek:
@DedeHai I would not mess with JSON buffer size on ESP8266 unless you intend to do very rigorous testing (i.e. plenty of segments and usermods and LEDs and ledmaps, etc).

netmindz:
I need to have a look through from my laptop rather than my phone, however the first thing I spot is that if we assign new effect IDs for the ones we are "replacing" the existing effect with then we will be unable to have both the particle system and AnimARTrix in the same build as we will go over the limit on the number of effects we can handle. 

While there are some effects in AnimARTrix that could possibly be combined into a single effect with suitable parameters, it does feel a bit wasteful of effect IDs to leave the old ones unused in our default builds.

I probably need to have a play with the new PS replacements to get a feel for how similar they are to get a sense of how likely it is that folks will want to retain them. 

If we push the legacy ones into a legacy FX usermod, but don't preserve their original effectID would that be acceptable I wonder, provided it was properly documented in the release notes as breaking change?

DedeHai:
> I would not mess with JSON buffer size on ESP8266 unless you intend to do very rigorous testing (i.e. plenty of segments and usermods and LEDs and ledmaps, etc).

alright, thanks for the input. I already implemented the option to disable 2D PS FX by default on ESP8266, will add #ifdefs for ESP8266 and throw a compiler error if users want to add incompatible compile options. I need to run tests but it probably is "choose one of the three": classic FX or PS 1D or PS 2D

DedeHai:
> While there are some effects in AnimARTrix that could possibly be combined into a single effect with suitable parameters, it does feel a bit wasteful of effect IDs to leave the old ones unused in our default builds.

I need to check the order FX are added but if UM FX are added after the "native" ones, there could be a "cleanup" call after adding the last regular FX, removing all unused entries from the vector and freeing the slots for UM FX (not looking at code RN, so not 100% sure this is easy to do).
 
> I probably need to have a play with the new PS replacements to get a feel for how similar they are to get a sense of how likely it is that folks will want to retain them.

I tried to mimic them as close as possible but since its a completely different approach, its not possible. The main difference for most replacements is that the FX are frame and not time based, so speed depends on number of LEDs if FPS are not the same. There will be users that like the new ones and some that want the old ones back. 

> If we push the legacy ones into a legacy FX usermod, but don't preserve their original effectID would that be acceptable I wonder, provided it was properly documented in the release notes as breaking change?

From my experience users get annoyed if stuff changes in their existing presets, which is understandable so I think the current approach with keeping the IDs serves users better. I will have a look if unused IDs can be freed up to be available for UM FX.

blazoncek:
You could have an option to let user choose old vs. new effects. The question is, where to put that option.
I would recommend to create a new settings page intended to configure effects.

There were requests to "permanently" hide unwanted effects and this could be the place to fulfil that as well.

The implementation may not be trivial however, if someone would also want to configure "default" effect values or other (future) "hidden" parameters.

blazoncek:
Before I go too deep into review I'd like to mention this is not yet compatible with effect blending.

Namely, handling of palettes during a transition.

DedeHai:
> You could have an option to let user choose old vs. new effects. The question is, where to put that option. I would recommend to create a new settings page intended to configure effects.

it is more about code size: there are no other reasons to disable the old effects (apart from having two doing almost the same). But yes, if both new and old are enabled, that config page to select which ones to hide would be useful.

DedeHai:
> Namely, handling of palettes during a transition.

I saw that conflict, did not yet look at what the issue is, will try to work something out that works for both.

wled00/FX.cpp
```diff
@@ -29,7 +34,7 @@
   uint8_t  *fftResult = nullptr;
   float    *fftBin = nullptr;
   um_data_t *um_data;
-  if (usermods.getUMData(&um_data, USERMOD_ID_AUDIOREACTIVE)) {
+  if (UsermodManager::getUMData(&um_data, USERMOD_ID_AUDIOREACTIVE)) {
```
blazoncek, line 37: I think this is an oversight from @netmindz. AR effects no longer use that approach.

The correct way is: `um_data_t *um_data  = getAudioData();`

blazoncek:
One more thing. I do not know how others feel but IMO we should start splitting FX.cpp. It is becoming unmanageable.
I would split it into FX.cpp with all legacy effects, FX_2D.cpp with 2D effects, FX_AR.cpp with audio effects (including audio 2D) and finally FX_PS.cpp.

wled00/FX.cpp
netmindz: Yeah possibly missed during one of the rounds on reactoring we did around that. Example perhaps didn't match the actual code exactly so missed from the find and replace

wled00/cfg.cpp
```diff
@@ -1240,4 +1240,4 @@ void serializeConfigSec() {
   if (f) serializeJson(root, f);
   f.close();
   releaseJSONBufferLock();
-}
+}
```
netmindz, line 1243: Not essential to revert, but I recommend that Devs always double check the git diff --staged before commit or git show before push to ensure they are only making their intended change. If you use your graphical tools is far too easy to just click buttons and overlook exactly which files it's staging and assume the contents are accurate - treated like a "save all"

wled00/FX.h
```diff
@@ -977,4 +1008,4 @@ class WS2812FX {  // 96 bytes
 extern const char JSON_mode_names[];
 extern const char JSON_palette_names[];
 
-#endif
+#endif
```
netmindz, line 1011: Might be worth checking your IDE settings to not automatically remote trailing end of line chars

wled00/FXparticleSystem.cpp
```diff
@@ -0,0 +1,2329 @@
...
+    if(renderbuffer) {
+      deallocatePSmemory((void*)renderbuffer, renderBufferSize * sizeof(CRGB));
+      renderbuffer = nullptr;
+      renderBufferSize = 0;
+    }
+  }
+}
+
+// transfer the frame buffer to the segment and handle transitional rendering (both FX render to the same buffer so they mix)
+void transferBuffer(uint32_t width, uint32_t height, bool useAdditiveTransfer) {
```
blazoncek, line 2269: I am not comfortable with this implementation as it moves pixel blending to a second place while not providing "universal" solution.
Segment blending (as opposed to current pixel blending) should be implemented elsewhere (either by introducing triple-buffering or some other (yet undetermined) method).
I specifically do not approve changing blending mode during effect run.

My original idea was to remove calls to NPB/WS2812FX::setPixelColor() in Segment::setPixelColor() entirely and rely on Segment RGBW buffer (AKA setUpLeds()) entirely (also removing any grouping & spacing logic from Segment::setPixelColor().
When all segments are processed (FX fcn called) blend them in WS2812FX::show() using their opacity and blending mode.
Unfortunately this approach will have huge demands on RAM and CPU (moving pixels between buffers) and will not be suited for S2 or ESP8266.

DedeHai:
tested combinations that work with ESP8266 without touching JSON buffer size: the only thing that does not work is enabling both 1D and 2D particle system, which is now prohibited by an #ifdef. Enabling the replaced FX still produces a working UI.

DedeHai:
Updated to play nice with blending styles. @blazoncek  regarding 
>I specifically do not approve changing blending mode during effect run.

that (intermediate) disabling is necessary to not have one FX fade out and the other fade in as that is not how PS transitions work anymore when "fade" style is set. It is also only disabled during the buffer transfer. I don't see a better way except if another blending style specifically for the PS is added, but that is just confusing. If you have any other good ideas, I am open to suggestions.

blazoncek:
> If you have any other good ideas

I do not know if it is a "good idea" but am just thinking out loud:
Forget about blending inside PS system. There should be no notion of two separate runs of PS effects to know of each other (although you think it looks really good but it does so in edge cases only). Leave blending (mixing) to strip logic (as mentioned above) or segment logic (inside setPixelColor()).
It may sound counterproductive or counterintuitive but consider this scenario: 32x32 matrix with 5 segments set up, 0,0-16,16; 16,0-32,16; 0,16-16,32; 16,16-32,32 and 0,0-32,32 (or even a more complex one). Each of them running separate effect, the last segment overlays all others. It will be a lot of work for PS system to *correctly* blend that.

If done outside of PS it is as simple* as `color_blend()` for each pixel.
\* _not really that simple but simpler still_

DedeHai:
thanks for the input. I agree it would be better to do blending in one place BUT: blending between FX inside the PS is only done for one segment, not multiple. The reason it does that is to share the buffer AND the particles for one segment: if you had your proposed scenario, each segment would require two render buffers and twice the particles, which would exceed RAM already at 32x32 and multiple segments + transitions. Disabling the "particle blending" is as easy as changing two or three if's as it already disables it when not using 'fade' style. The "two PS FX can interact" part is more of an addition than a necessity, it could be a user selectable option as well.

DedeHai:
@netmindz feel free to squash-merge this

netmindz:
> @netmindz feel free to squash-merge this

Sorry I wasn't sure what you meant by this and got distracted before I replied. Was this just a note to use squash when I merge, but not yet or a request for merge with squash now?

DedeHai:
Just a suggestion to squash merge so commits are compacted. Merge when you took a look at it, any remaining issues are minor and can be worked out with more feedback from users

DedeHai:
@netmindz regarding the max number of FX: I think increasing `mode` (and `_modeT`) to int16_t and making the "reserved" to be -1 instead of the current 255 should be possible without too much fuss. I can take a stab at it in a new PR.

blazoncek:
@DedeHai there are numerous old effects that can be removed (and replaced with other existing effects with added sliders and options) some of which you already modified or consolidated. I would recommend to do that.

These are: 19, 53, 60, 77, 93, 94, 99, 102, 103, 117

blazoncek:
> I think increasing `mode` (and `_modeT`) to int16_t ... should be possible without too much fuss.

Are you aware of *all* consequences? I guess not otherwise you wouldn't make such a bold claim.

DedeHai:
> Are you aware of _all_ consequences? I guess not otherwise you wouldn't make such a bold claim.

I am not. please do enlighten me: which is the tricky part?

blazoncek:
Sync etc.

DedeHai:
@coderabbitai review

netmindz:
Are we good for you to merge now? If there are any outstanding issues I think we should open an issue to track them rather than blocking this PR any further

DedeHai:
yes, its good to merge. I will be pushing updates from time to time as I am still improving things as I see them in testing (or from feedback). 
I did not yet figure out why it sometimes crashes, whether its due to async timing of commands, memory handling conflicts with other FX or just a simple index-overflow.

netmindz:
I'm layed up in bed ill at the moment but can't wait to have a play when I'm doing a bit better

Issue #4396, opened 2024-12-14
Weird flicker when toggling on/off Athom bulbs with 0.15.0-b6


Andro-Marian:
### What happened?

When the fade-in starts and when fade-out is finishing. A weird flicked happens.
https://drive.google.com/file/d/1YgvZzD1sl7MFx1y-zQKS_JEvJhUdDfPS/view?usp=sharing

When turning on is starting to increase brightness, then turning instant off than on again. And vice versa.

### To Reproduce Bug

Flash the  0.15.0-b6 or the release 0.15 to any Athom 7w or 15w.
Turn it on/off by the web interface.

I think was introduces in **0.15.0-b6**. Works fine on **0.15.0-b5**

Noticed a flicker when going on or off.

### Expected Behavior

To be smooth as before.

### Install Method

Binary from WLED.me

### What version of WLED?

0.15-(b6/release)

### Which microcontroller/board are you seeing the problem on?

ESP8266

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

DedeHai:
I noticed this too when working on the new PWM outputs. It is related to fading value being offset from the starting and target value but I could not pin-point where this actually happens.

Andro-Marian:
Not sure what was changed on Beta 6. Thinking if the https://github.com/Aircoookie/WLED/issues/3196 was introduced first on b6

JorinL:
Hey there,
before opening a complete new - according to the video - it could be also my issue.

I think i might have the same thing - it "looks" pretty the same.

I have this problem on my 3 PWM strips.
When I cut the transitiontime it is gone - with transition it flashes when turning on before dimming up and when turning off while dimming down a flash and dimming further down.

I added a small video of the current situation.
https://github.com/user-attachments/assets/f6009672-1853-451c-b2fd-f404a0fe5acc

marcolino7:
Hi,
same behavoiur here, WLED 0.15 on PWM RGBW strip. I reverted back to 0.14 and issue gone

DedeHai:
I found one issue in the code related to CIE brightness curve calculation. Is the issue gone if you enable dithering in the PWM LED settings?

JorinL:
Thanks for fixing :-)
Tried the Alpha 16. Works :-)

SuulsaK:
[...]

PR #4518, opened 2025-01-24
Nightly release - Add bin.gz artifacts

zuckschwerdt:
This adds the bin.gz artifacts to the nightly builds.

I'm affected by #4396 and like to try #4419 as fix. I run this on a Shelly RGBW2 (ESP8266 with 2M) and need the `WLED_0.15.0_ESP02.bin.gz` which is currently not published for the nightly builds.


netmindz:
Thanks for this fix

PR #4524, opened 2025-01-27
ColorFromPaletteWLED: improved speed and fixed issue

DedeHai:
- fixed issue: blending was also done when color was on a key-index-color which is now skipped
- speed improvement: conversion is skipped if color is key-color


DedeHai:
@blazoncek is correct, I tested again and the original was actually faster. I thought the scaling could be done in 32bit at similar speed but apparently that is not the case. Reverted some of the changes.

netmindz:
@DedeHai this wasn't merged to 0_15_x and throws conflict when i try. Can you take a look please?

DedeHai:
@netmindz this was mislabelled, the fix is for a function added after 0.15 so it does not apply

PR #4536, opened 2025-02-03
Ant and PacMan effects

BobLoeffler68:
I added two new effects:  Ant and PacMan   Please take a look at them and test them and make any suggestions.  I'm a noob at this, so I will probably code these in the most inefficient manner.  :-)  I don't know if the speeds are going to be good on other esp chips (e.g. ESP8266).  I've only tested on an ESP32 WROOM 32D so far.  Both effects are pretty rough right now and experts could make them so much better, I would think.  I want to add at least another feature to reverse the direction of the PacMan effect.

## Summary by CodeRabbit

- New Features
	- Added new LED animation modes “Ants” and “Pac-Man” for enhanced visual effects.
	- Updated the Rolling Balls effect while removing an outdated effect.
- Chores
	- Streamlined the configuration by focusing on a single target environment.
	- Increased the upload speed for improved performance.


blazoncek:
Thanks.
You'll need to revert changes in platformio.ini and package-lock.json

BobLoeffler68:
> I like the general idea of this effect and the animation is well done. Suggestions:

Thanks!

> * you should return `FRAMETIME` as anything else is legacy and leads to problems with transitions and overlay

Yes, I knew changing the FRAMETIME would be ugly because it slows down the FPS rate, but I couldn't figure out how to do it any other way.  Do you have a suggestion of how to do it after looking at the code?

> * IMHO it should be more versatile by user-selectable number of yellow dots, selectable overall-speed and maybe white-dot spacing

Ah, good idea with the user-selectable yellow dots.

The speed is selectable with that first slider, but that is doing it by changing the FRAMETIME.  I need a suggestion for a better way to do it.  :-(

Yes, white dot spacing would be good to make it more obvious that PacMan is eating the dots.

> maybe I have some more points after testing :)

That would be awesome!  I am planning on changing the direction that they start moving.  Right now they move from LED 0 to SEGLEN, so I will add an option checkbox to start them moving from SEGLEN to 0.

Thanks!

DedeHai:
> Yes, I knew changing the FRAMETIME would be ugly because it slows down the FPS rate, but I couldn't figure out how to do it any other way. Do you have a suggestion of how to do it after looking at the code?

you need to render every frame, meaning rendering the same frame again if time has not advanced enough between calls.


> That would be awesome! I am planning on changing the direction that they start moving. Right now they move from LED 0 to SEGLEN, so I will add an option checkbox to start them moving from SEGLEN to 0.

that can be done by selecting "reverse" on the segment as well

BobLoeffler68:
> you need to render every frame, meaning rendering the same frame again if time has not advanced enough between calls.

Rendering every frame is still a new concept to me.  :-)  Thanks.

> that can be done by selecting "reverse" on the segment as well

Oh, that's right!  I won't worry about that one and will check that off of my list.  :-)

DedeHai:
I tested the two effects, I found this:
- first run of pacman does not clear / initialize correctly
- ants sometimes randomly disappear (or shrink?) for a brief moment

wled00/FX.cpp
```diff
@@ -3115,6 +3115,280 @@ static uint16_t rolling_balls(void) {
...
+
+    for (int i = 1; i < 5; i++) {                                               // draw the 4 ghosts (and black dots surrounding each ghost)
+      SEGMENT.setPixelColor(character[i].pos+1, BLACK);
+      SEGMENT.setPixelColor(character[i].pos, character[i].color);
+      SEGMENT.setPixelColor(character[i].pos-1, BLACK);
+      character[i].pos = character[i].pos-1;
+    }                                                                           // do not draw the power dot since PacMan ate it
+  }
+
+  return 20 + ((22 * (uint32_t)(255 - SEGMENT.speed)) / SEGLEN);                // FRAMETIME and FRAMETIME_FIXED were too fast
```
DedeHai, line 3387: one way to solve timing would be like this:
- use aux0 for the two states (use bits as flags)
- this frees up aux1 which can be used as a counter
- use strip.now and SEGENV.step to set the tick count (for example 100ms)
```
if(strip.now > SEGENV.step) {
  SEGENV.step = strip.now + 100;
  SEGENV.aux1++; 
}
```
and then you can use something like
`if(SEGENV.aux1 % 5 == 0) ` to execute something every 0.5s

drawing pixels can be done in every loop, advancing characters only if timing is fulfilled

BobLoeffler68:
> I tested the two effects, I found this:

Thanks for testing!

> * first run of pacman does not clear / initialize correctly

It should work if you set the Transition to 0.

> * ants sometimes randomly disappear (or shrink?) for a brief moment

An ant can "merge" with another ant of the same color, so it looks like it disappears, but it's just using the same LEDs as another ant.  That is probably what you are seeing.  I haven't seen any actually disappear or shrink.

wled00/FX.cpp
```diff
@@ -3115,6 +3115,280 @@ static uint16_t rolling_balls(void) {
...
+  const uint16_t maxNumAnts = 32;                                 // 255/16 + 1 ( * 2 ?????)
+  uint16_t dataSize = sizeof(rball_t) * maxNumAnts;
+  if (!SEGENV.allocateData(dataSize)) return mode_static();                    //allocation failed
+
+  int confusedAnt;                                                             // the first random ant to go backwards
+  rball_t *ants = reinterpret_cast<rball_t *>(SEGENV.data);
+
+  // number of ants based on intensity setting to max of 16 (32?)
+  uint8_t numAnts = SEGMENT.intensity/16 + 1;
+  if (SEGLEN > 127)
```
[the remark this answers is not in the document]
BobLoeffler68: I'll give that a try.

wled00/FX.cpp
```diff
@@ -3115,6 +3115,280 @@ static uint16_t rolling_balls(void) {
...
+    numAnts *= 2;                                                              // double the number of ants for longer strips/segments
+  
+  antSize = map(SEGMENT.custom1, 0, 255, 1, 5);                                // the size/length of each ant is user selectable (1 to 5 pixels) with a slider
```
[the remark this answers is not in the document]
BobLoeffler68: Yeah, I could do that.  I just didn't think anyone would want a 20-pixel sized ant!  LOL

wled00/FX.cpp
```diff
@@ -3115,6 +3115,280 @@ static uint16_t rolling_balls(void) {
...
+
+  uint8_t bgColorIdx = map(SEGMENT.custom2, 0, 255, 0, 5);                     // background color based on the what the user selects from the Background Color slider
```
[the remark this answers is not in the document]
BobLoeffler68: True, but the way I did it allows the user to have 3 different colored ants instead of only 2.  I guess we can see what the users want after they play with it a little.

wled00/FX.cpp
```diff
@@ -3115,6 +3115,280 @@ static uint16_t rolling_balls(void) {
...
+    case 3:  bgcolor = 0x845000;  break;  // brown
+    case 4:  bgcolor = 0x00aaaa;  break;  // cyan-ish
+    case 5:  bgcolor = 0x999999;  break;  // light gray (whitish)
+  }
+  
+  if (SEGENV.call == 0) {
+    confusedAnt = random(0,numAnts-1);
+    for (int i = 0; i < maxNumAnts; i++) {
+      ants[i].lastBounceUpdate = strip.now;
+      ants[i].velocity = 10.0f * float(random16(1000, 5000))/5000.0f;          // Random number from 1 to 5
```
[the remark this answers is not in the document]
BobLoeffler68: ok, I'll change that.  Thx.

wled00/FX.cpp
```diff
@@ -3115,6 +3115,280 @@ static uint16_t rolling_balls(void) {
...
+      ants[i].lastBounceUpdate = strip.now;
+    }
+    // check if reached past the beginning of the strip. If so, wrap around.
+    if (thisHeight <= 0.0f && ants[i].velocity < 0.0f) {
+      thisHeight = 1.0f;
+      ants[i].lastBounceUpdate = strip.now;
+      ants[i].height = thisHeight;
+    }
+    // check if reached past the end of the strip. If so, wrap around.
+    if (thisHeight >= 1.0f && ants[i].velocity > 0.0f) {
```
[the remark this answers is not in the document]
BobLoeffler68: Ok, thanks for the suggestion.  I'll look at that effect and code.

wled00/FX.cpp
```diff
@@ -3115,6 +3115,280 @@ static uint16_t rolling_balls(void) {
...
+*
+*   The first slider is for speed.
+*   Checkbox1 is for displaying white dots that PacMan eats.  Enabled will show white dots.  Disabled will not show any white dots (all leds will be black).
+*   Checkbox2 not used.
+*   Checkbox3 not implemented yet.  Will be for switching the direction of travel.  Enabled will start them going from LED 0 to SEGLEN; disabled will start them going from SEGLEN to 0. 
+*
+*/
+static uint16_t mode_pacman(void) {
+  //allocate segment data
+  const uint16_t numGhosts = 4;
```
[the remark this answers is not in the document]
BobLoeffler68: Ok, thx.

wled00/FX.cpp
```diff
@@ -3115,6 +3115,280 @@ static uint16_t rolling_balls(void) {
...
+  if (blinkingPowerdotTimer > 250) {                                            // every 250 milliseconds (1/4th of a sec)
+    if (character[5].color == ORANGEYELLOW)
+      character[5].color = BLACK;
+    else
+       character[5].color = ORANGEYELLOW;
+    SEGENV.step = strip.now;
+  }
+
+  // PacMan is eating the power dot!
+  if (character[0].pos >= SEGLEN) {
```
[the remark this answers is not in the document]
BobLoeffler68: Ah yes, good point.  Thx.

wled00/FX.cpp
```diff
@@ -3115,6 +3115,280 @@ static uint16_t rolling_balls(void) {
...
+    }
+    for (int i = 0; i < 5; i++) {                                               // move each character back one pixel, but not the orange-ish powerdot
+      character[i].pos = character[i].pos - 1;
+    }
+    SEGENV.aux1 = 1;                                                            // ghosts need to be blue now = true/yes
+  }
+
+  // When the ghosts are blue, and the last ghost is within the first 25% (or 33%) of LEDs in the segment...
+  if (SEGENV.aux1 == 1 && character[0].pos <= startBlinkingGhostsLED) {  
+    for (int i=1; i<5; i++) {                                                   // For all 4 ghosts...
```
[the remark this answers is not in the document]
BobLoeffler68: Yep, will do.

wled00/FX.cpp
```diff
@@ -3115,6 +3115,280 @@ static uint16_t rolling_balls(void) {
...
+      if (character[i].color == BLUE)
+          character[i].color = BLACK;
+      else
+          character[i].color = BLUE;
+    }
+    SEGENV.aux1 = 1;                                                            // ghosts still blue = true/yes
```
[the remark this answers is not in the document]
BobLoeffler68: I'll check on that.  Thx.

wled00/FX.cpp
BobLoeffler68: Ok, I'll work on that for the next version.
Thanks!!!!!

wled00/FX.cpp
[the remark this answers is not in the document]
DedeHai: I used a palette and got a lot more than 3 ant colors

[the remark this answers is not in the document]
blazoncek: Let me chime in.
You can use `color_from_palette()` to select color from selected palette or predefined color slot.
```c++
uint32_t fgColor = SEGMENT.color_from_palette(i, (true|false), (true|false), 0, [bri]);
uint32_t bgColor = SEGMENT.color_from_palette(i, (true|false), (true|false), 1, [bri]);
```
If user selects "default" palette, color slot will be used otherwise palette entry will be used.

You can also use, in a similar fashion, `SEGMENT.color_wheel(index)`.

blazoncek:
> It should work if you set the Transition to 0.

Do not assume such approach. The effect has to handle everything.

wled00/FX.cpp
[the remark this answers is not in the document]
DedeHai: you can do all that: there is the "auto generated" palettes from 2 or 3 colors, so if you choose the correct "hue" index it will use the three selected colors. Or make it a checkmark to use colors instead of palette or use color_wheel() which I think does either use colors or palette.
you can use one of the checkmarks as well to do "random colors" and then just assign a random hue index instead of fixed ones for more variation.
Just look at those functions (color from palette and colorwheel) and experiment.

DedeHai:
> > > It should work if you set the Transition to 0.
> > 
> > 
> > Do not assume such approach. The effect has to handle everything.
> 
> Good point. Is there a way to set the transition to 0 in the code?

no, you need to handle transitions in the FX.

blazoncek:
> Is there a way to set the transition to 0 in the code?

There is, but you don't do that from effect. You have to handle transitions (if needed at all).
@DedeHai has provided valuable information on how to do that.

wled00/FX.cpp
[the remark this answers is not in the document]
blazoncek: Color slots are just slots for stored color values. How you use them is up to you.
I have merely provided information how you can use best of both worlds: color slots and palettes.

The magic word is "Default palette". Default palette behaves a bit differently than other palettes as it is not necessarily a palette (as shown above). It can be color slot 0, 1 or 2 depending on how you choose to use it.

blazoncek:
> How do I revert changes to those two files in Github Desktop?

I don't use Github Desktop app but here are some instructions for git:
https://git-scm.com/book/ms/v2/Git-Basics-Undoing-Things
https://stackoverflow.com/questions/2733873/how-can-i-revert-a-single-file-to-a-previous-version

Basically you check-out a file from previous commit. And then commit it again.

NEVER FORCE PUSH YOUR CHANGES!

wled00/FX.cpp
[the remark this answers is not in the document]
BobLoeffler68: > you can do all that: there is the "auto generated" palettes from 2 or 3 colors, so if you choose the correct "hue" index it will use the three selected colors. Or make it a checkmark to use colors instead of palette or use color_wheel() which I think does either use colors or palette. you can use one of the checkmarks as well to do "random colors" and then just assign a random hue index instead of fixed ones for more variation. Just look at those functions (color from palette and colorwheel) and experiment.

Ok, thx.

blazoncek:
`git checkout <commit> -- path/to/filespec` is the correct way to revert a file to a previous commit.

BobLoeffler68:
> git checkout <commit> -- path/to/filespec is the correct way to revert a file to a previous commit.

So these should've worked, right?

C:\Users\boblo\Documents\GitHub\repos\WLED-AC>git checkout https://github.com/Aircoookie/WLED/commit/34d0f17597f31af1f29e994c4dec0eaddc0b924e -- C:\Users\boblo\Documents\GitHub\repos\WLED-AC\package-lock.json

C:\Users\boblo\Documents\GitHub\repos\WLED-AC>git checkout https://github.com/Aircoookie/WLED/commit/34d0f17597f31af1f29e994c4dec0eaddc0b924e -- C:\Users\boblo\Documents\GitHub\repos\WLED-AC\platformio.ini

I saw several websites that said that you can use just the filename of the file if you were running the command from the working directory.  But I'll use the full file path from now on.  Thanks.

wled00/FX.cpp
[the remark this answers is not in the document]
DedeHai: the new COB strips are pretty high density, I have one where 20 pixels is only like 5cm.

netmindz:
@coderabbitai review

platformio.ini
```diff
@@ -10,14 +10,16 @@
 # ------------------------------------------------------------------------------
 
 # CI/release binaries
-default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, nodemcuv2_compat, esp8266_2m_compat, esp01_1m_full_compat, esp32dev, esp32dev_V4, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_16MB_opi, esp32s3dev_8MB_opi, esp32s3_4M_qspi, esp32_wrover
+;default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, nodemcuv2_compat, esp8266_2m_compat, esp01_1m_full_compat, esp32dev, esp32dev_V4, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_16MB_opi, esp32s3dev_8MB_opi, esp32s3_4M_qspi, esp32_wrover
```
netmindz, line 13: please revert this change

platformio.ini
```diff
@@ -10,14 +10,16 @@
...
 
 src_dir  = ./wled00
 data_dir = ./wled00/data
 build_cache_dir = ~/.buildcache
 extra_configs =
   platformio_override.ini
 
+default_envs = esp32dev
+
```
netmindz, line 22: please revert this change

platformio.ini
```diff
@@ -125,7 +127,7 @@ framework = arduino
 board_build.flash_mode = dout
 monitor_speed = 115200
 # slow upload speed but most compatible (use platformio_override.ini to use faster speed)
-upload_speed = 115200
+upload_speed = 921600
```
netmindz, line 130: please revert this change

netmindz (changes requested):

wled00/FX.h
```diff
@@ -322,8 +322,11 @@ extern byte realtimeMode;           // used in getMappedPixelIndex()
 #define FX_MODE_WAVESINS               184
 #define FX_MODE_ROCKTAVES              185
 #define FX_MODE_2DAKEMI                186
+#define FX_MODE_ANTS                   187
```
blazoncek, line 325: Please re-use old (and removed) effects IDs.

wled00/FX.cpp
```diff
@@ -3115,6 +3115,280 @@ static uint16_t rolling_balls(void) {
 static const char _data_FX_MODE_ROLLINGBALLS[] PROGMEM = "Rolling Balls@!,# of balls,,,,Collisions,Overlay,Trails;!,!,!;!;1;m12=1"; //bar
 
 
+/*
```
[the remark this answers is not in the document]
blazoncek: IMO this is ok as-is.

wled00/FX.cpp
```diff
@@ -3115,6 +3115,290 @@ static uint16_t rolling_balls(void) {
...
+*   
+*   Checkbox2 is for Overlay mode (enabled is Overlay, disabled is no overlay)
+*   Checkbox3 is for whether the ants will bump into each other (disabled) or just pass by each other (enabled)
+*/
+static uint16_t mode_ants(void) {
+  //allocate segment data
+  uint32_t bgcolor = BLACK;
+  static constexpr uint16_t MAX_ANTS = 32;                                     // Maximum number of ants
+  static constexpr uint8_t DEFAULT_ANT_SIZE = 1;                               
+  uint8_t antSize = DEFAULT_ANT_SIZE;
```
blazoncek, line 3136: Unnecessarily using `uint8_t` or `uint16_t` increases code size as compiler needs to add width checks.
Code is also slower.

If you do not need 8-bit or 16-bit numbers (for wrapping) it is best to use `int` or `unsigned`.

wled00/FX.cpp
```diff
@@ -3115,6 +3115,290 @@ static uint16_t rolling_balls(void) {
...
+      ants[i].velocity = 10.0f * float(hw_random16(1000, 5000))/5000.0f;       // Random number from 1 to 5
+      if (i == confusedAnt)                                                    // make ant[i] go in the opposite direction
+        ants[i].velocity = -ants[i].velocity;
+      ants[i].height = (float(hw_random16(0, 10000)) / 10000.0f);              // Random number from 0 to 1  (used for the position of the ant on the strip)
+    }
+  }
+
+  float cfac = float(scale8(8, 255-SEGMENT.speed) +1)*20000.0f;                // this uses the Aircoookie conversion factor for scaling time using speed slider
+
+ 	if (!SEGMENT.check2) SEGMENT.fill(bgcolor);                                  // fill all LEDs with background color
```
blazoncek, line 3172: Indentation?

blazoncek (commented):
I would like you to also remove changes to package-lock.json and reuse removed effects IDs (marked in the source) for new effects.

Otherwise the code is clean and readable, no comments there.

wled00/FX.cpp
```diff
@@ -3124,38 +3124,27 @@ static const char _data_FX_MODE_ROLLINGBALLS[] PROGMEM = "Rolling Balls@!,# of b
 *   First slider is for the ants' speed.
 *   Second slider is for the # of ants.
 *   Third slider is for the Ants' size.
-*   
+*   Checkbox1 is for using the palettes (enabled) or the color slots (disabled)
```
DedeHai, line 3127: in other FX, this is also done using something like `if(palette==0)`

wled00/FX.cpp
```diff
@@ -3169,8 +3158,8 @@ static uint16_t mode_ants(void) {
 
   float cfac = float(scale8(8, 255-SEGMENT.speed) +1)*20000.0f;                // this uses the Aircoookie conversion factor for scaling time using speed slider
 
- 	if (!SEGMENT.check2) SEGMENT.fill(bgcolor);                                  // fill all LEDs with background color
- 
+  if (!SEGMENT.check2) SEGMENT.fill(SEGCOLOR(1));                              // fill all LEDs with background color (Bg) if the user didn't select Overlay checkbox
```
DedeHai, line 3161: much better :)

BobLoeffler68:
> * ants sometimes randomly disappear (or shrink?) for a brief moment

I found that this was happening when one of the ant colors was the same as the background color (e.g. a black ant on a black background).  You couldn't see that there was an ant there, until a different colored ant moved over it and then you would see just one or two pixels.  I fixed this so that the ant color will never be the background color if the user is using the color slots.  But it can still happen if they use a palette that has the same color as the background.  I don't think it will be a problem most of the time.

BobLoeffler68:
New option added to the Ant effect.  It is called "Gathering Food" and will add another pixel/LED (white, but will be yellow if the bg color is white) to each ant when it grabs a piece of food from one end of the LED segment.  It will then take it to the other end and "drop it". Then the ant will go get another piece of food.  This option will automatically enable the Pass By option so ants won't bump into each other.  (I found that the ants can get stuck at one end of the segment when the Pass By option is not enabled.)  Also, the normal flow of the ants is to move mostly in one direction and wrap around when they get to the last LED in the segment, but this option will change the flow to go back and forth between the ends of the segment.

There are two issues right now:

1) sometimes a few ants will turn white (or yellow if the bg color is white) when you enable the Gathering Food option if you previously enabled the Pass By option.  They will change to the correct color when they pick up some food, so it's not a critical bug.  I will try to fix this soon.

2) If the Gathering Food option is enabled and the Pass By option is disabled, and then you disable the Gathering Food option, the Pass By option will become enabled.  Confusing, but not a big deal as the user can then just disable the Pass By option again if they want to. I will try to fix this soon.

BobLoeffler68:
I'm also working on the transition issue in PacMan when it first starts up.

BobLoeffler68:
The transition in PacMan from a previous effect should look good now.

BobLoeffler68:
I just committed and did a push for the PacMan effect.  I added a new feature (a second power dot in the middle of the segment), and it works at first, but PacMan can't get past it.  PacMan will eat it (lines 3438-3448) and then chase the ghosts back to the beginning of the segment.  In line 3447, I'm setting FirstPowerDotEaten = TRUE, but then the white dots are displayed again (even though they shouldn't because line 3380 should see that it has been eaten and no white dots should be displayed). So at that point it seems like FirstPowerDotEaten is false, but it should still be true.  And when PacMan gets back to line 3438 again, it thinks the power dot has not been eaten, so it then goes back to the beginning of the segment instead of continuing to the end of the segment (and the other power dot).  It won't go past that power dot in the middle of the segment because it thinks FirstPowerDotEaten is false (but it should be true).  I'm setting that variable to false in only 2 places:
(1) Line 3364 in this if block:   if (SEGENV.call == 0) {
(2) Line 3477, but we never get to that line because LastPowerDotEaten is never true in line 3476.
Any help!  Thanks.

wled00/FX.cpp
```diff
@@ -3116,6 +3116,409 @@ static uint16_t rolling_balls(void) {
...
+  }
+
+  if (strip.now > SEGENV.step) {
+    SEGENV.step = strip.now;                 // "+ 100" creates a very jerky movement as the characters jump ahead several pixels each time they move
+    SEGENV.aux1++;
+  }
+
+  // draw white dots (or black LEDs) so PacMan can start eating them
+  if (SEGENV.aux0 & DIRECTION_FLAG && character[PACMAN].pos > 0) {
+    for (int i = character[4].pos; i > 0; i--) {                               // black out LEDs behind the last ghost (character[4]) in case they are on (transition from previous effect)
```
DedeHai, line 3375: this goes out of bounds with i-2

wled00/FX.cpp
```diff
@@ -3116,6 +3116,409 @@ static uint16_t rolling_balls(void) {
...
+      SEGMENT.setPixelColor(i-2, BLACK);
+    }
+
+    if (SEGMENT.check1) {                                                      // White Dots option is selected, so draw white dots (and black LEDs between them) in front of PacMan
+      if (!FirstPowerDotEaten) {
```
DedeHai, line 3380: better to save last "top" position of pacman and just start drawing from there instead of using a bool.

wled00/FX.cpp
```diff
@@ -3116,6 +3116,409 @@ static uint16_t rolling_balls(void) {
...
+    character[5].pos =   SEGLEN-1;                                              // last pixel of the segment
+
+    character[6].color = ORANGEYELLOW;                                          // orange-ish power dot (always blinks until it is eaten)
+    character[6].pos =   SEGLEN/2;                                              // pixel in the center of the segment
+
+    SEGMENT.setPixelColor(character[5].pos, character[5].color);                // draw the power dot in the last pixel/led
+    SEGMENT.setPixelColor(character[6].pos, character[6].color);                // draw the second power dot in the middle of the segment
+
+    SEGENV.aux0 = SEGENV.aux0 & ~GHOSTSBLUE_FLAG;                               // the ghosts are not blue yet, so set this to false (0)
+    FirstPowerDotEaten = false;
```
DedeHai, line 3364: this is a temporary variable, it is undefined in subsequent calls.

wled00/FX.cpp
```diff
@@ -3116,6 +3116,409 @@ static uint16_t rolling_balls(void) {
...
+  // draw white dots (or black LEDs) so PacMan can start eating them
+  if (SEGENV.aux0 & DIRECTION_FLAG && character[PACMAN].pos > 0) {
+    for (int i = character[4].pos; i > 0; i--) {                               // black out LEDs behind the last ghost (character[4]) in case they are on (transition from previous effect)
+      SEGMENT.setPixelColor(i-2, BLACK);
+    }
+
+    if (SEGMENT.check1) {                                                      // White Dots option is selected, so draw white dots (and black LEDs between them) in front of PacMan
+      if (!FirstPowerDotEaten) {
+        for (int i = SEGLEN-2; i > character[PACMAN].pos; i--) {               // start at the end of the segment (but not the last one because it's the orange power dot) and draw to the PacMan character
+          if (i != character[6].pos)                                           // don't draw a white dot where the middle power dot is
```
DedeHai, line 3382: just draw all pixels, overdraw it below with the yellow dot

wled00/FX.cpp
```diff
@@ -3116,6 +3116,409 @@ static uint16_t rolling_balls(void) {
...
+  if (character[PACMAN].pos >= SEGLEN) {
+    SEGENV.aux0 = SEGENV.aux0 & ~DIRECTION_FLAG;                                // reverse direction for all characters; 0=ghosts chasing pacman
+    for (int i = 1; i < numGhosts + 1; i++) {                                   // For all 4 ghosts...
+      character[i].color = GREEN; // BLUE;                                                // ...change their color to blue
+    }
+    for (int i = 0; i < 5; i++) {                                               // move each character back one pixel, but not the orange-ish power dots
+      character[i].pos = character[i].pos - 1;
+    }
+    SEGENV.aux0 = SEGENV.aux0 | GHOSTSBLUE_FLAG;                                // ghosts need to be blue now = true/yes
+    LastPowerDotEaten = true;
```
DedeHai, line 3460: all this variable does is being set true here, then set false again below, so its always false (or undefined)

wled00/FX.cpp
```diff
@@ -3116,6 +3116,409 @@ static uint16_t rolling_balls(void) {
...
+      character[6].color = BLACK;
+    else
+      character[6].color = ORANGEYELLOW;
+     
+    if ((SEGENV.aux0 & DIRECTION_FLAG) && !FirstPowerDotEaten)                 // draw/blink the first power dot (in the middle of the segment) only if it has not been eaten yet
+      SEGMENT.setPixelColor(character[6].pos, character[6].color);
+  }
+
+  // PacMan ate the first power dot! Chase the ghosts!
+  if ((character[PACMAN].pos >= character[6].pos) && !FirstPowerDotEaten) {
```
DedeHai, line 3438: FirstPowerDotEaten is undefined except in call==0

wled00/FX.cpp
```diff
@@ -3116,6 +3116,409 @@ static uint16_t rolling_balls(void) {
...
+     
+    if ((LastPowerDotEaten && SEGENV.aux0 & DIRECTION_FLAG) || (!LastPowerDotEaten && SEGENV.aux0 | ~DIRECTION_FLAG))    // draw/blink last power dot only if it has not been eaten yet
+      SEGMENT.setPixelColor(character[5].pos, character[5].color); 
+
+    if (character[6].color == ORANGEYELLOW)
+      character[6].color = BLACK;
+    else
+      character[6].color = ORANGEYELLOW;
+     
+    if ((SEGENV.aux0 & DIRECTION_FLAG) && !FirstPowerDotEaten)                 // draw/blink the first power dot (in the middle of the segment) only if it has not been eaten yet
```
DedeHai, line 3433: you should always draw it but only blink/set to orange if its not eaten (add a "bool draw" or something like that to the character struct, that way you can add as many dots as you want using the same code)

BobLoeffler68:
Added a new feature called Compact Dots in case the user wants a white dot in every LED instead of white dots with black LEDs between them.   That might look better in some situations, e.g. if a user uses the effect on their house using Govee or similar strings where the LED pixels/pucks are 8 inches or more apart.  Compact Dots won't do anything if the White Dots feature is disabled.  New version pushed to my repo here on GitHub.

wled00/FX.cpp
```diff
@@ -3360,163 +3362,151 @@ static uint16_t mode_pacman(void) {
     SEGMENT.setPixelColor(character[5].pos, character[5].color);                // draw the power dot in the last pixel/led
     SEGMENT.setPixelColor(character[6].pos, character[6].color);                // draw the second power dot in the middle of the segment
 
+    SEGENV.aux0 = DIRECTION_FLAG;                                               // initial direction of movement for all characters.  bit=1 (true = PacMan chases ghosts)
     SEGENV.aux0 = SEGENV.aux0 & ~GHOSTSBLUE_FLAG;                               // the ghosts are not blue yet, so set this to false (0)
-    FirstPowerDotEaten = false;
-    LastPowerDotEaten = false;
+    SEGENV.aux0 = SEGENV.aux0 & ~FIRSTPWRDOTEATEN_FLAG;                         // the first powerdot has not been eaten yet, so set this to false (0)
```
DedeHai, line 3367: usually, &= and |= are used to set flags (better readability)

wled00/FX.cpp
```diff
@@ -3291,53 +3292,54 @@ static uint16_t mode_ants(void) {
...
 } pacmancharacters_t;
 
 #define ORANGEYELLOW  (uint32_t)0xFF8800
 #define WHITEISH      (uint32_t)0x999999
-#define DIRECTION_FLAG  0b01
-#define GHOSTSBLUE_FLAG 0b10
+
+#define DIRECTION_FLAG          0b0001
+#define GHOSTSBLUE_FLAG         0b0010
+#define FIRSTPWRDOTEATEN_FLAG   0b0100
```
DedeHai, line 3320: instead of flagging them individually like this, add the flag to the character. that way you can have as many powerdots (or even ghosts) as you want.

wled00/FX.cpp
```diff
@@ -3360,163 +3362,151 @@ static uint16_t mode_pacman(void) {
...
       }
-      else {
-        for (int i = SEGLEN-2; i > character[PACMAN].pos; i--) {               // start at the end of the segment (but not the last one because it's the orange power dot) and draw to the PacMan character
-            SEGMENT.setPixelColor(i, WHITEISH);                                // white dots
-            SEGMENT.setPixelColor(i-1, BLACK);                                 // black LEDS between each white dot
-          i--;                                                                 // skip a dot so the dots are not next to each other
+      else {                                                                    // draw white dots with black LEDs between them
+        for (int i = SEGLEN-1; i > character[PACMAN].topPos; i--) {
+          SEGMENT.setPixelColor(i, WHITEISH);                                   // white dots
+          SEGMENT.setPixelColor(i-1, BLACK);                                    // black LEDS between each white dot
```
DedeHai, line 3391: putting the `if (SEGMENT.check2) {  `  here makes code more compact, no loop duplication

wled00/FX.cpp
```diff
@@ -3360,163 +3362,151 @@ static uint16_t mode_pacman(void) {
...
       character[6].color = BLACK;
     else
       character[6].color = ORANGEYELLOW;
-     
-    if ((SEGENV.aux0 & DIRECTION_FLAG) && !FirstPowerDotEaten)                 // draw/blink the first power dot (in the middle of the segment) only if it has not been eaten yet
-      SEGMENT.setPixelColor(character[6].pos, character[6].color);
   }
 
+  // blink last power dot (at the end of the segment) only if it has not been eaten yet
+  if (((character[PACMAN].topPos < character[5].pos) && (SEGENV.aux0 & DIRECTION_FLAG)) || ((character[PACMAN].topPos < character[5].pos) && (SEGENV.aux0 | ~DIRECTION_FLAG)))
```
DedeHai, line 3419: same as above: add flags to the characters themselves, so they "know" their current state (blinking, eaten, etc.) and you can just loop over all dots and blink the ones which have flags set instead of the dots being "dumb" and having to determine what to do with them at every corner.

DedeHai (commented):
this is good progress in general, but there is still a lot of redundant code that makes it hard to understand.

BobLoeffler68:
Hi @DedeHai  I've cleaned up the PacMan code and it's looking much better, but I still have one big issue.  I added the ability to select up to 5 power dots, but it's not working correctly. If I hard-code numPowerDots in line 3325, it works fine.  But when I allow the user to select the number of power dots with the intensity slider (line 3324), it doesn't work correctly.  No more than 3 power dots are displayed even if I select 4 or 5, and the power dots are not put in the correct LEDs.  Can you take a look at it?  I just pushed the latest code to my branch.  Thanks.

DedeHai:
looks a lot cleaner, I think you are getting the hang of it. I can check whats up with the dots.

BobLoeffler68:
> looks a lot cleaner, I think you are getting the hang of it. I can check whats up with the dots.

I really, really appreciate your help and guidance!

DedeHai:
fixed it. I did get a crash at one point when changing number of dots near the end of the animation, no idea why.

wled00/FX.cpp
```diff
@@ -3116,6 +3116,398 @@ static uint16_t rolling_balls(void) {
...
+  unsigned long lastBounceUpdate;
+  bool hasFood;
+  float velocity;
+  float height;
+} ant_t;
+
+static uint16_t mode_ants(void) {
+  //allocate segment data
+  uint32_t bgcolor = BLACK;
+  static constexpr unsigned MAX_ANTS = 32;                                     // Maximum number of ants
```
blazoncek, line 3145: Why `static`? This may force compiler to put variables into global memory, never to be reclaimed.

wled00/FX.cpp
```diff
@@ -3116,6 +3116,398 @@ static uint16_t rolling_balls(void) {
...
+typedef struct Ants {
+  unsigned long lastBounceUpdate;
+  bool hasFood;
+  float velocity;
+  float height;
+} ant_t;
+
+static uint16_t mode_ants(void) {
+  //allocate segment data
+  uint32_t bgcolor = BLACK;
```
blazoncek, line 3144: `bgcolor` should be `SEGCOLOR(1)` (as per metadata below) and not BLACK IMO.

wled00/FX.cpp
```diff
@@ -3116,6 +3116,398 @@ static uint16_t rolling_balls(void) {
...
+    // check if reached past the beginning of the strip.
+    if (thisHeight <= 0.0f && ants[i].velocity < 0.0f) {
+      if (SEGMENT.check1) {                                               // if looking for food, stop and go back the other way
+        thisHeight = 0.0f;
+        ants[i].velocity = -ants[i].velocity;                             // reverse direction
+        ants[i].lastBounceUpdate = strip.now;
+        ants[i].height = thisHeight;
+        ants[i].hasFood = true;                                           // found food
+        passBy = true;                                                    // when looking for food, pass by other ants without bumping into them
+        SEGMENT.check3 = true;                                            // when looking for food, pass by other ants without bumping into them
```
blazoncek, line 3200: Invalid request. Effect should not change *user controllable parameters*.

wled00/FX.cpp
```diff
@@ -3116,6 +3116,398 @@ static uint16_t rolling_balls(void) {
...
+            else
+              ants[j].velocity = -ants[j].velocity;
+            thisHeight = ants[i].height + ants[i].velocity*(strip.now - ants[i].lastBounceUpdate)/cfac;
+          }
+        }
+      }
+    }
+
+    uint32_t color;
+    if (SEGMENT.palette != 0 ) {                                                        // if a Palette is selected (besides the Default palette), use the palette's colors
```
blazoncek, line 3252: This `if` is unnecessary as that is what color_from_palette() does internally.
4th parameter determines which color (from the 3) it takes.

wled00/FX.cpp
```diff
@@ -3116,6 +3116,398 @@ static uint16_t rolling_balls(void) {
...
+#define PACMAN 0                                                                // PacMan is character[0]
+
+static uint16_t mode_pacman(void) {
+  constexpr unsigned numGhosts = 4;
+  unsigned maxPowerDots = SEGLEN / 5;                                          // Maximum number of power dots depends on segment length, max is 1 every 5 Pixels
+  
+  unsigned numPowerDots = map(SEGMENT.intensity, 0, 255, 1, maxPowerDots);                 // number of Power Dots (between 1 and 5) based on intensity slider setting
+
+  //update power dots
+  unsigned everyXLeds = (SEGLEN - 10) / numPowerDots;
```
blazoncek, line 3327: What if I only have 5 LEDs in segment?

wled00/FX.cpp
```diff
@@ -3116,6 +3116,398 @@ static uint16_t rolling_balls(void) {
...
+        SEGMENT.setPixelColor(character[i].pos, character[i].color);
+        SEGMENT.setPixelColor(character[i].pos-1, BLACK);
+        character[i].pos -= 1;                                                  // update their positions backwards for the next frame draw
+      }
+    }
+  }
+
+  return FRAMETIME;
+}
+static const char _data_FX_MODE_PACMAN[] PROGMEM = "PacMan@Speed,# of Power Dots,,,,White dots,Compact dots,;,,;!;1;m12=0,o1=1";
```
blazoncek, line 3508: There is no need to specify `;,,;` if there are no color selectors. jus use `;;`.

blazoncek (commented):
I think this could do without floats and there are some other aspects I'd like to be reconsidered and probably changed.

I did not trace execution so can't say anything about behaviour.

wled00/FX.cpp
DedeHai: that is cought above. min is 27.

BobLoeffler68:
> fixed it. I did get a crash at one point when changing number of dots near the end of the animation, no idea why. the powerdot spacing may need some fine-tuning, but using round() with floats is not a good way.

Thanks!  I did find a new bug with the dynamic setting of the number of power dots.  After approximately half of the power dots are eaten while the setting is set at the default (50% of the slider), change it to 100% and you will see the new powerdots in the pixesl that have already been eaten.  I can fix that by adding "&& character[i+5].pos > character[PACMAN].topPos" to line 3423:

  // now draw the power dots in the segment only if they have not been eaten yet
  for (int i = 0; i < numPowerDots; i++) {
    if (!character[i+5].eaten && character[i+5].pos > character[PACMAN].topPos)
      SEGMENT.setPixelColor(character[i+5].pos, character[i+5].color);
  }

But PacMan will still start chasing the ghosts at that pixel because line 3407 is placing the powerdot there, even though we can't see it because of my change above.  I tried doing the following, but then none of the powerdots are set/displayed (besides the last one).  :-(

  // update power dot positions: can change if user selects a different number of power dots
  for (int i = 1; i < maxPowerDots; i++) {
    if (character[i+5].pos > character[PACMAN].topPos)
      character[i+5].pos = 10 + i * everyXLeds;
  }

Any ideas?  And do we want to use maxPowerDots in there?  Or should it be numPowerDots?  I played with both but it didn't seem to help with this bug.

DedeHai:
> After approximately half of the power dots are eaten while the setting is set at the default (50% of the slider), change it to 100% and you will see the new powerdots in the pixesl that have already been eaten.

that is by design. if you want to fix it, there are two ways:
- in the part that sets number of powerdots, add a loop to check all powerdots, set all to eaten that have x<maxpacmanposition
- re-initialize the animation if the slider is moved

BobLoeffler68:
> * re-initialize the animation if the slider is moved

I think this is the cleanest and least confusing way to do it, so I'll do this for the next version.  Thx.

BobLoeffler68:
Hi @DedeHai and @blazoncek   I just pushed the latest version.  I've fixed all of the bugs that I know about and looked at everything you have mentioned, except the floats.  I see a lot of other effects with floats and I just did my effects the same as those, since I don't know how to not use floats.  Thanks for all of your help!  Both effects look really good now.

BobLoeffler68:
Hi @DedeHai and @blazoncek I just pushed the latest version.  I had found a crashing issue in PacMan when using the White Dots option on a 2D matrix.  It's fixed now.  I think both effects are done now and we can merge this branch into the main WLED (I think that is the correct terminology).  Please take a look when you have time.  If you think it's good, let me know what I need to do.  Thx!

DedeHai:
I need to look at it once more, I think the pacman effect is ok to merge, I am not too sure about the ants effect, I need to test it some more and check the code. As mentioned before, the ant effect may be better done using the PS: potentially shorter code, much more rendering options (colors, size of ants etc.)

BobLoeffler68:
> I need to look at it once more, I think the pacman effect is ok to merge, I am not too sure about the ants effect, I need to test it some more and check the code. As mentioned before, the ant effect may be better done using the PS: potentially shorter code, much more rendering options (colors, size of ants etc.)

Ok, let me know about the Ants code after you've had time to review it.  Thanks!

wled00/FX.h
```diff
@@ -323,6 +325,7 @@ extern byte realtimeMode;           // used in getMappedPixelIndex()
 #define FX_MODE_WAVESINS               184
 #define FX_MODE_ROCKTAVES              185
 #define FX_MODE_2DAKEMI                186
+
```
DedeHai, line 328: revert

wled00/FX.cpp
```diff
@@ -10286,6 +10700,11 @@ void WS2812FX::setupEffectData() {
   addEffect(FX_MODE_TV_SIMULATOR, &mode_tv_simulator, _data_FX_MODE_TV_SIMULATOR);
   addEffect(FX_MODE_DYNAMIC_SMOOTH, &mode_dynamic_smooth, _data_FX_MODE_DYNAMIC_SMOOTH);
 
+  // --- Bob L's 1D effects ---
```
DedeHai, line 10695: please remove this comment, your contribution is marked in the code

wled00/FX.cpp
```diff
@@ -3116,6 +3116,412 @@ static uint16_t rolling_balls(void) {
...
+        character[PACMAN].topPos = 0;                                           // set the top position of PacMan to LED 0 (beginning of the segment)
+      }
+    }
+  }
+
+  // display the characters
+  if (character[PACMAN].direction) {                                            // Going forward from the beginning of the segment...
+    if (SEGENV.aux1 % map(SEGMENT.speed, 0, 255, 15, 1) == 0) {                 // User-selectable speed of PacMan and the Ghosts
+      SEGMENT.setPixelColor(character[PACMAN].pos, character[PACMAN].color);    // draw PacMan
+      SEGMENT.setPixelColor(character[PACMAN].pos-1, BLACK);                    //   and the black dot behind him
```
DedeHai, line 3491: instead of drawing pixels black: why don't you clear all pixels at the beginning and draw them all fresh on each frame?

BobLoeffler68:
`  unsigned everyXLeds = ((SEGLEN - 10) << 8) / numPowerDots;
  for (int i = 1; i < maxPowerDots; i++) {
      character[i+5].pos = 10 + (i * everyXLeds) >> 8;
  }`

The everyXLeds code is not looking right on my 141 LED strip, so I'm going to play around with it.

DedeHai:
> The everyXLeds code is not looking right on my 141 LED strip, so I'm going to play around with it.

ups, there is a missing bracket: `character[i+5].pos = 10 + ((i * everyXLeds) >> 8);`

BobLoeffler68:
> > The everyXLeds code is not looking right on my 141 LED strip, so I'm going to play around with it.
> 
> ups, there is a missing bracket: `character[i+5].pos = 10 + ((i * everyXLeds) >> 8);`

Yes, that was it.  Thx!  New version pushed.  Looks very optimized now.

DedeHai:
I tested the ant effect yesterday and while the effect itself looks ok, the code is a mess.
It also uses 1.5k of flash, which is a lot for a rather simple (and specific) effect. In its current state, I don't think it should be added as a default effect (you could still move it to the user-effects usermod which is in PR)

BobLoeffler68:
> I tested the ant effect yesterday and while the effect itself looks ok, the code is a mess. It also uses 1.5k of flash, which is a lot for a rather simple (and specific) effect. In its current state, I don't think it should be added as a default effect (you could still move it to the user-effects usermod which is in PR)

I basically used the code from the rolling_balls FX and made modifications to it.  Last night I started looking at your PS code, so maybe I will try to convert it.  Not sure how that will go since I'm not a coding expert.  If that doesn't go well, I'll move it to the user-effects usermod.

DedeHai:
you can start looking at `mode_particleChase()` it probably comes closest in how I would implement it: it does manipulate and place particles "manually" i.e. not using a source to emit particles. You can create the ants either using several particles with a defined spacing (one pixel in 1D particle space has a "length" of 32) or you can use a single particle and increase its size by setting `PartSys->advPartProps[i].size` up to 255. make sure to initialize the PS with advanced=true or advPartProps is not available and it will crash if you try to access it.

BobLoeffler68:
Ok, thx.  I'll take a look at that soon.

BobLoeffler68:
Hi @DedeHai and @blazoncek   I haven't had any time to work on this for almost 2 months, but am hoping to this weekend or next weekend.  Life and family keep getting in my way.  :-)  There are 3 things that I want to get done:

1. Make sure the PacMan FX gets into the .16 version.
2. Remove the Ants FX code so I can work on it later (to clean it up, or convert it to PS, or put into a usermod fx).
3. Try to help figure out that 40k memory issue when using a preset for an ledmap with 2 FX in it.

Which one should I do first?  Do I need to close this PR (since I will probably remove the Ants FX) and then create a new branch and PR?

blazoncek:
I'd try to figure out what's causing memory drain 1st.
Then make PR for each effect separately.

BobLoeffler68:
> I'd try to figure out what's causing memory drain 1st. Then make PR for each effect separately.

Ok, I'll do that first.  My published code on GitHub is the same as my local "rolled back" code that I tried to merge/squash on 5/25/2025 in the custom-fx channel on the WLED discord.  It said that local code was already up-to-date with my published code.  So maybe I need to go back further and checkout an older commit.  I have a big gitlog saved as a text file.  Or the reflog if either of those would help.  Or would it be easier to create a new repository by cloning the latest code from WLED AC?

BobLoeffler68:
@blazoncek and @DedeHai   I installed SourceTree in case that will help us figure out how to get my published code back to a point where I can help find the memory error.  I've attached a screenshot of my main branch's history.  In May I tried going back to a good commit (b19ec2c3), but my published code on GitHub was up-to-date with my local branch, so I need to go further back.  Is it up-to-date with my PR branch because of the highlighted/selected branch merge (b9e7c2e5)?

BobLoeffler68:
BobL

BobLoeffler68:
Closed this PR and will open a new one for just the PacMan effect.

PR #4543, opened 2025-02-07
Soap FX optimization

DedeHai:
- saves 600 bytes of flash
- 2x speed increase thanks to local buffer

## Summary by CodeRabbit

- **New Features**
	- Expanded the range of lighting effects with over 50 new visualizations, including dynamic patterns, engaging animations, 2D displays, and audio-reactive modes for a richer user experience.
- **Refactor**
	- Optimized performance for smoother transitions and improved overall responsiveness.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -7469,33 +7469,39 @@ static const char _data_FX_MODE_2DDISTORTIONWAVES[] PROGMEM = "Distortion Waves@
 //@Stepko
 //Idea from https://www.youtube.com/watch?v=DiHBgITrZck&ab_channel=StefanPetrick
 // adapted for WLED by @blazoncek, optimization by @dedehai
-void soapProcessPixels(bool isRow, int size1, int size2, uint8_t* noise3d, int amplitude, int shift, CRGB* ledsbuff) {
-    for (int i = 0; i < size1; i++) {
-        int amount = ((int)noise3d[isRow ? XY(0, i) : XY(i, 0)] - 128) * 2 * amplitude + 256 * shift;
-        int delta = abs(amount) >> 8;
-        int fraction = abs(amount) & 255;
-        for (int j = 0; j < size2; j++) {
-            int zD, zF;
-            if (amount < 0) {
-                zD = j - delta;
-                zF = zD - 1;
-            } else {
-                zD = j + delta;
-                zF = zD + 1;
-            }
-            CRGB PixelA = CRGB::Black;
-            if ((zD >= 0) && (zD < size2)) PixelA = isRow ? SEGMENT.getPixelColorXY(zD, i) : SEGMENT.getPixelColorXY(i, zD);
-            else                           PixelA = ColorFromPalette(SEGPALETTE, ~noise3d[isRow ? XY(abs(zD), i) : XY(i, abs(zD))] * 3);
-            CRGB PixelB = CRGB::Black;
-            if ((zF >= 0) && (zF < size2)) PixelB = isRow ? SEGMENT.getPixelColorXY(zF, i) : SEGMENT.getPixelColorXY(i, zF);
-            else                           PixelB = ColorFromPalette(SEGPALETTE, ~noise3d[isRow ? XY(abs(zF), i) : XY(i, abs(zF))] * 3);
-            ledsbuff[j] = (PixelA.nscale8(ease8InOutApprox(255 - fraction))) + (PixelB.nscale8(ease8InOutApprox(fraction)));
-        }
-        for (int j = 0; j < size2; j++) {
-            if (isRow) SEGMENT.setPixelColorXY(j, i, ledsbuff[j]);
-            else SEGMENT.setPixelColorXY(i, j, ledsbuff[j]);
-        }
+void soapProcessPixels(bool isRow, uint8_t* noise3d, int amplitude, int shift, CRGB* ledsbuff) { //1153477-1152873
+  const int cols = SEG_W;
+  const int rows = SEG_H;
+  int rowcol, colrow;
+  rowcol = isRow ? rows : cols;
+  colrow = isRow ? cols : rows;
+
+  for (int i = 0; i < rowcol; i++) {
+    int amount = ((int)noise3d[isRow ? i * cols : i] - 128) * 2 * amplitude + 256 * shift;
+    int delta = abs(amount) >> 8;
+    int fraction = abs(amount) & 255;
+    for (int j = 0; j < colrow; j++) {
+      int zD, zF;
+      if (amount < 0) {
+          zD = j - delta;
+          zF = zD - 1;
+      } else {
+          zD = j + delta;
+          zF = zD + 1;
+      }
+      CRGB PixelA = CRGB::Black;
+      if ((zD >= 0) && (zD < colrow)) PixelA = isRow ? SEGMENT.getPixelColorXY(zD, i) : SEGMENT.getPixelColorXY(i, zD);
+      else                            PixelA = ColorFromPalette(SEGPALETTE, ~noise3d[isRow ? (abs(zD)%cols) + i*cols : i + (abs(zD)%rows)*cols] * 3);
+      CRGB PixelB = CRGB::Black;
+      if ((zF >= 0) && (zF < colrow)) PixelB = isRow ? SEGMENT.getPixelColorXY(zF, i) : SEGMENT.getPixelColorXY(i, zF);
+      else                            PixelB = ColorFromPalette(SEGPALETTE, ~noise3d[isRow ? (abs(zF)%cols) + i*cols : i + (abs(zF)%rows)*cols] * 3);
+      ledsbuff[j] = (PixelA.nscale8(ease8InOutApprox(255 - fraction))) + (PixelB.nscale8(ease8InOutApprox(fraction)));
     }
+    for (int j = 0; j < colrow; j++) {
+      if (isRow) SEGMENT.setPixelColorXY(j, i, ledsbuff[j]);
+      else SEGMENT.setPixelColorXY(i, j, ledsbuff[j]);
+    }
+  }
 }
 
 uint16_t mode_2Dsoap() {
@@ -7555,9 +7561,9 @@ uint16_t mode_2Dsoap() {
   CRGB ledsbuff[MAX(cols,rows)];
 
   amplitude = (cols >= 16) ? (cols-8)/8 : 1;
-  soapProcessPixels(true, rows, cols, noise3d, amplitude, shiftX, ledsbuff); // rows 1166192 vs 1165634
+  soapProcessPixels(true, noise3d, amplitude, shiftX, ledsbuff);  // rows
   amplitude = (rows >= 16) ? (rows-8)/8 : 1;
-  soapProcessPixels(false, cols, rows, noise3d, amplitude, shiftY, ledsbuff); // cols
+  soapProcessPixels(false, noise3d, amplitude, shiftY, ledsbuff); // cols
 
   return FRAMETIME;
 }
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -7469,40 +7469,54 @@ static const char _data_FX_MODE_2DDISTORTIONWAVES[] PROGMEM = "Distortion Waves@
 //@Stepko
 //Idea from https://www.youtube.com/watch?v=DiHBgITrZck&ab_channel=StefanPetrick
 // adapted for WLED by @blazoncek, optimization by @dedehai
-void soapProcessPixels(bool isRow, uint8_t* noise3d) {
-  const int cols = SEG_W;
-  const int rows = SEG_H;
-  const auto XY = [&](int x, int y) { return (x%cols) + (y%rows) * cols; };
-  CRGB ledsbuff[MAX(cols,rows)];
-  const int rowcol = isRow ? rows : cols;
-  const int colrow = isRow ? cols : rows;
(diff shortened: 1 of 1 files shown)


blazoncek:
Please use new lambda XY() as it simplifies review and prevents errors.

DedeHai:
it does, just not in the first commit ;)

wled00/FX.cpp
```diff
@@ -7468,7 +7468,42 @@ static const char _data_FX_MODE_2DDISTORTIONWAVES[] PROGMEM = "Distortion Waves@
...
+      if (amount < 0) {
+          zD = j - delta;
+          zF = zD - 1;
+      } else {
+          zD = j + delta;
+          zF = zD + 1;
+      }
+      CRGB PixelA = CRGB::Black;
+      if ((zD >= 0) && (zD < colrow)) PixelA = isRow ? SEGMENT.getPixelColorXY(zD, i) : SEGMENT.getPixelColorXY(i, zD);
+      else                            PixelA = ColorFromPalette(SEGPALETTE, ~noise3d[isRow ? (abs(zD)%cols) + i*cols : i + (abs(zD)%rows)*cols] * 3);
```
blazoncek, line 7494: I was referring to this where XY() is used to get the index of array.

DedeHai: I did that as a speed improvement, but can revert back.

wled00/FX.cpp
```diff
@@ -7526,52 +7562,9 @@ uint16_t mode_2Dsoap() {
...
-      else                          PixelA = ColorFromPalette(SEGPALETTE, ~noise3d[XY(abs(zD),y)]*3);
-      CRGB PixelB = CRGB::Black;
-      if ((zF >= 0) && (zF < cols)) PixelB = SEGMENT.getPixelColorXY(zF, y);
-      else                          PixelB = ColorFromPalette(SEGPALETTE, ~noise3d[XY(abs(zF),y)]*3);
-      ledsbuff[x] = (PixelA.nscale8(ease8InOutApprox(255 - fraction))) + (PixelB.nscale8(ease8InOutApprox(fraction)));
-    }
-    for (int x = 0; x < cols; x++) SEGMENT.setPixelColorXY(x, y, ledsbuff[x]);
-  }
-
+  soapProcessPixels(true, noise3d, amplitude, shiftX, ledsbuff);  // rows
```
blazoncek, line 7565: You don't need to pass `ledsbuff` as it is completely local.

wled00/FX.cpp
```diff
@@ -7468,7 +7468,43 @@ static const char _data_FX_MODE_2DDISTORTIONWAVES[] PROGMEM = "Distortion Waves@
 //Soap
 //@Stepko
 //Idea from https://www.youtube.com/watch?v=DiHBgITrZck&ab_channel=StefanPetrick
-// adapted for WLED by @blazoncek
+// adapted for WLED by @blazoncek, optimization by @dedehai
+void soapProcessPixels(bool isRow, uint8_t* noise3d, int amplitude, int shift, CRGB* ledsbuff) { //1153477-1152873
+  const int cols = SEG_W;
+  const int rows = SEG_H;
+  const auto XY = [&](int x, int y) { return (x%cols) + (y%rows) * cols; };
+  int rowcol, colrow;
```
blazoncek, line 7476: Can't these be `const`?

blazoncek (commented):
You could push it a bit further. 😉 
Perhaps using `std::swap()`?

blazoncek:
I was thinking about this to improve readability:
```c++
static void soapPixels(bool isRows, uint32_t *pixels, uint8_t *noise3d) {
  const int cols = SEG_W;
  const int rows = SEG_H;
  const auto XY = [&](int x, int y) { return (x%cols) + (y%rows) * cols; };
  const int hor = isRows ? rows : cols;
  const int ver = isRows ? cols : rows;
  const int amplitude = max(1, ((ver >= 16) ? (ver-8)/8 : 1) * (1 + SEGMENT.custom1) >> 6);

  CRGBW ledsbuff[ver];

  for (int y = 0; y < hor; y++) {
    int amount   = ((int)noise3d[isRows ? y*cols : y] - 128) * 2 * amplitude + (128 - SEGMENT.custom2)*2;
    int delta    = abs(amount) >> 8;
    int fraction = abs(amount) & 255;
    for (int x = 0; x < ver; x++) {
      int zD;
      int zF;
      if (amount < 0) {
        zD = x - delta;
        zF = zD - 1;
      } else {
        zD = x + delta;
        zF = zD + 1;
      }
      const int indxA = isRows ? XY(zD,y) : XY(y,zD);
      const int indxB = isRows ? XY(zF,y) : XY(y,zF);
      const int noisA = isRows ? XY(abs(zD),y) : XY(y,abs(zD));
      const int noisB = isRows ? XY(abs(zF),y) : XY(y,abs(zF));
      CRGB PixelA;
      CRGB PixelB;
      if ((zD >= 0) && (zD < ver)) PixelA = pixels[indxA];
      else                         PixelA = ColorFromPalette(SEGPALETTE, ~noise3d[noisA]*3);
      if ((zF >= 0) && (zF < ver)) PixelB = pixels[indxB];
      else                         PixelB = ColorFromPalette(SEGPALETTE, ~noise3d[noisB]*3);
      ledsbuff[x] = (PixelA.nscale8(ease8InOutApprox(255 - fraction))) + (PixelB.nscale8(ease8InOutApprox(fraction)));
    }
    for (int x = 0; x < ver; x++) {
      if (isRows) SEGMENT.setPixelColorXY(x, y, pixels[XY(x,y)] = ledsbuff[x]);
      else        SEGMENT.setPixelColorXY(y, x, pixels[XY(y,x)] = ledsbuff[x]);
    }
  }
}
```
Where `pixels[]` is an array of previous pixel values since `getPixelColor()` does not work with gaps.

DedeHai:
A local buffer would also help with speed, so your call. I think you should be able to push to this PR (or merge this and update later?)
However: making the buffer CRGBW is wasteful on RAM, W is never used and CRGB is used in the function anyway so lots of back and forth conversions, better make the buffer CRGB which converts only once (in sPC).

calling the variables hor/ver and the indices x/y even though they get flipped around (hor beeing actually vertical y coordinates and y actually being x direction if doing columns) may be more confusing than helpful.

blazoncek:
I hope I have not ruined it. This produced smallest footprint on my system.

wled00/FX.cpp
```diff
@@ -7469,40 +7469,54 @@ static const char _data_FX_MODE_2DDISTORTIONWAVES[] PROGMEM = "Distortion Waves@
...
-    int amount = ((int)noise3d[isRow ? i * cols : i] - 128) * amplitude;
-    int delta = abs(amount) >> 8;
+static void soapPixels(bool isRow, uint8_t *noise3d, CRGB *pixels) {
+  const int  cols = SEG_W;
+  const int  rows = SEG_H;
+  const auto XY   = [&](int x, int y) { return x + y * cols; };
+  const auto abs  = [](int x) { return x<0 ? -x : x; };
+  const int  tRC  = isRow ? rows : cols; // transpose if isRow
+  const int  tCR  = isRow ? cols : rows; // transpose if isRow
+  const int  amplitude = 2 * ((tCR >= 16) ? (tCR-8) : 8) / (1 + ((255 - SEGMENT.custom1) >> 5));
```
DedeHai, line 7479: I had >> 5 in the beginning but found it matches less with the original look on default setting. if you don't mind that, this is better.

edit: I tested this and it does change the look as well as the original logic, on high custom1 values it looks very weird. maybe you can keep my logic using max(1, xxxxx) as it will match better with the initial 
` amplitude = (cols >= 16) ? (cols-8)/8 : 1;`

wled00/FX.cpp
```diff
@@ -7469,40 +7469,54 @@ static const char _data_FX_MODE_2DDISTORTIONWAVES[] PROGMEM = "Distortion Waves@
...
+  const int  shift = 0; //(128 - SEGMENT.custom2)*2;
```
DedeHai, line 7480: original code was 
```
  int shiftX = 0; //(SEGMENT.custom1 - 128) / 4;
  int shiftY = 0; //(SEGMENT.custom2 - 128) / 4;
```
  so I am guessing you keep this as a reminder of the option.

wled00/FX.cpp
```diff
@@ -7469,40 +7469,54 @@ static const char _data_FX_MODE_2DDISTORTIONWAVES[] PROGMEM = "Distortion Waves@
...
-      else                            PixelA = ColorFromPalette(SEGPALETTE, ~noise3d[isRow ? XY(abs(zD), i) : XY(i, abs(zD))] * 3);
-      CRGB PixelB = CRGB::Black;
-      if ((zF >= 0) && (zF < colrow)) PixelB = isRow ? SEGMENT.getPixelColorXY(zF, i) : SEGMENT.getPixelColorXY(i, zF);
-      else                            PixelB = ColorFromPalette(SEGPALETTE, ~noise3d[isRow ? XY(abs(zF), i) : XY(i, abs(zF))] * 3);
+      int yA = abs(zD);
+      int yB = abs(zF);
+      int xA = i;
+      int xB = i;
+      if (isRow) {
+        std::swap(xA,yA);
```
DedeHai, line 7502: nice!

wled00/FX.cpp
```diff
@@ -7512,24 +7526,22 @@ uint16_t mode_2Dsoap() {
...
-  const auto XY = [&](int x, int y) { return (x%cols) + (y%rows) * cols; };
+  const auto XY = [&](int x, int y) { return x + y * cols; };
 
-  const size_t dataSize = SEGMENT.width() * SEGMENT.height() * sizeof(uint8_t); // prevent reallocation if mirrored or grouped
+  const size_t segSize = SEGMENT.width() * SEGMENT.height(); // prevent reallocation if mirrored or grouped
+  const size_t dataSize = segSize * (sizeof(uint8_t) + sizeof(CRGB)); // pixels and noise
   if (!SEGENV.allocateData(dataSize + sizeof(uint32_t)*3)) return mode_static(); //allocation failed
 
   uint8_t  *noise3d    = reinterpret_cast<uint8_t*>(SEGENV.data);
+  CRGB     *pixels     = reinterpret_cast<CRGB*>(SEGENV.data + segSize * sizeof(uint8_t));
```
DedeHai, line 7536: just FYI: this almost doubles the FPS!

wled00/FX.cpp
blazoncek: Go ahead, I did not have much time to compare now-and-then but from the top of my head this was "original" if custom1==0.
I would avoid max() or min() nd just try to find a sweet spot.

wled00/FX.cpp
blazoncek: I briefly experimented and saw minor difference so I wanted to keep shift as a reminder.

netmindz:
@coderabbitai review

wled00/FX.cpp
```diff
@@ -7468,42 +7468,86 @@ static const char _data_FX_MODE_2DDISTORTIONWAVES[] PROGMEM = "Distortion Waves@
...
+      ledsbuff[j] = (PixelA.nscale8(ease8InOutApprox(255 - fraction))) + (PixelB.nscale8(ease8InOutApprox(fraction)));
+    }
+    for (int j = 0; j < tCR; j++) {
+      CRGB c = ledsbuff[j];
+      if (isRow) std::swap(j,i);
+      SEGMENT.setPixelColorXY(i, j, pixels[XY(i,j)] = c);
+      if (isRow) std::swap(j,i);
+    }
+  }
+}
```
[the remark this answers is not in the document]
blazoncek: abs lambda produces smaller code than std::abs

wled00/FX.cpp
```diff
@@ -7472,11 +7472,11 @@ static const char _data_FX_MODE_2DDISTORTIONWAVES[] PROGMEM = "Distortion Waves@
 static void soapPixels(bool isRow, uint8_t *noise3d, CRGB *pixels) {
   const int  cols = SEG_W;
   const int  rows = SEG_H;
-  const auto XY   = [&](int x, int y) { return x + y * cols; };
+  const auto XY   = [&](int x, int y) { return (x%cols) + (y%rows) * cols; };
```
blazoncek, line 7475: I was too zealous with optimisations. Apparently. 😄 
Though I saw glitches but couldn't pin-point where I introduced them.

To keep optimisation, could `xA` & `xB` be modulo-ed instead?

DedeHai: probably (which is what I did in the orignal code, before bringing XY back) I will check.

blazoncek: This will do it:
```c++
      int yA = abs(zD)%tCR;
      int yB = abs(zF)%tCR;
```

wled00/FX.cpp
blazoncek: I find this the best compromise:
```c++
const int  amplitude = 1 + (tCR * (SEGMENT.custom3 >> 2))/8;
```
And is close to original source from @st3p40 while using slider to boost amplitude.
It may require default value of 4-8 to be equal to original.

NOTE: changed custom1 to custom3 as there is no need for 255 values.

wled00/FX.cpp
DedeHai: > Though I saw glitches but couldn't pin-point where I introduced them.

I saw the exact same glitches previously while optimizing but could not remember how I fixed them so I asked chatGPT to compare the two versions and the changed XY function popped up, which I had completely missed.

DedeHai:
that AI needs to be put on a leash ;)

that said, nothing more to add from my side, I briefly tested it on 32x32 and it looks good now. IMHO ready to merge.

PR #4549, opened 2025-02-12
Pixel-Based Effect Classes

TripleWhy:
Ignore some of the changes of this PR...

What I am trying to figure out with this PR is if you'd be interested in on overhaul of how effects work. I created a virtual base Effect class, and implemented 2 effects as implementations of that:

[wled00/effects/Effect.h](https://github.com/wled-dev/WLED/pull/4549/files#diff-50deaad265c6f363e2e19fea29e8e0b49edfece29cd6f00fcdfbafe9d4dac60b)
[wled00/effects/StaticEffect.h](https://github.com/wled-dev/WLED/pull/4549/files#diff-1c7a35da2e134021ba759990d60150321ccf0074ae3d9357d7dc38663d1a50d3)
[wled00/effects/PaletteEffect.h](https://github.com/wled-dev/WLED/pull/4549/files#diff-4df54429aaaa37469cde72e0db4514bd4a9e1aab4c0921fb42c13eae139c56ba)

Note that this effect class does not handle a segment as a whole like current effects do, but instead handles one pixel at a time. See how this is called in [wled00/FX_fcn.cpp](https://github.com/wled-dev/WLED/pull/4549/files#diff-f109c5828073461346e40af85d68967e51a37b8353f8f228ae720dc54fa64550R1536) `WS2812FX::service()`.

Advantages of this approach: 

- not every effect needs to understand the inner workings of how segments work
- when using sparse setups, such as matrices where not all pixels map to physical LEDs, this could potentially safe some work. (Or at least implementing this kind of saving would require one change to `service` rather than requiring changing of all effects.)
- theoretically, this design should work will with potential future sparse 3D setups (where instead of implementing dense LED cubes, you throw a bunch of LEDs into space, then map their 3D positions)
- things like the 2D expansion that is currently used in the palette effect only could/would have to be implemented globally for all effects rather than once per effect
- EDIT: Another potential advantage might be that this design would allow effects being layered on top of each other without losing information due to brightness being applied immediately (however I don't think this would work for effects that use color values from previous frames)

Disadvantages:

- theoretically one virtual function call per pixel. (A good compiler should be able to resolve the virtual function once per `service()` call instead of per pixel iteration, but c++ provides no way of enforcing that.)
- currently, this code relies on c++17 and therefore only compiles on esp 8266

Outlook: currently, I only implemented fading transitions (untested). Theoretically, it would be possible to use an extended version of this Effect class to handle transitions, so that there would only have to be one code base for effects and transitions.

What do you think? Are you interested in this kind of work? Should I invest some time into advancing it?

## Summary by CodeRabbit

* **Refactor**
  * Reimplemented effect system architecture with enhanced state management and rendering pipeline.
  * Improved effect-to-effect transition handling and ownership mechanics.
  * Updated Solid, Palette, Bouncing Balls, and Tetrix effects with revised implementations featuring improved processing logic.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -13,6 +13,9 @@
 #include "wled.h"
 #include "FX.h"
 #include "fcn_declare.h"
+#include "effects/StaticEffect.h"
+#include "effects/PaletteEffect.h"
+#include <memory>
 
 #if !(defined(WLED_DISABLE_PARTICLESYSTEM2D) && defined(WLED_DISABLE_PARTICLESYSTEM1D))
   #include "FXparticleSystem.h"
@@ -863,7 +866,7 @@ static uint16_t chase(uint32_t color1, uint32_t color2, uint32_t color3, bool do
   uint16_t counter = strip.now * ((SEGMENT.speed >> 2) + 1);
   uint16_t a = (counter * SEGLEN) >> 16;
 
-  bool chase_random = (SEGMENT.mode == FX_MODE_CHASE_RANDOM);
+  bool chase_random = (SEGMENT.getEffectId() == FX_MODE_CHASE_RANDOM);
   if (chase_random) {
     if (a < SEGENV.step) //we hit the start again, choose new color for Chase random
     {
@@ -10127,36 +10130,37 @@ static const char _data_RESERVED[] PROGMEM = "RSVD";
 // use id==255 to find unallocated gaps (with "Reserved" data string)
 // if vector size() is smaller than id (single) data is appended at the end (regardless of id)
 // return the actual id used for the effect or 255 if the add failed.
-uint8_t WS2812FX::addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name) {
-  if (id == 255) { // find empty slot
-    for (size_t i=1; i<_mode.size(); i++) if (_modeData[i] == _data_RESERVED) { id = i; break; }
+uint8_t WS2812FX::addEffect(std::unique_ptr<EffectFactory>&& factory) {
+  uint8_t id = factory->getEffectId();
+  if (id == 255u) { // find empty slot
+    for (size_t i=1; i<_effectFactories.size(); i++) if (_effectFactories[i] == nullptr) { id = i; break; }
   }
-  if (id < _mode.size()) {
-    if (_modeData[id] != _data_RESERVED) return 255; // do not overwrite an already added effect
-    _mode[id]     = mode_fn;
-    _modeData[id] = mode_name;
+  for (size_t i = _effectFactories.size(); i < id; ++i) {
+    _effectFactories.push_back(nullptr);
+  }
+  if (id < _effectFactories.size()) {
+    if (_effectFactories[id] != nullptr) return 255; // do not overwrite an already added effect
+    _effectFactories[id] = std::move(factory);
     return id;
-  } else if (_mode.size() < 255) { // 255 is reserved for indicating the effect wasn't added
-    _mode.push_back(mode_fn);
-    _modeData.push_back(mode_name);
-    if (_modeCount < _mode.size()) _modeCount++;
-    return _mode.size() - 1;
+  } else if (_effectFactories.size() < 255) { // 255 is reserved for indicating the effect wasn't added
+    _effectFactories.push_back(std::move(factory));
+    return _effectFactories.size() - 1;
   } else {
-    return 255; // The vector is full so return 255
+    return 255u; // The vector is full so return 255
   }
 }
 
-void WS2812FX::setupEffectData() {
+void WS2812FX::setupEffectData(size_t modeCount) {
   // Solid must be first! (assuming vector is empty upon call to setup)
-  _mode.push_back(&mode_static);
-  _modeData.push_back(_data_FX_MODE_STATIC);
+  addEffect(std::make_unique<EffectFactory>(StaticEffect::effectInformation));
+  addEffect(std::make_unique<EffectFactory>(PaletteEffect::effectInformation));
   // fill reserved word in case there will be any gaps in the array
-  for (size_t i=1; i<_modeCount; i++) {
-    _mode.push_back(&mode_static);
-    _modeData.push_back(_data_RESERVED);
+  for (size_t i=1; i<modeCount; i++) {
+    _effectFactories.push_back(nullptr);
   }
   // now replace all pre-allocated effects
   // --- 1D non-audio effects ---
+/*
   addEffect(FX_MODE_BLINK, &mode_blink, _data_FX_MODE_BLINK);
   addEffect(FX_MODE_BREATH, &mode_breath, _data_FX_MODE_BREATH);
   addEffect(FX_MODE_COLOR_WIPE, &mode_color_wipe, _data_FX_MODE_COLOR_WIPE);
@@ -10387,5 +10391,6 @@ addEffect(FX_MODE_PS1DGEQ, &mode_particle1DGEQ, _data_FX_MODE_PS_1D_GEQ);
 addEffect(FX_MODE_PSFIRE1D, &mode_particleFire1D, _data_FX_MODE_PS_FIRE1D);
 addEffect(FX_MODE_PS1DSONICSTREAM, &mode_particle1Dsonicstream, _data_FX_MODE_PS_SONICSTREAM);
 #endif // WLED_DISABLE_PARTICLESYSTEM1D
+*/
 
 }
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -13,8 +13,9 @@
 #include "wled.h"
 #include "FX.h"
 #include "fcn_declare.h"
-#include "effects/StaticEffect.h"
+#include "effects/BouncingBallsEffect.h"
 #include "effects/PaletteEffect.h"
+#include "effects/StaticEffect.h"
 #include <memory>
 
 #if !(defined(WLED_DISABLE_PARTICLESYSTEM2D) && defined(WLED_DISABLE_PARTICLESYSTEM1D))
@@ -10154,6 +10155,7 @@ void WS2812FX::setupEffectData(size_t modeCount) {
   // Solid must be first! (assuming vector is empty upon call to setup)
   addEffect(std::make_unique<EffectFactory>(StaticEffect::effectInformation));
   addEffect(std::make_unique<EffectFactory>(PaletteEffect::effectInformation));
+  addEffect(std::make_unique<EffectFactory>(BouncingBallsEffect::effectInformation));
   // fill reserved word in case there will be any gaps in the array
   for (size_t i=1; i<modeCount; i++) {
     _effectFactories.push_back(nullptr);
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -16,6 +16,7 @@
 #include "effects/BouncingBallsEffect.h"
 #include "effects/PaletteEffect.h"
 #include "effects/StaticEffect.h"
+#include "effects/TetrixEffect.h"
 #include <memory>
 
 #if !(defined(WLED_DISABLE_PARTICLESYSTEM2D) && defined(WLED_DISABLE_PARTICLESYSTEM1D))
@@ -10155,6 +10156,7 @@ void WS2812FX::setupEffectData(size_t modeCount) {
   // Solid must be first! (assuming vector is empty upon call to setup)
   addEffect(std::make_unique<EffectFactory>(StaticEffect::effectInformation));
   addEffect(std::make_unique<EffectFactory>(PaletteEffect::effectInformation));
+  addEffect(std::make_unique<EffectFactory>(TetrixEffect::effectInformation));
   addEffect(std::make_unique<EffectFactory>(BouncingBallsEffect::effectInformation));
   // fill reserved word in case there will be any gaps in the array
   for (size_t i=1; i<modeCount; i++) {
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -628,7 +628,6 @@ typedef struct Segment {
     Segment &setMode(uint8_t fx, bool loadDefaults = false);
     Segment &setPalette(uint8_t pal);
     Segment &setName(const char* name);
-    uint8_t differs(const Segment& b) const;
     void    refreshLightCapabilities();
 
     // runtime data functions
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -882,7 +882,7 @@ class WS2812FX {  // 96 bytes
     inline void trigger()                                     { _triggered = true; }  // Forces the next frame to be computed on all active segments.
     inline void setShowCallback(show_callback cb)             { _callback = cb; }
     inline void setTransition(uint16_t t)                     { _transitionDur = t; } // sets transition time (in ms)
-    inline void appendSegment(const Segment &seg = Segment()) { if (_segments.size() < getMaxSegments()) _segments.push_back(seg); }
+    inline void appendSegment(Segment &&seg = Segment())      { if (_segments.size() < getMaxSegments()) _segments.push_back(std::move(seg)); }
     inline void suspend()                                     { _suspend = true; }    // will suspend (and canacel) strip.service() execution
     inline void resume()                                      { _suspend = false; }   // will resume strip.service() execution
 
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -21,6 +21,7 @@
 
 #include "const.h"
 #include "bus_manager.h"
+#include "effects/Effect.h"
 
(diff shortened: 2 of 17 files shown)


DedeHai:
I like the base idea of only calculating pixels that actually exist, as you say, big advantage for mapped setups. 

- how does this impact code size and execution speed for normal setups?
- how would you apply blurring which relies on neighbouring pixels?

TripleWhy:
I didn't produce a clean version that only contains the effect changes, so it's a bit hard to say. But the FPS are indeed a bit lower than on main, which is what I would expect because of the virtual functions and no loops that can be optimized as a whole per effect. I believe it opens up more room for other optimizations though.

Test: 16x16 dense 2D matrix.

This PR:
38 FPS
RAM:   [======    ]  56.7% (used 46464 bytes from 81920 bytes)
Flash: [=======   ]  73.7% (used 770055 bytes from 1044464 bytes)

Main 2d6ad41ed43a7a7d854b88a6d93424a6f4836737:
47 FPS
RAM:   [======    ]  57.3% (used 46964 bytes from 81920 bytes)
Flash: [========  ]  84.2% (used 879751 bytes from 1044464 bytes)

Blurring:
What is blurring, how is it currently implemented? Where can I look at the code to see some?

If it is just part of some effects, I guess it would continue working the way it did before? If it requires reading the current state of the pixels, that is already part of this PR: The effect function has x, y, and current color as inputs and outputs one new color.

DedeHai:
that is a **significant** drop in FPS, unless this can be optimized, its a no-go.

look at `void Segment::blur2D()` for example. there are other functions to manipulate pixels in that file as well that need to work.

blazoncek:
Thanks.
This is a hefty change I'm afraid nobody would want to undertake (to reimplement all existing effects if that is the goal).

Regarding "segment workings awareness": I've had a lengthy discussion of pros and cons of the approach PixelBlaze has taken in the past (and why I chose not to go that way with WLED). Some things are easy to implement the way PixelBlaze does it, some are nearly impossible. So the best would be to extend segment API to allow both approaches. My 1st attempt was to create a `setPixelColor()` that takes relative coordinate (range 0-1, but it can also be -1-1) which would then require appropriate *mapping* to map certain coordinate into actual pixel drawing. This would also allow sparse set-ups. I tried but in the end abandoned the attempt as it would require writing a new effect that would use it.
The issue (I see) is how would you mix regular/legacy and new/coordinate-agnostic effects or. to put it the other way around how would you mix mappings?

If you are unaware of PixelBlaze, please Google it. It has some nice approaches (including sparse set-ups) but is unfortunately a closed source project.

blazoncek:
> Considering @blazoncek's segment blending

According to @DedeHai the speed increase with blending approach is about 10%. As it does away with traversing bus logic.
This (segment layering/blending) alone does not prohibit any change in how effect functions are created. However encapsulating them into class(es) may prove challenging to use them in usermods or other custom creations (like overlays, etc) but I am not very well versed in C++ to know for sure (see work by @willmmiles for converting usermods into libraries).

As mentioned earlier I'm afraid who will take on the effort of changing each and every effect out of about 200+ currently available.

But if we put that aside my main question is "what was the driving force for this change"? If I understand implementation correctly effect no longer renders entire "canvas" in a single call but only returns a single pixel color for supplied coordinates (`WS2812FX::service()` being responsible to call effect function for each pixel in segment). Correct?
Except for the (now redundant) calls to `SEGMENT.virtualXXXXX()` I do not see any real difference. Effect will still need to do something with X and Y but without knowing what max X and max Y are. Hence it will not be able to scale an image for example or draw rectangle on the border of frame.

I've considered PixelBlaze approach back in 2022 which takes a similar approach as this PR for some of its "patterns" but ultimately decided it would be too much hassle to modify every effect and the benefit would be minor (as it currently stands). **If someone could figure out the approach to allow two different APIs to coexist (i.e. old where effect draws entire canvas in a single call and new where it has no knowledge of canvas but responds to *virtual* (X,Y) coordinates) that would be the best IMO.**

> Are there effects that use the pixel color from the previous frame? What do they do with it?

There are plenty. Especially 2D effects that do a lot of pixel moving (i.e. shift all pixels left, etc). Some use old pixel values to *evolve* pixels over time. If you want to omit that then effect will need to store such information elsewhere. Making rewrite even more difficult.

TripleWhy:
> According to @DedeHai the speed increase with blending approach is about 10%. As it does away with traversing bus logic.

Expect another uplift for every call to getPixelColor that blending can eliminate. In my version here where I went from one call per pixel to zero calls, the improvement was ~28% in FPS.

> However encapsulating them into class(es) may prove challenging to use them in usermods or other custom creations (like overlays, etc)

Well that should be about the same effect API change as it is for the main effects. IIRC usermods are unmaintained though...  
Anyway without breaking things, options for progress are limited. I waited to introduce this concept until after 0.15 was released; I don't see an issue with introducing major changes when you start developing a new version.

> As mentioned earlier I'm afraid who will take on the effort of changing each and every effect out of about 200+ currently available.

Don't worry about that, I'd do it if you would merge such a change. But its a bit of work, so that's why I'm asking if you want this at all before doing it.

> If I understand implementation correctly effect no longer renders entire "canvas" in a single call but only returns a single pixel color for supplied coordinates (WS2812FX::service() being responsible to call effect function for each pixel in segment). Correct?

Correct.

> Except for the (now redundant) calls to SEGMENT.virtualXXXXX() I do not see any real difference.

Well first of all it avoids some code repetitions. And it enables some features to be implemented globally, rather than once per effect. It also removes the check and adjust each and every single one of the effects if you make some change (like making indexes unsigned, implementing effect layering, or changing the API for setting and getting pixel colors). I'm sure you remember the last time you had to do that, but you shouldn't have to.

In the long run, it would make a lot of sense to iterate physical LEDs and map segments to it rather than the other way round as it is now.
- That would avoid unnecessary work for sparse setups.
- That would enable non-grid mapped and 3D setups. No longer do you have to iterate an infinite space, but a fixed number of LEDs.
- It would allow information to be passed from one effect to the next layered one without losing information, due to applied and un-applied brightness etc. (this is what `oldColor` was intended for).
- Segment blending would be something like 1 - 5 lines of code (plus blend math functions/lambdas and configuration code).

> Effect will still need to do something with X and Y but without knowing what max X and max Y are.

Some effects need that, yes. There are solutions to that, for example:

- Keep everything as it is, the effects read the dimensions from globals at the moment, keeping that wouldn't break anything.
- Pass the max values to the effect as well, in addition to x and y.
- Use relative coordinates.

> **If someone could figure out the approach to allow two different APIs to coexist (i.e. old where effect draws entire canvas in a single call and new where it has no knowledge of canvas but responds to virtual (X,Y) coordinates) that would be the best IMO.**

That's simple: Keep everything as it is in main, and add a new (full canvas) effect function that does what I do in this POC. However, it would lose pretty much all of the advantages, as it would not enable most of (any of?) the refactoring possibilities discussed in this thread.

>> Are there effects that use the pixel color from the previous frame? What do they do with it?
>
> There are plenty. Especially 2D effects that do a lot of pixel moving (i.e. shift all pixels left, etc). Some use old pixel values to evolve pixels over time.

Well that doesn't sound safe... especially when layering or transitions are involved.


> "what was the driving force for this change"?

Well to be perfectly honest it's just that the current state of the code base is so poor and I would like to improve it, which would also make working on it less hard. I'm an idealist, don't question my motives 😅

I listed a few advantages in the PR's description, but here are a few more :)  
(I wrote this before I wrote most of the above, so there are some repetitions)

- As you identified correctly, relative coordinates (be it floats or scaled integers) would be possible with this, allowing for dimension-agnostic effects.
- Dimension-agnostic effects could be used to achieve the anamorphic mode switch of the palette effect globally instead of per effect, in addition to other mentioned advantages (sparse setups, 3d setups, 2d expansion (ok that one is enabled by the api change in general)).
- getPixelColor would be unambiguous and return the previous frame's color, not the previous effect's color. It would still only return the final color after blending, not the current effect's previous color though.
- Avoids information loss due to segment brightness. (see above)
- Effect transitions (the ones disabled with WLED_DISABLE_MODE_BLEND) would be much simpler to implement
- Segment layering (the one Blaž is(?) working on) would be easy to implement

DedeHai:
> Don't worry about that, I'd do it if you would merge such a change. But its a bit of work, so that's why I'm asking if you want this at all before doing it.

I have updated/migrated quite a few effects. I must admit I did not yet look at what your code does exactly (its a bit above my C++ understanding) but I can tell you that migrating an effect can take several hours, so you need to be aware that you are volunteering potentially hundreds of hours of tedious work in migrating and testing. If you think its worth it, also consider the particle system is now part of WLED and needs handling of this.

blazoncek:
There are two nice examples that may prove difficult to implement with *new* system: Tetrix and Bouncing Balls.
In the same category are all temporal effects (those that rely on time instead of place).

TripleWhy:
I re-implemented bouncing balls with the new approach, and took the opportunity to imagine what such an effect would look like in a dimension-agnostic world.

As for tetrix, and possibly other effects that do not iterate the pixel space and set each pixel in their main loop, but instead run their logic and then determine pixels they want to modify independently, or that store information in pixels: It's always possible to add a pixel buffer to those effects that they can work on in `nextFrame`, buffer can then be read and returned pixel by pixel later. According to @blazoncek there is more than enough RAM available for that.

Pixel buffers would also have the advantage that their dimensions can be independent of the display dimensions, you can then do all sorts of transformations with that during `getPixelColor`. For example: How wide do you really need your tetrix field to be? maybe at some point it makes sense to scale it up in order to have larger blocks that fill more than one pixel.

blazoncek:
Question regarding effect transitions or overlapping segments: How do you intend to tackle those? Assuming you can't read other segment's pixels. BTW do not forget that (currently) `getPixelColor()` returns modified output buffer's pixel value (restored as close as possible). By output buffer I mean NeoPixelBus' buffer.

I see that you have moved temporal property of the effect into nextFrame() method which is clever.

TripleWhy:
Alright, here is the Tetrix effect, with its own pixel buffer. The effect code itself is almost unaltered, just moved. Go ahead and diff them! This approach should work for all effects, so this is probably as close as this proposal will get to supporting both APIs, without giving up all of the advantages.

> Question regarding effect transitions or overlapping segments: How do you intend to tackle those? Assuming you can't read other segment's pixels. BTW do not forget that (currently) `getPixelColor()` returns modified output buffer's pixel value (restored as close as possible). By output buffer I mean NeoPixelBus' buffer.

layering: iterate physical LEDs, figure out a number of segments they belong to. Iterate them in order, and let each effect process one pixel, passing the getPixelColor result to the first effect, and the result of the effect ("current color") to the next effect. Was the original pan here. with gimp-style blending: don't pass colors to the effects, but apply blending functions instead: `current color = blend(current color, effect result)`.

About getPixelColor: I have thought about that quite a bit, and I believe that that is simply something that shouldn't be accessible to effects. It's lossy and they can never be sure what they get is even related to what they put there before. Instead, they should just keep everything they need in memory. Apparently there is now enough RAM for that.

transitions: The same as layering, really. Ideally in the same iteration pass. If it's a transition that uses either one effect or the other, just do that during iteration, no need to execute both. If its something like blending, execute both, blend them, use that as the result. I used to have a POC for that, but I believe its not in this PR anymore.

About your concerns on memory ownership and cleanup:
Not an issue. It simply works through the magic of unique_ptr. An effect class gets created and calls its constructor when you assign the effect to a segment (for example when the segment is created or when switching effects), and it gets deleted when you assign another effect or delete the segment. Then the effect destructor gets called and cleans everything up.

> I see that you have moved temporal property of the effect into nextFrame() method which is clever.

Well yes, separating nextFrame and getPixelColor is the fundamental idea here

blazoncek:
Good. It is a shame that you force-pushed, all in-line comments are lost due to that.

I do believe that creating effect (base) class is the correct way forward, however I'm not so sure that coordinate based effect calls are viable for every existing effect (at least not without some cumbersome or convoluted approach taken). Consider PS effects with hundreds of particles for instance. But perhaps someone can figure out that too.

I urge @DedeHai, @netmindz and @softhack007 to take a look and have a say.

As for me, until *all current effects* are modified and implemented with this new approach (or something entirely different) I will refrain from additional comments and code examination. I will continue working on segment layering/improving effect transitions assuming segment copy will still be used.

TripleWhy:
> It is a shame that you force-pushed, all in-line comments are lost due to that.

Crap. I didn't realize there were any.

> As for me, until _all current effects_ are modified and implemented with this new approach (or something entirely different) I will refrain from additional comments and code examination. I will continue working on segment layering/improving effect transitions assuming segment copy will still be used.

Well, that's the reason this PR exists: Should I go and modify _all current effects_, or won't that be merged anyway?

willmmiles:
First: I am responding mostly to the discussion - I have not yet taken the time to review the implementation details in depth.

As I understand things, there are a number of semi-related changes being proposed in this PR:
1. Formalizing the available FX mode list as a single array of objects, instead of two matched but independent arrays ('`mode` and `_modeData`).
2. Switching to storing FX specific state inside an object instance for each Segment; this permits structured setup and teardown.
3. Move the mode update render function from the "mode list" to a virtual call on the object instance.  ('nextFrame`).
4. Adding a "sampling" layer to sample from the rendered frame to the output pixels instead of rendering every pixel.  The intent is to offer improved performance for sparse setups.

Overall thoughts on those changes:
1. This is a good idea, and we should do it.  I am particularly excited as the link-time list assembly implementation from #4480 could be used to implement `addEffect()` at build time, eliminating the need for keeping vectors in RAM altogether.
2. This is also a good idea, and we should do it.  If it was up to me, I'd implement this in the first form as shallow wrapper around the existing FX functions without making any changes to the function code; and then migrate out common features to composable logic.  In particular, having a per-FX setup dovetails nicely with #4550 : the segment frame buffer (if one is needed) can be allocated and managed by the FX state class.
3. This is probably a net win, though I'm less condfident - I've been surprised by the compiler before.
4. I think the key is that FX should be *permitted*, but not *required* to be implemented in terms of a 'pull' in the blending stage.  The key reason comes back to performance: any FX that is most efficiently implemented by rendering to a flat buffer will be substantially penalized when blending to the output LEDs by the indirect call on every pixel.  There is a significant performance advantage to inlining the buffer lookup.  In practice, I think that that, instead of `getPixelColor()`, the key indirect call probably should be something closer to the interior loop of [`blendSegment()`](https://github.com/blazoncek/WLED/blob/dd1031b5c1cb404668dbbe687c5d0eabcd08dc58/wled00/FX_fcn.cpp#L1407C1-L1407C63).  This would allow those FX using a pixel buffer to share a common, optimized blending call; while those that don't need a whole pixel buffer are permitted to do something faster or simpler.  It also means that there's a quick and working shim layer for existing FX code, and any optimization that are applicable to only some FX can be added later.

Broadly: I think the overall goals are good, though I haven't reviewed the proposed implementation yet.  I strongly feel we need to merge some version of #4550 ASAP to really unlock some of the benefits of heading in this direction.

softhack007:
@coderabbitai please re-review this PR

Address these additional aspects in your review summary:
* specific pro's and con's of the suggested solution
* an assessment of user-visible impact
* memory usage impact
* a prioritized list of suggested improvements

softhack007:
I currently lean towards closing this PR, both due to inactivity and for being a very experimental idea that will need much more effort to understand if it adds value (in terms of flash size, speed, or code clarity). 

The "LazyColor::getColor()" idea seems to be inefficient in WLED 16.0.0 - all segments now use their own pixel buffer array, so I don't expect that any speed can be gained with lazy evaluation.

@DedeHai @willmmiles what do you think?

DedeHai:
while it is an interesting idea, this PR is a proof of concept only, hence the draft. 
The main advantage I see is with sparse setups. As those are an edge case additional RAM use or slow-down of the rendering pipeline would be a show stopper for me.
Also the fact that it needs a "double API" approach - some FX render per pixel, others still require a buffer to render to does not sound like a big advantage.

willmmiles:
I think it's probably safe to close this PR at this point.  We've learned from these ideas and we'll likely ultimately land on something similar; addressing the details of this specific implementation is disruptive enough and far enough down our priority list that we're not really in a good place to commit the necessary time at this point.

To summarize:
- We will eventually want to arrange that FX can manage their state storage via C++ classes; this has already been a pain point a couple of times.  However it must not be mandatory -- for many reasons (compatibility chief among them) we also want to support stateless effects implemented as pure functions.  (#5294 is tracking this, though I have yet to get around to a detailed response.)
- IMO, I wouldn't rule out a "sampling" architecture for compositing FX as an alternative to the current "overdraw" architecture at some point.  There will always be some FX that are easiest to implement in the other paradigm.  Unfortunately our rendering core isn't organized in a way that supports this well ... yet.  (#5464 is beginning to point the way: breaking out the composite operation to optimized paths for different cases.)  The key distinction is that any abstraction over composition will need to be handled at the Segment, not the pixel level.

PR #4566, opened 2025-02-17
Added Chase Race Effect to WLED

drdome:
Added a new effect, "Chase Race," based on the existing Chase effect.

Chase Race is designed for automotive racing displays, simulating three colored "cars" chasing each other in LED sequences. Unlike Chase, it accepts three input colors, with a fixed black background. This can be integrated with Home Assistant, allowing real-time race data to dynamically adjust colors and sequencing. Initial testing has been conducted using F1 race data.

## Summary by CodeRabbit

- **New Features**
  - Introduced a dynamic new LED effect, "Chase Race," featuring three racing colored bands separated by black.
  - Added additional LED animation modes—"Perlin Move," "Wavesins," "Flow Stripe," and "Blends"—for varied lighting experiences.
  - Enhanced audio reactive performance and overall effect responsiveness for a more immersive display.
- **Chores**
  - Updated project configuration to streamline the build environment, now targeting only `esp32dev`.


DedeHai:
thanks for your contribution.
the code seems a bit unfinished... 
there is the `do_palette` parameter that is unused, also making the effect a function and using another function to call it does only make sense if its used multiple times.
I used an AI to improve code size and adjusted parameters slightly. in your original version, there is overlapping if you set width to a high value.

consider this:
```

uint16_t chase_race(uint32_t color1, uint32_t color2, uint32_t color3) {
    uint16_t counter = strip.now * ((SEGMENT.speed >> 2) + 1);
    uint16_t a = (counter * SEGLEN) >> 16;
    
    // Use intensity setting to vary chase up to 1/2 string length
    unsigned size = 1 + ((SEGMENT.intensity * SEGLEN) >> 10);
    unsigned gap = 1 + size / 4;  // Smaller black band size as a fraction of color size
    
    // Calculate positions for each color band
    uint16_t positions[] = {
        a,                          // Start of color1
        (a + size) % SEGLEN,       // End of color1
        (a + size + gap) % SEGLEN, // Start of color2
        (a + size * 2 + gap) % SEGLEN, // End of color2
        (a + size * 2 + gap * 2) % SEGLEN, // Start of color3
        (a + size * 3 + gap * 2) % SEGLEN  // End of color3
    };
    
    // Define compact lambda for filling LED segments using modulo
    auto fillSegment = [](uint16_t start, uint16_t end, uint32_t color) {
        for (unsigned count = 0; count < SEGLEN && count < end - start + (end < start ? SEGLEN : 0); count++) {
            SEGMENT.setPixelColor((start + count) % SEGLEN, color);
        }
    };
    
    // Set background to black
    SEGMENT.fill(0);
    
    // Fill the three color segments
    fillSegment(positions[0], positions[1], color1);
    fillSegment(positions[2], positions[3], color2);
    fillSegment(positions[4], positions[5], color3);
    
    return FRAMETIME;
}
```

netmindz:
This feels a bit niche to be a core effect and I think should be a usermod instead

blazoncek:
@DedeHai cherry-pick from my fork. Already done that part.

netmindz:
Ah, so the code doesn't really do what the description says, it's only when you also basically provide the other half of the functionality with HA that it works. I should have read the code before commenting

wled00/FX.cpp
```diff
@@ -934,6 +977,14 @@ uint16_t mode_chase_color(void) {
 }
 static const char _data_FX_MODE_CHASE_COLOR[] PROGMEM = "Chase@!,Width;!,!,!;!";
 
+/* 
+* Chase Race with 3 color strips 
+*/
+uint16_t mode_chase_race()
```
DedeHai, line 983: remove alias function, it serves no purpose.

platformio.ini
```diff
@@ -10,8 +10,8 @@
 # ------------------------------------------------------------------------------
 
 # CI/release binaries
-default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, nodemcuv2_compat, esp8266_2m_compat, esp01_1m_full_compat, esp32dev, esp32dev_V4, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_16MB_opi, esp32s3dev_8MB_opi, esp32s3_4M_qspi, esp32_wrover
-
+#default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, nodemcuv2_compat, esp8266_2m_compat, esp01_1m_full_compat, esp32dev, esp32dev_V4, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_16MB_opi, esp32s3dev_8MB_opi, esp32s3_4M_qspi, esp32_wrover
+default_envs = esp32dev
```
DedeHai, line 14: revert any changes to platform.ini

drdome:
Made/removed changes to platformio.ini file.  Updated fx.cpp to remove unneeded alias per DedeHai

wled00/FX.cpp
```diff
@@ -933,6 +976,14 @@ uint16_t mode_chase_color(void) {
 }
 static const char _data_FX_MODE_CHASE_COLOR[] PROGMEM = "Chase@!,Width;!,!,!;!";
 
+/*
+ * Chase Race with 3 color strips
+ */
+uint16_t chase_race()
```
DedeHai, line 982: this is the alias function, it does nothing but call the other one.

platformio.ini
```diff
@@ -11,7 +11,6 @@
 
 # CI/release binaries
 default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, nodemcuv2_compat, esp8266_2m_compat, esp01_1m_full_compat, esp32dev, esp32dev_V4, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_16MB_opi, esp32s3dev_8MB_opi, esp32s3_4M_qspi, esp32_wrover, usermods
-
```
DedeHai, line 14: remove change

drdome:
Can we add this?  It's been open a while.

drdome:
Hi @Aircoookie, @Christian-Schiffer, and @DedeHai,

I hope you’re all doing well! I wanted to check in on this PR (#4566). I’ve addressed the earlier feedback from @DedeHai and made the requested updates to the effect. It’s been a little while, so I just wanted to see if there’s anything else needed before it can move forward.  

I’m happy to make additional tweaks, add documentation, or handle testing adjustments if needed. It has been a while so I could consider reworking the code, but if I were to do I wanted to make sure it will be considered for inclusion.

Thanks again for your time and all the work you put into maintaining WLED — I really appreciate the opportunity to contribute.

Best regards,
Drdome

willmmiles:
Sorry for the delay.   Unfortunately, right now we're in the working on a long term plan for managing FX ID code assignment, as our current 8-bit space is filling up, and many existing integrations and packets are limited to a variable of that size.  What this means is that any new IDs assigned in the core code now create additional backwards compatibility challenge for 0.16 and beyond.  At this point I think we'd like to hold off on adding many new FX to the core while we get that sorted.

@netmindz suggested above that this effect could be implemented as a usermod, so it wouldn't be limited by the core FX mangement challenges.  The `usermods/user_fx` folder has a template for implementing one this way.  Would you mind moving the code out to a `user_fx_chase_race` component?  I'd be happy to sponsor inclusion in that way.

DedeHai:
I agree, since this is a simple effect, it would be better to hold off with including it and move it to the user-FX UM for now.

also: the code is very much overcompilcated, you even define a "single LED version" but its not even a single LED effect, dont think that is userful. Here is a cleaner version that uses less than half the code size and from my quick check does exactly the same:

``` c++
uint16_t chase_race()
{
  if (SEGLEN<4) return FRAMETIME;

  unsigned counter = uint16_t(strip.now * ((SEGMENT.speed >> 2) + 1));
  unsigned a = uint16_t((counter * SEGLEN) >> 16);
  unsigned size = 1 + ((SEGMENT.intensity * SEGLEN) >> 10);
  unsigned gap  = 1 + size / 4;
  unsigned len  = size;

  SEGMENT.fill(0);
  for (int i = 0; i < 3; i++) {
    for (int j = 0; j < len; j++)
      SEGMENT.setPixelColor((a + j) % SEGLEN, SEGCOLOR(i));
    a += size + gap;
  }
  return FRAMETIME;
}
static const char _data_FX_MODE_CHASE_RACE[] PROGMEM = "Chase Race@!,Width;!,!,!;!";
```

PR #2413, opened 2021-12-12
FSEQ Playback from LittleFS and SD

fastbytes:
Credit goes to @constant-flow for the original idea and structure created for TPM2 playback!
https://github.com/Aircoookie/WLED/pull/2292

Adds support for playback of FSEQ files from xLights or Vixen
Includes support for both SD_MMC and SD via SPI (e.g. TTGO devices)
Uses FSEQFile class to prevent overlapping functions

NOTE: Created REALTIME_MODE_FSEQ using 10 instead of 9 (next available value) to keep compatibility with TPM2RECORD once both are merged

## Summary by CodeRabbit

* **New Features**
  * Added FSEQ animation recording playback with repeat/loop control and customizable LED range
  * Web editor now recognizes and supports FSEQ file editing
  * JSON API support for loading FSEQ recordings and managing playback behavior
  * Integrated playback state management into the main system loop

PR #4570, opened 2025-02-19
added support for fseq files

Andrejchrcek:
This PR adds a usermod for playing FSEQ files from xLights.  
- Added an "FSEQ" button to the main menu.  
- Implemented a basic file explorer for uploading, deleting, and playing FSEQ files.  

The feature is still in development, and further improvements are planned.  
Feedback and testing on different ESP32 boards are welcome!
Screenshot 2025-02-19 at 20 44 33

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "wled",
-  "version": "0.16.0-dev",
+  "version": "0.16.0-alpha",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "wled",
-      "version": "0.16.0-dev",
+      "version": "0.16.0-alpha",
       "license": "ISC",
       "dependencies": {
         "clean-css": "^5.3.3",
--- a/usermods/FSEQ/README.md
+++ b/usermods/FSEQ/README.md
@@ -1,0 +1,69 @@
+# SD & FSEQ Usermod for WLED
+
+This usermod adds support for playing FSEQ files from an SD card and provides a web interface for managing SD files and FSEQ playback. Follow the instructions below to install and use this usermod.
+
+---
+
+## Installation
+
+### 1. Modify `usermods_list.cpp`
+
+Add the following lines to register the FSEQ usermod. Make sure that you **do not** include any conflicting modules (e.g. `usermod_sd_card.h`):
+
+```cpp
+#ifdef USERMOD_FSEQ
+  #include "../usermods/FSEQ/usermod_fseq.h"
+#endif
+
+#ifdef USERMOD_FSEQ
+  UsermodManager::add(new UsermodFseq());
+#endif
+
+// #include "../usermods/sd_card/usermod_sd_card.h"
+
+
+### 2. Configure PlatformIO
+
+Add the following configuration to your platformio_override.ini (or platformio.ini) to enable the necessary build flags and define the usermod:
+
+
+[env:esp32dev_V4]
+build_flags = 
+  -D WLED_DEBUG
+  -D WLED_USE_SD_SPI
+  -D WLED_PIN_SCK=18    ; CLK
+  -D WLED_PIN_MISO=19   ; Data Out (POCI)
+  -D WLED_PIN_MOSI=23   ; Data In (PICO)
+  -D WLED_PIN_SS=5      ; Chip Select (CS)
+  -D USERMOD_FSEQ
+  -std=gnu++17
+
+
+
+### 3. Update the WLED Web Interface
+
+To integrate the new FSEQ functionality into the WLED UI, add a new button in your index.htm file. For example, insert the following button into the navigation area:
+
+<!-- New button for SD & FSEQ Manager -->
+<button onclick="window.location.href=getURL('/sd/ui');">
+  <i class="icons">&#xe0d2;</i>
+  <p class="tab-label">Fseq</p>
+</button>
+
+
+This button will take you to the SD & FSEQ Manager interface.
+
+
+
+###Usage
+	•	Web Interface:
+Access the SD & FSEQ Manager by clicking the Fseq button in the main UI. The interface allows you to view, upload, and delete SD card files as well as control FSEQ playback.
+	•	File Management:
+The SD file manager displays all files on the SD card. You can upload new files via the provided form and delete files using the red “Delete” button. 
+	•	FSEQ Playback:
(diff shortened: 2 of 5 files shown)


netmindz:
Now we have playback of gif files in the main code. Is there any chance this could be extended to support this file type as well or is there too little in common?

blazoncek:
@netmindz there is an old PR from @fastbytes #2413 and @constant-flow #2292 for FSEQ/TPM2 file playback we never got around to properly migrate/update for WLED.
These have nothing in common with GIF playback except perhaps SD card support.

I am really sorry those old PRs were left hanging but one man just couldn't do everything. I hope this fares better or old PRs are revived.

Andrejchrcek:
xLights can now discover WLED using the network discovery feature.
FPP Connect detects the WLED device and allows uploading FSEQ files directly.

Screenshot 2025-02-21 at 09 07 51

netmindz:
No worries if there isn't anything in common. I had a vague memory I'd seen a comment about GIF just being the first to be done then extend, but maybe that was just aspirational rather than any software architecture/design

netmindz:
> xLights can now discover WLED using the network discovery feature. FPP Connect detects the WLED device and allows uploading FSEQ files directly.

So someone presumable has already written FSEQ support in a fork somewhere if xLights has an upload option?

blazoncek:
Since I'm already getting tonnes of mail, let's join the fray.

What I'll veto is the change in UI. Please remove Fseq button from UI as this is a usermod.
Usermods (can) have their own HTTP handler (as you've shown) and you can get a pointer to it into Info dialog.

Andrejchrcek:
@coderabbitai review

Andrejchrcek:
I have tested the code, and it works fine for me. However, it would be great if someone could test it on different boards to ensure compatibility.

noman1228:
ESP32-DevkitV4 (wroom32-u) tested working
Freenove Wrover wrover-e) confirmed working
DoItDevKit1 - Confirmed Working
D32 - Confirmed Working
ESP32-TETH- Confirmed Working

PaoloTK:
I'm interested in testing this in production. I have a pretty aggressive use case, so can provide feedback on robustness hopefully. I have 4 Quinled-Dig-Octa Brainboards running around 2k pixels each. I would like to run FSEQ off an SD card on each board and keep the boards synced between each other over WiFi. How feasible is this currently?

Andrejchrcek:
Hi, thanks for your interest. Syncing FSEQ files over Wi‑Fi using Falcon Player (FPP) should work. I’ve built a custom remote that sends the same data as the FPP server. I can’t confirm 100% without testing, but it ought to work. Let me know if you run into any issues.

netmindz:
This PR needs to be updated to follow the new usermod setup that does not require changing `usermods_list.cpp` but does require the library.json

npike:
Pretty excited to see this PR as someone who has a lot of WLED devices deployed but is lately suffering from congestion on the 2.4ghz wifi. Thank you for your work on this @Andrejchrcek 

I tried this on a DigUno ESP32+AE+, configured as:

```
int8_t configPinSourceSelect = 5;
int8_t configPinSourceClock = 18;
int8_t configPinPoci = 19; 
int8_t configPinPico = 23;
```

This DigUno has a string of 5v ws2811 pixels on GPIO16.

**What works**
- Fseq button on main UI, when clicked loads a two tab html interface
- Uploading a fseq file seems to work, although there is no indication of upload progress.
- fseq tab shows all uploaded files - pressing "play" button does _something_ (see what doesn't work below)

**What doesn't seem to work**
- xlights/fpp connect doesn't seem to discover this device, even when manually adding it via IP address
- fpp doesn't seem to discover this device as a remote player
- Attempting to play an uploaded fseq file just produces random looking colors for about half a second, before returning to the previously selected WLED effect
- Periodically, WLED shows an unexpected live data message for a few seconds: "WLED is receiving live USB Adalight/TPM2 data"  (this is curious because nothing is connected to the USB port..). Sometimes this changes to "WLED is receiving live Hyperion data"

netmindz:
@Andrejchrcek can you either grant maintainers permission to your branch so we can apply that changes for being a 0.16 usermod or make the changes yourself please

netmindz:
As this is a pretty sizable bit of code, I think this probably should be an out of tree usermod, can you give some details on how to do this and how to then build please @willmmiles

willmmiles:
> As this is a pretty sizable bit of code, I think this probably should be an out of tree usermod, can you give some details on how to do this and how to then build please @willmmiles

I'm not sure I agree, actually -- rather I think it should be broken up in to smaller components.  I have not done an in-depth review, but to my cursory lookover, there are a number of features wrapped up here:

- Adding SD card support.  With the addition of GIF and video animation support in mainline, the ability to access external storage could be broadly useful.  This should be a module on its own, or outright integrated in to the main codebase, and at the very least made available to the gif player.
- Adding a web manager/viewer for animation files on storage.  This has substantial overlap with some of the things @DedeHai is working on, both in #4982 and to a lesser extent #4956.  The key features here should get integrated with the existing efforts.
- The FSEQ playback engine.  The current implementation here treats playback as a realtime data source; I think it might be better implemented as an effect, similar to the gif player, that can operate inside the segment mapping -- particularly given the 0.16 blending framework.
- The FPP UDP protocol handler.  Not much to say here, it is what it says on the tin

As a general-purpose integration with another LED control framework, I think this feature is a good candidate for mainline inclusion -- it's not as tied to someone's unique project as some other modules ;)   The things that make it bigger than most are IMO good ideas that we probably want for other things too.

blazoncek:
I fully agree with @willmmiles

DedeHai:
I concur. With the features in the pipeline (Matrix tool,a little further out the Video player as an (external) part of that) we should think of breaking this up into more paletable junks. Is @Andrejchrcek still around to work on this?

Danit2:
Hello.

First of all, thank you very much for this great project.

What about support for FSEQ files? Will this be included in version 16?

It would be great if WLED supported FSEQ files from the SD card and also synchronized with the FPP player.

As a cool alternative, it would also be nice if the FSEQ files could be stored directly on the ESP32. With a 16 MB ESP, there should be plenty of space available.

In my case, this would massively reduce the network load. I currently control my WLED controllers via DDP and WiFi, but I always have the problem that the controllers are not 100% synchronized. 

Thank you very much for your work.

Andrejchrcek:
This update introduces enhanced compatibility with the latest WLED version. Additionally, I've identified and fixed several minor bugs to improve overall stability and functionality.

I am currently considering adding a feature for file unzipping.

PR #4573, opened 2025-02-22
Consolidate classic FX

DedeHai:
This PR removes some of the classic FX "aliases" and moves them into the main FX to be used with checkmarks. 

cherry picked from @blazoncek's branch, updated, cleaned and tested.

Removed effects:
- Dynamic Smooth -> use Danamic with check3
- Candle Multi -> use Candle with check 3
- Solid Glitter -> use Glitter (alternative: PS Sparkler)
- Ripple Rainbow -> use Ripple with check 1
- Sinelon Dual -> use Sinelon with check 2
- Sinelon Rainbow -> use Sinelon with check 1
- Meteor Smooth -> use Meteor
- Scanner Dual -> use Scanner with check 3
- Running Dual -> use Running with with check 2
- Chase 2 ->  use Theater without check 3
- Dissolve Rnd -> use Dissolve with with check 1
- Saw -> use Running with with check 3
- Theater Rainbow ->  use Theater with check 
- Scan Dual -> use Scan with check 3

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -507,15 +507,15 @@ static const char _data_FX_MODE_THEATER_CHASE[] PROGMEM = "Theater@!,Gap size,,,
  * Running lights effect with smooth sine transition base.
  * Idea: Make the gap width controllable with a third slider in the future
  */
-static uint16_t running_base(bool saw) {
-  const bool dual = SEGMENT.check3;
+static uint16_t mode_running_lights() {
+  const bool dual = SEGMENT.check2;
   unsigned x_scale = SEGMENT.intensity >> 2;
   uint32_t counter = (strip.now * SEGMENT.speed) >> 9;
   const bool moving = SEGMENT.check1;
 
   for (unsigned i = 0; i < SEGLEN; i++) {
     unsigned a = i*x_scale - counter;
-    if (saw) {
+    if (SEGMENT.check3) { // Saw mode
       a &= 0xFF;
       if (a < 16)
       {
@@ -539,23 +539,7 @@ static uint16_t running_base(bool saw) {
 
   return FRAMETIME;
 }
-
-/*
- * Running lights effect with smooth sine transition.
- */
-uint16_t mode_running_lights(void) {
-  return running_base(false);
-}
-static const char _data_FX_MODE_RUNNING_LIGHTS[] PROGMEM = "Running@!,Wave width,,,,Animate palette,,Dual;!,!;!;;o1=0";
-
-
-/*
- * Running lights effect with sawtooth transition.
- */
-uint16_t mode_saw(void) {
-  return running_base(true);
-}
-static const char _data_FX_MODE_SAW[] PROGMEM = "Saw@!,Width,,,,Animate palette;!,!;!;;o1=0,o3=0";
+static const char _data_FX_MODE_RUNNING_LIGHTS[] PROGMEM = "Running@!,Width,,,,Animate palette,Dual,Saw;L,!,R;!";
 
 
 /*
@@ -1173,16 +1157,6 @@ uint16_t mode_larson_scanner(void) {
   return FRAMETIME;
 }
 static const char _data_FX_MODE_LARSON_SCANNER[] PROGMEM = "Scanner@!,Trail,Delay,,,Animate palette,Bi-delay,Dual;!,!,!;!;;m12=0,c1=0,o1=0,o3=0";
-
-/*
- * Creates two Larson scanners moving in opposite directions
- * Custom mode by Keith Lord: https://github.com/kitesurfer1404/WS2812FX/blob/master/src/custom/DualLarson.h
- */
-//uint16_t mode_dual_larson_scanner(void){
-//  SEGMENT.check3 = true;
-//  return mode_larson_scanner();
-//}
-//static const char _data_FX_MODE_DUAL_LARSON_SCANNER[] PROGMEM = "Scanner Dual@!,Trail,Delay,,,Animate palette,Bi-delay;!,!,!;!;;m12=0,c1=0,o3=1";
 
 
 /*
@@ -2306,76 +2280,59 @@ uint16_t mode_meteor() {
   byte* trail = SEGENV.data;
 
   const unsigned meteorSize = 1 + SEGLEN / 20; // 5%
-  unsigned counter = strip.now * ((SEGMENT.speed >> 2) +8);
-  uint16_t in = counter * SEGLEN >> 16;
-  const bool gradient = SEGMENT.check1;
+  uint16_t meteorstart;
+  if(meteorSmooth) meteorstart = map((SEGENV.step >> 6 & 0xFF), 0, 255, 0, SEGLEN -1);
+  else {
+    unsigned counter = strip.now * ((SEGMENT.speed >> 2) + 8);
+    meteorstart = (counter * SEGLEN) >> 16;
+  }
 
   const int max = SEGMENT.palette==5 || !SEGMENT.check1 ? 240 : 255;
   // fade all leds to colors[1] in LEDs one step
   for (unsigned i = 0; i < SEGLEN; i++) {
-    if (random8() <= 255 - SEGMENT.intensity) {
-      int meteorTrailDecay = 128 + hw_random8(127);
-      trail[i] = scale8(trail[i], meteorTrailDecay);
-      int index = trail[i];
-      int idx = 255;
-      int bri = SEGMENT.palette==35 || SEGMENT.palette==36 ? 255 : trail[i];
-      if (!gradient) {
-        idx = 0;
-        index = map(i,0,SEGLEN,0,max);
-        bri = trail[i];
+    uint32_t col;
+    if (hw_random8() <= 255 - SEGMENT.intensity) {
+      if(meteorSmooth) {
+        if (trail[i] > 0) {
+          int change = trail[i] + 4 - hw_random8(24); //change each time between -20 and +4
+          trail[i] = constrain(change, 0, max);
+        }
+        col = SEGMENT.check1 ? SEGMENT.color_from_palette(i, true, false, 0, trail[i]) : SEGMENT.color_from_palette(trail[i], false, true, 255);
       }
-      SEGMENT.setPixelColor(i, SEGMENT.color_from_palette(index, false, false, idx, bri)); // full brightness for Fire
+      else {
+        trail[i] = scale8(trail[i], 128 + hw_random8(127));
+        int index = trail[i];
+        int idx = 255;
+        int bri = SEGMENT.palette==35 || SEGMENT.palette==36 ? 255 : trail[i];
+        if (!SEGMENT.check1) {
+          idx = 0;
+          index = map(i,0,SEGLEN,0,max);
+          bri = trail[i];
+        }
+        col = SEGMENT.color_from_palette(index, false, false, idx, bri);  // full brightness for Fire
+      }
+      SEGMENT.setPixelColor(i, col);
     }
   }
 
   // draw meteor
   for (unsigned j = 0; j < meteorSize; j++) {
-    int index = (in + j) % SEGLEN;
-    int idx = 255;
-    int i = trail[index] = max;
-    if (!gradient) {
-      i = map(index,0,SEGLEN,0,max);
-      idx = 0;
+    unsigned index = (meteorstart + j) % SEGLEN;
+    if(meteorSmooth) {
+        trail[index] = max;
+        uint32_t col = SEGMENT.check1 ? SEGMENT.color_from_palette(index, true, false, 0, trail[index]) : SEGMENT.color_from_palette(trail[index], false, true, 255);
+        SEGMENT.setPixelColor(index, col);
     }
-    SEGMENT.setPixelColor(index, SEGMENT.color_from_palette(i, false, false, idx, 255)); // full brightness
-  }
-
-  return FRAMETIME;
-}
-static const char _data_FX_MODE_METEOR[] PROGMEM = "Meteor@!,Trail,,,,Gradient;!;!;1";
-
-
-// smooth meteor effect
-// send a meteor from begining to to the end of the strip with a trail that randomly decays.
-// adapted from https://www.tweaking4all.com/hardware/arduino/adruino-led-strip-effects/#LEDStripEffectMeteorRain
-uint16_t mode_meteor_smooth() {
-  if (SEGLEN <= 1) return mode_static();
-  if (!SEGENV.allocateData(SEGLEN)) return mode_static(); //allocation failed
-
-  byte* trail = SEGENV.data;
-
-  const unsigned meteorSize = 1+ SEGLEN / 20; // 5%
-  uint16_t in = map((SEGENV.step >> 6 & 0xFF), 0, 255, 0, SEGLEN -1);
-  const bool gradient = SEGMENT.check1;
-
-  const int max = SEGMENT.palette==5 || !gradient ? 240 : 255;
-  // fade all leds to colors[1] in LEDs one step
-  for (unsigned i = 0; i < SEGLEN; i++) {
-    if (/*trail[i] != 0 &&*/ hw_random8() <= 255 - SEGMENT.intensity) {
-      int change = trail[i] + 4 - hw_random8(24); //change each time between -20 and +4
-      trail[i] = constrain(change, 0, max);
-      SEGMENT.setPixelColor(i, gradient ? SEGMENT.color_from_palette(i, true, false, 0, trail[i]) : SEGMENT.color_from_palette(trail[i], false, true, 255));
+    else{
+      int idx = 255;
+      int i = trail[index] = max;
+      if (!SEGMENT.check1) {
+        i = map(index,0,SEGLEN,0,max);
+        idx = 0;
+      }
+      uint32_t col = SEGMENT.color_from_palette(i, false, false, idx, 255); // full brightness
+      SEGMENT.setPixelColor(index, col);
     }
-  }
-
-  // draw meteor
-  for (unsigned j = 0; j < meteorSize; j++) {
-    unsigned index = in + j;
-    if (index >= SEGLEN) {
-      index -= SEGLEN;
-    }
-    trail[index] = max;
-    SEGMENT.setPixelColor(index, gradient ? SEGMENT.color_from_palette(index, true, false, 0, trail[index]) : SEGMENT.color_from_palette(trail[index], false, true, 255));
   }
 
   SEGENV.step += SEGMENT.speed +1;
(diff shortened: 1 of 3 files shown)


netmindz:
I've move away from trying to track all changes in the changelog.md file, but for breaking changes like, it's probably our best place to record these changes, so it's then easy to find when we put out the release notes

blazoncek:
> I've move away from trying to track all changes in the changelog.md file

I usually did it once in a while by checking commit messages and writing down changes.

wled00/FX.cpp
```diff
@@ -563,47 +525,21 @@ static uint16_t running_base(bool saw, bool dual=false) {
...
-
-
-/*
- * Running lights effect with sawtooth transition.
- */
-uint16_t mode_saw(void) {
-  return running_base(true);
-}
-static const char _data_FX_MODE_SAW[] PROGMEM = "Saw@!,Width;!,!;!";
+static const char _data_FX_MODE_RUNNING_LIGHTS[] PROGMEM = "Running@!,Width,,,,Rainbow,Dual,Saw;L,!,R;!";
```
TripleWhy, line 542: The first checkbox doesn't produce any rainbows. In the code, the value is called `moving`, but it doesn't have anything to do with movement either. It's `color_from_palette`'s `wrap` argument (which isn't a super self-explanatory name to a user either...). Both the variable and the UI string should be renamed.

blazoncek: Moving in the instance of palette means if the palette itself is moving (i.e. hue of the palette is changing).
Some palettes wrap at the end and some don't. This "moving" parameter tells that palette should wrap as it is changing so that transition from ent to start is smooth.

Hmm. Not the best explanation, I know.

DedeHai: the reason it is named 'rainbow' is legacy: some FX use(d) this exactly for that: moving the palette, creating a rainbow effect when using rainbow palette (i.e. colowheel). I agree that is now a bit nonsensical since many palettes are available, I am open to suggestions for renaming all "move palette" checkmarks into something more meaningful. in other places, such behaviour is also called "move" or "shift".

DedeHai:
this has been pending for a while. are there any objections to removing consolidated FX? It is a breaking change to existing presets. IMHO 0.16 will bring so many new things (layered segments, PS, transitions) it seems like a good point in time to some de-cluttering of the effect list.

blazoncek:
I approve but I think other team members need to be ok with it.
After all, we'll get plenty of questions "where did XXX go?".

willmmiles:
Is it worth implementing a compatibility parser for old presets?  ie. a shim function that translates the old FX ids into new ones?

blazoncek:
Are you thinking of a separate application that will modify existing presets?

TripleWhy:
That could have been and continue to be useful tool, but I believe it's much too complex to be part of the embedded wled software. Not to forget that it can never drop any code, it will just grow larger over time. The user would also need to specify from which version of wled to which other version to convert, because that isn't saved in the preset itelf, and simply judging from IDs that are currently unused doesn't work, because they get re-used for new effects over time (at least that is what happened in the past).

Such a tool could do more than just replacing IDs. For example it could reset effect options that were unused and set arbitrarily in older versions, then later became used, to either their new defaults or to a value that closely matches the old behavior.

But it might also be a pain to keep up to date.

netmindz:
I suppose I'm uncertain from the description what we are trying to achieve.

If we think we have effects that are redundant because you can achieve the same result by just changing the parameters of another effect, then yes we should consolidate

Alternatively, if we are just looking to reduce the code size but maintain the same list of effect names, then we refactor so we maintain the effect but it's implementation is calling the common code with the right arguments set

blazoncek:
It is the former, some effects are in fact duplicates if we allow extended parameters (custom1, custom2, custom3, etc) to be used.
I've already removed those from my fork, but for upstream we need to make it obvious for users that they are replaced.

I would vote for a separate app to modify existing presets if someone is willing to make one.

willmmiles:
I was thinking of a little bit of shim code at parse time - the list in the description above is short enough that I don't think it'd cost much.  Ideally I'd put that kind of thing in for handling a 0.15->0.16 migration, and then remove in 0.17.

DedeHai:
@netmindz the goal is not to save on flash but to remove redundant code and to clean up the FX list that has now grown quite large.

@willmmiles I would also opt for a built-in function rather than a conversion tool as this change might affect a lot of peoples setups (like christmas light setups) where an automatic conversion will save them a lot of hassle. It would mean we must not re-use the freed-up IDs in 0.16 at least. 
What about downgrading? if the now free IDs are re-used in the future and someone decides to downgrade to 0.16, the presets could get messed up.

willmmiles:
We could follow the same policy -- 0.16 deprecates the IDs, mapping them internally to the new implementatoin; 0.17 leaves them unallocated; 0.18 permits them to be re-used.

That said, in the future I'd like to get rid of numeric IDs altogether.  We already pay the code space for storing name strings, and not having to worry about ID contention makes safer to support FX usermods.

shupershuff:
It would be ideal if there was an external parser (eg a webapp) to detect deprecated or changed preset config

blazoncek:
> It would be ideal if there was an external parser (eg a webapp) to detect deprecated or changed preset config

If you can lend a hand, we'd be grateful.

blazoncek:
I would have preferred this being merged *before* adding new effects like #4682.

DedeHai:
since current path forward is to reserve the freed up ID's to be able to write parser for existing presets, it does not really matter when this is merged.

TripleWhy:
However, PRs age really poorly, so if you want to merge something, sooner is much better than later ;)

netmindz:
I think it's a good thing to have numeric ID for an effect,  I'm not a fan of reusing an ID, you would never do that on a database. Moving to uint16 has implications for WLED sync and DMX 
We do need a way to try and group similar effects together from a usability perspective

DedeHai:
> We do need a way to try and group similar effects together from a usability perspective

the idea is also to shorten the FX list, yes. There is a working html that lets you select/unselect which FX are in the list but I have misplaced the discord link to it. @dosipod may know where to find it.

Is moving to uint16 even an option with regards to backwards compatibility?

blazoncek:
Just FYI regarding switch to `uint16_t` for effect IDs:
* most users already complain about *too many* effects (my opinion as well)
* there are several places that will need update and it will be a nightmare to support backwards compatible UDP and ESP-NOW sync
* users will always complain about: either missing effects or broken integration or backwards compatibility or presets no longer working or showing incorrect effect, etc.

My own opinion is to stick with at most 255 effects and consolidate what can be consolidated and then employ something similar to @ewoudwijma 's ARTIFX (or however it is named) or @Aircoookie 's (yet to be implemented) dynamic effects.

That having said, I've already done the reduction in my fork (and even moved all audio effects into audioreactive usermod) and will base all of my future work based on that assumption.

ewoudwijma:
In MoonLight I skipped ID's completely and just store the effect name(s) in use in the data model, next to that also a 'family' will be stored showing the origin of the effect, eg wled, ps, animatrix, FastLED ...
Each family in a different .h file to be included or excluded as the user wishes

DedeHai:
since 0.16 will add many new FX the list is growing. Where do we stand on this?

blazoncek:
I've reduced effects in my fork a long time ago (as already mentioned). But I don't consider it my task to bring that upstream as I can see there are different opinions on how to do that (as were in the past).

BTW if it was my decision in the past we'd have renumbered and grouped effects already, but when I ported Audioreactive usermod from Soundreactive fork the decision was to keep effect IDs the same as in Soundreactive. IMO understandable but a bad decision we are paying for today.

netmindz:
I'm (currently) strongly against removing effectIDs entirely, swapping to uint16 is much less disruptive and gives us such a larger range that we can do things like assigning a range for custom/third party effects with reduced risk of clashing

In parallel, I think we also need to look at a number of UX issues that relate to things like grouping effects, maybe categories, things like a favourites list maybe, multiple tags (PS, volume based etc)

DedeHai:
I'd like to throw in one more thing: if the combined FX strings exceed the JSON buffer size (`serializeModeNames()`) there UI will be buggy. This already happens on ESP8266 when including all PS effects. Just something to look out for in the future.

netmindz:
See https://docs.google.com/document/d/15nSEHdtqQtY4FOqBK5-QgvbXdqOF2mY9APDJH7i5yfw/edit?tab=t.0 and https://discord.com/channels/473448917040758787/1416383094365093988/1416383098077184041

blazoncek:
@DedeHai the Blink/Blink Rainbow/Strobe/Strobe Rainbow ca all be made by just one effect using checkboxes.

ewowi:
> @DedeHai the Blink/Blink Rainbow/Strobe/Strobe Rainbow ca all be made by just one effect using checkboxes.

You could say the old fx-id be more like a category and in that effect you have a switch subeffect statement to call sub effects ...

DedeHai:
> @DedeHai the Blink/Blink Rainbow/Strobe/Strobe Rainbow ca all be made by just one effect using checkboxes.

done.

This is progressing well I think, the removed FX are now fully compatible with old presets and the user gets informed, at least if they are using the UI. 
What is still missing is a tool to auto-convert the presets.
My idea for that would be to make it a htm page, if an old preset is called, a button to said page is invoked at the top of the presets list (maybe even in red), the page informs users to backup their presets as the changes cannot be reverted if they decide to downgrade. If they accept, a JS function applies the changes from the switch-case I already made directly to the presets and uploads the new json file, replacing the old one.

ewowi:
(Posted this on discord but also here for the record):

maybe you have already a solution but could this work: 

- add uint8_t fc_id; // fx category: 7 bits for category (128 categories), 1 bit telling if it is classic id or not
- Keep fx_id uint8_t
- Existing effects get a new fc_id, choose whatever you like to group them, set classic id bit to true
- If a preset contains only fx_id (existing presets), not fc_id, find the fx_id with classic id == true, update the preset by adding the fc_id
- If a preset contains both fc_id and fx_id just match it
- new effects can have an already existing fx_id as long as they have another fc_id and the classic bit is false 
- In a far away future remove the 1 bit

TripleWhy:
Do you plan to block the old FX IDs from being re-used, forever?  
Do you plan to never re-use any ID in the future, no matter how many new effects are created and how many are removed?  
Can you actually make sure this will never happen?  
Remember, this has happened before.

* If so:
  - add a comment that both makes this policy very clear
  - make sure the comment is in a place where it would be seen if someone tried to violate this policy
  - add good reasoning to said comment, so people won't assume it's outdated
  - ideally, make it impossible to violate the policy accidentally, e.g. by moving the IDs into an enum, and then adding static_asserts that test if any of the blocked IDs are defined
* if not:
  - do not put the conversion code onto the microchip:
    * it blows up the code size for, ideally, a one time use
    * you can never reduce this code size, it will keep growing forever
    * maintaining it is not a fun job
  - if you must embed the conversion, make sure that presets are versioned!

About versioning:
* ideally its just a number in a version field
* inspired by ewowi's suggestion: you could simply rename the fx field, for some kind of implicit versioning. That would save the version field in the json, but make versioning impossible to intuitively understand for a human reader. It would also make conversion code even more complex

DedeHai:
> Do you plan to block the old FX IDs from being re-used, forever?

The idea is to block them in 0.16 and hopefully have a better solution in 0.17 and remove the mapping code again.

The main issue with moving away from 8bit FX numbers is backwards compatibility with external control software / sync / HA etc. there are some ideas we discussed but nothing is decided.

TripleWhy:
Well then, see above: make sure there is some way for 0.17 to tell if the preset came from 0.16, or something older

ewowi:
@triplewhy , I might have created a mess by having both discord and github for discussion 🙈

But on your remark: does it block fx_ids to be reused:

Answer: no, but there can be only one fx_id with the classic bit set to yes.

Each category (fc_id) could have the same fx_id as long as for these the classic bit is false 

The only requirement is you enforce that the combination <fc_id, classic bit,fx_id> is unique for an effect now and forever

blazoncek:
As FX ID change is *one time* issue for existing presets (now or in the future) I would not bother with any workaround or auxiliary tools. Once FX is removed or replaced it is up to the (annoyed) user to fix it. However, there have to be a clear and explicit instructions for each and every FX in question.

There are at least 18 FX that can be removed (and replaced by other effects that may need modification) according to my count.

One thing that hurts me still is the fact that there was a time where I've reordered (and grouped) effects when I ported Soundreactive to upstream. Unfortunately I was overruled and new effects kept their IDs from Soundreactive and no reordering was done.

Changing FX ID to 16 bit (or some other form) is a non-trivial task that will break backwards compatibility. Consider sync with devices from 0.10 onwards, which currently works as designed with all versions.

My 2c.

blazoncek:
Add Wipe/Sweep(Random) to the list.

DedeHai:
> As FX ID change is _one time_ issue for existing presets (now or in the future) I would not bother with any workaround or auxiliary tools.

> However, there have to be a clear and explicit instructions for each and every FX in question.

it is easiest to just have an auxiliary tool in 0.16 (and 0.16 only) to do it automatically as writing up clear instructions will almost take the same amount of time than writing a tool.


I agree on all other points.

> Changing FX ID to 16 bit (or some other form) is a non-trivial task that will break backwards compatibility. Consider sync with devices from 0.10 onwards, which currently works as designed with all versions.

syncing will only work with non-removed FX IDs, so this PR is a breaking change in that regard unless we keep the IDs reserved forever.

blazoncek:
> syncing will only work with non-removed FX IDs, so this PR is a breaking change in that regard unless we keep the IDs reserved forever.

Sync will still work and original (non replaced/removed) effects will still play. Replaced effects will play incorrect effec, but will still play.

You can also add Chase effects to the list, albeit with a bit of twist. Some used SEGCOLOR(1) for background and some SEGCOLOR(0). Code available in my fork.

wled00/util.cpp
```diff
@@ -592,14 +592,8 @@ void enumerateLedmaps() {
...
-  while (d < 42) {
-    r = hw_random8();
-    x = abs(pos - r);
-    y = 255 - x;
-    d = MIN(x, y);
-  }
-  return r;
+  uint8_t offset = hw_random8(172); // 256 - 2*42
+  return (pos + 42 + offset);
 }
```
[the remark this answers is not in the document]
DedeHai: offset is 0-171, pos +42 is restringing the new color to be outside of the minimum restraint, the max random number is pos+42+171 which is not closer to the original color on the other side of the wheel. so this is identical to original behaviour. if not explain why.

wled00/FX.cpp
```diff
@@ -262,49 +259,14 @@ uint16_t color_wipe(bool rev, bool useRandomColors) {
     } else
     {
       SEGMENT.setPixelColor(index, back? col0 : col1);
-      if (i == ledIndex) SEGMENT.setPixelColor(index, color_blend(back? col0 : col1, back? col1 : col0, uint8_t(rem)));
```
blazoncek, line 300: Why did you remove smooth transition?

DedeHai: because it acts on one single pixel for about 0.2s and I could not see any transition from that, even on lowest speed. So either this effect was slower initially or I did not test it right.

blazoncek:
This is my (current) list of effect IDs that can be freed (some already are), by slightly modifying other existing effects.
Not yet final version.
```c++
#define FX_MODE_COLOR_WIPE_RANDOM        4  // candidate for removal (was Wipe; use Wipe with check 3)
#define FX_MODE_COLOR_SWEEP              6  // candidate for removal (was Sweep; use Wipe with check 1)
#define FX_MODE_DUAL_SCAN               11  // candidate for removal (use Scan with check 1)
#define FX_MODE_THEATER_CHASE_RAINBOW   14  // candidate for removal (was Theater Rainbow; use Theater with check 3)
#define FX_MODE_SAW                     16  // candidate for removal (was Saw; use Running Lights with check 2)
#define FX_MODE_DISSOLVE_RANDOM         19  // candidate for removal (was Dissolve random; use Dissolve with with check 3)
#define FX_MODE_STROBE                  23  // candidate for removal (was Strobe; use Blink with with check 2)
#define FX_MODE_STROBE_RAINBOW          24  // candidate for removal (was Strobe Rainbow; use Blink with with check 1 & check 2)
#define FX_MODE_BLINK_RAINBOW           26  // candidate for removal (was Blink Rainbow; use Blink with with check 1)
#define FX_MODE_CHASE_RANDOM            29  // candidate for removal (was Chase Random; use Chase with check 3)
#define FX_MODE_CHASE_RAINBOW           30  // candidate for removal (was Chase Rainbow; use Chase with check 1 & check 3)
#define FX_MODE_CHASE_RAINBOW_WHITE     33  // candidate for removal (was Chase Rainbow White; use Chase with check 1 & check 2 & check 3)
#define FX_MODE_COLOR_SWEEP_RANDOM      36  // candidate for removal (was Sweep Random; use Wipe with check 1 & check 3)
#define FX_MODE_RUNNING_COLOR           37  // candidate for removal (use Theater)
#define FX_MODE_LOADING                 47  // candidate for removal (use Gradient with check 1)
#define FX_MODE_RUNNING_DUAL            52  // candidate for removal (use Running)
#define FX_MODE_HALLOWEEN               53  // removed in 0.14!
#define FX_MODE_DUAL_LARSON_SCANNER     60  // candidate for removal (use Scanner with with check 1)
#define FX_MODE_METEOR_SMOOTH           77  // candidate for removal (use Meteor with check 1)
#define FX_MODE_TWINKLECAT              81  // candidate for removal (use Twinklefox with check 3)
#define FX_MODE_SPOTS_FADE              86  // candidate for removal (use Spots with check 1)
#define FX_MODE_SINELON_DUAL            93  // candidate for removal (use sinelon)
#define FX_MODE_SINELON_RAINBOW         94  // candidate for removal (use sinelon)
#define FX_MODE_RIPPLE_RAINBOW          99  // candidate for removal (use ripple)
#define FX_MODE_CANDLE_MULTI           102  // candidate for removal (use candle with multi select)
#define FX_MODE_SOLID_GLITTER          103  // candidate for removal (use glitter)
#define FX_MODE_PHASEDNOISE            109  // candidate for removal (use Phased with check 1)
#define FX_MODE_DYNAMIC_SMOOTH         117 // candidate for removal (check3 in dynamic)
```

netmindz:
Unnecessary breaking changes impacting everyone's presets that feature any affected effect is **not acceptable** if this is just about an artificial restriction to always keep effect ID as single byte

netmindz:
Duplication of the effect code, **while maintaining the effect ID and expected behaviour** is good software design and therefore perfectly ok

netmindz:
> > syncing will only work with non-removed FX IDs, so this PR is a breaking change in that regard unless we keep the IDs reserved forever.
> 
> Sync will still work and original (non replaced/removed) effects will still play. Replaced effects will play incorrect effec, but will still play.

Playing the wrong effect is not working!

netmindz:
I think we can rework the implementation a little here so that we don't cause breaking changes between versions. 

When calling a legacy effect, we shouldn't change our effectID, but rather delegate the implementation to the new function, that way we still sync the original ID, therefore keeping compatibility @DedeHai 

If/how we can do that while still having the nice feature of the warning, without too much boilerplate code checking for first frame, might take some thought

netmindz:
> BTW if it was my decision in the past we'd have renumbered and grouped effects already, but when I ported Audioreactive usermod from Soundreactive fork the decision was to keep effect IDs the same as in Soundreactive. IMO understandable but a bad decision we are paying for today.

I don't understand what you mean with this comments sorry. Surely weather or not IDs were kept consistent between forks or assigned a new ID does affect the number of IDs used? The current system doesn't apply any significance to the ID as we list effects alphabetically, not by ID so there is no grouping based on ID

TripleWhy:
From what I have gathered, everyone was on the same page in the end, including yourself. So is there any reason why you closed the PR?

netmindz:
Sorry, didn't mean to close, miss tapped

softhack007:
@DedeHai @netmindz @TripleWhy is this PR still "up to date", or does it have open ends to be fixed?

DedeHai:
This needs some more discussion in a meeting I think:
- FX will disapear from the list, users will complain (minor IMHO)
- the "Auto forwarding" of FX IDs may become an issue in the future
- I had proposed to add a button in the UI that will automatically update all presets so we can get rid of the FX ID forwarding in the future, however this will break backwards compatibility

PR #4575, opened 2025-02-26
Add ethernet option

thirstyice:
I figure I can't be the only one to do ethernet like this.

Maybe it would be worth providing a "Custom" option that allows you to set each setting individually?
Beyond the scope of this PR though.

### DIFF
--- a/wled00/data/settings_wifi.htm
+++ b/wled00/data/settings_wifi.htm
@@ -80,7 +80,7 @@
 						select.appendChild(option);
 
 						if (input.value === "" || input.value === "Your_Network" || found) input.replaceWith(select);
-						else select.remove(); 
+						else select.remove();
 					}
 				}
 
@@ -225,6 +225,7 @@ Static subnet mask:<br>
 				<option value="5">TwilightLord-ESP32</option>
 				<option value="3">WESP32</option>
 				<option value="1">WT32-ETH01</option>
+				<option value="13">Thirstyice</option>
 			</select><br><br>
 		</div>
 		<hr>
--- a/wled00/network.cpp
+++ b/wled00/network.cpp
@@ -107,10 +107,10 @@ const ethernet_settings ethernetBoards[] = {
 
   // ABC! WLED Controller V43 + Ethernet Shield & compatible
   {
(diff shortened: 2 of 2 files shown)


netmindz:
Can you include a link to the board this relates to? I had as quick Google and couldn't find it

thirstyice:
It's not a purchasable board (yet - I'm working on it), but it's the way I often hook up phy boards such as [this one](https://www.amazon.ca/dp/B07G9TP5YP) to my esp32 dev board

thirstyice:
There's still a little work for me to do - for me, the ethernet connects then immediately disconnects. Need to figure out if the problem is hardware or software.

netmindz:
Adding an option for a bespoke board doesn't make sense

blazoncek:
> So I could use GPIO17 for LED output.

You can use GPIO0 for LED output as well.

> I feel like the real solution is to add a "custom" option that lets users define the gpios as they like

Tell that to not so tech savvy people. But if you insist, why not make a PR?

> Anyway, if you don't want to include this then feel free to close it.

Adding another (and another and another etc) board, just because someone decided to use different GPIO than others (without proper justification) is problematic as it eats away precious flash memory. That does not mean the PR is worthless or should be closed without merging. I am just trying to get justification why this choice?

thirstyice:
> You can use GPIO0 for LED output as well.

My logic level converter holds the pin high at boot, which is a problem for GPIO0 when I need to reflash.

> Tell that to not so tech savvy people. But if you insist, why not make a PR?

This is my first foray into web-based settings on an esp, so I don't feel like I know enough about how it works to do anything more complicated than add an item to a dropdown (at least, not yet!)

> Adding another (and another and another etc) board, just because someone decided to use different GPIO than others (without proper justification) is problematic as it eats away precious flash memory.

This I definitely agree with.

>  I am just trying to get justification why this choice?

Mostly because I found myself compiling my own version just for this change, and I figured it would be easier if there was an option in the dropdown, and maybe someone else out there would think the same. I don't have a super strong conviction that this is a thing that needs to happen. I figured I'd toss it out there, but I won't be upset if it doesn't get merged.

PR #4578, opened 2025-03-01
Reduce segment copies

TripleWhy:
Use move instead of copy where possible, avoid complexity/memory usage of full copies in other locations.

(Currently, this version would compile with deleted Segment copy functions.)

From the json/differs commit:

> Move segment json-specific segment diffing to json.cpp
> - Avoid calling Segment copy constructor and destructor. This is fine
> as long as we only read primitive values.
> - Only diff stuff we are actually interested in.
> - Only diff stuff if it has an effect.

Note that the json copy code is effectively very similar to before, because the copy constructor also does a memcpy, plus additional work.

About `applyValuesToSelectedSegs`:  
This could be simplified a lot, if we were to alter the behavior such that all selected segments would be modified unconditionally. Which would also address this function comment:

> `// problem: if the first selected segment already has the value to be set, other selected segments are not updated`


wled00/led.cpp
```diff
@@ -28,21 +28,27 @@ void setValuesFromSegment(uint8_t s)
 // problem: if the first selected segment already has the value to be set, other selected segments are not updated
 void applyValuesToSelectedSegs()
 {
-  // copy of first selected segment to tell if value was updated
-  unsigned firstSel = strip.getFirstSelectedSegId();
-  Segment selsegPrev = strip.getSegment(firstSel);
```
blazoncek, line 33: This was an oversight in original. Const reference can (and should) be used. Fixed long time ago in my fork.

wled00/FX.h
```diff
@@ -883,7 +882,7 @@ class WS2812FX {  // 96 bytes
     inline void trigger()                                     { _triggered = true; }  // Forces the next frame to be computed on all active segments.
     inline void setShowCallback(show_callback cb)             { _callback = cb; }
     inline void setTransition(uint16_t t)                     { _transitionDur = t; } // sets transition time (in ms)
-    inline void appendSegment(const Segment &seg = Segment()) { if (_segments.size() < getMaxSegments()) _segments.push_back(seg); }
+    inline void appendSegment(Segment &&seg = Segment())      { if (_segments.size() < getMaxSegments()) _segments.push_back(std::move(seg)); }
```
blazoncek, line 885: I've replace this with emplace_back, does the same.

wled00/FX_fcn.cpp
```diff
@@ -1790,7 +1763,7 @@ void WS2812FX::resetSegments() {
   #else
   segment seg = Segment(0, _length);
   #endif
-  _segments.push_back(seg);
+  _segments.push_back(std::move(seg));
```
blazoncek, line 1766: This can be emplaced_back as well.

wled00/json.cpp
```diff
@@ -34,7 +62,8 @@ bool deserializeSegment(JsonObject elem, byte it, byte presetId)
   //DEBUG_PRINTLN(F("-- JSON deserialize segment."));
   Segment& seg = strip.getSegment(id);
   //DEBUG_PRINTF_P(PSTR("--  Original segment: %p (%p)\n"), &seg, seg.data);
-  const Segment prev = seg; //make a backup so we can tell if something changed (calling copy constructor)
+  alignas(Segment) std::array<byte, sizeof(Segment)> segmentBackup; //make a backup so we can tell if something changed (voiding copy constructor)
```
blazoncek, line 65: Interesting principle.

blazoncek (commented):
I've already fixed some of unnecessary copies by using emplace_back.
I do like `differs()` approach which does away with copying effect data.

wled00/led.cpp
TripleWhy: Judging by the comments and how it is used, that was clearly intentional. Just using a reference instead of a copy won't help here. Except if the `problem` function comment is something that should be changed, see PR description.

wled00/FX.h
TripleWhy: Not sure how this could be replaced with emplace here?

TripleWhy: That is, if you want to keep this `appendSegment` function. Removing it opens options.

blazoncek: Change its parameters to accept segment coordinates instead.

wled00/led.cpp
blazoncek: Comment is off. There is no need to copy the segment as it si not modified in the for loop (see `firstSel`).

TripleWhy: The loop iterates over all segments. When `i == firstSel`, the segment is *not* skipped, so `seg`, which is `selsegPrev` at that point in time if you use a reference, is modified, is it not?

blazoncek: `if (i == firstSel || !(seg.isActive() && seg.isSelected())) continue;` that's translated if. Isn't it?
At least it should be written as this.

wled00/led.cpp
```diff
@@ -28,21 +28,27 @@ void setValuesFromSegment(uint8_t s)
 // problem: if the first selected segment already has the value to be set, other selected segments are not updated
```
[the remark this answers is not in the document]
blazoncek: That's the case. Consider:
```c++
setValuesFromSegment(strip.getFirstSelectedSegId());
...
applyValuesToSelectedSegs();
```
That's due to the ifs in the function.
Why they were written as they are is beyond my knowing. Probably as some optimisation. I just kept original logic when updating.
If all ifs are removed, so is the problem. 😄

wled00/led.cpp
TripleWhy:
```c++
if (i != firstSel && (!seg.isActive() || !seg.isSelected())) continue;
```

this is the unmodified line.

if you want to invert it, the continue would fall away or move into the else branch:

```c++
if (!(i != firstSel && (!seg.isActive() || !seg.isSelected()))) {
  // the interesting stuff
} else {
  continue; // boring
}
```

```
   !(i != firstSel  &&  (!seg.isActive() || !seg.isSelected()))
=  !(i != firstSel) || !(!seg.isActive() || !seg.isSelected())
=   (i == firstSel) ||  ( seg.isActive() &&  seg.isSelected())
```

So: firstSel get's modified, no matter if it's active or selected. Additionally, other segments get modified, if they are both active and selected at the same time.

blazoncek: Then you found a bug.

TripleWhy: I'm not so sure about hat, it looks very intentional, and the comments align with it, too.

blazoncek: Prior to having segments (in 0.9) HTTP API only dealt with setting colors as global `col[]` and `colSec[]` etc. This was replicated in IR code (code was unaware of segments, i.e. how do you tell IR remote that you want to target 2nd segment?).

As `col[]` and `colSec[]` are part of official HTTP API there has to be a way to set them from a newer code. This is where these functions arise from - they set legacy values that are used in old code that has not been rewritten. They are only used in IR code and if preset uses HTTP API (legacy macro converted into preset).

blazoncek: Had some time to spare... This would be correct implementation:
```c++
// applies global legacy values (col, colSec, effectCurrent...)
void applyValuesToSelectedSegs() {
  for (unsigned i = 0; i < strip.getSegmentsNum(); i++) {
    Segment& seg = strip.getSegment(i);
    if (!(seg.isActive() && seg.isSelected())) continue;
    if (effectSpeed     != seg.speed)     {seg.speed     = effectSpeed;     stateChanged = true;}
    if (effectIntensity != seg.intensity) {seg.intensity = effectIntensity; stateChanged = true;}
    if (effectPalette   != seg.palette)   {seg.setPalette(effectPalette);}
    if (effectCurrent   != seg.mode)      {seg.setMode(effectCurrent);}
    uint32_t col0 = RGBW32(colPri[0], colPri[1], colPri[2], colPri[3]);
    uint32_t col1 = RGBW32(colSec[0], colSec[1], colSec[2], colSec[3]);
    if (col0 != seg.colors[0])            {seg.setColor(0, col0);}
    if (col1 != seg.colors[1])            {seg.setColor(1, col1);}
  }
}
```

TripleWhy:
Updated applyValuesToSelectedSegs to the suggested logic.

Removed all copy and diff code from json code, based on this assumption: https://github.com/wled/WLED/pull/4580#issuecomment-2692654622

blazoncek:
> based on this assumption: [#4580 (comment)](https://github.com/wled/WLED/pull/4580#issuecomment-2692654622)

This is actually a bug in the `setOption()` code. If the `on` state didn't change so shouldn't `stateChanged`.
`setOption()` is used to force transition for changes of `on` (or power) state. In the past it was used to set other bitfields which are now accessible directly (but such change does not invoke `startTransition()`).

So, IMO, `differs()` should stay. However, there is no need to copy entire segment as currently implemented since only segment attributes are compared.

TripleWhy:
I can remove the commit that removes the differs function from this PR.

> However, there is no need to copy entire segment as currently implemented since only segment attributes are compared.

I'm not sure what you are saying here. Are you favoring the version I had in this PR or the version from the other PR?

blazoncek:
I am just trying to clarify. I did like the solution in https://github.com/wled/WLED/pull/4578/commits/90a49c5a072c668a8c855beb765e029d79821e34

These: `stateChanged` and `stateUpdated()` or (God forbid) `colorUpdated()` are all helpers to (mostly) keep legacy code happy.
If you are curious, I don't like any of them.

TripleWhy:
Removed 4d2ac3a780e24eca7c5511098e3549cdcaf97273, this should be what you want

DedeHai:
is this PR still relevant with the new layering code?

blazoncek:
I do not think it is as layer blending did reduce segment copies in a similar way as this PR.

TripleWhy:
Merged main into this PR, and if you look at the diff now, there isn't really much left of the actual reduction, just some general adjustments

DedeHai:
I glanced over it, to me it looks like no change at all apart from som re-factoring which is kind of arbitrary / personal style.

TripleWhy:
All of these changes are clearly improvements, but yes, most of them are minor, and don't affect the number of segment copies

wled00/json.cpp
```diff
@@ -103,10 +98,7 @@ static bool deserializeSegment(JsonObject elem, byte it, byte presetId = 0)
     seg.intensity,
     seg.custom1,
     seg.custom2,
-    seg.custom3,
-    seg.check1,
```
DedeHai, line 107: so setting the chemarks here was redundant?

TripleWhy: Well they wer never used. Maybe they should, but it has been that way forever. Changing behavior is not this PR's intention. So in the spirit of the minimal copy struct here, removing these values makes it more minimal. It also makes it a bit clearer that something is missing. If you just see that struct and how it is used (apart from the actual diff function) you would assume that the bools are used, and used in a specific way

Issue #4550, opened 2025-02-13
Segment layering/blending


blazoncek:
With 0.14 we've got proper blending of color and palettes across all segments, 0.15 improved on that and allowed effects to be *blended* as well. Recent *blending styles* (#4158) introduced effect transitions with different styles.

What is missing IMO is correct segment blending/layering with different blending modes known to Photoshop users, i.e. lighten, darken, multiply, add, subtract, difference, etc.

I've recently come across a snippet that I find useful and relatively simple to implement in WLED. Unfortunately it would require re-implementation of pixel drawing functions.

Let's first see the code and then discuss the nuances:
```c++
void WS2812FX::blendSegments(const Segment &top, const Segment &bottom) {
  constexpr auto top        = [](uint8_t a, uint8_t b){ return a; };
  constexpr auto bottom     = [](uint8_t a, uint8_t b){ return b; };
  constexpr auto add        = [](uint8_t a, uint8_t b){ return a + b; };
  constexpr auto subtract   = [](uint8_t a, uint8_t b){ return b > a ? (b - a) : 0; };
  constexpr auto difference = [](uint8_t a, uint8_t b){ return b > a ? (b - a) : (a - b); };
  constexpr auto average    = [](uint8_t a, uint8_t b){ return (a + b) >> 1; };
  constexpr auto multiply   = [](uint8_t a, uint8_t b){ return a * b; };
  constexpr auto divide     = [](uint8_t a, uint8_t b){ return b != 0 ? a / b : 0; };
  constexpr auto lighten    = [](uint8_t a, uint8_t b){ return a>b ? a : b; };
  constexpr auto darken     = [](uint8_t a, uint8_t b){ return a<b ? a : b; };
  constexpr auto screen     = [](uint8_t a, uint8_t b){ return uint8_t(0xFF - ~a * ~b); }; // 255 - (255-a)*(255-b)
  constexpr auto overlay    = [](uint8_t a, uint8_t b){ return uint8_t(a<0x80 ? 2*a*b : (0xFF - 2 * ~a * ~b)); };

  using FuncType = uint8_t(*)(uint8_t, uint8_t);
  FuncType funcs[] = {
    top, bottom, add, subtract, difference, average, multiply, divide, lighten, darken, screen, overlay
  };

  const uint8_t blendMode = top.blendMode;
  auto func = funcs[blendMode % (sizeof(funcs) / sizeof(FuncType))];

  for (auto i = 0; i<top.length(); i++) {
[...]

PR #4580, opened 2025-03-01
Make Segment class non-copyable and move-only

JoaDick:
This is really pure coincidence... Apparently @TripleWhy and me have been working on the same topic at the same time:
Eliminating copies of the Segment class. See the first PR here: https://github.com/wled/WLED/pull/4578

The reasons for this alternative PR - after hesitating for a long time - are the following:
- This PR prohibits copying of the Segment and WS2812FX class entirely.
- It is less intrusive, preserving most of the existing code.
- It does _not_ treat a segment object as raw memory.

The last point of this list is the most important one IMHO. Using `memcpy()`, `reinterpret_cast()`, and alike on objects seems very convenient in the beginning. But they are also very prone to becoming a really bad pitfall in the future! And the performance gain - if any - is very disputable.
That is also why many C++ coding standards from the industry (MISRA, AUTOSAR, SEI-CERT ...) do not allow using these; see here for example:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-memset
https://wiki.sei.cmu.edu/confluence/display/cplusplus/OOP57-CPP.+Prefer+special+member+functions+and+overloaded+operators+to+C+Standard+Library+functions

As said, I was hesitating for a long time, and studied the PR and its discussion intensively. Just wait for the other PR to become available? Or incorporate its suggestions into this C++ compliant PR and give credit?

Well... here's the result. I really hope no one is offended by this alternative!

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -16,6 +16,7 @@
 #ifndef WS2812FX_h
 #define WS2812FX_h
 
+#include <array>
 #include <vector>
 #include "wled.h"
 
@@ -385,6 +386,8 @@ typedef enum mapping1D2D {
   M12_sPinwheel = 4
 } mapping1D2D_t;
 
+struct SegmentSettings;
+
 // segment, 68 bytes
 typedef struct Segment {
   public:
@@ -414,7 +417,7 @@ typedef struct Segment {
     };
     uint8_t  grouping, spacing;
     uint8_t  opacity;
-    uint32_t colors[NUM_COLORS];
+    std::array<uint32_t, NUM_COLORS> colors;
     uint8_t  cct;                 //0==1900K, 255==10091K
     uint8_t  custom1, custom2;    // custom FX parameters/sliders
     struct {
@@ -565,7 +568,7 @@ typedef struct Segment {
       stopY  = sStopY;
     }
 
-    Segment(const Segment &orig); // copy constructor
+    Segment(const Segment&) = delete;
     Segment(Segment &&orig) noexcept; // move constructor
 
     ~Segment() {
@@ -580,7 +583,7 @@ typedef struct Segment {
       deallocateData();
     }
 
-    Segment& operator= (const Segment &orig); // copy assignment
+    Segment& operator= (const Segment&) = delete;
     Segment& operator= (Segment &&orig) noexcept; // move assignment
 
 #ifdef WLED_DEBUG
@@ -626,7 +629,7 @@ typedef struct Segment {
     Segment &setMode(uint8_t fx, bool loadDefaults = false);
     Segment &setPalette(uint8_t pal);
     Segment &setName(const char* name);
-    uint8_t differs(const Segment& b) const;
+    uint8_t differs(const SegmentSettings& b) const;
     void    refreshLightCapabilities();
 
     // runtime data functions
@@ -784,6 +787,48 @@ typedef struct Segment {
 } segment;
 //static int segSize = sizeof(Segment);
 
+struct SegmentSettings
+{
+  explicit SegmentSettings(const Segment& b)
+    : start{b.start}
+    , stop{b.stop}
+    , offset{b.offset}
+    , grouping{b.grouping}
+    , spacing{b.spacing}
+    , opacity{b.opacity}
+    , mode{b.mode}
+    , speed{b.speed}
+    , intensity{b.intensity}
+    , palette{b.palette}
+    , custom1{b.custom1}
+    , custom2{b.custom2}
+    , custom3{b.custom3}
+    , startY{b.startY}
+    , stopY{b.stopY}
+    , colors{b.colors}
+    , options{b.options}
+  {}
+
+  uint16_t start;
+  uint16_t stop;
+  uint16_t offset;
+  uint8_t grouping;
+  uint8_t spacing;
+  uint8_t opacity;
+  uint8_t mode;
+  uint8_t speed;
+  uint8_t intensity;
+  uint8_t palette;
+  uint8_t custom1;
+  uint8_t custom2;
+  uint8_t custom3;
+  // also include check1/2/3 ?
+  uint8_t startY;
+  uint8_t stopY;
+  std::array<uint32_t, NUM_COLORS> colors;
+  uint16_t options;
+};
+
 // main "strip" class
 class WS2812FX {  // 96 bytes
   typedef uint16_t (*mode_ptr)(); // pointer to mode function
@@ -798,6 +843,10 @@ class WS2812FX {  // 96 bytes
   static WS2812FX* instance;
 
   public:
+    WS2812FX(const WS2812FX&) = delete;
+    WS2812FX(WS2812FX&&) = default;
+    WS2812FX& operator=(const WS2812FX&) = delete;
+    WS2812FX& operator=(WS2812FX&&) = default;
 
     WS2812FX() :
       paletteBlend(0),
@@ -883,7 +932,7 @@ class WS2812FX {  // 96 bytes
     inline void trigger()                                     { _triggered = true; }  // Forces the next frame to be computed on all active segments.
     inline void setShowCallback(show_callback cb)             { _callback = cb; }
     inline void setTransition(uint16_t t)                     { _transitionDur = t; } // sets transition time (in ms)
-    inline void appendSegment(const Segment &seg = Segment()) { if (_segments.size() < getMaxSegments()) _segments.push_back(seg); }
+    inline void appendSegment(Segment &&seg = Segment())      { if (_segments.size() < getMaxSegments()) _segments.push_back(std::move(seg)); }
     inline void suspend()                                     { _suspend = true; }    // will suspend (and canacel) strip.service() execution
     inline void resume()                                      { _suspend = false; }   // will resume strip.service() execution
 
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -91,18 +91,6 @@ uint8_t  Segment::_clipStartY = 0;
 uint8_t  Segment::_clipStopY = 1;
 #endif
 
-// copy constructor
-Segment::Segment(const Segment &orig) {
-  //DEBUG_PRINTF_P(PSTR("-- Copy segment constructor: %p -> %p\n"), &orig, this);
-  memcpy((void*)this, (void*)&orig, sizeof(Segment));
-  _t = nullptr; // copied segment cannot be in transition
-  name = nullptr;
-  data = nullptr;
-  _dataLen = 0;
-  if (orig.name) { name = static_cast<char*>(malloc(strlen(orig.name)+1)); if (name) strcpy(name, orig.name); }
-  if (orig.data) { if (allocateData(orig._dataLen)) memcpy(data, orig.data, orig._dataLen); }
-}
-
 // move constructor
 Segment::Segment(Segment &&orig) noexcept {
   //DEBUG_PRINTF_P(PSTR("-- Move segment constructor: %p -> %p\n"), &orig, this);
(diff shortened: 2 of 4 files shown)


TripleWhy:
Hmmm  
If you were to make `SegmentSettings` a member of `Segment`, and remove the relevant members from `Segment`, this approach would be clearly better, I think. Could have more advantages, too. However, that's probably a lot of work.

With this current proposal, note, that `SegmentSettings` stores more than what is needed for `applyValuesToSelectedSegs`. 🤷

The reason I did not forbid copies however is that it currently looks like it's going to be needed [again](https://github.com/wled/WLED/issues/4550#issuecomment-2687136622)... Which is also why I didn't bother avoiding memcpy, which is used internally by the copy constructor (which doesn't make it less dangerous, just less obvious).

TripleWhy:
Oh right, what I forgot to mention is that I also only call differs if that would actually do anything, i.e. if `stateChanged` isn't true already. And I haven't found a way to trigger that from the UI... So all of this copy and differ stuff might actually be obsolete? Does anyone know more about that?

blazoncek:
> So all of this copy and differ stuff might actually be obsolete?

Indeed it may be. `differs()` is a relic from 0.11 era where there was no `stateChanged` flag. Tracing the workflow may prove that.

blazoncek:
@JoaDick what was the driving force for removing/prohibiting segment copy?

BTW I'm working on a feature that depends on segment copy (or at least segment settings + effect data + pixel state data copy).

blazoncek:
> So all of this copy and differ stuff might actually be obsolete?

Checked the code: It isn't. `differs()` is still valid.

TripleWhy:
> Oh right, what I forgot to mention is that I also only call differs if that would actually do anything, i.e. if `stateChanged` isn't true already. And I haven't found a way to trigger that from the UI... So all of this copy and differ stuff might actually be obsolete? Does anyone know more about that?

I found the reason:

`deserializeSegment` calls `seg.setOption(SEG_OPTION_ON, getBoolVal(elem["on"], seg.on));`, unconditionally, [here](https://github.com/TripleWhy/WLED/blob/9dadc0628cabcdf4f02549184254d04db05775a8/wled00/json.cpp#L158).
`setOption` then [sets](https://github.com/TripleWhy/WLED/blob/9dadc0628cabcdf4f02549184254d04db05775a8/wled00/FX_fcn.cpp#L604) `stateChanged` (because `SEG_OPTION_ON` is neither `SEG_OPTION_SELECTED` nor `SEG_OPTION_RESET`). It doesn't compare the old and new values.

Ergo: `stateChanged` is always set, the result of `differs` never matters, whether all effect parameters should be compared or not is irrelevant, and the copy is never needed. At least if that stays as it is.

JoaDick:
> Hmmm If you were to make `SegmentSettings` a member of `Segment`, and remove the relevant members from `Segment`, this approach would be clearly better, I think. Could have more advantages, too. However, that's probably a lot of work.

Fully agree, and I'd appreciate such a refactoring in the future. For this PR I resisted to do so to keep it less intrusive.

> With this current proposal, note, that `SegmentSettings` stores more than what is needed for `applyValuesToSelectedSegs`. 🤷

As suggested by @blazoncek,  `applyValuesToSelectedSegs()` now stores just a reference to the first selected segment, and no more a copy or the new `SegmentSettings`.
 
> The reason I did not forbid copies however is that it currently looks like it's going to be needed [again](https://github.com/wled/WLED/issues/4550#issuecomment-2687136622)... Which is also why I didn't bother avoiding memcpy, which is used internally by the copy constructor (which doesn't make it less dangerous, just less obvious).

Thanks for this hint; I wasn't aware of that change and will have a closer look at it.

JoaDick:
> @JoaDick what was the driving force for removing/prohibiting segment copy?

The main reason is probably very similar to what @TripleWhy is striving for: providing a possibility that effects can be implemented in terms of a class and not just as a plain function. The current way stems back from very early roots - [Kitesurfer's WS2812FX](https://github.com/kitesurfer1404/WS2812FX/tree/master) - and is still suffering from the same limitations: keeping data persistent between frames is only possible through this kind of a workaround via `SEGENV`.
Effect implementations could be made much more simpler and understandable, if the persistent data would just be a class member. Also, certain helper functions could be embedded as class members, which would increase encapsulation.

In my playground, I tried something very similar (but much smaller) as @TripleWhy's proposal from here: https://github.com/wled/WLED/pull/4549
The idea is that these new kind of effects shall be implemented as a class, which derives from a specific base class. They just have to implement a virtual `showEffect()` method, where all the logic lives. This logic could mostly be migrated 1:1 from the existing method-based implementation into the new `showEffect()` method - i.e. just a renaming of the function.
An essential claim is also that both, the existing method-based and the new class-based implementations, can coexist and don't require immediate changes.
However, the copying of `Segment` class and its raw byte-based memory management is making things really difficult or even impossible.
If you're interested, I'd be happy to prepare a draft PR for review.

Another driving force are SW design considerations. In a nutshell, only so called value-classes shall be copyable. Most other kind of classes - in particular logic-classes like `Segment` - shall not be copyable. Although it might seem convenient sometimes, it will most certainly strike back in the future.

> BTW I'm working on a feature that depends on segment copy (or at least segment settings + effect data + pixel state data copy).

Thanks a lot for this hint - I'll have a closer look at that as well. Hopefully we can figure out an idea to bring all that great stuff together!

TripleWhy:
> As suggested by @blazoncek, applyValuesToSelectedSegs() now stores just a reference to the first selected segment, and no more a copy or the new SegmentSettings.

In my understanding that would alter the current behavior in funny ways. Either that, or I'm missing something

JoaDick:
> BTW I'm working on a feature that depends on segment copy (or at least segment settings + effect data + pixel state data copy).

Whoa, that's quite a lot and really impressive!
Hopefully I understood all that correctly in that short time, so please correct me if I got anything wrong.
Apparently it is about [this issue](https://github.com/wled/WLED/issues/4550#issuecomment-2692030691), and the copy of the Segment is required [here](https://github.com/blazoncek/WLED/blob/unified-blend/wled00/FX_fcn.cpp#L294C2-L294C111).

As a quick solution, I'd suggest to replace the copy constructor with a `clone()` method. That might sound like a cheap trick, because from a technical perspective it is doing just a copy. But from a design perspective, it will clearly state the intent when it is used, and accidental copies are avoided.
Nevertheless, this workaround doesn't solve the underlying problem: the raw memory handling of `Segment::data`, which of course has to be copied as well. And exactly that raw byte copying makes it nearly impossible to use "real" C++ objects and pointers to them inside `Segment::data`. Such rawly copied pointers still point to the old location :worried: To my understanding, this appears to be a major reason why the ParticleSystem had to implement such a complicated memory management.
However, `Segment::data` might become obsolete one day when we follow @TripleWhy's or my idea of a class based effect implementation. As a consequence, these new effects would also have to provide a `clone()` method - but that might be acceptable.

blazoncek:
> Apparently it is about [this issue](https://github.com/wled/WLED/issues/4550#issuecomment-2692030691), and the copy of the Segment is required [here](https://github.com/blazoncek/WLED/blob/unified-blend/wled00/FX_fcn.cpp#L294C2-L294C111).

Yes.

Just FYI the current implementation in that branch is working miraculously on ESP8266 yielding 55 FPS on a 16x16 matrix. 40+ FPS when 3 effects are stacked one atop another and blended.

AFAIK @DedeHai needed to implement (complex) memory management since PS needs a lot of memory and running i.e. 3 layers (with transition going on) exhausted available RAM. 

As for suggestion: I am still a newbie as far as C++ goes. I may not understand everything you or @TripleWhy are trying to pass on to me.

TripleWhy:
> Nevertheless, this workaround doesn't solve the underlying problem: the raw memory handling of `Segment::data`, which of course has to be copied as well. And exactly that raw byte copying makes it nearly impossible to use "real" C++ objects and pointers to them inside `Segment::data`. Such rawly copied pointers still point to the old location 😟 To my understanding, this appears to be a major reason why the ParticleSystem had to implement such a complicated memory management.

Well, you know what would [solve](https://github.com/wled/WLED/pull/4580#issuecomment-2692588941) all of that, right? ;)
You might not even need to copy the segment with this, or if you do, it should be a simpler avoid memcpy.

> However, `Segment::data` might become obsolete one day when we follow @TripleWhy's or my idea of a class based effect implementation. As a consequence, these new effects would also have to provide a `clone()` method - but that might be acceptable.

Yeah I thought it could be that way, too. But transitions from old um, states, to new states don't just include the effect and all of its parameters, but also the way you set up segments. These days I don't use segments much, but before 2D setups, you had to define separate presets with completely different segment definitions to achieve a simple color transition from left to right or top to bottom. The transition should still work when switching between those setups. Thus all old and new segment definitions and effect parameters need to be known during the transition.

JoaDick:
OK, might have been a bit too much information at once in my last comments... What I wanted to say with those many words:

- With some minor modifications, this PR could be made compatible with your concept branch. Without any performance impact 👍 
- Unnecessary and accidental copies of `Segment` are gone.
- But the existing problems caused by the raw memory management inside `Segment` still remain.

JoaDick:
> Well, you know what would [solve](https://github.com/wled/WLED/pull/4580#issuecomment-2692588941) all of that, right? ;) You might not even need to copy the segment with this, or if you do, it should be a simpler avoid memcpy.

This would surely be a first step in the right direction 👍 

> Yeah I thought it could be that way, too. But transitions from old um, states, to new states don't just include the effect and all of its parameters, but also the way you set up segments. [...] Thus all old and new segment definitions and effect parameters need to be known during the transition.

Very well summarized. I think we're having quite the same thoughts on this topic.

Now it would be really good to get feedback if there is general interest in approaching this direction towards class based effects.
We've already had some really good ideas and concepts for that - which might probably be too much for one PR without knowing the general direction. (Especially for a WLED newbie-contributor like me 😅 )

blazoncek:
I've read that again. Hopefully getting it right.

The main issue is that `_data` is raw byte array. Right?
Placing objects in it (in current implementation) is prone to errors especially if those objects allocate their own memory. Right?

Rewriting everything from scratch (this or the other PR) aside the easy "solution" is to prohibit memory allocation from within objects placed into `_data` (i.e. using placement new or other more unsafe methods).

This, of course, contradicts PS approach which *does* need extra memory and has to come with its own solution.

IMO the solution lies somewhere in the middle. The approach by @TripleWhy is very radical (and probably unfit for every existing effect, especially PS system). If the effect functions are converted into instantiated objects then 2 things are missing: object creation and destruction. When do we create effect object? At JSON parsing time, when effect ID actually change, or later when strip is being servidced? The same goes with destruction. However, when there are transitions active both effects should be running so we cannot destroy previous effect when selecting a new one.

I have mentioned somewhere that current/legacy approach to effects has an issue at the time effect no longer runs. There is no notification to effect function to cleanup after itself (destructor). Currently the `call` variable is the only way for effect to know when it started (or was restared!!!). Restarting would not need entire creation process (constructor), just reiniting some variables. A definition of *when* effect starts and *notification* when effect ceases to run (there will be no more executions of that function) are two missing things to safely store any data from within effect function apart from `_data` location. Being object or not.

All that being said I am not against the change, however, WLED is a very complex system so any change has to be thought over several times.

JoaDick:
> The main issue is that `_data` is raw byte array. Right? Placing objects in it (in current implementation) is prone to errors especially if those objects allocate their own memory. Right?

Exactly, this is a big drawback.

> Rewriting everything from scratch (this or the other PR) aside the easy "solution" is to prohibit memory allocation from within objects placed into `_data` (i.e. using placement new or other more unsafe methods).

I'd not prohibit that at once, but provide alternatives to be able to move away from it step by step.

> This, of course, contradicts PS approach which _does_ need extra memory and has to come with its own solution.

It is essential to remain compatible. Breaking changes that require adjusting "everything" are a no-go. Of course that might make some parts tricky...

> [...] then 2 things are missing: object creation and destruction. When do we create effect object? [...] The same goes with destruction. However, when there are transitions active both effects should be running so we cannot destroy previous effect when selecting a new one.

Exactly, these triggers are something I was looking for a long time ago. And gave up back then...
However, after now digging deeper into the code, it should be possible to implement a workaround - although it's probably also going to be tricky...

> Restarting would not need entire creation process (constructor), just reiniting some variables. A definition of _when_ effect starts and _notification_ when effect ceases to run (there will be no more executions of that function) are two missing things to safely store any data from within effect function apart from `_data` location. Being object or not.

This would surely be helpful. But IMHO rather an optimization, and not vital from the very start.

> All that being said I am not against the change, however, WLED is a very complex system so any change has to be thought over several times.

That's why I try to keep changes small and less intrusive. Better several small changes one after another than on big bang.
Thanks a lot your feedback, I highly appreciate it!

blazoncek:
@TripleWhy I can see your point so I'll cut that short: As I'm working on a feature that *requires* segment copy, this PR *does not* get a green light from me. Others may vote otherwise though.

IMO there should be a new PR that deals with informing effect function when it is first executed and when it is last executed, prior to any other change. The best place to start that hunt is in `resetIfRequired()`. Caveat, that function may be invoked several times during a life of effect function. It is sure to be invoked after the last successful run of effect function before new effect takes place. The other place is `startTranstion()`, this one is sure to be invoked before effect function changes (but can be called prior and after that as well).
That will go hand-in-hand with conversion of effect functions into classes however that is implemented.

JoaDick:
Ack, that's fine for me. Looking forward to seeing your layering branch on main!
Maybe we can then pick this discussion up again; in particular wrt. the raw memory management - which is the main issue, and not necessarily the copying itself.
And thanks for your hints on the triggers for a possible effect base class. I'll definitely have a closer look at them.

netmindz:
Should we resume this discussion now PS and the new blending has been merged?

blazoncek:
blending *requires* segment copies.

PR #4589, opened 2025-03-08
Add warning about unconfigurable outputs

PaoloTK:
Since 0.15 it is now only possible to configure the last output type. This can be confusing for users of pre-assembled boards that ship configured with multiple outputs, so a warning has been added to warn about this behavior.
It's been placed before the first output because some boards are pre-configured with a lot of outputs and thus it would be unlikely for the user to notice the warning since they'd have to scroll all the way to the bottom.
I decided to add a line before the warning because I found it weird to have the title for the section with a colon and then the warning before the outputs without it. It could be added before the title but all other suggestions are below the relevant section.

chrome_9iFrZoipti

### DIFF
--- a/wled00/data/settings_leds.htm
+++ b/wled00/data/settings_leds.htm
@@ -789,9 +789,11 @@ Swap: <select id="xw${s}" name="XW${s}">
 			</div>
 		</div>
 		<h3>Hardware setup</h3>
-		<div id="mLC">LED outputs:<br>
-		<hr class="sml">
-		<i>Only last output can be changed. Remove to edit others.<br></i></div>
+		<div id="mLC">
+			LED outputs:<br>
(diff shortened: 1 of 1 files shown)


wled00/data/settings_leds.htm
```diff
@@ -789,8 +789,9 @@ <h2>LED &amp; Hardware setup</h2>
 			</div>
 		</div>
 		<h3>Hardware setup</h3>
-		<div id="mLC">LED outputs:</div>
+		<div id="mLC">LED outputs:<br>
```
TripleWhy, line 792: proper indentation would be awesome

netmindz (approved):

netmindz:
Thank you for your pr

PR #4594, opened 2025-03-12
Adding perlin noise replacement for fastled functions

DedeHai:
I created integer-math based perlin noise functions to replace the 8-bit optimized fastled inoise functions. 
They execute about twice as fast and use no look-up tables. The functions are slightly different from the originals but I tried to match them as good as possible. Here is a comparison of the outputs:

image

image

saves about 2.7k of flash as a bonus.

These FX use perlin noise, please check if you can spot obvious differences. They are very subtle and I can only spot them in a direct side-by-side comparison.
- Noisefire
- Firenoise
- Midnoise (AR)
- Waverly (AR)
- Rotozoom 
- Sunradiation
- Polarlights
- Plasmaball
- Noise2d
- Metaballs
- Noise pal
- Phased  noise
- Fill noise

## Summary by CodeRabbit

- **New Features**
  - Introduced a new dynamic visual mode ("Perlin Move") for more fluid, organic animations.
  
- **Refactor**
  - Upgraded the noise generation approach across audio reactive and visual effects, enhancing the responsiveness and overall animation quality.

- **Chores**
  - Removed redundant debugging output for a cleaner setup process.


blazoncek:
I would have preferred if you'd use `#define inoise8 perlin8` instead of replacing every `inoise8` function call.

DedeHai:
the final goal is to replace fastled, then it does not make sense to me to keep the legacy function names. 
but actually adding that for compatibility is a good idea.
I also realized I commited changes to rotozoomer that I need to revert...

DedeHai:
did anyone have a chance to test this?
I have the rest of fastled replaced with local functions, some optimizations but this needs to be merged before I can move ahead with that. Once fastled is fully replaced, the next step is to try and optimize for speed by replacing CRGB and CHSV instances wherever possible (and useful) to reduce 24bit <-> 32bit color conversions.

netmindz:
If you are wanting others to test, I think it would be helpful if you could add some notes as what/how to test

Otherwise let's merge in and see if anyone complains 😜

DedeHai:
that is what I was trying with the 

> These FX use perlin noise, please check if you can spot obvious differences.

part in the commit comment :)

netmindz:
I would say the graphs are more accurate way to compare than looking at the effects and they look close enough to my eyes

netmindz (approved):

PR #4596, opened 2025-03-15
Fix wipe effect smoothness

Dschogo:
With update to 15.x many vars got changed to unsigned. Wipe relies on truncation (of an uint16_t) for modulo operation, and was therefore broken.

Using directly an uint16_t like my proposal *should* be overall faster than using an unsigned and then doing a modulo. (Not an expert)


netmindz:
Did you also do the cherry-pick to main @blazoncek ?

blazoncek:
> Did you also do the cherry-pick to main?

I think yes.

Issue #4597, opened 2025-03-16
Compiling does not include usermods


C0oL3r:
### What happened?

When i'm trying to compile WLED usermods are not included in the firmware.

### To Reproduce Bug

If i clone the repository and try to build it like it is it will compile good with audioreactive. But if i add for example Temperature usermod and compile it again it will add Temperature usermod but it will not include audioreactive. 
Then if i remove again Temperature it will not include any usermod anymore. I tried to delete .buildcache, .pio but no luck. Searching for string in the firmware file first compile it shows me about audioreactive, second about the Temperature and the third compile it will not find anything about any usermod.
Also even adding ${common.debug_flags} it will not change anything, the firmware will not be recompiled

### Expected Behavior

It should add the usermods i set to custom_usermods in platformio.ini. And also recompile if anything changes in platformio.ini.

### Install Method

Self-Compiled

### What version of WLED?

Nightly Release 20250316 0.16.0

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

blazoncek:
None of the usermods (except Audioreactive) are automatically compiled with WLED.
If you want to include usermods in your compile, you need to modify platformio.ini (or better override) as specified in usermod's readme file.

C0oL3r:
i did say that i am adding them in platformio.ini, first compile works with audioreactive, second is with audioreactive and Temperature but only Temperature is added, after that none will be added.

blazoncek:
Then perhaps @willmmiles may know more. Reopen if needed.

willmmiles:
I can't replicate this in my development environment; changes to platformio.ini are always picked up on the next build, and seem to build correctly with the specified usermods.

@C0oL3r , are you using an IDE or building with platformio at the command prompt?  What platform is your build system (Linux, Windows, macOS, GitHub codespaces, etc.)?

willmmiles:
Actually, let me hazard another guess.

For this custom environment:
```
[env:my_esp32dev]
extends = env:esp32dev
```
you will get the default usermods with esp32dev (ie. audioreactive).

For this environment spec:
```
[env:my_esp32dev_v2]
extends = env:esp32dev
custom_usermods = Temperature
```
you will get only Temperature; the `custom_usermods` settings has been overridden with the new value specified there.  The inherited value is lost.

For this environment spec:
```
[env:my_esp32dev_v3]
[...]

PR #4609, opened 2025-03-23
Use live cfg json instead of file for usermod settings page

willmmiles:
Ensure that the usermods settings page is using a correct and complete set of configuration items for the currently loaded firmware, regardless of what may have been left behind in `cfg.json` from previous firmware installs.

Implemented by:
- Separating the config the config JSON construction from the filesystem operation; 
- Arranging that the `json/cfg` endpoint uses live constructed JSON output instead of returning the disk file;
- Modifying the usermod settings page to request `json/cfg` instead.

Reference: #4597

## Summary by CodeRabbit

- **Refactor**
  - Redesigned the system’s configuration persistence workflow to improve modularity and ensure that all settings changes (such as WiFi credentials and presets) are reliably saved.
  - Updated the settings retrieval endpoint to streamline data fetching, enhancing consistency and performance.
  - Refined internal processes governing settings updates to further minimize potential inconsistencies, resulting in a more stable operation.

### DIFF
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -671,7 +671,7 @@ void deserializeConfigFromFS() {
     // call readFromConfig() with an empty object so that usermods can initialize to defaults prior to saving
     JsonObject empty = JsonObject();
     UsermodManager::readFromConfig(empty);
-    serializeConfig();
+    serializeConfigToFS();
     // init Ethernet (in case default type is set at compile time)
     #if defined(ARDUINO_ARCH_ESP32) && defined(WLED_USE_ETHERNET)
     initEthernet();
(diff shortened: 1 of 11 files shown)


blazoncek:
One thing of note: The original idea for settings rewrite was to abandon config processing using cfg.cpp/xml.cpp and rely on cfg.json parsing in JavaScript. Usermod setting page was the first implementation of that but further development stalled as @Aircoookie couldn't find time to do it.

AFAIK /json/cfg was intended to allow WLED configuration from external sources like LedFx or xLights, etc.

willmmiles:
> One thing of note: The original idea for settings rewrite was to abandon config processing using cfg.cpp/xml.cpp and rely on cfg.json parsing in JavaScript. Usermod setting page was the first implementation of that but further development stalled as @Aircoookie couldn't find time to do it.

Parsing the configuration JSON directly in the settings pages is a good idea.  The difference here is only in the source: this PR changes it to come from the live memory state, instead of whatever was last written to the filesystem.   Any existing code need only point to `/json/cfg` to get the live state instead, and it will work.   (And of course anyone who wants to read the filesystem state can always continue to use `/cfg.json`.)

blazoncek (approved):
Approved from my side.

PR #4611, opened 2025-03-25
Fix unused variabe compiler warning

marcone:
Variable is only used in debug logs, so is unused when compiled without debug.

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -751,7 +751,7 @@ void WLED::handleConnection()
   static bool scanDone = true;
   static byte stacO = 0;
   const unsigned long now = millis();
-  const unsigned long nowS = now/1000;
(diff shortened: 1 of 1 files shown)


wled00/wled.cpp
```diff
@@ -751,7 +751,7 @@ void WLED::handleConnection()
   static bool scanDone = true;
   static byte stacO = 0;
   const unsigned long now = millis();
-  const unsigned long nowS = now/1000;
+  [[maybe_unused]] const unsigned long nowS = now/1000;
```
DedeHai, line 754: why not wrap it in an ifndef instead?

marcone: Personal preference, plus it matches the style of existing code in other files.

blazoncek:
This part of the code is going to change tremendously once I finish ESP-NOW sync code so I vote against merge to prevent conflicts in the future.

PR #4617, opened 2025-03-27
Fix: Update platformio_override.sample.ini

huckley:
Add custom_usermods to 433MHz RF example

### DIFF
--- a/platformio_override.sample.ini
+++ b/platformio_override.sample.ini
@@ -534,5 +534,6 @@ monitor_filters = esp32_exception_decoder
 [env:esp32dev_usermod_RF433]
 extends = env:esp32dev
 build_flags = ${env:esp32dev.build_flags} -D USERMOD_RF433
+custom_usermods = usermod_v2_RF433
 lib_deps = ${env:esp32dev.lib_deps}
   sui77/rc-switch @ 2.6.4


DedeHai:
thanks for the updated example. can ou remove the old definition while you are at it?

platformio_override.sample.ini
```diff
@@ -533,6 +533,7 @@ monitor_filters = esp32_exception_decoder
 # 433MHz RF remote example for esp32dev
 [env:esp32dev_usermod_RF433]
 extends = env:esp32dev
-build_flags = ${env:esp32dev.build_flags} -D USERMOD_RF433
+build_flags = ${env:esp32dev.build_flags}
+custom_usermods = usermod_v2_RF433
 lib_deps = ${env:esp32dev.lib_deps}
```
Arcitec, line 538: The entire `lib_deps` statement should be deleted too. That's handled automatically by the PlatformIO Libraries feature we use now.

In this case it's handled here now:

https://github.com/wled/WLED/blob/main/usermods/usermod_v2_RF433/library.json

Arcitec (changes requested):

platformio_override.sample.ini
```diff
@@ -533,6 +533,7 @@ monitor_filters = esp32_exception_decoder
 # 433MHz RF remote example for esp32dev
 [env:esp32dev_usermod_RF433]
 extends = env:esp32dev
-build_flags = ${env:esp32dev.build_flags} -D USERMOD_RF433
+build_flags = ${env:esp32dev.build_flags}
```
Arcitec, line 536: This is actually redundant now and would work equally well if deleted. The `${env:esp32dev.build_flags}` is however vitally important if the user wants to add WLED build flags that configure the usermod, so the example should still stay.

But maybe change it to this:

```ini
;; Uncomment and append your own build flags here if you want to configure the usermod:
; build_flags = ${env:esp32dev.build_flags}
``

Arcitec (changes requested):

PR #4618, opened 2025-03-27
add example demonstrating advanced PlatformIO environment inheritance

Arcitec:
Adds a self-documented example demonstrating how a user can inherit from a core WLED configuration and add their own values without needing to copy all of the upstream values manually. This new method simplifies configuration maintenance for the users.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new configurable environment for ESP32 Ethernet devices that inherits settings from the default configuration.
  - Updated build settings now allow the customization of the release name and relay pin, ensuring optimal configuration without conflicts.

### DIFF
--- a/platformio_override.sample.ini
+++ b/platformio_override.sample.ini
@@ -538,9 +538,19 @@ lib_deps = ${env:esp32dev.lib_deps}
   sui77/rc-switch @ 2.6.4
 
 # ------------------------------------------------------------------------------
-# Advanced example: Inheriting all settings from an upstream "base" config, and
-# adding/overriding specific build flags, which is useful if you want to base
-# your environment on a stable core env but just override certain WLED defaults.
+# Advanced example: Inheriting all settings from an upstream WLED "release" config,
+# and adding/overriding specific build flags, which is useful if you want to base
+# your environment on a stable WLED release env but easily override some of its
+# defaults, such as the output pin numbers, the default ethernet board ID, etc.
+#
+# WARNING: The WLED project does NOT guarantee that your chosen "upstream" config
+# will exist or be maintained in the future. Environments can sometimes change
+# radically. The compiler will warn you if the upstream configuration doesn't
+# exist anymore, but it won't warn you if it's no longer maintained or has added
+# new flags that may impact your use-case. It is YOUR responsibility to periodically
+# check the status and any new defaults for your chosen upstream configuration.
+# However, it's *generally* safe to pick a well-known, popular release configuration
+# and tweak a few commonly used defaults such as its pin numbers and modes!
 # ------------------------------------------------------------------------------
 
 [env:esp32_eth_custom]
@@ -550,9 +560,10 @@ lib_deps = ${env:esp32dev.lib_deps}
 extends = env:esp32_eth
 
 ;; Remove conflicting upstream flags to fix compiler "variable redefined" warnings.
-;; WARNING: Must specify the EXACT upstream "build_flags" flag VALUES to unset;
-;; because if we don't specify any value then it unsets ALL instances of the
-;; flag, meaning that it would erase our new values too.
+;; WARNING: Must specify the EXACT upstream "build_flags" flag VALUES to unset.
+;; If no value is specified, then ALL values for that flag will be removed, which
+;; also deletes any new values you're trying to set. Also be aware that the
+;; compiler deletes ALL instances of the EXACT VALUES you specify below.
 ;; NOTE: If the upstream values are ever changed, the compiler will encounter
 ;; duplicate values and complain about "variable redefined" again, meaning that
 ;; these values will simply need to be revised. So maintenance is safe and easy.
@@ -567,5 +578,7 @@ build_unflags = ${env:esp32_eth.build_unflags}
 build_flags = ${env:esp32_eth.build_flags}
   ;; Set/add your new flag values:
   -D WLED_RELEASE_NAME=\"ESP32_Ethernet_Custom\"
-  -D RLYPIN=24
+  -D RLYPIN=15           ;; Relay pin.
+  -D WLED_ETH_DEFAULT=4  ;; "QuinLed-ESP32-Ethernet"
+  -D DATA_PINS=16,3,1,4  ;; LED output channels.
 
--- a/platformio_override.sample.ini
+++ b/platformio_override.sample.ini
@@ -536,3 +536,36 @@ extends = env:esp32dev
 build_flags = ${env:esp32dev.build_flags} -D USERMOD_RF433
 lib_deps = ${env:esp32dev.lib_deps}
   sui77/rc-switch @ 2.6.4
+
+# ------------------------------------------------------------------------------
+# Advanced example: Inheriting all settings from an upstream "base" config, and
+# adding/overriding specific build flags, which is useful if you want to base
+# your environment on a stable core env but just override certain WLED defaults.
+# ------------------------------------------------------------------------------
+
+[env:esp32_eth_custom]
+;; Inherit all settings from upstream config. We can then override specific settings.
+;; WARNING: The upstream environment's name, such as "esp32_eth" in this example,
+;; must also be specified in the other inherited settings (see example below).
+extends = env:esp32_eth
+
+;; Remove conflicting upstream flags to fix compiler "variable redefined" warnings.
+;; WARNING: Must specify the EXACT upstream "build_flags" flag VALUES to unset;
+;; because if we don't specify any value then it unsets ALL instances of the
+;; flag, meaning that it would erase our new values too.
+;; NOTE: If the upstream values are ever changed, the compiler will encounter
+;; duplicate values and complain about "variable redefined" again, meaning that
+;; these values will simply need to be revised. So maintenance is safe and easy.
+build_unflags = ${env:esp32_eth.build_unflags}
+  ;; Unset the old, conflicting flag values:
+  -D WLED_RELEASE_NAME=\"ESP32_Ethernet\"
+  -D RLYPIN=-1
+
+;; Inherit all upstream flags and then overwrite/append with some custom values.
+;; NOTE: You can add/modify ANY of the WLED build-time variables here (it doesn't
+;; have to be values that were in the original upstream config).
+build_flags = ${env:esp32_eth.build_flags}
+  ;; Set/add your new flag values:
+  -D WLED_RELEASE_NAME=\"ESP32_Ethernet_Custom\"
+  -D RLYPIN=24
+


Arcitec:
Thanks Mr. Bot for finding a small typo. Fixed. :)

DedeHai:
thanks, 
please don't force push branches in PR, it can mess up revies (nothing bad happened just as a general rule).

@willmmiles can you check the example is correct please? I am not too familiar with the inner workings of envs.

Arcitec:
Hey Dede! :) I agreed with Mr. Bot about adding two more comment lines to super clearly delineate where the old values should be unset/set. Unfortunately I had pushed like 10 seconds before I saw your comment. Apologies. Well, the diff is here:

https://github.com/wled/WLED/compare/378fa59596976a83d0977dfde34aedf06307c2a7..c0d6cfcfb2d87e24a934bfbfc2959a928fe9d59b

As for checking if the example is correct, first go into VSCode preferences (`Ctrl + ,`) and search for "scrollback" and set it to 100000 lines. Then open the PlatformIO sidebar (left side-panel of VSCode) and go to "Advanced: Verbose Build". This will build with output showing all the flags that are being used for each file being compiled.

You can then see that the old flags are removed and the new flags are set.

I spent about 2 hours going through PlatformIO's code (seeing how `build_unflags` is being processed by PlatformIO) and deeply testing this to ensure the final result is correct. Nobody had documented this usage anywhere even in PlatformIO's own discussions or docs. Took a lot of time but I definitely ensured it works correctly. :)

Arcitec:
Here are some of the references I used (useful for history/PR documentation):

- How interpolation works for referring to other environment values: https://docs.platformio.org/en/latest/projectconf/interpolation.html
- `build_flags`: https://docs.platformio.org/en/stable/projectconf/sections/env/options/build/build_flags.html
- `build_unflags` (which is barely documented at all and was a pain): https://docs.platformio.org/en/stable/projectconf/sections/env/options/build/build_unflags.html
- Seeing how `ProcessUnflags` removes flags (which is how I noticed that it looks for an exact match): https://github.com/platformio/platformio-core/blob/develop/platformio/builder/tools/piobuild.py#L255
- And a worthy sidenote: PlatformIO doesn't care about formatting when comparing flag equality. `-D RLYPIN=-1` and `-DRLYPIN=-1` are treated as identical (note the lack of space in one example). It parses all build flags into a format that lets it compare regardless of whitespace (and I assume quotation differences too). But the simplest way to specify the correct `build_unflags` as a user is simply to Ctrl-C and Ctrl-V the original rows from the upstream config.

platformio_override.sample.ini
```diff
@@ -536,3 +536,36 @@ extends = env:esp32dev
...
+  -D RLYPIN=-1
+
+;; Inherit all upstream flags and then overwrite/append with some custom values.
+;; NOTE: You can add/modify ANY of the WLED build-time variables here (it doesn't
+;; have to be values that were in the original upstream config).
+build_flags = ${env:esp32_eth.build_flags}
+  ;; Set/add your new flag values:
+  -D WLED_RELEASE_NAME=\"ESP32_Ethernet_Custom\"
+  -D RLYPIN=24
+
```
willmmiles, line 571: Given that we have to copy-and-paste parts of the original definition in to our new environment, I think it would be clearer and easier to understand if we were to copy and modify the full definitions instead of using `build_unflags`.

Eg. given:
```
[env:esp32_eth]
board = esp32-poe
platform = ${esp32.platform}
upload_speed = 921600
custom_usermods = audioreactive
build_flags = ${common.build_flags} ${esp32.build_flags} -D WLED_RELEASE_NAME=\"ESP32_Ethernet\" -D RLYPIN=-1 -D WLED_USE_ETHERNET -D BTNPIN=-1
;  -D WLED_DISABLE_ESPNOW ;; ESP-NOW requires wifi, may crash with ethernet only
lib_deps = ${esp32.lib_deps}
```

I think this:
```
[env:esp32_eth_custom]
extends = env:esp32_eth
; All build flags were copied from the original environment, then modified here
; So I can see exactly the flags I'm going to get
build_flags = ${common.build_flags} ${esp32.build_flags}
  -D WLED_RELEASE_NAME=\"ESP32_Ethernet_Custom\"
  -D RLYPIN=24
  -D WLED_USE_ETHERNET -D BTNPIN=-1
```

is simpler and clearer than this:
```
[env:esp32_eth_custom]
extends = env:esp32_eth
build_unflags = ${env:esp32_eth.build_unflags}  ; Have to reference the original build_unflags
   ; And also have to copy-and-paste in the ones I want to modify
  -D WLED_RELEASE_NAME=\"ESP32_Ethernet\"
  -D RLYPIN=-1
build_flags = ${env:esp32_eth.build_flags}
   ; Now I've had to mention these flags twice
  -D WLED_RELEASE_NAME=\"ESP32_Ethernet_Custom\"
  -D RLYPIN=24
```

My experience with end-users and scripting languages (I maintain a fairly complex scripting solution in my day job) is that our users found "copy and modify" (or better yet, "modify in place") much easier to understand than more sophisticated solutions, even when the sophisticated solutions offered better future-proofing.

Arcitec: @willmmiles Every other example in the `platform_overrides.sample.ini` file already does exactly what you suggested - which is to extend and entirely replace values with copy-pasted values.

What you get then is a very clunky, verbose config that must be maintained by the user in the future. With that technique, the user must **manually and relentlessly keep up with any changes upstream**, manually looking at all code changes in WLED to see if their chosen upstream profile's `build_flags` ever changes.

And will an upstream profile ever change? Yes it will. Like it just did in WLED 0.16.0, when `audioreactive` was converted to a library and as removed from the build flags of `esp32_eth` and all the other core environments.

If a user uses your technique, they'll have all kinds of bugs and compiler issues unless they relentlessly monitor all upstream changes.

The literal purpose of my new technique is to not have to copy-paste anything. The new technique compiles equally (perfectly) on 0.15.0, 0.16.0, etc, with *zero* changes in the user's own platform overrides. It automatically receives the latest upstream fixes and improved arguments with no maintenance by the user.

I know that the new technique requires the user to understand how it works. Which is why it's self-documented with heavy commenting. But in actual usage it's very simple:

- Extend something to inherit everything from that core environment.
- Set your new flags after the `;; Set/add your new flag values:` line.
- Unset any flags that caused compiler warnings due to already being set upstream. If there are no clashes, the user could delete the entire `build_unflags` from their downstream config. If there are clashes, they must unset the old values, which is a simple copy-paste of the specific upstream setting-line that they wanted to overwrite.
- And regarding knowing exactly what flags are being applied: You know that already by looking at the upstream config. But that's not interesting to most people. They just want to base their own downstream config on an upstream preset and add their own slight settings tweaks. That's what this new technique solves perfectly.

Here's the thing. There's great value in having a technique that just automatically imports everything from the upstream config without causing bugs for users when WLED evolves. Like it just did in 0.16.0 with the usermod system changes.

I hope this explains why the example is very useful. It's marked advanced and is deeply documented for a reason, but it's by far the best way to create derived profiles for most people. The alternative (copy-pasting the entire upstream flags) requires vigilance and obsessive observations by the users, unfortunately.

Neither solution is perfect, but this provides another way to work with PlatformIO in a safer manner.

willmmiles: >  The new technique compiles equally (perfectly) on 0.15.0, 0.16.0, etc, with zero changes in the user's own platform overrides.

However it wouldn't work with 0.14 to 0.15 -- the syntax for `WLED_RELEASE_NAME` changed, so the `build_unflags` wouldn't match.  Anything put in `build_unflags` is no less vulnerable to being changed out from underneath a custom environment in some future version.

From my perspective, it looks like asking people to take on understanding extra complexity for, in the end, no stronger guarantees.

blazoncek:
For what it's worth, *my own* opinion:
`platformio.ini` provides *base* settings for each of the supported chips in `[esp8266]` or `[esp32]`, etc. sections. These sections are provided with intent to be used in custom environments (in `platformio_override.ini`) as well as in environments provided by WLED (for the purposes of CI and official releases) and include minimum amount of settings to compile for certain chip.
These are merely *sections* not entire environments and don't do anything on their own.
The idea to *reuse* one of the (CI/official) build environments is flawed as environments may change (as @willmmiles pointed out) without prior notice while base *sections* are guaranteed to include minimum amount of settings for successful compile. I also fully agree with what @willmmiles wrote regarding cleanness or readability.
IMO suggesting users to reuse official (full) *environments* and overriding certain settings is not for first-time comers but for experienced developers that *know* the implications. Those would already know the technique.

That being said, the change is benign and only boils down to the question: Are we willing to deal with users complaining/asking why their custom builds fail out of the sudden when they pull from upstream since they followed instructions provided by WLED?

If the answer is yes then I'd suggest to add a disclaimer that environments may change without notice and such change may render derived environment non-functional.

Arcitec:
@blazoncek It's already marked as an advanced method in the first line of the example. :)

I agree 100% that there should be a clear warning in the example code which explains that the upstream environment may change or become unmaintained in the future. I will add that!

That "feature" is the entire reason for its existence though: Seeing a good release build (such as "ESP32_Ethernet") and tweaking a few WLED defaults for yourself. For most users, such a usage will not need any maintenance at all because they'll only be setting a few pin numbers, default ethernet board model, etc.

The alternative is to copy-paste an entire definition and maintain it manually and relentlessly.

I just need to explain more clearly the upsides and downsides of this approach in the comments itself.

blazoncek:
> The alternative is to copy-paste an entire definition and maintain it manually and relentlessly.

What is the difference compared to extending existing core environment? You will still need to "maintain it manually and relentlessly".

I understand that this approach may work for you (until we change `esp32_eth`) but I am against suggesting such approach in general.

netmindz:
Thank you for raising this @Arcitec 

I agree that we want to make it easy for users to base their builds of a known good set of values, which may change and we don't want them to have to track these changes so extends is the right approach 

I need to take some time to review our current config as well as your pr so that the best path for the user is also the easy path

Arcitec:
@netmindz That's excellent to hear! If you have any concerns or questions, I'm here for you and ready to revise anything! :)

The general approach requires all 3 parts:

- `extends` inherits 100% of the values from the upstream config. But any *other* fields that are set manually in the derived config **will overwrite** the upstream config.
- `build_flags` therefore uses [interpolation](https://docs.platformio.org/en/latest/projectconf/interpolation.html), which just means that we refer to the upstream config as our first line, which then recursively resolves the entire upstream config's field value. So in short, `build_flags = ${env:esp32_eth.build_flags}` means "Set it to the exact same value as upstream". After that, we append our own flags/overrides _after_ that line.
- `build_unflags` uses interpolation too, meaning that `build_unflags = ${env:esp32_eth.build_unflags}` sets it to the exact same value as upstream. We then append the other flags that we want to unset: Values that we wanted to change but already existed in the upstream `build_flags`, since the compiler doesn't allow people to define a flag multiple times and would spit out "variable redefined" warnings endlessly. So we must unset the conflicting (old) values. The "unflags" in PlatformIO works by scanning for the flag (or the exact _value_ match if provided) and deleting that from the final flags. This is the only way to get rid of the duplicate flags. It's easy to populate the list though, because the user can just open the upstream config and copy-paste the specific `build_flags` values they want to remove. And if those ever change to another value in the future, the compiler will again start complaining about duplicate flags, which is a very visible alert that the user just needs to copy-paste the new upstream value into unflags (this is documented in the example).

To verify that the correct flags are being applied, use this technique to do a verbose build to see all flags being applied:

https://github.com/wled/WLED/pull/4618#issuecomment-2759177352

Hope this helps! Let me know if there's any questions.

Arcitec:
Further activity occurs.

softhack007:
@netmindz please take a look.

I think the example is very basic, so good for me to add this into the examples file. I just find the comments really exhaustive, like ~40 lines of explanation for a buildenv with 7 line. Maybe we can shorten the text a bit?

Other than that - OK to merge.

softhack007 (approved):
Text could be shorter, but technically the example is correct.

DedeHai:
+1 on shorter text, anyone using this should be at least somewhat familiar with coding and a lot of info in the text is just common sense in that case.

PR #4624, opened 2025-03-29
fix timer overflow bug in handleRandomPalette()

DedeHai:
timer overflow was not handled properly: due to a misplaced bracket, the palette would stop updating after about 18hours. this fixes it and makes the code more clear. 

The fix would also work for 0.16 but @blazoncek has a more elaborate fix.


blazoncek:
This should not be included in 0.16 or above as I've added some more functionality that accompanies this fix for 0.16.

PR #4635, opened 2025-04-08
added new usermod photoresistor_sensor_mqtt_v2 with hass integration, improved hass intgeration for PIR_sensor_switch, fixed compile warnings 

rommo911:
added new usermod for ldr with hass mqtt integration :
- photoresistor_sensor_mqtt_v2 : 
 features : it will publish auto discovery message to hass mqtt and attach the light sensor to the wled integration automatically with good options 
pir switch usermod : added support for discovery message to integrate mqtt entity into wled integration ( check photo ) 
hass_ldr_mqtt
othert changes : fixed signed/unsigned comparasion compiler warning , added to gitignore local buildcache folder ( when desired )


usermods/photoresistor_sensor_mqtt_v2/library.json
```diff
@@ -0,0 +1,3 @@
+{
+  "name:": "photoresistor_sensor_mqtt_v2"
+}
```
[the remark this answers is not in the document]
rommo911: false error ?

DedeHai:
thank you for contributing.
the UM looks ok, please revert changes to other files, make them seperate PRs.

rommo911:
@DedeHai  thanks for the review ,i rebased it to have the UM as a single commit. tested on esp32 & esp8266 nodmcu & esp_01

DedeHai:
please dont force push. there are still changed files not part of the UM

rommo911:
alright. 
yes all UM are good here.

DedeHai:
no, you still mixed two PRs and changed unrelated files.

rommo911:
> no, you still mixed two PRs and changed unrelated files.

sorry , my bad. now should be good

usermods/PIR_sensor_switch/PIR_sensor_switch.cpp
```diff
@@ -317,7 +317,9 @@ void PIRsensorSwitch::publishHomeAssistantAutodiscovery()
     device[F("mf")]   = F(WLED_BRAND);
     device[F("mdl")]  = F(WLED_PRODUCT_NAME);
     device[F("sw")]   = versionString;
-    
+    JsonArray connections = device[F("connections")].createNestedArray();
```
DedeHai, line 320: this is related to the new UM how?

DedeHai:
I took a closer look at the UM. If the purpose of it is to read an analog value and send it to MQTT why not just name it that? It is not limited to a photoresistor even if that is what you use it for. It would be much better suited as a general "read analog input, publish the value", it could even be extended to multiple pins. I did not check if any usermod already does something like this.

usermods/PIR_sensor_switch/PIR_sensor_switch.cpp
rommo911: this PR has two UM modifs
1- added new usermod photoresistor_sensor_mqtt_v2 with hass integration auto discovery with wled integration (  the new UM ) 

2- improved hass integration for PIR_sensor_switch ( enable integrating the pir sensor in the wled hass device ; see the photo , previously the auto discovery does not support putting the mqtt entity of the pir into wled device ) , hence this change.

rommo911: let me know if this modification should go in a seperate PR ?

rommo911:
> I took a closer look at the UM. If the purpose of it is to read an analog value and send it to MQTT why not just name it that? It is not limited to a photoresistor even if that is what you use it for. It would be much better suited as a general "read analog input, publish the value", it could even be extended to multiple pins. I did not check if any usermod already does something like this.

it can be adapted as general purpose analog read, but the UM aims to support hass wled integration via mqqt as light sensor with auto discovery. so if its not used as light sensor, hass discovery needs adjustement.
1- multiple pins can be supported yes 
2- usage as general purpose can be also supported but needs to add more customization options for hass discovery .
but the main goal is to use it with LDR as sensor for hass users so i think this could be very useful to many peopole and it can be adjusted or improved next.

usermods/PIR_sensor_switch/PIR_sensor_switch.cpp
DedeHai: yes, don't mix different goals in on PR

DedeHai:
I am unfamiliar with how HA works. If you made this into a more generally useable UM, there could be any analog sensor attached. If HA uses a particular structure: adding a dropdown to the UM config to select what type HA should be detecting would be a good option. 
The multiple pin options is something I see optional and also you could add that later.

DedeHai:
I read the code some more. 
- formating could be streamlined to rest of WLED code (similar to google format)
- you forgot to add a pin selection
- this would be the third usermod for photoresistors, to get my vote to add it, it needs to bring configurability, otherwise it does almost exactly the same as existing "V1" apart from added HA specific things.

@blazoncek since you are more familiar with UMs capabilities, would you care to take a look?

usermods/PIR_sensor_switch/PIR_sensor_switch.cpp
blazoncek: This needs to be removed from this PR

usermods/photoresistor_sensor_mqtt_v2/README.md
```diff
@@ -0,0 +1,37 @@
...
+Setting | Description | Default
+--- | --- | ---
+Enabled | Enable/Disable the LDR functionality. | Disabled
+LDR Pin | The analog capable pin your LDR is connected to. | A0
+LDR update interval | update interval to read the analog pin ( in ms ) | 3000
+LDR inverted | invert readings based on your own hardware setup ( invert mapping analog to digital value to  0V = 100%  or 3.3V = 100% ) | false
+change threshould | the minimum limit to detect change and publish the value (in % ; float ; ex 0.8 ) . | 1.0
+
+## Author
+[@jeffrommo911wdh](https://github.com/rommo911)
```
blazoncek, line 25: Username needs fixing.

usermods/photoresistor_sensor_mqtt_v2/photoresistor_sensor_mqtt_v2.cpp
```diff
@@ -0,0 +1,204 @@
...
+    top["change threshould"] = change_threshold;
+  }
+
+  bool readFromConfig(JsonObject &root)
+  {
+    int8_t oldLdrPin = ldrPin;
+    JsonObject top = root[FPSTR(_name)];
+    bool configComplete = !top.isNull();
+    configComplete &= getJsonValue(top["Enabled"], ldrEnabled);
+    configComplete &= getJsonValue(top["LDR Pin"], ldrPin);
```
blazoncek, line 110: Even though it is (currently) supported, I would avoid spaces in JSON key names.

usermods/photoresistor_sensor_mqtt_v2/photoresistor_sensor_mqtt_v2.cpp
```diff
@@ -0,0 +1,204 @@
...
+    configComplete &= getJsonValue(top["LDR update interval"], update_interval);
```
blazoncek, line 111: These are `cfg.json` configuration item key names. They need not be descriptive. Use `addInfo()` to add description.

usermods/photoresistor_sensor_mqtt_v2/photoresistor_sensor_mqtt_v2.cpp
```diff
@@ -0,0 +1,204 @@
...
+    configComplete &= getJsonValue(top["LDR inverted"], inverted);
+    configComplete &= getJsonValue(top["change threshould"], change_threshold);
+    // pin changed - un-register previous pin, register new pin
+    PinManager::deallocatePin(oldLdrPin, PinOwner::UM_PHOTO_RESISTOR_MQTT);
+    if (ldrEnabled && ldrPin > 0)
```
blazoncek, line 116: Is a 0 invalid entry?

usermods/photoresistor_sensor_mqtt_v2/photoresistor_sensor_mqtt_v2.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+class photoresistor_sensor_mqtt_v2 : public Usermod
+{
+private:
+  unsigned long update_interval = 3000;
+  float change_threshold = 1;
+  static const char MQTT_TOPIC[];
+  static const char _name[];
+  const bool HomeAssistantDiscovery = true; // is HA discovery turned on by default
+  int ldrPin = A0;
```
blazoncek, line 11: Please use actual GPIO number and not Arduino pin monikers.

blazoncek (commented):
I do not like duplicates and would have preferred older usermods be updated. This will allow existing users to benefit from the updated code without modifying their environments.

usermods/photoresistor_sensor_mqtt_v2/README.md
rommo911: i'll put it in a seperate PR . its very useful for hass integration

rommo911:
> I do not like duplicates and would have preferred older usermods be updated. This will allow existing users to benefit from the updated code without modifying their environments.

alrighty , i am still new to WLED PR. i use it very much around my house with several UM & hass integration ( i have almost all versions running in my house , ESP01 / ESP8266 / ESP32 / ESP32-S2  ESP32-S3 . 

> I am unfamiliar with how HA works. If you made this into a more generally useable UM, there could be any analog sensor attached. If HA uses a particular structure: adding a dropdown to the UM config to select what type HA should be detecting would be a good option. The multiple pin options is something I see optional and also you could add that later.

i am working on modifying it to be as  @DedeHai  , analog sensor with multiple pins support & hass general WLED integration to be most correct, current mqtt hass dicovery add a seperate mqtt device that is identified outside the WLED device in hass. this integration would put the mqtt entity inside the WLED device menu in hass. 

TODO : 
1- add support for multiple pins 
2- rename to generic ADC mqtt sensor 
3- streamline json & config . 
4- add hass discovery option to UM settings. 
5- fix pins issues

rommo911:
@blazoncek i have issue with choosing pin A0 via pinManager , actually A0 is pin number 17 . and pin manager refuses to allocate pin 17 as its ONLY ADC pin and in the code its explicitly specified to return false if pin >=17 for ESP8266 ... 
should we consider also adding support for pin manager to allocate pin 17 for esp8266 in input only mode ?  or should i do a workaround only in the UM ?

DedeHai:
I think a more or less generic analog sensor UM with MQTT and proper HA support would be greatly appreciated by other users. There have been several requests in this direction.

blazoncek:
ESP8266 only has one (un-modifiable) ADC pin and is so pointless to allow selection of it.
I would recommend you ignore any input in such case.
For ESP32 (and variants) that does not apply.

usermods/PIR_sensor_switch/PIR_sensor_switch.cpp
rommo911: alright.

rommo911:
> I think a more or less generic analog sensor UM with MQTT and proper HA support would be greatly appreciated by other users. There have been several requests in this direction.

on it .

rommo911:
its looking like this at this moment 

image

for ESP8266 , removed pin choice and forced A0=17 , ignore changes on pin. 

HASS detect the sensor class & its unit. 
added option to disable hass discovery, added option to publish either % of ADC ( mapping to 100 ) or publish raw value. 
tested on ESP32 & ESP8266 ... 
for ESP32 only ADC 1 channels are allowed ( as wifi uses ADC 2 ) 

TODO : gotta check with pin for ESP32-C3 / S2 / S3 to ensure appropriate ADC pins are allowed,

would appreciate your feedback or advices while working on it to consider other aspects.   

spoiler alert : the UM setup / config / discovery is rather heavy complicated code but since its for ESP series , it can be handled quit good i guess.

rommo911:
> I think a more or less generic analog sensor UM with MQTT and proper HA support would be greatly appreciated by other users. There have been several requests in this direction.

it is ready now, i tested it well on multiple targets :) . let me know you opinion

usermods/adc_sensor_mqtt/adc_sensor_mqtt.cpp
```diff
@@ -0,0 +1,418 @@
...
+#else
+      adc_pin[i] = -1;
+#endif
+      adc_value[i] = 0;
+      adc_Percentage[i] = 0;
+      adc_last_value[i] = 0;
+      device_class[i] = "voltage"; // default device class
+      unit_of_meas[i] = "V";       // default unit of measurement
+    }
+    // customize here your hass adc device class and unit of measurement
```
DedeHai, line 66: What device classes are there? Would be great to have a dropdown menu in UM settings to select class for each input

DedeHai:
generally looks really good, I have no means to test any of it though.

usermods/adc_sensor_mqtt/adc_sensor_mqtt.cpp
rommo911: i though about it, they are so MANY with possible same class and different measurement units take a look at here  
https://www.home-assistant.io/integrations/sensor/#device-class

rommo911:
> generally looks really good, I have no means to test any of it though.

i tested it on 5 targets normal ESP32 and C3/S2/S3 and esp8266. no monkey business found since couple of days in my setup.

usermods/adc_sensor_mqtt/adc_sensor_mqtt.cpp
DedeHai: how about picking the 5 most relevant for WLED applications and add those? and add a link in the source code to the full list for users that want to implement something different?

rommo911: yeah that could be a solution. 
i never implemented a drop down list of items .... care to give me a link to a good example ?

DedeHai: sounds good.
I did a drop down list in deep-sleep usermod but I think that is only numbers and I suck at java/html so can't really help you there.

DedeHai:
is the change threshold in ADC bits? would that be better to do in % or something more accessible to the "unknowing" user?

rommo911:
> is the change threshold in ADC bits? would that be better to do in % of full scale or something more accessible to the "unknowing" user?

Yeah i though about it but it can be set to percentage but in this case i want it to be float value,  example , getting change threshold to 0.5% or 2.5% , or in this case the default mapped value. 
Let me change it to float and test it ans specify in the ui also how this behave. 
Any other things to consider ?

DedeHai:
@netmindz code looks good to me, but since I do not HA nor MQTT I can't say if its working correctly. I would be ok to merge as-is after conflicts are resolved but having someone test it would be preferred.

DedeHai:
@rommo911 if you are still interested in bringing this upstream we can reopen but so far no help arrived in testing this.

Issue #4634, opened 2025-04-08
AJ-SR04M sensor doesn't work with Animated Starcase usermod


lexxnsk:
### What happened?

Hello.
I have AJ-SR04M with 2 pins. It works fine with ESP8266 and some basic Code from Arduino IDE, but doesn't work with WLED Animated Staircase usermod. At the same time regular HC-SR04 works fine with WLED. I really appreciate if you can help me. Thanks.

### To Reproduce Bug

Connect AJ-SR04M sensor  - doesn't work.
replace with HC-SR04 - works

### Expected Behavior

Works without any differences

### Install Method

Self-Compiled

### What version of WLED?

0.15

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

blazoncek:
Does it say this sensor is supported?
If not, you'll need to contact usermod creator and request support.

lexxnsk:
@willmmiles hello, may I ask you to have a look?

lexxnsk:
@realrolfje hello, may I ask you to have a look?

lexxnsk:
#4637 https://github.com/wled/WLED/pull/4637

lexxnsk:
Hello. I've fixed the issue. Please, have a look and approve this PR https://github.com/wled/WLED/pull/4637
Thanks.

PR #4637, opened 2025-04-09
UM animated staircase: fix for AJ-SR04M sensor

lexxnsk:
When using AJ-SR04M sensor, pulseIn function always returns 0 when timeout value is about 5-10 thousands. I don't know why.
With HC-SR04 behavior is different - it returns real value. So to support both sensors, I've added this additional logic:
      unsigned long duration = pulseIn(echoPin, HIGH, 30000);
      return (duration > 0 && duration < maxTimeUs);

## Summary by CodeRabbit

- **Refactor**
  - Improved ultrasonic sensor detection for more accurate and reliable measurements.
  
- **Style**
  - Applied minor formatting enhancements.

### DIFF
--- a/usermods/Animated_Staircase/Animated_Staircase.cpp
+++ b/usermods/Animated_Staircase/Animated_Staircase.cpp
@@ -152,6 +152,7 @@ class Animated_Staircase : public Usermod {
       delayMicroseconds(2);
       digitalWrite(signalPin, HIGH);
       delayMicroseconds(10);
+      digitalWrite(signalPin, LOW);
       unsigned long duration = pulseIn(echoPin, HIGH, 30000);
       return (duration > 0 && duration < maxTimeUs);
     }
--- a/usermods/Animated_Staircase/Animated_Staircase.cpp
+++ b/usermods/Animated_Staircase/Animated_Staircase.cpp
@@ -152,8 +152,8 @@ class Animated_Staircase : public Usermod {
       delayMicroseconds(2);
       digitalWrite(signalPin, HIGH);
       delayMicroseconds(10);
-      digitalWrite(signalPin, LOW);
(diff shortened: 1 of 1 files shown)


lexxnsk:
[#](https://github.com/wled/WLED/issues/4634)

lexxnsk:
https://github.com/wled/WLED/issues/4634

usermods/Animated_Staircase/Animated_Staircase.cpp
```diff
@@ -152,8 +152,8 @@ class Animated_Staircase : public Usermod {
       delayMicroseconds(2);
       digitalWrite(signalPin, HIGH);
       delayMicroseconds(10);
-      digitalWrite(signalPin, LOW);
```
blazoncek, line 155: This line is required to drop signal pin LOW and end transmission. Please revert.

blazoncek (changes requested):
Please thoroughly check on all supported sensors.

usermods/Animated_Staircase/Animated_Staircase.cpp
lexxnsk: you're right.
It was mistakenly deleted

lexxnsk: I've reverted this line and committed this fix as well. Please, have a look.

blazoncek (approved):
Currently I have no way of testing it but I think adding 30ms wait for response may be a bit long and can produce visible hiccups in LED display (default FPS is 42 or about 23ms between frames).

Even though approved I would be cautius with the merge and would require thorough testing before doing so (using both sensors).

lexxnsk:
I will test it in coming weeks. I will update you with the results within this thread.
Thanks.

DedeHai:
@lexxnsk still waiting on your feedback.

Issue #4647, opened 2025-04-16
Syslog / Net debug - several logs created because of missing hostname


KrX3D:
### What happened?

Hello, i use the syslog / net debug feature.

I can set those options in my syslog server for "System identifier for logfile name"

![Image](https://github.com/user-attachments/assets/b3fa5845-5b76-495f-83b2-4e3e1b6e0b17)

For the first and last everything is ok, but when i use "Hostname (from syslog message)" which i use mostly i get several log files created because the hostname is missing after the time

![Image](https://github.com/user-attachments/assets/6eb17b49-017c-476c-932f-80eaaf265c39)

i.e.: syslog-Segment.log

![Image](https://github.com/user-attachments/assets/63097c6b-a4bc-4023-a5a0-3c995dcb4b65)

### To Reproduce Bug

#define WLED_DEBUG
#define WLED_DEBUG_HOST "192.168.1.3"
#define WLED_DEBUG_PORT 514

and have a syslog server which uses the hostname from the send log after the date

### Expected Behavior

instead of have just i.e. 

`Apr 16 20:46:20 Segment geometry: 0,30 -> 0,1`

to have

`Apr 16 20:46:20 HOSTNAME Segment geometry: 0,30 -> 0,1`

this should create than just one log file with the complete log

### Install Method

Self-Compiled

### What version of WLED?

0.16.0-alpha

### Which microcontroller/board are you seeing the problem on?

ESP32-S2

### Relevant log/trace output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

netmindz:
From you description it sounds like the debug messages do not contain a hostname, hence why if you select the option on your syslog server to extract the hostname from the message you get unexpected results.

To be honest I was unaware the debug feature was even trying to implement the syslog protocol, i thought it was raw logs, where did you see reference to WLED using syslog?

blazoncek:
@netmindz you are correct. Netdebug does not implement any protocol, it is intended to transmit raw debug messages.
However if it can be made syslog compatible it wouldn't hurt.
Though I do not see areal value beyond using that feature wit multiple units being debugged simultaneously.

But just a note of warning: netdebug will not provide you with crash dumps.

KrX3D:
will be solved with 

https://github.com/wled/WLED/pull/4664

PR #4650, opened 2025-04-19
Adds hostname (serverDescription) to log in net debug

KrX3D:
Fix for https://github.com/wled/WLED/issues/4647

Adds the hostname (serverDescription) to the send output. 

Output before:

Apr 16 20:46:20 Segment geometry: 0,30 -> 0,1

With this fix:

Apr 19 16:06:20 EleksTube Wifi state: 3

So now the syslog server creates only one file syslog-EleksTube.log instead of multiple (see picture in https://github.com/wled/WLED/issues/4647)

## Summary by CodeRabbit

- **Refactor**
  - Improved network debug output by buffering messages and sending them in batches, resulting in more organized and grouped UDP debug logs.
  - Enhanced hostname resolution efficiency for debug output.
  - Reduced network traffic by minimizing the number of UDP packets sent during debugging.

### DIFF
--- a/wled00/net_debug.cpp
+++ b/wled00/net_debug.cpp
@@ -2,9 +2,7 @@
 
 #ifdef WLED_DEBUG_HOST
 
-size_t NetworkDebugPrinter::write(uint8_t c) {
-  if (!WLED_CONNECTED || !netDebugEnabled) return 0;
-
+void NetworkDebugPrinter::resolveHostName() {
   if (!debugPrintHostIP && !debugPrintHostIP.fromString(netDebugPrintHost)) {
     #ifdef ESP8266
       WiFi.hostByName(netDebugPrintHost, debugPrintHostIP, 750);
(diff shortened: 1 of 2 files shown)


blazoncek:
Thank you. This seems like a good improvement on a first glance.

Would you care to take a look at implementing QuickDebug by @gmag11 as proposed in #3396 ?
That could be a huge benefit for debugging purposes.

KrX3D:
> Would you care to take a look at implementing QuickDebug by @gmag11 as proposed in #3396 ? That could be a huge benefit for debugging purposes.

yeah i can take a look, but cant promise when, since i got a couple other things to do at the moment.

netmindz:
The title and description don't really match you code changes. Adding a buffer is a significant change to behavior and resolving your hostname for every message will give poor performance

KrX3D:
Hi, so i will revert the net debug back to the original. I created a seperate syslog.h/.cpp which can used and is able to use the format RFC3164/5424 or even send it raw like in the net debug (using a bufer)

i will update this PR in the next days with the new files

KrX3D:
Hi, im closing this since like i said i created something seperate to the net debug here:
https://github.com/wled/WLED/pull/4664

for syslog there is also an option RAW which should send like the net debug the output

PR #4651, opened 2025-04-19
Removed PS memory manager and some minor improvements

DedeHai:
Removing the mem-manager should fix all PS crashes but comes at the expence of increased RAM usage.
This is in preparation to merging segment layering by @blazoncek

The local rendering buffer is now also added to segment data, particles are also stored in segment data, as it was prior to adding the memorymanager.

A more appropriate memory management can be added back in later.
Since the increased RAM usage is a problem for ESP8266, I removed the local rendering buffer for the 1D system on ESP8266, making rendering a lot slower but at least it works up to about 600 LEDs.

- removed some comments
- slight improvement in particle binning for collisions (might be a tiny bit faster)
- Improved particle brightness in PS Impact

## Summary by CodeRabbit

- **Refactor**
  - Simplified and unified particle system memory management, removing legacy memory manager and related logic.
  - Streamlined rendering and collision handling for improved performance and maintainability.
  - Adjusted particle count limits for ESP8266 devices.
  - Enhanced code clarity by updating variable types and removing redundant comments.
- **Bug Fixes**
  - Improved consistency in particle system initialization and parameter handling.
  - Fixed potential issues with particle lifetimes and standby times for certain effects.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -7998,7 +7998,7 @@ uint16_t mode_particlefire(void) {
   uint32_t i; // index variable
   uint32_t numFlames; // number of flames: depends on fire width. for a fire width of 16 pixels, about 25-30 flames give good results
 
-  if (SEGMENT.call == 0) { // initialization TODO: make this a PSinit function, this is needed in every particle FX but first, get this working.
+  if (SEGMENT.call == 0) { // initialization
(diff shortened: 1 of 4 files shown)


blazoncek (approved):
IMO these are good changes that will make merging segment blending/layering easier.

I am still not sold on motion blur being included within PS, though. 😉

DedeHai:
I do get your point about motion blurring being somewhat reduntant and can be easily done in the FX itself. I see it as an "ease of use" feature plus there is the speed advantage when doing it in the PS. Same goes for smear blur.

PR #4652, opened 2025-04-20
Only disable Arduino OTA when using -D WLED_DISABLE_OTA

blazoncek:
- since Update object is accessible even with ArduinoOTA disabled it should possible to use HTTP OTA regardless
- this (using WLED_DISABLE_OTA) saves about 12kB while still allowing HTTP OTA updates
- HTTP OTA updates can be blocked using PIN or OTA lock

## Summary by CodeRabbit

- **Refactor**
  - OTA update functionality is now always available, removing the previous option to disable it at build time. The fallback message for disabled OTA updates has been removed.
  - The "Enable ArduinoOTA" option in the settings UI is now hidden when OTA updates are disabled at compile time.

### DIFF
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -611,7 +611,9 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
   if (pwdCorrect) { //only accept these values from cfg.json if ota is unlocked (else from wsec.json)
     CJSON(otaLock, ota[F("lock")]);
     CJSON(wifiLock, ota[F("lock-wifi")]);
+    #ifndef WLED_DISABLE_OTA
     CJSON(aOtaEnabled, ota[F("aota")]);
+    #endif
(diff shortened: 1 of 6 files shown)


DedeHai:
do any of the WLED tools use arduino OTA? if not: should it be disabled by default assuming not many users actually use this method?

blazoncek:
IDK what users actually use but chances are majority only uses HTTP updates (HA and mobile app use that).
If we choose to disable ArduinoOTA by default then we should rename the directive to WLED_ENABLE_ARDUINO_OTA.

blazoncek:
Any other thoughts on this?

willmmiles:
I think we should rename the directive to WLED_ENABLE_ARDUINO_OTA either way.  No objections to making it disabled by default.

(Actually, if it was up to me, I'd move ArduinoOTA to a usermod - but that's a different patch.)

blazoncek:
Agreed, but let's do that in separate PR.

PR #4654, opened 2025-04-22
add support for up to 10 ESPNow remotes

DedeHai:
- upgraded ESPNow remotes to a vector array
- changed config to support arrays
- updated wifi settings page to:
  - display the list of linked remotes
  - add last seen remote to the list
  - limit to 10 remotes max
  - send out the list of remots as "RM0" to "RM9"
  
Since I have very little knowledge on HTML and java script I did the UI update with a lot of help from AI's. 
Please check the code for any mistakes or possible improvements. I tested it and it does work but I really do not know if there is better ways to do the java part.

image

## Summary by CodeRabbit

- **New Features**
  - Added support for managing multiple ESP-NOW remote MAC addresses in the WiFi settings, replacing the previous single-remote limitation.
  - Introduced a new user interface for adding, removing, and listing up to 10 ESP-NOW remotes directly from the settings page.

- **Bug Fixes**
  - Improved handling and validation of remote MAC addresses to prevent duplicates and enforce limits.
  - Enhanced message processing to accept ESP-NOW data only from authorized linked remotes.

- **Refactor**
  - Updated internal logic to store and process multiple remote devices, enhancing flexibility for advanced users.
  - Removed legacy single-remote filtering to support multiple remotes seamlessly.

### DIFF
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -39,7 +39,7 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
 #ifndef WLED_DISABLE_ESPNOW
   CJSON(enableESPNow, nw[F("espnow")]);
   linked_remotes.clear();
-  JsonArray lrem = nw[F("lrem")]; // array of MAC addresses
+  JsonArray lrem = nw[F("linked_remote")]; // array of MAC addresses
   if (!lrem.isNull()) {
     for (size_t i = 0; i < lrem.size(); i++) {
       std::array<char, 13> entry{};
@@ -739,7 +739,7 @@ void serializeConfig(JsonObject root) {
   JsonObject nw = root.createNestedObject("nw");
 #ifndef WLED_DISABLE_ESPNOW
   nw[F("espnow")] = enableESPNow;
-  JsonArray lrem = nw.createNestedArray(F("lrem"));
+  JsonArray lrem = nw.createNestedArray(F("linked_remote"));
   for (size_t i = 0; i < linked_remotes.size(); i++) {
     lrem.add(linked_remotes[i].data());
   }
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -39,20 +39,22 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
 #ifndef WLED_DISABLE_ESPNOW
   CJSON(enableESPNow, nw[F("espnow")]);
   linked_remotes.clear();
-  JsonArray lrem = nw[F("linked_remote")]; // array of MAC addresses
+  JsonVariant lrem = nw[F("linked_remote")];
   if (!lrem.isNull()) {
-    for (size_t i = 0; i < lrem.size(); i++) {
-      std::array<char, 13> entry{};
-      getStringFromJson(entry.data(), lrem[i], 13);
-      entry[12] = '\0';
-      linked_remotes.push_back(entry); // TODO: need to check for invalid MAC?
+     if (lrem.is<JsonArray>()) {
+      for (size_t i = 0; i < lrem.size(); i++) {
+        std::array<char, 13> entry{};
+        getStringFromJson(entry.data(), lrem[i], 13);
+        entry[12] = '\0';
+        linked_remotes.emplace_back(entry);
+      }
     }
-  }
-  else { // legacy support for single MAC address in config
-    std::array<char, 13> entry{};
-    getStringFromJson(entry.data(), nw[F("linked_remote")], 13);
-    entry[12] = '\0';
-    linked_remotes.push_back(entry);
+    else { // legacy support for single MAC address in config
+      std::array<char, 13> entry{};
+      getStringFromJson(entry.data(), lrem, 13);
+      entry[12] = '\0';
+      linked_remotes.emplace_back(entry);
+    }
   }
 #endif
 
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -38,8 +38,22 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
   JsonObject nw = doc["nw"];
 #ifndef WLED_DISABLE_ESPNOW
   CJSON(enableESPNow, nw[F("espnow")]);
-  getStringFromJson(linked_remote, nw[F("linked_remote")], 13);
-  linked_remote[12] = '\0';
+  linked_remotes.clear();
+  JsonArray lrem = nw[F("lrem")]; // array of MAC addresses
+  if (!lrem.isNull()) {
+    for (size_t i = 0; i < lrem.size(); i++) {
+      std::array<char, 13> entry{};
+      getStringFromJson(entry.data(), lrem[i], 13);
+      entry[12] = '\0';
+      linked_remotes.push_back(entry); // TODO: is this a good way to check for invalid MAC?
+    }
+  }
+  else { // legacy support for single MAC address in config
+    std::array<char, 13> entry{};
+    getStringFromJson(entry.data(), nw[F("linked_remote")], 13);
+    entry[12] = '\0';
+    linked_remotes.push_back(entry);
+  }
 #endif
 
   size_t n = 0;
@@ -725,7 +739,10 @@ void serializeConfig(JsonObject root) {
   JsonObject nw = root.createNestedObject("nw");
 #ifndef WLED_DISABLE_ESPNOW
   nw[F("espnow")] = enableESPNow;
-  nw[F("linked_remote")] = linked_remote;
+  JsonArray lrem = nw.createNestedArray(F("lrem"));
+  for (size_t i = 0; i < linked_remotes.size(); i++) {
+    lrem.add(linked_remotes[i].data());
+  }
 #endif
 
   JsonArray nw_ins = nw.createNestedArray("ins");
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -45,7 +45,7 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
       std::array<char, 13> entry{};
       getStringFromJson(entry.data(), lrem[i], 13);
       entry[12] = '\0';
-      linked_remotes.push_back(entry); // TODO: is this a good way to check for invalid MAC?
+      linked_remotes.push_back(entry); // TODO: need to check for invalid MAC?
     }
   }
   else { // legacy support for single MAC address in config
--- a/wled00/data/settings_wifi.htm
+++ b/wled00/data/settings_wifi.htm
@@ -157,29 +157,17 @@ Static subnet mask:<br>
 		function aR(id, mac) {
 			if (!/^[0-9A-F]{12}$/i.test(mac)) return; // check for valid hex string
 			if (rC >= 10) return;
-			let inputs = gId('rml').getElementsByTagName('input');
-			for (let i = 0; i < inputs.length; i++) {
-				if (inputs[i].value === mac) return;
+			let inputs = d.querySelectorAll("#rml input");
+			for (let i of (inputs || [])) {
+				if (i.value === mac) return;
 			}
-			let list = gId('rml');
-			let row = cE('div');
-			let inp = cE('input');
-			inp.type = 'text';
-			inp.name = id;
-			inp.value = mac;
-			inp.maxLength = 12;
-			inp.minLength = 12;
-			inp.onchange = uR;
-			row.appendChild(inp);
-			let btn = d.createElement('button');
-			btn.type = 'button';
-			//btn.className = 'btn btn-xs';
-			btn.innerText = '-';
-			btn.onclick = function() {
-				rR(row);
-			};
-			row.appendChild(btn);
-			list.appendChild(row);
+			let l = gId('rml'), r = cE('div'), i = cE('input');
+			i.type = 'text'; i.name = id; i.value = mac; i.maxLength = 12; i.minLength = 12; i.onchange = uR;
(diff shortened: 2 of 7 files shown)


wled00/cfg.cpp
```diff
@@ -38,8 +38,22 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
   JsonObject nw = doc["nw"];
 #ifndef WLED_DISABLE_ESPNOW
   CJSON(enableESPNow, nw[F("espnow")]);
-  getStringFromJson(linked_remote, nw[F("linked_remote")], 13);
-  linked_remote[12] = '\0';
+  linked_remotes.clear();
+  JsonArray lrem = nw[F("lrem")]; // array of MAC addresses
```
blazoncek, line 42: Why not keep original name?
You'd only need to check if it is string or an array. You can see example in `json.cpp`.

wled00/data/settings_wifi.htm
```diff
@@ -205,8 +261,12 @@ <h3>ESP-NOW Wireless</h3>
 			Enable ESP-NOW: <input type="checkbox" name="RE"><br>
 			<i>Listen for events over ESP-NOW<br>
 			Keep disabled if not using a remote or wireless sync, increases power consumption.<br></i>
-			Paired Remote MAC: <input type="text" name="RMAC" minlength="12" maxlength="12"><br>
-			Last device seen: <span class="rlid" onclick="d.Sf.RMAC.value=this.textContent;" style="cursor:pointer;">None</span> <br>
+      Last device seen: <span class="rlid">None</span>
+      <button type="button" class="btn btn-xs" onclick="addRemote('RM'+remoteCount,this.parentElement.querySelector('.rlid').textContent)">Add</button><br>
```
blazoncek, line 265: You could use `id` property in last device seen `<span>` and then use shorter form `gId("rlid")`.

wled00/set.cpp
```diff
@@ -91,8 +91,21 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
...
+      char rm[4];
+      snprintf(rm, sizeof(rm), "RM%d", n); // "RM0" to "RM9"
+      if (request->hasArg(rm)) {
+        const String& arg = request->arg(rm);
+        if (arg.isEmpty()) continue;
+        std::array<char, 13> mac{};
+        strlcpy(mac.data(), request->arg(rm).c_str(), 13);
+        strlwr(mac.data());
+        if (mac[0] != '\0') {
+          linked_remotes.push_back(mac);
```
blazoncek, line 105: Perhaps try `emplace_back()`.

blazoncek (changes requested):
I would like to not change existing config token name.

wled00/cfg.cpp
DedeHai: just out of my own ignorance... updated. can you point me to the case in json.cpp you are referring to? I am not very familiar with the json library.

wled00/set.cpp
DedeHai: I think emplace_back() has no benefits for a simple vector of just an array.

wled00/data/settings_wifi.htm
DedeHai: I tried this, since xml.cpp modifies it, it must be a `class` no? I cann add an `id` on top.

DedeHai:
in the html files, indentation seems to be messed up here on git, but it displays right in vscode. Do I need to change a setting or is this a github thing?

wled00/cfg.cpp
blazoncek: See: https://github.com/wled/WLED/blob/10b925acb7ec0a440491940bb3cde537da50c1d5/wled00/json.cpp#L145-L174
Or: https://github.com/wled/WLED/blob/10b925acb7ec0a440491940bb3cde537da50c1d5/wled00/json.cpp#L387-L417

When JSON data (file or stream) is read into memory key's type is dynamically determined. If you anticipate either array or object you can use `JsonVariant` and decide on the fly what kind of content it has. However, you can use `JsonObject` or `JsonArray` and check for null. If the content is an array, `JsonObject` will be null and vice-versa.

wled00/data/settings_wifi.htm
blazoncek: `class` is used for formatting purposes, `id` for identification. Each element *must* have unique `id` if it is used.
You can assign multiple classes to an element (and they are inherited) but `id` is specific to the element (tag).

wled00/set.cpp
blazoncek: Emplace creates the object (even POD) in-place, while `push_back` uses copy-constructor. For POD that means one `memcpy` more instructions. Not a big deal, but still.

blazoncek:
> indentation seems to be messed up

It is not. HTML/CSS/JS files use tabs, C/C++ files use spaces. Fix that in your VSC configuration and you are good.

blazoncek:
One more thing that sprung into mind: Instead of "Add" and "Remove" buttons you could use familiar "+" and "-" buttons.

wled00/data/settings_wifi.htm
DedeHai: if I remove the class and replace it with id, the xml.cpp update does not work. 
this does work:
```
Last device seen: <span class="rlid" id="ld">None</span>
<button type="button" onclick="aR('RM'+rC,gId('ld').textContent)">+</button><br>
```

DedeHai:
I think I have it all fixed and ready now.

wled00/data/settings_wifi.htm
```diff
@@ -136,12 +136,75 @@
...
+		}
+		// add remote MAC to the list
+		function aR(id, mac) {
+			if (!/^[0-9A-F]{12}$/i.test(mac)) return; // check for valid hex string
+			if (rC >= 10) return;
+			let inputs = gId('rml').getElementsByTagName('input');
+			for (let i = 0; i < inputs.length; i++) {
+				if (inputs[i].value === mac) return;
+			}
+			let list = gId('rml');
```
blazoncek, line 164: Reorder and reuse `list` object. With shorter name even more bytes could be saved.

wled00/data/settings_wifi.htm
```diff
@@ -136,12 +136,75 @@
 			getLoc();
 			loadJS(getURL('/settings/s.js?p=1'), false);	// If we set async false, file is loaded and executed, then next statement is processed
 			if (loc) d.Sf.action = getURL('/settings/wifi');
+			setTimeout(tE, 500); // wait for DOM to load before calling tE()
```
blazoncek, line 139: This should not be needed here. You can add call to `tE()` in xml.cpp if it is necessary to be invoked when all fields are set.

wled00/wled.h
```diff
@@ -538,7 +538,8 @@ WLED_GLOBAL bool     serialCanTX _INIT(false);
 WLED_GLOBAL bool enableESPNow        _INIT(false);  // global on/off for ESP-NOW
 WLED_GLOBAL byte statusESPNow        _INIT(ESP_NOW_STATE_UNINIT); // state of ESP-NOW stack (0 uninitialised, 1 initialised, 2 error)
 WLED_GLOBAL bool useESPNowSync       _INIT(false);  // use ESP-NOW wireless technology for sync
-WLED_GLOBAL char linked_remote[13]   _INIT("");     // MAC of ESP-NOW remote (Wiz Mote)
+//WLED_GLOBAL char linked_remote[13]   _INIT("");     // MAC of ESP-NOW remote (Wiz Mote)
+WLED_GLOBAL std::vector<std::array<char, 13>> linked_remotes; // MAC of ESP-NOW remotes (Wiz Mote)
```
blazoncek, line 542: @willmmiles can you explain why is it better to use `std::array<char,13>` instead of plain `char[13]` or what are the consquences?

IMO we don't need object of char[13] as we always know it will have exactly 12 characters and we are never going to manipulate it (except for setting content) so it is just as good to have POD.

blazoncek (approved):
A minor tweak here and there might do it good but it looks great already. Will need to test still.

wled00/data/settings_wifi.htm
DedeHai: I tried many variations, including calling tE() from xml.cpp. this variant, although a bit of a hack, is the only one that consistently worked. tE() was always evaluated too early, not properly showing/hiding the section.

wled00/wled.h
blazoncek: Found this: https://stackoverflow.com/questions/2392308/c-vector-of-char-array

wled00/data/settings_wifi.htm
```diff
@@ -136,12 +136,75 @@
...
+		function rstR() {
+			gId('rml').innerHTML = ''; // clear remote list
+			rC = 0;
+			uR(); // update remote list
+		}
+		// add remote MAC to the list
+		function aR(id, mac) {
+			if (!/^[0-9A-F]{12}$/i.test(mac)) return; // check for valid hex string
+			if (rC >= 10) return;
+			let inputs = gId('rml').getElementsByTagName('input');
```
blazoncek, line 160: What you also could do is: `let inputs = d.querySelectorAll("#rml input")`.

wled00/data/settings_wifi.htm
```diff
@@ -136,12 +136,75 @@
...
+			for (let i = 0; i < inputs.length; i++) {
```
blazoncek, line 161: And here: `for (let input of inputs||[]) ...`

wled00/data/settings_wifi.htm
```diff
@@ -136,12 +136,75 @@
...
+			inp.value = mac;
+			inp.maxLength = 12;
+			inp.minLength = 12;
+			inp.onchange = uR;
+			row.appendChild(inp);
+			let btn = d.createElement('button');
+			btn.type = 'button';
+			//btn.className = 'btn btn-xs';
+			btn.innerText = '-';
+			btn.onclick = function() {
```
blazoncek, line 178: shorter: `btn.onclick = (e)=>{..};`
And it would be better (more universal) to replace `row` from next line with actual object from which event triggered.

wled00/data/settings_wifi.htm
```diff
@@ -136,12 +136,75 @@
...
+		}
+		// remove a remote from the list
+		function rR(e) {
+			e.remove();
+			uR();
+		}
+		// update remote list
+		function uR() {
+			let macs = [];
+			let inputs = gId('rml').getElementsByTagName('input');
```
blazoncek, line 194: same here:
```javascript
d.querySelectorAll('#rml input').forEach((o,i)=>{
  if (o.value && o.value !== '') macs.push(o.value);
});
```

blazoncek:
If you wish, here is a slightly better JS (validation, markup):
```javascript
		// toggle visibility of ESP-NOW remote list based on checkbox state
		function tE() {
			// keep the hidden input with MAC addresses, only toggle visibility of the list UI
			gId('rlc').style.display = d.Sf.RE.checked ? 'block' : 'none';
		}
		// reset remotes: initialize empty list (called from xml.cpp)
		function rstR() {
			gId('rml').innerHTML = ''; // clear remote list
			uR(); // update remote list
		}
		function isMAC(v) { return /^[0-9A-F]{12}$/i.test(v); };
		// add remote MAC to the list
		function aR(mac) {
			let rC = gId("rml").childElementCount;
			if (rC >= 10) return;
			let id = 'RM' + rC;
			mac = mac.replace(/:/g, '').replace(/-/g, '').toUpperCase();
			if (!isMAC(mac)) mac = 'FFFFFFFFFFFF'; // check for valid hex string
			let inputs = d.querySelectorAll('#rml input[type="text"]');
			for (let input of inputs||[]) {
				if (input.value === '') return;
				if (input.value === mac) mac = '';
			}
			let r = cE('div');
			let i = cE('input');
			i.type = 'text';
			i.name = id;
			i.value = mac;
			i.style.width = '13ch'
			i.maxLength = 12;
			i.minLength = 12;
			i.onchange = uR;
			i.oninput = (e) => {
				e.target.style.color = !isMAC(e.target.value) ? 'red' : 'white';
			};
			r.appendChild(i);
			let btn = d.createElement('button');
			btn.type = 'button';
			btn.className = 'sml';
			btn.innerText = '-';
			btn.onclick = (e) => {
				r.remove();
				uR();
			};
			btn.style.marginBottom = '4px';
			r.appendChild(btn);
			gId('rml').appendChild(r);
			uR();
		}
		// update remote list
		function uR() {
			let j = 0;
			d.querySelectorAll('#rml input[type="text"]').forEach((i) => {
				i.style.color = !isMAC(i.value) ? 'red' : 'white';
				i.value = i.value.toUpperCase();
				i.name = 'RM' + j++;
			});
			gId('+').style.display = gId("rml").childElementCount < 10 ? 'inline' : 'none';
			return true; // Allow form submission to continue
		}
```
BTW filling "RMA" is unnecessary as it is no used anywhere.

DedeHai:
I updated the java code, it should now handle everything properly. I took a different approach though: the list is filled by xml.cpp, UI can add one or remove one at a time. The edge case where the list is full is handled like this (tested and working):
- + button is removed if list is full, stays removed even if the user deletes one
- saving the page lets the user then add a remote again
while this is a less elegant solution than the above proposed, it uses less code to handle that edge case. Extending the list in the future to more than 10 remotes by using A...Z as on top of 0-9 for the RMx IDs would make it even more of an edge case.

From my side, this is ready to merge.

DedeHai:
@netmindz would be good to have another opinion on these changes, especially the additions to `settings_wifi.htm`. Code is tested and working.

netmindz (approved):
I have not tested, but the PR looks good

PR #4657, opened 2025-04-23
bugfix in enumerating buttons and busses

DedeHai:
char value was changed from "55" to 'A' which is 65. need to deduct 10 so the result is 'A' if index counter is 10.

@netmindz this bugfix also needs to go into 0.15.x if approved by @blazoncek 

## Summary by CodeRabbit

- **Bug Fixes**
  - Adjusted the calculation of parameter key and form element suffix characters for bus indices 10 and above, ensuring correct labeling and display for advanced LED settings. This affects how certain settings are represented in both the user interface and JavaScript output.


blazoncek (approved):
Approved.
@DedeHai once merged just chery-pick into 0_15_x branch.

PR #4660, opened 2025-04-24
Added ability to test if Usermod is enabled from UsermodManager

rhkean:
* added ability to check if a usermod is enabled from UsermodManager
* added isEnabled() to Usermod base class
* standardized mod_id parameter in UsermodManager (some were uint8_t, some were uint16_t)

## Summary by CodeRabbit

- **New Features**
  - Added the ability to check if a user module is enabled.

- **Improvements**
  - Expanded support for a wider range of user module IDs.

### DIFF
--- a/wled00/fcn_declare.h
+++ b/wled00/fcn_declare.h
@@ -430,6 +430,7 @@ class Usermod {
     virtual void onUpdateBegin(bool) {}                                      // fired prior to and after unsuccessful firmware update
     virtual void onStateChange(uint8_t mode) {}                              // fired upon WLED state change
     virtual uint16_t getId() {return USERMOD_ID_UNSPECIFIED;}
+    virtual bool isEnabled() {return true;}                                  // if not overridden and object exists, assume enabled (many mods are enabled just by being there)
 
   // API shims
   private:
@@ -450,7 +451,7 @@ namespace UsermodManager {
   void loop();
   void handleOverlayDraw();
   bool handleButton(uint8_t b);
-  bool getUMData(um_data_t **um_data, uint8_t mod_id = USERMOD_ID_RESERVED); // USERMOD_ID_RESERVED will poll all usermods
+  bool getUMData(um_data_t **um_data, uint16_t mod_id = USERMOD_ID_RESERVED); // USERMOD_ID_RESERVED will poll all usermods
   void setup();
   void connected();
   void appendConfigData(Print&);
@@ -470,6 +471,7 @@ namespace UsermodManager {
   void onStateChange(uint8_t);
   Usermod* lookup(uint16_t mod_id);
   size_t getModCount();
+  bool isModEnabled(uint16_t mod_id = USERMOD_ID_RESERVED);
 };
 
 // Register usermods by building a static list via a linker section
--- a/wled00/um_manager.cpp
+++ b/wled00/um_manager.cpp
@@ -31,10 +31,17 @@ bool UsermodManager::handleButton(uint8_t b) {
   }
   return overrideIO;
 }
-bool UsermodManager::getUMData(um_data_t **data, uint8_t mod_id) {
+bool UsermodManager::getUMData(um_data_t **data, uint16_t mod_id) {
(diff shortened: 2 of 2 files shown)


blazoncek:
What is the benefit in your opinion (for current state and the future)?

For enabled feature I would expect modification of all existing usermods that employ enabled functionality. It would also make sense to add private variable to base class if this is something that's beneficial.

rhkean:
> I don't think enabling/disabling WiFi should be a usermod. In fact I've added a short snippet into JSON handling to start/stop AP mode on JSON API request. See json.cpp, `deserializeState()`. It would be better to enhance that, but be prepared that I will comment heavily (with good intentions even if it might nod sound so). And IMO BLE has a toll that we won't be willing to pay, but as a usermod it might have a chance. So far no-one brought it to living.
> 
interestingly, I came to that same conclusion today.  LOL
I am actually almost finished re-writing my disable/Enable WiFi as a core feature.  I've finished the code modifications.  Now I'm working on the WiFi settings page updates.

The BLE mod that I'm writing will use BLE advertising, so a paired connection should not be needed.... I'll keep it as a separate PR, though.  The trick is in managing the radio which is why I had to back up and implement the wifi-disable first.

> > I should include the modifications to the other usermods in this PR
> 
> Yes, if you are adding a feature, make sure it is used where appropriate. Some usermods indeed have `enabled()` public method and ability to detect if usermod is available is already included using `lookup()` function. There are a few usermods that already share data via public methods. `getUMData()` was intended to share data with core WLED features but I do not think we'll pursue that further as other approaches were made (i.e. ability to add usermod effects).

understood.  I'll make those changes and update the PR

willmmiles:
I'm not sure that we need this to be a global feature.  Every currently merged usermod that interacts with another usermod does so by either (a) making custom calls directly on the other usermod's object, or (b) uses getUMData().  If you're implementing a specialized communication between two modules, I'd recommend the former approach.  If you're instead trying to ask "does this usermod exist in this build", I would recommend a compile-time test; see https://github.com/wled/WLED/blob/354a0aef5205c2c56f4784a554d85eeb84972ec0/usermods/PWM_fan/setup_deps.py for an example of how to do that.

That said: if we do want to adopt an "enabled" state as a queryable property on all usermods, I think we should adopt MoonModules' approach to improve compatibility between the forks.  https://github.com/MoonModules/WLED-MM/blob/7cb8eebba61e0e14f15cbf036f68f6030e9f5ca0/wled00/fcn_declare.h#L301C2-L302C32   (I was pretty sure I'd seen a PR for this, but I can't find it now -- maybe it was just a suggestion on the Discord?)

rhkean:
The idea was to test if an installed mod (that could be enabled or disabled) was enabled in a simpler way than using the getUMData() method.  That felt like a lot of code to write to a) make a simple bool available to other usermods.  It makes sense to me if you're needing access to a large set of data like with the AR mod, but for a simple bool test.

If you all don't see it as a value-add, I can see your point of view.  We can close the PR, if you want.

willmmiles:
I agree `getUMdata` isn't a good solution for "simple" requests.  The common solution in current use is `static_cast<UsermodMyOtherClass*>(UsermodManager::lookup(USERMOD_ID_MYOTHERUSERMOD))->getDataIWant()`, which works because most usermods know exactly which other modules they want to interact with.  I believe the use case you described above falls in to this class -- your second module knows exactly which other module you want to interact with, so you're free to use `lookup()` to make calls directly on the module's own class type.

What I'm uncertain about is the use case for "the set of modules that expose an enable", especially since in this fork there's no API to do pretty much anything with them in a generic way on them except get their ID code.  (The MM fork also requires each usermod provide a name, which permits making a generic UI to manipulate the enable setting.)  Can you help me understand what is the use case for this set that isn't served by `lookup()`?

rhkean:
>  Can you help me understand what is the use case for this set that isn't served by `lookup()`?

Nope.... LOL  
you make great points.  I cannot see any advantage to this mod after this conversation... no speed or code size advantages, either.

Thank you for useful feedback and great conversation.  If you agree, I'll close this PR and focus on disabling WiFi... LOL

PR #4670, opened 2025-05-01
added Enable/Disable WiFi functionality

rhkean:
* Added ability to disable WiFi
* enabling or disabling of WiFi takes effect after restart
* AP can be temporarily started with a Button0 long press

## Summary by CodeRabbit

- **New Features**
  - Added a global WiFi enable/disable flag.
  - Clearer, separate handling for toggling Access Point (AP) and Station (STA) modes.

- **Bug Fixes**
  - Centralized and streamlined AP shutdown for more reliable AP disable.
  - Skips connection handling when WiFi is disabled and prevents AP startup in button-only AP mode.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -27,6 +27,8 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }} 
           sinceTag: v0.15.0
+          # Exclude issues that were closed without resolution from changelog
+          exclude-labels: 'stale,wontfix,duplicate,invalid'
       - name: Update Nightly Release
         uses: andelf/nightly-release@main
         env:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,7 +24,9 @@ jobs:
       with:
           token: ${{ secrets.GITHUB_TOKEN }} 
           sinceTag: v0.15.0
-          maxIssues: 500       
+          maxIssues: 500
+          # Exclude issues that were closed without resolution from changelog
+          exclude-labels: 'stale,wontfix,duplicate,invalid'       
     - name: Create draft release
       uses: softprops/action-gh-release@v1
       with:


wled00/button.cpp
```diff
@@ -322,15 +322,26 @@ void handleButton()
...
+            //   if(!mod->isWiFiEnabled())
+            //   {
+            //     mod->EnableWiFi();
+            //   }
+            // #endif
+            WLED::instance().initAP(true);
+          }
         } else {
-          WLED::instance().initAP(true);
+          doReboot = true;
```
DedeHai, line 344: very bad idea

wled00/fcn_declare.h
```diff
@@ -450,7 +450,7 @@ namespace UsermodManager {
   void loop();
   void handleOverlayDraw();
   bool handleButton(uint8_t b);
-  bool getUMData(um_data_t **um_data, uint8_t mod_id = USERMOD_ID_RESERVED); // USERMOD_ID_RESERVED will poll all usermods
+  bool getUMData(um_data_t **um_data, uint16_t mod_id = USERMOD_ID_RESERVED); // USERMOD_ID_RESERVED will poll all usermods
```
DedeHai, line 453: this is an unrelated change. while a good fix, it should not be in this PR

DedeHai (changes requested):
I briefly looked at the code, generally looks ok. did not test if it works as intended.

wled00/button.cpp
rhkean: I'm not sure why that is showing in this PR... That code was already there

wled00/fcn_declare.h
rhkean: true.  I think it's small enough to not be concerned about pulling it out, but if it needs to be pulled, I will

rhkean:
One thing that I forgot to note...

I do not have an Ethernet board to test the USER_ETHERNET features.  I think that I covered that area correctly, however, if someone with a board could test that Ethernet still works correct, I would greatly appreciate it.

wled00/button.cpp
DedeHai: It was not, you changed it

wled00/fcn_declare.h
DedeHai: Keeping changes separated helps future debugging

wled00/button.cpp
rhkean: yes, I misread the code snippet above... I realized what was being referenced when I opened the entire file.  It's fixed now.

wled00/fcn_declare.h
rhkean: I understand.  I'll pull it.

rhkean:
I'd also be interested in others' thoughts regarding the "turn WiFi off" comments from code rabbit.

wled00/json.cpp
```diff
@@ -477,9 +477,9 @@ bool deserializeState(JsonObject root, byte callMode, byte presetId)
   
   JsonObject wifi = root[F("wifi")];
   if (!wifi.isNull()) {
-    bool apMode = getBoolVal(wifi[F("ap")], apActive);
-    if (!apActive && apMode) WLED::instance().initAP();  // start AP mode immediately
-    else if (apActive && !apMode) { // stop AP mode immediately
+     bool apMode = getBoolVal(wifi[F("ap")], apActive);
```
DedeHai, line 480: Indentation

wled00/network.cpp
```diff
@@ -319,7 +319,7 @@ int findWiFi(bool doScan) {
 
 
 bool isWiFiConfigured() {
-  return multiWiFi.size() > 1 || (strlen(multiWiFi[0].clientSSID) >= 1 && strcmp_P(multiWiFi[0].clientSSID, PSTR(DEFAULT_CLIENT_SSID)) != 0);
+  return wifiEnabled && (multiWiFi.size() > 1 || (strlen(multiWiFi[0].clientSSID) >= 1 && strcmp_P(multiWiFi[0].clientSSID, PSTR(DEFAULT_CLIENT_SSID)) != 0));
```
DedeHai, line 322: Not sure this is correct, disabled can still be configured,  check where this is used

wled00/wled.cpp
```diff
@@ -649,42 +654,41 @@ void WLED::initConnection()
 
   if (WLED_WIFI_CONFIGURED) {
     showWelcomePage = false;
-    
-    DEBUG_PRINTF_P(PSTR("Connecting to %s...\n"), multiWiFi[selectedWiFi].clientSSID);
 
+    DEBUG_PRINTF_P(PSTR("Connecting to %s...\n"), multiWiFi[selectedWiFi].clientSSID);
```
DedeHai, line 658: Unnecessary change

wled00/wled.cpp
```diff
@@ -649,42 +654,41 @@ void WLED::initConnection()
...
-    if (apActive) {
-      DEBUG_PRINTLN(F("ESP-NOW initing in AP mode."));
-      #ifdef ESP32
-      quickEspNow.setWiFiBandwidth(WIFI_IF_AP, WIFI_BW_HT20); // Only needed for ESP32 in case you need coexistence with ESP8266 in the same network
-      #endif //ESP32
-      espNowOK = quickEspNow.begin(apChannel, WIFI_IF_AP);  // Same channel must be used for both AP and ESP-NOW
-    } else {
-      DEBUG_PRINTLN(F("ESP-NOW initing in STA mode."));
-      espNowOK = quickEspNow.begin(); // Use no parameters to start ESP-NOW on same channel as WiFi, in STA mode
+  #ifndef WLED_DISABLE_ESPNOW
```
DedeHai, line 674: Revert these changes

wled00/wled.cpp
```diff
@@ -816,8 +820,8 @@ void WLED::handleConnection()
       initConnection();
     }
     if (!apActive && now - lastReconnectAttempt > 12000 && (!wasConnected || apBehavior == AP_BEHAVIOR_NO_CONN)) {
-      if (!(apBehavior == AP_BEHAVIOR_TEMPORARY && now > WLED_AP_TIMEOUT)) {
-        DEBUG_PRINTF_P(PSTR("Not connected AP (@ %lus).\n"), nowS);
+      if (!(apBehavior == AP_BEHAVIOR_TEMPORARY && now > WLED_AP_TIMEOUT) && (apBehavior != AP_BEHAVIOR_BUTTON_ONLY)) {
```
DedeHai, line 823: Is this a bugfix?

wled00/wled.cpp
```diff
@@ -834,7 +838,7 @@ void WLED::handleConnection()
     DEBUG_PRINTLN();
     DEBUG_PRINT(F("Connected! IP address: "));
     DEBUG_PRINTLN(Network.localIP());
-    if (improvActive) {
+    if (wifiEnabled && improvActive) {
```
DedeHai, line 841: This should never be reached if wifi is  disabled no?

wled00/xml.cpp
```diff
@@ -198,6 +198,11 @@ void getSettingsJS(byte subPage, Print& settingsScript)
     printSetFormIndex(settingsScript,PSTR("AB"),apBehavior);
     printSetFormValue(settingsScript,PSTR("AS"),apSSID);
     printSetFormCheckbox(settingsScript,PSTR("AH"),apHide);
+    printSetFormCheckbox(settingsScript,PSTR("EN"),wifiEnabled);
+    settingsScript.printf_P(PSTR("gId('scan').disabled=%s;"), wifiEnabled ? PSTR("false") : PSTR("true"));
```
DedeHai, line 202: Call tw() instead?

blazoncek (changes requested):
I'm afraid I'll have to say no to this PR.

First, if you disable WiFi, how do you enable it again? Even AP mode on button press is disabled so there will be no way to enable it again. Or did I miss such behaviour?

Second, even though ESP-NOW uses WiFi circuitry it is not a WiFi protocol and should not be disabled. There is an option for enabling/disabling ESP-NOW in the settings.

Thirdly, there are unnecessary changes.

wled00/xml.cpp
rhkean: yes!  I was going to do that and then forgot.  thank you

rhkean:
> I'm afraid I'll have to say no to this PR.
> 
> First, if you disable WiFi, how do you enable it again? Even AP mode on button press is disabled so there will be no way to enable it again. Or did I miss such behaviour?
> 
the AP still starts with a button0 longpress (see 1st line in initAP().

> Second, even though ESP-NOW uses WiFi circuitry it is not a WiFi protocol and should not be disabled. There is an option for enabling/disabling ESP-NOW in the settings.
> 
I did not realize that ESP-NOW was not a WiFi protocol.  I'll remove those code changes.  That was my mistake.

> Thirdly, there are unnecessary changes.
Let me make the changes requested by you and @DedeHai and I'll pester you some more with another review.  If there are additional unnecessary changes that are not noted so far, please let me know and I'll clean those up too.

DedeHai:
ESPNow and wifi hardware are tightly entangled, it is not possible to disable wifi and have ESPNow enabled, at least in arduino framework afaik. 
Also the ability to disable wifi should be a usermod or at least a setting that first needs to be activated , hidden by default, with a big warning. Users tend to go nuts with random settings and I see much potential for bricked setups

wled00/network.cpp
rhkean: you're right.  I was cheating... I've cleaned it up and move that check to where it is supposed to be

wled00/wled.cpp
rhkean: yes, I'll remove it from this PR

wled00/wled.cpp
rhkean: it can reach this in an ETH configuration

rhkean:
I've made the code changes noted above.  But, I still have to test the rebuild.  I'll do that tonight and refresh the PR.

DedeHai:
My point is that it should not be possible to accidentally disable wifi and brick a setup

DedeHai:
I am not against making it a core feature if done right, was just thinking out loud. What I do want to avoid is many complaints about bricked setups

rhkean:
I've made the changes noted above and improved the WiFi settings page by moving the disable setting to the bottom of the page and included a warning with required confirmation checkbox.

DedeHai:
Did not test yet but from reading the code the double-check to disable it seems acceptable.
Does ESPNow work like this? also, wifi-sleep is no longer respected. Not sure wifi-sleep should be generally disabled when enabling ESPNow as it does not work well with wifi-sleep enabled.

rhkean:
> Did not test yet but from reading the code the double-check to disable it seems acceptable. Does ESPNow work like this? also, wifi-sleep is no longer respected. Not sure wifi-sleep should be generally disabled when enabling ESPNow as it does not work well with wifi-sleep enabled.

Can you point me to where the wifi-sleep is not respected?  I can't find any of my modifications that impacted that.
thanks

DedeHai:
> Can you point me to where the wifi-sleep is not respected? I can't find any of my modifications that impacted that. thanks

`WiFi.setSleep()` is not executed.

rhkean:
> > Can you point me to where the wifi-sleep is not respected? I can't find any of my modifications that impacted that. thanks
> 
> `WiFi.setSleep()` is not executed.

AH!  good catch!  thank you.  fixed.

blazoncek (changes requested):
I'm still not satisfied with proposed changes and I still do not like the fact that WiFi enabled state is saved into configuration (I will never approve that, no matter what but other team members may vote otherwise and I will accept that).

WLED has W in its name which stands for "wireless" (or WiFi).

There are also some irrelevant changes that need to be reverted.

I've worked with WiFi logic for several months while preparing ESP-NOW sync and have to say that it is still not ready for prime time due to the fact that I want WiFi to remain available at all times.

Let's put it this way:
* if Ethernet is connected it will be used and WiFi will not be used (effectively disabled, no need for any change)
* if you do not configure WiFi and set AP mode to never start, WiFi will not be enabled (no need for any change)
* if you want to *temporarily* disable WiFi, add JSON handling (that's on my TODO list)
* if your use case is to have device run on battery and reduce power consumption create a boot preset (or a playlist) that will *temporarily* disable WiFi

DedeHai:
fully agree with @blazoncek, if there were a JSON command to disable wifi instead, that would be a much more elegant solution.

wled00/set.cpp
```diff
@@ -20,6 +20,9 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
   if (subPage == SUBPAGE_WIFI)
   {
     unsigned cnt = 0;
+    wifiEnabled = !request->hasArg(F("EN"));
+    DEBUG_PRINTF("wifi: %s", wifiEnabled ? "enabled" : "disabled");
```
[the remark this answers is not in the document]
rhkean: I was following other examples in the code.  I'll modify this.  Thank you

rhkean:
wow!   Great feedback.  Thank you so much.

My use case is BLE.  My intentions were to implement disabling WiFi, then implement BLE, then eventually upgrade the BLE module with COEXISTENCE so the WiFi did not have to be disabled.

I'll make the changes you noted above, remove the config save and settings page changes, and add a JSON API call

package.json
```diff
@@ -25,10 +25,10 @@
   "dependencies": {
     "clean-css": "^5.3.3",
     "html-minifier-terser": "^7.2.0",
-    "web-resource-inliner": "^7.0.0",
-    "nodemon": "^3.1.9"
+    "nodemon": "^3.1.9",
```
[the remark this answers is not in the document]
rhkean: agreed.  I got tired of arguing with VSCode.... it kept marking it changed, so I finally just committed it... I'll remove it from the PR.

rhkean:
@blazoncek, 

OK.... made all of the changes mentioned above.  I've tested by putting a `wifiEnabled = false;` at the beginning of the setup() in wled.cpp.  the code successfully disables WiFi and still allows a button0 long press to start the AP even with a WiFi configuration.

I'll add the JSON API call next.

blazoncek:
For JSON API I'm thinking something in the lines of:
```c++
    if (!wifi[F("on")].isNull()) {
      bool sta = getBoolVal(wifi[F("on")], wifiEnabled);
      if (Network.isConnected() && !sta) WiFi.disconnect();
      if (sta) forceReconnect = true;
      wifiEnabled = sta;
    }
```

And simple logic in `handleConnection()` like:
```c++
  if (wifiConfigured && !wifiEnabled) return;
```
This will allow firing up AP on boot if WiFi is not configured (assuming AP was not disabled).

The logic behind the above is that you `WiFi.disconnect()` from configured SSID via JSON API. This (`disconnect(false)`) will leave WiFi circuitry enabled and operational (for ESP-NOW for example) while no longer trying to connect to SSID. If you'd want to completely shut down WiFi, you can call `WiFi.disconnect(true)` instead.
When enabling WiFi back on using JSON (via IR or button or some other means), forcing a reconnect will do all that's necessary.

rhkean:
now that's an elegant solution... I'll code it up and start testing after work!

(I've got to figure out how to use SerialJSON for now, bc I don't have an IR remote lying about)  :/
(at least until I get further along with the BLE stuff)

blazoncek:
Create two presets: `{"wifi":{"on":false}}` and `{"wifi":{"on":true}}` or a single `{"wifi":{"on":"t"}}` and assign them/it to buttons.

rhkean:
Added JSON API
added wifi["on"] and wifi["pwrOff"]

- button0 longpress will start the AP
- {"wifi":{"on":true}}  will connect to configured wifi
- {"wifi":{"on":true}}  will disconnect wifi leaving circuitry on
- {"wifi":{"on":true},"pwrOff":false} will disconnect wifi leaving circuitry on
- {"wifi":{"on":true},"pwrOff":true} will disconnect wifi and turn off the circuitry

rhkean:
> Added JSON API added wifi["on"] and wifi["pwrOff"]
> 
> * button0 longpress will start the AP
> * {"wifi":{"on":true}}  will connect to configured wifi
> * {"wifi":{"on":false}}  will disconnect wifi leaving circuitry on
> * {"wifi":{"on":false},"pwrOff":false} will disconnect wifi leaving circuitry on
> * {"wifi":{"on":false},"pwrOff":true} will disconnect wifi and turn off the circuitry

I'm trying to track down the cause, but I've found, what I think is, an issue.
if I create a preset `{"wifi":{"on":false,"pwrOff":true}}` and set it to "Apply on boot", it works perfectly if `wifiConfigured = false`

if, however, `wifiConfigured = true`, then it connects to the WiFi, then gets disabled, but does not release the IP... watching the debug output, it continues to report the leased IP instead of 0.0.0.0.  It also, seems to be taking more than 1 execution of the toggle preset to get it to reconnect to the network.  (the multiple toggling to get it to re-enable doesn't happen if I don't select pwrOff, but the IP issue does)

(side note, it appears that saving a preset with "apply at boot" causes that preset to execute immediately.  is that a bug?)

rhkean:
> I'm trying to track down the cause, but I've found, what I think is, an issue. if I create a preset `{"wifi":{"on":false,"pwrOff":true}}` and set it to "Apply on boot", it works perfectly if `wifiConfigured = false`
> 
> if, however, `wifiConfigured = true`, then it connects to the WiFi, then gets disabled, but does not release the IP... watching the debug output, it continues to report the leased IP instead of 0.0.0.0. It also, seems to be taking more than 1 execution of the toggle preset to get it to reconnect to the network. (the multiple toggling to get it to re-enable doesn't happen if I don't select pwrOff, but the IP issue does)
> 
figured it out...
WiFi init is started before the bootPresets are applied, but before the WiFi connects...

blazoncek:
Does it need to release IP? It will get a new one on reconnect if needed.
If that interferes with your processing use `WiFi.status()` and check if it is either WL_IDLE_STATUS or WL_DISCONNECTED (or !WL_CONNECTED).

As for boot preset: It is executed in async fashion (as are all presets).

rhkean:
> Does it need to release IP? It will get a new one on reconnect if needed. If that interferes with your processing use `WiFi.status()` and check if it is either WL_IDLE_STATUS or WL_DISCONNECTED (or !WL_CONNECTED).

I, personally, don't care as long as the radio is off.  It feels like bad practice, though... I'm not sure.  I'll spit out some more debug statements and do some more testing after work.  I'll pull in my BLE code and verify it doesn't crash.   LOL

> As for boot preset: It is executed in async fashion (as are all presets).

yeah, I figured.  as is the WiFi connect call, I'm seeing.   Hence the timing of it all.  I'll try to get it to behave with a little more consistency.

blazoncek:
Async behaviour is intended. It should be implemented everywhere.

rhkean:
I verified that the connection status is reporting as connect in the situation discussed earlier...  I've updated the code to handle it.

netmindz:
It maybe a stupid question, but why do you want to disable WIFI?
I think each new feature should have a clear use case, but this PR does not give any indication of why you would want this feature

rhkean:
additional testing showed that AP was not being enabled/disabled properly.... fixed now

rhkean:
> It maybe a stupid question, but why do you want to disable WIFI? I think each new feature should have a clear use case, but this PR does not give any indication of why you would want this feature

it's actually mentioned in many of the comments of this PR.
* power saving (especially helpful for battery power)
* my other mod (in progress) is BLE.  I'm hoping to get COEXISTENCE to work, but that will probably end up being a 2.0 version of the mod, bc it's not cooperating, so I'm focusing on single use antennae config initially

rhkean:
any decisions on this @blazoncek ? @DedeHai ?

DedeHai:
can you explain the current state?
why are there two parameters, enable and power?
from what I see in code they are redundant or at least unclear. if the goal is to disable wifi, one parameter is enough and disable should be fully disable, including AP and ESPnow, anything else is just a bit convoluted. Or can you elaborate the intention and use cases?

blazoncek:
> any decisions on this @blazoncek ?

I know this may work for you but I am not comfortable with the implementation as it is based on a workaround.

I am not sure why would you want to circumvent initial WiFi connect as it has no side effects and will run at most for about 10 or so seconds.

What I particularly don't like is disconnecting WiFi in event handling. IMO it is not the correct place to do it.

Unfortunately I'm lacking in time ATM for testing but have a slightly different approach in my dev branch. Once I test it I'll let you know if it is any better.

rhkean:
> can you explain the current state? why are there two parameters, enable and power? from what I see in code they are redundant or at least unclear. if the goal is to disable wifi, one parameter is enough and disable should be fully disable, including AP and ESPnow, anything else is just a bit convoluted. Or can you elaborate the intention and use cases?

great question
as noted in one of @blazoncek 's messages earlier, the call to `WiFi.disconnect(false)` will disconnect WiFI, however `WiFi.disconnect(true)` will power off the WiFi circuitry for better power saving.  So, I added the extra option for both as a "bonus" feature.  

for me, personally, I don't want to turn off the circuitry, bc I'm trying to use it for BLE, so I saw the benefit for both options.

rhkean:
> > any decisions on this @blazoncek ?
> 
> I know this may work for you but I am not comfortable with the implementation as it is based on a workaround.
> 
> I am not sure why would you want to circumvent initial WiFi connect as it has no side effects and will run at most for about 10 or so seconds.

My testing showed that the actual timing was not actually turning off the WiFi if it is configured when using a preset to run on boot.  I placed a gazillion log statements in so that I could trace it out and it played out in this order:
1. boot
2. init wifi and check if WiFi is configured
3. start scan
4. request boot presets
5. preset runs
6. scan completes and connects

if there is a better way to have the preset run on boot, but delay until after WiFi connects, that would be perfect I think

I thought about adding a `static disableWiFi(bool pwrOff)` method to the WLED class that could be called from a usermod if needed (such as the BLE mod I'm working on)
 
> What I particularly don't like is disconnecting WiFi in event handling. IMO it is not the correct place to do it.

agreed.  I wasn't too happy about that, either.  however, my BLE startup kept crashing.  :)

> Unfortunately I'm lacking in time ATM for testing but have a slightly different approach in my dev branch. Once I test it I'll let you know if it is any better.

I haven't had a chance to look at your fork yet.  I try to get to that this weekend.  If you have reworked the WiFi startup though, it'll probably work a lot smoother.  Please keep me in the loop.

I really would love for my contributions to be of value and not just a "well it works for me so I'll just run my own builds" scenario.  So, I greatly appreciate all the feedback.

blazoncek:
@rhkean what I meant with "why would you want to circumvent initial WiFi connect" is: let it connect and turn off WiFi after about 10 or so seconds using a playlist with two preset - one dummy that does nothing but runs for 10 (or so) seconds and another one that turns off WiFi.

rhkean:
> @rhkean what I meant with "why would you want to circumvent initial WiFi connect" is: let it connect and turn off WiFi after about 10 or so seconds using a playlist with two preset - one dummy that does nothing but runs for 10 (or so) seconds and another one that turns off WiFi.

well, obviously, I didn't think of that.   ROFL

here's the simplified implementation.  Only a `wifiEnabled` bool flag to test, and a JSON API call.

I noticed that the AP shutdown code was implemented 6 different times, so I consolidated it into a single WLED method.

xxv:
Hey folks. I just wanted to chime in here as a user who would love to turn off wifi (temporarily). My use case is battery power applications, where I have WLED running on a power bank. I don't need wifi to be on the entire time, but having it on for a little while when I plug it in is good. I found the AP opens: "Temporary (no connection after boot)" setting and thought that would be what I wanted, but it turns out to only disable broadcasting, not actually turning off the wifi hardware. I still leave it on because I don't need people connecting to my WLED controller and I can always unplug it if I need to change settings. Perhaps that could be another way that this change could be used, specifically helpful for battery-powered applications. If the setting actually saved power, that would be amazing!

It may also be wise to mention that it's only on for 5 minutes in the setting description somewhere, as it's not very clear what "no connection after boot" actually means in practice.

rhkean:
> Hey folks. I just wanted to chime in here as a user who would love to turn off wifi (temporarily). My use case is battery power applications, where I have WLED running on a power bank. I don't need wifi to be on the entire time, but having it on for a little while when I plug it in is good. I found the "AP opens: Temporary (no connection after boot)" setting and thought that would be what I wanted, but it turns out to only disable broadcasting, not actually turning off the wifi hardware. I still leave it on because I don't need people connecting to my WLED controller and I can always unplug it if I need to change settings. Perhaps that could be another way that this change could be used, specifically helpful for battery-powered applications. If the setting actually saved power, that would be amazing!
> 
> It may also be wise to mention that it's only on for 5 minutes in the setting description somewhere, as it's not very clear what "no connection after boot" actually means in practice.

if you have a Button0 (Button connected to GPIO-0), try the Never setting.  In code, it's actually called "Button only".  It will not turn on the AP unless you hold the button for 5-6 seconds.

blazoncek:
> I noticed that the AP shutdown code was implemented 6 different times, so I consolidated it into a single WLED method.

You should've checked my fork as suggested.

> It will not turn on the AP unless you hold the button for 5-6 seconds.

Beware, holding it too long and it will be factory reset.

rhkean:
> > I noticed that the AP shutdown code was implemented 6 different times, so I consolidated it into a single WLED method.
> 
> You should've checked my fork as suggested.

so... removed the shutdownAP method?  close the PR?

blazoncek:
Nothing, I'm just pointing out it clashes with what I've done some 9 months ago, yet I'm still not comfortable with my solution too.

rhkean:
> Nothing, I'm just pointing out it clashes with what I've done some 9 months ago, yet I'm still not comfortable with my solution too.

ah... that makes sense.  thanks

(side note: I've managed to get AP and BLE working together.  STA and BLE still crashes, though... probably an advertising interval issue or something)

xxv:
> if you have a Button0 (Button connected to GPIO-0), try the Never setting.  In code, it's actually called "Button only".  It will not turn on the AP unless you hold the button for 5-6 seconds.

That's good to know and I'll try it out for devices that have a button. Some of mine don't, so I'll need to find another solution like fixing the temporary AP shutdown. I still think it is a good idea that should happen either in this or a follow-up PR once the shutdown code is consolidated.

DedeHai:
finally got around to test this. In general, it works as intended but I think it still needs improvement:
- power consumption seems inconsistent - sometimes I measure 32mA, sometimes over 40mA after disabling wifi. may also be a glitch but hints at wifi not always entering the same state depening on what was going on at the time of "disable" processing. No LEDs connected for this test.
- wifi is not shut down properly if AP is active: setting AP active to "always", connecting to AP and disabling wifi will only disable the AP but not wifi, can still go back to my network and access it. AP is still visible but service is probably shut down so cant connect.

the way I tested this was with a preset: `{"wifi":{"on":false}}` on an ESP32 C3.

rhkean:
> finally got around to test this. In general, it works as intended but I think it still needs improvement:
> 
> * power consumption seems inconsistent - sometimes I measure 32mA, sometimes over 40mA after disabling wifi. may also be a glitch but hints at wifi not always entering the same state depening on what was going on at the time of "disable" processing. No LEDs connected for this test.
> * wifi is not shut down properly if AP is active: setting AP active to "always", connecting to AP and disabling wifi will only disable the AP but not wifi, can still go back to my network and access it. AP is still visible but service is probably shut down so cant connect.
> 
> the way I tested this was with a preset: `{"wifi":{"on":false}}` on an ESP32 C3.

Thank you @DedeHai 
I think that I'm just going to leave this here until @blazoncek gets time to pull his networking changes into main and then address these changes accordingly.
thoughts?

blazoncek:
Due to recent finds regarding Ethernet's PLL and WiFi circuitry I have to revisit entire ESP-NOW approach. Unfortunately it will take some time before that is done.
However, precisely due to that issue it will be necessary to implement WiFi on/off functionality and connection handling if WiFi is off. As I've already heavily modified connection logic for ESP-NOW merging this PR into main will only add to the complexity of future merge of ESP-NOW sync if that is ever done.

PR #4686, opened 2025-05-12
Add SK9822 to the supported LED list

rshamsnejad:
As a first time user of SK9822 LEDs, I was confused when I couldn't find the option in the drop-down list of supported LEDs. After a bit of googling I found that people advised to use the APA102 preset because they work in the same way. This pull request simply appends "/SK9822" to the APA102 string to prevent further confusion among new users.

## Summary by CodeRabbit

- **Style**
  - Updated the LED type description to display "APA102/SK9822" instead of "APA102" for improved clarity in the user interface.


blazoncek:
Thank you.
Unfortunately I'm not very fond of increasing flash use for something that can be explained in [documentation](https://kno.wled.ge/basics/compatible-led-strips/) as we are struggling with WLED binary size.
Yes, it is just 7 bytes in this case but there are other LED strips that have protocol similar to APA102 (as does SK9822) and accommodating all would be unacceptable.
But let's see what other members say.

DedeHai:
I agree, this should be in KB.
there are probably half a dozen compatible ones to APA102, can't have them all in the list.

PR #4690, opened 2025-05-13
AR: add compile-time flag for "Automatic Gain Control" option

Arcitec:
Automatic Gain Control is a very important aspect of the audioreactive plugin, and is vitally important when the external music volume constantly changes.

It makes sense to allow users to choose their preferred AGC behavior at compile-time, since they can already set the Gain and Squelch via flags.

Adds `SR_AGC` as a flag, which defaults to 0 (off).

---

Notes:

- This completes the compile-time control of the most common AudioReactive settings: Gain `SR_GAIN`, Squelch `SR_SQUELCH` and now AGC `SR_AGC`.
- I've read through the code several times to make sure everything is still initialized correctly, and of course tested it on hardware.
- I also have a version of this patch for the `0_15_x` branch, where the repo structure is different. I'll await your feedback first, but I can open a pull request for that branch too if necessary, to save you time. :+1:


Arcitec:
Oh, I pushed one small tweak: Since the AGC feature only exists for the `ARDUINO_ARCH_ESP32`, I pushed a small documentation adjustment to clarify that AGC is an ESP32-only feature.

PS: I disagree with CodeRabbit about putting the note at the end of the line, since their example looks like "Setting 3 = Lazy (Only ESP32)" which would confuse users. And since the readme doesn't use any punctuation within bullet points, it gets very awkward to add anything to the end of the sentence. I am open to suggestions if has a better sentence structure, but it's such a minor detail that I don't think it's worth arguing over. :smile:

netmindz (approved):

Arcitec:
@netmindz Thank you for taking a look at this. I just noticed the merge. :)

netmindz:
Thanks for your submission @Arcitec

Issue #4488, opened 2025-01-13
Effects "Bouncing Balls" and "Fire 2012" not working when disabling 2D support on C3


Samybigio:
### What happened?

when i compile the main branch (0.16 dev) source code, and disable 2d support, the effects "Bouncing Balls" and "Fire 2012" don't work. when applying bouncing balls, the segment fades to black, while applying fire2012 freezes the segment to a frame of the previous effect, when applying other effects the segment behaves fine. note that i tested 0.15 source code and 0,16 dev source code, and the issue was present only in the dev source code.

### To Reproduce Bug

compile the main branch (0.16 dev) source code, and disable 2d support with flag "-D WLED_DISABLE_2D", compile and flash the SW to a ESP32-C3, and the two mentioned effects will be broken.

### Expected Behavior

normal behavior of the 2 effects

### Install Method

Self-Compiled

### What version of WLED?

WLED 0.16-alpha

### Which microcontroller/board are you seeing the problem on?

ESP32-C3

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #4692, opened 2025-05-19
fix for https://github.com/wled/WLED/issues/4488

DedeHai:
virtual strip index is added even if strip is 1D. this change fixes FX using virtual strips not working when WLED_DISABLE_2D is used.

I recommend waiting to merge this until segment layering PR is merged.


DedeHai:
this fix does NOT need porting to 0.15 as the bug this PR fixes was introduced in 0.16

blazoncek:
```c++
#ifndef WLED_DISABLE_2D
  #define indexToVStrip(index, stripNr) ((index) | (int((stripNr)+1)<<16))
#else
  #define indexToVStrip(index, stripNr) (index) // since stripNr will never be > 0
#endif
```
Would have solved the issue as well.

PR #4693, opened 2025-05-19
Update to distortionwave FX

DedeHai:
- fixed "jumpyness" by improving offset calculation resolution
- added palette support
- added two rendering modes for use with palette: one uses hue-mapping, one uses brightness mapping
- added alternate animation mode (from initial source)
- extended scaling: zoom checkmark (depends on matrix size)

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -7412,7 +7412,7 @@ static const char _data_FX_MODE_2DAKEMI[] PROGMEM = "Akemi@Color speed,Dance;Hea
 
 // Distortion waves - ldirko
 // https://editor.soulmatelights.com/gallery/1089-distorsion-waves
-// adapted for WLED by @blazoncek
+// adapted for WLED by @blazoncek, improvements by @dedehai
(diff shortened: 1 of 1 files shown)


DedeHai:
any objections to merging this? 
existing presets should not be affected by the changes.

willmmiles:
LGTM from a quick code inspection.  Go for it.

On the micro-optimization side:  I notice that you've converted some `2*` to `x<<1`.  The compiler ought to be able to make that transformation itself if it's faster, given a literal multiplicand -- sometimes it isn't, particularly when it can leverage fused-multiply-add instructions.  This is especially likely for `*2` and `*4` cases where address calculation opcodes can sometimes be co-opted into doing arithmetic for us.  I'm not recommending changing it back, just something to be cognizant of.

DedeHai:
thanks for the review. good point about the multiply-add, I will consider it in the future. That variable that was a constant 2 bothered me as it was confusing, I chose a shift over a multiplication just out of habit, since its not really impacting speed a bit of a questionable choice ;)

Issue #3576, opened 2023-12-08
mqtt (wled/device/status) status info takes a long time to update


viktor6:
### What happened?

mqtt status takes a long time to update
esp8266   wled/wled/status   30 seconds
esp32      wled/wled/status   2 minutes  which is very long

### To Reproduce Bug

on/off esp32 wemos D1 

### Expected Behavior

30 c

### Install Method

Binary from WLED.me

### What version of WLED?

Installed version: WLED 0.14.0 (ESP32-D0WDQ6 build 2310130)

### Which microcontroller/board are you seeing the problem on?

ESP32

### Anything else?

On esp8266 it makes a status update from offline to online in 30 seconds
On esp32  it makes a status update from offline to online in 2 minutes

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

viktor6:
but on my esp32 it updates consistently in 2 minutes. I tried 3 different esp32s and more than once

viktor6:
I need to track when wled appears on the network and for further automation

dosipod:
Okay i so what you mean  when you listed on/off  you actually mean remove power or make it offline from the network  
![image](https://github.com/Aircoookie/WLED/assets/87589371/ff9e895a-770b-4d1f-985d-90453cd41b35)
I can see use case for car garage

viktor6:
And on esp8266 it is about 30 seconds

blazoncek:
You don't seem to understand MQTT. `/status` topic is **retained** with LWT set.
When WLED device is without power you will receive LWT (which is "offline"), when it is powered on you will receive "online". WLED will not re-broadcast it on its own but only together with any other state change (see example below).

![Screenshot 2023-12-08 at 20 33 16](https://github.com/Aircoookie/WLED/assets/7722392/d14fc3c4-4a9e-4a89-be7f-b6dbf33bd72d)

dosipod:
> And on esp8266 it is about 30 seconds

I also got the same strange result you pointed out when I tested with esp8266 , it will be reconnected after i remove power  in very much less time  then esp32 . Must be something with mqtt re-connection on esp32  could be heavy after power down  , I did not even trap for offline /online but only for wled/WLED/g as you see in the example which is the same test I did on esp32  


![image](https://github.com/Aircoookie/WLED/assets/87589371/eb3cf615-10aa-48e3-aebf-773abdd0f849)

dosipod:
@viktor6  
[...]

PR #4698, opened 2025-05-22
Fix running initMqtt twice on bootup

zuckschwerdt:
This fixes a bug where at reboot `initMqtt()` will run twice in quick succession and fail.
E.g. /status messages will be duplicated and bounce and MQTT will be unavailable for the first 30 seconds. This change fixes this.

On Startup `WLED::initInterfaces()` will run `initMqtt()` and start to connect.
But then `WLED::loop()`, with `lastMqttReconnectAttempt==0` will also again run `initMqtt()`.
Since the first connect has not had time to finish `initMqtt()` can not guard against this and will again start to connect.
The `mqttClientID` will collide and both connections will be ended. Then 30 seconds later the loop again runs a `initMqtt()` and only that connection will be permanent.

This issue can be seen with published messages at startup like this
```
@0 secs : wled/mywled/status : offline
@0 secs : wled/mywled/status : offline
@1 secs : wled/mywled/status : online
@1 secs : wled/mywled/status : online
@1 secs : wled/mywled/status : offline
@30 secs : wled/mywled/status : online
```

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -738,9 +738,6 @@ void WLED::initInterfaces()
   e131.begin(e131Multicast, e131Port, e131Universe, E131_MAX_UNIVERSE_COUNT);
   ddp.begin(false, DDP_DEFAULT_PORT);
   reconnectHue();
-#ifndef WLED_DISABLE_MQTT
-  initMqtt();
-#endif
   interfacesInited = true;
   wasConnected = true;
 }


blazoncek:
That may be the cause for what someone reported a while ago.

zuckschwerdt:
> That may be the cause for what someone reported a while ago.

Yes, #3576 fits the error I see and will be fixed by this.

netmindz:
Thanks for your PR

PR #4710, opened 2025-06-06
fix particle brightness distribution with new gamma correction

DedeHai:
Particle System depends on linear brightness distribution, gamma corrected values are non-linear.
- added invers gamma table
- reversing gamma after brightness distribution makes it linear once again

there is no other way to fix this unfortunately. It comes at the cost of a second lookup table (additional 256bytes of RAM used) and a slight hit to FPS in PS effects but its nothing to worry about.

## Summary by CodeRabbit

- **New Features**
  - Improved LED particle rendering by introducing inverse gamma correction for more accurate brightness and color representation.

- **Bug Fixes**
  - Enhanced consistency of particle brightness across different modes by applying gamma correction and its inverse during rendering.

- **Chores**
  - Updated internal handling of gamma correction settings to support both gamma and inverse gamma lookup tables for improved visual output.

### DIFF
--- a/wled00/FXparticleSystem.cpp
+++ b/wled00/FXparticleSystem.cpp
@@ -584,7 +584,7 @@ void ParticleSystem2D::render() {
       continue;
     // generate RGB values for particle
     if (fireIntesity) { // fire mode
-      brightness = (uint32_t)particles[i].ttl * (3 + (fireIntesity >> 5)) + 20;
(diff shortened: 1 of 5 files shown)


blazoncek:
You could have used calcGammaTable() to fill both gamma and inverse gamma tables instead of adding another function that needs to be called at the same time. for loop is the same for both.

DedeHai:
Good point.

PR #4711, opened 2025-06-07
Add pre-applied gamma to c3g & (some) fastled palettes to match original look

DedeHai:
With the new and proper implementation of gamma correction for colors, the palettes imported from http://seaviewsensing.com/pub/cpt-city without any correction by @blazoncek have a slightly different, washed-out look because all three color channels are corrected with the same gamma.
Originally, the palettes were pre-corrected with gammas of 2.6 for red, 2.2 for green and 2.5 for blue. 
In order to match the originals after applying the new default gamma of 2.2 the red and blue channels need a slight pre-correction of 2.6/2.2=1.182 for red and 2.5/2.2=1.136 for blue.

The palettes used from FastLed require an inverse-correction of 2.2 to get a close match to the original looks. I also tried with 1.8 and 1.5 to get something "in between" which also looks quite ok but slightly different. Without any inverse-correction, they do not look good, especially party-colors and rainbow colors. 
FastLed palettes that use HTML colors (clouds, lava, ocean, forest) I left untouched as I think they do look slightly better than the originals, up for discussion.

Also: changed the default value for gamma correction to 2.2.

## Summary by CodeRabbit

- **New Features**
  - Added three new color palettes with improved gamma correction for more accurate color representation.

- **Improvements**
  - Updated many existing color palettes for better color consistency and accuracy.
  - Adjusted the default gamma correction value for improved color processing.

- **Documentation**
  - Expanded explanations regarding gamma correction and palette handling for better user understanding.

### DIFF
--- a/wled00/palettes.h
+++ b/wled00/palettes.h
@@ -2,106 +2,109 @@
 #define PalettesWLED_h
 
 /*
- * Color palettes for FastLED effects (65-73).
+ * WLED Color palettes
+ *
+ * Note: palettes imported from http://seaviewsensing.com/pub/cpt-city are gamma corrected using gammas (1.182, 1.0, 1.136)
+ *       this is done to match colors of the palettes after applying the (default) global gamma of 2.2 to versions
+ *       prior to WLED 0.16 which used pre-applied gammas of (2.6,2.2,2.5) for these palettes.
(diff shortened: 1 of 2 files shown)


blazoncek:
I'm fine with that though I'm also fine with palettes prior to update.

PR #4712, opened 2025-06-07
Fix Segment::fade_out for 2d expansion

willmmiles:
Ran in to this one during testing last night.   Since we're using get/setPixelColorRaw, we need the true pixel count, not the virtual length.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved pixel fading behavior to ensure fade-out covers the entire 2D area for segments, providing more consistent visual effects.
  - Enhanced fade and blur effects by accurately processing the full physical pixel area in segments, resulting in smoother transitions.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -989,7 +989,8 @@ void Segment::fade_out(uint8_t rate) const {
   if (!isActive()) return; // not active
   rate = (256-rate) >> 1;
   const int mappedRate = 256 / (rate + 1);
-  for (unsigned j = 0; j < vLength(); j++) {
+  const size_t length = is2D() ? (vWidth() * vHeight()) : vLength();
+  for (unsigned j = 0; j < length; j++) {
     uint32_t color = getPixelColorRaw(j);
     if (color == colors[1]) continue; // already at target color
     for (int i = 0; i < 32; i += 8) {
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1038,12 +1038,12 @@ void Segment::blur(uint8_t blur_amount, bool smear) const {
 #endif
   uint8_t keep = smear ? 255 : 255 - blur_amount;
   uint8_t seep = blur_amount >> 1;
-  unsigned rlength = rawLength();
+  unsigned vlength = vLength();
   uint32_t carryover = BLACK;
   uint32_t lastnew; // not necessary to initialize lastnew and last, as both will be initialized by the first loop iteration
   uint32_t last;
   uint32_t curnew = BLACK;
-  for (unsigned i = 0; i < rlength; i++) {
+  for (unsigned i = 0; i < vlength; i++) {
     uint32_t cur = getPixelColorRaw(i);
     uint32_t part = color_fade(cur, seep);
     curnew = color_fade(cur, keep);
@@ -1057,7 +1057,7 @@ void Segment::blur(uint8_t blur_amount, bool smear) const {
     last = cur; // save original value for comparison on next iteration
     carryover = part;
   }
-  setPixelColorRaw(rlength - 1, curnew);
+  setPixelColorRaw(vlength - 1, curnew);
 }
 
 /*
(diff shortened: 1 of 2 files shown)


wled00/FX_fcn.cpp
```diff
@@ -989,7 +989,8 @@ void Segment::fade_out(uint8_t rate) const {
   if (!isActive()) return; // not active
   rate = (256-rate) >> 1;
   const int mappedRate = 256 / (rate + 1);
-  for (unsigned j = 0; j < vLength(); j++) {
+  const size_t length = is2D() ? (vWidth() * vHeight()) : vLength();
+  for (unsigned j = 0; j < length; j++) {
```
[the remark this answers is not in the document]
willmmiles: Although the backing pixel buffer is always allocated to the full physical segment size, effect rendering uses only a subset of it when mirror or grouping are enabled (`vWidth() * vHeight()` for 2D segments, `vLength()` for 1D segments).  Color processing beyond that subset is not required.  (The full pixel buffer is always allocated so the output configuration on a Segment object may be changed later without needing to reallocate the buffer.)

DedeHai (approved):

wled00/FX_fcn.cpp
```diff
@@ -989,7 +989,8 @@ void Segment::fade_out(uint8_t rate) const {
   if (!isActive()) return; // not active
   rate = (256-rate) >> 1;
   const int mappedRate = 256 / (rate + 1);
-  for (unsigned j = 0; j < vLength(); j++) {
+  const size_t length = is2D() ? (vWidth() * vHeight()) : vLength();
```
blazoncek: This shouldn't be necessary as `vLength()`/`virtualLength()`takes into account 2D (and its mapping).

willmmiles: That would be correct if `fade_out` was using `getPixelColor()/setPixelColor()`, which apply the mapping.  Right now it's using `getPixelColorRaw()/setPixelColorRaw()`, so it needs the "unmapped" length.

The tradeoff is between calculating the color fade extra times, vs applying the mapping.  Given that the author of the function chose the `Raw` forms I took this to mean that they intended the former, which I agree is probably faster in most cases.

The bug is pretty clearly apparent with any of the Grav* AR FX with bar expansion -- it doesn't fade out properly.

blazoncek: Right. The author would be me.
That only means `fade_out()` should use regular `get/setPixelColor()` and skip speed optimisations. This might hold true for all *universal* functions.
Or re-implement mapping to correctly iterate expanded pixels.

willmmiles: OK - I think the optimization is helpful here, but if you feel strongly that we should go the other way, that's OK by me.  (Now that you mention it, there are a couple of other locations with the same issue - I should've checked more thoroughly when I spotted the one.)

Long term I think we should move any and all mapping to the blending stage, where we are guaranteed we're only calculating it once per frame.  (Yes, I'm aware there are a couple of FX that do special things in certain mapping modes - I'm sure we can find solutions for them.)

blazoncek: IMO the next best thing is to do `for (int i=0; i<length(); i++) ...` as that will cover all pixels.
I think that will still be faster than mapping.
This is feasible as actual rendering is done in blending function.

willmmiles: I thought about `length()` too; I went with `vWidth()*vHeight()` as it's smaller for any case with grouping or mirroring.

blazoncek: If everything would be properly implemented those 2 functions would be unavailable in WLED_DISABLE_2D.
But I was lazy.

willmmiles: Sometimes lazy is good. :)

I'll update with a fix for all cases.

wled00/FX_fcn.cpp
```diff
@@ -1036,12 +1038,12 @@ void Segment::blur(uint8_t blur_amount, bool smear) const {
 #endif
   uint8_t keep = smear ? 255 : 255 - blur_amount;
   uint8_t seep = blur_amount >> 1;
-  unsigned vlength = vLength();
+  unsigned rlength = rawLength();
```
blazoncek: `blur()` does not need modification as 2D is handled by `blur2D()`

willmmiles: Fair enough.  I'd included it for completeness in the set of things using `get/setPixelColorRaw()`, but you're correct to note that it's depending on the proverbial "sufficiently smart compiler" to discard the extra tests.  I've reverted this change.

PR #4713, opened 2025-06-07
Update crude parallel I2S detection

willmmiles:
My ESP32s are reporting a maximum of 32 buses, so the parallel I2s tickbox was missing from the LED settings UI.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved hardware detection logic for ESP32 boards in the LED settings interface, ensuring more accurate display and configuration options for supported devices.


blazoncek (approved):
👍

PR #4722, opened 2025-06-10
add inverse gamma 32 function and fix colors in pride

DedeHai:
more FX that do not use palettes may need gamma inversion as well so adding a function to do that.

## Summary by CodeRabbit

- **New Features**
  - Improved color blending in the Pride 2015 color waves effect for enhanced visual accuracy.

- **Bug Fixes**
  - Adjusted gamma correction handling for more accurate color representation in certain effects.

- **Chores**
  - Updated the default gamma correction value for new configurations to improve color fidelity.


blazoncek (approved):
I would have chosen a slightly different method/function names (i.e. `inverseGama32()` or `revertGamma32()`) but otherwise looks ok to me.

PR #4727, opened 2025-06-12
Added on/off event to trigger auto save event

encrust-ace:
This change will add on/off event of LED to trigger auto save event, which will be applied after power restore.

## Summary by CodeRabbit

- **New Features**
  - Auto-save now also triggers when the power state of the LED strip changes, ensuring settings are saved automatically after turning the strip on or off.


blazoncek:
FYI checking `bri` and `offMode` is identical.
And yes I know the logic is convoluted.

encrust-ace:
Great, I am closing the PR, but we can try to add user selectable events in setting

Issue #3738, opened 2024-02-03
Reverse twinklecat effect. Slowly fade twinkles on, then instantly off.


benjamw:
**Is your feature request related to a problem? Please describe.**
No. It's a new effect request.

**Describe the solution you'd like**
The twinklecat effect is very nice, and I noticed in the code that it is a simple adjustment on the twinklefox effect.
The effect that I would like to see is the opposite of the twinklecat (twinkledog maybe?) where the twinkling lights fade from the background to the selected color or palette segment, and then once max brightness is reached, instantly go back to background color.

I imagine this would look really cool with a bright background and a dark foreground colors.

**Describe alternatives you've considered**
No alternatives exist.

This is probably a really easy fix, and I saw the location in the code where it would be possible to just add it myself, but my C++ is very sad, and I'm not exactly sure what all the variables are doing. Particularly, the `ph` var that looks like it drives the whole thing.

PR #4728, opened 2025-06-12
Add TwinkleFadeIn effect

benjamw:
This adds a new effect which is the reverse of Twinklecat: **Twinklefadein** (See Issue #3738)

Twinklecat instantly turns on random lights and then slowly fades them out. This effect slowly fades in random lights, and then instantly turns them off.

It is a very simple and small change, but has not been tested.

The only questions I had were about the FX numbering scheme:
1. Should new effects be placed at the bottom of the list regardless of where they naturally sort?
2. Should new effects be placed in the list where they naturally sort, but keep the numbering the same?
3. Should new effects be placed in the list where they naturally sort, and renumber the list to keep the numbering sequential?

I opted for option #1, but can easily adjust it if needed.

Thanks

## Summary by CodeRabbit

- **New Features**
  - Introduced a "Reverse" option for the Twinklecat effect, allowing users to choose between LEDs fading on or fading off.
- **Improvements**
  - Updated the Twinklecat effect settings to include a new checkbox for reversing the twinkle behavior.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -4820,7 +4820,7 @@ static const char _data_FX_MODE_WAVESINS[] PROGMEM = "Wavesins@!,Brightness vari
 //////////////////////////////
 //     Flow Stripe          //
 //////////////////////////////
-// By: ldirko  https://editor.soulmatelights.com/gallery/392-flow-led-stripe , modifed by: Andrew Tuline
+// By: ldirko  https://editor.soulmatelights.com/gallery/392-flow-led-stripe , modifed by: Andrew Tuline, fixed by @DedeHai
 uint16_t mode_FlowStripe(void) {
   if (SEGLEN <= 1) return mode_static();
   const int hl = SEGLEN * 10 / 13;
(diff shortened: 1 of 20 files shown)


blazoncek:
If it is a small and simple change, why not update existing effect with one of the custom checkboxes?

willmmiles:
To expand: FX ids are at somewhat of a premium these days.  I also think it'd be better to add an "invert" checkbox to TwinkleCat vs giving it a new name.

wled00/FX.h
```diff
@@ -373,7 +373,9 @@ extern byte realtimeMode;           // used in getMappedPixelIndex()
 #define FX_MODE_PS1DSONICBOOM          215
 #define FX_MODE_PS1DSPRINGY            216
 #define FX_MODE_PARTICLEGALAXY         217
-#define MODE_COUNT                     218
+#define FX_MODE_TWINKLEFADEIN          218
+
+#define MODE_COUNT                     219
```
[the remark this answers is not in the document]
netmindz: Some FXs have been removed, but their ID not reused, so MODE_COUNT needs to be one greater than the highest mode, not always equal to the count of FX_MODE_*

benjamw:
Thank you for the feedback. That makes a lot more sense to use a checkbox instead of a whole new effect. (Which also implies that "TwinkleCat" itself could have been a checkbox. But that's a whole other issue.)

I've converted the function param to be a checkbox called "Reverse" and committed that.

Note: I did use AI (cursor) to assist me, as I am unfamiliar with the PROGMEM mode string syntax, so please let me know if that is incorrect.

wled00/FX.h
```diff
@@ -373,6 +373,7 @@ extern byte realtimeMode;           // used in getMappedPixelIndex()
 #define FX_MODE_PS1DSONICBOOM          215
 #define FX_MODE_PS1DSPRINGY            216
 #define FX_MODE_PARTICLEGALAXY         217
+
```
DedeHai: revert

wled00/FX.cpp
```diff
@@ -2665,7 +2667,7 @@ uint16_t mode_twinklecat()
 {
   return twinklefox_base(true);
 }
-static const char _data_FX_MODE_TWINKLECAT[] PROGMEM = "Twinklecat@!,Twinkle rate,,,,Cool;!,!;!";
+static const char _data_FX_MODE_TWINKLECAT[] PROGMEM = "Twinklecat@!,Twinkle rate,,,,Cool;!,!;!,Reverse";
```
DedeHai: check2 is after cooling checkbox:
`"Twinklecat@!,Twinkle rate,,,,Cool,Reverse;!,!;!";`

please make sure to test your code so it does what you want it to.

DedeHai (changes requested):

wled00/FX.cpp
benjamw: I plan on testing it. Thanks for the updates.

benjamw:
I just compiled and tested, and it's working as expected.

For future reference, is there a way to edit the icons used for the checkboxes?

DedeHai:
> I just compiled and tested, and it's working as expected.
> 
> For future reference, is there a way to edit the icons used for the checkboxes?

the icons are fixed but you can choose the checkmark number

netmindz:
Not had chance to test, but the code changes are looking much more compact :)

PR #4729, opened 2025-06-13
Digital clock overlay

Myriadbits:
Added an usermod for my clock project (called Cloxel). This usermod basically puts a digital clock on top of the effects. The effects are faded to make the clock more readable. It works best with a 32x8 LED matrix. It includes multiple fonts and some color/text customization settings.
An additional feature is that by using the API it is possible to show temporary (short) message on the WLED screen (useful to let kids know that it is time for dinner).  
See https://www.instructables.com/Cloxel-WLED-Pixel-Clock/ for some pictures.


blazoncek:
FYI you can use segment layers in 0.16 with multiple blending modes that will allow various effects with clock overlay.

netmindz:
Hi @Myriadbits  - looks like a nice addition. It looks like coderabbitai has picked up on a few potential issues with your code, sometimes it doesn't get things right, so please comment on any false positives

Myriadbits:
> FYI you can use segment layers in 0.16 with multiple blending modes that will allow various effects with clock overlay.

@blazoncek : It sounds exactly like something I could use in this mod. Where can I find more information on this feature? What calls should I look at?

DedeHai:
@Myriadbits if you write it as a FX instead of a direct overlay, the new layering feature takes care of the rest if you create two or more overlapping segments you can choose how they are mixed.

Making it into  an effect also gives direct control over colors using the colorwheel, brightness using the segment slider making it easier to use and more versatile if you want to use the FX sliders and checkmarks.

Question:
at first glance, the added fonts are very similar to the current WLED fonts. anything better about the ones you added?

blazoncek:
@everyone there is an analog clock FX waiting in one of the PRs. With 0.16 it can be used as an overlay.

@Myriadbits you can't, unless you examine the code. It is not intended to be used in usermods and is intended to replace any and all overlays available (in due time).

Scrolling text FX set up with appropriate segment name is a flexible way to have time and/or date displayed as overlay over any effect.

blazoncek:
@DedeHai as we managed to reduce binary with Tasmota framework it may be feasible to add more fonts to Scrolling text FX.

DedeHai:
> @DedeHai as we managed to reduce binary with Tasmota framework it may be feasible to add more fonts to Scrolling text FX.

I was recently thinking about that. Could also use a different font encoding to save unused bits (currently about 1k worth) at the cost of needing a converter script / more cryptic encoding. A more elaborate version could be to add support to read a font from a file, similar to IR.json

Myriadbits:
@blazoncek  The scrolling text/clock overlay didn't look that nice to me. I find the scrolling annoying/distracting and the font is not that clean. That is why I created this user mod. 

@DedeHai So that would mean the code will end up in the WLED code itself, not in an usermod. That will take some extra time, but is that something that the maintainers want?
Regarding the fonts (I created them myself) are more 'squarish' (if that is a correct word :) ). If size is an issue, I could remove them except the 3x6 font as I personally find that the nicest.

blazoncek:
First off, scrolling only happens if text is wider than it will fit on screen.
Second, there are 4 fonts to choose from.

DedeHai:
most if not all of what your usermod is trying to achieve can be done with the updated scrolling text, see https://github.com/wled/WLED/pull/4742
the API also allows to display custom messages AFAIK.
What is left is the fonts, if they are nicer, we could replace the current ones. Got a bitmap comparison of the fonts?

edit:
if the auto-scrolling is an issue, I was thinking of adding a "Fixed" checkmark to the FX that will center the text and not scroll even if the text does not fit.

Myriadbits:
@DedeHai : I understand your idea. Personally I like it way better if the text overlay is part of WLED core itself instead of a usermod. Adding a 'Fixed' checkmark sounds like a great idea. It could also be an option box, with a fixed time and maybe an option to show the time and date after each other (for instance each 5 seconds) and/or show the current moving time/date.
Thanks for the ideas and feedback!

I will cancel this PR and (if not already present) try to add the options inside WLED itself. (and maybe add the some fonts as well).

blazoncek:
I veto fixed text checkmark.

PR #4732, opened 2025-06-14
Maint: convert IR library to Arduino-IRremote

Graimalkin:
IRremoteESP8266 appears abandoned ( last updated July 2023 ).

It's based on Arduino-IRremote, which *IS* being maintained, and is not quite a drop in swap.

So, I swapped them. 😛 

I did have to make a choice here to either re-order all of ir_codes.h or swap the byte order for results.decodedRawData so that it could mimic results.value. I went with swapping the byte order via esp8266Value(), because changing all of ir_codes.h sounds unpleasant to review, and like it's just asking for issues.

Also added in some debugging output for the serial monitor so it's easier to see what's going on.

License changes from GNU GPL -> MIT, so that seems fine here.

This is part of a bigger change that allows WLED to work with MagiQuest wands from Great Wolf Lodge. The wands aren't working with the old IRremoteESP8266 library, but they do with this one! I'm just breaking the changes up for ease of processing. (https://github.com/Graimalkin/WLED/pull/2)

## Testing:
Tested with an ESP32 dev board, TSOP4838 IR sensor, and generic LED light strip. Also grabbed one of the white 44 key IR remotes from Amazon -- https://www.amazon.com/dp/B0982BVCSD

Colors changed, brightness goes up and down, can swap to presets.

*I'd recommend trying on a few more hardware stacks if you have it.*

image

## Summary by CodeRabbit

- **Refactor**
  - Updated infrared remote control support to use a newer IR library, improving compatibility and reliability.
  - Enhanced debug output for IR remote decoding.
  - Maintained compatibility with previous IR code values.

- **Chores**
  - Updated dependency for the IR remote library to a newer version from a different source.

### DIFF
--- a/wled00/ir.cpp
+++ b/wled00/ir.cpp
@@ -1,13 +1,13 @@
 #include "wled.h"
 
 #ifndef WLED_DISABLE_INFRARED
+#include <IRremote.hpp>
 #include "ir_codes.h"
 
 /*
  * Infrared sensor support for several generic RGB remotes and custom JSON remote
  */
 
-IRrecv* irrecv;
 decode_results results;
 unsigned long irCheckedTime = 0;
 uint32_t lastValidCode = 0;
@@ -698,34 +698,76 @@ static void decodeIR(uint32_t code)
 
 void initIR()
 {
-  if (irEnabled > 0) {
-    irrecv = new IRrecv(irPin);
-    if (irrecv) irrecv->enableIRIn();
-  } else irrecv = nullptr;
+  if (irEnabled > 0 ) {
+    IrReceiver.begin(irPin, false);
+  }
 }
 
 void deInitIR()
 {
-  if (irrecv) {
-    irrecv->disableIRIn();
-    delete irrecv;
+  IrReceiver.end();
+}
+
+// to make it easy to tell what's incoming when we're debugging
+const char* decodeTypeToStr(uint8_t type) {
+  switch (type) {
+    case LG :       return "LG";
+    case NEC:       return "NEC";
+    case SONY:      return "SONY";
+    case SAMSUNG:   return "SAMSUNG";
+    case MAGIQUEST: return "MAGIQUEST";
+    // Add other protocol cases as needed...
+    default:        return "UNKNOWN";
   }
-  irrecv = nullptr;
+}
+
+// reverse the bit-order of a single byte
+static uint8_t reverse8(uint8_t b) {
+  b = (b & 0xF0) >> 4 | (b & 0x0F) << 4;
+  b = (b & 0xCC) >> 2 | (b & 0x33) << 2;
+  b = (b & 0xAA) >> 1 | (b & 0x55) << 1;
+  return b;
+}
+
+// mimic the old ESP8266 results.value so we don't have to rebuild them
+static uint32_t esp8266Value(uint32_t raw) {
+  // shift off the low “address” byte
+  uint32_t w = raw >> 8;   // e.g. 0x00A35CFF
+  uint8_t b2 = reverse8((w >> 16) & 0xFF);
+  uint8_t b1 = reverse8((w >>  8) & 0xFF);
+  uint8_t b0 = reverse8((w      ) & 0xFF);
+  return (((uint32_t)b0 << 16)
+        | ((uint32_t)b1 <<  8)
+        | ((uint32_t)b2      ))
+        & 0x00FFFFFF;
 }
 
 void handleIR()
 {
   unsigned long currentTime = millis();
   unsigned timeDiff = currentTime - irCheckedTime;
-  if (timeDiff > 120 && irEnabled > 0 && irrecv) {
+  if (timeDiff > 120 && irEnabled > 0 ) {
     if (strip.isUpdating() && timeDiff < 240) return;  // be nice, but not too nice
     irCheckedTime = currentTime;
-    if (irrecv->decode(&results)) {
-      if (results.value != 0 && serialCanTX) { // only print results if anything is received ( != 0 )
-        Serial.printf_P(PSTR("IR recv: 0x%lX\n"), (unsigned long)results.value);
+    if (IrReceiver.decode()) {
+      auto &results = IrReceiver.decodedIRData;
+      if (results.numberOfBits > 0) {
+        if (serialCanTX && results.protocol != UNKNOWN ) { // only print results if anything is received ( != 0 )
+            Serial.printf_P(PSTR("  Protocol Received: %s\n"), decodeTypeToStr(results.protocol));
+            Serial.printf_P(PSTR("  Raw Data: %d\n"), results.decodedRawData);
+            Serial.printf_P(PSTR("  Address: 0x%X\n"), results.address);
+            Serial.printf_P(PSTR("  Command: 0x%lX\n"), (unsigned long)results.command);
+            Serial.printf_P(PSTR("  Num Bits: %d\n"), results.numberOfBits);
+
+            Serial.println();
+            Serial.println(F("========================="));
+        }
+
+        uint32_t code = esp8266Value((uint32_t)results.decodedRawData);
+        Serial.printf_P(PSTR("  Code: 0x%06lX\n"), (unsigned long)code);
+        decodeIR(code);
       }
-      decodeIR(results.value);
-      irrecv->resume();
+      IrReceiver.resume();
     }
   }
 }
--- a/wled00/ir.cpp
+++ b/wled00/ir.cpp
@@ -751,19 +751,19 @@ void handleIR()
     irCheckedTime = currentTime;
     if (IrReceiver.decode()) {
       auto &results = IrReceiver.decodedIRData;
-      uint32_t code = esp8266Value((uint32_t)results.decodedRawData);
-      if (results.numberOfBits != 0 && serialCanTX && results.protocol != UNKNOWN ) { // only print results if anything is received ( != 0 )
+      uint32_t value = esp8266Value((uint32_t)results.decodedRawData);
+      if (results.numberOfBits != 0 && serialCanTX) { // only print results if anything is received ( != 0 )
           Serial.printf_P(PSTR("  Protocol Received: %s\n"), decodeTypeToStr(results.protocol));
           Serial.printf_P(PSTR("  Raw Data: %d\n"), results.decodedRawData);
           Serial.printf_P(PSTR("  Address: 0x%X\n"), results.address);
           Serial.printf_P(PSTR("  Command: 0x%lX\n"), (unsigned long)results.command);
           Serial.printf_P(PSTR("  Num Bits: %d\n"), results.numberOfBits);
-          Serial.printf_P(PSTR("  Code: 0x%06lX\n"), (unsigned long)code);
+          Serial.printf_P(PSTR("  Code: 0x%06lX\n"), (unsigned long)value);
 
           Serial.println();
           Serial.println(F("========================="));
       }
-      decodeIR(code);
+      decodeIR(value);
       IrReceiver.resume();
     }
   }
--- a/wled00/ir.cpp
+++ b/wled00/ir.cpp
@@ -734,8 +734,14 @@ void handleIR()
 
       // we need to reverse the bits for NEC, SAMSUNG and SONY remotes to not break backwards compatability
       uint32_t value;
-      if (results.protocol == NEC || results.protocol == SAMSUNG || results.protocol == SONY) {
-        value = bitreverse32Bit(results.decodedRawData);
+      if (results.protocol == NEC || results.protocol == SAMSUNG || results.protocol == SONY)
+      { 
+        if ( results.decodedRawData == 0 ) {
+          value = 0xFFFFFFFF; // if we get a 0, we assume it is a repeat code
+        }
+        else { 
+          value = bitreverse32Bit(results.decodedRawData);
+        }
       }
       else {
         value = results.decodedRawData;
--- a/wled00/ir.cpp
+++ b/wled00/ir.cpp
@@ -751,22 +751,19 @@ void handleIR()
     irCheckedTime = currentTime;
     if (IrReceiver.decode()) {
       auto &results = IrReceiver.decodedIRData;
-      if (results.numberOfBits > 0) {
-        if (serialCanTX && results.protocol != UNKNOWN ) { // only print results if anything is received ( != 0 )
-            Serial.printf_P(PSTR("  Protocol Received: %s\n"), decodeTypeToStr(results.protocol));
-            Serial.printf_P(PSTR("  Raw Data: %d\n"), results.decodedRawData);
-            Serial.printf_P(PSTR("  Address: 0x%X\n"), results.address);
-            Serial.printf_P(PSTR("  Command: 0x%lX\n"), (unsigned long)results.command);
-            Serial.printf_P(PSTR("  Num Bits: %d\n"), results.numberOfBits);
+      uint32_t code = esp8266Value((uint32_t)results.decodedRawData);
+      if (results.numberOfBits != 0 && serialCanTX && results.protocol != UNKNOWN ) { // only print results if anything is received ( != 0 )
+          Serial.printf_P(PSTR("  Protocol Received: %s\n"), decodeTypeToStr(results.protocol));
+          Serial.printf_P(PSTR("  Raw Data: %d\n"), results.decodedRawData);
+          Serial.printf_P(PSTR("  Address: 0x%X\n"), results.address);
+          Serial.printf_P(PSTR("  Command: 0x%lX\n"), (unsigned long)results.command);
+          Serial.printf_P(PSTR("  Num Bits: %d\n"), results.numberOfBits);
+          Serial.printf_P(PSTR("  Code: 0x%06lX\n"), (unsigned long)code);
 
-            Serial.println();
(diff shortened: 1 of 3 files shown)


blazoncek:
Thank you!
I hope this also reduces the final WLED binary size. Have you compared those?
If it increases it instead it will be a tough call.
Other than that one of the possible rewrites is to convert everything to JSON based decoding.

wled00/ir.cpp
```diff
@@ -698,34 +698,76 @@ static void decodeIR(uint32_t code)
...
+      if (results.numberOfBits != 0 && serialCanTX) { // only print results if anything is received ( != 0 )
+          Serial.printf_P(PSTR("  Protocol Received: %s\n"), decodeTypeToStr(results.protocol));
+          Serial.printf_P(PSTR("  Raw Data: %d\n"), results.decodedRawData);
+          Serial.printf_P(PSTR("  Address: 0x%X\n"), results.address);
+          Serial.printf_P(PSTR("  Command: 0x%lX\n"), (unsigned long)results.command);
+          Serial.printf_P(PSTR("  Num Bits: %d\n"), results.numberOfBits);
+          Serial.printf_P(PSTR("  Code: 0x%06lX\n"), (unsigned long)value);
+
+          Serial.println();
+          Serial.println(F("========================="));
```
netmindz: Please follow the normal WLED debug logging rather than raw Serial.print*

wled00/ir.cpp
```diff
@@ -1,13 +1,13 @@
 #include "wled.h"
 
 #ifndef WLED_DISABLE_INFRARED
+#include <IRremote.hpp>
```
netmindz, line 4: includes should remain in the header file not the cpp

DedeHai:
checked for code size: looks like it saves about 3k on a C3 build in current state. 
I would request more testing. IR is a feature used in many setups.

wled00/ir.cpp
willmmiles: Wait, what?  This is a clear improvement - any changes to the library won't invalidate your whole build cache!  Include should be scoped to the file they actually matter for.

blazoncek: I do agree with @willmmiles 
WLED is a mess regarding includes. Some even my fault.

wled00/ir.cpp
blazoncek: This code is supposed to replace "IR recv" string printed out to serial with actual IR code received.
This may be needed to create custom JSON IR override. However there is no need for such detailed output as only IR code is needed.

There is one caveat we've learned on devices with CDC USB: if nobody is reading the output ESP may enter extensive waits when UART transmit buffer is full. So this piece of code may need a revision.

wled00/ir.cpp
netmindz: Sorry. I'm talking rubbish, please ignore

wled00/ir.cpp
Graimalkin: This can be slimmed down, it was in there to make it easy to debug what was going on with the IR input while testing hardware stacks.

netmindz:
While the library we were using had started life as being "based off" the library you are swapping to here, there is mention on GitHub about almost total rewrite for version 2.0

There is also the question about which remotes are supposed by which library. The change brings support for the remote you are trying to use if I'm understanding correctly, but which remotes might we loose support for by swapping?

wled00/ir.cpp
Graimalkin: > Please follow the normal WLED debug logging rather than raw Serial.print*

@netmindz - You're looking for DEBUG_PRINTF_P(), correct?

I don't see anything in the contributing docs about logging syntax, so just went with what was already in place here.

blazoncek: @Graimalkin you made it debug-like output.
Stick with only IR code and `Serial.print()` is ok. For anything else `DEBUG_PRINT...()` is the way to go.

netmindz: Sorry, that will have been my bad guidance. I didn't realise we ever used raw Serial in AC, for MM we have USER_PRINT for non-debug messages

Graimalkin: Made changes to the logging here.
- serial now only outputs Protocol received and value, and only if WLED_DEBUG isn't defined
- if WLED_DEBUG is defined, the full IR results are returned to make it easier to debug what's going on here.

Let me know if you want this cleaned up more, but I think there's value in saving it for folks who want to debug remotes that don't do the reversing bit trick. ( Since results.value is a mix of Address / Command broken out of Raw Data ).

Debugging: 
image

Non-Debug serial out:
image

Graimalkin:
> While the library we were using had started life as being "based off" the library you are swapping to here, there is mention on GitHub about almost total rewrite for version 2.0
> 
> There is also the question about which remotes are supposed by which library. The change brings support for the remote you are trying to use if I'm understanding correctly, but which remotes might we loose support for by swapping?

So right now, platformio.ini is by default enabling [these IR remote types](https://github.com/wled/WLED/blob/d2d5c423b768001e38f5cb3637bb8d34cf86b046/platformio.ini#L97):

```
  -D DECODE_HASH=true
  -D DECODE_NEC=true
  -D DECODE_SONY=true
  -D DECODE_SAMSUNG=true
  -D DECODE_LG=true
```

All of these are supported by the new Arduino-IRremote library.

Of these, it looks like in the IRremoteESP8266 (old) repo, NEC based remotes ( like my white 44 key one ), Sony, and Samsung have their results.value info come in in LSB (least significant bit) order but their address and command have been hit by reverseBits which does what it sounds like - reverses the bits. This puts these in MSB order. So when I reconstruct results.value from decodedRawData, it needs to be reversed if we want the values to match what is already in ir_codes.h.

Arduino-IRremote on converting from MSB first to LSB first -- https://github.com/Arduino-IRremote/Arduino-IRremote/tree/master#how-to-convert-old-msb-first-32-bit-ir-data-codes-to-new-lsb-first-32-bit-ir-data-codes

I did make a change to the code to only convert those remotes that I found in IRremoteESP8266 that were being reversed ( and that arduino-IRremote mentioned needing it ). You can see those changes here before the debug logging changes:
https://github.com/wled/WLED/pull/4732/commits/b0dce70a1b7548865df4be31f241ba1cd5d19423

Anyways, I went ahead and ordered a few more IR remotes that are supported per [docs](https://kno.wled.ge/interfaces/infrared/) and mentioned in [ir_codes.h](https://github.com/Graimalkin/WLED/blob/maint--upgrade-IR-library-to-arduino-IRremote/wled00/ir_codes.h). These should be here in a couple of weeks, so I'm going to mark this a DRAFT for now, and re-open when the remotes come in and I can test them out.

blazoncek:
Just so you are informed: Many users (including me) use various remotes using JSON IR remote option.
For example: I have an old Apple TV remote that I repurposed for WLED.

If the new library reverses bits all of these remotes will stop working until JSON files are updated (and accompanying JSON files in WLED-Docs) or this reversal is taken care of.

EDIT: I'm very fond of [these](https://www.aliexpress.com/item/1005001965865394.html).

Graimalkin:
> Just so you are informed: Many users (including me) use various remotes using JSON IR remote option. For example: I have an old Apple TV remote that I repurposed for WLED.
> 
> If the new library reverses bits all of these remotes will stop working until JSON files are updated (and accompanying JSON files in WLED-Docs) or this reversal is taken care of.
> 
> EDIT: I'm very fond of [these](https://www.aliexpress.com/item/1005001965865394.html).

I believe the Apple TV remote is NEC based, so would need the reversebit to work. I think it'll work with these changes as is. If you get a chance, can you check it out?

Most NEC remotes should work as-is. There are some weird 16 bit address extended NEC remotes that might potentially see issues - but I'm not sure without testing with one. 

That little remote you linked looks like another NEC remote. I grabbed these ones for testing with (In addition to the 44 key one I have):

https://www.aliexpress.us/item/2255800121523134.html?spm=a2g0o.order_list.order_list_main.10.63e318028Xdvoc&gatewayAdapt=glo2usa

image
image

blazoncek:
> If you get a chance, can you check it out?

It will be more than 2 weeks before I'm back home. Where all my equipment is.

Graimalkin:
> > If you get a chance, can you check it out?
> 
> It will be more than 2 weeks before I'm back home. Where all my equipment is.

It's all good. It's about 2 weeks until I get those remotes in also. 👍

wled00/ir.cpp
```diff
@@ -698,34 +698,64 @@ static void decodeIR(uint32_t code)
...
+      DEBUG_PRINTF_P(PSTR("IR Protocol Received: %s\n"), decodeTypeToStr(results.protocol));
+      DEBUG_PRINTF_P(PSTR("  Raw Data: %d\n"), results.decodedRawData);
+      DEBUG_PRINTF_P(PSTR("  Address: 0x%X\n"), results.address);
+      DEBUG_PRINTF_P(PSTR("  Command: 0x%lX\n"), (unsigned long)results.command);
+      DEBUG_PRINTF_P(PSTR("  Num Bits: %d\n"), results.numberOfBits);
+      DEBUG_PRINTF_P(PSTR("  Value: 0x%0lX\n"), (unsigned long)value);
+
+      #ifndef WLED_DEBUG
+        if (results.numberOfBits != 0 && serialCanTX) { // only print results if anything is received ( != 0 )
+            Serial.printf_P(PSTR("  Protocol Received: %s, Value: 0x%0lX\n"), decodeTypeToStr(results.protocol), (unsigned long)value);
```
DedeHai, line 760: I don't think the encoding type is of any use to the average user and may add more questions than it answers as this is auto-detected and nothing that is selectable. Or what's the reasoning to add it?

Graimalkin: I left the encoding type for users to be able to do an easy sanity check that the IR signal they received is from the remote device they're using. If you're trying to add a Samsung remote, you'd expect to see SAMSUNG as the encoding device. Also, since platformio.ini only enables a very *small* set of brand remotes, for folks who are trying to add a LG based remote, they would get UNKNOWN with their code - which could prompt them to dig deeper into what's going on. Remote brands that aren't covered wouldn't be decoded properly and would most likely be treated as a HASH based decode.

Graimalkin:
So I got my collection of remotes in, and I'm taking a look at this some more today.

I'm currently testing with:
- white 24-key IR remote with R,G,B + 12 color-tones
- blue 40-key IR remote with keys for 25%, 50%, 75% and 100%
- white 44-key IR remote with up/down arrows for the colors R,G and B
- black 6-key IR remote with CH up/down + Vol up/down

All of these are NEC based remotes, and they all seem to be working correctly. I can change colors, dim, turn off, and even repeat press for dimming up / down.

I also grabbed one of the TV remotes around here (Samsung) which according to platformio.ini is being decoded, to see if the codes stayed the same between MAIN / this branch. These remotes aren't in ir_codes.h, or supported in the drop down. BUT they should be getting decoded. They're going to use a variant of the NEC standard, so 🤷. Here's some testing from it:

```
Button                                 MAIN                               branch
power                                 0xE0E040BF                      0xE0E040BF
vol up                                0xE0E0E01F                      0xE0E0E01F
vol down                              0xE0E0D02F                      0xE0E0D02F
channel up                            0xE0E048B7                      0xE0E048B7
channel down                          0xE0E008F7                      0xE0E008F7
1                                     0xE0E020DF                      0xE0E020DF
2                                     0xE0E0A05F                      0xE0E0A05F
3                                     0xE0E0609F                      0xE0E0609F
```

That's about it for the remotes that I have available for testing with.

Let me know if there are next steps, or if I should just abort.

wled00/ir.cpp
blazoncek: I have to agree with @DedeHai . "Regular" user, who may be instructed to monitor serial port to see IR codes, doesn't need to know which brand his/her/its remote belongs to, he/she/it only needs to know IR code itself (to construct JSON file, for example).

However, if you insist on providing this information to the enduser, please move all of the strings into flash.

blazoncek (commented):
While I am very comfortable using IRremoteESP8266, I do not exclude switching to a library that is more robust, has smaller footprint and/or uses less resources.

When I looked at Arduino-IRremote I noticed [this](https://github.com/Arduino-IRremote/Arduino-IRremote?tab=readme-ov-file#problems-with-neopixels-fastled-etc) pragraph. If it is true, it might be a dealbreaker or may require additional safeguards implemented.

As for testing, unfortunately life took a turn and I do not have time to do any proper testing. Sorry. IMO it may be best to ask for help with testing on Discord or forum. @scottrbailey was involved with IR in the past so he hight be able to help.

DedeHai:
If you can, please test on all platforms, especially ESP8266, C3 and S2 as those have a single core.

willmmiles:
> When I looked at Arduino-IRremote I noticed [this](https://github.com/Arduino-IRremote/Arduino-IRremote?tab=readme-ov-file#problems-with-neopixels-fastled-etc) pragraph. If it is true, it might be a dealbreaker or may require additional safeguards implemented.

TL;DR: WLED will be fine.   That warning is intended for people trying to do IR and LEDs using bit-banging on slower hardware and 2017-era libraries.

--

So as it happens, I was looking through our current IR library earlier because I was curious as to whether or not it used the RMT hardware on ESP32.  Sending and receiving IR data is what the RMT device was intended for, after all!

It turns out that no, IRremoteESP8266 is already just using a generic GPIO interrupt, and timestamping when the line changed state.  Inefficient perhaps compared to using the dedicated hardware; but works well enough and isn't particularly platform specific.  (This is good for me, in the context of replacing the RMT driver, as there's no way to share the RMT interrupt between two different drivers ... )

I just took a quick look through the Arduino-IRremote code, it seems like they've taken a different approach: essentially sampling the receiver with a 20kHz (!!) timer interrupt.  This also works, but it'll be really wasteful of CPU - you pay the cost for all those interrupts even when the IR line is idle.

That said: I don't anticipate that it will cause any (new) issues with NeoPixelBus LED output.  Any time we're using real hardware (I2S, RMT, UART, whatever) we're insulated from small CPU timing gaps; and even the bit-banging driver allows interrupts through after every pixel is written.  (I should know; I PR'd that feature to NeoPixelBus myself to fix some glitching/crashes on an 8266 ;)

DedeHai:
not using hardware pin-interrupts is a very bad design...

DedeHai:
Ran a quick test, the polling eats up 10% of CPU time i.e. I see a 10-15% drop in FPS. Not acceptable.

willmmiles:
> not using hardware pin-interrupts is a very bad design...

In fairness to the devs, (ab)using a timer is almost certainly the most portable solution.  It's the equivalent of bit-banging -- highest CPU costs, most vulnerable to timing issues, but works absolutely everywhere with every protocol on any pin that can do GPIO.

Having spent time digging through FastLED and NeoPixelBus - both libraries that aim to reach for good performance everywhere, with a large number of custom drivers included - I can appreciate why a library maintainer (or team) might opt to say "no, we are not interested in being a library of optimized drivers for every chip ever -- we just do the one simple thing all the time; if you don't want that, go elsewhere".

Unfortunately, the long term impact of such a choice is that we, the developer community, never get to settle on the "one battle-tested library that everyone uses"; we end up with a panopoly of competing libraries, forcing people to choose between performance vs protocol support vs hardware support vs active maintenance vs ease-of-use .. sigh.

DedeHai:
That may have sounded wrong. Not questioning the devs of that library: if it's to be universal and for arduino purposes, using timer polling is the easiest option, no doubt. For our purposes however, it's a bad design so it can't be used.
The easy part is to read the signal, the hard part is to decode it and support all the formats. So if we get a library that just does the decoding, writing a function that reads pin interrupt timings should not be a huge effort.

willmmiles:
> Ran a quick test, the polling eats up 10% of CPU time i.e. I see a 10-15% drop in FPS. Not acceptable.

Agreed - that's a significant regression; I don't want to merge that for all users.   I would like to see some path forward for the ultimate goal here, though - my family has good memories of Great Wolf Lodge, we're hoping to visit sometime soon. ;)

As I see it, that leaves:
- Someone forks IRremoteESP8266 and adds MagiQuest protocol support, at least until the devs there get back on their feet.  (I see recent commits in the repo, but it's hard to know how active they'll be, if they'll move through the backlog).  We're already  keeping a stable of forked dependencies in the project; it's not the first time and won't be the last.
- Split the IR support out to a "core usermod".  (I'd like to do this anyways, if only to make it possible to exclude the IR library dependency entirely when it's not needed!)  Then it's possible to have two (or more) different implementations using different backing libraries via each module's `library.json`. This allows easy build time selection.  (Or potentially runtime selection, if someone includes both.)
- Both of the above?
- Any other ideas?


> The easy part is to read the signal, the hard part is to decode it and support all the formats. So if we get a library that just does the decoding, writing a function that reads pin interrupt timings should not be a huge effort.

I did a quick search and I couldn't find one - basically every library that does decoding also includes one or more hardware drivers to read into their own idiosyncratic decoding format.  (Also the hardware drivers are legitimately nontrivial -- you need precise enough timestamping, which isn't standard on most platforms, so it has to be set up by the driver.)  IMO it's probably easier to port the decoding from one library to another than it is to replace the drivers for one of the existing libraries.  YMMV, though.

https://github.com/IRMP-org/IRMP  might be worth looking at; it supports both timer-based and pin-irq-based sampling.  Although curiously the README consistently links over to IRremoteESP8266 (?!?).

Graimalkin:
Ok, going to go ahead and close this then as not a good change due to the CPU increase.

Thanks for all of the feedback folks!

PR #4735, opened 2025-06-17
Port  genPresets from WLED-MM 

netmindz:
author ewoudwijma@gmail.com

Generate presets and playlists to act as demo mode or during testing

## Release Notes

* **New Features**
  * Added a preset generator tool that automatically creates a complete presets file from all installed effects. The generator intelligently organizes effects into playlists grouped by dimension and category, with smart default parameter handling. Generated presets can be saved directly to the device via a dedicated web interface.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wled/WLED/pull/4735?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


DedeHai:
is this a feature that is useful for "daily use" or mainly for debugging? Not fully clear to me from the code what it actually can do.

blazoncek:
I would have preferred this as an external, uploadable file and not part of the standard build.

willmmiles:
> I would have preferred this as an external, uploadable file and not part of the standard build.

Hm, could it be implemented as a usermod?  `addToJsonInfo()` permits script injection.

netmindz:
> Generate presets and playlists to act as demo mode or during testing

Reading the description is more helpful than the code ;)

> > I would have preferred this as an external, uploadable file and not part of the standard build.

> Hm, could it be implemented as a usermod? addToJsonInfo() permits script injection.

The issue with making a usermod is then you need to remember to add to every build

It serves many purposes, both for formal testing and also useful for users to get a feel of what is supported

e.g Want a playlist that has all the audio effects - you've got it, want a playlist that lets you see all the new PS effects, you've got it, wanting to run a soak test of all effects  - you've got it

netmindz:
You get a playlist for each "type" of effect, 1D/2D/2F etc All 2D, AnimARTrix

netmindz:
Might be worth trying throwing it at copilot to see if it can refactor the JavaScript to make it more compact. As fewer WLED users use 2D than on MM night also be worth only creating the 2D items if you have a 2D setup

willmmiles:
> > > I would have preferred this as an external, uploadable file and not part of the standard build.
> 
> > Hm, could it be implemented as a usermod? addToJsonInfo() permits script injection.
> 
> The issue with making a usermod is then you need to remember to add to every build

Yes, that's exactly the goal.  If I'm understanding @blazoncek correctly, he's raising an objection to including this feature in all builds (presumably on space usage grounds).  Putting it in a module makes it easy to exclude if space is tight and you don't need it for your application.  Personally I'd be happy to include this as a module in the default module set for our reference build.

(Looking forward, I am still hoping to move any current core feature with a `WLED_DISABLE_X` option to a library module; call them "core mods" instead of "usermods", but the API is the same.  The implication is that we'll have some list of core modules we include in standard builds, but allow modders with custom builds to enable/disable stuff cleanly via platformio_override.ini, and minimize/reduce the #ifdef spam in the code.)

blazoncek:
Sorry for late reply. What I wanted to say is that IMO generation of presets from existing effects is not what a regular user needs.

So it does not belong into regular UI IMO.

Unfortunately we cannot have UI built in a similar fashion as object code with WLED_DISABLE_XXXX compile parameter for C++, so some other method is needed.

I'm also not sure if usermod is good way as it can't inject huge amount of text in info dialog.

DedeHai:
Would making this an external but more elaborate tool be useful? 
The approach I have used in https://github.com/wled/WLED/pull/4982 allows to download and open a tool in one click and stores it in FS. Only downside is that it requires internet connection on the client device when first loading it.

netmindz:
this version is a usermod, probably needs cleanup before suitable for anyone else, but very useful for my testing

PR #4736, opened 2025-06-17
WLED_DISABLE_2D does not compile

ingDIY:
the compilation fails with tons of errors if you try to compile using WLED_DISABLE_2D
- **message**: _enclosing class of constexpr non-static member function 'bool Segment::is2D() const' is not a literal type_
- **LineNumber**: _766_

to reproduce the issue you can _Run Build Task_ with this example environment:
```
[env:test_2D-on]
extends = env:esp32dev
build_flags = ${env:esp32dev.build_flags}
  -D WLED_MAX_BUTTONS=1
  -D BTNPIN=-1
```
it will compile, while running this:
```
[env:test_2D-off]
extends = env:esp32dev
build_flags = ${env:esp32dev.build_flags}
  -D WLED_MAX_BUTTONS=1
  -D BTNPIN=-1
  -D WLED_DISABLE_2D
```
fails.

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -763,7 +763,7 @@ class Segment {
     inline void drawCharacter(unsigned char chr, int16_t x, int16_t y, uint8_t w, uint8_t h, CRGB c, CRGB c2 = CRGB::Black, int8_t rotate = 0) const { drawCharacter(chr, x, y, w, h, RGBW32(c.r,c.g,c.b,0), RGBW32(c2.r,c2.g,c2.b,0), rotate); } // automatic inline
     inline void fill_solid(CRGB c) const { fill(RGBW32(c.r,c.g,c.b,0)); }
   #else
-    inline constexpr bool is2D() const                                            { return false; }
+    inline bool is2D() const                                                      { return false; }
     inline void setPixelColorXY(int x, int y, uint32_t c) const                   { setPixelColor(x, c); }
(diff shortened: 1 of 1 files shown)


DedeHai:
confirmed, fix is working.
@blazoncek  any objections to merging this?

willmmiles (approved):
LGTM.  At C++20 we could use `consteval` but until then we'll have to rely on dead branch elimination.

netmindz:
Do we know how common it is for people to use WLED_DISABLE_2D? Should we have a build env to confirm that works as expected? Maybe not actually release that bin, just part of our testing 

@blazoncek

blazoncek:
Back in 2022 (when implementing 2D) I wanted to keep binary small to fit OTA on ESP01 with 1M flash.
It has grown beyond that even with 2D disabled.

My suggestion is to completely remove WLED_DISABLE_2D conditional compile.

DedeHai:
I am all for removing the compile flag. It seems quite a few people are using this flag for some reason. 
@ingDIY are you using this flag or just something you discovered by accident?

ingDIY:
> I am all for removing the compile flag. It seems quite a few people are using this flag for some reason. @ingDIY are you using this flag or just something you discovered by accident?

Hello!
I am currently using WLED_DISABLE_2D. All my LEDs are running in single strips, 1D only, so I enabled this option just to remove all 2D effects from the list.
If you think it's better to remove this flag, it's ok also for me.

PR #4737, opened 2025-06-18
Bugfix: convert cctBlend value back to "%" for UI

DedeHai:
This fixes a bug where when using CCT blending the value changes each time the LED config page is saved.
Reason is:
input is in %
value is stored as 0-127
value is not converted back to % when sending it to the UI, increasing the value each time.

- added proper rounding when storing the value
- added conversion to % when sending it to UI

I checked the conversion is done correctly for all 100 input values.

### DIFF
--- a/wled00/xml.cpp
+++ b/wled00/xml.cpp
@@ -288,7 +288,7 @@ void getSettingsJS(byte subPage, Print& settingsScript)
     printSetFormCheckbox(settingsScript,PSTR("CCT"),strip.correctWB);
     printSetFormCheckbox(settingsScript,PSTR("IC"),cctICused);
     printSetFormCheckbox(settingsScript,PSTR("CR"),strip.cctFromRgb);
-    printSetFormValue(settingsScript,PSTR("CB"),Bus::getCCTBlend());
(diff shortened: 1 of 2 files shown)


DedeHai:
no backporting to 0.15 needed.

wled00/xml.cpp
```diff
@@ -288,7 +288,7 @@ void getSettingsJS(byte subPage, Print& settingsScript)
     printSetFormCheckbox(settingsScript,PSTR("CCT"),strip.correctWB);
     printSetFormCheckbox(settingsScript,PSTR("IC"),cctICused);
     printSetFormCheckbox(settingsScript,PSTR("CR"),strip.cctFromRgb);
-    printSetFormValue(settingsScript,PSTR("CB"),Bus::getCCTBlend());
+    printSetFormValue(settingsScript,PSTR("CB"),(Bus::getCCTBlend() * 100 + 64) / 127); // 100% = 127, +64 for rounding
```
blazoncek: Move that to getCCTBlend() and replace getCCTBlend() in bus_manager.cpp with Bus::_cctBlend.

PR #4738, opened 2025-06-19
Fix validation after removing a usermod

willmmiles:
When removing a usermod from an existing environment, PlatformIO does not clean up the `.pio/libdeps` folder.   This confused the `validate_modules.py` script: to avoid replicating the usermod mapping logic, it assumed every usermod library it could see in the build ought to be included in the build.   So when a usermod was removed, it would incorrectly fault claiming it wasn't included.

Instead, have `load_usermods.py` leave the actual expected module list in the environment for `validate_modules.py` to pick up.

As a side effect, this disables the ability to check for modules that aren't compatible with the current platform, as they're filtered out before the list capture in `load_usermods.py` is run.  (This is the normal behaviour of `lib_deps`, for better or for worse.)  I've just removed the nonfunctional code, as out-of-tree modules are expected to be put in `lib_deps`, so they'll have that behaviour anyways; we might as well just be consistent about it.  We can add that feature back later if it turns out to be worthwhile.

### DIFF
--- a/pio-scripts/load_usermods.py
+++ b/pio-scripts/load_usermods.py
@@ -77,17 +77,18 @@ def wrapped_ConfigureProjectLibBuilder(xenv):
   for dep in result.depbuilders:
      cached_add_includes(dep, processed_deps, extra_include_dirs)
 
+  wled_deps = [dep for dep in result.depbuilders if is_wled_module(dep)]
+
   broken_usermods = []
-  for dep in result.depbuilders:
-    if is_wled_module(dep):
-      # Add the wled folder to the include path
-      dep.env.PrependUnique(CPPPATH=str(wled_dir))
-      # Add WLED's own dependencies
-      for dir in extra_include_dirs:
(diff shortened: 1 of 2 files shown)


netmindz:
Sounds like a good improvement. Not a blocker for this PR, but is it worth us considering have some CI that tests for actions like this - building then changing the usermod list?

willmmiles:
> Sounds like a good improvement. Not a blocker for this PR, but is it worth us considering have some CI that tests for actions like this - building then changing the usermod list?

Maybe! Truthfully, `validate_modules.py` probably isn't necessary on regular day-to-day builds; now that most of the kinks with `load_usermods.py` are worked out, it only really adds value when trying to diagnose build system failures.  That said, it's fast enough that it didn't seem to be worth turning off, either -- until it became a source of build breakage itself!  We could instead develop a test sequence that applies the validation script only when running the build tests, and reduce the risk that it causes problems in its own right.  There's probably room for more extensive validation there too, if we think it's worth the effort.

PR #4741, opened 2025-06-20
fixed compiler warnings

ingDIY:
fixed compiler warnings for uninitialized vars
fixed compiler warnings for unused variables

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stability of the Particle Fireworks effect by ensuring internal variables are properly initialized, reducing the risk of unexpected behavior.

- **Refactor**
  - Enhanced configuration handling to only process EEPROM and interface settings when relevant features (such as MQTT or HueSync) are enabled, optimizing performance and resource usage.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -7856,10 +7856,11 @@ uint16_t mode_particlefireworks(void) {
   // check each rocket's state and emit particles according to its state: moving up = emit exhaust, at top = explode; falling down = standby time
   uint32_t emitparticles, frequency, baseangle, hueincrement; // number of particles to emit for each rocket's state
   // variables for circular explosions
-  [[maybe_unused]] int32_t speed, currentspeed, speedvariation, percircle;
+  [[maybe_unused]] int32_t speed = 2; // just init to min value
+  [[maybe_unused]] int32_t currentspeed, percircle;
   int32_t counter = 0;
-  [[maybe_unused]] uint16_t angle;
-  [[maybe_unused]] unsigned angleincrement;
+  [[maybe_unused]] uint16_t angle = 0; // just init to zero
+  [[maybe_unused]] unsigned angleincrement = 2730; // minimum 15°
   bool circularexplosion = false;
 
   // emit particles for each rocket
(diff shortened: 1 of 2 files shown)


DedeHai:
thanks for the input.
I do not like to increase code size just to shut up the compiler though.

ingDIY:
> thanks for the input. I do not like to increase code size just to shut up the compiler though.

I don't think that the code will be bigger just because some var is initialized and the are better managed compile masks

DedeHai:
> I don't think that the code will be bigger just because some var is initialized and the are better managed compile masks

try it.
the init values need to be stored somewhere.

ingDIY:
I tryed, and the compile output is even smaller. I am comparing the current PR with the main branch:
Immagine 2025-06-20 173819

"a", "b" and "c" are my custom builds.

Initializing variables is good practice to avoid warnings about uninitialized use, which can lead to undefined behavior.

What do you think about it?
Thanks!

DedeHai:
> Initializing variables is good practice to avoid warnings about uninitialized use, which can lead to undefined behavior.
that is true but not initializing variables that are _for sure_ initialized later serves no purpose.

I do see the sum of your commits saves on flash, did not investigate why yet. But the changes you made to FX.cpp which is adding technically unneeded inits uses a few bytes of extra flash.

ingDIY:
> > Initializing variables is good practice to avoid warnings about uninitialized use, which can lead to undefined behavior.
> that is true but not initializing variables that are _for sure_ initialized later serves no purpose.
> 
> I do see the sum of your commits saves on flash, did not investigate why yet. But the changes you made to FX.cpp which is adding technically unneeded inits uses a few bytes of extra flash. 
> 

Dear @DedeHai , technically yes, initializing variables requires flash but, in this case, we are talking about 6 or 8 bytes that compared to 4MB of flash is a very small increase.

Anyway I don't want to convince anyone about this good coding practice.😌

Anyway if we talk about flash reduction, this PR saves about 600 bytes, feel free to check it, please!

Thanks for collaborating in this project😁

DedeHai:
the reason this PR saves flash: it removes config and breaks everything.

PR #4742, opened 2025-06-20
enhancement & bugfixes in scrolling text

DedeHai:
- FX function now evaluates full string: allows custom text plus multiple tokens in any order
- fixed token evaluation order to prevent early exits (for example #YYYY was never reached as #YY was evaluated before)
- fixed day strings (argument must be weekday() )

also 
- change #MMMM to #MONL for consistency (numbers vs. strings)

As en example, this is now possible:
`It's #DAYL - #HH:#MM0:#SS0 !``

There may be a few edge cases that are displayed incorrectly, one of them is this example:
`#SSii0` is evaluated as an "seconds" token, the 0 is still within 6 chars (max token length) so it is interpreted as a "leading zero" token, the first `i` is then skipped as the correct syntax expects the `0` right after the token. 
This case could be handled but I omitted it for code simplicity.

## Summary by CodeRabbit

- **New Features**
  - Added support for new date/time tokens such as `#MONL` and `#DAYL` in scrolling text displays.

- **Refactor**
  - Improved parsing of date/time tokens for more flexible and accurate text formatting in scrolling text effects.

- **Bug Fixes**
  - Enhanced handling of zero-padding and token recognition in scrolling text, ensuring correct formatting and display.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -50,7 +50,7 @@ tabulate==0.9.0
     # via platformio
 typing-extensions==4.12.2
     # via anyio
-urllib3==2.3.0
+urllib3==2.5.0
     # via requests
 uvicorn==0.34.0
     # via platformio
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -6132,7 +6132,7 @@ uint16_t mode_2Dscrollingtext(void) {
         else if (!strncmp_P(token,PSTR("#HHMM"),5))  sprintf_P(temp, zero?PSTR("%02d:%02d")     :PSTR("%d:%02d"),    AmPmHour,         minute(localTime));
         else if (!strncmp_P(token,PSTR("#YYYY"),5))  sprintf_P(temp,          PSTR("%04d")                 ,         year(localTime));
         else if (!strncmp_P(token,PSTR("#MONL"),5))  sprintf  (temp,          ("%s")                       ,         monthStr(month(localTime)));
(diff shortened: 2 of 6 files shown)


blazoncek:
Removing #DDDD will break my presets. 🤨

blazoncek:
And expect funny issues opened when people use Unicode characters.

DedeHai:
> Removing #DDDD will break my presets. 🤨

that comes as a surprise given that DAYL was added by you 4 weeks ago :)

edit:
I have no preference, it just seemed to me like most XX and XXXX are numbers and not strings.

DedeHai:
> And expect funny issues opened when people use Unicode characters.

nothing new, it was not supported before by the scrolling text FX, nothing changed in that regard, just its now possible to add in the clock-tokens to the text.

blazoncek:
> nothing changed in that regard

It did. Now you can combine text and tokens.

DDDD was added way back. 😉

DedeHai:
I think we can handle a few more complaints :)

yes, DDDD was added before, so was your addition of the duplicate unintentional? 
I think DAYL is a better choice in naming.

blazoncek:
DDDD is known from other *worlds*.
Please *don't* remove.

blazoncek:
You misunderstood. 😄 
We can have both.

DedeHai:
why add two tokens for the same? New tokens were introduced in 0.16...

blazoncek:
Because you *can* have both.
You can, however, disregard my comments and do as you please.

DedeHai:
DDDD it is then, I see no need for a duplicate and I am fine with either version.

PR #4748, opened 2025-06-25
Revert disable OTA logic & optional Arduino OTA

blazoncek:
- new compile flag WLED_ENABLE_AOTA
- modify WLED_CONNECTED macro
- bugfix in Network isConnected() when static IP is set

With #4700 WLED_DISABLE_OTA did not entirely disable OTA, just Arduino OTA. This PR reverses that and introduces new compile flag WLED_ENABLE_AOTA (undefined by default) which (if set) enables Arduino OTA. Otherwise only HTTP OTA is available.

WLED_CONNECTED macro is modified to call `Network.isConnected()` instead of duplicating its code.

`Network.isConnected()` is updated to account if static IP is set for wired connection.

## Summary by CodeRabbit

- **New Features**
  - Added the ability to enable or disable OTA (Over-The-Air) updates at compile time, affecting both UI elements and server endpoints.
- **Bug Fixes**
  - Improved network connectivity checks to more accurately reflect WiFi and Ethernet status.
  - Prevented potential errors in UI toggling by adding safety checks for element existence.
- **Style**
  - Unified the use of feature flags and connectivity checks for OTA updates across the app.
  - Encapsulated the Software Update section in the settings UI for better structure and conditional visibility.
- **UI Improvements**
  - Updated the update button visibility in the info modal based on device capabilities.
  - Refined OTA-related UI elements to show or hide depending on OTA enablement status.

### DIFF
--- a/wled00/data/index.htm
+++ b/wled00/data/index.htm
@@ -329,7 +329,7 @@
 	<div>
 		<button class="btn ibtn" onclick="requestJson()">Refresh</button>
 		<button class="btn ibtn" onclick="toggleNodes()">Instance List</button>
-		<button class="btn ibtn" onclick="window.open(getURL('/update'),'_self');">Update WLED</button>
+		<button class="btn ibtn" onclick="window.open(getURL('/update'),'_self');" id="updBt">Update WLED</button>
 		<button class="btn ibtn" id="resetbtn" onclick="cnfReset()">Reboot WLED</button>
 	</div>
 	<br>
--- a/wled00/data/index.js
+++ b/wled00/data/index.js
@@ -685,6 +685,7 @@ function parseInfo(i) {
 		gId("filter2D").classList.remove('hide');
 		gId('bs').querySelectorAll('option[data-type="2D"]').forEach((o,i)=>{o.style.display='';});
 	}
+	gId("updBt").style.display = (i.opt & 1) ? '':'none';
 //	if (i.noaudio) {
 //		gId("filterVol").classList.add("hide");
 //		gId("filterFreq").classList.add("hide");
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -22,6 +22,9 @@
 // You are required to disable over-the-air updates:
 //#define WLED_DISABLE_OTA         // saves 14kb
 #ifdef WLED_ENABLE_AOTA
+  #if defined(WLED_DISABLE_OTA)
+    #warning WLED_DISABLE_OTA was defined but it will be ignored due to WLED_ENABLE_AOTA.
+  #endif
   #undef WLED_DISABLE_OTA
 #endif
 


blazoncek:
@KrX3D can you verify/check this PR?

KrX3D:
i tested it with both


```
#define WLED_ENABLE_AOTA
#define WLED_DISABLE_OTA
```

in all combinations and it always worked, it compiled and could be uplaoded via usb

----

but it took me some time to figure out that i need to uncheck "Only allow update from same network/WiFi"

and only looking atthe log i could find out that i need to uncheck it:


```
Jun 26 00:34:46 EleksTube WLED: Loop delayed more than 4ms.
Jun 26 00:34:46 EleksTube WLED: Attempted OTA update from different/non-local subnet!
Jun 26 00:34:46 EleksTube WLED: Attempted OTA update from different/non-local subnet!
Jun 26 00:34:46 EleksTube WLED: Attempted OTA update from different/non-local subnet!
Jun 26 00:34:46 EleksTube WLED: Attempted OTA update from different/non-local subnet!
Jun 26 00:34:46 EleksTube WLED: Attempted OTA update from different/non-local subnet!
```

since i could select the BIn file and hit update, and for 2 devices (both an s2 mini) i got 2 results.

1. it shows updating and after like 15 sec i get:
image

and need to manually reset the device. i dont know if it really was flashed

2. on the other device i got imidately after hitting the update button a "Access denied" when the page reloads


for both it is not clear that the checkbox creates the problem.

for 2. i think it is 

`static const char s_accessdenied[]   PROGMEM = "Access Denied";`
 in the wled_server.cpp? it would be nice if the message is extened to "normal" users would know why  this happened. maybe to the same text which is in the log?

for 1. not sure why it behaves differently, both are an s2 mini and are flashed with exactly the same bin file and are located in the same VLAN

blazoncek:
Thank you @KrX3D
Unfortuantely, due to security concerns, it is no longer possible to update WLED from a different subnet (VLAN) by default. Hence the checkmark in the settings page. This was introduced in #4700 and it will be mentioned in release notes once 0.16 hits the streets.

Since I am unable to perform any testing ATM due to not having access to an ESP device I have to rely on willing users to do the testing.

This PR only introduces ability to completely disable OTA updates (Access denied) or additionally enable Arduino OTA (Arduino protocol UDP 3232 IIRC).
Arduino OTA was always included in anything 0.15 or below if OTA was allowed.

Arduino OTA requires about 10k of flash and is seldom used (mostly with developers) while HTTP OTA requires 4k (14k if both are included).

Your issues might be related to browser cache.

KrX3D:
i tested those new changes and they seems also ok.

a couple more things.

 if i have checked Only allow update from same network/WiFi: and try tro flash it from a different vlan it takes like i said around 10-15 sec. during that time i get around 700 lines of:

```
Jun 26 18:55:54 EleksTube WLED: Attempted OTA update from different/non-local subnet!
Jun 26 18:55:54 EleksTube WLED: Attempted OTA update from different/non-local subnet!
Jun 26 18:55:54 EleksTube WLED: Attempted OTA update from different/non-local subnet!
```

and than the esp just hangs up and i get the page shown on the picture in my last post.

-----

and i found out why i get on those 2 devices different results.

the one where i get the picture in my last post i connected via IP 192.168.XXX.XXX and HTTP which also give the 700 lines 
and the other where i get "Access Denied" i connect via HTTPS and my reverse proxy like elekstube.lan

and by getting the access denied the log only gets 3 lines and not 700, and the esp also doesnt hang up


```
Jun 26 19:01:56 EleksTube WLED: Loop delayed more than 4ms.
Jun 26 19:01:58 EleksTube WLED: Attempted OTA update from different/non-local subnet!
Jun 26 19:01:58 EleksTube WLED: Attempted OTA update from different/non-local subnet!
Jun 26 19:01:58 EleksTube WLED: Attempted OTA update from different/non-local subnet!
Jun 26 19:01:58 EleksTube WLED: Loop delayed more than 4ms.
Jun 26 19:02:00 EleksTube WLED: Local time: 19:02
```

so its not an:

> Your issues might be related to browser cache.

and also it would be good to have on both ways the Access denied message


> Thank you @KrX3D Unfortuantely, due to security concerns, it is no longer possible to update WLED from a different subnet (VLAN) by default. Hence the checkmark in the settings page. This was introduced in #4700 and it will be mentioned in release notes once 0.16 hits the streets.
> 

its ok that the checkbox hindes to upload from another vlan, what i wanted to suggest was that the "Access denied"  which just looks like this:

image


should also mention WHY it is denied, and so it would be good to have that same message as in the log like:

```
Access denied
Attempted OTA update from different/non-local subnet!
```

so that also new users would understand why it doesnt work

blazoncek:
The primary goal of this PR (and original #4700) is related to preventing OTA from untrusted sources. This PR removes Arduino OTA from standard builds, freeing up couple of kB or flash and RAM. It also fixes a few underlying bugs. It does not change original behaviour of #4700 which is at the source for comments from @KrX3D .

So IMO notifying user about inability to update OTA from another VLAN is of secondary importance (and can be described in KB if necessary).

Are there any other objections to this PR or should it be merged to wrap what was started in #4700?

DedeHai:
no objections from me

willmmiles (approved):

rapek412-prog (approved):

netmindz:
Given we have moved from Arduino to PlatformIO, we should be removing Arduino OTA entirely. We say the only way to build is to use platformio so, no point maintaining an update method used by Arduino

blazoncek:
With this PR enabling Arduino OTA (protocol) is optional and not enabled by default.
I sill use it on my development builds to enable me OTA update from within PlatformIO environment (actually espota.py uses this protocol).

netmindz:
I thought the esp OTA was different to the Arduino OTA, if was wrong about that, then yes we keep support

Can you give an example of your platformio config where you have this working?

blazoncek:
UI OTA works using HTTP while ArduinoOTA has its own protocol (and port in use).

You can update via wget with first method while using espota.py tool for second.
~~I will provide an example when I'm back at my computer.~~

EDIT: Here you go:
```
upload_protocol = espota
;upload_port="192.168.xx.xx"
;or
upload_port="wled-test8266.local"
```

PR #4749, opened 2025-06-26
Version2 Custom Firmware

mjgriffin1113:
## Summary by CodeRabbit

* **New Features**
  * Introduced a modular pattern animation system enabling new visual effects such as custom shapes, breathing squares, spiral waves, pulsing crosses, moving stripes, corner flashes, and various frequency-controlled (Hz) effects.
  * Added hardware-reliable frequency flashing modes for physiological testing.

* **Improvements**
  * Increased the default frame rate for smoother effect animations.
  * Expanded the total number of available effect modes.

* **Bug Fixes**
  * Improved frame timing accuracy for effect updates.

* **Chores**
  * Updated default build environments for improved compatibility and build process.

* **Other**
  * Disabled the captive portal DNS server, removing the automatic splash page when using access point mode.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -7475,6 +7475,7 @@ struct Frame {
     uint16_t height;          // Pattern height  
     uint16_t baseDuration;    // Base duration in ms (modified by speed)
     uint16_t basePulseFreq;   // Base pulse frequency in Hz (modified by intensity)
+    uint8_t brightness;       // Brightness level (0-255)
 };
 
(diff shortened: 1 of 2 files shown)


blazoncek:
There are some interesting (but wasteful) approaches but overall this gave me WTF!? for disabling captive portal.

Sorry for bluntness.

WLED now supports usermods that can implement custom effects and you can find examples in the existing usermods on how to implement that.

If the usermod (or effects it produces) are requested enoug it is almost as simple as inluding those in FX.cpp in the future.

Issue #4586, opened 2025-03-04
ESP32-S3 and C3 hostname


b3-4r:
### What happened?

Hostname of WLED installed on C3 and S3 controllers is 'esp32s3-####', not 'WLED-servername' as is the case with older MCUs.

### To Reproduce Bug

Install WLED on an s3 or c3 board using the unofficial webinstaller.

### Expected Behavior

Hostname follows server description from UI settings page.

### Install Method

Self-Compiled

### What version of WLED?

WLED 0.16.0-alpha (build 2412040)

### Which microcontroller/board are you seeing the problem on?

ESP32-S3

### Relevant log/trace output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

blazoncek:
You are wrong. mDNS name is defined in WiFi settings, not UI settings.

b3-4r:
mDNS name is different from the hostname as reported to e.g. the DHCP server

blazoncek:
Only mDNS is used for communicating with device. Modified "serverDescription" as set for "hostname" is meaningless.

Rodney2K:
mDNS is different!

Start-Situation:

WiFi Settings:
mDNS -> http://      testmdns        .local
Client IP: 192.168.188.191
User Interface Settings:
server description:         WLED

After configuration save:
nslookup 192.168.188.191 ->
Server:  fritz.box
Address:  192.168.188.1
Name:    **wled-WLED**.fritz.box
Address:  192.168.188.191

 -> but after POWER CUT
nslookup 192.168.188.191
Server:  fritz.box
Address:  192.168.188.1
Name:    **esp32c3-f0cac5**.fritz.box
Address:  192.168.188.191

blazoncek:
Use mDNS explorer. It might be your Fritzbox is faulty and reports nDNS entries via DNS lookup. `nslookup` is used for regular DNS queries, not mDNS.

b3-4r:
.local is the mDNS address, the x.fritz.box addresses are FQDNs on a fritz.box network, construed as hostname.domain.tld, in their case by the fritzbox. Hostname or FQDN aren't usually something that can be used to access the webUI, rather a network-level identifier one would use e.g. to identify which devices are connected in a router's active DHCP client list.

mDNS address and network node hostname are distinct, this issue doesn't pertain to mDNS address

[...]

PR #4751, opened 2025-06-27
Hostname unification

### LINKED ISSUE #1242: Simplify "device name" across app
I do not know if the server description is anything special but it would be nice to use a single `Device ID`
to set the following

* Server name (title)
* mdns hostname
* apname

blazoncek:
- fixes #1242
- fixes empty MQTT topic
- fixes missing hostname on change #4586

## Summary by CodeRabbit

* **New Features**
  * Network Settings: hostname input + mDNS enable with validation and live mDNS URL preview.
  * Ethernet Type selector with mutual exclusivity between Ethernet and ESP-NOW.
  * MQTT client/device IDs and hostname default to unique MAC-based values when empty.

* **Improvements**
  * Consolidated hostname/mDNS handling across setup and runtime; mDNS only active when enabled.
  * Safer string handling and immediate application of WiFi options.
  * Streamlined network UI and device name validation.

* **Bug Fixes**
  * Fixed unsafe string/buffer handling and prevented unconditional mDNS updates.

* **Style**
  * Renamed settings buttons and enhanced invalid-input styling.


wled00/network.cpp
```diff
@@ -402,8 +402,8 @@ void WiFiEvent(WiFiEvent_t event)
         ETH.config(INADDR_NONE, INADDR_NONE, INADDR_NONE);
       }
       // convert the "serverDescription" into a valid DNS hostname (alphanumeric)
-      char hostname[64];
-      prepareHostname(hostname);
+      char hostname[33];
```
netmindz: Why the change from 64 to 33 and this seems out of step with other parts of the code where you are using 25

netmindz:
With this change, can we still have no mDNS but still the user-defined hostname? (used in DHCP request for "regular" DNS)

wled00/network.cpp
blazoncek: None of the UI fields are 64 bytes.

blazoncek:
> With this change, can we still have no mDNS but still the user-defined hostname?

Yes, the fallback is UI device name.

netmindz:
Yeah that doesn't quite make sense then. 

While I accept that not putting anything in there will maintain existing behaviour, I thought that part of the reason for this change is that current behaviour is unclear. 

Most users would expect the network hostname to be with the network settings, not automatically generated from a setting under UI.

So this change to move hostname to be consistent for regular DNS and mDNS is an improvement, but it would be sad if you couldn't see your own regular hostname without being forced to enable mDNS as the only way to disable is to remove the hostname

blazoncek:
Then the only option is to add a checkmark for enabling mDNS.
I would postpone that for a later time.

blazoncek:
@netmindz @willmmiles @DedeHai @softhack007 @Aircoookie I would like we all agree on the proposed change since it may disrupt some people's network configuration and we should be aware of this change.
Unfortunately I do not see a solution apart from adding another field for (non-mDNS) hostname or adding checkmark for enabling mDNS (slightly bigger change).

As a pro for this change it fixes wrong/missing hostname on change or boot.

DedeHai:
this is above my networking know-how. I exclusively work with IPs.

blazoncek:
> this is above my networking know-how. I exclusively work with IPs.

It is not really technical. This PR switches the hostname generation to mDNS name instead of UI name (by default). UI name is used as a fallback when user does not want to use mDNS and empties mDNS name.

In current situation UI name is WLED by default and if user does not change that there may be several devices with same hostname. Default mDNS name is a combination of "wled-" and MAC address.

DedeHai:
Ah I get it, still don't really understand the background, did not have time yet to read the issues solved with this. 
I never had any issues with all devices having the same name and adding a cryptic hex number to it does not change the situation from my point of view. If I need to know what device it is, I rename it manually. So if there is a technical reason to do so, no objection.

Can't really comment on "disrupting network". I do not use names, I assign fixed IPs. 

All I want to add: it must not be convoluted, like input fields need to be clear. If I change mDNS name, it should not change the host name and vice versa (if that is how it currently is). If any naming settings are linked, they MUST be in the same settings page and synced directly in that page so its obvious. 

edit: 
I glanced at the issues this is solving and uinifying makes sense.

blazoncek:
I'm pulling this to draft as perhaps we need to discuss the following:
- mDNS name
- hostname for DHCP (dnsmasq), Arduino, WiFi identification, etc
- device name (UI)

IMO hostname and mDNS name should be identical (not necessarily the case with this PR) however device name should not be used in the networking context (it is with this PR if user erases mDNS).

I have modified the entire name handling and introduced explicit mDNS selection as a POC in my fork (unpublished ATM).

wled00/data/settings_wifi.htm
```diff
@@ -189,35 +190,75 @@
...
+		</div>
+		<hr class="sml">
+		<h3>ESP-NOW Wireless</h3>
+		<div id="NoESPNOW" class="hide">
+			<i class="warn">This firmware build does not include ESP-NOW support.<br></i>
+		</div>
+		<div id="ESPNOW">
+			Enable ESP-NOW: <input type="checkbox" name="RE" onchange="tE()"><br>
+			<i>Listen for events over ESP-NOW<br>
+			Keep disabled if not using a remote or wireless sync, increases power consumption.<br></i>
```
DedeHai, line 252: name it ESPNow-Sync instead of wireless sync?

blazoncek:
@willmmiles does the latest push fulfill what we [talked about](https://github.com/wled/WLED/issues/1242#issuecomment-2950748210) (single hostname across all networking options)? It does not change UI name, though, which remains WLED by default but can be overridden during compile.

netmindz:
I think we should just have an explicit hostname field, that is then used by the DHCP client to self register against your domain used for you LAN, for example wled123.fritz.box and then also the same hostname for mDNS wled123.local 

By default it set to WLED dash mac address, with colons removed. For existing installs we populate this new field with the current logic of the sanitised version of the server name, to maintain compatibility

blazoncek:
@netmindz this is exactly what this PR does.

netmindz:
It is very close, I'm just testing at the moment, but the hostname I got for my existing install did not use the device name, but i need to double check i was in the correct state before

netmindz:
If we are going to display the full URL that your mDNS address will be, then that should be a link

netmindz:
I terms of placement, I  think it should be just before the static ip rather than right at the top

netmindz:
yeah, I've just double checked and as it stands at the moment, this PR is a breaking change, the existing way of setting the hostname does not carry over to the new setup, i just get wled-e73b10

We need to populate the new hostname value based on the existing logic

wled00/cfg.cpp
```diff
@@ -57,14 +57,22 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
 #endif
 
   JsonObject id = doc["id"];
-  getStringFromJson(cmDNS, id[F("mdns")], 33);
-  getStringFromJson(serverDescription, id[F("name")], 33);
+  // legacy behaviour
+  getStringFromJson(hostName, id[F("mdns")], sizeof(hostName));
+  if (strlen(hostName) == 0) {
+    mDNSenabled = false; // if no host name is set, disable mDNS
+    sprintf_P(hostName, PSTR("wled-%.*s"), 6, escapedMac.c_str() + 6);
```
netmindz, line 64: This needs to be changed to use the existing logic, for example my current hostname of window is because i have the WLED device called Window

netmindz (changes requested):
We need to migrate people's existing hostname

wled00/cfg.cpp
blazoncek: This uses existing logic. If `{"id":{"mdns":....}, ...}` exists, it uses that name. Only if someone disabled mDNS (by entering empty name) then new hostname is generated.

netmindz: That's using the mDNS name, personally I've never edited that as I'm using regular DNS which uses the device name from the UI settings

netmindz: Not sure which we choose, but whatever we choose it won't be the right option for 100% of users. 

Certainly if the device name in UI settings is still "WLED" then we could definitely ignore that

netmindz: the non-mDNS hostname is wled-{serverDescription}, so actually you get "wled-window" not just "window"

wled-window.ftriz.box for example - when name in server description in the UI settings is "Window"

blazoncek: `cmDNS` and `serverDescription` are only *created* during 1st boot. Each consecutive boot will use stored values if they were modified by user or not.

The only difference this PR does (regarding network naming) is the change of DHCP lease names (and dnsmasqd responses to DNS querires (non-mDNS) if your router supported that). Those were missing in case of ESP32 (producing `esp-xxxxxx` registration) or constructed from `serverDescription` (producing `wled-zzzzzzz` name, where zzzzzz conformed to `[0-9a-zA-Z\-\_]+` regexp). DNS queries were never officially supported or advertised by WLED.

The mDNS name and hostname are now (with this PR) the same (as is usual or common practice).

However, I am loosing interest in pursuing this PR further (since none of the OP responded to my queires, these include: @tablatronix, @huggy-d1, @mateuszdrab, @b3-4r, @rodney2k, @BBaoVanC, @schildbach and @colbyjdx) so do whatever pleases you (merge, edit or abandon), I am not going to update it any more.

tablatronix:
I must have this muted didn't see this

milhnl:
Hi all!

First of all, I'd like to thank @blazoncek for all the hard work. I don't intend to 'steal' his PR in any way, but saw that he didn't want to pursue this any further, so I tried to get this to a state where it can be merged. The result is over in [my fork](https://github.com/milhnl/WLED/tree/unify-hostname).

There were two issues:

- `main` had moved on, so there were some new compilation errors after rebasing. Those are fixed.
- @netmindz wanted the hostname to remain `wled-$serverDescription` if it was not set to avoid a breaking change. This behaviour is restored.

This PR was tested on both an ESP8266 and ESP32, but the latter reported `esp-xxxxxx` as hostname to `dnsmasq` on `main`, so I'm not sure whether the legacy hostname even worked there.

If you need anything more, just let me know!

blazoncek:
@milhnl you are welcome to PR into `unify-hostname` branch instead of `main`. That way no *stealing* will occur.

EDIT: FYI I disliked behaviour of `prepareHostname()`.

milhnl:
> EDIT: FYI I disliked behaviour of prepareHostname().

I agree. But not breaking backwards compatibility seems more important to the maintainers, which I get, even though it's not in a documented feature.

blazoncek:
Sometimes you need to break things to make something new. 😉

milhnl:
After that, there's two commits at in my PR. No clue how these not-to-main PRs work in the GitHub view. But they're there.

PR #4760, opened 2025-07-07
Add new effect for a single EQ bar with peak marker and decay

gwmullin:
Currently there is not anything that can duplicate the geq-style functionality for a selected frequency bin on a single segment - so I wrote one.

## Summary by CodeRabbit

* **New Features**
  * Introduced a new audio-reactive LED effect that displays a single equalizer bar, responsive to a user-selected frequency bin.
  * The effect supports both 1D and 2D LED setups, with customizable peak decay and bar fade speeds.
  * Peak levels are visually highlighted for enhanced audio visualization.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -7272,19 +7272,17 @@ uint16_t mode_single_eqbar(void) {
 
   // Draw the main bar (but not peak pixel)
   for (int i = 0; i < barHeight; i++) {
-    if (i < barHeight) {
-      // Draw the main bar according to the palette. Don't draw the top pixel, that's drawn by the peak handling.
-      SEGMENT.setPixelColor(i, SEGMENT.color_from_palette(i, true, PALETTE_SOLID_WRAP, 0));
-    }
+    SEGMENT.setPixelColor(i, SEGMENT.color_from_palette(i, true, PALETTE_SOLID_WRAP, 0));
   }
 
   if (peakDecay == 0 && barHeight > 0){
-    // No peak pixel if decay is set to zero, just draw the bar.
+    // No peak pixel if decay is set to zero, just draw the peak pixel of the bar.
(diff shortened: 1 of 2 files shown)


netmindz:
While I agree with the general sentiment that we don't want a new effect that is just a variant of an existing one, I'm unclear @blazoncek as to how we could add this new behaviour to the existing effect.

Could you please elaborate on how the user would achieve this effect using only parameters

blazoncek:
`custom3` can be used to select bin and `custom1` already defines width/# of bands.

netmindz:
> `custom3` can be used to select bin and `custom1` already defines width/# of bands.

So a slider that maps custom3 to which FFT to use, only if custom1 is sat to 1 otherwise has no affect?

blazoncek:
Yes, that's what was on my mind.

netmindz:
If you save a preset, then a WLED upgrade adds an extra param to get effect, what value is used for that param when the preset loads @blazoncek ?

0, Max or the default value set for that effect?

If would be nicer if the new slider selected the first (or maybe last) FFT value to use, rather than only working if the bar count is 1

However I'm keen not to break anyone's existing presets

blazoncek:
Depends on the "default" values specified in effect metadata. If there are none, value is not modified.

It may break only rare case that user only has width/# of bands == 1 and custom3 from previous preset is used (selecting random bin, no default value).

DedeHai:
@gwmullin please test https://github.com/wled/WLED/pull/4764

gwmullin:
> @gwmullin please test #4764

Sorry for slow reply - This doesn't appear to work at all in a single segment (I changed '2f' to '12f' in _data_FX_MODE_2DGEQ locally to allow it to appear for non-2d configs).

DedeHai:
U assume by single segment you mean 1D, it works fine on a single 2D segment. I can check if I can come up with something.

netmindz:
Please continue any further discussion on https://github.com/wled/WLED/pull/4764

PR #4761, opened 2025-07-09
Add analog in support for AR on C3, S2, S3

DedeHai:
DRAFT, only for reference, this needs to go upstream to MM first.

uses DMA sampling in the background, not compatible with analog buttons. 

## Summary by CodeRabbit

* **New Features**
  * Broader analog-microphone support across ESP32 variants; "Generic Analog" microphone option now always available on ESP32 boards.

* **Improvements**
  * Unified configuration and labeling for analog mics on ESP32.
  * Faster initialization and improved stability/recovery for analog audio input on newer ESP32 targets.


netmindz:
Not a criticism, just curious as to why you have started with a pr against here Vs MM, was it just that you had already written this code and so sharing here for the benefit of others who might want to pick up before the updated AR mod gets applied here?

DedeHai:
I have problems working with MM: I can't compile unless I base it directly off of MM, but then I can not do any commits. If I make a branch in my fork of AC, it will give me lots of compile errors when I set base to MM. So I did the development in AC and thought I might as well do a draft PR to see the diff and what the rabbit has to say. MM PR is next.

blazoncek:
@netmindz it might be in best interest of WLED to expedite migrating MM's version of AR into upstream. If that is desired at all and if there is someone that will do the migration.

Otherwise the amount of differences will only grow with time making it more difficult to migrate with each commit.

DedeHai:
now that I have a better grasp on AR code and @softhack007 is not available, I can take a stab at it. How would one go about this? cherry pick all the commits or is there a better way of grabbing the current state of only the AR files and merging them in?

I do however not approve of all features in MM AR, some are just too technical for the non-audio-people, like having 5 different FFT windows selectable at runtime for example. But that is easy to wrap in some `#ifdef`s

blazoncek:
I totally agree with you!

KIS - keep it simple
I would extend this to the code itself.

DedeHai:
I would be happy to do a sweeping cleanup of that code, but I do not want to step on anyones toes.

netmindz:
> @netmindz it might be in best interest of WLED to expedite migrating MM's version of AR into upstream. If that is desired at all and if there is someone that will do the migration.
> 
> Otherwise the amount of differences will only grow with time making it more difficult to migrate with each commit.

It's basically a 3 step process 

1. Copy current MM version to here
2. Hide experimental features
3. Add back in the AC approach for the AR palettes

That's why I don't want any extra "downstream" features added to the version in this repo or otherwise we are adding more steps to the process

netmindz:
> I have problems working with MM: I can't compile unless I base it directly off of MM, but then I can not do any commits. If I make a branch in my fork of AC, it will give me lots of compile errors when I set base to MM.

Let's try and jump on a call at some point and we can look at these issues together. I'm always swapping between the two

DedeHai:
> It's basically a 3 step process
> 
> 1. Copy current MM version to here
> 2. Hide experimental features
> 3. Add back in the AC approach for the AR palettes
> 
> That's why I don't want any extra "downstream" features added to the version in this repo or otherwise we are adding more steps to the process

fully agree that there should only be one "upstream" version that gets features added. 
Untangling the current state is a bit of a challenge. Copying the MM version will remove any git history, not fond of that. I'll check if I can do a merge of just the AR UM. Maybe git subtree is an option?

netmindz:
I agree in principle that we should use git as much as possible, however I fear that the commits might not be clean enough, so while in theory we can cherry-pick the commits, I think we might face some significant challenges.

This would of course be much simpler if we had out of tree usermods, should we perhaps split it out entirely making it a library rather than being in the source itself ?

willmmiles:
> This would of course be much simpler if we had out of tree usermods, should we perhaps split it out entirely making it a library rather than being in the source itself ?

Theoretically, the current build scripts should already support this, though I haven't tested it fully: any `lib_dep` with a name starting with "wled-" should be treated the same as an in-tree usermod for the purposes of build flags and include paths and whatnot.

willmmiles:
Following up: I just tested it with a quick rip of the example usermod, seems like it works OK:

```
[env:esp32_wrover_usermods]
extends = env:esp32_wrover
custom_usermods = audioreactive
lib_deps = ${env:esp32_wrover.lib_deps}
  wled-usermod-example =  https://github.com/willmmiles/wled-usermod-example.git
```

..builds for me with two usermods reported by the validation script.  (I'm out of town this weekend and can't test an upload, though.)  Hope this helps!

Issue #4389, opened 2024-12-13
WLED 0.15.0 Introduces Flashing/Noise on GPIOs 1, 16 on QuinLED-Dig-Quad


hacustom:
### What happened?

After upgrading a QuinLED-Dig-Quad to WLED 0.15.0, LED segments on GPIOs 1 and 16 experience significant noise and flashing.  An LED segment on GPIO 3 operates fine without any issue.

Segment 1 GPIO 3 - 629 LEDs
Segment 2 GPIO 1 - 161 LEDs
Segment 3 GPIO 16 - 171 LEDs

This particular setup has been running flawlessly for just over 3 years.  The issue was immediately resolved by downgrading to WLED 0.14.3.

### To Reproduce Bug

- Upgrade a QuinLED-Dig-Quad to WLED 0.15.0
- Activate any preset or animation
- Flashing/Noise is prominent on LED segments on GPIO 1 and 16

### Expected Behavior

- No noise/flashing on any LED segment

### Install Method

Binary from WLED.me

### What version of WLED?

WLED 0.15.0 2412100

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell
I don't have any logs at the moment, I have since downgraded to WLED 0.14.3.  However, I would be more than happy to re-upgrade and capture any logs/traces/configs that would aid in troubleshooting.
```

### Anything else?

This appears to be similar to report #3976.  That was submitted for 0.15.0b2 and not the full release however.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

blazoncek:
GPIO 1 & 16 will use RMT. RMT is susceptible to flickering in some circumstances.
0.15 added a lot of code that may make those circumstances more frequent.

GPIO 3 uses I2S.

hacustom:
Interesting.  

I had a spare Dig-Quad on hand, and I upgraded it to WLED 0.15.0 and configured the LED outputs to be identical as the main one for my house lights.  I wired up a spare string of LEDs I had leftover and I'm seeing the same behavior as with my other - tons of flashing/noise.  If I stick to a solid color, it's OK.  There's a bunch of noise every time I change color but once it reaches the target color the noise stops.  Any animation is unusable.

hacustom:
Just for kicks, I setup my spare Dig-Quad from scratch, installing it from https://install.quinled.info/dig-quad/ just to see if it made any difference and it didn't.

blazoncek:
[...]

Issue #4315, opened 2024-11-24
Mixed bus timings in parallel I2S cause incorrect output


willmmiles:
When:
- On ESP32
- Using more than 5 digital output buses, so the system switches to parallel I2S
- Buses of different timings are selected

Only the timing of the first digital bus will be used, resulting in incorrect timings for later buses.

The automatic switchover to using parallel I2S should be aware of the individual bus type timings, and fall back to using RMT channels when buses of different timings are selected.  Otherwise incorrect timings may be used.

Reference:

>>> @Makuna (hope your reading this) we are still not 100% sure if mixing different led strip types on the same "parallel I2S bus" is supported (esp32), or it may lead to errors and spurious crashes? Our most important use case would be to have both ws2812b (RGB) and sk6812 (RGBW) on the same parallel I2S driver.
>> 
>> It this "safe", or would you say better to use the parallel bus for only one type, and have single RMT busses for other types?
>
>The "Feature" can be unique per parallel channel without any concern.  So, mixing RGB with RGBW or even 8 bit per element and 16 bit per element isn't an issue.
>The "Method" uses the i2s bus speed of the first NPB instanced (constructed) on that i2s bus as the speed for all channels that are latter constructed on that i2s bus.  The reset time can be unique per i2s parallel channels without issues.  The ramification of this is that not all methods use the same speed.  In the NPB file "NeoBits.h" is the list and the key class member is "BitSendTimeNs".  You can see that WS2805 has a different value than the WS2812x, but SK6812 has the same as WS2812x.  So, if the speed is the same, there is no issue at all.  If the speed is different, the only effect is that it will run at the speed of the first one instanced.  Should not crash or cause other issues than timing.
>NOW, with this all stated, the big caveat is that all parallel channels must be instanced (constructed) before the first Begin() is called and no other are instanced after the first Begin() is called.  That could cause issues.

_Originally posted by @Makuna in https://github.com/Aircoookie/WLED/issues/4312#issuecomment-2496030173_

softhack007:
🤔 Thinking about short-term fixing, it would be "save" to only allow ws2812b and sk6812 in the parallel bus driver, and fall back to RMT for any other type. 

This would prevent timing issues with the parallel bus, and it would give us some more time to implement a better solution   after 0.15 is released.

willmmiles:
[...]

PR #4762, opened 2025-07-09
bus_wrapper: Use parallel I2S first when enabled

willmmiles:
On ESP32, if parallel I2S is enabled, use it first, before selecting RMT outputs.  This provides a quick workaround for RMT output issues (#4389), though it doesn't warn about #4315.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for ESP32 devices when using parallel I2S output, ensuring correct channel mapping and bus selection for various ESP32 variants.

### DIFF
--- a/wled00/bus_wrapper.h
+++ b/wled00/bus_wrapper.h
@@ -1436,7 +1436,8 @@ class PolyBus {
       // ESP32-S2 only has 4 RMT channels
       if (_useParallelI2S) {
         if (num > 11) return I_NONE;
-        if (num < 8) offset = 1;    // use x8 parallel I2S0 channels (use last to allow Audioreactive)
+        if (num < 8) offset = 1;    // use x8 parallel I2S0 channels followed by RMT
+                                    // Note: conflicts with AudioReactive if enabled
       } else {
         if (num > 4) return I_NONE;
         if (num > 3) offset = 1;  // only one I2S0 (use last to allow Audioreactive)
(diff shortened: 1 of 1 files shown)


wled00/bus_wrapper.h
```diff
@@ -1441,15 +1449,15 @@ class PolyBus {
...
+        if (num < 8) offset = 1;    // use x8 parallel I2S LCD channels
       } else {
         if (num > 3) return I_NONE; // do not use single I2S (as it is not supported)
       }
       #else
       // standard ESP32 has 8 RMT and x1/x8 I2S1 channels
       if (_useParallelI2S) {
         if (num > 15) return I_NONE;
-        if (num > 7) offset = 1;  // 8 RMT followed by 8 I2S
+        if (num < 8) offset = 1;  // 8 RMT followed by 8 I2S
```
blazoncek: Update comment.

wled00/bus_wrapper.h
```diff
@@ -1428,7 +1436,7 @@ class PolyBus {
       // ESP32-S2 only has 4 RMT channels
       if (_useParallelI2S) {
         if (num > 11) return I_NONE;
-        if (num > 3) offset = 1;    // use x8 parallel I2S0 channels (use last to allow Audioreactive)
+        if (num < 8) offset = 1;    // use x8 parallel I2S0 channels (use last to allow Audioreactive)
```
blazoncek: Comment

wled00/bus_wrapper.h
willmmiles: Arg, sorry for the sloppiness.  Thanks for catching it!

blazoncek:
IMO this should also go into 0_15_x branch

netmindz:
Unfortunately it won't cherry-pick without some conflict resolution, so I have not merged into 0_15_x 

Are you able to cherry-pick and resolve @willmmiles ?

Arcitec:
@netmindz I've cherry picked and rebased the fix to 0.15.x here: #4766

Thank you Will! If I understand correctly, these changes makes WLED prefer Parallel I2S instead of RMT to avoid the RMT flickering issues. I see that it selects offset 1 (I2S) when the channel number `num` is `0 - 7`. So if the channel number is 8 or higher, it would use offset 0 (RMT).

In other words, it seems like everyone will be on Parallel I2S output mode now unless they have 9 or more channels, after which the last channels will start using RMT?

Good change! :+1:

PR #4763, opened 2025-07-09
Prevent parallel I2S use if different LED types are used.

blazoncek:
- helps with #4315
- add-on to #4762
- increases parallel I2S LED limit to 600 per bus

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1203,21 +1203,27 @@ void WS2812FX::finalizeInit() {
   #if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32C3)
   // determine if it is sensible to use parallel I2S outputs on ESP32 (i.e. more than 5 outputs = 1 I2S + 4 RMT)
   unsigned maxLedsOnBus = 0;
+  unsigned busType = 0;
   for (const auto &bus : busConfigs) {
     if (Bus::isDigital(bus.type) && !Bus::is2Pin(bus.type)) {
       digitalCount++;
+      if (busType == 0) busType = bus.type; // remember first bus type
+      if (busType != bus.type) {
+        DEBUG_PRINTF_P(PSTR("Mixed digital bus types detected! Forcing single I2S output.\n"));
+        useParallelI2S = false; // mixed bus types, no parallel I2S
+      }
       if (bus.count > maxLedsOnBus) maxLedsOnBus = bus.count;
     }
   }
   DEBUG_PRINTF_P(PSTR("Maximum LEDs on a bus: %u\nDigital buses: %u\n"), maxLedsOnBus, digitalCount);
-  // we may remove 300 LEDs per bus limit when NeoPixelBus is updated beyond 2.9.0
-  if (maxLedsOnBus <= 300 && useParallelI2S) BusManager::useParallelOutput(); // must call before creating buses
+  // we may remove 600 LEDs per bus limit when NeoPixelBus is updated beyond 2.8.3
+  if (maxLedsOnBus <= 600 && useParallelI2S) BusManager::useParallelOutput(); // must call before creating buses
   else useParallelI2S = false; // enforce single I2S
+  digitalCount = 0;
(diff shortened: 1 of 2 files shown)


blazoncek:
Large led count may need thorough testing with several usermods enabled.
AFAIK minimal free heap is about 50k for AWS to function properly. S2 may struggle with that.

blazoncek:
@coderabbitai please review

netmindz:
would we also want to merge this into main @blazoncek ?

Arcitec:
@netmindz 

- Disabling parallel I2S when different LED chips are mixed looks like a correct change.
- If there's any doubt, _that_ change could be extracted into a separate PR and merged separately since it looks like a good and easily verifiable change.
- The code that increases LED limit and changes the "memory usage algorithm" is confusing. The most confusing part is the new Javascript functions (such as `is32()` and `isS2()`) which analyzes some limits to try to detect the board model. Is there any better way to detect the board model from Javascript?
- Increasing the LED limit might still cause memory overflows on some chips. Probably not though. Blaz thinks that memory overflow might only be a problem on ESP32-S2.

blazoncek:
@netmindz yes

Arcitec:
I have now ran a local test with WLED `0_15_x` branch with a custom backport of all of these latest fixes:

- Blaz's fix for the I2S color conversion bug: https://github.com/wled/WLED/commit/232dc044e795bf26e78b6a257968c57f36599732
- Will's change which makes Parallel I2S the default instead of RMT: https://github.com/wled/WLED/pull/4762
- Blaz's PR (this one) which disables Parallel I2S if the output channels use different LED ICs, and also raises the Parallel I2S LED IC limit from 300 to 600 and changes the LED Settings HTML page to calculate and detect chips differently.

image

---

Test Results:

- The test was done on a `ESP32-WROOM-32E(16MB)`.
- LED outputs are on GPIO 16, 12, 4, 2.
- I tested all outputs and they all display correct colors without any flickering.

---

I also had a second look at the LED Settings HTML changes of this PR. I see now that it receives the limit from an API call and then looks at the Max Analog, Max Digital and Max Virtual bus counts to identify the chip type.

That solution works, but if those limits ever change, there's a risk of subtle bugs in the future where someone may forget to also change the chip identification in `settings_leds.htm`. Then the chip can be subtly misidentified. One of the variables is even dynamic in the code: `#define WLED_MAX_ANALOG_CHANNELS (LEDC_CHANNEL_MAX*LEDC_SPEED_MODE_MAX)`. Sure, it's not likely that they will change, but it's still a bit risky to look directly at changeable parameters from a different file to identify chips.

I think that part of the code should be changed to add a backend API change which returns the chip as a string, such as `"ESP32"`, `"ESP8266"`, `"ESP32-C3"` etc, by putting the chip "ID/name" in a `#define` string on the backend whose contents is based on the compile-flags such as `CONFIG_IDF_TARGET_ESP32S2`.

And then make the backend settings API return that string in the "get current settings" response.

And then change `settings_leds.htm` from this:

```js
		function is8266() { return maxA ==  5 && maxD ==  3; } // NOTE: see const.h
		function is32()   { return maxA == 16 && maxD == 16; } // NOTE: see const.h
		function isC3()   { return maxA ==  6 && maxD ==  2; } // NOTE: see const.h
		function isS2()   { return maxA ==  8 && maxD == 12 && maxV == 4; } // NOTE: see const.h
		function isS3()   { return maxA ==  8 && maxD == 12 && maxV == 6; } // NOTE: see const.h
```

To this:

```js
		function is8266() { return chipType === "ESP8266"; }
		function is32()   { return chipType === "ESP32"; }
		function isC3()   { return chipType === "ESP32-C3"; }
		function isS2()   { return chipType === "ESP32-S2"; }
		function isS3()   { return chipType === "ESP32-S3"; }
```

Other than that part, I think everything here looks very good. Thank you Blaz for being such an amazing dude and amazing developer. :heart:

However, I am not able to test the memory limits to check for overflows with huge amounts of LEDs and many usermods. Is that only likely to be a problem on ESP32-S2 with many usermods enabled at the same time, though?

PR #4764, opened 2025-07-10
Adding center bin selection to 2D GEQ

DedeHai:
setting custom3 to 0 gives the "old" behaviour, this is the default for new presets. Existing presets will have the custom3 slider at the center, changing presets that do not use the full width.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -7258,11 +7258,10 @@ uint16_t mode_2DGEQ(void) { // By Will Tatam. Code reduction by Ewoud Wijma.
 
   for (int x=0; x < cols; x++) {
     int band = map(x, 0, cols, 0, NUM_BANDS);
-    //if (NUM_BANDS < 16) band = map(band, 0, NUM_BANDS - 1, 0, 15); // always use full range. comment out this line to get the previous behaviour.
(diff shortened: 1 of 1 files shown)


DedeHai:
this is the proper implementation of https://github.com/wled/WLED/pull/4760 and replaces that PR.

wled00/FX.cpp
```diff
@@ -7235,6 +7235,7 @@ uint16_t mode_2DGEQ(void) { // By Will Tatam. Code reduction by Ewoud Wijma.
   if (!strip.isMatrix || !SEGMENT.is2D()) return mode_static(); // not a 2D set-up
 
   const int NUM_BANDS = map(SEGMENT.custom1, 0, 255, 1, 16);
+  const int CENTER_BIN = map(SEGMENT.custom3, 0, 31, 0, 15);
```
netmindz, line 7238: Can we use the constant rather than a magic number or is AC missing that constant?

DedeHai: I never came across such a constant if it exists. It could be defined at the top of FX.cpp.

blazoncek: It is in the usermod. I have moved AR effects into usremod in my fork which would make such use of constant easier.

PR #4768, opened 2025-07-12
Add mDNS resolution for network bus

### LINKED ISSUE #4268: DDP with mdns names - feature request
Sometimes i need to change the IP of my WLED Controller. Since i am using DDP i need to change the according ip in the DDP settings

A test field which could be filled with the IP or the mDNS names like wled-1234.local

fixed IP - but in rare cases like changes of the router fixed IPs are lost.

Thank you for your ideas for making WLED better!

blazoncek:
Fixes #4268

Only for ESP32

## Summary by CodeRabbit

* **New Features**
  * Added support for specifying a hostname or custom text for each network-based LED output (excluding ESP8266 devices).
  * Hostnames are now resolved to IP addresses automatically on ESP32 devices for network LED outputs.
  * The web UI for LED outputs now includes a "Host" input field for compatible devices.

* **Improvements**
  * Custom text/hostname is now saved and restored with LED bus configurations.
  * Invalid input fields are visually highlighted in red for easier identification.

* **Bug Fixes**
  * Minor typo and formatting corrections in comments and code.

### DIFF
--- a/wled00/bus_manager.h
+++ b/wled00/bus_manager.h
@@ -133,6 +133,7 @@ class Bus {
     virtual uint16_t getUsedCurrent() const                     { return 0; }
     virtual uint16_t getMaxCurrent() const                      { return 0; }
     virtual size_t   getBusSize() const                         { return sizeof(Bus); }
+    virtual const String getCustomText() const                  { return String(); }
 
     inline  bool     hasRGB() const                             { return _hasRgb; }
     inline  bool     hasWhite() const                           { return _hasWhite; }
@@ -215,7 +216,7 @@ class Bus {
     uint8_t  _autoWhiteMode;
     // global Auto White Calculation override
     static uint8_t _gAWM;
-    // _cct has the following menaings (see calculateCCT() & BusManager::setSegmentCCT()):
+    // _cct has the following meanings (see calculateCCT() & BusManager::setSegmentCCT()):
     //    -1 means to extract approximate CCT value in K from RGB (in calcualteCCT())
     //    [0,255] is the exact CCT value where 0 means warm and 255 cold
     //    [1900,10060] only for color correction expressed in K (colorBalanceFromKelvin())
@@ -375,8 +376,9 @@ struct BusConfig {
   uint16_t frequency;
   uint8_t milliAmpsPerLed;
   uint16_t milliAmpsMax;
+  String text;
 
-  BusConfig(uint8_t busType, uint8_t* ppins, uint16_t pstart, uint16_t len = 1, uint8_t pcolorOrder = COL_ORDER_GRB, bool rev = false, uint8_t skip = 0, byte aw=RGBW_MODE_MANUAL_ONLY, uint16_t clock_kHz=0U, uint8_t maPerLed=LED_MILLIAMPS_DEFAULT, uint16_t maMax=ABL_MILLIAMPS_DEFAULT)
+  BusConfig(uint8_t busType, uint8_t* ppins, uint16_t pstart, uint16_t len = 1, uint8_t pcolorOrder = COL_ORDER_GRB, bool rev = false, uint8_t skip = 0, byte aw=RGBW_MODE_MANUAL_ONLY, uint16_t clock_kHz=0U, uint8_t maPerLed=LED_MILLIAMPS_DEFAULT, uint16_t maMax=ABL_MILLIAMPS_DEFAULT, String sometext = "")
   : count(std::max(len,(uint16_t)1))
   , start(pstart)
   , colorOrder(pcolorOrder)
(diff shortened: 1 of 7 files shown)


DarthWeber:
Thanks - but which branch do i need? I cant find "fix-4768" and in the main branch i see no changes (only ip field).

DedeHai:
@DarthWeber you can compile this PR or donwnload the compiled bins under the "Checks" tab -> Artifacts
FYI: any (recent) commit has compiled bins for download, click on the checkmark to get them:
image

DarthWeber:
Thanks, but also with the bin under artifacts e7157e5 i got no host field under the ddp segment. Seems to be without PR, the build is still 2412040

DarthWeber:
got it compiled. But it doesn't work as expected: I need to fill in the ip. I can not let them empty. DDP works only if i fill in the right ip. Double-checked the hostname, it is right.

Screenshot 2025-07-22 122715

And another strange thing: The multirelay usermod is not available although i use -D USERMOD_MULTI_RELAY. It is available with 15-beta1.  But that is maybe another issue.

blazoncek:
enter anything in IP fields.

DarthWeber:
i tried 8.8.8.8 and 0.0.0.0 - but it doesnt work

blazoncek:
Well, it does in my case. And in fact it replaces IP with device's actual IP when you review settings.
I would expect that mDNS issues on your network.

DarthWeber:
got it. I deleted the DDP output, entered the hostname and 8.8.8.8 and i got the resolved ip after saving. But this works only the first time an a new output. If you enter something in the ip again, it does not resolve again the hostname.

blazoncek:
To prevent resource exhaustion new mDNS resolution is done after 10 min.
Otherwise you would see LED stuttering due to delays.

DarthWeber:
Works since 24h like a charme. Only two minor problems:
- in that release i cant include the multirelay usermod. 
- one or two stucks on the receiving controller, whis is still 0.15-beta1

blazoncek:
Is everyone comfortable with merging this?
It does have its quirks. Namely, when you reboot and ESP starts, buses are created *before* there is WiFi connectivity. This means mDNS is not used to resolve IP address and stored IP is used instead. However, to mitigate this mDNS is retried later when you turn WLED off and on again.

willmmiles (approved):
I'm happy to accept this as-is.  We can arrange to resolve on the network coming up in a future patch.

wled00/data/settings_leds.htm
```diff
@@ -52,6 +52,11 @@
 			maxL  = l; // maxL - max LEDs (will serve to determine ESP >1664 == ESP32)
 			maxCO = o; // maxCO - max Color Order mappings
 		}
+		function is8266() { return maxA ==  5 && maxD ==  3; } // NOTE: see const.h
+		function is32()   { return maxA == 16 && maxD == 16; } // NOTE: see const.h
+		function isC3()   { return maxA ==  6 && maxD ==  2; } // NOTE: see const.h
+		function isS2()   { return maxA ==  8 && maxD == 12 && maxV == 4; } // NOTE: see const.h
+		function isS3()   { return maxA ==  8 && maxD == 12 && maxV == 6; } // NOTE: see const.h
```
softhack007, line 59: This looks very fragile. Especially if someone changes the number of virtual buses by overriding the value in const.h, then "isS3" will stop working.

wled00/bus_manager.h
```diff
@@ -368,8 +376,9 @@ struct BusConfig {
   uint16_t frequency;
   uint8_t milliAmpsPerLed;
   uint16_t milliAmpsMax;
+  String text;
 
-  BusConfig(uint8_t busType, uint8_t* ppins, uint16_t pstart, uint16_t len = 1, uint8_t pcolorOrder = COL_ORDER_GRB, bool rev = false, uint8_t skip = 0, byte aw=RGBW_MODE_MANUAL_ONLY, uint16_t clock_kHz=0U, uint8_t maPerLed=LED_MILLIAMPS_DEFAULT, uint16_t maMax=ABL_MILLIAMPS_DEFAULT)
+  BusConfig(uint8_t busType, uint8_t* ppins, uint16_t pstart, uint16_t len = 1, uint8_t pcolorOrder = COL_ORDER_GRB, bool rev = false, uint8_t skip = 0, byte aw=RGBW_MODE_MANUAL_ONLY, uint16_t clock_kHz=0U, uint8_t maPerLed=LED_MILLIAMPS_DEFAULT, uint16_t maMax=ABL_MILLIAMPS_DEFAULT, String sometext = "")
```
softhack007, line 381: some text? I thought this is about an mDNS name.
Would it be possible to find a more meaningful name for this parameter?

wled00/bus_manager.h
```diff
@@ -133,6 +133,7 @@ class Bus {
     virtual uint16_t getUsedCurrent() const                     { return 0; }
     virtual uint16_t getMaxCurrent() const                      { return 0; }
     virtual size_t   getBusSize() const                         { return sizeof(Bus); }
+    virtual const String getCustomText() const                  { return String(); }
```
softhack007, line 136: Two ideas for improvement:
- Return a const string & (aka reference) to avoid string duplication in heap
- return text, instead of an empty string. Or add a comment to explain why this must return an empty string.

PR #4769, opened 2025-07-12
Add mDNS support for MQTT server

### LINKED ISSUE #4671: Resolve mDNS address for MQTT broker
When connecting to an MQTT broker, please allow putting an mDNS domain into the "broker" field. Currently (on version 0.15.1.beta2) it seems only IP addresses and regular DNS names are supported.

blazoncek:
- fixes #4671
- reduce some topic string parsing
- moves LWT into onConnect

## Summary by CodeRabbit

* **New Features**
  * Improved mDNS support for MQTT server resolution on ESP32 devices.

* **Improvements**
  * Unified and streamlined MQTT topic formatting for consistent and reliable topic generation.
  * Enhanced reliability of online status reporting via MQTT by centralizing Last Will and Testament message setup.

* **Documentation**
  * Added and clarified comments for better understanding of MQTT payload parsing and logic.

### DIFF
--- a/wled00/mqtt.cpp
+++ b/wled00/mqtt.cpp
@@ -55,8 +55,8 @@ static void onMqttConnect(bool sessionPresent)
   DEBUG_PRINTLN(F("MQTT ready"));
 
 #ifndef USERMOD_SMARTNEST
-  snprintf_P(mqttStatusTopic, sizeof(mqttStatusTopic)-1, sTopicFormat, MQTT_MAX_TOPIC_LEN, mqttDeviceTopic, "status");
-  mqtt->setWill(mqttStatusTopic, 0, true, "offline"); // LWT message
+  snprintf_P(subuf, sizeof(subuf)-1, sTopicFormat, MQTT_MAX_TOPIC_LEN, mqttDeviceTopic, "status");
+  mqtt->publish(subuf, 0, true, "online"); // retain message for a LWT
 #endif
 
   publishMqtt();
@@ -231,6 +231,10 @@ bool initMqtt()
   mqtt->setClientId(mqttClientID);
   if (mqttUser[0] && mqttPass[0]) mqtt->setCredentials(mqttUser, mqttPass);
 
+  #ifndef USERMOD_SMARTNEST
+  snprintf_P(mqttStatusTopic, sizeof(mqttStatusTopic)-1, sTopicFormat, MQTT_MAX_TOPIC_LEN, mqttDeviceTopic, "status");
+  mqtt->setWill(mqttStatusTopic, 0, true, "offline"); // LWT message
+  #endif
   mqtt->setKeepAlive(MQTT_KEEP_ALIVE_TIME);
   mqtt->connect();
   return true;


schildbach:
So far, I had no luck with this PR. I installed `WLED_0.16.0-alpha_ESP32.bin` from this PR to my Athom LS8P, replaced Sync Settings → MQTT → Broker with "assistant.local" and rebooted. It does not seem to connect to my MQTT server.

There is no such problem if I use a regular DNS name or IP address. And a variety of ESPHome devices can connect using the mDNS name without problem as well.

blazoncek:
You don't need to add `.local` to host name.

schildbach:
> You don't need to add `.local` to host name.

How do you mean? That's what my mDNS domain is called, and broadcasted via avahi-publish.

blazoncek:
.local is needed for your host computer to know it needs to do mDNS. querying mDNS directly does not need it.

schildbach:
> .local is needed for your host computer to know it needs to do mDNS. querying mDNS directly does not need it.

Yes exactly. How does WLED know I want to resolve an mDNS address if I don't add the correct `.local` domain? Domains without `.local` are plain DNS so those would try to get resolved via my DNS server – and fail, right?

blazoncek:
Just enter the hostname. What else? Similar as, if you are old enough to remember it, WINS.
Will add stripping of .local

schildbach:
> Just enter the hostname. What else? Similar as, if you are old enough to remember it, WINS. Will add stripping of .local

Again, how will it know it's an mDNS domain if I strip the `.local`?

blazoncek:
See the code.

schildbach:
I see changes to MQTT topic parsing which seem irrelevant to this PR. However I'm a Java dev with limited understanding of CPP code so I can't say for sure.

Can you describe in words?

Typically, in these kind of hostname fields there are these cases:

 * formatted as IPv4 → use IP
 * formatted as IPv6 → use IP
 * ending with `.local` → resolve IP via mDNS
 * just host, no domain → add DNS search domain, resove IP via DNS
 * host.domain → resolve IP via DNS

wled00/mqtt.cpp
```diff
@@ -214,8 +214,10 @@ bool initMqtt()
     mqtt->setServer(mqttIP, mqttPort);
   } else {
     #ifdef ARDUINO_ARCH_ESP32
-    if (strlen(cmDNS) > 0 && strchr(mqttServer, '.') == nullptr) { // if mDNS is enabled and server does not have domain
-      mqttIP = MDNS.queryHost(mqttServer);
+    String mqttMDNS = mqttServer;
+    mqttMDNS.replace(F(".local"), ""); // remove .local if present
+    if (strlen(cmDNS) > 0 && mqttMDNS.length() > 0 && mqttMDNS.indexOf('.') < 0) { // if mDNS is enabled and server does not have domain
```
schildbach: Is `.replace()` limited to the end of strings? Otherwise it would strip in the middle of DNS names too which is probably not wanted.

In general, I'd recommend explicitly checking for `.local` at the end, and only if present resolve via mDNS (stripping the domain if it isn't done by the mDNS component anyway). It seems the above code just checks for the presence of dots, which is imho not correct. Plain hostnames (without dots) are typically expected to be resolved via DNS (using the search domain).

schildbach: Also, the impl from this PR would fail for those who use dots in their mDNS name. Yes, some people do this and I understand this is bending the mDNS spec a bit, but it would be nice to support these cases anyway.

schildbach: One more comment about `.replace()`: is it case-insensitive? Both DNS and mDNS names/domains are case-insensitive.

blazoncek: You are welcome to add on to this at a later time. For now just test if it fulfills your needs with .local domain.
You have to understand that I'm doing that in my spare time and I have no need for this feature and limited resources in ESP may need a few compromises.

FYI resolving mDNS host only requires hostname, no domain. Domain is only there to determine if regular DNS is needed or not.

schildbach: Fair enough, and thanks for your time investment. I was just reviewing this PR as you asked me to do.

schildbach:
I retested the 24f2306 binary and `assistant.local` now works for me.

schildbach:
However, `assistant.LOCAL` doesn't work although I think it should.

`ASSISTANT.local` works.

blazoncek:
> `.local` part needs to appear _at the end_ of the string

Agreed, but that is a bit more complex. However, that stripping is only done for mDNS resolution. If host is not found, regular DNS query is made. See `else`.

rapek412-prog (approved):

PR #4776, opened 2025-07-14
using segment buffer instead of local buffer for PS 

DedeHai:
tested up to 8000 LEDs and PS still works on a single segment. 
Speed is about the same, even slightly improved but insignificant.

## Summary by CodeRabbit

* **New Features**
  * Enhanced particle system rendering to support RGBW color output for more vibrant and accurate lighting effects.
  * Added a method to calculate maximum mapping length for 2D segments to improve buffer allocation.

* **Improvements**
  * Particle emission now starts from the bottom of the frame for a more visually intuitive effect.
  * Internal optimizations to color handling and framebuffer management for improved performance and maintainability.

* **Compatibility**
  * Updated integration of particle effects with RGBW LED hardware.

### DIFF
--- a/wled00/FXparticleSystem.cpp
+++ b/wled00/FXparticleSystem.cpp
@@ -17,8 +17,8 @@
 // local shared functions (used both in 1D and 2D system)
 static int32_t calcForce_dv(const int8_t force, uint8_t &counter);
 static bool checkBoundsAndWrap(int32_t &position, const int32_t max, const int32_t particleradius, const bool wrap); // returns false if out of bounds by more than particleradius
-static void fast_color_add(CRGBW &c1, const CRGBW &c2, uint8_t scale = 255); // fast and accurate color adding with scaling (scales c2 before adding)
-static void fast_color_scale(CRGBW &c, const uint8_t scale); // fast scaling function using 32bit variable and pointer. note: keep 'scale' within 0-255
+static uint32_t fast_color_add(CRGBW c1, const CRGBW c2, uint8_t scale = 255); // fast and accurate color adding with scaling (scales c2 before adding)
+static uint32_t fast_color_scale(CRGBW c, const uint8_t scale); // fast scaling function using 32bit variable and pointer. note: keep 'scale' within 0-255
 #endif
 
 #ifndef WLED_DISABLE_PARTICLESYSTEM2D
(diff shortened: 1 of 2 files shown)


DedeHai:
FYI: PS Vortex runs at an impressive 33FPS at 128x64 pixels.

DedeHai:
1D->2D expansion is not implemented yet.

blazoncek:
> 1D->2D expansion is not implemented yet.

You will need to use regular `setPixelColor()` for that or re-implement mapping (prone to errors in the future).
IMO it would be ok to loose a few FPS as expansion always deals with reduced number of pixels.

DedeHai:
I have already implemented 1D->2D locally using setPixelColor(), crashes all the time though and need to investigate on where I am overrunning a buffer. 
Am I correct to assume that the mapping (and with it vLength) never changes while an FX is in the middle of rendering?

willmmiles:
> Am I correct to assume that the mapping (and with it vLength) never changes while an FX is in the middle of rendering?

I believe that's the design intent, but I think the current implementation might technically permit some races.   The write on [FX_fcn.cpp:1210](https://github.com/wled/WLED/blob/b7bfd6fc67645f1475c5998d5d4fea4cff3ee256/wled00/FX_fcn.cpp#L1210) is not a memory fence; there is no strict requirement that `_isServicing` is committed to RAM before the loop begins manipulating Segments.

Ideally `suspend`/`isServicing` would be replaced by a mutex, though preserving the early exit of the render path on contention would be very handy.  Alternatively, we could architect a solution to ensure that state-altering operations are always performed on the main task -- it'd be a little more work to get started, but might be a more practical solution globally.  (Technically, today, even usermods might have to contend with their `readFromJsonState()` being called simultaneously with `loop()`).

DedeHai:
> I believe that's the design intent, but I think the current implementation might technically permit some races. The write on [FX_fcn.cpp:1210](https://github.com/wled/WLED/blob/b7bfd6fc67645f1475c5998d5d4fea4cff3ee256/wled00/FX_fcn.cpp#L1210) is not a memory fence; there is no strict requirement that `_isServicing` is committed to RAM before the loop begins manipulating Segments.

a race condition could explain the erratic crash dumps I see in https://github.com/wled/WLED/issues/4779

blazoncek:
> The write on [FX_fcn.cpp:1210](https://github.com/wled/WLED/blob/b7bfd6fc67645f1475c5998d5d4fea4cff3ee256/wled00/FX_fcn.cpp#L1210) is not a memory fence

The proper place to check for is `_suspend` as it will break the servicing loop if set.

> Ideally `suspend`/`isServicing` would be replaced by a mutex,

Please do. That's beyond my knowledge. Would adding `volatile` keyword help?  I would use `_suspend`.

Check if you can see debug output specified [here](https://github.com/wled/WLED/blob/b7bfd6fc67645f1475c5998d5d4fea4cff3ee256/wled00/FX_fcn.cpp#L1683). As `waitForIt()` should wait until strip is no longer servicing but is forgiving and will continue if more than 1 frame has passed.

DedeHai:
FPS comparison shows a 15% increase in speed overall

Issue #4779, opened 2025-07-17
1D->2D expansions can lead to crashes


DedeHai:
### What happened?

While debugging the PS code I found it crashes quite often when using 1D FX in a 2D segment. 
My investigation showed, that this also happens with non PS effects. The larger the segment, the easier it is to trigger the crash. 

### To Reproduce Bug

I was able to reproduce this on S3 and ESP32:
set up a 2D matrix, 32x32 or larger, on 64x32 it happens more often. I used two outputs in my tests both WS281x with equal amounts of LEDs (512 for example).
I set a bootup preset with Aurora effect, then switch to Android effect (just to see if it crashes as I tested with no LEDs connected). Crash dump below was with Aurora.
Change the 1D->2D expansion type randomly, at 32x32 it usually takes 10-20 changes, on larger ones it happenes sometimes after just one or two changes.

### Expected Behavior

-

### Install Method

Self-Compiled

### What version of WLED?

Latest main.

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell
Unfortunately, trace outputs are very erratic and inconclusive. What often pops up is something like this:

JSON buffer released. (12)
-- Stopping transition: S=0x3ffd87bc T(0x3fff15f0) O[0x3fff098c]
-- Destroying segment: 0x3fff098c [0,32:0,32] 352->(0x3fff16fc) T[0]
JSON buffer locked. (11)
Waited for strip to finish servicing.
Waited for strip to finish servicing.
-- Started transition: S=0x3ffd87bc T(0x3fff15f0) O[0x3fff098c] OP[0x3fff1a68]
JSON buffer released. (11)
JSON buffer locked. (11)
Waited for strip to finish servicing.
JSON buffer released. (11)
Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.

Core  1 register dump:
PC      : 0x40082d99  PS      : 0x00050032  A0      : 0x40087b94  A1      : 0x3ffbf870  
=> 0x40082d99: i2s_ll_get_intr_status at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/hal/esp32/include/hal/i2s_ll.h:398
      (inlined by) i2s_intr_handler_default at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/driver/i2s.c:434
[...]

PR #4783, opened 2025-07-19
Bugfixes in FX data allocation

DedeHai:
- realloc was not handled properly and existing data pointer was discarded when `d_realloc()` returned null . This fixes the memory leak
-  tracking of `_usedSegmentData` was also wrong: `Segment::addUsedSegmentData(-_dataLen);` must not be called if `d_malloc()` fails, only for realloc.
- also added *intermediate* fix for waitForIt(), see https://github.com/wled/WLED/issues/4779

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -159,8 +159,19 @@ bool Segment::allocateData(size_t len) {
     return false;
   }
   // prefer DRAM over SPI RAM on ESP32 since it is slow
-  if (data) data = (byte*)d_realloc(data, len);
-  else      data = (byte*)d_malloc(len);
+  if (data) {
+    void* newData = d_realloc(data, len);
+    if (newData)
+      data = (byte*)newData; // realloc returns null if it fails but does not free the original pointer
+    else {
+      d_free(data); // free old data if realloc failed
+      data = nullptr; // reset pointer to null
+      Segment::addUsedSegmentData(-_dataLen); // subtract original buffer size
+      _dataLen = 0; // reset data length
+    }
+  }
+  else data = (byte*)d_malloc(len);
+
   if (data) {
     memset(data, 0, len);  // erase buffer
     Segment::addUsedSegmentData(len - _dataLen);
@@ -170,7 +181,6 @@ bool Segment::allocateData(size_t len) {
   }
   // allocation failed
   DEBUG_PRINTLN(F("!!! Allocation failed. !!!"));
-  Segment::addUsedSegmentData(-_dataLen); // subtract original buffer size
   errorFlag = ERR_NORAM;
   return false;
 }
@@ -1677,7 +1687,7 @@ void WS2812FX::setTransitionMode(bool t) {
 
 // wait until frame is over (service() has finished or time for 1 frame has passed; yield() crashes on 8266)
 void WS2812FX::waitForIt() {
-  unsigned long maxWait = millis() + getFrameTime();
+  unsigned long maxWait = millis() + getFrameTime() + 100; // TODO: this needs a proper fix for timeout!
   while (isServicing() && maxWait > millis()) delay(1);
   #ifdef WLED_DEBUG
   if (millis() >= maxWait) DEBUG_PRINTLN(F("Waited for strip to finish servicing."));


DedeHai:
thanks for reviewing. I found a few more bugs. will add them to this PR.

Edit:
I also updated the realloc code: realloc can lead to fragmentation that hardly clears up. I updated the realloc to only be allowed if it does it in-place. if the returned pointer changes, free all memory and allocate a new block, hopefully filling up gaps. I saw some improvement in fragmentation that way but it may be completely arbitraray.
also, there are more places that use realloc() incorrectly. fixed those too.

DedeHai:
I am testing ESP32 for larger setups with the latest code. 
I see one rather big issue:
if I set it up as 4panels, 64x64pxels in total, the output works but the webu UI does not. Free heap is 58k but largest available block ist only 8k, web UI is stuck loading. 
@willmmiles is this something you already looked into?  IMHO having working LEDs and no UI is bad, it means users have to reset the device just because the LED buffers hog all memory and starve the UI. 
Should the minimum block size be monitored (free heap is kind of irrelevant) and action taken (what?) if its too small?

willmmiles:
IIRC, the web server core doesn't typically do any allocations bigger than a single packet (1500 bytes).  AsyncJSONResponse serializes on a packet-by-packet basis; it doesn't need a contiguous block either, other than the globally allocated JSON buffer.

Do you get HTTP 500s from specific requests?  The web server core monitors free space before accepting each call.

wled00/FX_fcn.cpp
```diff
@@ -159,8 +159,17 @@ bool Segment::allocateData(size_t len) {
     return false;
   }
   // prefer DRAM over SPI RAM on ESP32 since it is slow
-  if (data) data = (byte*)d_realloc(data, len);
-  else      data = (byte*)d_malloc(len);
+  if (data) {
+    void* newData = d_realloc(data, len); // note: realloc returns null if it fails but does not free the original pointer
+    if (!newData || newData != data) // realloc failed or used a new memory block causing fragmentation. free and allocate new block
+      d_free(newData);
+      d_free(data);
```
willmmiles: You don't need to free the old buffer, `realloc()` has already `free()`d it for you.

willmmiles: (unless it returned null)

DedeHai: but only if it does not fail, i.e. if newData is valid. Or ist that incorrect?

DedeHai:
> Do you get HTTP 500s from specific requests? The web server core monitors free space before accepting each call.

I did not investigate what exactly happens. I looked more into why heap is fragmented so much. Sometimes I do get 503, other times its just stuck on loading without error in the browser.

wled00/FX_fcn.cpp
willmmiles: Correct.

But if realloc *did* allocate a new buffer for you, there's no point in releasing it and trying again; malloc could just give you the new buffer back, since there wasn't enough space to extend the old one.

willmmiles:
> > Do you get HTTP 500s from specific requests? The web server core monitors free space before accepting each call.
> 
> I did not investigate what exactly happens. I looked more into why heap is fragmented so much. Sometimes I do get 503, other times its just stuck on loading without error in the browser.

Under extreme heap pressure it will just drop the connection entirely.  It needs to be pretty extreme though (biggest block <2048, free heap < 8192).

wled00/FX_fcn.cpp
DedeHai: you are right, I had an error in my thought process thinking that if a new segment is created and the old data copied it would help fragmentation if deleting all and allocating new but that is exactly what realloc() actually does.

DedeHai:
I have the UI failing at ~8k smallest block and 70k free heap. so must be something else at play.

I am actually working on PS memory management, these are just things I picked up along the way and had a play and some printouts just to make sure its not the PS causing it. It may still be at fault but its pretty hard to say as I am debugging on several fronts...

willmmiles:
> I have the UI failing at ~8k smallest block and 70k free heap. so must be something else at play.

There are a lot of transient allocations in the TCP stack.  The view from the web server might be much tighter than elsewhere in the code.  You can try building with `-D ASYNCWEBSERVER_DEBUG_TRACE` to get it to print out what it thinks is going on.

DedeHai:
This is what I get. 64x32 2D matrix, ESP32, selecting some RAM heavy PS FX:
```
[59876]{71428}MultiMessage 1073677576 - 0 (0/0/3076 - 0/0)
[59922]{74296}BasicMessage 1073701280 - 0 (0/0/16 - 0/0)
[61465]{79756}MultiMessage 1073677452 - 0 (0/0/1931 - 0/0)
[63144]{51236}MultiMessage 1073701280 - 0 (0/0/1926 - 0/0)
[66265]*** Rejecting client 3FFF5700 (52763): 0, 6132/8976
[66267]*** Sent 503 to 3FFF5700 (52763), result 53
[66322]*** Client 3FFF5700 (0) disconnected
[66847]*** Rejecting client 3FFF5F70 (52764): 0, 6132/8856
[66849]*** Sent 503 to 3FFF5F70 (52764), result 53
[66904]*** Client 3FFF5F70 (0) disconnected
[85115]*** Rejecting client 3FFF58CC (52777): 0, 6132/8600
[85117]*** Sent 503 to 3FFF58CC (52777), result 53
[85162]*** Client 3FFF58CC (0) disconnected
[88268]*** Rejecting client 3FFF5F74 (52778): 0, 6132/8376
[88270]*** Sent 503 to 3FFF5F74 (52778), result 53
[88323]*** Client 3FFF5F74 (0) disconnected
[89442]*** Rejecting client 3FFF58CC (52779): 0, 6132/8624
[89444]*** Sent 503 to 3FFF58CC (52779), result 53
[89488]*** Client 3FFF58CC (0) disconnected
[127760]{60340}(3fff58cc) WR created
[127767]{58052}(3fff58cc) WR handler ready /ws
[127767]Queue: 1 entries, 0 running, 1 queued
[127767]{58052}(3fff58cc) WR handler running
[127780]{57356}(3fff58cc) WR added response 3ffff580
[127846]{56268}MultiMessage 1073677576 - 0 (0/0/1924 - 0/0)
[127848]{50764}(3fff58cc) WR destructing
[127849]Removing 3FFF58CC from queue
[127849]Queue: 0 entries, 0 running, 0 queued
[127860]{57884}(3fff58cc) WR destructed
[136632]{58328}MultiMessage 1073677576 - 0 (0/0/1924 - 0/0)
[140525]*** Rejecting client 3FFFF690 (52801): 0, 8180/11608
[140527]*** Sent 503 to 3FFFF690 (52801), result 53
[140528]*** Rejecting client 3FFFF768 (52802): 0, 8180/10704
[140579]*** Client 3FFFF690 (0) disconnected
```
Last FX is running happily, just UI is dead.


Edit:
after some timeout it starts working again:
```
[152988]*** Client 3FFF58CC (0) disconnected
[201604]*** Sent 503 to 3FFF60C0 (52812), result 53
[201605]*** Rejecting client 3FFF561C (52827): 0, 8180/11132
[201649]*** Client 3FFF60C0 (0) disconnected
[305258]*** Client 3FFF561C (0) disconnected
[351877]{60216}(3fff5e70) WR created
[351882]{58956}(3fff5e70) WR handler ready /
[351882]Queue: 1 entries, 0 running, 1 queued
[351882]{58956}(3fff5e70) WR handler running
[351907]{58240}(3fff5e70) WR added response 3fff59a4
[351908](3fff59a4) ack 0
[351911](3fff59a4)AL1436 0
[351912]*** Rejecting client 3FFF58CC (52909): 1, 5620/8720
[351914](3fff59a4) ack 1713
[351915](3fff59a4)AL1436 0
[351970](3fff59a4) ack 1436
```

after changing FX a few times (even non-PS FX) UI breaks once more.

DedeHai:
got the other error again: it just stops responding, largest free block is only 1.2k at this point:

```
[284446]{73184}MultiMessage 1073599392 - 0 (0/0/1939 - 0/0)
[285598]{72920}MultiMessage 1073599988 - 0 (0/0/1938 - 0/0)
[287015]{73152}MultiMessage 1073599424 - 0 (0/0/1939 - 0/0)
Free heap: 75220
Largest free block: 16372 bytes
[288430]{57812}MultiMessage 1073599424 - 0 (0/0/1934 - 0/0)  ---> stopped responding completely here
Free heap: 54264
Largest free block: 1204 bytes
```

after this, largest free block increases to 4k but UI never recovers.

willmmiles:
> [351912]*** Rejecting client 3FFF58CC (52909): 1, 5620/8720

Sorry the output is somewhat inscrutable.  The values are pointer to the AsyncClient object, the remote port number, the number of web requests queued, the largest free RAM block, and the available heap.

So yeah, it's just out of memory. :(

wled00/FX_fcn.cpp
```diff
@@ -449,8 +459,8 @@ void Segment::setGeometry(uint16_t i1, uint16_t i2, uint8_t grp, uint8_t spc, ui
   }
   // re-allocate FX render buffer
   if (length() != oldLength) {
-    if (pixels) pixels = static_cast<uint32_t*>(d_realloc(pixels, sizeof(uint32_t) * length()));
-    else        pixels = static_cast<uint32_t*>(d_malloc(sizeof(uint32_t) * length()));
+    if (pixels) d_free(pixels); //pixels = static_cast<uint32_t*>(d_realloc(pixels, sizeof(uint32_t) * length())); // using realloc can cause additional fragmentation instead of reducing it
```
blazoncek: That's a misleading comment.

DedeHai: it is what I have seen in tests: when using realloc, fragmentation was worse. When using free/malloc it was somewhat better but both approaches do cause fragmentation. The idea behind this is that free + malloc allows it to fill gaps left behind. 
Got any test scenario that is good to test which works better in practice?

DedeHai:
> So yeah, it's just out of memory. :(

memory is a bit of an issue currently. If I set up: 64x64 using 8 Panels, 512 LEDs each and have a bootup preset = solid the UI already does not work anymore. there is "plenty" of free heap (60k) but largest block is too small it seems (around 10k). 
The limit for a 2D setup without PSRAM seems to be around 2000 LEDs to work *somewhat* reliably. 
Question is how to handle it, IMHO a broken UI is inacceptable.

wled00/FX_fcn.cpp
blazoncek: The sole purpose of realloc() functions is to *reduce* fragmentation. free + malloc will only fill gaps if garbage collector did its job. Unfortunately I have no knowledge of MM on ESP platforms.

IIRC @softhack007 was having a lot of issues with memory fragmentation in the past but I no longer remember what was the outcome (except that FX memory is no longer released and it only grows once allocated).

blazoncek:
@DedeHai this was a great find and a poor performance on my side. I should've thought this better.

However I would prefer to fix the issue at its core, like so:
```c++
void *p_realloc(void *ptr, size_t size) {
  int caps1 = MALLOC_CAP_SPIRAM  | MALLOC_CAP_8BIT;
  int caps2 = MALLOC_CAP_DEFAULT | MALLOC_CAP_8BIT;
  void *newbuf = nullptr;
  if (psramSafe) {
    if (heap_caps_get_free_size(caps2) > 3*MIN_HEAP_SIZE && size < 512) std::swap(caps1, caps2);  // use DRAM for small alloactions & when heap is plenty
    newbuf = heap_caps_realloc_prefer(ptr, size, 2, caps1, caps2); // otherwise prefer PSRAM if it exists
    if (newbuf) return newbuf; // realloc successful
    else {
      p_free(ptr); // free old buffer if realloc failed (to keep consumer allocation logic simple)
      return p_malloc(size); // fallback to malloc if realloc failed (buffer will not be copied!!!)
    }
  }
  newbuf = heap_caps_realloc(ptr, size, caps2); // fallback to default realloc
  if (newbuf) return newbuf; // realloc successful
  else {
    p_free(ptr); // free old buffer if realloc failed
    return heap_caps_malloc(size, caps2); // fallback to malloc if realloc failed
  }
}
```
I do agree that this hides the issue of failed reallocations downstream but IMO that is an acceptable tradeoff as the code is simpler.

wled00/FX_fcn.cpp
DedeHai: there is zero garbage collection, its first come, first serve from the bottom up.
so if a block of memory divides a larger block, free -> malloc *should* move it to the bottom, realloc will keep it there if it fits. 
Using realloc *can* prevent fragmentation but it can also make it worse if large buffers are at play. At least that is my understanding and also what I have seen in tests.

blazoncek:
> Question is how to handle it, IMHO a broken UI is inacceptable.

Indeed it is. 😁 Inventing the logic to switch between DRAM and PSRAM is the key. And if the PSRAM is fast enough for LED and effect data.
When I last dealt with memory allocations (about a year and half ago) AWS was eating 80kB on start and needed additional 50kB for runtime. LED, segment and FX data would eat as much.

wled00/FX_fcn.cpp
blazoncek: Then go a step further and simplify p_realloc and d_realloc to use p_free/p_malloc and d_free/d_malloc instead.
It is the cleanest solution.
However ESP8266 will need the same.

DedeHai:
@blazoncek can you explain the intent of this function?
```c++
void *d_malloc(size_t size) {
  int caps1 = MALLOC_CAP_DEFAULT | MALLOC_CAP_8BIT;
  int caps2 = MALLOC_CAP_SPIRAM  | MALLOC_CAP_8BIT;
  if (psramSafe) {
    if (size > MIN_HEAP_SIZE) std::swap(caps1, caps2);  // prefer PSRAM for large alloactions
    return heap_caps_malloc_prefer(size, 2, caps1, caps2); // otherwise prefer DRAM
  }
  return heap_caps_malloc(size, caps1);
}
```
is the idea to put all rendering buffers in PSRAM, even if there is plenty of DRAM available? What am I missing?

willmmiles:
FWIW, deferring the mapping to the last step in the render process - as we now do with blending - should make things more performant when using PSRAM.  PSRAM on ESP32 has a cache in SRAM which is paged in and out: so when you hit a PSRAM address, that page gets swapped in, and then "nearby" accesses are just as fast as any other SRAM until the page needs to be swapped out again.  So if you're doing a lot of linear accesses (ie. start to end, in row order) it's pretty fast; if you're doing a lot of random or non-linear accesses (such as when applying ledmaps, inversions, mirroring, 2d expansion, etc.) it can be really slow paging in and out all the time.  The more of the mapping logic that can be done later, the faster it will go.

blazoncek:
d_ should prefer DRAM while p_ should prefer PSRAM.
But.
If allocations are small p_ will prefer DRAM, if allocation is huge d_ will prefer PSRAM.

Assuming PSRAM is available.

blazoncek:
> PSRAM on ESP32 has a cache in SRAM

Is that true for rev.1 and rev.3 ESPs? @softhack007 had very bad results with PSRAM on classic ESP32 in the past. I also have a test units rev.1 and rev.3 so can compare when time permits.

willmmiles:
> > PSRAM on ESP32 has a cache in SRAM
> 
> Is that true for rev.1 and rev.3 ESPs? @softhack007 had very bad results with PSRAM on classic ESP32 in the past. I also have a test units rev.1 and rev.3 so can compare when time permits.

I'll review it in more detail, but as far as I could tell, it's fundamental to the memory architecture -- there isn't any way for the CPU to directly address PSRAM, it all goes through the MMU and is backed by some SRAM.   MMU bugs in those accesses are entirely possible, though.

willmmiles:
*Especially* MMU and paging issues with dual-core systems -- there seem to be a lot of SMP coherency-related fixes and workarounds in the ESP-IDF source code...

DedeHai:
> d_ should prefer DRAM while p_ should prefer PSRAM. But. If allocations are small p_ will prefer DRAM, if allocation is huge d_ will prefer PSRAM.

huge as in larger than MIN_HEAP_SIZE (which is only 2k)?

blazoncek:
> huge as in larger than MIN_HEAP_SIZE

Whatever the code says. 😄 BTW at the time of writing MIN_HEAP_SIZE was 8k.

DedeHai:
8k min heap makes a lot more sense :) at least for ESP32's.
I still think that if you have 100k of DRAM available, that should be used and only start consider using PSRAM if DRAM starts running low. OR as Will suggested: for buffers that have mostly sequential access, like the "global" `_pixels` buffer.

webUI / webserver issue: I ran some testing on ESP32: as you can see in the logs above, at 8k of heap (actual usable heap) its already very unstable. @willmmiles the reason the webUI does not work right is closely tied to available heap (obviously) but once it breaks its broken for good: I can get it to half-load the UI, throwing some "json abruptly ends" kind of errors, once the heap recovers to over 20k with a 10k contiguous part it still stays broken until reboot. So keeping a minimum heap of about 8k makes sense but something still breaks the UI. 2k min heap is uselsess. Also: on ESP32 `esp.getfreeheap()` will report all heap, including IRAM, that is why in the printout above it shows 70k heap while in reality, only about 15k of that was useable.
I will provide more fixes/improvents in a new PR as those are more "experimental".
It would be good to find out why the UI is becoming unstable and especially why it does not recover from a "low memory" event and/or how to detect that and re-initialize. 

@blazoncek I found one more subtle bug I could not pinpoint: when transitions are enabled but allocation of pixel buffer fails, FX are still executed with `SEGLEN` equal to zero, causing crashes in FX that divide by that: "Android" is one of them.

DedeHai:
@willmmiles I see you changed the MIN_HEAP_SIZE to 2k in your async webserver update. at the same time WLED_REQUEST_HEAP_USAGE is set to 12k. Does that mean you expect any request to not be processed correctly below 12k of heap? If so, the MIN_HEAP_SIZE should be larger than that no?

blazoncek:
> I found one more subtle bug I could not pinpoint

Not a bug of segment logic per-se. `_vLength` will be 0 if segment is not active (i.e. `stop==0`).

If a request for segment's pixel buffer fails (during start of transition), a copy of "old" segment is made, however its `stop` will be 0 making it inactive. Since `service()` doesn't care if segment's copy is inactive, it runs the effect (it does care for base segment).

We can add `if (segO->isActive())` check prior to running old segment's effect. The effect will be that transition will not run correctly.
Another option is to nullify transition (or at least remove old segment from transition similar effect as above) in such case. A bit more complex.

willmmiles:
I'm aware there are still unresolved issues with surviving heap exhaustion in the web server stack.  I haven't gone all the way to the bottom yet - I believe there are some problems even in the wifi drivers themselves where they can hang up connections, leaking packet buffers, if there's an allocation failure on some critical path.  I'd tried to track down as many places as possible and make sure things are handled gracefully, but I haven't managed to get them all yet.  Sorry - still some work to do there. :(

The heap availability checks in AsyncWebServerWLED do use MALLOC_CAPS to check only for memory that `malloc()` or `new` can access - I'd forgotten about this, it might be why you see different numbers elsewhere than the web server reports.

There are actually several layers of heap checks in the web server: 
- an absolute hard minimum (8kb total, 2kb block, on ESP32) where it will just abort and drop the connection.  This particular threshold was arrived at empirically - once it hits that level, there typically isn't enough RAM left for the LwIP stack to serve a 503.  (LwIP in ESP32 always allocates a full packet for any outbound write, so it needs a contiguous block of ~1600 bytes.)  If this happens, the debug trace will give you a "Dropping client" message.
-  the "minimum heap to queue a request" value: if there's at least that much heap, it will put the request on a queue to service later.  If this threshold is hit, you get a "Rejecting client" message.  In WLED this is set by WLED_REQUEST_MIN_HEAP.
- the "minimum heap to serve a request" value; if there's at least that much heap, the server will dequeue another request and attempt to handle it.  (And it will always try to serve at least one request at a time, as long as the hard minimum is met, because otherwise heap exhaustion would leave it entirely dead to the world.)  The idea is that under memory pressure we can often queue a couple of requests to handle later, even if we can't afford the memory to handle them in parallel.  In WLED this threshold is set with WLED_REQUEST_HEAP_USAGE.

DedeHai:
Update:
- the added `realloc_malloc()` functions are wrappers to try a realloc including freeing original pointer with malloc fallback as suggested. Changing the base realloc is not a good idea as it does not behave like realloc() and can lead to future bugs if it is expected to do so.
- changed the d_malloc/realloc functions to prefer DRAM up to the point where the largest contiguous block gets small
- changed MIN_HEAP_SIZE back to 8k
- Also added UI info printout on error 7 (render buffer allocation failed)

from my side this is ready to merge.

blazoncek:
> from my side this is ready to merge

Why not also add a fix for inactive "old segment"? All it needs is add `&& segO->isActive()` in `service()` (L: 1240).

> `SEGLEN` equal to zero, causing crashes in FX that divide by that

It will solve this.

BTW I also noticed (while poking in my fork) that there are no safeguards if `WS2812FX::_pixels` fails to allocate (very unlikely though). There are also several places that would benefit a `deallocateData()` if segment becomes inactive.

DedeHai:
> Why not also add a fix for inactive "old segment"? All it needs is add `&& segO->isActive()` in `service()` (L: 1240).

that should be a seperate PR, I already wrote it down in my notes.


> There are also several places that would benefit a `deallocateData()` if segment becomes inactive.

I also have a fix for that: release FX data if ram runs low. PS can allocate quite a lot of FX memory which should be released but not if there is no shortage of ram.

edit:
working on some ram allocation tricks RN, the API is a mess though...

netmindz:
Very happy to see some optimisations being done for better memory management for WLED.

Not something for this PR, but at some point I think it would be good to a bit of a review of PSRAM usage overall as especially with octal, we could make greater use of it and in general I'm a big fan of providing stability where we can - i.e better to perhaps drop the frame rate than be unresponsive in the UI or reboot rather than avoiding SPRAM due to speed concerns

DedeHai:
> i.e better to perhaps drop the frame rate than be unresponsive in the UI or reboot rather than avoiding SPRAM due to speed concerns

the new alloc functions should take care of that, parameters could be fine-tuned a bit more for specific use-cases but the general approach of @blazoncek d_alloc/p_alloc is sound.
Ran a few tests for RAM availability, as usual espressif docs are a bit vague on that to say the least but I found some secred stashes for C3 and ESP32 we can tap into, S2 and S3 do not have those but they have plenty of SRAM packed.

wled00/util.cpp
```diff
@@ -664,12 +673,20 @@ void *d_realloc(void *ptr, size_t size) {
...
+// realloc with malloc fallback, original buffer is freed if realloc fails but not copied!
+void *d_realloc_malloc(void *ptr, size_t size) {
+  void *newbuf = d_realloc(ptr, size); // try realloc first
+  if (newbuf) return newbuf; // realloc successful
+  d_free(ptr); // free old buffer if realloc failed
+  return d_malloc(size); // fallback to malloc
+}
+
 void *d_calloc(size_t count, size_t size) {
   int caps1 = MALLOC_CAP_DEFAULT | MALLOC_CAP_8BIT;
```
willmmiles, line 691: This isn't new in this commit, but one thing to be aware of is that `MALLOC_CAP_DEFAULT` by itself doesn't mean "use internal memory" -- it means "use the memory pool that malloc() draws from".  This can matter if PSRAM is available to `malloc()`, either because the IDF was built with `CONFIG_SPIRAM_USE_MALLOC`, or if `heap_caps_malloc_extmem_enable()` was called; both allow this to return a PSRAM buffer.  This isn't necessarily wrong per se -- certainly `p_*alloc()` above doesn't care -- but it might be worth thinking about here in the `d_*alloc()` functions.

If we want to insist on using internal memory, use `MALLOC_CAP_INTERNAL`.  If we want to limit it to the pools that `malloc()` can also reach, use `MALLOC_CAP_INTERNAL | MALLOC_CAP_DEFAULT`.  As you've noticed, there are some internal SRAM pools that `malloc()` won't touch, such as the block of shared data/instruction RAM intended for dynamically loadable or self-modifying code; and when SPI RAM is enabled, the [IDF configuration can reserve some internal memory in a separate non-malloc()-accessible pool for DMA buffers](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/kconfig-reference.html#config-spiram-malloc-reserve-internal).  Using only `MALLOC_CAP_INTERNAL` will permit usage of those pools, if we want to.

blazoncek: By all means! I set up the framework, more skilled people should tune it. 😉

DedeHai: > Using only `MALLOC_CAP_INTERNAL` will permit usage of those pools, if we want to.

you beat me by a few hours, I ran tests on exactly this and already prepared a few new function calls that work as intended. Conclusion of my endevours in a nutshell:
S3, S2 and C3 all have IRAM and DRAM as well as RTCRAM enabled to be used as heap. I initially thought they do not but turns out that was a false test. 
ESP32 however does not and can not BUT: as we are using all 32bit colors now, we can tap into IRAM which is 32bit accessible only. I tested that and it works, just need to run some more tests and make sure the buffers NEVER get accessed in any other way. It will give us about 50k of extra internal RAM (until more functions are put in IRAM that is).

willmmiles: > By all means! I set up the framework, more skilled people should tune it. 😉

Now if you *really* want to get fancy, you can play [fun and games with the MMU](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/mm.html) to coalesce fragmented blocks to a region of contiguous virtual address space so they look and feel like a flat buffer ... ;)

willmmiles: > Now if you really want to get fancy, you can play [fun and games with the MMU](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/mm.html) to coalesce fragmented blocks to a region of contiguous virtual address space so they look and feel like a flat buffer ... ;)

Eh, scrub that.  The MMU implementation isn't general enough to be useful for this purpose.  It looks like it's meant to be used for task stacks and flash/SPIRAM caching only, pretty much.

DedeHai:
Any objectons to merge this as-is and then build upon it instead of polishing what was supposed to be a quick bugfix?

we can continue the discussion in a follow-up PR with improvements to allocation handling.

blazoncek:
> > Why not also add a fix for inactive "old segment"? All it needs is add `&& segO->isActive()` in `service()` (L: 1240).
> 
> that should be a seperate PR, I already wrote it down in my notes.

Why? It is a FX allocation bugfix.

DedeHai:
> > > Why not also add a fix for inactive "old segment"? All it needs is add `&& segO->isActive()` in `service()` (L: 1240).
> > 
> > 
> > that should be a seperate PR, I already wrote it down in my notes.
> 
> Why? It is a FX allocation bugfix.

to me it is a different issue, related to segment buffer and has different implications I need to look into and test.

willmmiles:
+1 for merge now.  I've been looking at locking/task safety for #4779 -- I think one of the sub-tasks on the way there will be to make `Segment` work with default move and copy constructors.  That basically translates to wrapping `Segment::name`, `Segment::data`, and `Segment::pixels` in managed-memory structures.  So the sooner we merge these changes, the less merge conflict resolution I'll have to deal with ... :)

blazoncek:
> * also added _intermediate_ fix for waitForIt()

So why then this? It does not belong in this PR then.

DedeHai:
> > * also added _intermediate_ fix for waitForIt()
> 
> So why then this? It does not belong in this PR then.

you are correct, however I already tested that along the way. my comment was less about mixing up PRs and more about getting this one finished.
As mentioned I have some follow-up improvements that need this one merged and its easier to work in increments than having to maintain changes in different branches.

wled00/util.cpp
blazoncek: > If we want to insist on using internal memory, use `MALLOC_CAP_INTERNAL`

Causes bootloop when allocating (and initialising) ledmap on ESP32.
EDIT: ... when paired with MALLOC_CAP_32BIT.

All is well.

PR #4790, opened 2025-07-26
Allow to define default led effect

Tarik2142:
Allows you to declare a default LED (selected after first power on) effect via `#define DEFAULT_LED_EFFECT <effect number>`
Example: `#define DEFAULT_LED_EFFECT 157` will use "Gravcentric" effect on first startup.

## Summary by CodeRabbit

* **New Features**
  * The default LED effect can now be customized through a configuration setting, allowing users to set their preferred initial effect.

### DIFF
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -742,6 +742,7 @@ WLED_GLOBAL bool notifyHue    _INIT(true);                        // send notifi
 #ifndef DEFAULT_LED_EFFECT
 #define DEFAULT_LED_EFFECT 0
 #endif
+static_assert(DEFAULT_LED_EFFECT >= 0 && DEFAULT_LED_EFFECT < MODE_COUNT, "DEFAULT_LED_EFFECT must be between 0 and MODE_COUNT-1");
 WLED_GLOBAL byte effectCurrent _INIT(DEFAULT_LED_EFFECT);
 WLED_GLOBAL byte effectSpeed _INIT(128);
 WLED_GLOBAL byte effectIntensity _INIT(128);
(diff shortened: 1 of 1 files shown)


netmindz:
At some point I think we need to swap to allowing builds with config.json included, as there are so many things you can pre-configure and adding a build flag for each possible value is not really maintainable

Setting values for things like the pins used by certain hardware so that resetting back to factory defaults still works is one thing, setting your preferred effect on power on does not feel in the same category

Can I ask your usecase for why it's essential that a given effect is used after a fresh install or factory reset? The bootup orange is expected behavior and confirms you have the correct color order set

Tarik2142:
For SR testing it is better to have SR effect selected after flashing. This is required for production but I thought it might be useful for normal use

blazoncek:
FYI it is always possible to pre-build file system and upload it to ESP (OTA or via USB).
It can include minimal (non-hardware dependant) `cfg.json` or any other WLED file including presets.

netmindz:
> FYI it is always possible to pre-build file system and upload it to ESP (OTA or via USB). It can include minimal (non-hardware dependant) `cfg.json` or any other WLED file including presets.

Do you have a step by step set of instructions for this?

blazoncek:
https://discord.com/channels/473448917040758787/718943978636050542/1230376307854938154
https://discord.com/channels/473448917040758787/795233398314237972/950710007530389514
https://discord.com/channels/473448917040758787/795233398314237972/950724884177240064

For anyone else: You'll need a Discord account and join WLED server to see those posts.

PR #4791, opened 2025-07-26
Improvements to heap-memory and PSRAM handling

DedeHai:
Besides better use of PSRAM and free heap handling in general there is a huge improvement for ESP32: with these changes it can work up to 64x64 pixels reliably without PSRAM (UI starts failing at half that size currently). Even got it to work up to about 80x64 pixels but UI access becomes unstable

- Segment `allocateData()` uses more elaborate DRAM checking to reduce fragmentation and allow for larger setups to run on low heap
- Segment `allocateData()` does not limit amount if PSRAM is available (put segment  data in PSRAM if DRAM is depleted) -> prefer functionality over speed
- Segment data allocation fails if minimum contiguous block size runs low to keep the UI working
- Segment data allocation frees buffer if it is not small to not hog RAM excessively: if running a PS effect and then switching to something simpler, there is no need to keep a large buffer
- Freeing large buffers in general instead of using realloc helps to defrag RAM as a large buffer can keep blocking contiguous RAM
- Increased `MAX_SEGMENT_DATA` to account for better segment data handling
- Memory allocation functions try to keep enough DRAM for segment data: if PSRAM is available, putting buffers in PSRAM has little impact on FPS (I saw 2% max drop in my tests) but putting effect data in PSRAM has a larger impact
- Added constant `PSRAM_THRESHOLD` to improve PSARM usage: anything smaller is kept in DRAM, anything larger is put in PSRAM depending on how much heap is available. If heap is plenty, no PSRAM is used
- ESP32 makes use of IRAM (no 8bit access) for pixeluffers, freeing about ~35k of heap
- Global pixel buffer `_pixels` is always put in PSRAM if available (except if it fits in ESP32's 32bit IRAM)
- Increase MIN_HEAP_SIZE to reduce risk of breaking UI due to low memory for JSON response (does not fix it entirely but ist a bit more stable)
- Fix to properly get available heap on all platforms: added function `getFreeHeapSize()` and `getContiguousFreeHeap()`
- Bugfix for effects that divide by SEGLEN: don't run FX in service() if segment is not active
- Syntax fix in AR: calloc() uses (numelements, size) as arguments 

the code may need some polishing (add debug outputs) and some fine-tuning of parameters as I did only test for borderline large 2D setups to see how well it performs under memory starvation but tested on all platforms. 
There is a significant improvement if a board has PSRAM: there is virtually no limit to effect data and the heap is kept healthy and it is very stable, even with 10 overlapping segments of 32x32 pixels each.

## Summary by CodeRabbit

* **New Features**
  * Optional PSRAM-backed JSON and unified buffer allocator; enhanced memory diagnostics and runtime heap/contiguous-heap introspection.

* **Bug Fixes**
  * Fixed audio-reactive FFT buffer allocation to prevent undersized allocations.

* **Refactor**
  * Major memory-allocation overhaul: DRAM/PSRAM selection, PSRAM-preferred pixel/segment buffers, safer allocate/free paths and improved error handling to reduce fragmentation.

* **Chores**
  * Platform memory thresholds, per-request heap controls and minor LED memory accounting updated.


wled00/FX.h
```diff
@@ -88,18 +88,17 @@ extern byte realtimeMode;           // used in getMappedPixelIndex()
...
+// heap memory limit for effects data, pixel buffers try to reserve it if PSRAM is available
 #ifdef ESP8266
   #define MAX_NUM_SEGMENTS  16
   /* How much data bytes all segments combined may allocate */
-  #define MAX_SEGMENT_DATA  5120
+  #define MAX_SEGMENT_DATA  (MAX_NUM_SEGMENTS*640) // 10k by default
 #elif defined(CONFIG_IDF_TARGET_ESP32S2)
   #define MAX_NUM_SEGMENTS  20
-  #define MAX_SEGMENT_DATA  (MAX_NUM_SEGMENTS*512)  // 10k by default (S2 is short on free RAM)
+  #define MAX_SEGMENT_DATA  (MAX_NUM_SEGMENTS*1024)  // 20k by default (S2 is short on free RAM)
```
blazoncek: This will not cut it on every build for S2. Mine is one example, having only about 20-30k of free heap after everything is loaded and operational.

wled00/FX.h
```diff
@@ -88,18 +88,17 @@ extern byte realtimeMode;           // used in getMappedPixelIndex()
...
 #define FPS_CALC_SHIFT 7 // bit shift for fixed point math
 
-/* each segment uses 82 bytes of SRAM memory, so if you're application fails because of
-  insufficient memory, decreasing MAX_NUM_SEGMENTS may help */
+// heap memory limit for effects data, pixel buffers try to reserve it if PSRAM is available
 #ifdef ESP8266
   #define MAX_NUM_SEGMENTS  16
   /* How much data bytes all segments combined may allocate */
-  #define MAX_SEGMENT_DATA  5120
+  #define MAX_SEGMENT_DATA  (MAX_NUM_SEGMENTS*640) // 10k by default
```
blazoncek: That's a lot (and I mean **a lot**) for ESP8266.

blazoncek (commented):
IMO you overestimated the free ram on S2 and ESP8266.

wled00/FX.h
DedeHai: its no longer used in the same way: it now tries to keep this much ram free for FX data if there is PSRAM, otherwise it will limit the use to this amount but if heap runs low, it will respect minheap. so even if you set this to 100k nothing bad happens. one exception (which is no different in current implementation): if a usermod or something else uses heap to deplete it.

wled00/FX.h
DedeHai: if you can give me a scenario that is critical in ram usage, I can test.

blazoncek: 4 usermods (PIR, temperature, 4LD, AR) and 16x16 matrix

netmindz:
@troyhacks  - You might be interested in this

wled00/FX.h
DedeHai: works fine:
Free heap: 16384, Contiguous: 11344

Main:
Free heap: 15720, Contiguous: 7272

troyhacks:
> @troyhacks  - You might be interested in this

Absolutely - I've been doing this sort of thing in all my personal WLED-MM builds for the last year and agree with the findings. 😁 

PSRAM is the solution for more pixels and the speed decrease is negligible - especially compared to it not working. 🤣

@DedeHai has done it more in-depth than my liberal sprinkling of `heap_caps_malloc_prefer()` (and variants) but I'm sure a more measured approach helps things too.

wled00/FX.h
blazoncek: Did you enable MQTT and use GPIO3?

blazoncek:
I've sifted through code and can say that these are solid changes. However ...

Please use space after the keywords. I.e. `if (condition) {...` 😄

wled00/FX.h
DedeHai: no, I dont have the hardware to do that, you'd need to check with your setup.
edit:
since this is all about free heap : if it works in main, I see no reason it should not work with this PR

blazoncek:
Using my development ESP I get the following (excerpt):
```
---WLED 0.15.2-b1 2507270 INIT---

esp32 v3.3.5-1-g85c43024c
arduino-esp32 v1.0.x

CPU:   ESP32-D0WDQ6 rev.1, 2 core(s), 240 MHz.
FLASH: 4 MB, Mode 2 (DIO), speed 40 MHz.
heap 250892
JSON buffer allocated: 32767
Registering usermods ...
heap 217500
Reading config
Reading settings from /wsec.json...
Heap before buses: 215000
Starting usermod config.
heap 214436
Initializing strip
Mixed digital bus types detected! Forcing single I2S output.
Maximum LEDs on a bus: 200
Digital buses: 3
Heap after buses: 208908
Loading custom palettes
Loading custom ledmaps
strip buffer 0x400934a4 size: 1772B
Heap after strip init: 207940B
Playlist unloaded.
Request to apply preset: 101
heap 207564
Usermods setup
heap 193892
Starting WiFi
heap 186320
initServer
heap 134180
initIR
heap 133804
setup() done!
```

When UI comes to life there is about 80k heap free.

```
WiFi-E: Connected! @ 11s
heap 180264
Init STA interfaces
ESP-NOW stopping on connect.
mDNS started
ESP-NOW initing in STA mode.
ESP-NOW  inited in STA mode (channel: 11/1).
heap 135872
Reconnecting MQTT
MQTT ready
Publish MQTT
send NTP
NTP recv, l=48
* sunrise (minutes from UTC) = 216
Sunrise: 05:36
* sunset  (minutes from UTC) = -324
Sunset: 20:36
Local time: 21:29
Trigger time: 05:36
Trigger time: 20:36
---DEBUG INFO---
Runtime: 30s
Unix time: 1753644559,867
Free heap/contiguous: 85096/83488
Free 32bit-heap: 48408
```

DedeHai:
heap is reported incorrectly on ESP32 unless you use this PR.
edit: I see the last two lines are new. did you add the code from the PR to your build? A bit confused about the 0.15.2 at the top

blazoncek:
> heap is reported incorrectly on ESP32 unless you use this PR.

That's this PR's code.

ESP8266
```
---WLED 0.15.2-b1 2507270 INIT---

esp8266 @ 160 MHz.
Core: 3.1.2
FLASH: 4 MB
heap 28824
Registering usermods ...
heap 28552
Reading config
Reading settings from /wsec.json...
Reading settings from /cfg.json...
ESP-NOW linked remote: 3c:61:05:14:12:68
Heap before buses: 28112
Starting usermod config.
heap 27952
Initializing strip
Heap after buses: 27032
Loading custom palettes
Loading custom ledmaps
strip buffer 0x3fff5874 size: 1028B
Heap after strip init: 25480B
heap 24288
Usermods setup
heap 24080
Starting WiFi
heap 23464
initServer
heap 20056
initIR
heap 20056
setup() done!

WiFi-E: Connected! @ 6s
heap 18744
Init STA interfaces
ESP-NOW stopping on connect.
mDNS started
ESP-NOW initing in STA mode.
ESP-NOW  inited in STA mode (channel: 11/1).
heap 17008
Reconnecting MQTT
MQTT ready
Publish MQTT
send NTP
NTP recv, l=48
* sunrise (minutes from UTC) = 216
Sunrise: 05:36
* sunset  (minutes from UTC) = -324
Sunset: 20:36
Local time: 21:40
Trigger time: 05:36
Trigger time: 20:36
WS client connected.
heap 13576
heap 11296
Sending WS data to a single client.
---DEBUG INFO---
Runtime: 30s
Unix time: 1753645245,298
Free heap/contiguous: 13544/12544
```

blazoncek:
Yes, pulled to my dev.

DedeHai:
ESP32 looks similar to what I saw in tests, heap should be increased by about 35k minus all ledbuffers
For ESP8266 there is indeed not much left. With min-heap set to 8k the segment buffers+segdata have about 4k which should be enough for most effects.
What I did notice in my tests: ESP8266 will not save if number of LEDs is set above >650, neither in main nor in this PR. Is that intentional (or is the reason known)?

blazoncek:
ESP32 with PSRAM (rev.3) reports 75k/3972k heap/PSRAM. 32x32 matrix, 3 usermods, no MQTT.
Effects run at 42 FPS. Unfortunately no debug ATM.

blazoncek:
> ESP8266 will not save if number of LEDs is set above >650

Debug will tell you what's happening. I would guess output failed to initialise.

wled00/FX.h
DedeHai: maybe the condition to delete and reallocate of the segmentdata should not be `FAIR_DATA_PER_SEG` but something a bit larger, might help with fragmentation on ESP8266

blazoncek:
Just for posterity: My tests with WROVER rev.3 ESP32 (4MB PSRAM) showed that FPS is running at 26 for 64x32 matrix using effect (Soap, Octopus) that uses effect data buffer. With buffers in DRAM FPS is at stable 42.
I do have rev.1 ESP32 with 4MB PSRAM which is pending testing.

DedeHai:
Can you clarify: which buffers were in PSRAM? I also saw a FPS drop if effect-data is in PSRAM but only minor influence if just the pixel-buffers are in PSRAM. Since I have no ESP32 with PSRAM (Only S2, S3) this would be interesting to distinguish.

blazoncek:
Pixel buffers (effect data is always in DRAM).

DedeHai:
thanks for the clarification. This means that for ESP32 maybe the parameters need to be adjusted to keep pixelbuffers in DRAM and use PSRAM more carefully.

blazoncek:
@DedeHai I have to take my words back. It looks like my tests were flawed.
FPS with Soap on 64x32 matrix is only 27 even when using DRAM/IRAM. Indeed I can see no meaningful difference when using DRAM (fastest by about 1 frame), IRAM or PSRAM for pixel buffers.

EDIT: This is for rev.3 ESP32. rev.1 is still pending.

troyhacks:
@blazoncek There's also "things" that I've found during my extensive (and poorly documented) performance testing that were less intuitive.

I'd have to rerun the tests (and actually write them down this time), but in general I've noticed some random combinations of compiler optimizations and other build alternations that caused bigger FPS losses than I'd expected. 

I remain positive on PSRAM because "works" versus "crash" is often down to just having enough of any kind of RAM. I haven't seen major impacts on speed, but also I approve of a lower speed over it crashing.

Sure, 21845 pixels isn't the fastest on an original ESP32 CPU, but it works just fine when you give it enough RAM.

(21845 isn't a random number. That's the actual max limit when you have 16-bit integers and RGB pixels. 😁 My ESP32-P4 build has mostly been modified to break that barrier - but I don't have 22000+ pixels at home so I can only test at the venue where we have an installation of that size in active use, ha.)

blazoncek:
Thanks @troyhacks for the input.
At some point @DedeHai (or possibly @willmmiles ) mentioned that speed may be influenced by PSRAM caching. Which is very likely IMO (probably in the same way as flash is used/cached).

I have been using code from this PR in my own branch (with a few modifications of my own) and only have positive experience. However I only tested set-up of 64x32 on rev.3 ESP32 WROVER (4MB PSRAM) and QSPI S3 (2MB PSRAM). FPS drop was barely noticeable when using PSRAM or IRAM and free heap was improved by the size of buffers which is good.

IMO the minimum acceptable FPS for any effect is about 25-30, lower than that and you can see the lag.

DedeHai:
I will update this PR with some of the insights from @blazoncek but I need to wait on my WROVER to actually test the code.
I also wrote a defrag function that in its x'th iteration now works quite well while still keeping it as simple as possible. I think that should go in a new PR because it can introduce new bugs: the defrag blocks out free heap while it is running which may have some impact on HA requests, it runs usually for shorter than 1ms so chances are low but over time it will happen.

troyhacks:
One of the things we tried with WLED-MM was to allocate the biggest memory spaces we thought we'd need right off the top at boot to avoid fragmentation. 

Then if we ended up needing more we'd reallocate... which could fail and crash - but on boot up it would allocate the new bigger memory spaces on boot.

And you likely already know this, but there's Espressif code for checking "biggest allocation we can do with current fragmentation in a required heap type" which is much more reliable than "free memory" which absolutely lies to you because of fragmentation. 

This code also catches your heap issues and gives you a better sense of what's going on. I found a few spots during debugging when something tries to allocate a ton of RAM for no good reason.

https://github.com/troyhacks/WLED/blob/mdev/wled00%2Fwled00.ino#L14

DedeHai:
allocating a large chunk at bootup will only contribute to fragmentation in current AC implementation with large buffers for each segment. I ran many (many) tests and found that its hard if not impossible to try to control malloc(), that is why I tried defragmenting and that works much better than trying to be very smart about allocations, those are notoriously unpedictable due to wifi stack and other background processes constantly shuffeling heap around.
And yes, if you check the code of this PR, I fixed quite a lot of free-heap stuff.

willmmiles:
I think what's being suggested is moving segment/data/pixel buffer allocation to a separate heap from the main malloc() heap, reserving the maximum amount of space available to use at bootup.   The buffer heap can use a different allocator, and could potentially even do clever things like using "smarter" pointers to support moving buffers around to defragment internally.

In fact, I think it might be possible to leverage the existing IDF multi heap logic by reserving some memory at compile time, then passing it in as new heap blocks via `heap_caps_add_region_with_caps()` and making up a new `#define MALLOC_CAP_WLED (1<<30)` to flag it for our own usage only.   Then we're guaranteed the only allocations (and any fragmentation) in our regions are ours alone to deal with.

DedeHai:
I did not know about `heap_caps_add_region_with_caps()`, that would indeed open new possibilities. Allocating and managing a block "manually" was what I thought about but discarded as too complex. If we can leverage the built in heap management that option is back on the menu. I can try if this works and if it gives us any advantage.

DedeHai:
I played around with `heap_caps_add_region_with_caps()` and I am not sure this is the route we want to take. It can be used to restrict the rest of the system to use WLED reserved memory (and vice versa) but it really does that: if we restrict, we need to make sure the rest of the system gets enough free heap or we do not solve anything and since UM's and future drivers can grab any amount of heap pre-determining the amount in setup() is risky. Doing it dynamically kind of misses the point: the new `d_alloc()` functions already try to keep a minimum of heap free and do so dynamically at every call.
If I reserve 60k of heap in setup, about 25k is available after wifi has connected. If I reserve any less than that, the UI can still stop responding. IMHO it just shifts the problem and can even amplify it. The only thing that would benefit from such a heap-division would be the defrag-algorithm but I had some ideas that may work well even without the heap area restrictions.

willmmiles:
Yup, that's exactly the tradeoff -- any reservation for pixel buffers we do has to guess at how much memory to leave for the rest of the system.  The up side is that if it's correct, we won't accidentally OOM the core system when managing pixel buffers; the down side is that if it's wrong, the system is *always* unstable -- and this is especially true if it's just a little bit wrong.

I don't feel strongly either way on this one, just throwing out all the options before we go to heroic efforts to tame the default allocator.

Another task in the "memory management" corner is to figure out what's going on with the web server usage, both at rest and under load.  Some systems (definitely seen this on S2s) seem to have an unexpectedly large at-rest allocation when the network comes up.  On other systems (baseline ESP32) I've seen dynamic heap usage of over 60kb during a web transaction -- all released once the transaction is completed.  I don't know where it's all going yet.  ESP8266 can produce a web response with <10k of RAM -- what's the issue on the newer platforms?

And of course graceful OOM handling is still a known issue.  I've fixed a couple of cases in the web server since 2023, but I don't think I've got them all, and the "is it safe to handle or at least queue this request" logic is still weak.  Also I have one last OOM-handling patch to upstream to AsyncTCP (pre-allocating error notification event objects, so they can't be dropped and leak connection handles).

DedeHai:
if the "UI stops responding if heap runs low" can be fixed, there is no need to reserve any memory. What I am seeing is that whenever heap runs below ~8k of contiguous free block-size I keep getting errors and it never recovers, even if the OOM event is just for a brief moment.

DedeHai:
quick test on ESP32 shows that reserving heap yields much worse results than not doing so. If reserving just a few kB too little the UI and especially settings do not work at all (settings are not saved or even reverted, UI stops responding ever so often). 
If reserving more, buffer allocations fail all the time. It worked WAY better in previous tests without the reservation. I deem this experiment as failed.

willmmiles:
> if the "UI stops responding if heap runs low" can be fixed, there is no need to reserve any memory. What I am seeing is that whenever heap runs below ~8k of contiguous free block-size I keep getting errors and it never recovers, even if the OOM event is just for a brief moment.

Investigating in more detail is on my to-do list.  (I should probably open some issues for these to track them...)  I don't currently consider it a burning issue -- if it breaks when you allocate too many LEDs/segments, thus far we've been able to get away with saying "just don't do that then".

Truthfully, I'm not 100% sure it's solvable without going to IDF v5 and working with Espressif on it -- last I looked at it, while LwIP handles allocation failure gracefully, but IIRC the WiFi hardware driver tended to get confused and lock up.  (I have a vague recollection when testing web server code that older versions would just immediately `panic()`, so maybe that's an improvement??)

willmmiles:
> quick test on ESP32 shows that reserving heap yields much worse results than not doing so. If reserving just a few kB too little the UI and especially settings do not work at all (settings are not saved or even reverted, UI stops responding ever so often). If reserving more, buffer allocations fail all the time. It worked WAY better in previous tests without the reservation. I deem this experiment as failed.

Good enough for me.  We really could use some way of talking back in configuration error cases.  "Not saving" is usually nothing more than "I couldn't apply that config for some reason, so I didn't."   It was a little frustrating in the RMT-HI development as broken bus code could make the configuration "stuck" and unable to get out of a bad state.

blazoncek:
One possible way to tackle OOM issue is to prevent creation of new segments if available heap drops too low.

That should also include starting transitions if enabled.

This would also be the "cleanest" solution IMO. The only problem I see with this approach is the determination of "heap too low" condition as AsynTCP/AWS can eat up a hefty amount of heap while processing a request. Which is random by its nature.

DedeHai:
@blazoncek do you know how the `psramSafe` came to be and what its exact purpose is? 
I got my WROVER and there it does not do any good, it bootloops anyway if I upload a standard ESP32 build. What was the reason that builds without BOARD_HAS_PSRAM still need to support PSRAM?
Are there controllers that do not use a WROVER and still have PSRAM that require this?

blazoncek:
`psramSafe` is only to be set to true (on ESP32) if build environment is set correctly for PSRAM use. It should be set to true on S2 and S3 as they can always use PSRAM.

DedeHai:
So are you saying its a useless parameter and it should be a #ifdef?

blazoncek:
If you say so. 😄

DedeHai:
Using an ifdef would save some flash. All p_alloc functions can be skipped

wled00/FX_fcn.cpp
```diff
@@ -1177,7 +1184,10 @@ void WS2812FX::finalizeInit() {
     mem += bus.memUsage(Bus::isDigital(bus.type) && !Bus::is2Pin(bus.type) ? digitalCount++ : 0); // includes global buffer
     if (mem <= MAX_LED_MEMORY) {
       if (BusManager::add(bus) == -1) break;
-    } else DEBUG_PRINTF_P(PSTR("Out of LED memory! Bus %d (%d) #%u not created."), (int)bus.type, (int)bus.count, digitalCount);
+    } else {
+      errorFlag = ERR_NORAM_PX; // alert UI
+      DEBUG_PRINTF_P(PSTR("Out of LED memory! Bus %d (%d) #%u not created."), (int)bus.type, (int)bus.count, digitalCount);
+    }
   }
```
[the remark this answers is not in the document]
DedeHai: the allocate_buffer() function already takes care of a fallback if PSRAM is not available. If PSRAM is available, this will not fail as it is done at setup() stage and available PSRAM at that point is vast.

wled00/wled.cpp
```diff
@@ -243,13 +243,37 @@ void WLED::loop()
...
+    // 32-bit DRAM (not byte accessible, only available on ESP32)
+    size_t dram32_free = heap_caps_get_free_size(MALLOC_CAP_32BIT | MALLOC_CAP_INTERNAL) - dram_free; // returns all 32bit DRAM, subtract 8bit DRAM
+    //size_t dram32_largest = heap_caps_get_largest_free_block(MALLOC_CAP_32BIT | MALLOC_CAP_INTERNAL); // returns largest DRAM block -> not useful
+    DEBUG_PRINTF_P(PSTR("DRAM 32-bit:  Free: %7u bytes | Largest block: N/A\n"), dram32_free);
+    #else
+    // Fast RTC Memory (not available on ESP32)
+    size_t rtcram_free = heap_caps_get_free_size(MALLOC_CAP_RTCRAM);
+    size_t rtcram_largest = heap_caps_get_largest_free_block(MALLOC_CAP_RTCRAM);
+    DEBUG_PRINTF_P(PSTR("RTC RAM:      Free: %7u bytes | Largest block: %7u bytes\n"), rtcram_free, rtcram_largest);
+    #endif
```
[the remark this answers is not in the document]
DedeHai: true but since this is a debug output for devs its ok.

DedeHai:
I consider this PR ready to merge.
Here is a summary of the logic behind it:
- allocate_buffer() function tries to optimise allocations such that heap is kept healthy (i.e. a minimum of contiguous heap is ensured). Any memory that is slow (PSRAM / RTCRAM) is de-prioritized in order to keep speed at an optimum. If there is plenty of heap i.e. more than 1.5x min heap + remaining segment data for FX + buffer size a buffer is not put into PSRAM unless enforce is requested. 
- RTCRAM was as fast as DRAM in my tests, still espressif documentation states otherwise. Small buffers (i.e. segment name) are put in RTCRAM regardless to reduce heap fragmentation. Small persistant buffers can hinder larger buffers is what I saw in my defrag tests.
- PSRAM is used if requested only if DRAM is running below a threshold, `_pixels` global buffer is always placed in PSRAM if available.

I tested this code on all platforms to confirm it works as intended. (ESP32, ESP32-WROVER v1.0, S3, S2, C3 and ESP8266)

DedeHai:
Any objections to merge this to get more feedback from users? Otherwise it will just sit here waiting ;)

netmindz:
From my perspective it looks ready for merge, I'm not marking at code review approved as there are others more experience with C than me, but from the comment thread it looks like there is approval even if that hasn't specifically been flagged in GitHub so I'm happy for you to merge

blazoncek:
Well, I do *not approve* changes to MAX_NUM_SEGMENTS and MAX_SEGMENT_DATA. Even with improved heap handling.
However, this should not prevent merging if you are comfortable with the change and you are updating the documentation as well.

DedeHai:
> Well, I do _not approve_ changes to MAX_NUM_SEGMENTS and MAX_SEGMENT_DATA. Even with improved heap handling. However, this should not prevent merging if you are comfortable with the change and you are updating the documentation as well.

If this turns out to be a bad choice, there is nothing preventing us from updating it. Actually, I increased MAX_NUM_SEGMENTS to account for PSRAM, now that I added the BOARD_HAS_PSRAM as a compile time check, I will change that.

wled00/util.cpp
```diff
@@ -629,92 +629,186 @@ int32_t hw_random(int32_t lowerlimit, int32_t upperlimit) {
...
-  void *newbuf = realloc(ptr, size); // try realloc first
-  if (newbuf) return newbuf; // realloc successful
-  free(ptr); // free old buffer if realloc failed
-  return malloc(size); // fallback to malloc
+void *p_realloc_malloc(void *ptr, size_t size) {
+  void *buffer = heap_caps_realloc(ptr, size, MALLOC_CAP_SPIRAM | MALLOC_CAP_8BIT);
+  if (buffer) return buffer; // realloc successful
+  p_free(ptr); // free old buffer if realloc failed
+  return p_malloc(size); // fallback to malloc
 }
```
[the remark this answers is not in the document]
DedeHai: no, its the whole point of the function

PR #4792, opened 2025-07-28
fixed ADA light Serial conditions for USB CDC (some ESP32 C3 / S2 / S3 )

rommo911:
after a while searching, i found out that USB Serial CDC is not working for ADA light. 
it seems there are incompatibility in isPinOK when checking for pins are used in case of pins of native USB. 
this fixes the issue and now the ada light is working fine.

## Release Notes

* **Bug Fixes**
  * Enhanced USB CDC serial communication handling to properly detect and enable serial capabilities when using USB CDC boot mode, improving serial communication reliability across different hardware configurations.

### DIFF
--- a/wled00/wled_serial.cpp
+++ b/wled00/wled_serial.cpp
@@ -69,11 +69,7 @@ void sendBytes(){
 
 void handleSerial()
 {
-  #if ARDUINO_USB_CDC_ON_BOOT
-    if (!Serial) return;// arduino docs: `if (Serial)` indicates whether or not the USB CDC serial connection is open. For all non-USB CDC ports, this will always return true
-  #else
-    if (!(serialCanRX)) return; // check via pin manager if UART pin is NOT allocated by other function.
-  #endif 
+  if (!(serialCanRX && Serial)) return; // arduino docs: `if (Serial)` indicates whether or not the USB CDC serial connection is open. For all non-USB CDC ports, this will always return true
 
   static auto state = AdaState::Header_A;
   static uint16_t count = 0;
--- a/wled00/wled_serial.cpp
+++ b/wled00/wled_serial.cpp
@@ -69,9 +69,11 @@ void sendBytes(){
 
 void handleSerial()
 {
-  if (serialCanRX == false && Serial) {
-    return;
-  }; // arduino docs: `if (Serial)` indicates whether or not the USB CDC serial connection is open. For all non-USB CDC ports, this will always return true
+  #if ARDUINO_USB_CDC_ON_BOOT
+    if (!Serial) return;// arduino docs: `if (Serial)` indicates whether or not the USB CDC serial connection is open. For all non-USB CDC ports, this will always return true
+  #else
+    if (!(serialCanRX)) return; // check via pin manager if UART pin is NOT allocated by other function.
+  #endif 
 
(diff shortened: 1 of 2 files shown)


wled00/wled_serial.cpp
```diff
@@ -69,7 +69,9 @@ void sendBytes(){
 
 void handleSerial()
 {
-  if (!(serialCanRX && Serial)) return; // arduino docs: `if (Serial)` indicates whether or not the USB CDC serial connection is open. For all non-USB CDC ports, this will always return true
+  if (serialCanRX == false && Serial) {
```
softhack007: Hi @rommo911, could it be that your proposed change does not lead to the same behaviour when not having USB-CDC ?

old condition truth table:
|canRX  | Serial | result |
|------ | ------- | ------ |
| false |  -  | true = return|
| true |  false  | true = return|
| true |  true  | false = continue|

new condition truth table:
|canRX  | Serial | result |
|------ | ------- | ------ |
| false |  true  | true = return|
| false |  false  | false = continue|
| true |  -  | false = continue|


As the new condition you propose seems to be broken - how about just removing the "early exit" for the case of USB_CDC ?

````C++
#if ARDUINO_USB_CDC_ON_BOOT
   if (!Serial) return;
#else
   if (!(serialCanRX && Serial)) return;
#endif
````

@blazoncek not sure if ``serialCanRX`` is used in other places in the code, so we still need the change in wled.cpp?

**Review by softhack007 — CHANGES_REQUESTED** [review]
the new condition in wled_serial.cpp does not preserve the same behaviour when not using USB-CDC. 
Please correct.

**Comment by softhack007** [comment]
> Additionally, the logic for handling serial input is simplified by modifying the early return condition in the serial handler to focus on RX pin availability when Serial is active.

@coderabbitai it seems that rabbits are not good at logic ;-) 
You missed that the new "simplified" condition in wled_serial.cpp does _not_ preserve previous behaviour in case that ARDUINO_USB_CDC_ON_BOOT is not defined. So this PR has unwanted side-effects for other configurations.

File: wled00/wled_serial.cpp (hunk 1 above)
Comment (blazoncek) [reply wled00/wled_serial.cpp]: serialCanRX is only used to determine if ADA/Improv can be used. Nothing else AFAIK.

**Comment by blazoncek** [comment]
I do not know if forcing serialCanRX or serialCanTX to true is ok. It does not seem so to me.

File: wled00/wled_serial.cpp [hunk 2]
```diff
@@ -69,9 +69,11 @@ void sendBytes(){
 
 void handleSerial()
 {
-  if (serialCanRX == false && Serial) {
-    return;
-  }; // arduino docs: `if (Serial)` indicates whether or not the USB CDC serial connection is open. For all non-USB CDC ports, this will always return true
+  #if ARDUINO_USB_CDC_ON_BOOT
+    if (!Serial) return;// arduino docs: `if (Serial)` indicates whether or not the USB CDC serial connection is open. For all non-USB CDC ports, this will always return true
+  #else
+    if (!(serialCanRX)) return; // check via pin manager if UART pin is NOT allocated by other function.
```
softhack007: to get the previous behaviour with "not USB-CDC": either use the original line from WLED source code, or change line 75 to
````C++
     if (!serialCanRX || !Serial) return;
````

Issue #4794, opened 2025-07-29
dark color mixing regression?


TripleWhy:
### What happened?

I noticed that recent versions of the main branch produce artifacts when using dim brightness levels. I traced it to the merge of #4658.

Here is what 4a3af814bfb9fec195ab35b676db01fb0d5d87b4 (before merge) looks like with the palette effect at global brightness 4:

![Image](https://github.com/user-attachments/assets/56d03ef9-12a4-4c63-af5d-07a37917db5e)

<img width="995" height="995" alt="Image" src="https://github.com/user-attachments/assets/ce1c7a5b-b8c3-4b7b-a276-9b7bf4a88e47" />

And this is 23a51e0982a5aaab02371246fc38e11863dcde5d (merge commit) with the same preset:

![Image](https://github.com/user-attachments/assets/a7c0c22a-19ca-479f-b498-751c36413d3f)

<img width="995" height="995" alt="Image" src="https://github.com/user-attachments/assets/b5d1860e-c774-4cf7-b000-7bf6b6e89984" />

For completeness, I also added screenshots of the UI. There it looks the other way round.

### To Reproduce Bug

Use low global brightness.

### Expected Behavior

Some colors produce incorrect colors or off pixels on physical LEDs

### Install Method

Self-Compiled

### What version of WLED?

WLED 0.16.0-alpha (build 2412040)

### Which microcontroller/board are you seeing the problem on?

ESP8266

### Relevant log/trace output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

blazoncek:
Due to new way gamma is handled (which is now correctly applied as a last step).

TripleWhy:
So you are saying this is the correct behavior?
(and everyone who likes colors should simply turn off gamma correction? ^^)

blazoncek:
I can't answer that but gamma is supposed to be applied only as a last step in image manipulation. And now it is. In the past Gamma was applied immediately when color was selected and every calculation worked on gamma modified color.

TripleWhy:
Well I believe it doesn't really matter when you convert between color spaces, but you need to make sure that the math you are using is the correct one that is applicable for each color space. If you simply move the conversion to a different place, but don't replace all math that happens in between, I can see how that would produce different results.

[...]

PR #4798, opened 2025-07-31
proof of concept: use video scaling instead of NPB luminance

DedeHai:
NPB luminance setting dims dark colors to black, this gives really ugly results at low brightness:
Examples:
- palette effect with rainbow is just red, green, blue even with black stripes 
- DNA Spiral: parts of the spiral are missing at brightness < 10

The proposed method applies video scaling: any color that is non zero will stay non zero
NPB luminance is not used anymore.

fixes https://github.com/wled/WLED/issues/4794

### DIFF
--- a/wled00/bus_wrapper.h
+++ b/wled00/bus_wrapper.h
@@ -5,6 +5,8 @@
 //#define NPB_CONF_4STEP_CADENCE
 #include "NeoPixelBusLg.h"
 
+uint8_t globalBrightness = 255; // global brightness for digital busses, set by setBrightness()
+
 //Hardware SPI Pins
 #define P_8266_HS_MOSI 13
 #define P_8266_HS_CLK  14
@@ -775,11 +777,29 @@ class PolyBus {
     return true;
   }
 
+#define R(c) (byte((c) >> 16))
+#define G(c) (byte((c) >> 8))
+#define B(c) (byte(c))
+#define W(c) (byte((c) >> 24))
+
   [[gnu::hot]] static void setPixelColor(void* busPtr, uint8_t busType, uint16_t pix, uint32_t c, uint8_t co, uint16_t wwcw = 0) {
-    uint8_t r = c >> 16;
-    uint8_t g = c >> 8;
-    uint8_t b = c >> 0;
-    uint8_t w = c >> 24;
+
+  // apply global brightness using video scaling: keep a minimum of 1 for each channel if non zero
+  uint32_t addRemains = 0;
+  addRemains  = R(c) ? 0x00010000 : 0;
+  addRemains |= G(c) ? 0x00000100 : 0;
+  addRemains |= B(c) ? 0x00000001 : 0;
+  addRemains |= W(c) ? 0x01000000 : 0;
+
+  const uint32_t TWO_CHANNEL_MASK = 0x00FF00FF;
+  uint32_t rb = (((c & TWO_CHANNEL_MASK) * globalBrightness) >> 8) &  TWO_CHANNEL_MASK; // scale red and blue
+  uint32_t wg = (((c >> 8) & TWO_CHANNEL_MASK) * globalBrightness) & ~TWO_CHANNEL_MASK; // scale white and green
+  c = (rb | wg) + addRemains;
+
+    uint8_t r = R(c);
+    uint8_t g = G(c);
+    uint8_t b = B(c);
+    uint8_t w = W(c);
     RgbwColor col;
     uint8_t cctWW = wwcw & 0xFF, cctCW = (wwcw>>8) & 0xFF;
 
@@ -897,99 +917,7 @@ class PolyBus {
   }
 
   static void setBrightness(void* busPtr, uint8_t busType, uint8_t b) {
-    switch (busType) {
-      case I_NONE: break;
-    #ifdef ESP8266
-      case I_8266_U0_NEO_3: (static_cast<B_8266_U0_NEO_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U1_NEO_3: (static_cast<B_8266_U1_NEO_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_DM_NEO_3: (static_cast<B_8266_DM_NEO_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_BB_NEO_3: (static_cast<B_8266_BB_NEO_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U0_NEO_4: (static_cast<B_8266_U0_NEO_4*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U1_NEO_4: (static_cast<B_8266_U1_NEO_4*>(busPtr))->SetLuminance(b); break;
-      case I_8266_DM_NEO_4: (static_cast<B_8266_DM_NEO_4*>(busPtr))->SetLuminance(b); break;
-      case I_8266_BB_NEO_4: (static_cast<B_8266_BB_NEO_4*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U0_400_3: (static_cast<B_8266_U0_400_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U1_400_3: (static_cast<B_8266_U1_400_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_DM_400_3: (static_cast<B_8266_DM_400_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_BB_400_3: (static_cast<B_8266_BB_400_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U0_TM1_4: (static_cast<B_8266_U0_TM1_4*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U1_TM1_4: (static_cast<B_8266_U1_TM1_4*>(busPtr))->SetLuminance(b); break;
-      case I_8266_DM_TM1_4: (static_cast<B_8266_DM_TM1_4*>(busPtr))->SetLuminance(b); break;
-      case I_8266_BB_TM1_4: (static_cast<B_8266_BB_TM1_4*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U0_TM2_3: (static_cast<B_8266_U0_TM2_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U1_TM2_3: (static_cast<B_8266_U1_TM2_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_DM_TM2_3: (static_cast<B_8266_DM_TM2_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_BB_TM2_3: (static_cast<B_8266_BB_TM2_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U0_UCS_3: (static_cast<B_8266_U0_UCS_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U1_UCS_3: (static_cast<B_8266_U1_UCS_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_DM_UCS_3: (static_cast<B_8266_DM_UCS_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_BB_UCS_3: (static_cast<B_8266_BB_UCS_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U0_UCS_4: (static_cast<B_8266_U0_UCS_4*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U1_UCS_4: (static_cast<B_8266_U1_UCS_4*>(busPtr))->SetLuminance(b); break;
-      case I_8266_DM_UCS_4: (static_cast<B_8266_DM_UCS_4*>(busPtr))->SetLuminance(b); break;
-      case I_8266_BB_UCS_4: (static_cast<B_8266_BB_UCS_4*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U0_APA106_3: (static_cast<B_8266_U0_APA106_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U1_APA106_3: (static_cast<B_8266_U1_APA106_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_DM_APA106_3: (static_cast<B_8266_DM_APA106_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_BB_APA106_3: (static_cast<B_8266_BB_APA106_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U0_FW6_5: (static_cast<B_8266_U0_FW6_5*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U1_FW6_5: (static_cast<B_8266_U1_FW6_5*>(busPtr))->SetLuminance(b); break;
-      case I_8266_DM_FW6_5: (static_cast<B_8266_DM_FW6_5*>(busPtr))->SetLuminance(b); break;
-      case I_8266_BB_FW6_5: (static_cast<B_8266_BB_FW6_5*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U0_2805_5: (static_cast<B_8266_U0_2805_5*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U1_2805_5: (static_cast<B_8266_U1_2805_5*>(busPtr))->SetLuminance(b); break;
-      case I_8266_DM_2805_5: (static_cast<B_8266_DM_2805_5*>(busPtr))->SetLuminance(b); break;
-      case I_8266_BB_2805_5: (static_cast<B_8266_BB_2805_5*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U0_TM1914_3: (static_cast<B_8266_U0_TM1914_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U1_TM1914_3: (static_cast<B_8266_U1_TM1914_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_DM_TM1914_3: (static_cast<B_8266_DM_TM1914_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_BB_TM1914_3: (static_cast<B_8266_BB_TM1914_3*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U0_SM16825_5: (static_cast<B_8266_U0_SM16825_5*>(busPtr))->SetLuminance(b); break;
-      case I_8266_U1_SM16825_5: (static_cast<B_8266_U1_SM16825_5*>(busPtr))->SetLuminance(b); break;
-      case I_8266_DM_SM16825_5: (static_cast<B_8266_DM_SM16825_5*>(busPtr))->SetLuminance(b); break;
-      case I_8266_BB_SM16825_5: (static_cast<B_8266_BB_SM16825_5*>(busPtr))->SetLuminance(b); break;
-    #endif
-    #ifdef ARDUINO_ARCH_ESP32
-      // RMT buses
-      case I_32_RN_NEO_3: (static_cast<B_32_RN_NEO_3*>(busPtr))->SetLuminance(b); break;
-      case I_32_RN_NEO_4: (static_cast<B_32_RN_NEO_4*>(busPtr))->SetLuminance(b); break;
-      case I_32_RN_400_3: (static_cast<B_32_RN_400_3*>(busPtr))->SetLuminance(b); break;
-      case I_32_RN_TM1_4: (static_cast<B_32_RN_TM1_4*>(busPtr))->SetLuminance(b); break;
-      case I_32_RN_TM2_3: (static_cast<B_32_RN_TM2_3*>(busPtr))->SetLuminance(b); break;
-      case I_32_RN_UCS_3: (static_cast<B_32_RN_UCS_3*>(busPtr))->SetLuminance(b); break;
-      case I_32_RN_UCS_4: (static_cast<B_32_RN_UCS_4*>(busPtr))->SetLuminance(b); break;
-      case I_32_RN_APA106_3: (static_cast<B_32_RN_APA106_3*>(busPtr))->SetLuminance(b); break;
-      case I_32_RN_FW6_5: (static_cast<B_32_RN_FW6_5*>(busPtr))->SetLuminance(b); break;
-      case I_32_RN_2805_5: (static_cast<B_32_RN_2805_5*>(busPtr))->SetLuminance(b); break;
-      case I_32_RN_TM1914_3: (static_cast<B_32_RN_TM1914_3*>(busPtr))->SetLuminance(b); break;
-      case I_32_RN_SM16825_5: (static_cast<B_32_RN_SM16825_5*>(busPtr))->SetLuminance(b); break;
-      // I2S1 bus or paralell buses
-      #ifndef CONFIG_IDF_TARGET_ESP32C3
-      case I_32_I2_NEO_3: if (_useParallelI2S) (static_cast<B_32_IP_NEO_3*>(busPtr))->SetLuminance(b); else (static_cast<B_32_I2_NEO_3*>(busPtr))->SetLuminance(b); break;
-      case I_32_I2_NEO_4: if (_useParallelI2S) (static_cast<B_32_IP_NEO_4*>(busPtr))->SetLuminance(b); else (static_cast<B_32_I2_NEO_4*>(busPtr))->SetLuminance(b); break;
-      case I_32_I2_400_3: if (_useParallelI2S) (static_cast<B_32_IP_400_3*>(busPtr))->SetLuminance(b); else (static_cast<B_32_I2_400_3*>(busPtr))->SetLuminance(b); break;
-      case I_32_I2_TM1_4: if (_useParallelI2S) (static_cast<B_32_IP_TM1_4*>(busPtr))->SetLuminance(b); else (static_cast<B_32_I2_TM1_4*>(busPtr))->SetLuminance(b); break;
-      case I_32_I2_TM2_3: if (_useParallelI2S) (static_cast<B_32_IP_TM2_3*>(busPtr))->SetLuminance(b); else (static_cast<B_32_I2_TM2_3*>(busPtr))->SetLuminance(b); break;
-      case I_32_I2_UCS_3: if (_useParallelI2S) (static_cast<B_32_IP_UCS_3*>(busPtr))->SetLuminance(b); else (static_cast<B_32_I2_UCS_3*>(busPtr))->SetLuminance(b); break;
-      case I_32_I2_UCS_4: if (_useParallelI2S) (static_cast<B_32_IP_UCS_4*>(busPtr))->SetLuminance(b); else (static_cast<B_32_I2_UCS_4*>(busPtr))->SetLuminance(b); break;
-      case I_32_I2_APA106_3: if (_useParallelI2S) (static_cast<B_32_IP_APA106_3*>(busPtr))->SetLuminance(b); else (static_cast<B_32_I2_APA106_3*>(busPtr))->SetLuminance(b); break;
-      case I_32_I2_FW6_5: if (_useParallelI2S) (static_cast<B_32_IP_FW6_5*>(busPtr))->SetLuminance(b); else (static_cast<B_32_I2_FW6_5*>(busPtr))->SetLuminance(b); break;
-      case I_32_I2_2805_5: if (_useParallelI2S) (static_cast<B_32_IP_2805_5*>(busPtr))->SetLuminance(b); else (static_cast<B_32_I2_2805_5*>(busPtr))->SetLuminance(b); break;
-      case I_32_I2_TM1914_3: if (_useParallelI2S) (static_cast<B_32_IP_TM1914_3*>(busPtr))->SetLuminance(b); else (static_cast<B_32_I2_TM1914_3*>(busPtr))->SetLuminance(b); break;
-      case I_32_I2_SM16825_5: if (_useParallelI2S) (static_cast<B_32_IP_SM16825_5*>(busPtr))->SetLuminance(b); else (static_cast<B_32_I2_SM16825_5*>(busPtr))->SetLuminance(b); break;
-      #endif
-    #endif
-      case I_HS_DOT_3: (static_cast<B_HS_DOT_3*>(busPtr))->SetLuminance(b); break;
-      case I_SS_DOT_3: (static_cast<B_SS_DOT_3*>(busPtr))->SetLuminance(b); break;
-      case I_HS_LPD_3: (static_cast<B_HS_LPD_3*>(busPtr))->SetLuminance(b); break;
-      case I_SS_LPD_3: (static_cast<B_SS_LPD_3*>(busPtr))->SetLuminance(b); break;
-      case I_HS_LPO_3: (static_cast<B_HS_LPO_3*>(busPtr))->SetLuminance(b); break;
-      case I_SS_LPO_3: (static_cast<B_SS_LPO_3*>(busPtr))->SetLuminance(b); break;
-      case I_HS_WS1_3: (static_cast<B_HS_WS1_3*>(busPtr))->SetLuminance(b); break;
-      case I_SS_WS1_3: (static_cast<B_SS_WS1_3*>(busPtr))->SetLuminance(b); break;
-      case I_HS_P98_3: (static_cast<B_HS_P98_3*>(busPtr))->SetLuminance(b); break;
-      case I_SS_P98_3: (static_cast<B_SS_P98_3*>(busPtr))->SetLuminance(b); break;
-    }
+    globalBrightness = b;
   }
 
   [[gnu::hot]] static uint32_t getPixelColor(void* busPtr, uint8_t busType, uint16_t pix, uint8_t co) {


DedeHai:
@blazoncek I am not sure this is the proper/best way to implement it, feel free to move stuff around.

blazoncek:
It is not. At least not with `NeoPixelBusLg`.

If you intend to replace luminance (and since 0.16 with segment blending you can) use base class `NeoPixelBus` instead. And it should be done in `show()` which outputs segment data to LEDs.

The other way is to move gamma handling into `NeoPixelBusLg` as it is designed with that in mind. I did intend to do that at some point (it is not very good practice to introduce too many changes at once).

DedeHai:
I thought so :)
Also this draft does not preserve color ratios well, some palettes get ugly hue distortion at low brightness. 
How can I test if NPB handles gamma better? That would indeed be the best way.

blazoncek:
Base NPB *does not* handle gamma, nor luminance. It is left for the consumer to do any brightness adjustments on RGB values sent to NPB.

DedeHai:
I am not going to mess with NPB, I want to focus on other things so I will leave that for someone else. I will update this PR with an improved solution that can be used in the meantime.

DedeHai:
@blazoncek I am a bit lost on how to properly implement this as I don't understand all the busses i.e. digitalbus vs polybus and need some guidance: where to set the brightness if not using luminance?

willmmiles:
> @blazoncek I am a bit lost on how to properly implement this as I don't understand all the busses i.e. digitalbus vs polybus and need some guidance: where to set the brightness if not using luminance?

I think the key question is, "do we get any benefits from letting different bus types use custom luminance or gamma calculations"?   Ie. should this be the responsibility of the bus, or should we have a single formula we always use?

IF we want to use a common formula for all buses, the calculation code belongs in [`WS2812FX::show()`](https://github.com/wled/WLED/blob/b76ef231fcb97073285488dff0dadfdcd2ace188/wled00/FX_fcn.cpp#L1657) before we pass the pixel color to the bus -- same as we're doing for gamma correction.   This will give us consistent results across all output types.  If we do this, we can replace `NeoPixelBusLg<>` with plain `NeoPixelBus<>` and maybe save some cycles and code space.   

IF instead we think there's some advantage to letting the bus decide how to apply the brightness calculation, then we should integrate the new calculation (and possibly move gamma correction) directly in to `NeoPixelBusLg`, which exists specifically to apply a luminance and gamma adjustment to pixel colors on the way through to the hardware buffer.

willmmiles:
> I think the key question is, "do we get any benefits from letting different bus types use custom luminance or gamma calculations"? Ie. should this be the responsibility of the bus, or should we have a single formula we always use?

At first glance I'd push in favor of using a common formula.  It might be that there are special cases (network outputs?) where passing a global brightness separately is preferable -- do we know of any?

blazoncek:
@willmmiles is right. IIRC at some point @Makuna considered abandoning NeoPixelBrightnessBus (or whatever it was called) but ultimately replaced it with NeoPixelBusLg.

At the time it was impossible to use NeoPixelBus (non Lg or Brightness) as entire WLED logic relied on NeoPixelBus to do the brightness scaling and WLED also handled gamma in an inappropriate way. Both these issues are resolved and indeed brightness scaling and/or gamma correction can be done in show() (and not in a bus).

We can stick with Lg however and let it do brightness/gamma for digital LEDs. For analog we will need to add gamma adjustment to bus as it already have brightness. For network/virtual bus we'd also need to do the same as it expects gamma to be applied.

willmmiles:
> We can stick with Lg however and let it do brightness/gamma for digital LEDs. For analog we will need to add gamma adjustment to bus as it already have brightness. For network/virtual bus we'd also need to do the same as it expects gamma to be applied.

I did note that `NeoPixelBusLg` does brightness before gamma correction.   Right now we're doing it the other way around (gamma, then brightness).

I'll also add that, using this branch and doing the brightness ourselves for digital busses, switching from `NeoPixelBusLg` to `NeoPixelBus` saves 3.5kb of code space (if that still matters).

To confirm, then, you think we should go the other way around, ie.push the gamma calculation down so each bus type does it?    NeoPixelBusLg bakes its brightness algorithm in fairly deeply -- to adopt the new one here, we can't really use NeoPixelBusLg either way.

One up side of doing the brightness and gamma in the bus object is that we could later support different gamma values for each bus -- could be helpful if you're tweaking mismatched hardware or something.

blazoncek:
> I did note that `NeoPixelBusLg` does brightness before gamma correction.

That would be the correct way.

> switching from `NeoPixelBusLg` to `NeoPixelBus` saves 3.5kb of code space (if that still matters)

I think it does and IMO worth doing (as it will also simplify bus wrapper).

> you think we should go the other way around, ie.push the gamma calculation down so each bus type does it?

Not yet 100% sure but, to answer your next question, we don't need to use Lg, we can implement brightness and gamma in bus manager (`Bus` class and derived classes). It is fairly easy implementation and as you say it allows us to have different brightness scaling and/or gamma for each bus.

DedeHai:
actually applying brightness before gamma would be the correct way:
the whole point of gamma correction is to account for the fact that the light output of LEDs is linear, the perceived brightness is non-linear. 
with the new correct way of applying gamma after all the color manipulations (in linear color space) is the correct way. since we are using a linear function for brightness, that should also be done while still in linear space: you can see in the code how I had to add a crude hue preservation for lowest light values in order to preserve darkest colors.
Flash usage is always an issue, we currently have some to spare thanks to tasmota core but we will reach a point where it is an issue, so moving away from the `NeoPixelBusLg` sounds like a good thing if we do not need any of its features.
doing gamma per bus sounds a bit overkill but would give users a means to fine-tune different types of strips, see [these measurements](https://cpldcpu.com/2022/08/15/does-the-ws2812-have-integrated-gamma-correction/). Mixing strips on a single controller with the requirement to match them in color accuracy is an edge case IMHO. 
What was requested and I do support is a means to adjust individual color channels to correct mismatched production batches of strips.
My suggestion would therefore be to have this:
- apply color channel correction
- apply brightness
- apply gamma
if this produces good results needs to be tested, but at least in theory thats how it should be. In order to achieve this, it would need to be per bus, using individual channel correction but global brightness and global gamma.

this is my understanding of how it *should* be, not claiming this is the best way.

DedeHai:
I ran a test with applying brightness before gamma: it looks awful.

edit: 
I did not dig too deep on NPB gamma correction but using `NeoGammaTableMethod` in `bus_wrapper.h` and passing uncorrected colors does work to apply gamma, tested without the changes of this PR, plain main with only this change.
Unless I did miss something, this is unusable: when I now set brightness < 15 its just all black. Is this supposed to be that way?

blazoncek:
> I ran a test with applying brightness before gamma: it looks awful.

The issue IMO is linear versus logarithmic brightness slider.

> using `NeoGammaTableMethod` in `bus_wrapper.h`

If you haven't noticed, the gamma (functions) rewrite in WLED was with this step in mind. Hence `NeoGammaWLEDMethod` class.
Unfortunately not yet implemented (until consensus is made regarding where to put corrections).

DedeHai:
I did see the comment in the code about `NeoGammaWLEDMethod` and if the `NeoGammaTableMethod` is not fundamentally different, which I assume its not as the output looks very similar, using gamma and brightness from NPBLg is a no-go.
I will update the code of this PR with a new suggestion in show().

DedeHai:
replacing NPB was easier than I thought, did that in the last commit.
@TripleWhy [the last commit](https://github.com/wled/WLED/pull/4798/commits/eb2de8a57fb95684b4469922300717bcd151bbcf) is probably as good as it gets with regards to color preservation at low brightness.

DedeHai:
Ran some speed comparison:
This PR gives me 2% higher FPS in my scenario (ESP C3, 64x32 DDP pixels at 40FPS)

I deem this PR ready to undraft&merge if this is the route we want to go, which I highly recommend for quality reasons, see https://github.com/wled/WLED/issues/4794

@netmindz what is your opinion on this? 
TL;DR: 
- low brightness color accuracy is bad 
- using NPBlg (currently used) allows for gamma correction, but if used, it gets even worse
- Ditching the lg variant of NPB saves some flash (I tested 1.3k on C3, @willmmiles tested >3.5k)
- This PR fixes the issue at hand but the question is, do we want to fix it some other way (i.e. per bus and not globally)

TripleWhy:
As a user, I like the result much better than the current version in main. It even produces some mixed colors at brightness 1, which the old version didn't do.

It doesn't appear to produce even brightness at least until around brightness 30, above it, I can't tell. It looks much more even to me without gamma correction. Even brightness is the point of gamma correction, isn't it?

One minor problem though: Using the power button in the UI doesn't turn off any LEDs.

Did you also want my opinion as a developer? I'd have to get into the gamma topic first.

DedeHai:
thanks for the feedback, the "not turning off LEDs" part is a good find, I did not check that part of code logic.

> It doesn't appear to produce even brightness at least until around brightness 30, above it, I can't tell. It looks much more even to me without gamma correction. Even brightness is the point of gamma correction, isn't it?

I do not know what you mean by that. Gamma is to convert linear color space to perceived color space. if you mean linear brightness increase, then you need to enable "apply gamma to brightness" which should still work the same as before.

TripleWhy:
What I mean is: the point of gamma correction is that the final result looks even to my human eyes. Or so I thought. If not, what is the point and why are we even bothering?

DedeHai:
it has no obligation to look "even" (please try to use more technical terms, that can mean anything).
why we bother:
add two colors colors, say one is B=60, the other one is R=120. dim that down by 50%  you get B=30 R=60. Now convert that to perceived brightness of each channel and the hue changes. this is what gamma correction fixes. Do a side-by side comparison of a palette, say orangery with and without gamma correction at different brightness and you will see it.

wled00/FX_fcn.cpp
```diff
@@ -1654,7 +1654,29 @@ void WS2812FX::show() {
     if (_pixelCCT) { // cctFromRgb already exluded at allocation
       if (i == 0 || _pixelCCT[i-1] != _pixelCCT[i]) BusManager::setSegmentCCT(_pixelCCT[i], correctWB);
     }
-    BusManager::setPixelColor(getMappedPixelIndex(i), realtimeMode && arlsDisableGammaCorrection ? _pixels[i] : gamma32(_pixels[i]));
+
+    uint32_t c = _pixels[i]; // need a copy, do not modify _pixels directly (no byte access allowed on ESP32)
+    if(c > 0) {
```
blazoncek: Spaces...

wled00/FX_fcn.cpp
```diff
@@ -1654,7 +1654,29 @@ void WS2812FX::show() {
...
+      if(!(realtimeMode && arlsDisableGammaCorrection))
+        c = gamma32(c); // apply gamma correction if enabled
+      if(_brightness < 255) {
+        uint8_t r = R(c), g = G(c), b = B(c), w = W(c);
+        // determine dominant channel for hue preservation
+        uint8_t maxc = (r > g) ? ((r > b) ? r : b) : ((g > b) ? g : b);
+        uint8_t halfMax = (maxc - 1) >> 1; // maxc is always > 0
+        uint32_t addRemains = 0;
+        // video scaling: make sure colors do not dim to zero if they started non-zero unless they distort the hue
+        addRemains  = r && r > halfMax ? 0x00010000 : 0;
```
blazoncek: Why don't you reuse or update`color_fade()`?
IMO updating `color_fade()` would be beneficial.

wled00/FX_fcn.cpp
```diff
@@ -1654,7 +1654,29 @@ void WS2812FX::show() {
...
+        addRemains |= g && g > halfMax ? 0x00000100 : 0;
+        addRemains |= b && b > halfMax ? 0x00000001 : 0;
+
+        const uint32_t TWO_CHANNEL_MASK = 0x00FF00FF;
```
blazoncek: Same with `fast_color_scale()`.

wled00/bus_wrapper.h
```diff
@@ -3,7 +3,9 @@
 #define BusWrapper_h
 
 //#define NPB_CONF_4STEP_CADENCE
-#include "NeoPixelBusLg.h"
+#include "NeoPixelBus.h"
+
+uint32_t globalBrightness = 255; // global brightness for digital busses, set by setBrightness()
```
blazoncek: What's the purpose of this? It should not be used in bus wrapper.

wled00/bus_wrapper.h
```diff
@@ -897,99 +899,7 @@ class PolyBus {
   }
 
   static void setBrightness(void* busPtr, uint8_t busType, uint8_t b) {
```
blazoncek: You can remove this function entirely. And from bus manager.

wled00/colors.cpp
```diff
@@ -589,10 +589,13 @@ uint8_t NeoGammaWLEDMethod::gammaT_inv[256];
 void NeoGammaWLEDMethod::calcGammaTable(float gamma)
 {
   float gamma_inv = 1.0f / gamma; // inverse gamma
-  for (size_t i = 0; i < 256; i++) {
-    gammaT[i] = (int)(powf((float)i / 255.0f, gamma) * 255.0f + 0.5f);
-    gammaT_inv[i] = (int)(powf((float)i / 255.0f, gamma_inv) * 255.0f + 0.5f);
+  for (size_t i = 1; i < 256; i++) {
```
blazoncek, line 591: Why 1? You are forcing `gamma[0]` to always be 0 for a few CPU cycles that are irrelevant.
IMO keep start at 0 and remove static assignment.

wled00/colors.cpp
```diff
@@ -589,10 +589,13 @@ uint8_t NeoGammaWLEDMethod::gammaT_inv[256];
...
+    gammaT[i] = (int)(powf((float)i / 255.0f, gamma) * 255.0f + 0.75f); // +0.75 for more aggressive rounding at low values
```
blazoncek: I disagree with this change. Does not line up with commonly accepted gamma calculation.

wled00/FX_fcn.cpp
DedeHai: as always, speed. this is a core function, using color_fade drops FPS by 2%.
but good point on updating that, some FX may benefit: its used in color_from_palette().

wled00/FX_fcn.cpp
DedeHai: same reason, much larger impact.

wled00/bus_wrapper.h
DedeHai: leftover from earlier tests.

wled00/colors.cpp
DedeHai: its always zero.

wled00/colors.cpp
DedeHai: well then commonly accepted is incorrect. this is a much better match with https://en.wikipedia.org/wiki/SRGB#Transformation

DedeHai: actually, it should be +0.9 not +0.75 for an even better match due to rounding errors.

wled00/FX_fcn.cpp
blazoncek: Then declare thos functions inline. You'll get most speed that way.

Or did I miss the subtle difference?

wled00/colors.cpp
blazoncek: Didn't we say "save on code size" if speed does not matter. It certainly does not here.

But yes, [0] is always 0 as is [255] always 255.

wled00/colors.cpp
blazoncek: Then just use ceil() for always rounding up.

wled00/FX_fcn.cpp
DedeHai: if those are inlined, they will bloat code unfortunately. There is no way in C to enforce an inline in just one place AFAIK so what I am basically doing is inlining manually. At this point, there is no real difference anymore to color_fade(). Mabe @willmmiles knows a way to force-inline it just here using some compiler flags?

wled00/colors.cpp
DedeHai: the inverse table would use powf() with a negative base, which is not allowed.

wled00/colors.cpp
DedeHai: no, that is not the same. could test though if it produces better looks, but ceil would make all numbers that are currently 0 into 1 (except at i=0)

DedeHai:
I had an idea:
since many are used to and seem to like the old way of applying brightness and 0.16 is a breaking change in that regard: 
adding a checkmark in settings to make "apply gamma after Effects" optional. If not set, gamma is applied to colors and palettes.

TripleWhy:
> why we bother: add two colors colors, say one is B=60, the other one is R=120. dim that down by 50% you get B=30 R=60. Now convert that to perceived brightness of each channel and the hue changes. this is what gamma correction fixes. Do a side-by side comparison of a palette, say orangery with and without gamma correction at different brightness and you will see it.

I see, thanks for the clarification.

> it has no obligation to look "even" (please try to use more technical terms, that can mean anything).

Well I was trying to use simpler terms, because my first attempt was confusing as well. Let's try a third time: With the default palette, the palette effect, and without the rainbow color wheel option, there are colors that look darker than others to me. Most notable left and right of blue: between blue and cyan / blue and magenta.  
I think this effect is a little less pronounced with the rainbow color wheel option enabled. It's really hard to track because I can't quantify it.

You can of course argue that that is the correct behavior, I don't know that. I'm just pointing out that the old version of gamma correction (before moving it around) did not produce that.

Another difference (between old version and this PR, ignoring current main):
In the old version, I could count 12 different hues, and they appeared to more or less look like the hue slider in an hsv color selector.  
In this PR's version, with the same setup, I can count 13 colors, but only 7 hues. The other colors are just darker versions of the colors next to it.  
Again, that could very well be the more correct behavior, after all, the default palette isn't an hsv sweep, just a linear r->g->b transition, just pointing out differences.

wled00/FX_fcn.cpp
willmmiles: I can't say I've ever tried it, but a quick google suggests the technique is:

```cpp
// fcn_declare.h
// Declare the inline version.  If you call this, it will always be inlined.
inline uint32_t color_fade_inline(uint32_t c1, uint8_t amount, bool video) __attribute__((always_inline)) {
   /* Note body must appear here, or it can't be inlined anyways */
}

// Optionally-inlined version.  Compiler can make up its own mind
uint32_t color_fade(uint32_t c1, uint8_t amount, bool video) {
    return color_fade_inline(c1, amount, video);
}

// Avoid inlining version
// Declaration only; body is the same as color_fade() above, but is located in a cpp file.
// Will never be inlined in any other file without LTO (as the body is not available to the compiler to inline)
uint32_t color_fade_noinline(uint32_t c1, uint8_t amount, bool video)
```

wled00/colors.cpp
blazoncek: ceil(a) == round(a+.99999999999999)

At least that's what they told me.

wled00/colors.cpp
blazoncek: then just revert inverse gamma calculation.
you'll always have rounding errors since you are storing result in integer.

DedeHai: no, it will revert incorrectly.

blazoncek: FYI (gamma==2.2)
```
Gamma Table:
[  0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   1   1   1
   1   1   1   1   1   1   1   2   2   2   2   2   2   2   3   3   3   3
   3   4   4   4   4   5   5   5   5   6   6   6   6   7   7   7   8   8
   8   9   9   9  10  10  11  11  11  12  12  13  13  13  14  14  15  15
  16  16  17  17  18  18  19  19  20  20  21  22  22  23  23  24  25  25
  26  26  27  28  28  29  30  30  31  32  33  33  34  35  35  36  37  38
  39  39  40  41  42  43  43  44  45  46  47  48  49  49  50  51  52  53
  54  55  56  57  58  59  60  61  62  63  64  65  66  67  68  69  70  71
  73  74  75  76  77  78  79  81  82  83  84  85  87  88  89  90  91  93
  94  95  97  98  99 100 102 103 105 106 107 109 110 111 113 114 116 117
 119 120 121 123 124 126 127 129 130 132 133 135 137 138 140 141 143 145
 146 148 149 151 153 154 156 158 159 161 163 165 166 168 170 172 173 175
 177 179 181 182 184 186 188 190 192 194 196 197 199 201 203 205 207 209
 211 213 215 217 219 221 223 225 227 229 231 234 236 238 240 242 244 246
 248 251 253 255]
Inverse Gamma Table:
[  0  21  28  34  39  43  46  50  53  56  59  61  64  66  68  70  72  74
  76  78  80  82  84  85  87  89  90  92  93  95  96  98  99 101 102 103
 105 106 107 109 110 111 112 114 115 116 117 118 119 120 122 123 124 125
 126 127 128 129 130 131 132 133 134 135 136 137 138 139 140 141 142 143
 144 144 145 146 147 148 149 150 151 151 152 153 154 155 156 156 157 158
 159 160 160 161 162 163 164 164 165 166 167 167 168 169 170 170 171 172
 173 173 174 175 175 176 177 178 178 179 180 180 181 182 182 183 184 184
 185 186 186 187 188 188 189 190 190 191 192 192 193 194 194 195 195 196
 197 197 198 199 199 200 200 201 202 202 203 203 204 205 205 206 206 207
 207 208 209 209 210 210 211 212 212 213 213 214 214 215 215 216 217 217
 218 218 219 219 220 220 221 221 222 223 223 224 224 225 225 226 226 227
 227 228 228 229 229 230 230 231 231 232 232 233 233 234 234 235 235 236
 236 237 237 238 238 239 239 240 240 241 241 242 242 243 243 244 244 245
 245 246 246 247 247 248 248 249 249 249 250 250 251 251 252 252 253 253
 254 254 255 255]
New Inverse Gamma Table:
[  0  15  25  31  36  41  45  48  51  54  57  60  62  65  67  69  71  73
  75  77  79  81  83  85  86  88  90  91  93  94  96  97  99 100 101 103
 104 105 107 108 109 110 112 113 114 115 116 118 119 120 121 122 123 124
 125 126 128 129 130 131 132 133 134 135 136 137 137 138 139 140 141 142
 143 144 145 146 147 148 148 149 150 151 152 153 154 154 155 156 157 158
 158 159 160 161 162 162 163 164 165 165 166 167 168 169 169 170 171 171
 172 173 174 174 175 176 177 177 178 179 179 180 181 181 182 183 183 184
 185 185 186 187 187 188 189 189 190 191 191 192 193 193 194 194 195 196
 196 197 198 198 199 199 200 201 201 202 202 203 204 204 205 205 206 207
 207 208 208 209 210 210 211 211 212 212 213 213 214 215 215 216 216 217
 217 218 218 219 220 220 221 221 222 222 223 223 224 224 225 225 226 227
 227 228 228 229 229 230 230 231 231 232 232 233 233 234 234 235 235 236
 236 237 237 238 238 239 239 240 240 241 241 242 242 243 243 244 244 245
 245 245 246 246 247 247 248 248 249 249 250 250 251 251 252 252 252 253
 253 254 254 255]
Transformed Values (Original, Gamma Applied, Inverse Gamma Applied : New Inverse Gamma Applied):
  0 ->   0 ->   0 :   0
  1 ->   0 ->   0 :   0
  2 ->   0 ->   0 :   0
  3 ->   0 ->   0 :   0
  4 ->   0 ->   0 :   0
  5 ->   0 ->   0 :   0
  6 ->   0 ->   0 :   0
  7 ->   0 ->   0 :   0
  8 ->   0 ->   0 :   0
  9 ->   0 ->   0 :   0
 10 ->   0 ->   0 :   0
 11 ->   0 ->   0 :   0
 12 ->   0 ->   0 :   0
 13 ->   0 ->   0 :   0
 14 ->   0 ->   0 :   0
 15 ->   1 ->  21 :  15
 16 ->   1 ->  21 :  15
 17 ->   1 ->  21 :  15
 18 ->   1 ->  21 :  15
 19 ->   1 ->  21 :  15
 20 ->   1 ->  21 :  15
 21 ->   1 ->  21 :  15
 22 ->   1 ->  21 :  15
 23 ->   1 ->  21 :  15
 24 ->   1 ->  21 :  15
 25 ->   2 ->  28 :  25
 26 ->   2 ->  28 :  25
 27 ->   2 ->  28 :  25
 28 ->   2 ->  28 :  25
 29 ->   2 ->  28 :  25
 30 ->   2 ->  28 :  25
 31 ->   2 ->  28 :  25
 32 ->   3 ->  34 :  31
 33 ->   3 ->  34 :  31
 34 ->   3 ->  34 :  31
 35 ->   3 ->  34 :  31
 36 ->   3 ->  34 :  31
 37 ->   4 ->  39 :  36
 38 ->   4 ->  39 :  36
 39 ->   4 ->  39 :  36
 40 ->   4 ->  39 :  36
 41 ->   5 ->  43 :  41
 42 ->   5 ->  43 :  41
 43 ->   5 ->  43 :  41
 44 ->   5 ->  43 :  41
 45 ->   6 ->  46 :  45
 46 ->   6 ->  46 :  45
 47 ->   6 ->  46 :  45
 48 ->   6 ->  46 :  45
 49 ->   7 ->  50 :  48
 50 ->   7 ->  50 :  48
 51 ->   7 ->  50 :  48
 52 ->   8 ->  53 :  51
 53 ->   8 ->  53 :  51
 54 ->   8 ->  53 :  51
 55 ->   9 ->  56 :  54
 56 ->   9 ->  56 :  54
 57 ->   9 ->  56 :  54
 58 ->  10 ->  59 :  57
 59 ->  10 ->  59 :  57
 60 ->  11 ->  61 :  60
 61 ->  11 ->  61 :  60
 62 ->  11 ->  61 :  60
 63 ->  12 ->  64 :  62
 64 ->  12 ->  64 :  62
 65 ->  13 ->  66 :  65
 66 ->  13 ->  66 :  65
 67 ->  13 ->  66 :  65
 68 ->  14 ->  68 :  67
 69 ->  14 ->  68 :  67
 70 ->  15 ->  70 :  69
 71 ->  15 ->  70 :  69
 72 ->  16 ->  72 :  71
 73 ->  16 ->  72 :  71
 74 ->  17 ->  74 :  73
 75 ->  17 ->  74 :  73
 76 ->  18 ->  76 :  75
 77 ->  18 ->  76 :  75
 78 ->  19 ->  78 :  77
 79 ->  19 ->  78 :  77
 80 ->  20 ->  80 :  79
 81 ->  20 ->  80 :  79
 82 ->  21 ->  82 :  81
 83 ->  22 ->  84 :  83
 84 ->  22 ->  84 :  83
 85 ->  23 ->  85 :  85
 86 ->  23 ->  85 :  85
 87 ->  24 ->  87 :  86
 88 ->  25 ->  89 :  88
 89 ->  25 ->  89 :  88
 90 ->  26 ->  90 :  90
 91 ->  26 ->  90 :  90
 92 ->  27 ->  92 :  91
 93 ->  28 ->  93 :  93
 94 ->  28 ->  93 :  93
 95 ->  29 ->  95 :  94
 96 ->  30 ->  96 :  96
 97 ->  30 ->  96 :  96
 98 ->  31 ->  98 :  97
 99 ->  32 ->  99 :  99
100 ->  33 -> 101 : 100
101 ->  33 -> 101 : 100
102 ->  34 -> 102 : 101
103 ->  35 -> 103 : 103
104 ->  35 -> 103 : 103
105 ->  36 -> 105 : 104
106 ->  37 -> 106 : 105
107 ->  38 -> 107 : 107
108 ->  39 -> 109 : 108
109 ->  39 -> 109 : 108
110 ->  40 -> 110 : 109
111 ->  41 -> 111 : 110
112 ->  42 -> 112 : 112
113 ->  43 -> 114 : 113
114 ->  43 -> 114 : 113
115 ->  44 -> 115 : 114
116 ->  45 -> 116 : 115
117 ->  46 -> 117 : 116
118 ->  47 -> 118 : 118
119 ->  48 -> 119 : 119
120 ->  49 -> 120 : 120
121 ->  49 -> 120 : 120
122 ->  50 -> 122 : 121
123 ->  51 -> 123 : 122
124 ->  52 -> 124 : 123
125 ->  53 -> 125 : 124
126 ->  54 -> 126 : 125
127 ->  55 -> 127 : 126
128 ->  56 -> 128 : 128
... irrelevant
```
Excep for i<32 most of the differences are minor/imperceptible.

DedeHai: I ran these and more sophisticated tests. and yes, its only relevant for the lower numbers, that is where it matters most. and I really don't see why you spend so much time nitpicking at two lines of code that are now more correct than they were before.

wled00/colors.cpp
DedeHai: reverted this in the last commit as the difference was not huge and at least on PS effects it looks more "natural" when pixels fade out sooner if darkened. Did not find a good scenario to test which one looks better so for now, I leave it the way it was.

DedeHai:
What should be done about the ABL? since we now took control over brightness IMHO it does not make sense to apply global brightness and afterwards apply ABL for each bus. My suggestion would be to calculate brightness for each bus in  `WS2812FX::show()` and then either scale each pixel value in `WS2812FX::show()` according to the bus brightness or scale the pixels in the `bus::show()`. The latter is a bit closer to how it currently is and would allow for future fine-tuning or color corrections. Any thoughts on this? Not 100% sure I know how to implement it but I can take a stab at it.

willmmiles:
> What should be done about the ABL? since we now took control over brightness IMHO it does not make sense to apply global brightness and afterwards apply ABL for each bus. 

Agreed.

> My suggestion would be to calculate brightness for each bus in `WS2812FX::show()` and then either scale each pixel value in `WS2812FX::show()` according to the bus brightness or scale the pixels in the `bus::show()`. The latter is a bit closer to how it currently is and would allow for future fine-tuning or color corrections. Any thoughts on this? Not 100% sure I know how to implement it but I can take a stab at it.

It might be better to do the scaling in `WS2812FX::show` with the per-bus brightness value.  I'd encourage adding color correction in `WS2812FX::show()` as well, that way it can apply to any bus type.

DedeHai:
ok, I will take a look at it. If I evaluated correctly it mostly means joining the ABL calculations into one. I am sure its not that easy ;)

DedeHai:
I have taken a more in-depth look at this and while I still think its doable it has wider implications than I though:
- ABL per bus is currently done on bus level, moving that up to strip level has some side-effects
- global ABL calculation is currently not respecting mapping, which is generally not an issue but in mixed setups, it can lead to wrong results
- ABL is not implemented for white, with the solution I am thinking of, this may be possible to implement
- moving all brightness scaling to one place (i.e. `strip.service()` ) means the mapping has to be done at that level as well. This is not trivial but I have some ideas on how to do it based on a chained-map generated from the normal map. 
- What that  would do is: re-arranging the `_pixels[]` buff in-place, calculate brightness for each bus for ABL, setting brightness for each bus then scaling the `_pixels[]` according to bus brightness.

this approach adds quite a lot of complexity and I am not sure we want to go that way. 
The other option is to keep the per-bus ABL scaling as it is now, keep ignoring the white channel in global ABL and do the per-bus scaling in `bus.service()`. It also means keeping the duplicate code for ABL calculation and probably the "restore color lossy" function.

willmmiles:
Thanks for taking the time to do a deep dive.  Sounds pretty clear that we need to do a per-bus pixel processing pass (ABL/brightness, per-bus color correction, whathaveyou) after rendering/mapping has taken place.  Given that different bus types can have different pixel buffer formats, it stands to reason that this would be best implemented inside the bus class.

Re global ABL, we could maybe separate the "estimate" and "apply" phases, eg. instead of `BusDigital::estimateCurrentAndLimitBri`, we could arrange that `BusDigital::getUsedCurrent()` always returns an estimate for the current contents of the pixel buffer at the current bus brightness setting.   Then the global layer can just call in to the bus instead of replicating the calculation code.  For reporting the global current usage after each frame, we could have `Bus::show()` return it, instead of caching the result to pick it up later.

DedeHai:
> Given that different bus types can have different pixel buffer formats, it stands to reason that this would be best implemented inside the bus class.

from a logic standpoint, yes this makes sense.

> Re global ABL, we could maybe separate the "estimate" and "apply" phases, eg. instead of `BusDigital::estimateCurrentAndLimitBri`, we could arrange that `BusDigital::getUsedCurrent()` always returns an estimate for the current contents of the pixel buffer at the current bus brightness setting. Then the global layer can just call in to the bus instead of replicating the calculation code. For reporting the global current usage after each frame, we could have `Bus::show()` return it, instead of caching the result to pick it up later.

I had exactly the same thought process. It sounds easy, the problem is: we currently do not have a buffer that contains each bus, mapping only happens when transferring the `_pixels` buffer to NPB. Hence my assessment above that this would require a new mapping approach (which I sketched out already). I see two options:
- apply gamma globally, calculate global ABL as it is done currently, transfer to NPB, for each bus: retrieve the buffer pixel by pixel, calculate ABL brightness (if global is not used), then scale pixel by pixel (either global brightness or ABL brightness). This is the easy way but not the fastest.
- apply gamma globally, re-map the `_pixels` buffer into bus-order, calculate ABL: can now be done globally OR per bus in one function, scale pixelcolors according to calculated bus brightness (global brightness or ABL), transfer all busses to NPB

willmmiles:
> * apply gamma globally, calculate global ABL as it is done currently, transfer to NPB, for each bus: retrieve the buffer pixel by pixel, calculate ABL brightness (if global is not used), then scale pixel by pixel (either global brightness or ABL brightness). This is the easy way but not the fastest.

Honestly I think this is pretty close to what we should do.  It's basically what we're doing now, and provides a place for doing any other unique-to-each-bus pixel manipulation or post-processing.  I think it'll be faster than re-mapping pixel buffers.  In the end we can do exactly two passes over each bus: one to set the un-corrected color value and estimate power usage, and one to apply any final correction.

I'm thinking something like:
* apply global gamma
* transfer to NPB, summing power usage as we go
* compute global ABL
* for each bus, for each pixel: apply brightness (global or local), color correction, any other post-processing
* tell NPB to send buffer to hardware

DedeHai:
alright, I can prototype that. 

> I think it'll be faster than re-mapping pixel buffers.

not so sure about that, remapping pixel buffer with the correct pre-sorted mapping should be as fast if not faster, its basically a lokup table with minimal checks. Its just more complex to pre-compute that table and harder to hunt bugs.

blazoncek:
> * apply global gamma
> * transfer to NPB, summing power usage as we go
> * compute global ABL
> * for each bus, for each pixel: apply brightness (global or local), color correction, any other post-processing
> * tell NPB to send buffer to hardware

AFAIK I've already implemented as such, except points 1, 2 & 3 are swapped (3,1,2).
Applying gamma 1st will indeed provide better results if ABL kicks in since calculated current will be less than without gamma. Low values will still get clipped though. Perhaps a better approach would be to do a logarithmic reduction of brightness instead of linear for ABL. It should preserve lower values better.

willmmiles:
> > I think it'll be faster than re-mapping pixel buffers.
> 
> not so sure about that, remapping pixel buffer with the correct pre-sorted mapping should be as fast if not faster, its basically a lokup table with minimal checks. Its just more complex to pre-compute that table and harder to hunt bugs.

It adds all the complexity, it doesn't save us a pass over the pixels, and the random access will be absolutely brutal on performance if any of those buffers end up in PSRAM.   One of the nice things about the NPB buffers is that they're almost always in SRAM, since the hardware layer needs to read 'em in ISRs.  Another issue is that if we try to re-use the global pixel buffer, there's no guarantee a bus-ordered version will fit -- multiple buses can output the same pixel.  I think also no guarantee there's a viable copy order that ensures that we don't have to have an arbitrarily deep scratch pad somewhere else.

(To be fair, my gut reaction "lookup tables slow" is definitely based on experience with big CPUs, where lookup tables are often almost as bad as linked lists in terms of cache management -- very hard to predict which memory needs to be accessed when, so performance tanks unless the table AND data fit in cache.  Not so much an issue on these micros main SRAMs, though, but any time PSRAM comes in to play, all the same issues arise.)



> > * apply global gamma
> > * transfer to NPB, summing power usage as we go
> > * compute global ABL
> > * for each bus, for each pixel: apply brightness (global or local), color correction, any other post-processing
> > * tell NPB to send buffer to hardware
> 
> AFAIK I've already implemented as such, except points 1, 2 & 3 are swapped (3,1,2). Applying gamma 1st will indeed provide better results if ABL kicks in since calculated current will be less than without gamma.

Yes, that's correct - the current implementation is pretty close!  All that's really needed is moving the global ABL to where it can share the estimate pass with per-bus ABL.

We *might* get some benefit from inverting the bus mapping and sampling from the global pixel buffer instead of iterating over it -- particularly for sparse setups -- but I think this can be looked at later.  Eg.
```
auto bus_power = std::vector<unsigned>(busses.size());   // FUTURE: small vector optimization - could this fit on the stack?
for(auto bus_index = 0; bus_index < busses.size(); ++bus_index) {
  auto& bus = *busses[bus_index];
  unsigned power = 0;
  for(auto i = 0; i < bus.getLength(); ++i) {
      auto pixel_color = _pixels[getInverseMappedPixelIndex(bus.getStart() + i)];
      pixel_color = calculate_gamma_for(pixel_color);
      power += bus.powerFor(pixel_color);   // Could be merged with bus.setPixelColor if dynamic dispatch is slow
      bus.setPixelColor(i, pixel_color);
  }   
  bus_power[bus_index] = power;
}
// global ABL analysis here
```

blazoncek:
> All that's really needed is moving the global ABL to where it can share the estimate pass with per-bus ABL.

IMO that makes no sense. Strip-level ABL operates on mapped pixels (pixels that may not exist and may be scattered across multiple buses in undetermined order). It will reduce overall brightness to be within global limit. Strip-level ABL already ignores non-existent pixels (i.e. pixels not belonging to a digital bus).

Per-bus ABL, however, operates on physical pixels and retrieves (modified) pixel values from NPB. This may cause a single bus to be limited while the rest may operate at desired brightness (where no ABL is needed).

EDIT: And ABL is either strip-level or per-bus, not both.

DedeHai:
> Strip-level ABL operates on mapped pixels (pixels that may not exist and may be scattered across multiple buses in undetermined order). It will reduce overall brightness to be within global limit. Strip-level ABL already ignores non-existent pixels (i.e. pixels not belonging to a digital bus).

the global ABL will limit too much on sparse mappings, it would be better to first map them to busses and then calculate the total brightness, no matter if global or per bus. This means passing colors to NPB then read them back (any loss here if passed unscaled?), calculate ABL if used, then scale each color in a final pass.
This is however not ideal: when no ABL is used, colors could be scaled before passing to NPB which is faster.
Applying the mapping to `_pixels` is possible, the mapping does not allow for double-use pixels, at least in the current implementation. There is however the random access during re-mapping which can be slow on PSRAM.
The best approach may be: determine if ABL is used at all, use global scaling if not, use per-bus scaling if it is. This would make normal setups stay fast and make ABL "more correct".

DedeHai:
RE: PSRAM-Speed
I ran a simple test on an S3 that compares access to a DRAM buffer to PSRAM. 
Read access is done like this:
``` C++
  for (size_t i = 0; i < count; i++) {
    sum += mem[indices[i]];
  }
```
Write acces like this:
``` C++
  for (size_t i = 0; i < count; i++) {
    mem[indices[i]] = i;
  }
```
The first access to PSRAM in the test is always slower, most likely as that is when PSRAM buffer get chached. 
Result: buffers up to ~32k will be (mostly) in cache and any access is quite fast (may differ on ESP32). How this changes if using multiple buffers in PSRAM I did not test and it all depends on how clever the caching is. 
Partial random access is done in blocks of ~5% buffer size, randomly distributed.

```
Memory access speed test (6144 elements, ~24 KB)
=== Partial Random Access ===
DRAM           : Write    353 us, Read    411 us
PSRAM          : Write    506 us, Read    411 us
PSRAM vs DRAM Write: 143.3%   -> note: this is slower due to initial caching of PSRAM buffer (first access in test sequence)
PSRAM vs DRAM Read : 100.0%

=== Random Access ===
DRAM           : Write    344 us, Read    418 us
PSRAM          : Write    337 us, Read    411 us
PSRAM vs DRAM Write: 98.0%
PSRAM vs DRAM Read : 98.3%

=== Sequential Access ===
DRAM           : Write    343 us, Read    420 us
PSRAM          : Write    345 us, Read    411 us
PSRAM vs DRAM Write: 100.6%
PSRAM vs DRAM Read : 97.9%



Memory access speed test (8192 elements, ~32 KB)
=== Partial Random Access ===
DRAM           : Write    453 us, Read    554 us
PSRAM          : Write    836 us, Read    562 us
PSRAM vs DRAM Write: 184.5% -> note: this is slower due to initial caching of PSRAM buffer (first access in test sequence)
PSRAM vs DRAM Read : 101.4%

=== Random Access ===
DRAM           : Write    460 us, Read    554 us
PSRAM          : Write    534 us, Read    572 us
PSRAM vs DRAM Write: 116.1%
PSRAM vs DRAM Read : 103.2%

=== Sequential Access ===
DRAM           : Write    463 us, Read    547 us
PSRAM          : Write    529 us, Read    575 us
PSRAM vs DRAM Write: 114.3%
PSRAM vs DRAM Read : 105.1%



From 32k buffer size upward, PSRAM starts getting slower due to cache size of about 32k, this hits random access hard

Memory access speed test (10240 elements, ~40 KB)
=== Partial Random Access ===
DRAM           : Write    572 us, Read    691 us
PSRAM          : Write   1747 us, Read   2905 us
PSRAM vs DRAM Write: 305.4%
PSRAM vs DRAM Read : 420.4%

=== Random Access ===
DRAM           : Write    573 us, Read    691 us
PSRAM          : Write   4911 us, Read   3888 us
PSRAM vs DRAM Write: 857.1%
PSRAM vs DRAM Read : 562.7%

=== Sequential Access ===
DRAM           : Write    565 us, Read    699 us
PSRAM          : Write   1414 us, Read   2498 us
PSRAM vs DRAM Write: 250.3%
PSRAM vs DRAM Read : 357.4%



Memory access speed test (16384 elements, ~64 KB)
=== Partial Random Access ===
DRAM           : Write    904 us, Read   1101 us
PSRAM          : Write   4049 us, Read   4189 us
PSRAM vs DRAM Write: 447.9%
PSRAM vs DRAM Read : 380.5%

=== Random Access ===
DRAM           : Write    906 us, Read   1102 us
PSRAM          : Write  18906 us, Read  11410 us
PSRAM vs DRAM Write: 2086.8%
PSRAM vs DRAM Read : 1035.4%

=== Sequential Access ===
DRAM           : Write    904 us, Read   1101 us
PSRAM          : Write   3171 us, Read   3521 us
PSRAM vs DRAM Write: 350.8%
PSRAM vs DRAM Read : 319.8%
```

DedeHai:
Key take-aways from my test:
- buffers up to 7500 LEDs worth (30kB) are still quite fast in PSRAM, some extra time (~0.3ms) is needed to cache such buffers
- buffers exceeding this limit will be 3x slower than DRAM buffers if access is _somewhat_ sequential
- random access on large buffers is painfully slow: 10x-20x slower than DRAM

Assuming reading colors back from the bus is fast enough (which it probably is) doing the mapping in  `_pixels[]` makes little sense. So the approach should be as @willmmiles proposed:

>apply global gamma
transfer to NPB, summing power usage as we go
compute global ABL
for each bus, for each pixel: apply brightness (global or local), color correction, any other post-processing
tell NPB to send buffer to hardware

I will try to implement this and check if there are any unforseen complications.

Arcitec:
Thanks for working on improving the gamma handling. I think your plan in the last message makes sense.

But regarding this earlier question:

> I had an idea:
> since many are used to and seem to like the old way of applying brightness and 0.16 is a breaking change in that regard:
> adding a checkmark in settings to make "apply gamma after Effects" optional. If not set, gamma is applied to colors and palettes.

If I understand correctly, 0.16 applies gamma **before** doing effects/color mixing, whereas 0.15 applied gamma at the **end** instead? The difference will certainly change how things wrap around or clip during effect calculations.

It only makes sense to apply gamma at the *end* as the final step. That way, colors have more headroom during effect calculations, and you ensure that effects happen in the correct, linear space (not a scaled/shifted color space). Gamma is a destructive process that causes clipping and shifting of colors. So let's not modify the colors until the end, at the output stage. That is definitely more correct and will ensure more consistent effect calculations.

So in the plan from the previous message, move gamma as the **final** stage before display on the output device.

DedeHai:
> If I understand correctly, 0.16 applies gamma **before** doing effects/color mixing, whereas 0.15 applied gamma at the **end** instead?

no, the other way around. 0.16 is doing it the more mathematically correct way to produce better hue accuracy

Arcitec:
Ah okay, yeah doing it last is much better. The quoted plan said "apply global gamma" first so I thought that's what it was doing. :)

Well, anyway, regarding the user checkbox, I personally don't think it makes sense to keep the wrong behavior even as an option. Sure, maybe someone was used to a certain effect being slightly different due to clipping in the old code, but ehhhhhhhhhhhhhhhhhhhh, can't they just... get used to looking at the correct effect. ;)

DedeHai:
Finally finished the re-work of the bus brightness handling and the ABL. Please review. 
See [commit comment](https://github.com/wled/WLED/pull/4798/commits/00062b2460ae024e7f7bc89b6295ae076ca7b1a9) for details.

Hope no changes got lost in manual merging of my WIP branch. I reworked the code several times over to arrive at this version which IMHO is the best compromise between speed and code consistency. Global ABL would be faster if done on strip level, only scaling once but it convolutes code and adds computational overhead even if the limit does not apply. @blazoncek took that approach in his dev branch (see table)

I did some testing on the ESP32 C3 but not extensively. Here is a summary of my findings:

Flash usage: only checked for C3 and ESP32
- C3: saves 1.1k
- ESP32: saves 5.7k

Test with 2x 256 LEDs (2 outputs) on C3, 32x16 matrix, palette effect

| Version   | Limiter Disabled | Global Limit (non-limiting / limiting) | Per-Bus Limit (non-limiting / limiting) |
|-----------|------------------|----------------------------------------|------------------------------------------|
| PR        | 76.3 FPS         | 75.6 FPS (-0.7) / 70.0 FPS (-6.3)     | 74.8 FPS (-1.5) / 69.3 FPS (-6.3)        |
| Main      | 75.0 FPS         | 74.1 FPS (-0.9) / 74.0 FPS (-0.9)     | 68.0 FPS (-6.1) / 61.3 FPS (-12.8)       |
| Blaz dev  | 67.0 FPS         | 64.0 FPS (-3.0) / 64.0 FPS (-3.0)     | 63.5 FPS (-3.5) / 60.0 FPS (-7.0)        |

In summary:
- this PR fixes low brightness color handling for much nicer colors
- moved brightness scaling from NPB into WLED code, saves us some flash and gives a slight speed boost
- complete rework of ABL, now also respects white channel and mapping and calculations are more consistent. Also it is done on bus level only, no more ABL code on strip level

wled00/FX.cpp
```diff
@@ -7528,9 +7528,9 @@ uint16_t mode_2Ddistortionwaves() {
       byte valueG = gdistort + ((a2-( ((xoffs - cx1) * (xoffs - cx1) + (yoffs - cy1) * (yoffs - cy1))>>7 ))<<1);
       byte valueB = bdistort + ((a3-( ((xoffs - cx2) * (xoffs - cx2) + (yoffs - cy2) * (yoffs - cy2))>>7 ))<<1);
 
-      valueR = gamma8(cos8_t(valueR));
-      valueG = gamma8(cos8_t(valueG));
-      valueB = gamma8(cos8_t(valueB));
+      valueR = cos8_t(valueR);
+      valueG = cos8_t(valueG);
+      valueB = cos8_t(valueB);
```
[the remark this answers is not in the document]
DedeHai: these corrections are all intentional.

wled00/bus_wrapper.h
```diff
@@ -141,65 +141,65 @@
...
 //APA106
-#define B_8266_U0_APA106_3 NeoPixelBusLg<NeoRbgFeature, NeoEsp8266Uart0Apa106Method, NeoGammaNullMethod> //3 chan, esp8266, gpio1
-#define B_8266_U1_APA106_3 NeoPixelBusLg<NeoRbgFeature, NeoEsp8266Uart1Apa106Method, NeoGammaNullMethod> //3 chan, esp8266, gpio2
-#define B_8266_DM_APA106_3 NeoPixelBusLg<NeoGrbFeature, NeoEsp8266DmaApa106Method, NeoGammaNullMethod>  //3 chan, esp8266, gpio3
-#define B_8266_BB_APA106_3 NeoPixelBusLg<NeoGrbFeature, NeoEsp8266BitBangApa106Method, NeoGammaNullMethod> //3 chan, esp8266, bb (any pin but 16)
+#define B_8266_U0_APA106_3 NeoPixelBus<NeoRbgFeature, NeoEsp8266Uart0Apa106Method> //3 chan, esp8266, gpio1
+#define B_8266_U1_APA106_3 NeoPixelBus<NeoRbgFeature, NeoEsp8266Uart1Apa106Method> //3 chan, esp8266, gpio2
+#define B_8266_DM_APA106_3 NeoPixelBus<NeoGrbFeature, NeoEsp8266DmaApa106Method>  //3 chan, esp8266, gpio3
+#define B_8266_BB_APA106_3 NeoPixelBus<NeoGrbFeature, NeoEsp8266BitBangApa106Method> //3 chan, esp8266, bb (any pin but 16)
 //FW1906 GRBCW
```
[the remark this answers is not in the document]
DedeHai: @blazoncek is the rabbit right? do you know anything about this?

blazoncek:
I do like the idea of BusManager::initializeABL() and the swap of NPB, however I do not like the pre-calculation of sum of all pixels on bus level. In current implementation it assumes a pixel is only ever modified once, which may not be true in the future.

There are a few other erros regarding reverse scaling of PolyBus::getPixelColor() too.

DedeHai:
Thanks for reviewing. 

> In current implementation it assumes a pixel is only ever modified once, which may not be true in the future.

I don't understand what you mean by that. If a pixel is sent out to the bus, it should have its final value. What manipulation at bus level are you thinking of?

> There are a few other erros regarding reverse scaling of PolyBus::getPixelColor() too.

very well possible. any hints?

blazoncek:
> What manipulation at bus level are you thinking of?

Multiple calls to BusDigital::setPixelColor() for the same pixel. Actually 2 strips may (partially) overlap if you choose so in settings (not relevant in this case, but still).

However, I've dug deeper now and do see the benefit of pre-calculating sum of all channels in the setPixelColor() phase.

> any hints?

Check restoreColorLossy() occurrences.

wled00/bus_wrapper.h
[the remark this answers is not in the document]
blazoncek: Regardless if rabbit is correct or not, users have color order override so it really doesn't matter.

DedeHai:
> However, I've dug deeper now and do see the benefit of pre-calculating sum of all channels in the setPixelColor() phase.

it makes it a tad faster :)

> Check restoreColorLossy() occurrences.

In the case I use to read back from NPB there is no need to restore the brightness, the ABL calculates on top of the applied brightness as that is applied before summing.

DedeHai:
any objections to merging this @willmmiles?

willmmiles:
> any objections to merging this @willmmiles?

I haven't done a thorough test, but the code looks good to me.  Go for it.

DedeHai:
thanks for reviewing. I did a lot of tests on the actual color scaling, not so many on the new ABL since I do not have all the hardware but the output power values for WS2812 were withing a reasnable tolerance. If you did not see any flaws in logic, this is good to go live.

PR #4808, opened 2025-08-03
POC implementation for deferred settings

willmmiles:
When a state changing request is received, defer handling that request until a known point on the main loop task.   This mitigates any possible cases of editing memory out from underneath the render task (or, for that matter, any other integration or usermod that's serviced on the main task).

To implement this, the JSON buffer lock is reimplemented as a binary semaphrore, as it needs to pass ownership from one task to another.  This required contexts which can back off and retry to indicate such to the kernel.  There might be better solutions, but this sufficies for demonstrating the overall approach.

This is IMO the quickest path to safety re #4779, though not the most performant.  It soft-guarantees that we don't alter the FX state while rendering, since all those operations are now handled on the one task.  I'd talked about setting up a mutex around WS2812FX in #4779; this approach is simpler to implement now and buys us time to arrange a more granular locking approach in the future.

The biggest drawback is that high FPS systems might observe lag during setting-set operations.  Still that's better than risking crashes, I think!

### DIFF
--- a/wled00/fcn_declare.h
+++ b/wled00/fcn_declare.h
@@ -508,7 +508,7 @@ size_t printSetFormIndex(Print& settingsScript, const char* key, int index);
 size_t printSetClassElementHTML(Print& settingsScript, const char* key, const int index, const char* val);
 void prepareHostname(char* hostname);
 [[gnu::pure]] bool isAsterisksOnly(const char* str, byte maxLen);
-bool requestJSONBufferLock(uint8_t moduleID=255);
+bool requestJSONBufferLock(uint8_t moduleID=255, bool wait = true);
 void releaseJSONBufferLock();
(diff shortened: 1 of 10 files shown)


DedeHai:
FYI, the crashes in 1D->2D expansions I saw may have had another main cause, I think I forgot to mention that in the PR:
https://github.com/DedeHai/WLED/blob/9bd9d321bcd6a298f47db3eb22192a198d4d3588/wled00/FX_fcn.cpp#L760
fixed one bug here, there was an out-of-array buffer access by using the _raw() function.

DedeHai:
I did not yet look at the code but the general approach of not handling settings at any random moment has a lot of benefits. One of them is that it (probably) allows for cleaner heap allocation when FX change. I ran many tests and the main issue with heap fragmentation is that themporary buffers are placed all over and leave holes when freed.

PR #4810, opened 2025-08-03
Make sync send/receive modes configurable at compile-time

Arcitec:
Simplifies device setup by being able to configure the desired Sync button behavior at compile-time.

Usage:

```
  -D SYNC_SEND_DIRECT=1
  -D SYNC_SEND_BUTTON=1
```

## Summary by CodeRabbit

* **Chores**
  * Improved configurability of notification synchronization settings, allowing default behaviors to be customized during compilation. No changes to user-facing features or application logic.

### DIFF
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -727,17 +727,52 @@ WLED_GLOBAL send_notification_t    notifyG  _INIT(0b00001111);
 #define notifyAlexa  notifyG.Alexa
 #define notifyHue    notifyG.Hue
 #else
-WLED_GLOBAL bool receiveNotificationBrightness _INIT(true);       // apply brightness from incoming notifications
-WLED_GLOBAL bool receiveNotificationColor      _INIT(true);       // apply color
-WLED_GLOBAL bool receiveNotificationEffects    _INIT(true);       // apply effects setup
-WLED_GLOBAL bool receiveNotificationPalette    _INIT(true);       // apply palette
-WLED_GLOBAL bool receiveSegmentOptions         _INIT(false);      // apply segment options
-WLED_GLOBAL bool receiveSegmentBounds          _INIT(false);      // apply segment bounds (start, stop, offset)
-WLED_GLOBAL bool receiveDirect _INIT(true);                       // receive UDP/Hyperion realtime
-WLED_GLOBAL bool notifyDirect _INIT(false);                       // send notification if change via UI or HTTP API
-WLED_GLOBAL bool notifyButton _INIT(false);                       // send if updated by button or infrared remote
-WLED_GLOBAL bool notifyAlexa  _INIT(false);                       // send notification if updated via Alexa
-WLED_GLOBAL bool notifyHue    _INIT(true);                        // send notification if Hue light changes
+    #ifndef SYNC_RECV_BRIGHTNESS
+    #define SYNC_RECV_BRIGHTNESS true
+    #endif
+    #ifndef SYNC_RECV_COLOR
+    #define SYNC_RECV_COLOR true
+    #endif
+    #ifndef SYNC_RECV_EFFECTS
+    #define SYNC_RECV_EFFECTS true
+    #endif
+    #ifndef SYNC_RECV_PALETTE
+    #define SYNC_RECV_PALETTE true
+    #endif
+    #ifndef SYNC_RECV_SEGMENT_OPTIONS
+    #define SYNC_RECV_SEGMENT_OPTIONS false
+    #endif
+    #ifndef SYNC_RECV_SEGMENT_BOUNDS
+    #define SYNC_RECV_SEGMENT_BOUNDS false
+    #endif
+    #ifndef SYNC_RECV_DIRECT
+    #define SYNC_RECV_DIRECT true
+    #endif
+    #ifndef SYNC_SEND_DIRECT
+    #define SYNC_SEND_DIRECT false
+    #endif
+    #ifndef SYNC_SEND_BUTTON
+    #define SYNC_SEND_BUTTON false
+    #endif
+    #ifndef SYNC_SEND_ALEXA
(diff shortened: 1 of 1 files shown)


netmindz:
We should not be adding compile flags for configuration, if you want your device to have a certain config after you flash it, then flash a filesystem with suitable config file, we need to reduce the number of build flags, not adding more

Arcitec:
@netmindz The issue with WLED's code is that the "Sync" button in WLED's main UI does **absolutely nothing whatsoever** since WLED 0.15 and newer. Activate "Sync" = Nothing happens. No code runs. Nothing gets activated.

My preferred fix would literally be:

- Disable Philips Hue's sync sending flag (why is that on by default in the code? typo, I bet, or someone decided to sneak a personal setting into the main WLED code).
- Enable the "send direct" and "send button" flags by default, so that all changes on the device are detected.

That's a two-line change and it makes the Sync button work as users expect it: Clicking the Sync button will make their other WLED devices on the network sync with the master's settings. It's also a harmless change since nothing is sent/processed unless the user clicks the main UI's "Sync" button.

What do you think about solving it that way?

Old:

https://github.com/wled/WLED/blob/3f90366aa86151e73424dd7c756f95ecbfaf143d/wled00/wled.h#L737-L740

New:

```cpp
WLED_GLOBAL bool notifyDirect _INIT(true);                       // send notification if change via UI or HTTP API
WLED_GLOBAL bool notifyButton _INIT(true);                       // send if updated by button or infrared remote
WLED_GLOBAL bool notifyAlexa  _INIT(false);                       // send notification if updated via Alexa
WLED_GLOBAL bool notifyHue    _INIT(false);                        // send notification if Hue light changes
```

Arcitec:
Flashing a filesystem has several issues:

- It gets wiped during factory resets. This is the biggest issue.
- Includes all settings, meaning the "first setup" wizard to choose WiFi network etc won't run.

So I'd prefer to fix the Sync button issue in WLED's code instead, either via compile-flags if there's no agreement on what the defaults should be (hence this PR), or alternatively by fixing WLED to have sane defaults for the Sync button. :)

Defaults should be something that matches the vast majority of users and how they are likely to use WLED. With syncing, and pressing the "Sync" button, the majority of users expect that it begins syncing with other WLED devices on the network. That can be fixed by improving the defaults.

And obviously, while I've fixed this issue personally, I want it to work out of the box for everyone who visits wled.me and downloads a binary, because that puts the project in a better light. Having a non-functional "Sync" button by default in the main UI doesn't really make sense.

I'd be happy to make the PR to fix the defaults to make the Sync UI button work. :heart:

netmindz:
I agree that users have found the new behaviour confusing, there have been several reports of if being "broken" but I'm still confused as to why it was changed and what the expected new behaviour is (out of the box)

Arcitec:
@netmindz Yeah. While I was investigating this and talking to people, @blazoncek said that the original pre-0.15 behavior was that clicking on "Sync" in the UI automatically enabled the relevant sync flags internally.

The latest WLED code uses a two-tier system:

Tier 1:

- The UI "Sync" button. Clicking this button toggles the master "send notifications" flags here:

https://github.com/wled/WLED/blob/3f90366aa86151e73424dd7c756f95ecbfaf143d/wled00/wled.h#L662-L663

- If that flag is false (it's always off by default at startup unless the user changes a settin for that), then notifications (syncing) is never sent. It's the master on/off switch. So all the tier 2 settings are slaves under the tier 1 master switch.

Tier 2:

- The settings then chooses what actually gets sent when the master switch is on. Those defaults no longer make any sense (all syncing is off except to send Philips Hue signals, which is a bit "wtf"):

https://github.com/wled/WLED/blob/3f90366aa86151e73424dd7c756f95ecbfaf143d/wled00/wled.h#L737-L740

My proposal is to change those flag values to something that makes sense by default:

- Sync direct + button-induced changes to WLED.
- Do not send notifications to Amazon Alexa and Philips Hue. People who have such extra devices can enable those manually.

```cpp
WLED_GLOBAL bool notifyDirect _INIT(true);                       // send notification if change via UI or HTTP API
WLED_GLOBAL bool notifyButton _INIT(true);                       // send if updated by button or infrared remote
WLED_GLOBAL bool notifyAlexa  _INIT(false);                       // send notification if updated via Alexa
WLED_GLOBAL bool notifyHue    _INIT(false);                        // send notification if Hue light changes
```

With that change, the WLED UI's "Sync" button would make sense to users again. When people's master "Sync" device changes its settings, other WLED devices in the network would then follow it as users expect.

I think it should be improved like this, so that WLED is pre-configured to sync with other WLED devices whenever a user clicks on "Sync".

I have not created a pull request to change the defaults because I wasn't sure if you want to do that, but I honestly think the change above should be made so that WLED is pre-configured to work out of the box. The current situation where people ask "is it broken?" isn't ideal. :)

I can make a quick PR if it helps, or anyone can commit the change directly. I think the improvement should also happen in 0.15.2.

blazoncek:
> I'm still confused as to why it was changed and what the expected new behaviour is

Previous behaviour was tied to "notify on direct change". So if you wanted to "notify" other "stuff" you were always notifying everything (aka sync).

I agree to set "notify on direct change" to true by default as it will no longer also trigger sync behaviour.

Arcitec:
@blazoncek Ahh that makes sense. So in the past, setting them to "true" would have made the devices always sync by default. When things were refactored to have the master on/off switch and granular control, this flag's default was simply forgotten.

I was sick for a few days. I am preparing the pull requests to fix this for 0.16 and 0.15.2 now, since it's a "major" user-facing bug in expected behavior, and it would make user's lives easier if Sync just works. :)

Edit: Pull requests are ready:

- 0.16: https://github.com/wled/WLED/pull/4844
- 0.15.2: https://github.com/wled/WLED/pull/4845

PR #4819, opened 2025-08-06
Add new effects: music strobe and flame jet

aerosh:
Added audio-reactive strobe and flame jet effects.
The strobe effect adjusts the trigger threshold, the fade time, and the palette.
The flame jet effect adjusts the palette, trigger threshold, fade time, and flame height. The effect looks good on cylindrical matrices.

## Summary by CodeRabbit

* **New Features**
  * Added two audio-reactive LED effects: Music Strobe and Flame Jet. Both respond dynamically to audio input and are selectable from the effects list.

* **Improvements**
  * Increased the total number of available effects to include these additions.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -10,7 +10,7 @@
 # ------------------------------------------------------------------------------
 
 # CI/release binaries
-default_envs = d1_mini, nodemcuv2, esp8266_2m, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, nodemcuv2_compat, esp8266_2m_compat, esp01_1m_full_compat, esp32dev, esp32dev_V4, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_16MB_opi, esp32s3dev_8MB_opi, esp32s3_4M_qspi, esp32_wrover, usermods
+default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, nodemcuv2_compat, esp8266_2m_compat, esp01_1m_full_compat, esp32dev, esp32dev_V4, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_16MB_opi, esp32s3dev_8MB_opi, esp32s3_4M_qspi, esp32_wrover, usermods
 
 src_dir  = ./wled00
 data_dir = ./wled00/data
--- a/platformio.ini
+++ b/platformio.ini
@@ -10,7 +10,7 @@
 # ------------------------------------------------------------------------------
 
 # CI/release binaries
-default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, nodemcuv2_compat, esp8266_2m_compat, esp01_1m_full_compat, esp32dev, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_16MB_opi, esp32s3dev_8MB_opi, esp32s3_4M_qspi, esp32_wrover, usermods
+default_envs = d1_mini, nodemcuv2, esp8266_2m, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, nodemcuv2_compat, esp8266_2m_compat, esp01_1m_full_compat, esp32dev, esp32dev_V4, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_16MB_opi, esp32s3dev_8MB_opi, esp32s3_4M_qspi, esp32_wrover, usermods
 
 src_dir  = ./wled00
 data_dir = ./wled00/data
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -10644,6 +10644,100 @@ uint8_t WS2812FX::addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name)
   }
 }
 
+/*
+  Simple bit strob
+*/
+uint16_t mode_music_strob(void) {
+  um_data_t *um_data = getAudioData();
+  if (!um_data) return FRAMETIME; 
+
+  bool samplePeak = *(bool*)um_data->u_data[3]; 
+  int volumeRaw   = *(int16_t*)um_data->u_data[1]; 
+
+  long mappedSpeed = map(SEGMENT.speed, 0, 255, 0, 20);
+  uint8_t fadeBy = 255 - mappedSpeed;
+  SEGMENT.fade_out(fadeBy);
+
+  uint8_t threshold = SEGMENT.intensity;
+
+  if (samplePeak && volumeRaw > threshold) {
+    SEGMENT.fill(SEGCOLOR(0));
+  }
+  
+  return FRAMETIME;
+}
+
+static const char _data_FX_MODE_MUSIC_STROB[] PROGMEM = "Music Strob@Fade speed,Threshold;!,!;!;1v;ix=128,sx=224,si=1";
+
+/*--------------------------*/
+
+/*
+  Flame jet
+*/
+uint16_t mode_flame_jet(void) {
+  if (!SEGMENT.is2D()) return mode_static();
+
+  um_data_t *um_data = getAudioData();
+  if (!um_data) return FRAMETIME;
+
+  bool samplePeak = *(bool*)um_data->u_data[3];
+  int volumeRaw   = *(int16_t*)um_data->u_data[1];
+
+  long mappedFade = map(SEGMENT.intensity, 0, 255, 0, 64);
+  uint8_t fade_by = 255 - mappedFade;
+
+  uint8_t threshold = SEGMENT.custom1;
+
+  const uint16_t max_height = SEG_H;
+
+  SEGMENT.fade_out(fade_by);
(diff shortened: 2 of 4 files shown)


blazoncek:
Thanks. A video would be beneficial.

aerosh:
> Thanks. A video would be beneficial.

The video shows an example of the work.
The strobe effect was turned on on the garland, and the flame jet effect was turned on on the lamp. The lamp has a 16x16 matrix rolled into a cylinder.
You can also customize these effects if someone thinks it should work a little differently :)

https://youtube.com/shorts/_6KdyX1JDU8?feature=share

aerosh:
> Thanks. A video would be beneficial.

Sorry, the music from the previous video may not be available in some countries.
If you're having trouble watching, I've re-recorded the video with different music for this case.

[link](https://youtube.com/shorts/xiNLuAIpEds?feature=share)

DedeHai:
I like the flame jet effect, but how is it 2D only? looks 1D to me, did not check the code yet.
The music strobe seems like it could be done with one of the existing effects, not sure. did you check if any of the existing ones use a similar threshold and you could just extend that?

aerosh:
Thanks for the quick feedback, @DedeHai 

**Regarding Flame Jet:**
You are absolutely right! The core concept is definitely 1D, and my initial implementation was unnecessarily limited to 2D matrices. I've refactored the `mode_flame_jet` function to support both 1D and 2D segments. It now checks `SEGMENT.is2D()` and uses the appropriate logic for each case. I've also updated the metadata string to reflect this.

**Regarding Music Strobe:**
That's a great point, and I did a thorough check of the existing effects to avoid redundancy. Here's what I found:

- The standard `Strobe` effects are all time-based, not audio-reactive.

- The audio-reactive effects that use `samplePeak` (like `Ripple Peak`, `Waterfall`, `Puddlepeak`) trigger different kinds of visual effects (e.g., a ripple from a random point, a single pixel flash) and, importantly, they don't have a user-configurable volume threshold. They trigger on any detected peak.

My `Music Strob` effect fills a specific gap: it provides a simple, full-segment flash that is triggered only when a beat's volume exceeds a user-defined `Intensity` threshold. I believe this functionality is distinct and useful enough to warrant its own effect rather than complicating an existing one with more options.

I've pushed the changes for `Flame Jet`. Let me know what you think!

wled00/FX.cpp
```diff
@@ -10644,6 +10644,118 @@ uint8_t WS2812FX::addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name)
...
 
+/*
+  Simple bit strob
+*/
+uint16_t mode_music_strob(void) {
+  um_data_t *um_data = getAudioData();
+  if (!um_data) return FRAMETIME; 
+
+  bool samplePeak = *(bool*)um_data->u_data[3]; 
+  int volumeRaw   = *(int16_t*)um_data->u_data[1];
```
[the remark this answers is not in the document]
netmindz: um_data is not strongly typed as different usermods can send different data.  Sadly we can't return a strongly typed result from getAudioData as the AudioReactive usermod might not be included in the build and so dummy data is returned for effects that expect audio

netmindz:
I think I follow the basic idea of the effects, but at the moment they feel like they might need a little more refinement, especially the strobe. You might be better looking at only certain frequencies rather than total volume, plus you might need to do things like compare against a running average, you might also possibly find the zero crossing data useful (think this is possibly only in WLED-MM at the moment)

wled00/FX.cpp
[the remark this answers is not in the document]
netmindz: those helper methods would certainly help code readability, but it does mean adding ar_ functions into the main codebase that relate to a usermod, so you might as well at that stage add strongly typed return of  getAudioData as both is adding AR specific code into the general codebase

netmindz:
Please never use force push

softhack007:
> it provides a simple, full-segment flash that is triggered only when a beat's volume exceeds a user-defined `Intensity` threshold.

This will not work well with AGC. The AGC feature in audioreactive tries to utilize the available signal range up to 85% except for some peaks. Peak/beat detection in AR really needs rework - planned for Q1/2026 - but I would suggest to combine peak detection with your code, otherwise you'll have parts of the music that will trigger a "constant flash" over more than 15 seconds.

wled00/FX.cpp
```diff
@@ -10644,6 +10644,118 @@ uint8_t WS2812FX::addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name)
...
+      for (uint16_t y_offset = 0; y_offset < SEGENV.aux0; y_offset++) {
+        uint16_t current_y = start_y - y_offset;
+        if (current_y >= SEG_H) continue; // Boundary check
+
+        uint32_t base_color = SEGCOLOR(0);
+        CRGB fastled_color = CRGB(base_color);
+
+        uint8_t heat_reduction = map(y_offset, 0, SEG_H, 0, 180);
+        fastled_color.g = qsub8(fastled_color.g, heat_reduction);
+        fastled_color.b = qsub8(fastled_color.b, heat_reduction * 2);
```
softhack007, line 10711: What happens when one of the color channels is less than ``heat_reduction`` ? Will the value wrap around (= higher value), or will it be clamped at 0?

wled00/FX.cpp
```diff
@@ -10644,6 +10644,118 @@ uint8_t WS2812FX::addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name)
...
+      }
+
+    } else {
+      // --- 1D LOGIC ---
+      const uint16_t max_height = SEGLEN;
+      uint16_t start_index = 0; // Or SEGLEN - 1 if you want it from the other end
+
+      for (uint16_t i_offset = 0; i_offset < SEGENV.aux0; i_offset++) {
+        uint16_t current_index = start_index + i_offset;
+        if (current_index >= SEGLEN) continue; // Boundary check
```
softhack007, line 10733: Better to use ``break; `` instead of ``continue``

wled00/FX.cpp
```diff
@@ -10644,6 +10644,118 @@ uint8_t WS2812FX::addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name)
...
+
+        uint32_t base_color = SEGCOLOR(0);
+        CRGB fastled_color = CRGB(base_color);
+        
+        uint8_t heat_reduction = map(i_offset, 0, SEGLEN, 0, 180);
+        fastled_color.g = qsub8(fastled_color.g, heat_reduction);
+        fastled_color.b = qsub8(fastled_color.b, heat_reduction * 2);
```
softhack007, line 10740: Same question as for the 2D case - how is qsub8 handling results <0 ?

platformio.ini
```diff
@@ -10,7 +10,7 @@
 # ------------------------------------------------------------------------------
 
 # CI/release binaries
-default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, nodemcuv2_compat, esp8266_2m_compat, esp01_1m_full_compat, esp32dev, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_16MB_opi, esp32s3dev_8MB_opi, esp32s3_4M_qspi, esp32_wrover, usermods
+default_envs = nodemcuv2, esp8266_2m, esp01_1m_full, nodemcuv2_160, esp8266_2m_160, esp01_1m_full_160, nodemcuv2_compat, esp8266_2m_compat, esp01_1m_full_compat, esp32dev, esp32dev_V4, esp32_eth, lolin_s2_mini, esp32c3dev, esp32s3dev_16MB_opi, esp32s3dev_8MB_opi, esp32s3_4M_qspi, esp32_wrover, usermods
```
softhack007, line 13: Veto ;-) to any changes in platformio.ini.

The modification to platformio.ini is not related to the scope of the PR. Please remove unrelated changes.

wled00/FX.cpp
```diff
@@ -10644,6 +10644,118 @@ uint8_t WS2812FX::addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name)
...
+  uint8_t threshold = SEGMENT.intensity;
+
+  if (samplePeak && volumeRaw > threshold) {
+    SEGMENT.fill(SEGCOLOR(0));
+  }
+  
+  return FRAMETIME;
+}
+
+static const char _data_FX_MODE_MUSIC_STROBE[] PROGMEM = "Music Strobe@Fade speed,Threshold;!,!;!;1v;ix=128,sx=224,si=1";
```
softhack007, line 10670: Please verify if the metadata string is correct. I think that "1v" is for ``virtual strip`` effects, not for fx that support both 1D and 2D.

wled00/FX.cpp
```diff
@@ -10644,6 +10644,118 @@ uint8_t WS2812FX::addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name)
...
+      if (SEGENV.aux0 >= max_height) {
+        SEGENV.aux0 = 0;
+      }
+    }
+  }
+  
+  return FRAMETIME;
+}
+
+static const char _data_FX_MODE_FLAME_JET[] PROGMEM = "Flame Jet@Base Speed,Trail Length,Threshold;!;!;1v;ix=192,sx=128,c1=128,pal=35,si=1";
```
softhack007, line 10757: Same here - please verify if the metadata is correct.

PR #4827, opened 2025-08-09
Bugfix for brightness factor upon save: fixes #4824

### LINKED ISSUE #4824: Brightness factor is not strictly enforced
### What happened?

When using brightness factor to limit the average power draw of a PWM LED strip, sometimes the PWM duty cycle is able to exceed this value.

I don’t know if this behavior also affects digital LEDs.

### To Reproduce Bug

Set a brightness factor of eg 20%, set brightness to 100%/255, then open the LED preferences screen and save (no changes are needed). The PWM duty cycle will then become 100% until the next brightness change.

I also observed an animation glitch where the brightness exceeded the brightness factor for a moment, but I have not been able to reproduce this.

### Expected Behavior

The LED output must absolutely never exceed the configured brightness factor.

### Install Method

Binary from WLED.me

### What version of WLED?

0.15.1

ESP32

### Relevant log/trace output

```shell

```

### Anything else?

I don’t know which flavor of ESP32 this device uses, but it’s whatever is in the GLEDOPTO GL-C-211WL. (They put a custom RF shield on the carrier that has the MCU, and the info screen just says ESP32.)

As a side note, since brightness factor could be used to keep power consumption within hardware limits and could thus constitute a safety issue, ideally it would be preserved when resetting the config by hardware button. (Maybe it already is; I haven’t tested it yet.)

### Code of Conduct

DedeHai:
## Summary by CodeRabbit

* **New Features**
  * Added support for parallel I2S output on ESP32 platforms, including UI options and configuration handling.
  * Enhanced segment geometry and mapping options, with improved 2D and mapping support.
  * Added new environment and platform configuration options for ESP32 V4 and esp32_pico-D4.
  * Improved accessibility with new tooltips and UI element IDs in the web interface.

* **Improvements**
  * Refined color handling to consistently use a dedicated primary color variable for all interfaces and integrations.
  * Expanded and standardized hardware limits and configuration for ESP32-S2 and other ESP32 variants.
  * Enhanced LED bus management, including memory usage estimation and platform-specific optimizations.
  * Improved segment and effect timing for smoother animations and more precise frame control.
  * Updated documentation and changelog with recent releases and requirements.

* **Bug Fixes**
  * Fixed issues with temperature unit display in BME280 usermod and WiFi reconnect logic.
  * Corrected array handling and loop logic in several effects and 2D modes.
  * Addressed potential glitches when accessing the filesystem during LED updates.

* **Refactor**
  * Modernized internal bus and segment management for better performance and maintainability.
  * Replaced fixed arrays with dynamic vectors for bus configuration.

* **Chores**
  * Updated version numbers and build scripts.
  * Improved debug output and error handling for easier troubleshooting.

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -183,9 +183,9 @@ void WLED::loop()
     bool aligned = strip.checkSegmentAlignment(); //see if old segments match old bus(ses)
     BusManager::removeAll();
     strip.finalizeInit(); // will create buses and also load default ledmap if present
-    BusManager::setBrightness(bri); // fix re-initialised bus' brightness #4005
     if (aligned) strip.makeAutoSegments();
     else strip.fixInvalidSegments();
+    BusManager::setBrightness(scaledBri(bri)); // fix re-initialised bus' brightness #4005 and #4824
     doSerializeConfig = true;
   }
   if (loadLedmap >= 0) {


netmindz:
@coderabbitai your generated summary is completely wrong. It does not relate to the commit

PR #4600, opened 2025-03-18
Get TTGO-T-Display usermod to build

willmmiles:
Adopt spiff72's upstream changes, and update the TTGO-T-Display usermod to the new library framework.  This allows it to be included in `custom_usermods` like any other usermod.  All other build requirements are handled by a library script, so the TFT_eSPI library does not need to be hand modified.

Fixes #4375.

NOTE: I do not actually have this hardware, so this is untested beyond "it builds"; I've double checked that the correct build flags are handed to each source, but that's about as far as I can go.   Can someone with one of these boards please give it a try?

PR #4828, opened 2025-08-09
Remove V1 usermod

netmindz:
Remove legacy usermod

## Summary by CodeRabbit

* **Refactor**
  * Streamlined usermod integration to rely on the central usermod manager for lifecycle events, removing legacy callback paths.

* **Chores**
  * Removed unused usermod template code and related documentation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wled/WLED/pull/4828?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


willmmiles:
There are several v1 usermods still in the tree -- they should get ported to v2 before we merge this.  As it turns out, the v1 usermod binding is actually fully compatible with the "usermods as libraries" approach, though you can only enable one (it'll throw a linker error otherwise).

netmindz:
How does that work then? Does the compiler just treat the CPP fille as if it's a duplicate header definition as the function is empty and magically switch to the version in the other directory if on the include path?

willmmiles:
Well blow me sideways - [I wrote it months ago, but I never merged that patch!](https://github.com/willmmiles/WLED/commit/687e1a64d4875f6a5de94f34862818b676083536)  It was in my TTGO-T-Display usermod branch, PR #4600.  I'd flagged it as draft as I didn't have the hardware to test it.  (I have something "close enough" on my test block now; works fine, I should dust it off and submit.)

But yes - that's pretty much how it works, through the power of [weak symbols](https://en.wikipedia.org/wiki/Weak_symbol).   The definitions in `wled00/usermod.cpp` can be flagged to be linked in only if no other source file defines them.

Sorry for the confusion.

PR #4830, opened 2025-08-11
fix: Added platform_packages for esp32s2, c3, s3

localhostgokul:
## Summary by CodeRabbit

- Chores
  - Aligned ESP32‑S2, ESP32‑C3, and ESP32‑S3 build environments to use the same platform packages as the base ESP32 configuration.
  - Improves build consistency across ESP32 variants, reducing discrepancies and potential dependency mismatches.
  - Enhances reliability of firmware builds for these boards without altering features or runtime behavior.
  - No changes to flags, partitions, or libraries; this update is limited to build configuration.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -296,6 +296,7 @@ board_build.partitions = ${esp32.default_partitions}   ;; default partioning for
 [esp32s2]
 ;; generic definitions for all ESP32-S2 boards
 platform = ${esp32_idf_V4.platform}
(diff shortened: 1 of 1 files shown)


DedeHai:
the rabbit may be on to something. did you test this with IDF V4?

netmindz:
Yeah I agree @DedeHai this doesn't look right

If also a little odd that the PR has no description only a title @localhostgokul

PR #4832, opened 2025-08-11
Add new multiple DRGBW realtime DMX mode

mxklb:
Currently missing realtime DMX Mode: Dimmer + RGBW

## Summary by CodeRabbit

* **New Features**
  * Added a new DMX mode: "Dimmer + Multi RGBW" — per-LED RGBW with a master dimmer.
  * Settings UI updated to include and reorder DMX mode options for easier selection.
  * Improved DMX/E1.31 handling so the master dimmer and LED indexing work correctly for multi-LED RGBW setups.
  * RDM/personality support expanded so controllers can detect and use the new mode.


netmindz:
Personally I don't see the need for a dimer channel as your lighting desk should be able to handle this for you, but given we already have for RGB it makes sense to add an option for RGBW too for consistency

PR #4834, opened 2025-08-13
0.15 - updated AsyncWebServer and AsyncTCP

willmmiles:
Backport #4796.  The major improvement here is that the updated AsyncTCP fixes some errors in the handling of unexpectedly closed connections that could cause double frees or use-after-free that result in memory corruption, particularly under sustained load.  The newer AsyncWebServer is backwards compatible and will not activate the queuing feature, though it does add the "emergency 503 on low heap" logic.


netmindz (approved):

Arcitec:
This is brilliant. Thank you for backporting it too! ❤️

PR #4836, opened 2025-08-14
Remove old V3 IDF

netmindz:
Make all ESP32 builds use the V4 ESP-IDF

## Summary by CodeRabbit

* **New Features**
  * Faster builds via a global build cache.
  * Support for an optional user override configuration.

* **Refactor**
  * Standardized ESP32 builds on an ESP-IDF v4 configuration.
  * Simplified and de-duplicated ESP32 environment settings; added a distinct V4-targeted environment.

* **Chores**
  * Defaults updated to prefer ESP-IDF v4 variants; removed legacy ESP32 audioreactive default.
  * Updated environment guidance including OTA/install cautions.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -247,6 +247,7 @@ platform = ${esp32_idf_V4.platform}
 platform_packages =
 build_unflags = ${common.build_unflags}
 build_flags = ${esp32_idf_V4.build_flags}
+lib_deps = ${esp32_idf_V4.lib_deps}
 
 tiny_partitions = tools/WLED_ESP32_2MB_noOTA.csv
 default_partitions = tools/WLED_ESP32_4MB_1MB_FS.csv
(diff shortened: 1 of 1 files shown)


platformio.ini
```diff
@@ -243,37 +243,27 @@ build_flags =
...
-  ${esp32_all_variants.lib_deps}
-  ${env.lib_deps}
+
 board_build.partitions = ${esp32.default_partitions}   ;; default partioning for 4MB Flash - can be overridden in build envs
 # additional build flags for audioreactive - must be applied globally
 AR_build_flags = ;; -fsingle-precision-constant ;; forces ArduinoFFT to use float math (2x faster)
 AR_lib_deps =  ;; for pre-usermod-library platformio_override compatibility
 
 
 [esp32_idf_V4]
```
DedeHai, line 265: If all are V4, do we need to state so? I.e. remove this?

platformio.ini
```diff
@@ -243,37 +243,27 @@ build_flags =
   -D WLED_ENABLE_GIF
 
 [esp32]
-#platform = https://github.com/tasmota/platform-espressif32/releases/download/v2.0.2.3/platform-espressif32-2.0.2.3.zip
-platform = espressif32@3.5.0
-platform_packages = framework-arduinoespressif32 @ https://github.com/Aircoookie/arduino-esp32.git#1.0.6.4
+platform = ${esp32_idf_V4.platform}
+platform_packages =
```
DedeHai, line 247: I saw a PR about adding these as they are missing for s3 etc. So are the packages required? Should they be added to the base definitions?

DedeHai:
Generally looks good to me

platformio.ini
netmindz: My (limited) understanding is that you only need to define if you are overriding the default tools for the platform

platformio.ini
netmindz: Yeah I wasn't sure if it made more sense to update the regular esp32 block or move things over. 

When considering users with their own forks or at least platformio_override I thought it might possibly be best to use explicit V4 params in our example so it was clearer what was being used, but then for backwards compatibility set the default esp32 to reference V4

I'm happy to hear other suggestions 

One other thing to bare in mind is what happens if we start adding some V5 builds into the mix too

DedeHai: thats a good point. in that case keep the V4. Does it need to be added to S2/S3/C3 envs?

netmindz: yes,  c3, s2 and s3 all use platform = ${esp32_idf_V4.platform}

PR #4841, opened 2025-08-16
bugfix to prevent "infinite" loop in palette blend

DedeHai:
this fixes a very long loop when an overflow was happening in palette blending.
- reset prevPaletteBlends to prevent overflow
- add safety check in case overflow should still happen in another combination

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of segment transitions by resetting internal palette-blend counters when starting a copied transition, ensuring consistent color behavior.
  * Added a safety cap on blend passes during FADE blending to prevent excessive iterations, reducing the risk of slowdowns or freezes and improving overall stability.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -282,6 +282,7 @@ void Segment::startTransition(uint16_t dur, bool segmentCopy) {
       _t->_oldSegment = new(std::nothrow) Segment(*this); // store/copy current segment settings
       _t->_start = millis();                              // restart countdown
       _t->_dur   = dur;
+      _t->_prevPaletteBlends = 0;
       if (_t->_oldSegment) {
         _t->_oldSegment->palette = _t->_palette;          // restore original palette and colors (from start of transition)
         for (unsigned i = 0; i < NUM_COLORS; i++) _t->_oldSegment->colors[i] = _t->_colors[i];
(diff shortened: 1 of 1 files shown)


wled00/FX_fcn.cpp
```diff
@@ -368,6 +369,7 @@ void Segment::beginDraw(uint16_t prog) {
     // minimum blend time is 100ms maximum is 65535ms
     #ifndef WLED_SAVE_RAM
     unsigned noOfBlends = ((255U * prog) / 0xFFFFU) - _t->_prevPaletteBlends;
+    if(noOfBlends > 255) noOfBlends = 255; // safety check
```
blazoncek, line 372: No need for this check.

blazoncek (approved):
There is no need to do a check if noOfBlends is > 255 as prog <= 0xffff and _prevPaletteBlends <= 0xff.

Resetting _prevPaletteBlends when transition restarts is enough for correct fix.

wled00/FX_fcn.cpp
DedeHai: are you 100% sure there are no possible combinations where transition time changes  i.e. prog gets smaller and _prevPaletteBlends is not reset?

blazoncek: If we trust in math then yes.
I.e. `prog` is a `_t->_progress` alias. `_progress` only increases with time. One exception is the one you found, when transition is restarted (`_start` and `_dur` change) to create a segment copy.

Issue #4206, opened 2024-10-18
[INFO] Code execution speed considerations for developers


DedeHai:
I want to collect some info here about things I have learned while writing code for the ESP32 family MCUs. Please feel free to add to this.

This is a work in progress. 

### Comparison of basic operations on the CPU architectures

| Operation                   | ESP32@240MHz (MOPS/s) | S3@240MHz (MOPS/s) | S2@240MHz (MOPS/s) | C3@160MHz (MOPS/s) |
|-----------------------------|----------------|-------------|-------------|-------------|
| Integer Addition             | 237.76         | 237.99      | 182.17      | 127.08      |
| Integer Multiply             | 237.05         | 238.06      | 182.17      | 120.74      |
| Integer Division             | 118.94         | 119.03      | 101.29      | **4.63**        |
| Integer Multiply-Add         | 158.49         | 158.66      | 136.63      | 127.22      |
| 64bit Integer Addition       | 19.50          | 20.81       | 18.11       | 36.82       |
| 64bit Integer Multiply       | 27.55          | 30.22       | 27.79       | 15.50       |
| 64bit Integer Division       | 2.71           | 2.71        | 2.65        | 1.02        |
| 64bit Integer Multiply-Add   | 19.80          | 21.88       | 19.16       | 20.30       |
| Float Addition               | 237.55         | 238.04      | 7.77        | 1.93        |
| Float Multiply               | 237.69         | 237.97      | 4.14        | 1.24        |
| Float Division               |** 1.42**           | **4.47**        |**0.86**        | **0.79**        |
| Float Multiply-Add           | 474.85         | 475.91      | 6.43        | 1.76        |
| Double Addition              | 6.50           | 6.18        | 6.51        | 1.51        |
| Double Multiply              | 2.23           | 2.37        | 2.23        | 0.70        |
| Double Division              | 0.48           | 0.54        | 0.30        | 0.41        |
| Double Multiply-Add          | 5.65           | 5.61        | 5.65        | 1.40        |

This table was generated using code from https://esp32.com/viewtopic.php?p=82090# 

Even though the ESP32 and the S3 have hardware floating point units, they still do floating point division in software so it should be avoided in speed critical functions. 

As to why integer divisions on the C3 are so slow is unknown, the datasheet clearly states that it can do 32-bit integer division in hardware.

### Bit shifts vs. division
[...]

PR #4843, opened 2025-08-16
2 Effects: Xmas Twinkle, Elastic Collisions

NPAssoc:
(My apologies for not successfully squashing sub commits. It is the first time I've done this.)

**Xmas Twinkle** is an effect that simulates mid-century Christmas tree twinkle lights that use a bimetallic strip to switch the bulb on and off. Because its contact is kind of sticky, the blinking is random. As the bulb heats up or cools down the blinking rate varies. Each bulb sort of has a mind of its own.

In the  effect, you can control the average rate of blinking and the density of blinking LEDS in a segment, i.e. what percentage of the LEDs blink or are off. LEDs are either on or off. Each LED is assigned a random, but unchanging, index into the current color palette. It works great with a C-9 color palette. A Checkbox causes a new color index to be generated each time a bulb turns on.

It looks great on 2D displays, imitating fake computer consoles in 1960s science fiction movies. It also works great flashing a single LED.

This simulation uses a short term random cycle to control each blinking LED, and a long term random number to control its average rate. Every 20 seconds an LED's long term rate is adjusted randomly. A general function, 'skewedRandom()', part of the code but may be useful elsewhere, is used to skew the long term random numbers to emphasize ranges of twinkle times.

The code uses 5 bytes/blinking LED to keep track of the long term time interval, time to next on or off cycle, time to next long term recalculate, and color palette index. Manual bit fields are used to cram all this data into 5 bytes.

**Elastic Collisions** simulates balls randomly hitting each other and bouncing elastically. Balls also bounce off the edges of a display. You can control, their Speed (velocity), Number of balls; 1-30, Uniformity 0-255, and regeneration time 15 seconds to 1 hour. Ball colors are random indices into the current palette.

Balls have a mass that is the cube of their diameter. Collisions conserve their energy and momentum as per the laws of physics

 A Uniformity of 255 is special. The balls are initialized with equal mass in a row and only one moving. When it collides with another ball, all momentum is transferred to the next ball and it stops, much like the swinging ball puzzle in "Professor T".

a few seconds before a regenerating a new set, the walls/ends "collapse" and the balls drift off the display.

It works very well on both 2D and 1D displays.

• The code matches the general coding style of other effects as suggested in the WLED advanced documentation and is well annotated.

• The code uses explicit shifting and masking to get at the saved data, the C++ bit fields did not work for me.

• It was extensively developed standalone in an Arduino environment, then merged into the main branch for more testing. The code is well isolated. There is only one block of code in FX.cpp.

• It has been extensively tested on single 300 LED strip segment and a 40x32 2D LED display. It has never crashed or misbehaved.

• Note this code uses single precision floating point. It may be possible to convert it to all fixed point, and maybe even the particle simulation objects, if the effect is found to be worthwhile.

## Summary by CodeRabbit

* **New Features**
  * Added the “Xmas Twinkle” lighting effect, with configurable timing and color variation.
  * Added the “Elastic Collisions” effect, featuring bouncing spheres, collisions, and configurable motion.
  * Elastic Collisions supports 1D and 2D displays, wall behavior, regeneration, and adjustable speed, count, uniformity, and lifetime.

* **Documentation**
  * Added setup instructions, parameter descriptions, and usage details for Elastic Collisions.


wled00/FX.h
```diff
@@ -374,7 +374,9 @@ extern byte realtimeMode;           // used in getMappedPixelIndex()
 #define FX_MODE_PS1DSONICBOOM          215
 #define FX_MODE_PS1DSPRINGY            216
 #define FX_MODE_PARTICLEGALAXY         217
-#define MODE_COUNT                     218
+#define FX_MODE_XMASTWINKLE            218
```
blazoncek: Please reuse gaps from removed effects if possible. See lines above.
May need #4573 to be merged 1st. /cc @netmindz

blazoncek (changes requested):
It would be beneficial to post a video of expected effect behaviour to easily judge if existing effects behave similarly.

For elastic collision, there is already a similar effect called Blobs. It would be preferred if that one would be updated to handle collisions instead of adding a new one.

For twinkle, there are several other *twinkles* already in place. Does none of them work similarly?

Other than that, I stopped reviewing when I saw a madness of indentations. Please fix all indentations.

NPAssoc:
> On Aug 16, 2025, at 11:14 AM, Blaž Kristan ***@***.***> wrote:
> 
> It would be beneficial to post a video of expected effect behaviour to easily judge if existing effects behave similarly.
> 
I just posted two videos on Discord "# custom-fx" thread. Let me know if there is a better place.
> For elastic collision, there is already a similar effect called Blobs. It would be preferred if that one would be updated to handle collisions instead of adding a new one.
> 
I’ll take a look. 
> For twinkle, there are several other twinkles already in place. Does none of them work similarly?
> 
No, it is quite a bit different. Existing twinkle effects, fade in and out randomly. My effect blinks on and off in the same position. Each light is assigned a fixed color index normally, so it doesn’t change color—much like twinkle lights on a real Christmas tree string. It also simulates the heating and cooling of a real twinkle light bulb. You can adjust the percentage of lights over a segment and their blink rate. A checkbox controls color assignment.

Because of this, the patten of blinking is more “interesting”.

It works well on a single LED.
> Other than that, I stopped reviewing when I saw a madness of indentations. Please fix all indentations.
> 
Indentations are fine in my Visual Studio Code, where a tab is only two spaces, but I’ll convert tabs to spaces so tabbing is consistent with any spaces/tab. I’ll get to it tomorrow. :-)

DedeHai:
First off I really like these effects and the effort you put into making them physically accurate. There are a few issues that will prevent them making it into default WLED effects:

- As already mentioned, very similar effects do already exist: Twinklecat and PS Blobs / PS Pinball
- Using floats all over is problematic: on my C3 test setup it drops to 8FPS. For comparison: PS Blobs runs at 65FPS
- And last but not least: code size. This PR adds over 4k of flash for just two effects (9k on my C3 test setup!). For comparison: the whole 1D particle system with 15 effects uses 17k. Flash is precious.

When playing around, some sliders were unresponsive by themseles and only changed things wen sliding another one. Needs some refinement.

Here are my suggestions:
- Update Twinklecat with a checkmark and add some of your code to it, keep it as simple as possible
- If you want to keep your effects as options in WLED code: move it to the user FX usermod: here the "rules" are less strict as these are not maintained by the devs but by the creators and they are not compiled in by default.

NPAssoc:
Hello DedeHai, et al,

Thank you for the review. Your points are well taken.

I actually wrote these effects over a year ago in my own environment but I had trouble figuring out submission procedures until I had more time—a contract job interfered. I only got back to moving them to WLED recently.

Since a year has gone by, the effect landscape has changed a bit, hence the overlap. Also the Particle System did not exist at the time I wrote Elastic Collisions.

Some of the tabbing problems were a result of the port from my independent project, I’ff fix that up.

> On Aug 17, 2025, at 3:29 AM, Damian Schneider ***@***.***> wrote:
> 
> DedeHai
>  left a comment 
> (wled/WLED#4843)
>  <https://github.com/wled/WLED/pull/4843#issuecomment-3194196289>
> First off I really like these effects and the effort you put into making them physically accurate. There are a few issues that will prevent them making it into default WLED effects:
> 
> As already mentioned, very similar effects do already exist: Twinklecat and PS Blobs / PS Pinball
> Using floats all over is problematic: on my C3 test setup it drops to 8FPS. For comparison: PS Blobs runs at 65FPS
With careful coding, I could probably get rid of all the floating point. If I want fractional numbers, I  could use fractional fixed point with a virtual binary point shifted left. I’ve worked on main frame computers in the very distant past that had no floating point.

Even more intriguing would be to convert Elastic Collisions to use the Particle System. That  would eliminate a whole lot of duplicate code. Would there be any documentation of PS beyond the code, by chance?
> And last but not least: code size. This PR adds over 4k of flash for just two effects (9k on my C3 test setup!). For comparison: the whole 1D particle system with 15 effects uses 17k. Flash is precious.
Agreed.
> When playing around, some sliders were unresponsive by themseles and only changed things wen sliding another one. Needs some refinement.
> 
I could fix that by detecting slider deltas, but having only .aux0, .aux1 and .step to store stuff between invocations, I’d have to do something tricky, unless you have another suggestion or I’m missing something.
> Here are my suggestions:
> 
> Update Twinklecat with a checkmark and add some of your code to it, keep it as simple as possible
I’ll look at that.
> If you want to keep your effects as options in WLED code: move it to the user FX usermod: here the "rules" are less strict as these are not maintained by the devs but by the creators and they are not compiled in by default.
If you feel that these effect algorithms are worthwhile, I’d be willing to put in the time to revise them and/or incorporate them into existing effects in a way that wouldn’t intrude on them.

I’m glad to do the work, but only if there is sufficient interest.

Thank you.

Best regards,

Nick Pisarro

DedeHai:
> Even more intriguing would be to convert Elastic Collisions to use the Particle System. That would eliminate a whole lot of duplicate code. Would there be any documentation of PS beyond the code, by chance?

unfortunately not but I did try to comment the code and I made many FX that use it so you can take it from whichever PS FX is closest to what you want to do. 
In a nutshell: it provides particles with some properties and optional advanced properties. Functions to apply basic physics, collision handling and rendering. There are two versions for 1D and 2D with different optimizations to keep it as speedy as possible, an ESP32 can handle a few thousand particles at high frame rates and it uses no floats so its quite fast even on S2/C3 and ESP8266.
If you want to play with it, please go ahead, any questions on coding details I can answer in discord.

I also have a trick in at least one of the PS FX to use aux to track all sliders: just sum them all up and store.

I can only speak on my behalf and the "Elastic Collisions" effect is just too bulky for what it is and there is PS Blobs which is quite close in overall looks and much more versatile. Except it does not handle mass. 
The new blinking lights effect could indeed be very interesting on pucks or bullet strings to make them look "old school". Since its very close to twinklecat, extending that FX makes more sense to me then we feel that there are too many FX already in the list plus there is the issue of limited IDs. Both these issues will be addressed in the future but we do not have a good solution just yet.

wled00/FX.cpp
```diff
@@ -7477,6 +7477,738 @@ uint16_t mode_2DAkemi(void) {
...
+#define TIME_TO_EVENT         0x7fe00000    // 10 bits >> 21
+#define TIME_TO_EVENT_SHIFT   21
+#define MAX_CYCLE             0x001ff800    // 10 bits >> 11
+#define MAX_CYCLE_SHIFT       11
+#define T_RETWINKLE           0x000007ff    // 11 bits >> 0
+} XTwinkleLight;
+
+// For creating skewed random numbers toward the shorter end.
+// The sum of percentages must = 100%
+const uint8_t pSize = 20;
```
DedeHai: static arrays are a no-go due to RAM usage. Approximate the table with a formula, use SEGMENT.data for the rest.

NPAssoc:
On Aug 17, 2025, at 7:59 AM, Damian Schneider ***@***.***> wrote:
> I also have a trick in at least one of the PS FX to use aux to track all sliders: just sum them all up and store.

I’m already using .aux0 & 1 but there is spare byte there. I’ll xor the sliders. This will work because only one slider changes at a time.
> I can only speak on my behalf and the "Elastic Collisions" effect is just too bulky for what it is and there is PS Blobs which is quite close in overall looks and much more versatile. Except it does not handle mass.
> The new blinking lights effect could indeed be very interesting on pucks or bullet strings to make them look "old school". Since its very close to twinklecat, extending that FX makes more sense to me then we feel that there are too many FX already in the list plus there is the issue of limited IDs. Both these issues will be addressed in the future but we do not have a good solution just yet.
> 
I’ve been doing a bit of poking around and recoding.

It seems to me that Colortwinkles is the closest to my effect, rather than either TwinkleCat or Fox.

Yes, most of the code is for Elastic Collisions. I’ll look at recoding it to use the PS objects.

Looking around, Xmas Twinkles has about the same code load as other effects.

At a basic level, Elastic Collisions and Blobs are similar, but there are several differences.
  • Elastic Collisions uses a fractional coordinate space and anti-alias drawing. This is a major difference. See how smoothly Elastic Collisions particles move, particularly at very low speeds. Blobs movement is very jerky and Blob objects are awkward because the blobs can only have integer sizes and velocities. Blob objects don’t appear to touch before bounding off each other.
  • Collisions works on 1D as well as 2D displays. Try Collisions on an LED string. Try a uniformity of 255, which is special.
  • Collision objects have mass and momentum. Mass is just radius cubed. Collision object initial velocities are the inverse of size to keep their momentums more equitable.

I have laboriously recoded the effects to use fixed point—ripping out all the floats. I use Q16.16 (32 bit with the binary point in the middle) arithmetic. This had a minor effect on Xmas Twinkles, but it took two weeks to redo Elastic Collisions. While in theory, you can do most calculations just as accurately using fixed point arithmetic as floating point, the bookkeeping necessary to avoid overflows and precision loss is pretty breathtaking. I’m afraid this did nothing to help the code load of Collisions. I also had to include a fast int64 floating point SQRT function.

A radical idea is to replace PS Blobs code with Elastic Collisions code. The differences, which could be adjusted, are:
  • Blob size distribution is different.
  • How the effects regenerate. Control with a check box?
  • You’d gain smooth movement and sizing.
  • You’d gain working on a 1D string.

Here is my project list as I see it:

• Catch all slider changes so they affect the effects immediately.

• Use retired effect numbers for any effects I keep standalone.

• Fix column indents in the source code.

• I’ve already ripped out all the floating point, but I have not reduced code size.

• Options for Xmas Twinkle: Add it to Colortwinkles as a checkbox—labeled “Classic", to TwinkleCat or keep it standalone. I really think it is best standalone. It supports 0d, vs. all the other twinkle effects. I believe it is sufficiently different to be its own effect, rather than a ‘hidden’ feature. But it is your call.

• Look at recoding Elastic Collisions using PS objects. This could reduce its code footprint. PS objects would need to support fractional coordinates. It might be possible to “fake it” by multiplying all coordinates by a factor, 256, say and reducing it out in a draw function. Using Q16.16 arithmetic, like I do now, runs into overflow problems when you multiply, but Q24.8 may avoid that.  But it would require using int32_t for PSparticle position.

Having the particle systems work with fractional coordinates could open up the possibility of adding still more sophisticated effects. I noticed a trend in making effects more ‘interesting’—that have sophisticated life cycles, like PS Hourglass, rather than just blink and flash.

Hmmm, Maybe you have some ideas here? I’m just throwing thoughts out. I need to understand the PS world more.

• I’ll keep making my changes into this pull request, but if you want, I can eventually merge out my changes with a new PR.


Best regards,

Nick

DedeHai:
>Looking around, Xmas Twinkles has about the same code load as other effects.

it does, but colortwinkle for example uses 1/3 of your code. You could achieve an almost identical effect with much less code. It is way over-engineered the way it is, there are some things like static arrays that are inacceptable. If you want I can show you what I did with the help of claude.ai to mimic your effect with much less code and no static ram used and I am sure it can be improved upon. I am sorry if I am being picky but we spent countless hours to cut down code size and make as many features fit as possible. Spending 1k of code on a (sophisticated) blink effect is just way too much IMHO.

Some of your statements about the PS are just not correct, you should look at it more closely.

> At a basic level, Elastic Collisions and Blobs are similar, but there are several differences. 
• Elastic Collisions uses a fractional coordinate space and anti-alias drawing.

so does the PS.


> See how smoothly Elastic Collisions particles move, particularly at very low speeds. Blobs movement is very jerky

they are jerky on purpose to mimic the old effect. look at ballpit.


> and Blob objects are awkward because the blobs can only have integer sizes and velocities. 

nothing wrong with that, size goes from 0-255, velocity is ±127. Its a tradeoff between accuracy, speed and RAM. Try instantiating 1000 of your spherese and see what you get ;)

-> Blob objects don’t appear to touch before bounding off each other. 

that is true, the size calculation is not perfectly accurate and could use improvement but I was not able to make it better without taking away speed, remember: some effects use hundreds of particles.


> Collisions works on 1D as well as 2D displays. Try Collisions on an LED string. 

I did, looks nice. you did not try the PS version I suggested I assume.


> Try a uniformity of 255, which is special. • Collision objects have mass and momentum. Mass is just radius cubed. Collision object initial velocities are the inverse of size to keep their momentums more equitable. 

I do not doubt that one can do nice things by spending lots of flash and CPU time (and floats). 

> A radical idea is to replace PS Blobs code with Elastic Collisions code. 

No. But you could present a PS version of your code. It will take you less than 2 weeks to change a few lines in PS Blobs and put it in a new FX ;)
If you want to have smoother movement, you can try to improve the rendering of larger PS particles.

As stated initially, you can add your effect to the user-FX usermod. Adding another "physics engine" just for one effect is just not going to fly. With the PS it will look almost the same and "good enough" for 20% of the code size.

p.s.
if you want to see how much better the PS Blobs looks, check out the old Blobs in 0.15

DedeHai:
I briefly looked through your code, you may want to read this: https://github.com/wled/WLED/issues/4206

NPAssoc:
> On Sep 6, 2025, at 1:36 AM, Damian Schneider ***@***.***> wrote:
> 
>  briefly looked through your code, you may want to read this: #4206 <https://github.com/wled/WLED/issues/4206>
Very cool. Thanks

I see floating point operations/sec. for  older chips are pretty grim. I assume those are emulated.

I’m finally going to look at greatly reducing the code load, perhaps by using PS to do it. I didn’t know about static arrays. If I think I’ll need to modify PS, I’ll run it by you.

By the way, I do limit the number of elastic spheres to 32, because they are computationally intensive. Is that too limiting?

I too, have been using Claude & chatGPT to write code, though their recommendations often have to be carefully debugged.

I’ll probably have a question or two. — Nick

P.S. PS Blobs does not show in the Effects list if you don’t have a 2D panel defined.

DedeHai:
> Very cool. Thanks I see floating point operations/sec. for older chips are pretty grim. I assume those are emulated.

S2 and C3 are newer chip, it all comes down to hardware support: no FPU no good float performance. 

> I’m finally going to look at greatly reducing the code load, perhaps by using PS to do it. 

if you find the results are not what you wanted, you can put your original effect (even with floats) into the effects usermod, which is the place created exactly for effects that are not well suited for everyone or are more setup specific.


> I didn’t know about static arrays.

any variable will reside in RAM, if at global scope, they are always in RAM. thats why C(++) has the scope brackets { }

>If I think I’ll need to modify PS, I’ll run it by you. 

please do, maybe run it by me before spending too much time on it as I tried many things and can tell you if something is doable before its even implemented. In general: things that operate on a per-particle basis, every single instruction counts.

>By the way, I do limit the number of elastic spheres to 32, because they are computationally intensive. Is that too limiting? 

for your effect, no. for a particle system: yes. 

>I too, have been using Claude & chatGPT to write code, though their recommendations often have to be carefully debugged. I’ll probably have a question or two.

They do ok if you know what you want and how to instruct them, but its often more time consuming than just writing the code yourself and yes, you have to carefully check the code and often re-structure and clean it up but they are fantastic to write quick test functions to evaluate something.

>PS Blobs does not show in the Effects list if you don’t have a 2D panel defined.

its a 2D only effect, PS has two versions: 1D and 2D both with different optimizations to minimize RAM use and maximize speed. I was thinking of writing code to enable "1.5D" rendering for 2D effects i.e. render them to a strip but I found its better to write an effect specifically. I think I wrote over 30 effects, each with many parameters to tune to ones liking. You can for example turn PS Volcano into a "whirlpool" effect by choosing the right parameters.

DedeHai:
just for reference https://gist.github.com/DedeHai/bbeaa71fa613610e6a7683f4688a48b5 my suggested simplified version of the twinkle effect.

NPAssoc:
Thanks, I’ll check it out.

I got pulled onto some other projects, but I’ll update my WLED source and take a look.

I’ll probably make my Elastic Collisions effect a usermod, at least for  now.

Best,

Nick

NPAssoc:
Getting back into WLED: I brought my source up to date and compiled. I have to remerge in my changes, then try your code in place of it.

Nick

NPAssoc:
Hi DedeHai,

I plugged in your code to try out. I basically like the blinking pattern but I found two serious bugs and an optimization in the code.

• TwinkleLight.timing.nextEvent holds an absolute time, rather than an increment. Even though it is centiseconds, after a few minutes it overflows and the display starts flashing 40 times/second. I now store an increment there, in ms. millis() will also wrap around, eventually, and I handle this too.

• skewedTime() parameter was defined as uint8_t, not uint16_t—too small to hold the parameter passed to it.

‘ skewedTime() has floating point in it. ‘normalized’ can be defined a uint32_t and the order of calculation can be reordered to preserve precision.

And it does need some more tuning, so other than a simpler random generator, I don’t know how much it saves over my version. I’m going to work on making Elastic Collisions a UserMod and then revisit this. I’ll keep you posted.

Nick

DedeHai:
My code is just a rough sketch as a suggestion, definitely nothing finished. The main difference is code size and complexity.

NPAssoc:
OK,

When I revisit it, I will probably use elements from both yours and mine.

NPAssoc:
I've rewritten Xmas Twinkle with suggested and simplified skewed Random number generator, now converted to Fixed Point. I added a new control, "Avg. Duty Cycle" that controls on/off cycle length. See if you like that, and the effect name.

Elastic Collisions is now a UserMod as it is acres of code and I'm not sure how to use the Particle System for it, particularly since I do sub-pixel calculations.

I also fixed all the code indentation.

My changes are in the "ElasticAndTwinkle" branch. I didn't know if I should merge it into the 'main' branch of my PR.

softhack007:
Could it be that we now have a third "elastic collisions" effect?
* this PR
* the "rolling balls" effect with "collision" option https://github.com/wled/WLED/blob/8204e1bafe0ab5795b67c644cd5cecd2ed391b36/wled00/FX.cpp#L3053
* The "PS Pinball" 1D effect that can also do "rolling balls" https://github.com/wled/WLED/blob/8204e1bafe0ab5795b67c644cd5cecd2ed391b36/wled00/FX.cpp#L9524
* The 2D "PS Ballpit" effect that also supports collisions https://github.com/wled/WLED/blob/c05097351d24053fedb2a30ee5fca07f1ea58c6f/wled00/FX.cpp#L8452
* not sure if there is another variant, for example in the user_fx usermod?

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
```diff
@@ -0,0 +1,540 @@
...
+
+    // Initial guess from highest bit.
+    int lz = 63 - countLeadingZeros64(n);
+    uint64_t res = 1ULL << (lz / 2);
+
+    // Newton–Raphson refinement (3–4 iterations are plenty)
+    res = (res + n / res) >> 1;
+    res = (res + n / res) >> 1;
+    res = (res + n / res) >> 1;
+
```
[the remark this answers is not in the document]
softhack007: @NPAssoc this alternate "fast SQRT" should not be necessary
* we already have an integer-based [``sqrt32_bw()``](https://github.com/wled/WLED/blob/c05097351d24053fedb2a30ee5fca07f1ea58c6f/wled00/wled_math.cpp#L225), 
   and I doubt that you can beat the performance ;-)
* on boards with FPU (esp32, esp32-S3) using the native ``sqrtf()`` function should be much faster
* doing 4 Newton-Raphson iterations is overkill, unless your Initial estimate is very bad. Additionally, the integer division you use _will_ drag down performance on esp32-C3 and esp8266
* if you absolutely want this for speed, look into the "quake3 fast inverse sqrt" for better speed (example code is here: https://github.com/kosme/arduinoFFT/blob/6b3ef9732db800bb2c658eefb7ad8939eb44e6e2/src/arduinoFFT.cpp#L468-L486)

softhack007:
> Elastic Collisions is now a UserMod as it is acres of code

@NPAssoc maybe you could add this effect to the [``user_fx`` usermod](https://github.com/wled/WLED/tree/main/usermods/user_fx) rather than creating a dedicated usermod for one effect?

NPAssoc:
> Could it be that we now have a third "elastic collisions" effect?

I looked at those effects and they are quite a bit different than "Elastic Collisions". With some work, it might be possible to bend "PS Ballpit" around to work like my effect, but it would take a lot of work. And "Elastic Collisions" works on both 1D and 2D arrays, which Ballpit does not. Give "Elastic Collisions a try and see.

NPAssoc:
> > Elastic Collisions is now a UserMod as it is acres of code
> 
> @NPAssoc maybe you could add this effect to the [`user_fx` usermod](https://github.com/wled/WLED/tree/main/usermods/user_fx) rather than creating a dedicated usermod for one effect?

I'll take a look…

DedeHai:
given its complexity, I would say it's a good candidate for its own (out of tree) usermod, user FX is possible too but optional

NPAssoc:
> On May 5, 2026, at 1:07 PM, Frank Möhle ***@***.***> wrote:
> @NPAssoc <https://github.com/NPAssoc> this alternate "fast SQRT" should not be necessary
> 
I needed a 64 bit integer square root.

Nick

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
```diff
@@ -373,6 +376,7 @@ nfixed sliderToSpeed(uint8_t slider) {
   
   // Calculate ax² + bx + c in Q16.16
   int64_t result = ((int64_t)a_q16 * x * x) + ((int64_t)b_q16 * x) + c_q16;
```
softhack007, line 380: I'm wondering if this gets the math right? 
if X is fixed point and has a shift > 0, then ``a*X*X`` and ``b*x`` should both need a shift back, so the "fixed point" stays in the right position.

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
softhack007, line 1: @NPAssoc I'll ask the rabbit to double-check if the new fixed point math still has any flaws - fixed point can be a real beast ;-)

@coderabbitAI please double-check all fixed point calculations (``nfixed`` type, and intermediate steps) to verify that fixed point math works as expected, and fixed point resolution is kept where needed. Also look for *undefined behaviour* in bitshift ops (compiler-defined behaviour is acceptable, all code is compiled with gcc on 32bit architectures only).

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
```diff
@@ -0,0 +1,544 @@
...
+			lowY = 0;
+		if (highY > gridH)
+			highY = gridH;
+
+    /* Loop over a range of pixels on a panel to see how bright the LEDs
+     * there should be to represent this object. */
+		for (int lY = lowY; lY < highY; lY++) {
+			for (int lX = lowX; lX < highX; lX++) {
+				// LED pixel center in high-resolution space
+        const nfixed halfPixel = 1 << (SPHERE_PREC_SHIFT - 1);
```
softhack007, line 302: indentation seems broken here

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
```diff
@@ -0,0 +1,544 @@
...
+        // float alpha = RGBEffect::clamp(1.0f - RGBEffect::smoothstep(FLOAT_IT(edge0), FLOAT_IT(edge1), dist), 0.0f, 1.0f);
+        nfixed alpha = clamp((1 << SPHERE_PREC_SHIFT) - smoothstep(edge0, edge1, dist), 0, 1 << SPHERE_PREC_SHIFT) + 0;
+        // nfixed alpha = clamp((1 << SPHERE_PREC_SHIFT) - smoothstep(edge0, edge1, dist), 1 << (SPHERE_PREC_SHIFT - 2), 1 << SPHERE_PREC_SHIFT);
+        // alpha = 1 << SPHERE_PREC_SHIFT;
+	
+				// Store intensity in LED array (0-1 range)
+				if (draw)
+				{
+					drawColor = sphereColor;
+					drawColor.nscale8(alpha * 255 >> SPHERE_PREC_SHIFT);
```
softhack007, line 315: not sure about correct operator precedence between `*` and `>>` -> maybe add brackets so the compiler understands you right.

usermods/elastic_collisions/Elastic_Collisions.cpp
```diff
@@ -0,0 +1,544 @@
...
+				// Store intensity in LED array (0-1 range)
+				if (draw)
+				{
+					drawColor = sphereColor;
+					drawColor.nscale8(alpha * 255 >> SPHERE_PREC_SHIFT);
+				}
+				else
+					drawColor = CRGB::Black;
+	
+				if (alpha > 0.0)
```
softhack007: comparing with "0.0" (double) will slow the loop down heavily, as the compiler is obliged to do the comparison in "double" instead of integer. (see overall analysis by the rabbit)

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
```diff
@@ -0,0 +1,544 @@
...
+      ++SEGMENT.aux0;       // Increments SPHERES_ALLOCATED
+    }
+    
+    SEGMENT.aux0 = (numSpheres << SPHERES_DESIRED_SHIFT) | (SEGMENT.aux0 & SPHERES_ALLOCATED);
+    SEGMENT.step =  millis() + BOUNCE_CYCLE_TIME;
+    SEGMENT.aux1 = elasticLifetime();
+  }
+
+  // If it is time to do something.
+  if (millis() > SEGMENT.step) {
```
softhack007, line 488: in general, effects should prefer ``strip.now`` over ``millis()``; ``strip.now`` can be synchronized over several WLED effect instances. If you want to keep ``millis()``, add short comment why ``strip.now`` can't be used here.

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
```diff
@@ -0,0 +1,544 @@
...
+      spheres[i].update(speed);
+
+      // If nearing a regeneration, let the walls fall and the spheres fly off!
+      if (SEGMENT.aux1 > WALL_COLLAPSE_INTR)
+        spheres[i].handleWallCollision(internalX, internalY);
+    }
+
+    // Check for collisions with other spheres.
+    for (int i = 0; i < (SEGMENT.aux0 & SPHERES_ALLOCATED); ++i)
+        for (int j = i + 1; j < (SEGMENT.aux0 & SPHERES_ALLOCATED); ++j)
```
softhack007, line 510: you might gain some speed by pre-calculating ``SEGMENT.aux0 & SPHERES_ALLOCATED`` before the loop, and storing this into a local variable; the result will not change inside the nested loop.

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
```diff
@@ -0,0 +1,544 @@
...
+  bool areSpheresColliding(MBSphere sp) {
+    nfixed dist = fixedDist(sp.x - this->x, sp.y - this->y);
+    return dist <= this->radius + sp.radius;
+	}
+
+  /* Make sure two spheres haven't gotten too close.
+   * Note: There is a pathological case where two spheres
+   * can crash into each other so hard, that one actually
+   * ends up insde the other. This function prevents that. */
+  void enforceMinDist(MBSphere *sp) {
```
softhack007, line 142: Nitpick comment by the rabbit:
> Avoid unnecessary copying in areSpheresColliding()
> Passing MBSphere by value creates a copy every time; it’s not needed and wastes cycles (and adds heap fragmentation) on embedded targets. Make it a const reference and mark the method const.
> ```diff
> - bool areSpheresColliding(MBSphere sp)
> + bool areSpheresColliding(const MBSphere& sp) const
> ```

softhack007: sorry, I did not see that you now pass `sp` by pointer. The rabbit suggestion is outdated, please ignore :-)

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
```diff
@@ -0,0 +1,544 @@
...
+		 * an LED array. I want to draw a solid circle of diameter 'r' and position 'x'
+		 * and 'y' in the LED array, anti-aliasing the pixels."
+		 * Optimize the loop to only working on pixels near the object. Don't do the
+		 * whole panel. */
+    nfixed edge0 = radius - (1 << SPHERE_PREC_SHIFT) / 2;  // Soft transition start
+    nfixed edge1 = radius + (1 << SPHERE_PREC_SHIFT) / 2;  // Soft transition end
+    int lowX = (x - edge1 >> SPHERE_PREC_SHIFT) - 1;        // We don't need to cut it too close.
+    int highX = (x + edge1 >> SPHERE_PREC_SHIFT) + 2;
+    int lowY = ((y - edge1 >> SPHERE_PREC_SHIFT)) - 1;
+    int highY = ((y + edge1 >> SPHERE_PREC_SHIFT)) + 2;
```
softhack007, line 277: nitpick: maybe check if operator precedence is used correctly here. '+' and '-' have lower precedence than '>>', so your assignments are actually handled as ``int highX = (x + (edge1 >> SPHERE_PREC_SHIFT)) + 2;`` etc. adding explicit brackets might make the logic clearer.

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
softhack007: Edit: could be correct - x is not fixed point.

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
DedeHai: generally, shifts, modulo and binary operators have low priority but always put brackets!

DedeHai:
just a general comment: on controllers with an FPU float calculations are about 10x faster than 64bit calculations, on S2 and C3 without an FPU 64bit it's the other way around.

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
NPAssoc: > Edit: could be correct - x is not fixed point.

The coefficients have been imperially adjusted over what the AI suggested so I get the range I want.

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
NPAssoc: CodeRabbit is wrong! << has lower precedent in C++ than +/-, not higher. I checked the original C++ volume by Bjarne Stroustrup. I'm fixing the accidental f.p comparisons.

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
NPAssoc: Double checked. I am moving a couple of open braces, but indentation is correct.

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
NPAssoc: << and >> are of lower precedence than other arithmetic in C++. I checked this carefully.

usermods/elastic_collisions/Elastic_Collisions.cpp
NPAssoc: Got it! :-)

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
NPAssoc: Optimizing compilers should pull that out of the loop automatically.

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
NPAssoc: Yes, I checked the C++ "Bible". '<<' and '>>' have lower precedence. This is not true for some other languages. (The code would crash, if I got this wrong.)

NPAssoc:
> just a general comment: on controllers with an FPU float calculations are about 10x faster than 64bit calculations, on S2 and C3 without an FPU 64bit it's the other way around.

And the ESP32-S3 has a multiply-add instruction that runs at _twice the clock rate_ of the rest of the core.

Ah, "the mind reels." Ideally, one would like to see the actual generated code. You can do that in VSCode, but it is not straightforward. I'm sure there are a few more "Gotchas" in there. (I'm a former CPU designer.) By the way, check out th upcoming ESP32-S31.

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
softhack007: yes, looks OK. 

Side note: never trust any AI when it comes to math. Most AI tools make _horrible_ mistakes when calculating numbers.

for example: i had `x = 3+2` in my code, the AI insisted that the value of x is `7` 😆

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
softhack007: I agree about the operator precedencies - the official reference https://en.cppreference.com/c/language/operator_precedence says the same as good-old Bjarne stated 40 years ago ;-)

Still, it would be good for maintenance if you add some braces around longer expressions, so people reading your code don't draw wrong conclusions ;-)

@coderabbitai double-check your findings related to C/C++ Operator precedence.

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
softhack007: in theory - yes they should. but in practice, sadly it usually does not happen.

One reason is that SEGMENT is accessing a global object member via pointers. We are in a multi-threaded OS, so the compiler cannot prove that SEGMENT.aux0 has not changed without seeing the whole program - the expression might be re-evaluated in each loop iteration (we are building WLED with -Os not with -O2).

usermods/elastic_collisions_fx/Elastic_Collisions.cpp
softhack007: ok, i thinks that's clear now. As I said earlier, a few braces might help innocent readers to understand the code correctly ;-)

NPAssoc:
> On May 6, 2026, at 7:25 AM, Frank Möhle ***@***.***> wrote:
> 
> You might gain some speed by pre-calculating SEGMENT.aux0 & SPHERES_ALLOCATED before the loop, and storing this into a local variable; the result will not change inside the nested loop.
> 

I looked at this.

If the calculation was more complicated, it would be worthwhile to precalculate “EGMENT.aux0 & SPHERES_ALLOCATED” before the loop. But performing this calculation only adds a single “AND” instruction.

Nick

NPAssoc:
I agree with Damian.

user_fx usermod, for the most part, are fairly simple 1D effects.

Note also, that “Xmas Twinkle” works well as a 0D (i.e. a single LED) effect. The other twinkle effects, as far as I can tell, wouldn’t.

Nick

softhack007:
> I looked at this. If the calculation was more complicated, it would be worthwhile to precalculate “EGMENT.aux0 & SPHERES_ALLOCATED” before the loop. But performing this calculation only adds a single “AND” instruction. Nick

Hi @NPAssoc, I've played a bit with your effect recently. Looks nice, while I still agree with @DedeHai that the physics and anti-aliased drawing could be done with the FXParticleSytem core, too.

It looks like the ACTUAL hot paths are in the AI-generated functions `drawMe()` (paints a ball) and `fixedDist()` / `fixed64Sqrt()` (sqrt on int64).

The main improvement idea for `drawMe()` is to avoid calculating fixed64Sqrt(). 
* If you are fully outside or fully inside the circle, you can skip the sqrt that is just needed for anti-aliasing the circle edge. 
* Computing "distance^2 = `dx*dx + dy*dy`" is sufficient to decide if a point is inside the circle; with distance^2 we can avoid calculating the square root. If "distance > radius", then also "distance^2 > radius^2".

 I can give you some concrete code improvements later (currently away from WLED). 

cheers, Frank.

NPAssoc:
> On May 27, 2026, at 8:49 AM, Frank Möhle ***@***.***> wrote:
> 
> Hi @NPAssoc <https://github.com/NPAssoc>, I've played a bit with your effect recently. Looks nice, while I still agree with @DedeHai <https://github.com/DedeHai> that the physics and anti-aliased drawing could be done with the FXParticleSytem core, too.
> 
> It looks like your real hot paths are actually the AI-generated functions drawMe() (paints a ball) and fixedDist() / fixed64Sqrt() (sqrt on int64). I can give you some concrete code improvements later (currently away from WLED).
> 
> The main improvement idea for drawMe() is to avoid calculating fixed64Sqrt(). If you are fully outside or fully inside the circle, you can skip the sqrt that is just needed for anti-aliasing the circle edge. For all other points, computing "distance^2" is sufficient; with distance^2 we can avoid calculating the square root. If "distance > radius", then also "distance^2 > radius^2”.
> 
Thanks for the comments. I too, am off tilting at other windmills including a paying gig.

I’d be glad to work on this some more in the future, but for now, I’d like to get released what I have.

To move forward, should I merge my work branch, ‘ElasticAndTwinkle’ into the main branch of my PR after bringing that up to date?

What happens next.

Nick

usermods/elastic_collisions_fx/library.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "Elastic Collisions",
```
DedeHai, line 2: I would change the name to "Elascit Collisions FX" to indicate this is an additional effect. or move it to user_FX usermod

wled00/FX.cpp
```diff
@@ -7749,6 +7749,132 @@ void mode_2DAkemi(void) {
...
+
+// Originally by Nick Pisarro, Jr. This version by DedeHai, and updatged by Nick.
+
+// We need to keep data for each twinkle light. 8 bytes/light
+typedef struct {
+  uint32_t nextEvent : 16;      // Time left to next state change (ms.)
+  uint32_t maxCycle : 16;       // Working maximum cycle time
+  uint32_t retwnkleTime : 16;   // Time left til we recalculate 'maxCycle'
+  uint32_t colorIdx : 8;        // May be fixed or change with each flash
+  uint32_t isOn : 1;
```
DedeHai, line 7763: add a comment that 7bits of padding are added which could be used in the future

wled00/FX.cpp
```diff
@@ -7749,6 +7749,132 @@ void mode_2DAkemi(void) {
...
+  uint32_t lastTime = SEGMENT.step;
+  uint32_t currTime = millis();
+  if (currTime < lastTime)
+    SEGMENT.step = lastTime = 0;
+  
+  // The interval may be zero if the refresh rate is fast enough.
+  uint32_t interval = currTime - lastTime;
+  
+  // Initialize on first run
+  if (! SEGMENT.call || SEGMENT.aux0 != numLights) {
```
DedeHai: use SEGMENT.call == 0 for clarity

wled00/FX.cpp
```diff
@@ -7749,6 +7749,132 @@ void mode_2DAkemi(void) {
...
+
+// Based on the speed, bias the maxtime toward faster or slower times.
+int16_t skewedMax()
+{
+  int32_t slowWeight = SEGMENT.speed;    // 0.0 - 1.0 as fixed Q24.8
+  // ">> 9, below divides by 2 and converts Q24.8 to Q32.0.
+  return (skewedTime(8800, 0) * slowWeight + (8800 - skewedTime(8800, 0)) * (256 - slowWeight) >> 9) + 200;
+}
+
+void mode_XmasTwinkle(void) {
```
DedeHai, line 7784: Not sure why you want to call it "Xmas Twinkle". It's an old-school twinkle effect that I would not associate with xmas but your call.

usermods/elastic_collisions_fx/library.json
NPAssoc: Now 'elastic_collisions_fx'.

wled00/FX.cpp
NPAssoc: √

wled00/FX.cpp
NPAssoc: Testing a variable for zero/not zero by just referencing it in a boolean clause is pretty standard c stuff, but sure!

wled00/FX.cpp
NPAssoc: I'm simulating the old time Christmas tree twinkle lights. I only remember them on Christmas trees.

wled00/FX.cpp
DedeHai: it is, and it is often used but most (all?) other FX do the zero check, so its mostly for consistency

wled00/FX.cpp
NPAssoc: I thought the 'PROGMEM' keyword took care of that? I originally designed 'skewedRandom()' to be used in several places, I thought it might be a useful tool for use in other effects. Now it's used in only one place. I'll see if A.I. can come up with a function that simulates that one use and isn't too complicated. If your comment is just for the 'pSize' variable, I can change that to a #define. Let me know.

NPAssoc: Indeed, checking with Gemini, 'PROGMEM'  locates a constant in 'PROM/ROM', not RAM. Note that I use 'pgm_read_byte_near()' to access the data for microcontrollers that have different address spaces for ROM & RAM. I made 'dmTableSize' a #define, instead of 'const int'.

PR #4846, opened 2025-08-17
Make version information consistent across update interfaces

Arcitec:
The duplication of logic and the formatting differences between the "OTA Updates" and "Security & Updates" pages made it very difficult to find the exact version details.

With this change, both update-pages now share the same consistent and detailed formatting, making it easy for users to identify which exact version and binary of WLED they've installed.

The version format has also been improved to make it much easier to understand.

---

## OTA Update Page (only reachable via Info -> OTA Update):

Before:

ota-before

After:

ota-after

## Security & Updates Page:

Before (doesn't say anything useful):

upd-before

After:

upd-after

---

The new format is easier for users to understand, and only adds 10 characters. Most builds hover around 75 characters total, and I've verified that the 128 byte buffer would fit extremely long `WLED_RELEASE_NAME` names. Much, much longer than any person would ever use. :)

With these changes, it's finally easy to find the vital information about the installed build. I didn't even know that it was possible to find it in the UI until I started digging around in the code and found that it was hiding on the well-hidden OTA page, hehe. It makes sense to have it on the Security & Updates page too.

I also fixed the braces of two functions that weren't consistent with the coding style of the rest of that file.

## Summary by CodeRabbit

- New Features
  - Improved version display across the app. The About and Update pages now show “Installed version: WLED <version>” for clarity.
  - Added a richer version details section that includes the version, build number, release name, and processor model, presented consistently across relevant UI sections.
  - Replaced a previous generic server error line in the About section with the installed version information for a more informative and user-friendly experience.

### DIFF
--- a/wled00/data/settings_sec.htm
+++ b/wled00/data/settings_sec.htm
@@ -76,7 +76,7 @@
 		A huge thank you to everyone who helped me create WLED!<br><br>
 		(c) 2016-2024 Christian Schwinne <br>
 		<i>Licensed under the <a href="https://github.com/wled-dev/WLED/blob/main/LICENSE" target="_blank">EUPL v1.2 license</a></i><br><br>
-		Server message: <span class="sip"> Response error! </span><hr>
+		Installed version: <span class="sip">WLED ##VERSION##</span><hr>
 		<div id="toast"></div>
 		<button type="button" onclick="B()">Back</button><button type="submit">Save</button>
 	</form>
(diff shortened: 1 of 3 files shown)


Arcitec:
Regarding coderabbit's nitpicks about HTML escaping: I certainly wish that `printSetClassElementHTML()` did escaping of HTML characters when it outputs strings into HTML, but it's pointless and costly to do that processing on a weak CPU where LED effects are more important. It would also waste RAM by having to make a new string buffer for the converted string. So I escape quotes manually via `\"` which is a quick and easy 2-byte solution. Ignore the silly rabbit.

Arcitec:
@netmindz Oh I see you added it to 0.15.2. I'll contribute a PR for that too since I already have a backport locally.

Edit: #4849

Issue #4848, opened 2025-08-18
0.16.0 alpha Boot looping issue on ESP32


NotAmphetamine:
### What happened?

After flashing the ESP32 with the newest available 0.16.0 nighty binary the log console shows 

> [16:56:07]ets Jul 29 2019 12:21:46
> [16:56:07]
> [16:56:07]rst:0xc (SW_CPU_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
> [16:56:07]configsip: 0, SPIWP:0xee
> [16:56:07]clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
> [16:56:07]mode:DIO, clock div:2
> [16:56:07]load:0x3fff0030,len:1184
> [16:56:07]load:0x40078000,len:13132
> [16:56:07]load:0x40080400,len:3036
> [16:56:07]entry 0x400805e4
> [16:56:08]Ada

which repeats forever.

The used board is sourced from Aliexpress and appears to be ESP-32D Wifi+BT/N4XX with the following chip Info

>  - Chip Family: ESP32
>  - Chip Model: ESP32-D0WDQ5-V3 (revision 3)
>  - Number of Cores: 2
>  - Max CPU Frequency: 240MHz
>  - Has Bluetooth: YES
>  - Has Embedded Flash: NO
>  - Has Factory-Calibrated ADC: YES
> Uploading stub...
> Running stub...
> Stub running...
> Changing baud rate to 460800
> Changed.
>  - Flash Size: 4MB
>  - Flash Mode: dout
>  - Flash Frequency: 80MHz

The newest pre-release version ([link](https://github.com/wled/WLED/releases/tag/nightly)) has this issue, using an older version ([link](https://github.com/wled/WLED/actions/runs/16813565752)) fixes it.

### To Reproduce Bug

In order to reproduce this issue you just need to flash the aforementioned software on to the ESP (I tried EspTools in the macOS Terminal via Python3 and the ESP flasher dev MacOS) and observer the logs which will repeat indefinitely. Clearing the Flash or factory resetting via EspTools did not influence this issue in any way.

### Expected Behavior

The logs in the flasher window will simply show the segment you can find below and loop this forever until the serial port is closed. No WLED-AP Network appears and no output to LEDs is observed. When using the [web installer](https://wled-install.github.io/) nothing apart from showing an endless preparing installation screen.

### Install Method

[...]

PR #4852, opened 2025-08-19
Fix bootloop if config is reset

willmmiles:
A couple of small fixes for #4793:

- If the config was missing, the device will bootloop, because `validConfig()` returns false every time.  Fix this by disabling the reboot in `resetConfig()` if we already had no config.
- Fix a minor bug on ESP8266 with saving the action state.

This is the most likely cause of #4848 - once an invalid config is reset, the device is stuck in a bootloop.

## Summary by CodeRabbit

* **Bug Fixes**
  * Configuration reset now verifies a config file exists before proceeding, preventing unnecessary backup steps and avoiding unintended reboots when no file is present.

* **Improvements**
  * ESP8266 devices: Boot-loop recovery better preserves recovery state across restarts, reducing repeated boot loops.
  * Backups now validate configuration files before creating backups, avoiding backups of invalid files.

* **Chores**
  * Minor internal comments added (no user action required).

### DIFF
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -787,13 +787,16 @@ bool verifyConfig() {
 }
 
 // rename config file and reboot
+// if the file doesn't exist, such as after a reset, do nothing
 void resetConfig() {
-  DEBUG_PRINTLN(F("Reset config"));
-  char backupname[32];
-  strcpy(backupname, s_cfg_json);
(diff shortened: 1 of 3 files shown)


DedeHai (approved):
All good fixes, thanks for looking into the issue.

netmindz:
Don't forget to cherry pick to 0_15_x too (if / when we back port this feature)

Issue #4782, opened 2025-07-18
Guru Meditation Error: Core  1 panic'ed if pin not available


maxx-ukoo:
### What happened?

I have board with ethernet. Ethernet chip use pins 16/17 for communication.
It looks like by default WLED use pin 16 as data pin for W28xx strips.
As result W28xx object not initialized properly and app crash on attempt of control leds. This lead to AP not available and it is not possible to fix config without full binary reflash.

### To Reproduce Bug

Flash ethernet firmware on ethernet supported board with pin 16/17 as data pins and configure ethernet.

### Expected Behavior

If data pin not available WLED shouldn't run effects on led object

### Install Method

Self-Compiled

### What version of WLED?

2412040 v4 eth build from https://github.com/wled/WLED/compare/0_15_x...maxx-ukoo:WLED:mk_0_15_0_RTL8201?expand=1 branch

### Which microcontroller/board are you seeing the problem on?

Other

### Relevant log/trace output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

blazoncek:
Please use latest sources and compile from upstream.

Workaround is to define default GPIO for LEDs.

netmindz:
We should not be panic like this, there is code to check if the pin is available and we have checks like  ` if (!_valid) return;` so why are we still seeing the crash? Do you have the full crash log @maxx-ukoo ?

netmindz:
Please can you check if my commit was enough to resolve the issue you were having @maxx-ukoo ?

maxx-ukoo:
@netmindz i broke esp32 prog board, I am waiting for another prog board from ali. I will be able to test in few weeks.

qistoph:
https://github.com/wled/WLED/commit/e5ba97bbe2e077438b5a93ee16406b81f4a8ab75 introduces issues for my `PWM RGB`

I am running a "Wemos D1 mini" with RGB outputs on GPIO `5`, `4`, and `3`. Updating my firmware with `main`/`nightly` today the output didn't work anymore. Using bisect I found this commit to be the issue.

```
git bisect start
# status: waiting for both good and bad commits
# bad: [1fb9eb771e2cb6f9163f2ce695acb3c1bedd8249] Merge pull request #4856 from wled/copilot/fix-4855
git bisect bad 1fb9eb771e2cb6f9163f2ce695acb3c1bedd8249
# status: waiting for good commit(s), bad commit known
[...]

PR #4878, opened 2025-08-28
Fix operation of length 1 strips, such as PWM LEDs

willmmiles:
Revert e5ba97b because it disables all length 1 segments, commonly used for PWM outputs.  Segments of length 1 are valid and expected under many conditions.

This fixes the output issues reported by @naster_mick on Discord and @qistoph in #4782.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed blending behavior for single-pixel segments, ensuring they are included in standard blending instead of being skipped.
  * Ensures 1x1 segments use fade-based blending for consistent visual transitions.
  * Improves visual consistency and smoother fades across all segment sizes.
* **Chores**
  * No changes to public APIs.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1340,11 +1340,6 @@ void WS2812FX::blendSegment(const Segment &topSegment) const {
   uint8_t       opacity    = topSegment.currentBri(); // returns transitioned opacity for style FADE
   uint8_t       cct        = topSegment.currentCCT();
 
-  if (length == 1) {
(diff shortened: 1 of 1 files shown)


DedeHai (approved):

netmindz:
If we just revert, will we not then see a crash back during blend?

DedeHai:
yes, but that needs a different fix.

netmindz:
Sorry about this, my bad. I'll look at the proper fix of the panic issue

PR #4889, opened 2025-08-30
Stencil blendingmode and some speed optimisations

DedeHai:
as mentioned in https://github.com/wled/WLED/issues/4550 there is no easy way to overlay text with the new blending modes. this addition fixes that.

I also made some slight optimisations to `blendSegment()` and found that the compiler is very unpredictable. I assume that is mostly due to pipelining and core register use. In general, replacing bit-field accessed variables with an alias or a pre-computed value is faster, but here it is not always the case. I tried different variations and this was the best performing one. Replacing any other repeated calculations in `blendSegment()` does not improve performance.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1292,9 +1292,6 @@ void WS2812FX::service() {
 }
 
 // https://en.wikipedia.org/wiki/Blend_modes but using a for top layer & b for bottom layer
-static uint8_t _top       (uint8_t a, uint8_t b) { return a; }
-static uint8_t _bottom    (uint8_t a, uint8_t b) { return b; }
-static uint8_t _add       (uint8_t a, uint8_t b) { unsigned t = a + b; return t > 255 ? 255 : t; }
 static uint8_t _subtract  (uint8_t a, uint8_t b) { return b > a ? (b - a) : 0; }
 static uint8_t _difference(uint8_t a, uint8_t b) { return b > a ? (b - a) : (a - b); }
 static uint8_t _average   (uint8_t a, uint8_t b) { return (a + b) >> 1; }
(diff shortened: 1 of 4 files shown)


DedeHai:
A few more comments:
Unfortunately the compact function pointer array had to go: while nice, compact and to my surprise still fast it did not allow blending to operate on a full color, which is necessary to determine if the color is equal to the background in the new mode. I made the new blend mode be number 32 so it is out of the way in case someone finds a smart way to move back to a function array, all my attempts at that failed: it either increased flash size or made it slow. The version I settled on here is faster than the original but it increases flash use due to inlining.

The use of CRGBW instead of uint32_t reduces flash size and increases speed, although how exactly eludes me: the two aliases tC / bC don't do anything but make code more compact to read and came from a version where I passed uint32_t and then cast that: if instead of aliases the arguments are used directly, codesize increases by 250bytes, no idea why having aliases makes the compiler happy.

one more peculiarity: when I did initial tests in a different branch (heap_handling), using "O2" optimization flag on `blendSegment()` the code size increased by about 1.5k, the speed gain was a few percent in FPS. If I do the same on the version of this PR, flash size increases by over 10k or even 20k on C3! Is there any simple explanation for that @willmmiles ?

willmmiles:
> one more peculiarity: when I did initial tests in a different branch (heap_handling), using "O2" optimization flag on `blendSegment()` the code size increased by about 1.5k, the speed gain was a few percent in FPS. If I do the same on the version of this PR, flash size increases by over 10k or even 20k on C3! Is there any simple explanation for that @willmmiles ?

IIRC, often big changes in code size are caused by loop unrolling.

wled00/FX_fcn.cpp
```diff
@@ -1317,23 +1314,40 @@ static uint8_t _softlight (uint8_t a, uint8_t b) { return (b * b * (255 - 2 * a)
...
-  const auto func  = funcs[blendMode]; // blendMode % (sizeof(funcs) / sizeof(FuncType))
-  const auto blend = [&](uint32_t top, uint32_t bottom){ return RGBW32(func(R(top),R(bottom)), func(G(top),G(bottom)), func(B(top),B(bottom)), func(W(top),W(bottom))); };
+static uint32_t segblend(CRGBW tcol, CRGBW bcol, CRGBW bgcol, uint8_t blendMode) {
+  CRGBW tC(tcol); CRGBW bC(bcol); CRGBW c; // note: using aliases shrinks code size for some weird compiler reasons, no speed difference in tests
+                                           // note2: using CRGBW instead of uint32_t improves speed as well as code size
+  switch (blendMode) {
+    case 0:  return tcol.color32;                 // Top
+    case 1:  return bcol.color32;                 // Bottom
+    case 2:  return color_add(tcol, bcol, false); // Add
+    case 3:  c.r = _subtract(tC.r, bC.r);   c.g = _subtract(tC.g, bC.g);   c.b = _subtract(tC.b, bC.b);   c.w = _subtract(tC.w, bC.w);   break; // Subtract
```
willmmiles: A bit of a nitpick, but I'd recommend using an inline metafunction or a macro to do the expansion for each channel.

DedeHai: that would clean up the code significantly, I will check how to do that here.

DedeHai: ok done.

DedeHai:
here is the new blending mode in action: Black Hole FX with blur enabled over a background:
image

DedeHai:
changed to draft, this PR needs to wait  until WIP on transparency is done.

DedeHai:
@coderabbitai review

wled00/FX_fcn.cpp
```diff
@@ -1361,24 +1361,41 @@ static uint8_t _softlight (uint8_t a, uint8_t b) { return (b * b * (255 - 2 * a)
...
+    // default: use function pointer from array
+    const auto func = funcs[blendMode];
+    return RGBW32(func(R(t),R(b)), func(G(t),G(b)), func(B(t),B(b)), func(W(t),W(b)));
+  };
 
   const int     length     = topSegment.length();     // physical segment length (counts all pixels in 2D segment)
   const int     width      = topSegment.width();
   const int     height     = topSegment.height();
+  const uint32_t bgColor   = topSegment.colors[1]; // background color for blend modes that need it (e.g. stencil)
   const auto    XY         = [](int x, int y){ return x + y*Segment::maxWidth; };
```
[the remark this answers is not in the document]
DedeHai: this was an intentional design choice to not introduce new parameters and overhead and keep code simple. it is up to the user to not use a background in combination with stencil mode.

wled00/json.cpp
```diff
@@ -307,9 +307,7 @@ static bool deserializeSegment(JsonObject elem, byte it, byte presetId = 0)
   seg.check2 = getBoolVal(elem["o2"], seg.check2);
   seg.check3 = getBoolVal(elem["o3"], seg.check3);
 
-  uint8_t blend = seg.blendMode;
-  getVal(elem["bm"], blend, 0, 15); // we can't pass reference to bitfield
-  seg.blendMode = constrain(blend, 0, 15);
+  getVal(elem["bm"], seg.blendMode);
```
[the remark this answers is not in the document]
DedeHai: this is intentional: if a blend mode is not supported it should not be truncated but default to a safe one, which is done in the segment blending function.

softhack007:
@coderabbitai re-review all.

wled00/FX_fcn.cpp
```diff
@@ -1361,24 +1361,41 @@ static uint8_t _softlight (uint8_t a, uint8_t b) { return (b * b * (255 - 2 * a)
...
+    }
+    // default: use function pointer from array
+    const auto func = funcs[blendMode];
+    return RGBW32(func(R(t),R(b)), func(G(t),G(b)), func(B(t),B(b)), func(W(t),W(b)));
+  };
 
   const int     length     = topSegment.length();     // physical segment length (counts all pixels in 2D segment)
   const int     width      = topSegment.width();
   const int     height     = topSegment.height();
+  const uint32_t bgColor   = topSegment.colors[1]; // background color for blend modes that need it (e.g. stencil)
```
DedeHai: TODO: remove unused variable

DedeHai:
@softhack007 thanks for reviewing. I am not sure those FPS improvements are still accurate with the latest implementation but I did check the blending function and it is a solid 10% faster - since the rendering "bottle neck" currently is the "pass pixels to the bus" (in my test: rendering FX: 1-2m, rendering two layers to main buffer (this PR) 3ms, rendering main buffer to two buses: 6ms 
when adding several layers, there is an improvement, using a single layer it is almost insignificant when looking at the full render pipeline, i.e. it shaves off about 0.4us of a 10ms frame (rough numbers of my test cases)

PR #4892, opened 2025-08-31
Add support for bq2589x battery charger (Usermod v2)

memExpert:
## Summary:

Added new usermod v2: usermod_v2_bq2589x for supporting bq2589x family chips (BQ25890/BQ25892/BQ25895).  
Uses the memeexpert/bq2589x library (added as a dependency).  
Implements battery status display: voltage, current, temperature, charge level, and power source in WLED JSON API and web interface.  
Support for configuring pins, polling interval, max charge current, and temperature via usermod settings.  
Added methods for reading battery parameters and controlling OTG.  
Added documentation (readme.md) with installation and usage instructions.

## Changes:

- New module: usermod_v2_bq2589x
- Added library.json with memeexpert/bq2589x dependency. Updated constants and PinManager for new usermod support.
- Added configuration template and usage examples.

## Testing:

- Verified build and operation on ESP32-C3.
- Web interface displays battery parameters.
- Settings are correctly applied via Usermod Settings. Documentation and configuration examples are available in readme.md.

If you find any inaccuracies or want me to add/change any logic, let me know. It seems to me that this user will be useful to people who want to use WLED on batteries.

### DIFF
--- a/usermods/usermod_v2_bq2589x/readme.md
+++ b/usermods/usermod_v2_bq2589x/readme.md
@@ -1,0 +1,75 @@
+# bq2589x
+
+The BQ25895 is a highly-integrated 5-A switch-mode battery charge management
+and system power path management device for single cell Li-Ion and Li-polymer
+battery. The devices support high input voltage fast charging.
+
+Version 1.0
+
+
+## Installation 
+
+Use this code for your platformio.ini or platformio_override.ini
+
+```ini
+[env:esp32bq2589x]
+extends = env:esp32c3dev
+custom_usermods = usermod_v2_bq2589x # add "custom_usermds" to the extensible environment if it is not there
+build_flags = ${common.build_flags_esp32}
+              -DUSERMOD_V2_BQ2589X
+lib_deps = ${esp32c3.lib_deps}
+           memeexpert/bq2589x@^1.0.3
+```
+
+
+## Compilation
+
+You can change default parameters by add build flags like this:
+
+```ini
+-D BQ2589X_OTG_PIN=12                    # for change OTG default pin
+-D BQ2589X_NCE_PIN=4                     # disable charger
+-D BQ2589X_INT_PIN=2                     # input pin for interrupt by bq
+-D BQ2589X_SYS_PIN=1                     # input pin for reading sys voltage by ADC of your controller.
+-D BQ2589X_DEFAULT_REQUEST_INTERVAL=5000 # changing request interval
+-D BQ2589X_DEFAULT_BAT_MAX_CC=2100       # changing max charger current (main charge stage)
+-D BQ2589X_MAX_TEMPERATURE=80            # change maximal temperatures
+-DWLED_DEBUG                             # for enable debug print
+
+```
+
+
+## Usage
+
+After compiling and upload your code you can see some params from your battery
+by click "info" button in UI.  
+You can change battery max/min voltages, request frequency, pins and charger
+charger current in usermods config.
+
+
+## Tips
+
+Be sure to use a capacitor on the OTG line and the host controller power supply
+to avoid "blinking" and voltage drops in the host controller supply.
+
+
+## Documentation
+
+- [Original TI datasheet](https://www.ti.com/lit/ds/symlink/bq25895.pdf).
+- [Community comments](https://gist.github.com/somebox/b6cd88731a4b4acb6d2a2f2a00b18c7f)
+
+
+## Developers
+
+Maintainer: [memeexpert](https://github.com/memExpert).
+Write me an email (hancharmaksimq@gmail.com or pottermax2000@gmail.com) if you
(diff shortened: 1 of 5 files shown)


usermods/usermod_v2_bq2589x/readme.md
```diff
@@ -0,0 +1,75 @@
...
+## Installation 
+
+Use this code for your platformio.ini or platformio_override.ini
+
+```ini
+[env:esp32bq2589x]
+extends = env:esp32c3dev
+custom_usermods = usermod_v2_bq2589x # add "custom_usermds" to the extensible environment if it is not there
+build_flags = ${common.build_flags_esp32}
+              -DUSERMOD_V2_BQ2589X
```
netmindz: Do you actually use USERMOD_V2_BQ2589X flag anymore, the new usermod system should make that redundant in favor of the custom_usermods

wled00/pin_manager.h
```diff
@@ -68,7 +68,8 @@ enum struct PinOwner : uint8_t {
   UM_LDR_DUSK_DAWN     = USERMOD_ID_LDR_DUSK_DAWN,      // 0x2B // Usermod "usermod_LDR_Dusk_Dawn_v2.h"
   UM_MAX17048          = USERMOD_ID_MAX17048,           // 0x2F // Usermod "usermod_max17048.h"
   UM_BME68X            = USERMOD_ID_BME68X,             // 0x31 // Usermod "usermod_bme68x.h -- Uses "standard" HW_I2C pins
-  UM_PIXELS_DICE_TRAY  = USERMOD_ID_PIXELS_DICE_TRAY    // 0x35 // Usermod "pixels_dice_tray.h" -- Needs compile time specified 6 pins for display including SPI.
+  UM_PIXELS_DICE_TRAY  = USERMOD_ID_PIXELS_DICE_TRAY,   // 0x35 // Usermod "pixels_dice_tray.h" -- Needs compile time specified 6 pins for display including SPI.
+  UM_BQ2589X           = USERMOD_ID_BQ2589X
```
netmindz: include the comment, as per the other entries

usermods/usermod_v2_bq2589x/usermod_v2_bq2589x.cpp
```diff
@@ -0,0 +1,686 @@
...
+#ifndef BQ2589X_MAX_TEMPERATURE
+  #define BQ2589X_MAX_TEMPERATURE 80
+#endif
+
+
+// this function calls when charger generate event 
+// you can do it all that you want
+static void IRAM_ATTR intPinUp() {
+  // just a dummy for now
+}
```
netmindz: Not sure about your usermod adding a global function like this

usermods/usermod_v2_bq2589x/readme.md
memExpert: Ok, i will remove it.

wled00/pin_manager.h
memExpert: I'm sorry, my bad. Just forgot about it.

usermods/usermod_v2_bq2589x/usermod_v2_bq2589x.cpp
memExpert: I apologize, I do not quite understand what you are talking about. If i can not register my interruptions, I can remove it. But I saw in usermod_v2_rotary_encoder_ui_alt registration of interruption. If you are about the fact that I do not use it, then I just wanted to clearly indicate to the user that he can add some logic there if he need it. If you think this is unnecessary, I will remove that.

DedeHai:
thanks for contributing. I have a few questions:
- are there any ready-made boards for this IC that are available to users? otherwise this is very niche since there are other controllers that do have arduino support AFAIK.
- since the UM is called bq2589x: do all of these chips use an identical interface and did you check that?
- as you have forked the library: did you confirm all settings in that library are correct and verify with the datasheet? the original library seems a bit "how you doin" not even with a readme.

memExpert:
>     * are there any ready-made boards for this IC that are available to users? otherwise this is very niche since there are other controllers that do have arduino support AFAIK.

I using board from CG: https://aliexpress.com/item/1005005660079407.html?sku_id=12000033925273913
 
>     * since the UM is called bq2589x: do all of these chips use an identical interface and did you check that? 
>     * as you have forked the library: did you confirm all settings in that library are correct and verify with the datasheet? the original library seems a bit "how you doin" not even with a readme.

Thanks to your remark, I noticed that the BQ25892 uses a different address (6B instead of 6A). I didn’t check this in the forked library. The rest of the interfaces are the same.
I agree that the original library isn’t in great shape, but I was able to get correct data from the bq25895 chip using it as a base. If it’s helpful, I can improve the library by adding documentation and performing full testing, then provide a detailed report.

I’m attaching a screenshot as confirmation.

bq25895_in_use

PR #4895, opened 2025-08-31
speed optimisations, fix for restoreColorLossy, code cleanup

DedeHai:
A few tweaks to increase rendering speed just a bit more. On ESP32 this comes at a cost of ~500bytes of flash use, on C3 and ESP8266 it saves ~300bytes. The flash and speed numbers in the text below are for ESP32.

Edit: with the latest commit, code now also shrinks for ESP32 by 400 bytes.

- Created WLED_O2_ATTR alias for `__attribute__((optimize("O2")))
- Moved `fastColorScale()` function from PS to colors.h, changed it to use 32bit math and put it to good use wherever scaling accuracy does not matter too much, i.e. in `Segment::fadeToBlackBy()` and `Segment::blur()`, making these function much faster. I also made it an inline function as the function call overhead is using more flash than the function itself. Since no byte access is used in the function, this is safe for all color operations, even if in 32bit access RAM or at least I'd exptect so, did not test that but it will show soon enogh ;)

- Added new `_NPBbri` variable to bus: need to track total brightness scaling (new ABL code) for NPB buffers or `BusDigital::getPixelColor()` will return incorrect color if ABL is engaged. Function is currently not used but if we remove the global `_pixels[]` buffer, the Copy FX will need it. I did not test this particular code change, someone please check if the logic holds (the math should be correct, I checked that).

- `color_blend()`:adding WLED_O2_ATTR gives a rendering speed improvement of about 1% at the cost of 100 bytes of flash.
- `Segment::setPixelColor()`:  removed IRAM_ATTR, using WLED_O2_ATTR instead: significant FPS improvement of 4% for the cost of 350bytes of flash.
- `Segment::getPixelColor()` removed IRAM_ATTR, using WLED_O2_ATTR instead: faster in tests and even saves a bit of flash.
- `isPixelClipped()` 1D & 2D version: removing IRAM_ATTR will highly likely just inline the function as it is called only once. A small flash use reduction seems to confirm that. This is faster than forcing a function call to IRAM. Since its only used during transitions I did not measure the speed impact and I would not expect it to be huge. 
- `Get/setPixelColorX`: instead of casting `vWidth/vHeight` to int, cast X and Y to unsigned and save the negative check. Makes code size smaller so I assume it is faster too. I also thought about making X/Y coordinated unsigned in general to get more consistency throughout the code but decided I will leave that to a futuer endeavour.

**Code cleanup:**
- Changed parameter order in settings_leds.htm to match cpp file as it was confusing, no functional change.
- Bugfix in `adjust_color()` and fixed indentation
- Removed IRAM_ATTR_YN from unused gamma functions (just in case that attribute forces the compiler to put them there)

In summary: this PR and in combination with https://github.com/wled/WLED/pull/4889 there was a significant improvement in FPS in my test. From 68FPS to 78FPS  (4 Layers, 32x32 on ESP32), each PR contributes roughly half of that improvement. On a more general test on the C3 the gain was less significant but still visible (+2FPS at 50FPS just this PR alone).

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate colors when brightness limiting (ABL) is active.
  * Corrected color adjustment to avoid unintended shifts.
  * Improved bounds checking to prevent artifacts from out-of-range pixels.
* **Refactor**
  * Faster, more robust blur and color-scaling for smoother visuals and improved rendering performance.
  * Reduced overhead in particle and rendering paths for more consistent effects.
* **Style**
  * Minor formatting cleanup on the LED settings page (no functional changes).

### DIFF
--- a/wled00/colors.cpp
+++ b/wled00/colors.cpp
@@ -48,7 +48,9 @@ uint32_t WLED_O2_ATTR color_add(uint32_t c1, uint32_t c2, bool preserveCR) //121
       wg =  (wg * scale)       & ~TWO_CHANNEL_MASK;
     } else wg <<= 8; //shift white and green back to correct position
   } else {
-    // branchless per-channel saturation to 255 (extract 9th bit, subtract 1 if it is set, mask with 0xFF)
+    // branchless per-channel saturation to 255 (extract 9th bit, subtract 1 if it is set, mask with 0xFF, input is 0xFF+0xFF=0x1EF max)
+    // example with overflow: input: 0x01EF01EF -> (0x0100100 - 0x00010001) = 0x00FF00FF -> input|0x00FF00FF = 0x00FF00FF (saturate)
+    // example without overflow: input: 0x007F007F -> (0x00000000 - 0x00000000) = 0x00000000 -> input|0x00000000 = input  (no change)
     rb |= ((rb & 0x01000100) - ((rb >> 8) & 0x00010001)) & 0x00FF00FF;
     wg |= ((wg & 0x01000100) - ((wg >> 8) & 0x00010001)) & 0x00FF00FF;
     wg <<= 8; // restore WG position


DedeHai:
> > wled00/FX_2Dfcn.cpp (1)> `149-184`: **ODR/behavior mismatch with FX.h inline stub for isPixelXYClipped()**

> > > // FX.h (inside class Segment)
> > > - inline bool isPixelXYClipped(int x, int y) const { return isPixelClipped(x); }
> > > + bool isPixelXYClipped(int x, int y) const;

@coderabbitai there is an #ifdef #else that excludes one or the other.

DedeHai:
I am guessing it's hard to review this properly as there are many details I worked hours on an they are a bit hard to grasp without spending much time studying the code in detail. I tested each single change for speed and corect operations. While I cannot guarantee there are no bugs introduced this is generally safe to merge.

wled00/FX_2Dfcn.cpp
```diff
@@ -186,7 +186,7 @@ bool IRAM_ATTR_YN Segment::isPixelXYClipped(int x, int y) const {
 void IRAM_ATTR_YN Segment::setPixelColorXY(int x, int y, uint32_t col) const
 {
   if (!isActive()) return; // not active
-  if (x >= (int)vWidth() || y >= (int)vHeight() || x < 0 || y < 0) return;  // if pixel would fall out of virtual segment just exit
+  if ((unsigned)x >= vWidth() || (unsigned)y >= vHeight()) return;  // if pixel would fall out of virtual segment just exit
```
blazoncek, line 189: Why don't you go all the way and change every `int` into `unsigned`?

It would save you a few manual conversions like this and the code would be cleaner.

However, do check if any of the effects produce negative index/coordinates to avoid other errors.

wled00/colors.cpp
```diff
@@ -25,39 +25,35 @@ uint32_t IRAM_ATTR color_blend(uint32_t color1, uint32_t color2, uint8_t blend)
...
     } else wg <<= 8; //shift white and green back to correct position
-    return rb | wg;
   } else {
-    r = r > 255 ? 255 : r;
-    g = g > 255 ? 255 : g;
-    b = b > 255 ? 255 : b;
-    w = w > 255 ? 255 : w;
-    return RGBW32(r,g,b,w);
+    // branchless per-channel saturation to 255 (extract 9th bit, subtract 1 if it is set, mask with 0xFF)
+    rb |= ((rb & 0x01000100) - ((rb >> 8) & 0x00010001)) & 0x00FF00FF;
```
blazoncek, line 54: Clever.

wled00/FX_2Dfcn.cpp
DedeHai: "Get/setPixelColorX: instead of casting vWidth/vHeight to int, cast X and Y to unsigned and save the negative check. Makes code size smaller so I assume it is faster too. I also thought about making X/Y coordinated unsigned in general to get more consistency throughout the code but decided I will leave that to a futuer endeavour."

wled00/colors.cpp
DedeHai: Can't take all the credit, I had a little AI help ;)

DedeHai:
@blazoncek thanks for the review. can you please point out where the code is unclear so I can add comments?

wled00/FX_2Dfcn.cpp
blazoncek: `int` was introduced to combat issues @softhack007 noticed with a few SR effects (possibly others too) that produced negative indexes.
Originally indexes were `uint16_t`.

DedeHai: 1D uses int for "vitual strips" IIRC, I wanted to convert 2D to unsigned before but struggled with "ambiguous" function calls I was not sure how to resolve so I abandoned it.

blazoncek: > 1D uses int for "vitual strips"

not `int` actually. it encodes virtual strip number into upper 16 bits. which means it *can* produce incorrect results.

wled00/colors.cpp
blazoncek: This was the place I spent couple of seconds to decode. It would be helpful to either have an example or description that 256-1 [`(1<<8)-1` or `256-(256>>8)`] is 255 which saturates (`|=`) the output.

wled00/FX_2Dfcn.cpp
softhack007: > `int` was introduced to combat issues @softhack007 noticed with a few SR effects (possibly others too) that produced negative indexes. Originally indexes were `uint16_t`.

Hi, the type can stay "int" - what Blaz refers to is that some effects were stuffing out-of-range numbers (including negatives) into sPC and gPC. This was not just happening with old SR effects. 

``(unsigned)x >= vWidth()`` is a trick that has exactly the same behaviour as ``x<0 || x >= vWidth()" when x is _signed_ - it simply saves one CPU instruction. The trick does not work when x is unsigned.

Issue #4655, opened 2025-04-22
ESP-NOW not functioning on latest build


A6uh:
### What happened?

Wizmote not being detected on latest build. Manually inputting the known Wizmote address also doesn’t yield any results when pressing buttons.

### To Reproduce Bug

Flash WLED 0.16.0-alpha (build 2412040) to ESP-32
Attempt to pair ESP-NOW remote

### Expected Behavior

It should be detecting the ESP-NOW remote address, and after pairing, should respond to button input.

### Install Method

Binary from WLED.me

### What version of WLED?

WLED 0.16.0-alpha (build 2412040)

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell

```

### Anything else?

Reverting to WLED 0.15 release resolves the issue.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DedeHai:
works perfectly fine. make sure you have wifi sleep disabled.

A6uh:
Could you double check this with a freshly flashed ESP32? It appears that ESP-NOW only doesn't work if you don't connect the ESP to WiFi. Like connect to the WLED AP, and try to setup the remote. If you connect the ESP to WiFi, ESP-NOW functions correctly.

Unfortunately for the project I'm trying to use it for, it won't be anywhere near a WiFi network to connect to.

A6uh:
Here are some logs I collected. ESP-NOW was enabled and wifi sleep was disabled by default. When I connect to a wifi network is on line 446.

https://pastebin.com/jUeNbRmz

DedeHai:
interesting find, not sure that is a bug or intentional or a HW limitation. Did this ever work in older versions?

blazoncek:
@A6uh how far have you researched ESP-NOW? It will help you understand where the problem is after you do that.

Brief recap: As WiFi and ESP-NOW share same circuitry they can't work independently. If you want uninterrupted ESP-NOW packets between two ESP devices they have to:

1. be connected to the same WiFi channel
2. must have WiFi disabled and still use same channel for ESP-NOW
3. be in AP mode on the same channel

Once these are fulfilled, ESP-NOW will work as expected. WLED is designed to operate either in 1) or in 3) mode.

I've been working on correct ESP-NOW implementation for over a year now and best take is available in my fork but it may still have quirks.

[...]

PR #4495, opened 2025-01-16
Add Wired DMX Input support

netmindz:
Port the DMX Input feature from WLED-MM

PR #4903, opened 2025-09-02
remove early return from initconnection()

DedeHai:
this line was added in https://github.com/wled/WLED/commit/aed03cd03baca7100f7eef713905bc3bae6b55ce as a part of https://github.com/wled/WLED/pull/4495 to fix a bug which no longer seems relevant. 

@arneboe and @netmindz  please test and make sure it does not bring back that bug.

fixes https://github.com/wled/WLED/issues/4655 (tested and confirmed)

## Summary by CodeRabbit

- New Features
  - Additional services now initialize when the device runs in Access Point mode without prior Wi‑Fi setup, enabling broader functionality out of the box.
- Bug Fixes
  - Resolved an issue where certain features did not start in AP mode when no Wi‑Fi was configured, improving reliability during first-time setup and standalone use.
- Chores
  - Internal initialization flow streamlined to ensure consistent startup behavior across connection states.

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -649,7 +649,6 @@ void WLED::initConnection()
   if (!WLED_WIFI_CONFIGURED) {
     DEBUG_PRINTLN(F("No connection configured."));
     if (!apActive) initAP();        // instantly go to ap mode
-    return;
   } else if (!apActive) {
     if (apBehavior == AP_BEHAVIOR_ALWAYS) {
       DEBUG_PRINTLN(F("Access point ALWAYS enabled."));


arneboe:
The missing attributes where later fixed upstream in esp_dmx.
The hack was removed in https://github.com/wled/WLED/commit/68e9d701dea5620a4daa503ee8672e8707e7cbe7
But apparently I did not remove the early return. Sorry about that :(

DedeHai:
@netmindz did you get a chance to test this?

Arcitec:
@DedeHai @netmindz @blazoncek @willmmiles Hey, since it's on the same topic and touches the same code - I've been waiting to hear more about Blaz's network rewrite before porting my ESP-NOW and WiFi fixes. There's a lot of bugs and a complete spaghetti nest in the current networking, and it's very easy to make WLED get into a state where both WiFi and Ethernet are activated at the same time (a very old bug). The ESP-NOW code also doesn't get initialized when Ethernet-only is being used. I fixed that too. But was going to hold the porting work until the network rewrite was done.

But I'm wondering, should I just submit it anyway, and then Blaz can see how I fixed it, and implement the same fixes in his rewrite? What do you think? Seems like there's more and more interest in people asking for ESP-NOW fixes.

PS: My bug fixes are for the 0.15.x branch. Hopefully `main` networking is still very similar, then I can safely and easily port it to main too. But we should start with 0.15.x, which is where it's stable and verified, and then review those patches before porting it to main.

DedeHai:
You can do a draft PR any time to have a discussuion but please do so for main, 0.15 can be backported once tested.

Arcitec:
@DedeHai Alright, I'll start looking at the main code. I hope it's not already too different from 0.15, because untangling the old mess was already heavy work. I've never seen crazier networking code (like 5 different files all modifying the same messy state without any care for synchronization). But I'll get started next week on the `main` code.

netmindz:
Tried to test today, but my test setup for the existing code wasn't working so I need to build another hardware setup to test

PR #4905, opened 2025-09-03
add Rainbow Shimmer effect. Just like Rainbow Cycle but with a white …

Charming-Lime:
Adds a new effect called Rainbow Shimmer effect. It takes the Rainbow Cycle effect but adds a  white flash that runs across the segment. Adjustable frequency of flash, direction, speed of flash, and size.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -544,6 +544,45 @@ uint16_t mode_rainbow_cycle(void) {
 }
 static const char _data_FX_MODE_RAINBOW_CYCLE[] PROGMEM = "Rainbow@!,Size;;!";
 
+/*
+ * Cycles a rainbow over the entire string of LEDs, with a white flash that goes across it.
+ */
+uint16_t mode_rainbow_shimmer() {
+
+  unsigned counter = (strip.now * ((SEGMENT.speed >> 2) + 2)) & 0xFFFF;
+  counter = counter >> 8;
+
+  uint32_t shimmerSpeed = 100 + (255 - SEGMENT.custom2) * 40; //ranges from 100-10260ms
+  uint32_t shimmerSize = (SEGMENT.custom3 * SEGLEN / 2 >> 5) + 1;
+  uint32_t cycleTime = (255 - SEGMENT.custom1) * 150 + shimmerSpeed;
+
+  uint32_t percCycle = strip.now % cycleTime;
+  float shimmerIndex = (float)percCycle / (float)shimmerSpeed * (SEGLEN + 2*shimmerSize);
+  
+  shimmerIndex -= shimmerSize;
+
+  // reverse direction of shimmer
+  if(!SEGMENT.check1) {
+    shimmerIndex = (float)SEGLEN - shimmerIndex;
+  }
+
+
+  for (unsigned i = 0; i < SEGLEN; i++) {
+    // Standard rainbow effect.
+    uint8_t index = (i * (16 << (SEGMENT.intensity / 29)) / SEGLEN) + counter;
+    SEGMENT.setPixelColor(i, SEGMENT.color_wheel(index));
+
+    //shimmer logic
+    float distFromShimmerCenter = fabsf(shimmerIndex - i);
+    // Only process pixels that are within the shimmer's range.
+    if (distFromShimmerCenter < shimmerSize) {
+      SEGMENT.setPixelColor(i, color_blend(SEGMENT.getPixelColor(i), 0xFFFFFF, (uint8_t)(255 * (1.0f - (distFromShimmerCenter / shimmerSize)))));
+    }
+  }
+
+  return FRAMETIME;
+}
+static const char _data_FX_MODE_RAINBOW_SHIMMER[] PROGMEM = "Rainbow Shimmer@!,Size,Shimmer Frq,Shimmer Speed,Shimmer Length,Reverse;;c1=231,c2=221";
 
 /*
  * Alternating pixels running function.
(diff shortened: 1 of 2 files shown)


DedeHai:
Thanks for contributing.
I like the general idea, however, this is not a good way to do it.
First off, do not use floats.
With 0.16 there are elaborate ways to use layers. If you want to add a shimmer effect, it should do so as an overlay so it can be used with any effect. It aso should use palettes instead of just plain white, the default can still be white when doing that. If shimmer is its own independent effect, it could also be a bit more elaborate, like having different shimmer patterns, I am imagining a gradient wave as the base (as you have) but also adding the possibility to modulate that with a cosine to generate "multiple stripes" or maybe you can come up with some other things that look good.

Charming-Lime:
I understand the concern about floats; it was a deliberate decision to balance the desire for a smooth transition of the "shimmer" without obfuscating the code with a pseudo implementation of float using bit shifts into an arbitrary index that would be remapped to physical indices. In my testing the performance limitation was vastly dominated by the strip bus speed rather than limits on computational power, a 1000 pixel strip was still achieving 40 FPS, which was the same as the unmodified rainbow cycle. If you would prefer the latter implementation I will refactor.

I originally looked at implementing as an overlay, or how effects could be combined to achieve the same result, but I could not find documentation or effects designed for use that way. Considering 0.16 isn't released yet and there is no documentation outside of the comments on the PR, this would appear to be the first effect dependent on the new layer implementation; I admit I am playing catchup here. If the effect is intended to be secondary to a primary effect, adding color palette support is the natural extension of that line of thought. 

On the topic of non-standalone effects, from a user perspective there should be a way in the UI to mark an effect as secondary or describe how effects should be combined with other. Maybe a way to default a primary and secondary overlapping effect? Otherwise, the user might be left wondering the purpose or how to use a secondary effect.

DedeHai:
floats are only available on ESP32 and S3, not on S2, C3 or ESP8266. So no floats unless absolutely necessary or for things that run only once in a while, it also increases code size.
The thing is, we do not want to increase the number of effects unless its really something new/unique. Otherwise, extending an existing effect is ok. Here there is potential for something unique. 
Also there are already a few overlay effects. Scrolling text, sparkle, proably a few others. I would not call them secondary, those also look great on their own, which this one can too if done right.
edit:
 documenting how to do composit/overlay effects is something anyone in the community can do, its not hard to figure out.

Charming-Lime:
As requested, removed float, added Fx and Bg colors. Removed the rainbow section for use as overlay.

DedeHai:
thanks for the update, good progress. 
I have a few points:
- 64bit math is about as bad as float, please read https://github.com/wled/WLED/issues/4206
- if you adjust the speed, there is no need for 64bit
- your numbers in the code are wrong
- remove white spaces
- fix the metadata string (please thoroughly test your code, this is an obvious one)
- you are doing double-inverse in color_blend, which is unnecessary

also you did not add palettes as requested, please do so and we have the basic structure for this effect. What I would later also like to add is: 
- random interval as a slider, the more random, the further it strays from given interval (or something like that)
- possibility to add brightness modulators to change the looks, like making it two smaller shimmers instead of one large one for exmple by overlaying multiple gradients on top of the base gradient, or as mentioned, a cosine wave may do it. or even add perlin noise as an option. 

with palettes added, a user can create a custom palette to only partially apply the shimmer and with modulators it may look a lot better when not using it as an overlay. Please let me know if (and at what point) this gets beyond the time you are willing to invest.

wled00/FX.cpp
```diff
@@ -545,40 +545,45 @@ uint16_t mode_rainbow_cycle(void) {
...
+  uint32_t segmentSize;
+  //gaurd against buffer overflow, should never exceed this based
+  //on 1000 pixles per GPIO and 10 strips
+  if(SEGLEN>10000) {
+    segmentSize = 10000;
+  } else {
+    segmentSize=SEGLEN;
+  }
+  
+  uint32_t shimmerSpeed = 100 + (255 - SEGMENT.speed) * 40; //  [100,10300]
```
DedeHai, line 560: if you change 100 to 200 in this line, no overflow and you do not have to limit the length.

wled00/FX.cpp
```diff
@@ -544,6 +544,46 @@ uint16_t mode_rainbow_cycle(void) {
...
+  int32_t shimmerIndex = (percCycle<<8) / shimmerSpeed * (segmentSize + 2*shimmerSize); // maxamum >124285
+  
+  shimmerIndex -= shimmerSize << 8;
+
+  //change direction unless reverse is checked
+  if(!SEGMENT.check1) {
+    shimmerIndex = ((segmentSize) << 8) - shimmerIndex;
+  }
+
+  for (uint32_t i = 0; i < segmentSize; i++) {
```
DedeHai, line 573: must be unsigned, for some reason that is beyond my knowledge it fails with C3 when using uint32_t

wled00/FX.cpp
```diff
@@ -544,6 +544,46 @@ uint16_t mode_rainbow_cycle(void) {
...
+
+    // slightly dangerous, but shimmer index is bounded to be safe
+    uint32_t distFromShimmerCenter = abs((int32_t)shimmerIndex - ((int32_t)i << 8));
+    if (distFromShimmerCenter < (shimmerSize<<8)) {
+      SEGMENT.setPixelColor(i, SEGMENT.color_from_palette(255-(distFromShimmerCenter / shimmerSize),false,PALETTE_SOLID_WRAP,0));
```
DedeHai, line 578: you forgot the blend

Charming-Lime: No, I didn't. It is up to the user to select a color palette that blends (or contrasts if desired) with the selected background color. Any additional blending between layers will happen using #4658 Enhanced Layering facilities. Adding additional blending into the loop could limit the number of ways the effect could be used.

DedeHai: without the gradient, there are other effects that do almost the same.

Charming-Lime: Then what exactly should the gradient be with?

DedeHai: as it was, plus the suggestions: cosine, sub-gradiens, perlin

Charming-Lime: I wrote this effect to be a simple, lite occasional flair to a relatively plain and simple back round, as every other effect had too much noise, randomness, and was always visible. Adding all of those gradients are well outside the scope of what I want my effect to be. I think we have very different ideas of what this effect should be so I am going to close this and just maintain my own fork. I'm happy to adapt to make my effect more versatile and flexible, but I am not going to make something I don't want.

DedeHai: you misunderstood.

PR #4917, opened 2025-09-08
Bugfix for FX: Tri Fade

blazoncek:
- incorrectly calculated counter and progress

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved precision in the Tricolor Fade effect to prevent intermediate calculation overflow, resulting in smoother color transitions during long-running animations and high counter values.
  * Reduces potential visual artifacts without altering user settings or overall behavior.


blazoncek:
reference: https://github.com/wled/WLED/blame/5732b66c5ae54fb116e78375f9177b2ab9b526e4/wled00/FX.cpp#L1658

PR #4918, opened 2025-09-08
Bugfix for FX: Tri Fade

blazoncek:
- incorrectly calculated counter and progress


blazoncek:
reference: https://github.com/wled/WLED/blame/5732b66c5ae54fb116e78375f9177b2ab9b526e4/wled00/FX.cpp#L1658

Issue #4920, opened 2025-09-10
NOWIFISLEEP build flag not honored


rhkean:
### What happened?

I noticed that the NOWIFISLEEP build flag was not being honored and always set true for esp32's and false for esp8266's.

I'm committing a PR for it

### To Reproduce Bug

* add `DEBUG_PRINTF_P(PSTR("calling WiFi.setSleep(%d);"), !noWifiSleep);` just before the call in the `initConnection` function on `wled.cpp`
* see that `setSleep` is always called with `0`, regardless of the build flag

### Expected Behavior

-D NOWIFISLEEP=0 enables WiFi sleep

### Install Method

Self-Compiled

### What version of WLED?

nightly

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DedeHai:
there is no buildflag called NOWIFISLEEP.

rhkean:
discussion on PR opted to not include the build flag

PR #4921, opened 2025-09-10
issue #4920 add code to respect the NOWIFISLEEP build flag

rhkean:
added check to respect the build flag

## Summary by CodeRabbit

* **New Features**
  * Added optional build-time flag to control Wi‑Fi sleep behavior.
  * When provided, firmware uses the flag’s value; otherwise, platform defaults are applied (no change to standard builds).
  * Default behaviors remain: ESP32 builds default to Wi‑Fi sleep on; others off, unless the flag is set.
  * Applies to all supported boards; runtime settings and saved configurations are unaffected.

### DIFF
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -377,7 +377,9 @@ WLED_GLOBAL int8_t selectedWiFi  _INIT(0);
 WLED_GLOBAL byte apChannel       _INIT(1);                        // 2.4GHz WiFi AP channel (1-13)
 WLED_GLOBAL byte apHide          _INIT(0);                        // hidden AP SSID
 WLED_GLOBAL byte apBehavior      _INIT(AP_BEHAVIOR_BOOT_NO_CONN); // access point opens when no connection after boot by default
-  #ifdef ARDUINO_ARCH_ESP32
+  #ifdef NOWIFISLEEP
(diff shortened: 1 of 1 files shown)


DedeHai:
why do you think there should be such a build flag?

blazoncek:
In any case, this flag is implemented only partially. If it is to be merged it needs to be properly implemented everywhere.

rhkean:
@DedeHai , @blazoncek 

couple things...

1. It found its way into my platformio.ini file somehow.... oh... it's in the QuinLED dig platformio_override.ini files, interesting
2. I'm still working on the Bluetooth BLE mod and disabling sleep crashes WLED when enabling WiFi and BLE, but allowing it to sleep, allows the RTOS to manage the radio properly for time slicing between the 2 functions and it doesn't crash.
3. can you direct me to where else I need to set the `noWifiSleep` variable.  Everywhere that I looked appears to trigger off that initial setting in `wled.h` that I addressed in the PR (except the pixel_dice_tray usermod that addresses it specifically for that mod)

blazoncek:
I really don't se the need to add compile time option for WiFi sleep.
Why?
It is a configuration option that user can change so if you have issues with it enabled you will either need to force it in your usermod `setup()` or some place after the configuration is loaded. Not a place for compile time option.

rhkean:
ok

PR #4928, opened 2025-09-14
Fix LED buffer size calculation

### LINKED ISSUE #4614: Hardware Setup Length can't be saved after 659
### What happened?

When you want to save a number above the 659 in the LED & hardware setup it resets again to 30. 

Tried to set it to 720
Image
After save, its again 30
Image

### To Reproduce Bug

Go to LED & Hardware setup 
In Hardwaresetup LED outputs choose a length greater then 659. 
After save, its set at 30 again

Image

### Expected Behavior

Go to LED & Hardware setup 
In Hardwaresetup LED outputs choose a length greater then 659. 
After save, its saved the number

### Install Method

Binary from WLED.me

### What version of WLED?

Nightly Release 20250326

ESP8266

### Relevant log/trace output

```shell

```

### Anything else?

_No response_

### Code of Conduct

DedeHai:
- cherry picked from @blazoncek fork and fixed merge conflicts.
- added the two mendatory buffers (global and segment, each 32bit) to memory requirement calculation in UI

fixes #4614 "Hardware Setup Length can't be saved after 659"

this fix also nees to be back-ported to 0.15

@willmmiles can you please check this is correct? I am not 100% sure I fixed the conflicts correctly, it did work when testing on ESP32 and ESP8266.

## Summary by CodeRabbit

- New Features
  - LED settings UI now shows updated hardware-aware memory estimates and clearer warning wording; pin validation logic improved.

- Bug Fixes
  - More comprehensive memory checks now account for shared I2S/double-buffer usage to avoid creating outputs that would exceed available memory.
  - Improved compatibility and sizing across ESP8266/ESP32 variants to reduce init failures.

- Refactor
  - Default LED bus setup simplified for more consistent configuration creation and memory reporting.


DedeHai:
@blazoncek I see you pushed an update to your branch. Do I need to cherry pick that as well or is it cosmetic?

DedeHai:
@willmmiles I pulled these changes into my "heap handlint tests" branch. I can get **ESP8266** to work with 800 LEDs still but it is very(!) unstable, frequently refusing to fully load the UI (info->reboot still works), at lucky moments there is enough heap to run a PS effect, but its not really useable. On the positive side: at least it does not lock-up. 
Wit 400 LEDs it works quite well, 500 is still ok-ish, 600 is pushing it but the UI stays responsive most of the time.
What we should consider adding is a reboot after adding new buses if calculated mem-usage is more than like 50%. MM has something like that where a window pops up to recommend a reboot. (CC @netmindz )

without a reboot, heap is fragmented and nothing works right.

blazoncek:
> @blazoncek I see you pushed an update to your branch. Do I need to cherry pick that as well or is it cosmetic?

Hit 2 flies with a single strike. 😉 

BTW add transition buffer to calculation as transitions are enabled by default.

DedeHai:
> Hit 2 flies with a single strike. 😉

lucky strike ;)
 
> BTW add transition buffer to calculation as transitions are enabled by default.

I would argue that those are not strictly required, allocation will fail and transition will not happen. If we want the value to represent "will work without memory issues" I agree. ESP8266 will show the warning at around 220LEDs if transition buffer is added to the calculation, stability issues start appearing above 400 but out of memory for FX buffer in transitions can happen at 200 already, depending on heap fragmentation.

netmindz:
Yeah adding a recommendation to reboot sounds like a good addition @DedeHai

wled00/data/settings_leds.htm
```diff
@@ -266,7 +266,7 @@
 				// is the field a LED type?
 				var n = s.name.substring(2);
 				var t = parseInt(s.value);
-				memu += getMem(t, parseInt(n)); // calc memory
+				memu += getMem(t, n); // calc memory
```
blazoncek, line 270: `n` is used both as character and a number in `getMem()`.
I prefer to have number instead of character.

wled00/data/settings_leds.htm
```diff
@@ -304,7 +304,7 @@
 			var nList = d.Sf.querySelectorAll("#mLC input[name^=L]");
 			nList.forEach((LC,i)=>{
 				let nm = LC.name.substring(0,2);  // field name
-				let n  = parseInt(LC.name.substring(2)); // bus number
```
blazoncek: Same here.

wled00/data/settings_leds.htm
DedeHai: but isnt n passed as 0-9 and A-Z ? can 'A' be parsed into int?

DedeHai:
I am no longer too suer about this PR being a good idea "as is" for 0.16. At bootup, free heap is plenty, LED size is accepted and buffers are created, after wifi connection heap runs out and UI completely breaks down on ESP8266. This may need some adustment. The calculation is fine however for 0.15.2 porting.

blazoncek:
> At bootup

It means somebody modified cfg.json by uploading it or editing it using /edit endpoint.
If it is changed via UI, WiFi is initialised and most buffer (although smaller) are allocated. When saving such configuration it has to succeed before it is saved, meaning all buffers are allocated the same size as during bootup.

wled00/data/settings_leds.htm
blazoncek: You are correct, but the solution presented here is not correct. `n` has to be converted.

DedeHai:
> It means somebody modified cfg.json by uploading it or editing it using /edit endpoint.

EDIT:  or if someone updates from 0.15

> If it is changed via UI, WiFi is initialised and most buffer (although smaller) are allocated. When saving such configuration it has to succeed before it is saved, meaning all buffers are allocated the same size as during bootup.

curently, calculation for UI and bus creation differ. What I do not like is to limit it to the 4k including all buffers, that will just bring back #4614. ESP8266 can handle up to about 800 LEDs, maybe a bit less if there are no additional usermods etc. with my current test code. I may need to reduce that number to fully stabilize it.

DedeHai:
btw. what is the reason there are two functions to calculate bus memory? in polybus there is `memUsage()` and there is `getDataSize()` both doing the same thing. Any good reason not to get rid of `getDataSize()`?

blazoncek:
> both doing the same thing

Are they? I don't think so.
One estimates the other returns actual memory use.

Unfortunately NPB does not provide memory estimation/requirement routines. And you can only get single buffer size out of it once bus is initialised (RMT may use 2 buffers).

DedeHai:
> Are they? I don't think so.
> One estimates the other returns actual memory use.

ah, got it. do we need both? decision to allocate a bus is made on the estimation, what is the "actual use" used for? also: you fixed only one of them :)

wled00/FX_fcn.cpp
```diff
@@ -1171,14 +1171,41 @@ void WS2812FX::finalizeInit() {
   digitalCount = 0;
   #endif
 
+  DEBUG_PRINTF_P(PSTR("Heap before buses: %d\n"), getFreeHeapSize());
```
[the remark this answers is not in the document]
willmmiles: `getFreeHeapSize()` lives in "fcn_delcare.h".  Unfortunately `ESP.getFreeHeap()` doesn't correctly return available memory on all ESP32 platforms, so we have a local function to make the correct platform call.

wled00/data/settings_leds.htm
DedeHai: I see you added a fix to your fork, do you mind adding that to this PR too? I am struggling a bit with JS.

blazoncek: I will once I get back from my trip.

DedeHai: I think I managed to extract it. At least it worked in tests but please check.

wled00/bus_wrapper.h
```diff
@@ -1363,6 +1360,7 @@ class PolyBus {
       if (_useParallelI2S) {
         if (num > 15) return I_NONE;
         if (num < 8) offset = 1;  // 8 I2S followed by 8 RMT
+        if (num < 8) offset = 1;  // 8 I2S followed by 8 RMT
```
willmmiles: We don't need to do this twice ;)

wled00/FX_fcn.cpp
```diff
@@ -1178,17 +1178,41 @@ void WS2812FX::finalizeInit() {
...
+      unsigned i2sCommonSize = stepFactor * bus.count * (3*Bus::hasRGB(bus.type)+Bus::hasWhite(bus.type)+Bus::hasCCT(bus.type)) * (Bus::is16bit(bus.type)+1);
+      if (i2sCommonSize > maxI2S) maxI2S = i2sCommonSize;
+    }
+    #endif
+    if (mem + maxI2S <= MAX_LED_MEMORY) {
+      BusManager::add(bus);
+      DEBUG_PRINTF_P(PSTR("Bus memory: %uB\n"), memB);
     } else {
-      errorFlag = ERR_NORAM_PX; // alert UI
       DEBUG_PRINTF_P(PSTR("Out of LED memory! Bus %d (%d) #%u not created."), (int)bus.type, (int)bus.count, digitalCount);
```
willmmiles, line 1212: We lost the errorFlag setting here. Should we put it back?

wled00/bus_wrapper.h
```diff
@@ -469,20 +469,20 @@ class PolyBus {
...
+      if (_useParallelI2S && channel > 6) channel -= 7; // accommodate I2S1 which is used as 1st bus
+    }
+    #elif defined(CONFIG_IDF_TARGET_ESP32S2)
+    if (_useParallelI2S && channel > 7) channel -= 8; // accommodate I2S1 which is used as 1st bus
+    #elif defined(CONFIG_IDF_TARGET_ESP32S3)
+    if (_useParallelI2S && channel > 7) channel -= 8; // accommodate I2S1 which is used as 1st bus
     #endif
-
+  #endif
     void* busPtr = nullptr;
```
willmmiles, line 486: None of these changes are relevant, and the original code was clearer and had less replication.  (But then again, I wrote it.)  I would revert these changes.

DedeHai: fine with me, these changes are pulled from @blazoncek  fork, did not know if they fix something.

wled00/FX_fcn.cpp
DedeHai: merge collateral. I'll put it back.

wled00/bus_wrapper.h
DedeHai: got duplicated in the merge

wled00/bus_wrapper.h
```diff
@@ -1212,68 +1212,65 @@ class PolyBus {
...
+      case I_32_I2_FW6_5    : // fallthrough
+      case I_32_I2_2805_5   : size = (size + 2*count);     break; // 5 channels
+      case I_32_I2_SM16825_5: size = (size + 2*count)*2;   break; // 16 bit, 5 channels
       #endif
+      default               : size *= 2;                   break; // everything else uses 2 buffers
     #endif
-      // everything else uses 2 buffers
-      default:         size *= 2; break;
     }
     return size;
```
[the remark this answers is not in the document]
DedeHai: the reason _useParallelI2S is used is to ge the correct function pointer, not to change the math.

PR #4939, opened 2025-09-16
Safety Checks for UI, fix for cfg exceeding LED limit

### LINKED ISSUE #4936: No fallback if bus fails to be added in 0.16
### What happened?

At bootup when reading the config and the number of LEDs is set such that they exceed MAX_LED_MEMORY no bus is added. This results in two problems:
- no LED output at all
- UI fails to load as there seems to be a required json element missing

### To Reproduce Bug

I switched between different branches, that used the old and the new way of calculating LED memory on ESP8266 but I guess it will also work with:
- start with 0.15, add 1000 WS28x
- update to https://github.com/wled/WLED/pull/4928

### Expected Behavior

If a digital bus fails to be added at bootup, instead of exiting the loop its size should be truncated to the default amount of LEDs (30) instead of exiting the loop. This will overrun the MAX_LED_MEMORY slightly but that is the lesser evil. The users will need to adjust anyway.
Also the UI should not fail to load with an error (in index.js) if there is missing JSON elements, it should handle it more gracefully.

### Install Method

Self-Compiled

### What version of WLED?

0.16 PR

ESP8266

### Relevant log/trace output

```shell
web UI: TypeError: cannot read properties of null
```

### Anything else?

_No response_

### Code of Conduct

DedeHai:
- added some safety checks for UI to not throw errors if JSCON parameters are missing
- add all buses in config even if they exceed RAM limit, truncates long buses

fixes #4936

this is mostly what codepilot came up with.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1173,11 +1173,17 @@ void WS2812FX::finalizeInit() {
 
   // create buses/outputs
   unsigned mem = 0;
-  for (const auto &bus : busConfigs) {
-    mem += bus.memUsage(Bus::isDigital(bus.type) && !Bus::is2Pin(bus.type) ? digitalCount++ : 0); // includes global buffer
-    if (mem <= MAX_LED_MEMORY) {
-      if (BusManager::add(bus) == -1) break;
-    } else DEBUG_PRINTF_P(PSTR("Out of LED memory! Bus %d (%d) #%u not created."), (int)bus.type, (int)bus.count, digitalCount);
+  for (auto bus : busConfigs) {
+    unsigned busMemUsage = bus.memUsage(Bus::isDigital(bus.type) && !Bus::is2Pin(bus.type) ? digitalCount : 0);
+    // if memory exceeds limit, add it anyway but reduce LED count to default if it is larger (a bit dangerous, but prevents no buses being created at all)
+    if (mem + busMemUsage > MAX_LED_MEMORY) {
+      bus.count = min((int)bus.count, DEFAULT_LED_COUNT);
+      DEBUG_PRINTF_P(PSTR("Bus %d memory usage exceeds limit, setting count to %d\n"), (int)bus.type, bus.count);
+    }
+    if (BusManager::add(bus) != -1) {
+      mem += bus.memUsage(Bus::isDigital(bus.type) && !Bus::is2Pin(bus.type) ? digitalCount : 0);
+      if (Bus::isDigital(bus.type) && !Bus::is2Pin(bus.type)) digitalCount++;
+    } else break;
   }
   busConfigs.clear();
   busConfigs.shrink_to_fit();
--- a/wled00/data/index.js
+++ b/wled00/data/index.js
@@ -668,9 +668,20 @@ function parseInfo(i) {
 	if (loc)    name = "(L) " + name;
 	d.title      = name;
 	simplifiedUI = i.simplifiedui;
-	ledCount     = i.leds.count;
+	// safety checks for LED count data to prevent UI crashes
+	if (i.leds && typeof i.leds.count !== 'undefined') {
+		ledCount = i.leds.count;
+	} else {
+		console.warn('No LED count');
+		ledCount = 30; // fallback value
+	}
 	//syncTglRecv   = i.str;
-	maxSeg       = i.leds.maxseg;
+	if (i.leds && typeof i.leds.maxseg !== 'undefined') {
+		maxSeg = i.leds.maxseg;
+	} else {
+		console.warn('No max segment data');
+		maxSeg = 16; // Reasonable fallback for max segments
+	}
 	pmt          = i.fs.pmt;
 	if (pcMode && !i.wifi.ap) gId('edit').classList.remove("hide"); else gId('edit').classList.add("hide");
 	gId('buttonNodes').style.display = lastinfo.ndc > 0 ? null:"none";
@@ -912,12 +923,24 @@ function populateSegments(s)
 		gId(`segr${i}`).classList.add("hide");
 	}
 	if (segCount < 2) {
-		gId(`segd${lSeg}`).classList.add("hide"); // hide delete if only one segment
-		if (parseInt(gId("seg0bri").value)==255) gId(`segp0`).classList.add("hide");
+		// safety check for segment elements to prevent UI crashes
+		const segdElement = gId(`segd${lSeg}`);
+		if (segdElement) segdElement.classList.add("hide"); // hide delete if only one segment
+		const seg0briElement = gId("seg0bri");
+		const segp0Element = gId(`segp0`);
+		if (seg0briElement && segp0Element && parseInt(seg0briElement.value)==255) segp0Element.classList.add("hide");
 		// hide segment controls if there is only one segment in simplified UI
 		if (simplifiedUI) gId("segcont").classList.add("hide");
 	}
-	if (!isM && !noNewSegs && (cfg.comp.seglen?parseInt(gId(`seg${lSeg}s`).value):0)+parseInt(gId(`seg${lSeg}e`).value)<ledCount) gId(`segr${lSeg}`).classList.remove("hide");
+	// safety checks for segment control elements
+	const segSElement = gId(`seg${lSeg}s`);
+	const segEElement = gId(`seg${lSeg}e`);
+	const segrElement = gId(`segr${lSeg}`);
+	if (!isM && !noNewSegs && segSElement && segEElement && segrElement) {
+		const segLen = cfg.comp.seglen ? parseInt(segSElement.value) : 0;
+		const segEnd = parseInt(segEElement.value);
+		if (segLen + segEnd < ledCount) segrElement.classList.remove("hide");
+	}
 	gId('segutil2').style.display = (segCount > 1) ? "block":"none"; // rsbtn parent
 
 	if (Array.isArray(li.maps) && li.maps.length>1) {
@@ -2253,7 +2276,9 @@ function rptSeg(s)
 	var rev = gId(`seg${s}rev`).checked;
 	var mi = gId(`seg${s}mi`).checked;
 	var sel = gId(`seg${s}sel`).checked;
-	var pwr = gId(`seg${s}pwr`).classList.contains('act');
+	// safety check for segment power element to prevent UI crashes
+	const segPwrElement = gId(`seg${s}pwr`);
+	var pwr = segPwrElement ? segPwrElement.classList.contains('act') : false;
 	var obj = {"seg": {"id": s, "n": name, "start": start, "stop": (cfg.comp.seglen?start:0)+stop, "rev": rev, "mi": mi, "on": pwr, "bri": parseInt(gId(`seg${s}bri`).value), "sel": sel}};
 	if (gId(`seg${s}grp`)) {
 		var grp = parseInt(gId(`seg${s}grp`).value);
@@ -2380,7 +2405,13 @@ function setGrp(s, g)
 
 function setSegPwr(s)
 {
-	var pwr = gId(`seg${s}pwr`).classList.contains('act');
+	// safety check for segment power element to prevent UI crashes
+	const segPwrElement = gId(`seg${s}pwr`);
+	if (!segPwrElement) {
+		console.warn('No power elemen');
+		return;
+	}
+	var pwr = segPwrElement.classList.contains('act');
 	var obj = {"seg": {"id": s, "on": !pwr}};
 	requestJson(obj);
 }


wled00/FX_fcn.cpp
```diff
@@ -1173,11 +1173,17 @@ void WS2812FX::finalizeInit() {
...
+    // if memory exceeds limit, add it anyway but reduce LED count to default if it is larger (a bit dangerous, but prevents no buses being created at all)
+    if (mem + busMemUsage > MAX_LED_MEMORY) {
+      bus.count = min((int)bus.count, DEFAULT_LED_COUNT);
+      DEBUG_PRINTF_P(PSTR("Bus %d memory usage exceeds limit, setting count to %d\n"), (int)bus.type, bus.count);
+    }
+    if (BusManager::add(bus) != -1) {
+      mem += bus.memUsage(Bus::isDigital(bus.type) && !Bus::is2Pin(bus.type) ? digitalCount : 0);
+      if (Bus::isDigital(bus.type) && !Bus::is2Pin(bus.type)) digitalCount++;
+    } else break;
   }
```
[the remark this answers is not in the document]
DedeHai: That is intentional. Exceeding RAM limit is not catasrophical and users will be forced to update their config after moving to 0.16

willmmiles:
Truncating too long buses is not OK -- it creates the weird behavior that the configuration "mutates" when saved by the user.  (Why won't my settings stick??)   I think what we need to do is create a bus in an "invalid" state and let it carry forward without operating.  I can think of two ways to do this:

-  Define a "BusDummy" type that we create instead, that doesn't actually do anything but hold the BusConfig for later readback - get/setPixelColor are ignored, etc.

or

- Add a `static size_t totalMemUsage` inside `BusDigital`, maintained at construct/destruct time; if the check fails, leave `_valid = false`.   Also remove the `isOk()` checks in `serializeConfig()` and `getSettingsJS()` so failed buses will be displayed in the settings page for the user to fix.  (We could potentially echo back the validity, too, so the UI could point them out.)

I lean towards the latter.

willmmiles:
Truncating too long buses is not OK -- it creates the weird behavior that the configuration "mutates" when saved by the user.  (Why won't my settings stick??)   I think what we need to do is create a placeholder bus object (`BusPlaceholder`?) that doesn't actually *do* anything, but carries forward the settings for later inspection.

(We should probably also consider removing the `Bus::isOk()` checks in `getSettingsJS()` and `serializeConfig()` -- that way failed buses can still be edited in the settings page as well.)

DedeHai:
> Truncating too long buses is not OK -- it creates the weird behavior that the configuration "mutates" when saved by the user. (Why won't my settings stick??) I think what we need to do is create a placeholder bus object (`BusPlaceholder`?) that doesn't actually _do_ anything, but carries forward the settings for later inspection.
> 
> (We should probably also consider removing the `Bus::isOk()` checks in `getSettingsJS()` and `serializeConfig()` -- that way failed buses can still be edited in the settings page as well.)

very good point, I saw that and thought the same thing but did not think about a solution yet. Any suggestion in how to preserve the config?

wled00/FX_fcn.cpp
```diff
@@ -1180,14 +1180,17 @@ void WS2812FX::finalizeInit() {
...
-      DEBUG_PRINTF_P(PSTR("Out of LED memory! Bus %d (%d) #%u not created."), (int)bus.type, (int)bus.count, digitalCount);
+  for (auto bus : busConfigs) {
+    unsigned busMemUsage = bus.memUsage(Bus::isDigital(bus.type) && !Bus::is2Pin(bus.type) ? digitalCount : 0);
+    // if memory exceeds limit, add it anyway but reduce LED count to default if it is larger (a bit dangerous, but prevents no buses being created at all)
+    if (mem + busMemUsage > MAX_LED_MEMORY) {
+      bus.count = min((int)bus.count, DEFAULT_LED_COUNT);
+      DEBUG_PRINTF_P(PSTR("Bus %d memory usage exceeds limit, setting count to %d\n"), (int)bus.type, bus.count);
     }
+    if (BusManager::add(bus) != -1) {
+      mem += bus.memUsage(Bus::isDigital(bus.type) && !Bus::is2Pin(bus.type) ? digitalCount : 0);
```
blazoncek: No need to estimate when you have bus already created.
Use `BusManager::getBus(BusManager::getNumBusses()-1)->getBusSize();` instead.

wled00/data/index.js
```diff
@@ -668,9 +668,20 @@ function parseInfo(i) {
 	if (loc)    name = "(L) " + name;
 	d.title      = name;
 	simplifiedUI = i.simplifiedui;
-	ledCount     = i.leds.count;
+	// safety checks for LED count data to prevent UI crashes
+	if (i.leds && typeof i.leds.count !== 'undefined') {
```
blazoncek: `count` property is never undefined (see json.cpp) but it can be 0 which may yield incorrect results.
Forcing `_length` to 1 in `finalizeInit()` resolves the issue.

blazoncek (commented):
I would have fixed the `count` issues in `finalizeInit()` instead.

wled00/data/index.js
DedeHai: the whole point of the changes in index.js is to catch things that ought not to be wrong and not crash. And yes, it is sometimes very much undefined: if that field is missing for example. these changes are meant to catch cases that are not anticipated but I have seen happening. the point is to still finish loading the page if stuff is missing due to incomplete jsons and thus make the UI a bit more robust.

DedeHai: I am not really familiar with how things are processed in index.js, I have simply instructed copilot to fix the errors I have observed and these changes fix those errors. It is more of a workaround in case the json data is incomplete an I am sure there are other places that would need fixing but its a start.

willmmiles:
> very good point, I saw that and thought the same thing but did not think about a solution yet. Any suggestion in how to preserve the config?

Make up a new Bus subclass `BusPlaceholder` (or some better name) that ignores the IO calls (`_valid` is false, pure virtuals are no-ops) but retains the relevant config data.  Then, in `finalizeInit`, if a bus will exceed the memory limit, we add a `BusPlaceholder` instead.  Since the "type" value gets preserved, future round-trips through `finalizeInit` will attempt to create the "correct" bus class first.

DedeHai:
> Make up a new Bus subclass `BusPlaceholder` (or some better name) that ignores the IO calls (`_valid` is false, pure virtuals are no-ops) but retains the relevant config data. Then, in `finalizeInit`, if a bus will exceed the memory limit, we add a `BusPlaceholder` instead. Since the "type" value gets preserved, future round-trips through `finalizeInit` will attempt to create the "correct" bus class first.

thanks, I need to process that. Working on heap/ram stuff RN so this is on backburner. If you have the time and interest, feel free to contribute directly to this PR.

wled00/data/index.js
blazoncek: > due to incomplete jsons

good luck with that. unfortunately.
you'll be chasing your own shadow.

DedeHai: yea, basically. but these changes already help a lot. instead of the UI just throwing errors, it loads to the point where the info panel works, so a reset it possible to execute. or settings (sometimes), so better than nothing I guess.

wled00/FX_fcn.cpp
```diff
@@ -1180,14 +1180,17 @@ void WS2812FX::finalizeInit() {
...
+    unsigned busMemUsage = bus.memUsage(Bus::isDigital(bus.type) && !Bus::is2Pin(bus.type) ? digitalCount : 0);    
+    if (mem + busMemUsage > MAX_LED_MEMORY) {      
+      DEBUG_PRINTF_P(PSTR("Bus %d with %d LEDS memory usage exceeds limit\n"), (int)bus.type, bus.count);
+      use_placeholder = true;
     }
+    if (BusManager::add(bus, use_placeholder) != -1) {
+      mem += BusManager::busses.back()->getBusSize();
+      if (Bus::isDigital(bus.type) && !Bus::is2Pin(bus.type)) digitalCount++;
+    } else break;
   }
```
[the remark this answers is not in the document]
willmmiles: digitalCount is intentionally still incremented with placeholders so as to perform consistent resource mapping.

The direct access to `BusManager::busses` is also intentional: it is part of BusManager's public API.  The extra methods wrapping access to this object only serve to obfuscate the code and should be removed in the future.  Further, using constructs of the form `my_vector[my_vector.size()-1]` instead of `back()` are poor practice -- apart from being unnecessarily verbose and obscuring what the code is doing, they add an implicit dependency on O(1) indexing on the container type.

You are correct to note that the else branch is now dead.  The bus management API needs a fair amount of cleanup; I am leaving that condition in place for now in case it is needed in the future.

blazoncek:
@willmmiles IMO there is no need to complicate things with bus placeholder. The only thing needed to be done is ignore isOk() on some places and proceed without memory allocation and underlying physical bus initialisation.

DedeHai:
I did some experimenting in my dev branch and I think I have a good overall solution for ESP8266. It goes something like this:
- the UI as well as finalizeInit() do not take the MAX_LED_MEMORY as a hard limit since we also need buffers and those are not all obligatory
- instead the UI takes the MAX_LED_MEMORY to throw warnings at the user and denies writing the config if double that limit is exceeded, i.e. bus+globalbuffer+segmentbuffer (with a warning message that it cannot be handled).
- finalizeInit() uses the same approach but has a slightly lower estimate to make sure buses get added if UI does not deny it
- for existing configs that would exceed the limit the bus length gets truncated (thx @willmmiles for taking the effort)

still need to test how that behaves on ESP32 and maybe tweak it a little. For ESP8266 the new limit is about 720 WS281x LEDs, everything still works at that limit although a bit wonky sometimes. The user is _actively_ informed about stability issues when setting more than about 300 LEDs. 

Once fully tested, I will push an update to this PR

willmmiles:
> @willmmiles IMO there is no need to complicate things with bus placeholder. The only thing needed to be done is ignore isOk() on some places and proceed without memory allocation and underlying physical bus initialisation.

I considered that approach, but it would mean that all bus types would need have to have a "don't actually construct" argument, which struck me as rather silly.  I found that using a clearly defined placeholder type was a cleaner implementation in the current code.  I agree that there *is* a general problem of handling bus initialization failure that overlaps quite a bit here, but I think that needs to be approached in context of a bigger Bus API redesign.  (I haven't done a full pass over it yet, but I'd like to give serious thought to elimininating `IsOk()` in calls like `setPixelColor()` in favor of replacing a failed bus with a Placeholder instead.  This should yield a slight speedup as we don't need to run the check for every pixel, it gets embedded in the virtual dispatch instead.)

willmmiles:
> * for existing configs that would exceed the limit the bus length gets truncated (thx @willmmiles for taking the effort)

I'm still hard against that.  Either construct the bus or don't, but don't change the user's settings.

DedeHai:
> I'm still hard against that. Either construct the bus or don't, but don't change the user's settings.

doesn't your added code take care of that? i.e. truncating the created bus but keeping the config? sorry, did not yet have time to study in detail what you added.

edit: the underlaying issue I am trying to address: ESP8266 on 0.16 will have a bad time (to the point where UI access becomes impossible) if you are coming in with a config using >1000 LEDs. Not creating the bus is one option, creating it but smaller would not leave users with a "its all bricked" moment. Not sure how edge-case this is though.

blazoncek:
> all bus types would need have to have a "don't actually construct"

I don't think so. If they *fail to construct properly*, they are just not ok.

However, the approach here is that you want to check if memory exceeds MAX_LED_MEMORY before bus is created and it is done in `finalizeInit()`. If you move the check into `BusManager::add()` you can summarize existing buses, add estimate of current bus and then decide what to do. This would simplify `finalizeInit()` function.
I've foreseen expansion of virtual buses (see usermod bus PR) and this could be of such use. Virtual bus that does nothing, but will retain every option selected.

I, personally, would just drop such bus and every bus after it.

willmmiles:
The posted code creates a fake bus that holds the settings but is inactive -- it doesn't produce any output.  It seemed the best approach to me; making working busses that have different internal sizes than their settings would be complex to say the least.

Re tightened requirements on 0.16, I think we should come at this from two fronts:
- First, add a check in 0.15.2 that'll refuse the update if your config won't be valid.  #4930 puts the framework in place for that -- while the prototype there only checks the release name, we're free to add additional useful information in the metadata struct, such as a "recommended max LEDs" value.  I expect this to catch 90-95% of problem cases.  Users can still use the "override" flag to force an update, but at that point we're justified in saying "you were warned".
- Second, if the config is no longer valid in the new version, then we prioritize (a) keeping the web interface up to allow the user to fix the config and/or switch firmware; then (b) making it clear the config is unsupported (for example: if you upload a config with too many LEDs with this PR; then after you reboot, the settings page tells you why it's unhappy right away when you open it).  If the user wants to fix their config by lowering the number of LEDs, they're free to do so at that point; but I suspect most such users would rather downgrade than have their configuration changed.  As long as the web API stays up, the device isn't "bricked" - it's just not working ;)

willmmiles:
> > all bus types would need have to have a "don't actually construct"
> 
> I don't think so. If they _fail to construct properly_, they are just not ok.

I was trying to make the point that allowing a "bus is not ok" state puts a performance cost on every bus access.  If we do not allow "not ok" busses to exist at all -- replacing them with placeholders or the like -- then we don't ever have to check if the bus is ok, speeding up the core render loops.

> 
> However, the approach here is that you want to check if memory exceeds MAX_LED_MEMORY before bus is created and it is done in `finalizeInit()`. If you move the check into `BusManager::add()` you can summarize existing buses, add estimate of current bus and then decide what to do. This would simplify `finalizeInit()` function. I've foreseen expansion of virtual buses (see usermod bus PR) and this could be of such use. Virtual bus that does nothing, but will retain every option selected.

I agree that the division of work between `finalizeInit()` and `BusManager` isn't in the right place.  As usual my predilection for the minimal patch involves extra work to avoid changes to the code structure.  I'll see if I can spare another hour to do a better cleanup.
 
> I, personally, would just drop such bus and every bus after it.

We cannot drop busses as it destroys the user configuration.  IMO "my configuration didn't save" is always a bug.

DedeHai:
I have a new approach/suggestion whicht might not require to divert or drop any buses but its a bit of a stretch: accept all buses if memory permits (or even if it does not really). Let the heap checker in the main loop handle the rest. 
Reasoning: this whole situation only arises if a user 
a) upates from a version that is incompatible
b) manually manipulates the config
If the heap-check in the main loop is updated to free as much heap as required (for example by dropping all buses) the UI will still work (after a while) and users are able to update the config.

willmmiles:
> If the heap-check in the main loop is updated to free as much heap as required (for example by dropping all buses) the UI will still work (after a while) and users are able to update the config.

Unfortunately, if it drops all busses, none of them will appear in the settings page: the bus list there is generated from the live BusManager state.

DedeHai:
> > If the heap-check in the main loop is updated to free as much heap as required (for example by dropping all buses) the UI will still work (after a while) and users are able to update the config.
> 
> Unfortunately, if it drops all busses, none of them will appear in the settings page: the bus list there is generated from the live BusManager state.

since the major change in this matter are the added buffers, simply dropping those should work. The heap checker currently drops all segments but if the culprit is the global `_pixels` buffer, resetting segments is not enough. Adding a "heap panic" extension to the checker and maybe some function to free that buffer might do the trick without messing with buses.

Edit: this works 
```C++
    if (heap < MIN_HEAP_SIZE && lastHeap < MIN_HEAP_SIZE) {
      DEBUG_PRINTF_P(PSTR("Heap too low! %u\n"), heap);
      if(heapPanic) {
        DEBUG_PRINTLN(F("!Heap panic reset!"));
        strip.~WS2812FX();      // deallocate strip and all its memory
        new(&strip) WS2812FX(); // re-create strip object, respecting memory limits
      }
      forceReconnect = true;
      strip.resetSegments(); // remove all but one segments from memory
      heapPanic = true;      // enable last resort measures on next call if heap is still low
    } else if (heap < MIN_HEAP_SIZE) {
      DEBUG_PRINTLN(F("Heap low, purging segments."));
      strip.purgeSegments();
    } else {
      heapPanic = false;
    }
```

Edit2:
I finally get your point about the "dummy bus" to respect the config settings. While I do agree that messing with users setting is bad, truncating a bus wont directly do that unless a user saves them but you are correct that it may be confusing. If we do add stubs, those must be highlighted in settings to indicate so or the next question will be: "why won't LEDs light up?" 
To recap:
- I agree there should be a workaround for setups that now exceed the capabilities
- I disagree that a bus should just be dropped/duped, leaving users literally "in the dark", capping its length is a better indicator (imagine a ESP8266 with just one output, 900 LEDs)
- I dont have a good solution to the config issue, I imagine adding a new variable to distinguish configured length and actual length is a bad workaround.

Edit3:
with the above heap guard "fix" we could keep the old bus creation logic, thus any bus that was allowed before will still be allowed. if it breaks the UI, it will recover within a minute and not change any settings. My current code will not let a user save settings that are considered unsafe, so they are left with the choice to update the settings to safe levels or downgrade.

willmmiles:
> * I disagree that a bus should just be dropped/duped, leaving users literally "in the dark", capping its length is a better indicator (imagine a ESP8266 with just one output, 900 LEDs)
> * I dont have a good solution to the config issue, I imagine adding a new variable to distinguish configured length and actual length is a bad workaround.

Right, I don't think you can have both.  Either you seperately store "configured" vs "live" lengths, OR you have to null out buses that can't be allocated.  I really don't think we should be trying to support "partially functional" buses.  Also, if we override the user config, we open ourselves up to "why doesn't my configuration work??? I saved 900 LEDs but only 30 light up??" if there's ever a bug in the JS validation.

I applaud that you're pushing for graceful degredation.  What would change my mind on this is a better solution for posting error messages back to the main UI.  Something that would stay on screen (without covering up parts of the interface) and clearly indicate what the system is unhappy about, what recovery actions were taken, and where to go to fix it.  Otherwise I think sophisticated recovery behaviour will be more confusing to users than helpful.

> Edit3: with the above heap guard "fix" we could keep the old bus creation logic, thus any bus that was allowed before will still be allowed. if it breaks the UI, it will recover within a minute and not change any settings. My current code will not let a user save settings that are considered unsafe, so they are left with the choice to update the settings to safe levels or downgrade.

I'm on board with the heap guard fix, but I do think it's worth pointing out that the impact of it is the same as dropping the bus: the LEDs will stay dark if there isn't enough memory to allocate `_pixels` or segment buffers.

DedeHai:
The way I see it is once again a question of "how do we anger the least users" :)
Disregarding the general issue about buses that can't be allocated (to which I think your approach is good) it is mostly if not only about ESP8266 with more than 700 LEDs, which we saw might be not that uncommon.
With 0.14 and 4k of LED buffer memory the max should be around 1300 LEDs. Not sure about non WS281x types but I assume those are less common on this ESP. In my current approach (not in this PR) its like this:

- use the same calculation limit as it was, I tested a bus with 1200 LEDs.
- At bootup, there is enough heap to allocate this amount, including _pixels[] and segment buffer and the LEDs light up.
- After wifi connects, heap drops below 6k, bricking the UI
- When heap recovery kicks in, enough heap is available (almost 10k) and UI starts working
- LEDs are still on, lit in orange but cant be controlled as there is no segment buffer anymore.
- Settings show a big red warning that memory is exceeded more than 2 fold. 
- When trying to save, it sais that this many LEDs cant be handled and user has to reduce to about 700LEDs in order to allow saving settings.

what I did not try yet is if the heap is enough to do an OTA downgrade. IMHO this would be a good approach to solve this particular case.
If we say we only allow full functionality or non at all, the new limit for the ESP8266 is around 300 pixels, for ESP32 about 3000, which I personally think a lot of existing setups may exceed thus resulting in lots of complaints. A somewhat functional system (i.e. only one segment layer, not all effects work with transitions, sometimes segments will not get a buffer etc.) but working UI and working in general is limited to approx. 700LEDs on ESP8266 and 4000LEDs on ESP32 (tested).

willmmiles:
> The way I see it is once again a question of "how do we anger the least users" :) 

We're in agreement on that!

> * use the same calculation limit as it was, I tested a bus with 1200 LEDs.
> * At bootup, there is enough heap to allocate this amount, including _pixels[] and segment buffer and the LEDs light up.
> * After wifi connects, heap drops below 6k, bricking the UI
> * When heap recovery kicks in, enough heap is available (almost 10k) and UI starts working
> * LEDs are still on, lit in orange but cant be controlled as there is no segment buffer anymore.
> * Settings show a big red warning that memory is exceeded more than 2 fold.
> * When trying to save, it sais that this many LEDs cant be handled and user has to reduce to about 700LEDs in order to allow saving settings.

I'm worried that seeing it light the LEDs in orange might give the false impression that we're applying some artificial limit in the config page.  After all, clearly the software is capable of lighting them up -- so why can't I use them?

Still I think we're largely in agreement in the overall thrust of the solution.  Lighting the LEDs or not when under memory pressure is negotiable.

> 
> what I did not try yet is if the heap is enough to do an OTA downgrade. IMHO this would be a good approach to solve this particular case. 

+1 - we should guarantee that at least.

> If we say we only allow full functionality or non at all, the new limit for the ESP8266 is around 300 pixels, for ESP32 about 3000, which I personally think a lot of existing setups may exceed thus resulting in lots of complaints. A somewhat functional system (i.e. only one segment layer, not all effects work with transitions, sometimes segments will not get a buffer etc.) but working UI and working in general is limited to approx. 700LEDs on ESP8266 and 4000LEDs on ESP32 (tested).

I think we should set the "hard" limits where things will work with no layering or transitions.  Those are fun features but at least some real-world setups can get by without them.   In the code, we should do what we can to degrade gracefully (eg. cancel transitions first if we're out of memory for segments; have a strong and well-documented ordering for discarding segments if there's insufficient memory for all of them; etc.) and provide some guidance on the "definitely safe" limits on the web site.

willmmiles:
>> what I did not try yet is if the heap is enough to do an OTA downgrade. IMHO this would be a good approach to solve this particular case.
>+1 - we should guarantee that at least.

Short answer seems to be that this will need work...   I wasn't intending to test this, but I ran in to it with #4930 since I was just using the same 8266 I was last testing for this PR, configured for 500 LEDs.  It triggered the low heap handler during the upload and reset the wifi. :(

DedeHai:
how long did that OTA take? I just tested with my enhanced code and OTA works as long as it takes less than 45 seconds. could up that limit easily.

Let me summarize what we discussed and the options as I see them:
- updates from <0.16 can have more LEDs configured than can safely be handled
- if there are bugs in 0.16 onward, we want to handle failed buses more gracefully: currently, they are dropped and no longer show on the config page but still remain in the config file. Suggestion by @willmmiles is to create "dummy" buses as placeholders. These should be rare cases, I agree with this handling.
- mostly (only?) relevant for 8266 setups with > 700 LEDs configured: after updating, we have these options to handle it:
  - let bus creation fail silently, probably leaving users puzzled as why nothing works, settings will show the bus removed
  - defer bus creation to a dummy, leaving settings page intact. Settings page does inform about excessive memory use but LEDs wont do anything and cant be controlled, also leaving users puzzled.
  - truncate the bus to 30 LEDs showing at least some effort, config will show a 30LED bus. If users hit save, it will overwrite the config and they have to set that up again after downgrading. less puzzling but probably more annoying.
  - accept the way too large bus (works after the OTA reboot) leaving no heap to load the UI, i.e. its bricked until the heap-guard kicks in and drops the buffers after 30 seconds (tested and working). This may also leave users puzzled as "it works for a few seconds" but then stops doing so. The config page will show correct bus(es) warning is present. 
- In all cases, saving LED settings is denied until reasonable values are entered (about 700 LEDs currently)

To me, any of the options are possible, personally I think the last one might be the least disruptive but would collide with the dummy approach which has other benefits in the future (no isOK() checking). I do agree here that users need more info on what is going on. We could add a slightly more sophisticated "error" log. Instead of doing the current short "toast" message, an error / symbol could be added to the UI, if clicked, it opens an error log of the last x error messages that could be a bit more descriptive. Extending the `ERR_x` list in const.h is possible.

willmmiles:
> how long did that OTA take? I just tested with my enhanced code and OTA works as long as it takes less than 45 seconds. could up that limit easily.

I didn't time it; I had a bunch of extra serial logging enabled to track the validation process as the ESP8266 validation has to be done partway through the upload.  It wasn't super fast but I doubt it was a full 45 seconds.   In any event, I'm not ok with "fails if your wifi isn't fast enough".   Using lots of memory during OTA is normal and expected -- that's why it clears all the segments first.   I'll figure out a way to cue the heap recovery to please leave the wifi alone if an OTA is in progress, and ensure the OTA failure timeouts are short enough that it'll recover quickly if there's a real problem with the OTA itself.  (One of the fixes in the WIP patch is to guarantee that regular processing is restarted if the OTA aborts halfway - the original code wouldn't always recover.)

>  1.  let bus creation fail silently, probably leaving users puzzled as why nothing works, settings will show the bus removed
>  2. defer bus creation to a dummy, leaving settings page intact. Settings page does inform about excessive memory use but LEDs wont do anything and cant be controlled, also leaving users puzzled.
>  3. truncate the bus to 30 LEDs showing at least some effort, config will show a 30LED bus. If users hit save, it will overwrite the config and they have to set that up again after downgrading. less puzzling but probably more annoying.
>  4. accept the way too large bus (works after the OTA reboot) leaving no heap to load the UI, i.e. its bricked until the heap-guard kicks in and drops the buffers after 30 seconds (tested and working). This may also leave users puzzled as "it works for a few seconds" but then stops doing so. The config page will show correct bus(es) warning is present.
>
>In all cases, saving LED settings is denied until reasonable values are entered (about 700 LEDs currently)

(edited to add numbers)

My preference order would be 2, then 4; I still think "doesn't work at all" is a clearer indication that the configuration is untenable than "seems to work. then stops".  I don't think 1 or 3 are good plans; anything that overwrites the user config is bad.  (I for one don't usually remember the exact count of LEDs in a particular build - if I don't remember to back up the config, I end up doing "does 320 light them all?  No?  How about 350?  OK, maybe 335?")

I'd also note that 4 needs a fallback to 2 if even the bus creation fails (ie. someone uploads a bad backup file that has so many LEDs that we can't even construct the buses and buffers once).   Ultimately at the code level the difference between the two is only "what is the hard limit at bus creation time" -- the code itself between 2 and 4 should otherwise be identical.

(Or to put it another way, I think your improvements to the heap recovery are a good idea regardless of what we do for bad config handling.)

> I do agree here that users need more info on what is going on. We could add a slightly more sophisticated "error" log. Instead of doing the current short "toast" message, an error / symbol could be added to the UI, if clicked, it opens an error log of the last x error messages that could be a bit more descriptive. Extending the `ERR_x` list in const.h is possible.

We should probably open a separate issue for an improved error UI and sort out the details elsewhere.

DedeHai:
> In any event, I'm not ok with "fails if your wifi isn't fast enough". 

fully agreed, the heap recovery needs a "gate". I don't know enough about the OTA handler, it might have "is in progress" check. 

> My preference order would be 2, then 4; I still think "doesn't work at all" is a clearer indication that the configuration is untenable than "seems to work. then stops".

you guess is as good as mine :) I have no strong preference, users might have.

> I'd also note that 4 needs a fallback to 2 if even the bus creation fails 

absolutely. I did not poke at "bugged" cases, just the real set limits. And also some UM's might hog a ton of static memory or heap which we cannot anticipate.

>Ultimately at the code level the difference between the two is only "what is the hard limit at bus creation time" -- the code itself between 2 and 4 should otherwise be identical.

correct.

> We should probably open a separate issue for an improved error UI and sort out the details elsewhere.

right.

So I think we are in agreement. Shall I go ahead an merge my local changes to this PR? that would be:
- changes to LED setup to enforce some limits
- remove the 30 LED limit, as well as any restricting memory limit but keep the "dummy" fallback if a bus fails anyway (case 2 + 4)
- add my improvements to the heap handler
- any other minor changes

in order for these changes to effective, https://github.com/wled/WLED/pull/4928  should be merged first. We can still adjust that PR later if there are any bugs in it, they just need to be added to the 0.15 version PR as well.

willmmiles:
I've put a review on #4928.  After those fixes, yes please go ahead and merge your updates here!

DedeHai:
@willmmiles I added in my improvements to the low-heap handler in main loop. 
also added `disableForceReconnect`  which is enabled during OTA. it does not have a timout so if OTA gets interrupted due to network error, I think it wont reach the cleanup stage and stays disabled (and strip stays suspended, as it was, IMHO an edge case with no horrible consequences).
Also added a `isPlacehoder()` to the bus: could also use `isOK()` but I thought this might have some future benefit, can remove later if it turns out that its not useful.
from my side this is ready to merge, I have some more heap handling improvements I did not want to add to this PR which fix some issues on ESP8266. 
I think this is a good base to refine the bus management, still needs more work to come to a conclusion on how to handle bus & buffer memory as `MAX_LED_MEMORY` currently serves double-duty: once for bus allocation and once for UI memory limits. In UI, required buffers are included but `MAX_LED_MEMORY` is doubled as buffers _should_ respect min heap whereas busses do not (and currently can't as they use constructors and not the malloc_x() functions)

wled00/data/index.js
```diff
@@ -2378,7 +2403,13 @@ function setGrp(s, g)
...
 {
-	var pwr = gId(`seg${s}pwr`).classList.contains('act');
+	// safety check for segment power element to prevent UI crashes
+	const segPwrElement = gId(`seg${s}pwr`);
+	if (!segPwrElement) {
+		console.warn('No power elemen');
+		return;
+	}
+	var pwr = segPwrElement.classList.contains('act');
 	var obj = {"seg": {"id": s, "on": !pwr}};
```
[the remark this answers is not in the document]
DedeHai: @coderabbitai regarding your nitpick comments: 
-getPins() can be called on not OK buses, isOK() is used to check if a neopixel bus was created, the bus object with its config does exist.
- the unreachable fallback is not unreachable and the comment already states that this can happen if heap is low. this is a safety measure as the UI does not work properly if there is no segment.
- strip reinitialisation in heap panic is intentionally skipped: it is a last resort measure to free heap for a working UI.

willmmiles:
> also added disableForceReconnect which is enabled during OTA. it does not have a timout so if OTA gets interrupted due to network error, I think it wont reach the cleanup stage and stays disabled (and strip stays suspended, as it was, IMHO an edge case with no horrible consequences).

I fixed the update cleanup handling in #4960, it'll run no matter why the update failed.  (Although deleting the segments to free RAM is, alas, currently unrecoverable.)   I'd also added an "update in progress" test to the heap check -- I went the other way and had the heap checker inspect the `Update` state directly so it'd work with ArduinoOTA or other non-HTTP implementations.  (Although such backdoors will also skip the suspend() and the #4960 validation, sadly. :( )

Merge conflicts, here we come... ;)

> Also added a isPlacehoder() to the bus: could also use isOK() but I thought this might have some future benefit, can remove later if it turns out that its not useful.

Ultimately I want to go the other way around -- I don't think we should ever have any `isSomeBusType()` calls.  The Bus abstraction ought to be actually properly abstract over the needs of Bus clients -- that needs to be a goal of the Bus API improvements.

> I think this is a good base to refine the bus management, still needs more work to come to a conclusion on how to handle bus & buffer memory as MAX_LED_MEMORY currently serves double-duty: once for bus allocation and once for UI memory limits. In UI, required buffers are included but MAX_LED_MEMORY is doubled as buffers should respect min heap whereas busses do not (and currently can't as they use constructors and not the malloc_x() functions)

Agreed.  Personally I'd also like to clean up the duplication of the memory estimation code in C++ and JS.  Later I think we should add an HTTP API endpoint for letting the JS call the C++ code to do the memory estimation.  Hopefully should be straightforward after the bus management work is done.

DedeHai:
>  I'd also added an "update in progress" test to the heap check
> Merge conflicts, here we come... ;)

much better way of doing it. I will remove my crude "hold up with that reset" for one less conflict ;)
 
> > Also added a isPlacehoder() to the bus: could also use isOK() but I thought this might have some future benefit, can remove later if it turns out that its not useful.
> 
> Ultimately I want to go the other way around -- I don't think we should ever have any `isSomeBusType()` calls. The Bus abstraction ought to be actually properly abstract over the needs of Bus clients -- that needs to be a goal of the Bus API improvements.

sounds like a good goal. I will leave the check in for now.

> Agreed. Personally I'd also like to clean up the duplication of the memory estimation code in C++ and JS. Later I think we should add an HTTP API endpoint for letting the JS call the C++ code to do the memory estimation. Hopefully should be straightforward after the bus management work is done.

Fully agree. I also thought about that but could not come up with a good plan on how to do it, my JS / HTTP / WS knowledge is very limited. 

So is this good to merge after I remove the connection reset blocker mentioned above?

willmmiles:
> So is this good to merge after I remove the connection reset blocker mentioned above?

I haven't been able to do a thorough review or test it, but the overall approach seems OK.

DedeHai:
no problem, I'll wait with merging until you approve.

wled00/wled.cpp
```diff
@@ -169,19 +169,46 @@ void WLED::loop()
...
-      DEBUG_PRINTLN(F("Heap low, purging segments."));
-      strip.purgeSegments();
+   // free memory and reconnect WiFi to clear stale allocations if heap is too low for too long, check once per second
+  if ((uint16_t)(millis() - heapTime) > 1000) {
+    #ifdef ESP8266
+    uint32_t heap = getFreeHeapSize(); // ESP8266 needs ~8k of free heap for UI to work properly
+    #else
+    #ifdef CONFIG_IDF_TARGET_ESP32C3
+    // calling getContiguousFreeHeap() during led update causes glitches on C3, this is a dirty workaround (tested up to 1500 LEDs)
+    // this can (probably) be removed once RMT driver for C3 is fixed
```
softhack007, line 179: @DedeHai this reminds me of something. I've seen similar behaviour also on other esp32 boards - especially esp32 and -S3, and a few times on -S2 too.
* When calling ``ESP.getFreeHeap()`` while ``strip.isUpdating()``, this was almost a dead-sure combination causing LED flickering.
* My intermediate hack in WLED-MM was to skip memory checking with `ESP.getFreeHeap()` when `strip.isUpdating()`. It removed the annoying "almost every 5 second flicker".

My educated guess here was that something inside ESP.getFreeHeap() uses some kind of "big kernel lock" that suspends everything while counting free heap bytes.

Maybe it would also help in upstream to avoid heap size checking while RMT is sending out LEDs. Its definitely not just a _-C3 only problem_.

softhack007: Edit: a similar, unguarded call that you might want to protect is here:

https://github.com/wled/WLED/blob/4f968861d680d36f176c26a92288c994582ab292/wled00/json.cpp#L838

this is running in webServer context when you refresh the main UI, so there is a possibility that this leads to flickering, too.

DedeHai: doh, that is the command I was looking for last week and could not remember. The C3 flickers whenever flash is accessed or other blocking processes run, like the free heap check. Its also an issue in the image FX when running animated gif's, that is where I wanted to add a check to not load a new gif frame while LEDs are being updated.
The flickering is (or should be) fixed with the new RMT driver update, but for the C3 it is still pending. I could not get any flickering on a ESP32, S2 or S3 no matter what I tried.

softhack007: maybe my famous "be nice but not too nice" code would help:
````C++
   unsigned t0 = millis();
   while (strip.isUpdating() && (millis() - t0 < 15)) delay(1);    // be nice, but not too nice. Waits up to 15ms
````

you could also use `delay(0)`, its just important that the task scheduler gets a chance for context switching.

Don't use yield() on esp32, because yield() does not work like on 8266.

DedeHai:
@willmmiles did you get a chance to test this? I am working on quite a lot of changes in bus handling to allow RMT and I2S mixed buses and this PR has some overlap with that work and the bus placeholder would be really useful there so I would like to get this PR approved & merged.

wled00/wled.cpp
willmmiles: > My educated guess here was that something inside ESP.getFreeHeap() uses some kind of "big kernel lock" that suspends everything while counting free heap bytes.

Yes, it locks the heap and has to walk the free lists.  It's quite expensive.  We may someday want to think about some other approach to heap usage monitoring -- the web server has the same problem trying to estimate heap availability.

DedeHai:
thanks for the review. I will fix the conflicts, add that "wait until bus is done" to the heap check. Once this is merged https://github.com/wled/WLED/pull/5303 may need a little ironing.

PR #4940, opened 2025-09-16
fix: correctly parse the path in getloc

K3das:
If you put WLED behind a reverse proxy, `getLoc` will ignore the reverse proxy if it is only one subpath (for example, if I had WLED on `/lights/`, `getLoc` would not find the subpath, and settings pages would not load).

Instead of blindly trimming the last path, this change has getLoc trim anything after `/settings/`. 

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable base URL detection for apps served behind reverse proxies or subpaths.
  * Correct handling of URLs with trailing "settings" segments and file-like segments, preventing broken links or missing assets.
  * Works correctly for single-segment paths and instances without a "settings" section.

* **Refactor**
  * Simplified and more robust path parsing to improve navigation and asset loading across varied deployments.

### DIFF
--- a/wled00/data/common.js
+++ b/wled00/data/common.js
@@ -86,7 +86,7 @@ function getLoc() {
 		let paths = path.slice(1,path.endsWith('/')?-1:undefined).split("/");
 		if (paths.length > 1) paths.pop(); // remove subpage (or "settings")
 		if (paths.length > 0 && paths[paths.length-1]=="settings") paths.pop(); // remove "settings"
-		if (paths.length > 1) {
+		if (paths.length > 0) {
 			locproto = l.protocol;
 			loc = true;
 			locip = l.hostname + (l.port ? ":" + l.port : "") + "/" + paths.join('/');
--- a/wled00/data/common.js
+++ b/wled00/data/common.js
@@ -84,8 +84,14 @@ function getLoc() {
 		// detect reverse proxy
 		let path = l.pathname;
 		let paths = path.slice(1,path.endsWith('/')?-1:undefined).split("/");
-		if (paths.length > 1) paths.pop(); // remove subpage (or "settings")
-		if (paths.length > 0 && paths[paths.length-1]=="settings") paths.pop(); // remove "settings"
+
+		// remove settings paths to get the base url
+		const settingsIndex = paths.lastIndexOf("settings");
+		const pathLength = paths.length;
+		if (settingsIndex === pathLength - 1 || settingsIndex === pathLength - 2) {
+			paths = paths.slice(0, settingsIndex);
+		}
+
 		if (paths.length > 0) {
 			locproto = l.protocol;
 			loc = true;
--- a/wled00/data/common.js
+++ b/wled00/data/common.js
@@ -81,18 +81,10 @@ function getLoc() {
 			localStorage.setItem('locIp', locip);
 		}
 	} else {
-		// detect reverse proxy
-		let path = l.pathname;
-		let paths = path.slice(1,path.endsWith('/')?-1:undefined).split("/");
-
-		// remove settings paths to get the base url
-		const settingsIndex = paths.lastIndexOf("settings");
-		if (settingsIndex > 0) paths = paths.slice(0, settingsIndex);
-
-		if (paths.length && paths[paths.length - 1].includes(".")) {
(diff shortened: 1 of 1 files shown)


wled00/data/common.js
```diff
@@ -84,9 +84,22 @@ function getLoc() {
 		// detect reverse proxy
 		let path = l.pathname;
 		let paths = path.slice(1,path.endsWith('/')?-1:undefined).split("/");
-		if (paths.length > 1) paths.pop(); // remove subpage (or "settings")
-		if (paths.length > 0 && paths[paths.length-1]=="settings") paths.pop(); // remove "settings"
```
blazoncek, line 88: Try simply:
```js
		let settingsIndex = paths.lastIndexOf("settings");
		if (settingsIndex > 0) for (let i = paths.length; i > settingsIndex; i--) paths.pop(); // remove all after and including "settings"
```
instead.

K3das: Why use a for loop with pop instead of slice?

blazoncek: These 2 lines do everything you need.

K3das: That removes anything after and including `/settings/` which can also be done in two lines with slice and be much more readable:
```js
let settingsIndex = paths.lastIndexOf("settings");
if (settingsIndex > 0) paths.slice(0, settingsIndex);
```

That being said, the code in my diff checks that it's only the last or second to last instance of `settings` to maximally avoid conflicting with proxy paths including `settings` for some reason, but for what it's worth it doesn't make much of a difference and could probably be removed in favor of readability.

blazoncek: > in two lines with slice

Even better. I'm not a JS developer. 😉 
Learnings added. 🤣

blazoncek:
I've come up with this:
```js
let paths = l.pathname.split("/").slice(1); // first is always empty
let settingsIndex = paths.lastIndexOf("settings");
paths = paths.slice(0,settingsIndex); // if we don't have "settings", remove last entry (empty / or file)
```
Reasoning:
- main UI's path is always at `/` (reverse proxy: `/path/to/UI/`) and is terminated with `/` (`/index.htm` is invalid address)
- several auxiliary pages start as `/page.htm`
- any setting page will start with `/settings`

So, possible landing URL paths (that can call `common.js` and invoke `getLoc()`) are:
- /
- /settings (or /settings/)
- /settings/subpage (or /settings/subpage/)
- /file.htm
Which translate to:
- /path/to/UI/ *requires terminating /*
- /path/to/UI/settings (or /path/to/UI/settings/)
- /path/to/UI/settings/subpage (or /path/to/UI/settings/subpage/)
- /path/to/UI/file.htm

K3das:
Oooh, that's a really neat way to do it! I don't see why it wouldn't work.

wled00/data/common.js
```diff
@@ -81,12 +81,12 @@ function getLoc() {
 			localStorage.setItem('locIp', locip);
 		}
 	} else {
-		// detect reverse proxy
-		let path = l.pathname;
-		let paths = path.slice(1,path.endsWith('/')?-1:undefined).split("/");
-		if (paths.length > 1) paths.pop(); // remove subpage (or "settings")
-		if (paths.length > 0 && paths[paths.length-1]=="settings") paths.pop(); // remove "settings"
-		if (paths.length > 1) {
+		// guess the base url if WLED is behind a reverse proxy
```
blazoncek: Just strip trailing spaces...

blazoncek (approved):
... and you are good.

wled00/data/common.js
K3das: not sure how those got there, good catch

blazoncek:
Please do test as I currently do not have reverse proxy set up for WLED.

K3das:
works on my WLED instance where I was having issues :)

blazoncek:
Re "awaiting testing": Works on my end.

K3das:
Can this be merged?

K3das:
bump?

DedeHai:
@netmindz this might be more in your direction of expertise.

DedeHai:
@netmindz I have no idea what a reverse proxy even is. can you check if this makes sense?

blazoncek:
What a shame...

PR #4950, opened 2025-09-20
Hub75  fixes

netmindz:
## Summary by CodeRabbit

- New Features
  - Optional debug bus logging for HUB75 builds.

- Bug Fixes
  - Improved color rendering consistency on HUB75 matrix displays.

- Refactor
  - Removed per-pixel gamma undo to streamline display updates.
  - Panel chaining for HUB75 matrices temporarily disabled.

- Chores
  - Unified build configurations across ESP32/ESP32‑S3 HUB75 targets.
  - Updated LED type and pin handling for HUB75 support.

### DIFF
--- a/platformio_override.sample.ini
+++ b/platformio_override.sample.ini
@@ -550,7 +550,7 @@ build_unflags = ${common.build_unflags}
 build_flags = ${common.build_flags}
   -D WLED_RELEASE_NAME=\"ESP32_hub75\"
   -D WLED_ENABLE_HUB75MATRIX -D NO_GFX
-  -D LED_TYPES=65
+  -D LED_TYPES=65 -D DATA_PINS=64,64,1
   -D WLED_DEBUG_BUS
   ; -D WLED_DEBUG
 lib_deps = ${esp32_idf_V4.lib_deps}
@@ -566,7 +566,7 @@ build_flags = ${common.build_flags}
               -D WLED_RELEASE_NAME=\"ESP32_hub75_forum_pinout\"
               -D WLED_ENABLE_HUB75MATRIX -D NO_GFX 
               -D ESP32_FORUM_PINOUT ;; enable for SmartMatrix default pins
-              -D LED_TYPES=65
+              -D LED_TYPES=65 -D DATA_PINS=64,64,1
               -D WLED_DEBUG_BUS
 ; -D WLED_DEBUG
 
@@ -588,7 +588,7 @@ build_flags = ${common.build_flags} ${esp32s3.build_flags} -D WLED_RELEASE_NAME=
   -D WLED_ENABLE_HUB75MATRIX -D NO_GFX 
   -D S3_LCD_DIV_NUM=20 ;; Attempt to fix wifi performance issue when panel active with S3 chips
   -D ARDUINO_ADAFRUIT_MATRIXPORTAL_ESP32S3
-  -D LED_TYPES=65
+  -D LED_TYPES=65 -D DATA_PINS=64,64,1
   -D WLED_DEBUG_BUS
 
 
@@ -617,7 +617,7 @@ build_flags = ${common.build_flags} ${esp32s3.build_flags} -D WLED_RELEASE_NAME=
               -D WLED_ENABLE_HUB75MATRIX -D NO_GFX 
               -D S3_LCD_DIV_NUM=20 ;; Attempt to fix wifi performance issue when panel active with S3 chips
               -D MOONHUB_S3_PINOUT ;; HUB75 pinout
-              -D LED_TYPES=65
+              -D LED_TYPES=65 -D DATA_PINS=64,64,1
               -D WLED_DEBUG_BUS
 
 lib_deps = ${esp32s3.lib_deps}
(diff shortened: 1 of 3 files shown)


wled00/bus_manager.cpp
```diff
@@ -1021,9 +1021,6 @@ void __attribute__((hot)) BusHub75Matrix::setPixelColor(unsigned pix, uint32_t c
     if ((c == IS_BLACK) && (getBitFromArray(_ledsDirty, pix) == false)) return; // ignore black if pixel is already black
     setBitInArray(_ledsDirty, pix, c != IS_BLACK);                              // dirty = true means "color is not BLACK"
 
-    #ifndef NO_CIE1931
-    c = unGamma24(c); // to use the driver linear brightness feature, we first need to undo WLED gamma correction
```
netmindz, line 1025: @DedeHai  - I'm not sure if I'm making the right change here, I've not been able to follow all the ins and outs about where Gamma correction should be applied, in WLED-MM we do use NO_CIE1931, but with the current WLED I was seeing issues with this flag, but then we don't have an `unGamma24` in the current WLED codebase

DedeHai: I implemented "unGamma" but its either 8bit or 32bit. do HUB75 have built in gamma correction or why is that needed? if undoing gamma, colors will be all weird on linear panels.

DedeHai: but let me worry about that after this is in main, the whole brightness/gamma/ungamma took me a long time to fix properly. All I need to know is what the HUB75 driver does to raw colors.

DedeHai: i.e. does it scale the brightness or apply any gamma and if so can we disable it as that's what we need to do in WLED code.

DedeHai:
there are quite a few things that need an update, I will add a few remarks, I can fix brightness after this is merged, as that is easier for me than explaining it.

wled00/bus_manager.cpp
```diff
@@ -812,14 +812,15 @@ BusHub75Matrix::BusHub75Matrix(const BusConfig &bc) : Bus(bc.type, bc.start, bc.
   virtualDisp = nullptr;
 
   if (bc.type == TYPE_HUB75MATRIX_HS) {
```
DedeHai, line 814: this code needs to go belw, as it is in MM. cant create a virtual panel from a null pointer.

wled00/bus_manager.cpp
```diff
@@ -812,14 +812,15 @@ BusHub75Matrix::BusHub75Matrix(const BusConfig &bc) : Bus(bc.type, bc.start, bc.
   virtualDisp = nullptr;
```
DedeHai, line 812: please add aliases for the bc.pins[x] at the top, the code is very hard to follow and highly error prone with just indices.

platformio_override.sample.ini
```diff
@@ -563,11 +564,14 @@ board_build.flash_mode = dio
 extends = env:esp32dev_hub75
 build_flags = ${common.build_flags}
               -D WLED_RELEASE_NAME=\"ESP32_hub75_forum_pinout\"
-              -D WLED_ENABLE_HUB75MATRIX -D NO_GFX -D NO_CIE1931
+              -D WLED_ENABLE_HUB75MATRIX -D NO_GFX 
               -D ESP32_FORUM_PINOUT ;; enable for SmartMatrix default pins
+              -D LED_TYPES=65 -D DATA_PINS=64,64,1
```
DedeHai, line 569: a comment that DATA_PINS means x,y,chainlength would be nice

wled00/bus_manager.cpp
netmindz: By default the driver is already non linear, check the docs for the NO_CIE1931 flag

platformio_override.sample.ini
```diff
@@ -581,9 +585,12 @@ build_flags = ${common.build_flags} ${esp32s3.build_flags} -D WLED_RELEASE_NAME=
...
-  -D WLED_ENABLE_HUB75MATRIX -D NO_GFX -D NO_CIE1931
+  -D WLED_ENABLE_HUB75MATRIX -D NO_GFX 
   -D S3_LCD_DIV_NUM=20 ;; Attempt to fix wifi performance issue when panel active with S3 chips
   -D ARDUINO_ADAFRUIT_MATRIXPORTAL_ESP32S3
+  -D LED_TYPES=65 -D DATA_PINS=64,64,1
+  -D WLED_DEBUG_BUS
+
 
 lib_deps = ${esp32s3.lib_deps}
   ${esp32.AR_lib_deps}
```
DedeHai, line 596: AR_lib_deps is deprecated.

platformio_override.sample.ini
```diff
@@ -607,9 +614,11 @@ build_flags = ${common.build_flags} ${esp32s3.build_flags} -D WLED_RELEASE_NAME=
               -DLOLIN_WIFI_FIX ; seems to work much better with this
               -D WLED_WATCHDOG_TIMEOUT=0
               ${esp32.AR_build_flags}
```
DedeHai, line 616: AR_build_flags is deprecated as well

DedeHai:
as you know, something is still wrong with the config. I could not get it to save at all, it would just crash.

netmindz:
> as you know, something is still wrong with the config. I could not get it to save at all, it would just crash.

That is not the behaviour I was seeing. I could update panel dimensions and chain length correctly

PR #4951, opened 2025-09-21
fix buffer size calculation for ESP8266 (0.15 edition)

DedeHai:
- also add proper global buffer calculation in settings 
- fix RMT/I2S channel calculation

this is a back-port from @blazoncek fork with some additions


DedeHai:
@coderabbitai review

PR #4980, opened 2025-10-04
(0.15) RMT High-priority Interrupt driver backport

willmmiles:
Backport of #4890 for 0.15.  Solves RMT flickering issues on XTensa platforms.


softhack007:
@willmmiles one question before I merge this into WLED-MM
we are still on NeoPixelBus 2.7.5 - do you think your patch will work with that version, too?

Or maybe not needed for NeoPixelBus < 2.8.3 ?

softhack007:
@coderabbitai this PR was created with NeoPixelBus release 2.8.3 in mind. Would it also help to remove flickering artefacts when using  NeoPixelBus release 2.7.5?

PR #4985, opened 2025-10-05
Correct broken esp32dev_8M enviroment

wled-compile:
esp32dev_8M: add flash_mode, otherwise it cannot be compiled

## Summary by CodeRabbit

- Bug Fixes
  - Updated ESP32 8M build configuration to use DIO flash mode, improving firmware flashing reliability and reducing intermittent boot issues on affected devices.

- Chores
  - Adjusted build settings for ESP32 8M environment. No changes to user-facing features or interfaces.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -450,7 +450,7 @@ board_build.partitions = ${esp32.large_partitions}
 board_upload.flash_size = 8MB
 board_upload.maximum_size = 8388608
 ; board_build.f_flash = 80000000L
-; board_build.flash_mode = qio
+board_build.flash_mode = dio
 
 [env:esp32dev_16M]
 board = esp32dev


platformio.ini
```diff
@@ -450,7 +450,7 @@ board_build.partitions = ${esp32.large_partitions}
 board_upload.flash_size = 8MB
 board_upload.maximum_size = 8388608
 ; board_build.f_flash = 80000000L
-; board_build.flash_mode = qio
+board_build.flash_mode = dio
```
netmindz, line 453: Did qio fail or are you just choosing the slower dio as a safer option in case the device doesn't support qio?

wled-compile:
Other esp32dev enviroments (esp32dev, esp32dev_16M) have already dio. Therefore, to be consistent (and most compatible) i set dio

blazoncek:
Most compatible would be DOUT.

wled-compile:
dio is also defined in board definition file. Actually platformio should use it, it also work with original arduino-espressif32 (for example wled 0.15.x, but does not work (and requires explicit definition in enviroment) when tasmota core is used (not tested if it is specific for version tasmote 2023.06.02 WLED uses or generally). Can nyone confirm it? Maybe something is broken for me or I do something wrong.

blazoncek:
FYI flash chip not capable of QIO mode would be low performance and not suitable for large LED set-ups where flash (and PSRAM) speed is mandatory.

Of course one can override compile time flash mode during ESPTOOL upload/flashing.

softhack007:
OK to reduce speed.
Normally I would expect that esp32 with 8MB or 16MB have an "external" Flash chip, and this should be "qio" according to the datasheet (see below).

However you'll never know what surprises the next *cheap esp32* will bring, so maybe better safe (dio) than sorrry (qio).

image

PR #4991, opened 2025-10-08
Fix: InvalidProjectConfError 'No option 'platform_packages' in section: 'esp32s2'

### LINKED ISSUE #4829: InvalidProjectConfError: Invalid 'wled/src/platformio.ini' (project configuration file): 'No option 'platform_packages' in section: 'esp32s2''
### What happened?

I copied `platformio_override.sample.ini` to `platformio_override.ini` and modified the settings to match my board, wich is ESP8266 based.

### To Reproduce Bug

1.  cp  platformio_override.sample.ini platformio_override.ini` 
2.  pio run

### Expected Behavior

I would expect that the `platformio_override.sample.ini` is maintained and valid. Unfortunately it is not updated if platform settings in other files are renamed or changed.

### Install Method

Self-Compiled

### What version of WLED?

HEAD of main

ESP8266

### Relevant log/trace output

```shell

```

### Anything else?

InvalidProjectConfError: Invalid 'wled/src/platformio.ini' (project configuration file): 'No option 'platform_packages' in section: 'esp32s2''

### Code of Conduct

itCarl:
Hi,
i tried to fix #4829 with this. Hope this helps.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -285,6 +285,7 @@ lib_deps =
 [esp32s2]
 ;; generic definitions for all ESP32-S2 boards
 platform = ${esp32_idf_V4.platform}
+platform_packages =
 build_unflags = ${common.build_unflags}
 build_flags = -g
   -DARDUINO_ARCH_ESP32
@@ -303,6 +304,7 @@ board_build.partitions = ${esp32.default_partitions}   ;; default partioning for
 [esp32c3]
 ;; generic definitions for all ESP32-C3 boards
 platform = ${esp32_idf_V4.platform}
+platform_packages =
 build_unflags = ${common.build_unflags}
 build_flags = -g
   -DARDUINO_ARCH_ESP32


eku:
Nice try, but this does not solve the builds that require these values to be non empty.

itCarl:
> Nice try, but this does not solve the builds that require these values to be non empty.

Have a look at the `esp32` section. There is also a empty `platform_packages` parameter. So guess it does not have to be filled.

Can u tell me what builds need the `platform_packages`? I might have a look at them.

itCarl:
It seems that `platform_packages` were already present a while ago. 
Got first added here:
image
then it changed couple of times over the years.

and here it got removed:
image

I also think a guy alredy tried to fix this config issue but his pr got rejected. #4830

softhack007:
> Nice try, but this does not solve the builds that require these values to be non empty.

@itCarl looks like I've fixed this one already in main
(without knowing about your PR or #4829 ) with commit https://github.com/wled/WLED/commit/a73a2aaa335f0fb8248116cca952dfa121f4be0f

Can you check if this solves the problem?

PR #4994, opened 2025-10-09
Arc Expand Optimizations

Brandon502:
New Arc algorithm that is more optimized and fixes holes on larger matrices (32x8 64x32 etc). FPS on 64x32 Rainbow went from 45 FPS to 80 FPS on esp32dev hub75 build. Looks slightly different but unnoticeable on most effects. gPC is more accurate and seems to fix the last few effects that still use it. 

 Videos comparing the new (reversed) and previous version. Left is new, right is old.

https://github.com/user-attachments/assets/f974f134-6cc0-4785-abff-7e5e0002e223

https://github.com/user-attachments/assets/72fbba36-3844-4a24-a1f7-7290817c7e3a

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved arc rendering consistency on 2D layouts, removing early-frame square artifacts and reducing off-by-one diagonal distortions.
  * More accurate handling of circle-edge cases for precise pixel mapping.
* **Refactor**
  * Replaced previous sampling with a more efficient circle-based arc-drawing approach for reduced computation and improved responsiveness.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -663,7 +663,7 @@ uint16_t Segment::virtualLength() const {
         vLen = max(vW,vH); // get the longest dimension
         break;
       case M12_pArc:
-        vLen = sqrt32_bw(vH*vH + vW*vW); // use diagonal
+        vLen = (sqrt32_bw((vH*vH + vW*vW) << 2) + 1) >> 1; // diagonal rounded
         break;
       case M12_sPinwheel:
         vLen = getPinwheelLength(vW, vH);
@@ -742,36 +742,25 @@ void WLED_O2_ATTR Segment::setPixelColor(int i, uint32_t col) const
         if (vStrip > 0)                   setPixelColorRaw(XY(vStrip - 1, vH - i - 1), col);
         else for (int x = 0; x < vW; x++) setPixelColorRaw(XY(x, vH - i - 1), col);
         break;
-      case M12_pArc:
+      case M12_pArc: {
         // expand in circular fashion from center
-        if (i == 0)
-          setPixelColorRaw(XY(0, 0), col);
-        else {
-          float r = i;
-          float step = HALF_PI / (2.8284f * r + 4); // we only need (PI/4)/(r/sqrt(2)+1) steps
-          for (float rad = 0.0f; rad <= (HALF_PI/2)+step/2; rad += step) {
-            int x = roundf(sin_t(rad) * r);
-            int y = roundf(cos_t(rad) * r);
-            // exploit symmetry
+        if      (i == 0) setPixelColorXY(0, 0, col);
+        else if (i == 2) setPixelColorXY(1, 1, col);
+
+        int x = 0, y = i;  // i is the radius
+        int d = -(i >> 1); // Initial decision parameter
+
+        // Barrera's circle algorithm
+        while (x <= y) {
+          if (!(i == x && i == y)) { // prevent early square
             setPixelColorXY(x, y, col);
             setPixelColorXY(y, x, col);
           }
-          // Bresenham’s Algorithm (may not fill every pixel)
-          //int d = 3 - (2*i);
-          //int y = i, x = 0;
-          //while (y >= x) {
-          //  setPixelColorXY(x, y, col);
-          //  setPixelColorXY(y, x, col);
-          //  x++;
-          //  if (d > 0) {
-          //    y--;
-          //    d += 4 * (x - y) + 10;
-          //  } else {
-          //    d += 4 * x + 6;
-          //  }
-          //}
+          if (d <= 0) d += ++x;
+          else d -= --y;
         }
         break;
+      }
       case M12_pCorner:
         for (int x = 0; x <= i; x++) setPixelColorXY(x, i, col); // note: <= to include i=0. Relies on overflow check in sPC()
         for (int y = 0; y <  i; y++) setPixelColorXY(i, y, col);
@@ -947,9 +936,16 @@ uint32_t WLED_O2_ATTR Segment::getPixelColor(int i) const
         else            { y = vH - i - 1; };
         break;
       case M12_pArc:
-        if (i > vW && i > vH) {
-          x = y = sqrt32_bw(i*i/2);
-          break; // use diagonal
+        if (i >= vW && i >= vH) { // Barrera's circle algorithm
+          int x2 = 0, y2 = i, d = -(i >> 1);
+          int validCount = 0; // return 2nd non mirrored pixel if available
+          while (x2 <= y2 && validCount < 2) {
+            if      (y2 < vH && x2 < vW) {x = x2, y = y2; validCount++;}
+            else if (y2 < vW && x2 < vH) {x = y2, y = x2; validCount++;}
+
+            if (d <= 0) d += ++x2; else d -= --y2;
+          }
+          break;
         }
         // otherwise fallthrough
       case M12_pCorner:
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -743,18 +743,21 @@ void WLED_O2_ATTR Segment::setPixelColor(int i, uint32_t col) const
         else for (int x = 0; x < vW; x++) setPixelColorRaw(XY(x, vH - i - 1), col);
         break;
       case M12_pArc: {
-        // expand in circular fashion from center
-        if      (i == 0) setPixelColorXY(0, 0, col);
-        else if (i == 2) setPixelColorXY(1, 1, col);
+        // expand in circular fashion from top left corner
+        // Tony Barrera's circle algorithm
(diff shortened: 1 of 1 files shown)


blazoncek:
I find this *very* interesting! Thank you.
Unfortunately you do not show the behaviour on non-square segments. As mentioned by @coderabbitai.

wled00/FX_fcn.cpp
```diff
@@ -663,7 +663,7 @@ uint16_t Segment::virtualLength() const {
         vLen = max(vW,vH); // get the longest dimension
         break;
       case M12_pArc:
-        vLen = sqrt32_bw(vH*vH + vW*vW); // use diagonal
+        vLen = (sqrt32_bw((vH*vH + vW*vW) << 2) + 1) >> 1; // diagonal rounded
```
blazoncek, line 666: I don't get it.
Mathematically this translates to: (sqrt(vH<sup>2</sup> + vW<sup>2</sup>) * 2 + 1) / 2 so why not write it simply as that?

I.e.: `vLen = ((sqrt32_bw(vH*vH + vW*vW) << 1) + 1) >> 1;`

wled00/FX_fcn.cpp
```diff
@@ -742,36 +742,25 @@ void WLED_O2_ATTR Segment::setPixelColor(int i, uint32_t col) const
...
-          setPixelColorRaw(XY(0, 0), col);
-        else {
-          float r = i;
-          float step = HALF_PI / (2.8284f * r + 4); // we only need (PI/4)/(r/sqrt(2)+1) steps
-          for (float rad = 0.0f; rad <= (HALF_PI/2)+step/2; rad += step) {
-            int x = roundf(sin_t(rad) * r);
-            int y = roundf(cos_t(rad) * r);
-            // exploit symmetry
+        if      (i == 0) setPixelColorXY(0, 0, col);
+        else if (i == 2) setPixelColorXY(1, 1, col);
```
blazoncek: Is this a fix for anomaly? I.e. missing pixel.

wled00/FX_fcn.cpp
```diff
@@ -742,36 +742,25 @@ void WLED_O2_ATTR Segment::setPixelColor(int i, uint32_t col) const
...
+
+        int x = 0, y = i;  // i is the radius
+        int d = -(i >> 1); // Initial decision parameter
+
+        // Barrera's circle algorithm
+        while (x <= y) {
+          if (!(i == x && i == y)) { // prevent early square
```
blazoncek: Perhaps change NAND to negated input OR? It may be a tiny bit faster.
I.e. `i != x || i != y`

wled00/FX_fcn.cpp
```diff
@@ -742,36 +742,25 @@ void WLED_O2_ATTR Segment::setPixelColor(int i, uint32_t col) const
...
-            int x = roundf(sin_t(rad) * r);
-            int y = roundf(cos_t(rad) * r);
-            // exploit symmetry
+        if      (i == 0) setPixelColorXY(0, 0, col);
+        else if (i == 2) setPixelColorXY(1, 1, col);
+
+        int x = 0, y = i;  // i is the radius
+        int d = -(i >> 1); // Initial decision parameter
+
+        // Barrera's circle algorithm
```
blazoncek: The resources for Barrera's algorithm are very scarce, perhaps add a link to description?

wled00/FX_fcn.cpp
```diff
@@ -947,9 +936,16 @@ uint32_t WLED_O2_ATTR Segment::getPixelColor(int i) const
         else            { y = vH - i - 1; };
         break;
       case M12_pArc:
-        if (i > vW && i > vH) {
-          x = y = sqrt32_bw(i*i/2);
-          break; // use diagonal
+        if (i >= vW && i >= vH) { // Barrera's circle algorithm
```
blazoncek, line 942: As this part is only relevant when `i` is beyond longest dimension (vW or vH), i.e. covering "corner" case of arc expansion, why not just use `sqrt(i*i/2)`? Reasoning: The larger the `i` the smaller the error.

Brandon502:
> I find this _very_ interesting! Thank you. Unfortunately you do not show the behaviour on non-square segments. As mentioned by @coderabbitai.

Here's a video of both versions of arc running 64x32 freqmatrix, 64x22 rainbow, and 64x8 dj lights. You can see the hole in the larger sizes on the old version and gPC breaks in the corner. New gPC works similar to old and uses corner when it can. I think there's still a chance that it won't work on the very last pixel due to unneeded rounding on certain sizes. I don't think any effects currently rely on going backwards. I can add condition in sPC that always paints the bottom right pixel when i == vLength() -1 to fix though if necessary. 

https://github.com/user-attachments/assets/193c7510-74d0-4f9f-a535-131ea290fcdb


https://github.com/user-attachments/assets/d2b05112-1776-4a5d-85fe-5d7da15ed4dc

blazoncek (commented):
As already mentioned I am fascinated by this. So I did analyze it.
I hope this gets in, in some way or another.

wled00/FX_fcn.cpp
Brandon502: I'll have to test this later to be sure, but a quick AI search shows that your version may be off by one in certain cases.

blazoncek: Mathematically speaking: sqrt(a*b) == sqrt(a) * sqrt(b)
EDIT: sqrt(a * 4) == sqrt(a)*2
😁

wled00/FX_fcn.cpp
Brandon502: These cases and the if (!(i == x && i == y) line are to avoid this early square pattern I noticed when mirroring. 

https://github.com/user-attachments/assets/51fe8e81-17c3-420a-ae44-719d6ec9e390

I initially hard coded when i < 3 and skipped loop, but the current version was faster.

wled00/FX_fcn.cpp
Brandon502: Yes that should work. If you want it to be even faster adding bounds checks and using sPCRaw was about 5% faster if I remember correctly.

wled00/FX_fcn.cpp
Brandon502: https://softwareengineering.stackexchange.com/questions/287478/drawing-concentric-circles-without-gaps/357445#357445 is where I found the algorithm mentioned. The source link is dead unfortunately.

wled00/FX_fcn.cpp
Brandon502: Use x = y = sqrt32_bw(i*i/2); version of gPC? It didn't work correctly. Some circles over write previous circles.

wled00/FX_fcn.cpp
Brandon502: I agree they look the same, but the results end up being different. Just tested. a my version, b your version
vW=64 vH=32 a=72 b=71

Actual sqrt value is 71.55

Brandon502: For 64x32 71 actually works without holes. But certain rectangles need the rounding to fill the bottom right corner pixel(s)

wled00/FX_fcn.cpp
blazoncek: That's what I did in my fork.

wled00/FX_fcn.cpp
blazoncek: Ah! Integer math.

wled00/FX_fcn.cpp
blazoncek: > Some circles over write previous circles

Then gPC will return incorrect value in any case. Won't it? The only question is, when.

wled00/FX_fcn.cpp
blazoncek: By all means, include it in the code. As a reference.

wled00/FX_fcn.cpp
blazoncek: This prompted me to revise my previous code. It does have a bug. Thanks!
This (or something like this with correct fix for integer math) would be the correct implementation:
```c++
        if (i > vW && i > vH) {
          // we can't use diagonal if segment is not square
          if (vW != vH) {
            if (vW > vH) { y = vH - 1; x = sqrt16(i*i - vH*vH); }
            else         { x = vW - 1; y = sqrt16(i*i - vW*vW); }
            break;
          } else x = y = sqrt16(i*i/2);
          break;
        }
        // otherwise fallthrough
```

Brandon502: I tried a few different equations for gPC and couldn't get it to work as good as my solution. They kind of work in one direction but you can see them speed up when they start retrieving the wrong pixel. The current way isn't 100% perfect either, but works in both directions for all sizes that I've tried and when it gets the wrong pixel it doesn't have as noticeable a change.

blazoncek: You are correct! Mathematical assumptions are not always valid when doing integer math.
It is necessary to reverse-plot coordinates.

blazoncek (approved):
Approved from my side.

DedeHai (approved):

PR #4997, opened 2025-10-10
Adding DDP over WS, moving duplicate WS-connection to common.js

DedeHai:
- Enabling DDP over WebSocket: this allows for html pages to stream data to the LEDs much faster than through the JSON API
- Moved the duplicate function to establish a WS connection from the live-view htm files to common.js

uses only 100bytes of additional flash and is a feature I could make use of in the [matrix tool](https://github.com/wled/WLED/pull/4982). 
Also the new custom palette editor has a "preview" feature that currently uses the JSON API to do a live-preview on the LEDs. Using DDP instead would make it much faster.
I ran some experiments doing video streaming from a ESP-Hosted webpage to the LEDs and this also did work quite well.

## Summary by CodeRabbit

- New Features
  - Live View adds binary WebSocket protocol support and a reusable WebSocket connector for simpler integration.

- Performance
  - LED data is sent in chunked frames to reduce overhead and improve update latency.

- Compatibility
  - Unified and more consistent WebSocket initialization and message handling across views.

- Bug Fixes
  - Stronger packet validation and bounds checks to prevent malformed binary data issues.

- Refactor
  - Consolidated WebSocket setup into a shared client script for reliability.

### DIFF
--- a/wled00/data/common.js
+++ b/wled00/data/common.js
@@ -156,7 +156,7 @@ function sendDDP(ws, start, len, colors) {
 		pkt[2] = 0x00; // reserved
 		pkt[3] = 0x01; // 1 = RGB (currently only supported mode)
 		pkt[4] = 0x01; // destination id (not used but 0x01 is default output)
-		pkt[5] = (off >> 24) & 255; // DDP protocl 4-7 is offset
+		pkt[5] = (off >> 24) & 255; // DDP protocol 4-7 is offset
 		pkt[6] = (off >> 16) & 255;
 		pkt[7] = (off >> 8) & 255;
 		pkt[8] = off & 255;


wled00/ws.cpp
```diff
@@ -71,6 +71,15 @@ void wsEvent(AsyncWebSocket * server, AsyncWebSocketClient * client, AwsEventTyp
           // force broadcast in 500ms after updating client
           //lastInterfaceUpdate = millis() - (INTERFACE_UPDATE_COOLDOWN -500); // ESP8266 does not like this
         }
+      }else if (info->opcode == WS_BINARY) {
+        // interpreted binary data using DDP protocol
+        if (len < 10) return; // DDP header is 10 bytes
+        size_t ddpDataLen = (data[8] << 8) | data[9]; // data length in bytes from DDP header
+        if (len < (10 + ddpDataLen)) return; // not enough data, prevent out of bounds read
+        // could be a valid DDP packet, forward to handler
+        IPAddress clientIP = client->remoteIP();
```
blazoncek: `realtimeIP = client->remoteIP();`

wled00/ws.cpp
```diff
@@ -71,6 +71,15 @@ void wsEvent(AsyncWebSocket * server, AsyncWebSocketClient * client, AwsEventTyp
...
+        e131_packet_t *p = (e131_packet_t*)data;
+        handleE131Packet(p, clientIP, P_DDP);
```
blazoncek: `handleDDPPacket((e131_packet_t*)data);`

DedeHai: I did that at first, then wanted to support the other protocols and realized they cannot be destinguished reliably just by the data and additional variables over JSON would be needed. Since I only have use for DDP I dropped the others but kept the function call to set the IP, but setting the IP directly and calling handleDDPPacket makes a lot more sense. Thanks for the review.

wled00/e131.cpp
```diff
@@ -30,11 +30,19 @@ void handleDDPPacket(e131_packet_t* p) {
...
   uint8_t* data = p->data;
   unsigned c = 0;
   if (p->flags & DDP_TIMECODE_FLAG) c = 4; //packet has timecode flag, we do not support it, but data starts 4 bytes later
 
+  unsigned numLeds = stop - start; // stop >= start is guaranteed
+  unsigned maxDataIndex = c + numLeds * ddpChannelsPerLed; // validate bounds before accessing data array
+  if (maxDataIndex > dataLen) {
+    DEBUG_PRINTLN(F("DDP packet data bounds exceeded, rejecting."));
+    return;
+  }
```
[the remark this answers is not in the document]
DedeHai: there is no need to check the spec limit, we do accept out of spec packets assuming they contain correct data, it is up to the sender to ensure correct encoding, we just need to make sure not to access out of bound.
the max data length checking includes the 4 bytes so checking beforehand is redundant and therefore omitted.

wled00/data/common.js
```diff
@@ -116,3 +116,52 @@ function uploadFile(fileObj, name) {
...
+
+	getLoc(); // ensure globals (loc, locip, locproto) are up to date
+	let url = loc ? getURL('/ws').replace("http","ws") : "ws://"+window.location.hostname+"/ws";
+	let ws = new WebSocket(url);
+	ws.binaryType = "arraybuffer";
+	if (onOpen) { ws.onopen = onOpen; }
+	return ws;
+}
+
+// send led colors to ESP using WebSocket and DDP protocol (RGB)
```
netmindz: LED

wled00/e131.cpp
netmindz, line 1: Your changes here are just improvements rather than required for the WS support?

wled00/ws.cpp
```diff
@@ -71,6 +71,15 @@ void wsEvent(AsyncWebSocket * server, AsyncWebSocketClient * client, AwsEventTyp
...
         }
+      }else if (info->opcode == WS_BINARY) {
+        // interpreted binary data using DDP protocol
+        if (len < 10) return; // DDP header is 10 bytes
+        size_t ddpDataLen = (data[8] << 8) | data[9]; // data length in bytes from DDP header
+        uint8_t flags = data[0];
+        if ((flags & DDP_TIMECODE_FLAG) ) ddpDataLen += 4; // timecode flag adds 4 bytes to data length
+        if (len < (10 + ddpDataLen)) return; // not enough data, prevent out of bounds read
+        // could be a valid DDP packet, forward to handler
+        handleE131Packet((e131_packet_t*)data, client->remoteIP(), P_DDP);
```
netmindz: This is the only bit that makes me twitchy, the assumption that *any* binary messages are going to be DDP
I would prefer we had a way to indicate what this data is, with DDP being the first (and only) supported type rather than assume that we can "detect" the format and do some magic later when there are more than DDP being send as binary

wled00/e131.cpp
DedeHai: yes, this adds a safety check which honestly I am surprised did not cause any issues: if a DDP packet states "100 LEDs" but data contains less, the ESP will just crash due to OOB array access.

wled00/ws.cpp
DedeHai: I am no expert in protocol definition. How to solve it?
Detecting protocols by header as it is (crudely and incompletely) implemented is the simplest version. A AR sync packet could be detected that way as well. Adding another "header" or "request" layer on top could do more harm than good: what happens if two clients connect, one sending DDP one sending AR data for example? 
edit:
if in the future, custom needs arise for a completely different protocol, it could also use a header OR do a UI config override: if "some special protocol" is used, disable all other protocol support.

if past experience has tought me anything it is not to try to predict the future of data usage ;)

netmindz: I've just checked the WebSocket protocol docs and you can send a Sec-WebSocket-Protocol http header, which if we used that then you also have a mechanism to know if the server supports what you are trying to send. 
That doesn't however resolve the challenge of knowing when you get a binary message, which function to delegate to. 
While I take your point about predicting the future is hard, at the same point, writing something in a way that prevents further extension later is bad. I still think it would be worth adding a single byte at the start that indicates what the message type is

netmindz: i.e the first byte maps to the function, which might itself support multiple messages

DedeHai: adding a header byte would work for me and since no applications exist yet that use binary streaming over WS there are no compatibility issues either. should I implement this approach? say something like:
0 = generic (future implementation or auto detect from following bytes?)
1 = DDP
2 - 9 = reserved for other LED streaming protocols
11 = AR packet (not yet supported)

wled00/data/common.js
netmindz: Typo

wled00/ws.cpp
netmindz: Not quite sure how auto detection would work, but reserving 0 for that sounds a good idea, beyond that do we already have any constant that identifiers DDP?
I would avoid trying to allocate any particular values for future use, the AR example was just the first hypothetical that came to mind, I'm not sure it would be a good idea as latency is more important than guaranteed delivery, so possibly inadvisable, but possibly only option of anyone was to want to try and do something entirely in the browser.
There might be other usermods however that would benefit from high speed TCP rather than hammering the API

DedeHai: > Not quite sure how auto detection would work

for DDP or AR packets for example by looking at each packets header and finding unique identifiers, but this was just a thought, no real usecase in mind. I will go ahead and implement a minimal version

netmindz: You definitely could do some form of auto detection, my query was more how to ensure proper encapsulation and separation of concerns within the code if we did that.
For now let's just keep to the simple if ID btye indicates DDP, delegate to the relevant function, otherwise error

netmindz:
Last two rabbit suggestions look fair, but otherwise I think we are good to merge

netmindz (approved):

wled00/ws.cpp
softhack007, line 1: @DedeHai one question: I would like to keep WLED-MM ready (and compatible) with your work in progress on HTML/JS tools. Would it be sufficient if I merge changes in ws.cpp into my fork? Especially, leaving the UI scripts in data/ as they are? 

(I'm not good in JS, but the changes to ws.cpp seem to be well isolated and compatible)

DedeHai: I am still a beginner in JS, I just learned how to trick AI's into creating the code I want ;)
yes, you basically only need the ws.cpp code addition to forward binary packets as DDP. The changes I made to the e131.cpp are bugfixes to prevent buffer overruns but unrelated to the DDP over WS support.
the JS part I changed is just to simplify WS initialization for other tools but so far, none of my tools use that. I might add a "live preview" to the Matrix tool that will use DDP streaming, in that case you would either need to disable this function in MM, port the ws-init function to common.js or copy the ws-init code into the matrix-tool htm file.

softhack007: Ah OK, thanks 👍 
I'll start with the code in ws.cpp, Let's see how important the new "live preview in matrix tool" will be, maybe i'll take the pain to port JS functions over to WLED-MM :-) 

Or you tell me - once its ready - how I can disable the preview part without breaking everything.

PR #5001, opened 2025-10-13
Drag and drop LED output reordering

blazoncek:
Ability to reorder LED outputs.
May allow users to update/modify otherwise unmodifiable output.

## Summary by CodeRabbit

- New Features
  - Drag-and-drop reordering of LED outputs in Settings with automatic reindexing of blocks, labels and IDs.
  - Dynamic bus-type availability: LED/bus type options update across selectors based on current configuration.

- Refactor
  - Internal per-LED selector and initialization logic reorganized for consistent behavior and reliable reindexing.

- Bug Fixes
  - UI text/spacing tweaks and updated min/max constraints for per‑LED settings.

### DIFF
--- a/wled00/data/settings_leds.htm
+++ b/wled00/data/settings_leds.htm
@@ -227,45 +227,6 @@
 			let gRGBW = false, memu = 0;
 			let busMA = 0;
 			let sLC = 0, sPC = 0, sDI = 0, maxLC = 0;
-			const abl = d.Sf.ABL.checked;
-			let setPinConfig = (n,t) => {
-				let p0d = "GPIO: ";
-				let p1d = "";
-				let off = "Off Refresh";
-				switch (gT(t).t.charAt(0)) {
-					case '2': // 2 pin digital
-						p1d = "Clock "+p0d;
-						// fallthrough
-					case 'D': // digital
-						p0d = "Data "+p0d;
-						break;
-					case 'A': // PWM analog
-						if (numPins(t) > 1) p0d = "GPIOs: ";
-						off = "Dithering";
-						break;
-					case 'N': // network
-						p0d = "IP address: ";
-						break;
-					case 'V': // virtual/non-GPIO based
-						p0d = "Config: "
-						break;
-					case 'H': // HUB75
-						p0d = "Panel size (width x height), Panel count:"
-						break;
-				}
-				gId("p0d"+n).innerText = p0d;
-				gId("p1d"+n).innerText = p1d;
-				gId("off"+n).innerText = off;
-				// secondary pins show/hide (type string length is equivalent to number of pins used; except for network and on/off)
-				let pins = Math.max(gT(t).t.length,1) + 3*isNet(t) + 2*isHub75(t); // fixes network pins to 4
-				for (let p=1; p<5; p++) {
-					var LK = d.Sf["L"+p+n];
-					if (!LK) continue;
-					LK.style.display = (p < pins) ? "inline" : "none";
-					LK.required = (p < pins);
-					if (p >= pins) LK.value="";
-				}
-			}
 
 			// enable/disable LED fields
 			updateTypeDropdowns(change);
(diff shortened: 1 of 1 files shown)


DedeHai:
I tried this but what is the purpose?
- if I move outputs around, it does not allow me to change the LED type
- does not work on mobile 

can you elaborate?

blazoncek:
You can change only the last one, however now you can choose which is the last.
Doing it on mobile would require a hell-of-a-good-dexterity. Not intended for mobile.

DedeHai:
On a C3 I can not change the last one:
image

blazoncek:
You found a bug. 🤦‍♂️

netmindz:
Do we need reordering as a feature or is this a workaround for the inability to edit all values of all outputs?

blazoncek:
> Do we need reordering as a feature or is this a workaround for the inability to edit all values of all outputs?

You are welcome to find a solution how to dynamically enable certain LED types to prevent questions like "I've added 3 WS2812 outputs to my C3 but only 2 work. How come?"

You can, however, just remove the line I changed in last commit if you don't want such protection. It is responsible for disabling select box.

Disregarding above, there was a request long time ago to allow simpler reordering outputs.

wled00/data/settings_leds.htm
```diff
@@ -438,38 +441,28 @@
 		function lastEnd(i) {
 			if (i-- < 1) return 0;
 			var s = chrID(i);
-			v = parseInt(d.getElementsByName("LS"+s)[0].value) + parseInt(d.getElementsByName("LC"+s)[0].value);
-			var t = parseInt(d.getElementsByName("LT"+s)[0].value);
+			v = parseInt(gN("LS"+s).value) + parseInt(gN("LC"+s).value);
+			var t = parseInt(gN("LT"+s).value);
 			if (isPWM(t)) v = 1; //PWM busses
 			return isNaN(v) ? 0 : v;
```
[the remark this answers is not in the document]
blazoncek: @coderabbitai check again. You are mis-referencing here.

wled00/data/settings_leds.htm
```diff
@@ -521,23 +514,16 @@
...
-				let sel = d.getElementsByName("LT"+s)[0];
-				// 32 & S2 supports mono I2S as well as parallel so we need to take that into account; S3 only supports parallel
-				let maxDB = maxD - (is32() || isS2() || isS3() ? (!d.Sf["PR"].checked)*8 - (!isS3()) : 0); // adjust max digital buses if parallel I2S is not used
-				if (digitalB >= maxDB) disable(sel,'option[data-type="D"]'); // NOTE: see isDig()
-				if (twopinB  >= 2)     disable(sel,'option[data-type="2P"]'); // NOTE: see isD2P() (we will only allow 2 2pin buses)
-				disable(sel,`option[data-type^="${'A'.repeat(maxA-analogB+1)}"]`); // NOTE: see isPWM()
-				sel.selectedIndex = sel.querySelector('option:not(:disabled)').index;
+				enLA(gN("LL"+s)); // update LED mA
+				let sel = gN("LT"+s);
+				sel.selectedIndex = sel.querySelector('option:not([data-type])').index; // select On/Off by default
```
DedeHai: if you add
``` JS
updateTypeDropdowns(); // update valid bus options including this new one
sel.selectedIndex = sel.querySelector('option:not(:disabled)').index; // select first valid option
```
like in my original implementation, the current behaviour of selecting the first available is preserved

wled00/data/settings_leds.htm
```diff
@@ -812,6 +835,36 @@
 			}
 			return opt;
 		}
+		// dynamically enforce bus type availability based on current usage
+		function updateTypeDropdowns(change=false) {
```
DedeHai: remove the unused `change` parameter or add comment on its intent

wled00/data/settings_leds.htm
```diff
@@ -408,6 +410,7 @@
 			gId("pc").textContent = (sLC == sPC) ? "":"(" + sPC + " physical)";
 
 			// memory usage and warnings
+			memu += sLC * (4 + (gN("MS").checked ? 0 : 4)); // pixel buffer (4 bytes per pixel) + at least 1 segment buffer
```
DedeHai: this is completely unrelated to this PR and breaks the currently intended behaviour of memory calculation. while this is technically correct, it needs a proper fix.

DedeHai (changes requested):
tested this and its now working properly in general. I have a fix for the drag selection which is now broken (and very annoying). should I push it directly to this PR for review?

Rps420 (approved):

Rps420 (approved):

wled00/data/settings_leds.htm
netmindz: Can you answer this question please @blazoncek

blazoncek: There need to be 2 calls and 2 `selectedIndex` change to properly select 1st available type in all circumstances.

I am readying this update along with ability to have GPIO dropdown selection for LED buses.

For my fork it is already working as expected but for upstream I have an issue with Hub75 matrices as they use pre-defined multiple GPIO which cannot be user-selected and are compile-time dependency. This is also an issue currently, however, since those GPIO are not grouped and reserved as one if user selects Hub75 as an output. There is also no place where those GPIO are saved in configuration as used GPIO if Hub75 is selected as an output. Those GPIO should remain available if Hu75 is not selected as an output.

blazoncek:
> drag selection

What kind of issue?

DedeHai:
> > drag selection
> 
> What kind of issue?

Try drag-marking the number of LEDs for editing.
Apart from that: accidental drag reordering could also be an issue for some users, I thoght about adding only a small icon or marked area that can be grabbed to avoid that. Or make d&d a checkmark option below the buses: makes users aware that this is supported and they can enable it only if they need it.

wled00/data/settings_leds.htm
netmindz: Yeah adding the drop-down for gpio with assignments is non trivial, which is why we are using the smaller change for the 0.15 maintenance release while we get your larger set of changes right for 0.16

blazoncek:
Adding `d.querySelectorAll(".iST input[type=number]").forEach((e,i)=>{e.draggable=true;e.setAttribute("ondragstart","pDS(event);");})` to the `updateTypeDropdown()` resolves the issue in Safari and Chrome, but not in Firefox.

```javascript
function pDS(e) {
	e.preventDefault();
	e.stopPropagation();
}
```

wled00/data/settings_leds.htm
blazoncek: TBH my work is perfectly suited for 0.15 as I'm developing on 0.15 codebase.
However I see no point in postponing a new 0.15 release just for that.

DedeHai: IMHO the issue is not about "suitable" but more about "working", we do not want to introduce issues into 0.15 that do not fix anything be it just for the sake of not aggrevating users.

blazoncek:
I am revoking this PR in favour of dynamically updating LED types in #5014.

PR #5004, opened 2025-10-15
Add Cycle mode to Twinkle

benjamw:
This adds a checkbox to the Twinkle effect that enables Cycle mode, where the lights will turn from the background color to the foreground color, and once all the pixels have changed, will randomly change each pixel back to the background color, completing the cycle which then repeats.

This obviously disables the Segment Intensity slider as all the pixels will twinkle to the foreground color before twinkling back to the background color.

All code has been successfully built, tested, and found working as expected on my setup.

Full disclosure, AI was used to assist me in making these changes.

## Summary by CodeRabbit

* **New Features**
  * Added an optional "Cycle" mode to the Twinkle effect that produces a flowing on/off sequence across the strip, advancing LEDs one-by-one in a tracked, repeatable pattern.
  * Preserved the original random twinkle behavior when Cycle mode is not enabled.
  * Effect label updated to indicate Cycle mode availability for easier discovery.


wled00/FX.cpp
```diff
@@ -675,7 +681,7 @@ uint16_t mode_twinkle(void) {
     uint32_t it = strip.now / cycleTime;
     
     // Initialize on first call
-    if (SEGENV.call == 0) 
+    if ((SEGENV.call == 0) || (SEGENV.aux1 > 1))
     {
```
benjamw: The suggested change was to add a value to the non-existent `aux2`. This method assumes that the `aux1` value was set by `hw_random` in the original code on line 750 to be some value (hopefully greater than 1).

Even if the data is not initialized, once the cycle completes, it gets initialized and works fine.

DedeHai:
thanks for contributing. I think adding 80 lines of code to 25line code effect is not a good idea. it should either be made much simpler or be its own effect.

benjamw:
There was conversation in my previous effect contribution PR about effect IDs being a limited commodity. So I was respecting that and putting this change into the effect that it was most similar to.

I don't feel like the code size should be the determining factor of where an effect should go, but rather, which effect it is most similar to and where it fits logically with respect to the actual effect.

But I'm also not here to tell the maintainers of the code how to handle their project, just giving my reasoning for why I did what I did.

I'll look into making it smaller. Thanks for the feedback.

DedeHai:
so you have basically recreated the dissolve effect? why not use that one?

benjamw:
Well, this is embarrassing. I just saw what twinkle was doing and had an idea, and never even looked for a different effect that did a similar thing.

Dissolve does behave nearly as I want, and will be good enough.

Thanks, and sorry for wasting your time.

PR #5005, opened 2025-10-15
Fix blank area issue with Twinkle

benjamw:
There was an issue with my setup where the first ~100 lights in my ~500 light string would not twinkle and would stay as the background color.

This fixes that issue on my setup.

AI seems to think that because the `hw_random` method creates a 32 bit number, when it's combined and truncated, it somehow leaves out a portion of the segment when the segment size is much smaller than 2^32.
Switching to a 16 bit random number fixed it. I'm not sure why, but I've built and tested it, and it works as expected with the same segments as the previously broken one.

## Summary by CodeRabbit

* **Refactor**
  * Adjusted the Twinkle effect’s internal random generation to a consistent 16-bit source, improving stability and visual variety of sparkle patterns.
  * Slight changes to random sequencing may subtly alter existing Twinkle animations’ appearance; no action required from users.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -672,7 +672,7 @@ uint16_t mode_twinkle(void) {
     if (SEGENV.aux0 >= maxOn)
     {
       SEGENV.aux0 = 0;
-      SEGENV.aux1 = hw_random16(); //new seed for our PRNG (16-bit for better coverage)
+      SEGENV.aux1 = hw_random(); //new seed for our PRNG
     }
     SEGENV.aux0++;
     SEGENV.step = it;


DedeHai:
good call, I wonder how no one noticed this before, this bug was introduced a year ago. 
your AI is wrong however. the issue is that PRNG relies on 16bit overflows, so it must be `uint16_t`
please update and once approved, please cherry pick this PR for 0._15_x a this needs backpporting to 0.15.2 (I can do that if you do not know how)

wled00/FX.cpp
```diff
@@ -672,13 +672,13 @@ uint16_t mode_twinkle(void) {
     if (SEGENV.aux0 >= maxOn)
     {
       SEGENV.aux0 = 0;
-      SEGENV.aux1 = hw_random(); //new seed for our PRNG
+      SEGENV.aux1 = hw_random16(); //new seed for our PRNG (16-bit for better coverage)
```
DedeHai: this is not the issue. please revert.

wled00/FX.cpp
```diff
@@ -672,13 +672,13 @@ uint16_t mode_twinkle(void) {
...
     }
     SEGENV.aux0++;
     SEGENV.step = it;
   }
 
-  unsigned PRNG16 = SEGENV.aux1;
+  uint16_t PRNG16 = SEGENV.aux1;
```
DedeHai, line 681: this is what fixes it.

DedeHai (changes requested):

benjamw:
Should I just edit this PR to merge into `0_15_x`?
Or create a PR for both?

NOPE. That didn't work. I'll create a new PR for that branch.

wled00/FX.cpp
benjamw: Interestingly, in the `0_15_x` branch, the method used here is `random16`, which I'm assuming is the 16-bit version of `random`.

I'm not disagreeing with you. PRs should be as small as possible and I have reverted that change.

benjamw: I noticed because I got conflicts when I was cherry-picking.

benjamw: I was wrong. There is no `random` or `random32` method in the FastLED lib. So `random16` was the largest number available.

DedeHai: yes, I replaced all fastled pseudo random numbers (or almost all) with an implementation that uses the hardware random register that is available in all ESPs giveing almost true random numbers

benjamw: Very nice! Thanks for the info.

PR #5012, opened 2025-10-19
Improvement to "Welcome Page" check

DedeHai:
There are frequent questions on why the welcome page shows in AP or ETH mode. The current workaround is to change the wifi network name, which is not intuitive.
This PR adds a secondary condition: if the config backup exists, the welcome page is also skiped, meaning the config was saved at least once.
This can still be changed to an actual config parameter in the future if better control is needed.

## Summary by CodeRabbit

* **New Features**
  * Added configuration backup detection capability, allowing the system to check for existing backup files.
  * Modified welcome page behavior: no longer displayed when a configuration backup is present.

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -474,7 +474,7 @@ void WLED::setup()
 
   if (needsCfgSave) serializeConfigToFS(); // usermods required new parameters; need to wait for strip to be initialised #4752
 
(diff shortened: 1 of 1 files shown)


netmindz (approved):

netmindz:
I want to overhaul the whole onboarding process, but this is definitely an improvement until then

PR #5013, opened 2025-10-19
Use sequential loading and requests for all UI resources

DedeHai:
- Load common.js and style.css sequentially for all config pages (no more "xy is undefined" due to missing common.js)
- Restrict all requests in index.js to single connection

This **greatly** improves UI stability in low heap as well as bad wifi signal conditions and does not noticeably slow down any UI access.
I also increased the retries on failed requests from 1 to 5 but shortened the request interval so some requests may even be faster than before. 

Combined with #4939 this makes the ESP8266 work flawlessly, I was even able to increase the number of LEDs from current 720 to 900 and its still more stable than it is currently. With this PR the UI keeps running down to about 6k of free heap whereas it is already struggeling at 10k without these changes.

the changes add ~900bytes to flash usage.

## Summary by CodeRabbit

* **New Features**
  * Dynamic resource loader for sequential CSS/JS loading with retry and a new loadResources API
  * WebSocket connector now reuses an open socket and provides the socket to connection callbacks

* **Bug Fixes**
  * Page is hidden until resources finish loading to avoid flashes of unstyled content
  * Bounded retry logic and centralized error handling to improve startup resilience

* **Chores**
  * Startup refactored to async/Promise flows for more reliable initialization
  * Removed a redundant pre-flight memory check in WebSocket send logic

### DIFF
--- a/wled00/data/liveview.htm
+++ b/wled00/data/liveview.htm
@@ -38,7 +38,7 @@
         ctx.fillRect(Math.round((i - start) * w / skip), 0, Math.ceil(w), c.height);
       }
     }
-    function update() { // via HTTP (/json/live)
+    function update(retry=0) { // via HTTP (/json/live)
       if (d.hidden) {
         clearTimeout(tmout);
         tmout = setTimeout(update, 250);
@@ -60,7 +60,7 @@
       .catch((error)=>{
         //console.error("Peek HTTP error:",error);
         clearTimeout(tmout);
-        tmout = setTimeout(update, 2500);
+        if (retry<5) tmout = setTimeout(() => update(retry+1), 2500); // stop endlessly bugging the ESP if resource is not available
       })
     }
     function S() { // Startup function (onload)


wled00/data/index.js
```diff
@@ -2820,56 +2804,53 @@ function rSegs()
...
+			localStorage.setItem(lsKey, JSON.stringify({
+				p: palettesData,
+				vid: lastinfo.vid
+			}));
+			redrawPalPrev();
+			setTimeout(resolve, 99); // delay optional
+		});
 	});
-}
+	}
```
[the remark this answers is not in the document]
blazoncek: That appears to be a longstanding bug. However `lastinfo.vid` is undefined at this time.

wled00/data/common.js
```diff
@@ -51,6 +51,35 @@ function tooltip(cont=null) {
...
+		if (i >= files.length) {
+			if (init) {
+				d.documentElement.style.visibility = 'visible'; // make page visible after all files are laoded if it was hidden (prevent ugly display)
+				d.readyState === 'complete' ? init() : window.addEventListener('load', init);
+			}
+			return;
+		}
+		const file = files[i++];
+		const isCSS = file.endsWith('.css');
+		const el = d.createElement(isCSS ? 'link' : 'script');
```
blazoncek, line 67: Why did you choose to implement JS loading as that already have a function for that, `loadJS()`.

wled00/data/common.js
```diff
@@ -51,6 +51,35 @@ function tooltip(cont=null) {
...
+		if (isCSS) {
+			el.rel = 'stylesheet';
+			el.href = file;
+		} else {
+			el.src = file;
+		}
+		el.onload = () => {	setTimeout(loadNext, 0);};
```
blazoncek: Is there a particular reason to "schedule" next load without timeout? Wouldn't a simpler `loadNext()`be enough?

wled00/data/settings.htm
```diff
@@ -4,13 +4,33 @@
 	<meta charset="UTF-8">
 	<meta content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" name="viewport">
 	<title>WLED Settings</title>
-	<script src="common.js" type="text/javascript"></script>
 	<script>
+		// load common.js with retry on error
+		(function loadFiles() {
+			const l = document.createElement('script');
+			l.src = 'common.js';
+			l.onload = () => loadResources(['style.css'], S); // load style.css then call S()
```
blazoncek: You can bypass `S()` by using lambda function here.

wled00/data/settings.htm
```diff
@@ -4,13 +4,33 @@
...
+			document.head.appendChild(l);
+		})();
+
 		function S() {
 			getLoc();
 			loadJS(getURL('/settings/s.js?p=0'), false);	// If we set async false, file is loaded and executed, then next statement is processed
 		}
 	</script>
+</head>
+<body>
```
blazoncek: Why move these?

wled00/data/index.js
```diff
@@ -481,8 +482,7 @@ function restore(txt) {
 	return false;
 }
 
-function loadPresets(callback = null)
-{
+async function loadPresets(callback = null) {
```
blazoncek: You can remove callback as it is no longer necessary.

wled00/data/index.js
```diff
@@ -1709,77 +1690,80 @@ function setEffectParameters(idx)
...
-	var useWs = (ws && ws.readyState === WebSocket.OPEN);
-	var type = command ? 'post':'get';
-	if (command) {
-		command.v = true; // force complete /json/si API response
-		command.time = Math.floor(Date.now() / 1000);
-		var t = gId('tt');
-		if (t.validity.valid && command.transition==null) {
-			var tn = parseInt(t.value*10);
-			if (tn != tr) command.transition = tn;
+async function requestJson(command=null) {
```
blazoncek: `requestJson()` is used on button press as well as other places in the code. I hope this does not break such use.

wled00/data/common.js
DedeHai: my function loads JS and CSS, I did not understand the purpose of the already implemented one. maybe they can be combined?

wled00/data/common.js
DedeHai: This was suggested to me by claude.ai and the purpose is, if I understand correctly, to execute a loaded script if it needs to be executed. I am on thin ice on this one ;)

wled00/data/settings.htm
DedeHai: this is beyond my understanding of JS, can you elaborate or make an example?

wled00/data/settings.htm
DedeHai: I assume you are referring to the moved CSS? if it is on top, it will not override style.css and look all ugly.

wled00/data/index.js
DedeHai: Do you know how to test this?

DedeHai:
>  I did find a few oddities, too.

I did mostly generate this using a series of AI requests but must admit I do not understand all of it in detail. So very well possible some of the code can be streamlined or improved upon. I did test it extensively on all platforms though.

wled00/data/settings.htm
blazoncek: CSS = *Cascading* Style Sheet
The inheritance *cascades* and if you *insert* DOM object at a proper place, all is good.
However, since you just used `document.head.appendChild()` you are appending inserted object at the end of head, making style.css the last style definition (and so taking precedence from anything above it).
You can overcome this by searching for the first `<style>` element in head and insert style.css before it.

wled00/data/index.js
blazoncek: Except by clicking everything, IDK.

wled00/data/settings.htm
blazoncek: In the same way as you did everything else.
`()=>{...; getV(); ... }`
No fuss lambda function that replaces S(). May make code a tiny bit smaller and more legible.

netmindz:
Interesting idea to tweak the behaviour of the browser to make it a little more friendly to the microcontroller. 

If we had request queuing in the server this might be less of an issue, but until then this might be a good idea

wled00/data/common.js
blazoncek: You can just call loadNext() which will do the same.
Almost. If you call loadNext() it will be executed *within* onload callback, if you schedule it, it will run separately in its own context.

blazoncek:
> tweak the behaviour of the browser

I did that with the release of 0.14. It did not go well. Perhaps this implementation is better but needs to be thoroughly tested on multiple devices, especially in captive portal mode.

DedeHai:
> If we had request queuing in the server this might be less of an issue, but until then this might be a good idea

there is request queuing, but each queue eats up RAM and if heap is low, they do get dropped and throw a 503 so I dont think this can ever be resolved on the server side.

happysmartlight:
Hi @DedeHai  

I’ve tested it, and the issue #5077 has indeed been resolved!

Hopefully, this branch will be merged soon to improve performance.

wled00/data/common.js
DedeHai: I took a look at it and the two functions do server quite different purposes, unifying them does not really make sense and will just make the code more complex, at least that is how it looks to me.

wled00/data/index.js
```diff
@@ -616,10 +581,10 @@ function populateQL()
 	gId('pql').innerHTML = cn;
 }
 
-function populatePresets(fromls)
+async function populatePresets(fromls)
 {
 	if (fromls) pJson = JSON.parse(localStorage.getItem("wledP"));
-	if (!pJson) {setTimeout(loadPresets,250); return;}
+	if (!pJson) {await loadPresets(); return;}
```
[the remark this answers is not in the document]
DedeHai: does presetError() not result in a valid pJson? that is its purpose

happysmartlight:
Hi @DedeHai 

Did you make additional updates on your side?
What does it improve?
What was the reason or motivation behind these new changes?

DedeHai:
see individual commits.

netmindz:
What testing is required @DedeHai ? Could you add some notes, it seems like at least one person has already treated and confirmed the improvement, so what it still to be tested?

blazoncek:
@netmindz from my own experience these kind of changes need thorough testing on various mobile devices (especially Android) in captive portal mode.
Unfortunately one can only have so many.

DedeHai:
@netmindz I ran some basic tests, also in AP mode and fixed some issues afterwards but I still want to re-test it, also on ESP8266 and also on android, firefox, safari. after those are successful, we can dare to merge this for further testing.

edit:
as for tests required, I really don't know, the "dangerous" changes are exclusively in index.js and the one bug that stands (palette preview not loaded in AP mode) is already present before this PR so I did not further investigate it yet. Things to look out for would be errors on high-wifi traffic / low wifi signal which I do not know how to emulate. Other than that I just poke around the UI, adding segments, presets, palettes etc. and see if it all works.

wled00/data/index.js
```diff
@@ -2819,56 +2770,50 @@ function rSegs()
...
+		getPalettesData(0, () => {
+			localStorage.setItem("wledPalx", JSON.stringify({
+				p: palettesData,
+				vid: lastinfo.vid
+			}));
+			redrawPalPrev();
+			setTimeout(resolve, 99); // delay optional
+		});
 	});
 }
```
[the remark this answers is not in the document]
DedeHai: is there any downside to moving palette data loading to after requestJson() in the init sequence?

DedeHai:
@netmindz I did test this and found a few oddities I was later not able to reproduce. the last commit seems to solve #5080 but needs confirmation.
the biggest issue I found was on android the peek was not working at all and all the config pages were just black. Debugging showed that in the config pages, the "hidden" style element was not removed, leaving the pages blank. I then tried a different android device and it just worked. A full page refresh then fixed it. So I suspect the issue was a cached "common.js" which was lacking the init function that removes the "hidden" tag. I then added a timout to automatically remove the tag but then removed again as it does not make any sense to do that: if `common.js` is not loaded properly, the init of the page will fail anyway because `S()` is not called.
Other than that I did not find any bugs related to this PR so I think its good to merge.

edit:
tested with ESP8266 on Chrome, Safari, Firefox and Android. Also tested in AP mode in Chrome and Android. Not tested on iPhone and Opera.

DedeHai:
just found one small bug I am not able to fix: when saving any of the settings pages, the "xxx settings saved..." sometimes shows up with no CSS styling i.e. in white. But not always. Any ideas on why that is and how to fix it?

edit:
there is another issue: the ws-connection appears to be less stable. On a C3 with high FPS (>100) the connection breaks when using "peek" and clicking a button. Need to investigate.

DedeHai:
I was no longer able to reproduce the connection issue I saw on the C3, not sure what the issue was there, it was rock-solid in my recent tests even at 120FPS and in AP mode.
The only real issue I am seeing is with cached pages: if I switch back and forth between versions and dont do a full page refresh, the UI can hang or act all weird.

wled00/data/common.js
```diff
@@ -51,6 +51,38 @@ function tooltip(cont=null) {
...
+			d.head.appendChild(el);
+		}
+		el.onload = () => {	loadNext(); };
+		el.onerror = () => {
+			i--; // load this file again
+			setTimeout(loadNext, 100);
+		};
+	};
+	loadNext();
+}
```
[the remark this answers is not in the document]
DedeHai: this is intentional, a page can not work without these required resources.

DedeHai:
In the latest commit I removed the hiding of the "(un)freeze" button in segments. Does anyone know the reason why it was hidden if a segment is not collapsed?

netmindz:
Why have you removed the Awaiting Testing label without comment from someone else's testing @DedeHai ?

DedeHai:
I added that as a reminder to myself as I did not deem it tested enough. I ran many tests on this on different platforms, in AP mode, on mobile, different browsers. I can add the label back in if anyone still wants to and is willing to run more tests.

DedeHai:
The only thing that I found that could cause trouble is browser cache: if the browser uses cached files from a previous version, the UI can break. I did not find a good way to enforce an automatic full reload so if there is an easy way to do that it should be added.

willmmiles:
>  I did not find a good way to enforce an automatic full reload so if there is an easy way to do that it should be added.

Today: change `VERSION` for each build -- it's used as the cache key.   Later we can patch up `cdata.js` to also include static content hashes.

willmmiles:
> > I did not find a good way to enforce an automatic full reload so if there is an easy way to do that it should be added.
> 
> Today: change `VERSION` for each build -- it's used as the cache key. Later we can patch up `cdata.js` to also include static content hashes.

I guess later was today (#5120)!  I'm still living in the past, I guess..

PR #5015, opened 2025-10-22
Improvements to the PWM Fan usermod

qistoph:
This is a combined PR for some improvements on the PWM Fan usermod.

**More precise RPM calculation**
Use milliseconds between actual measurements to calculate the desired speed instead of the configured seconds between checks.

**Show locked state when tacho is used**
The user interface shows **Speed**: `locked`/`auto` when the tacho pin is not used. This is useful information that is lost when the tacho pin is enabled, by showing **Speed**: `1500rpm`. This PR changes it to:
- **Speed**: `1500 rpm (locked)`
- **Speed**: `1500 rpm (auto)`

**Linear PWM control**
Before this PW the PWM speed is configured and calculated using
- Desired temperature (default: 35°C)
- Number of PWM steps (default: 7)
- Temperature difference per PWM step (default: 0.5°C)

desmos-graph

This PR changes the calculation to allow linear PWM control, instead of steps, and includes defaults to match the previous outcomes.
- Desired temperature (default: 34.5°C)
- Temperature at which fan should be at maximum speed (default: 38"C)

desmos-graph(1)

**Disable lock from UI with speed 0**
There was no way to disable the fan lock from the user interface. It would stay on until a reset or JSON API request. A user without access to the JSON API would not be able to turn to automatic fan control from the interface.

This PR changes the client side javascript call to unlock the automatic control when the user selects speed 0.

**Side effect**: the user can no longer disable the fan using the manual override, but would have to change the fan settings.

If any of these changes are undesired or require further improvement, please let me know. I'd be happy to help and look forward to have these changes included in the beautiful WLED.

## Summary by CodeRabbit

* **New Features**
  * Configurable maximum temperature ceiling to control fan max speed.
  * UI/state shows max-temperature control and explicit RPM labels for auto vs locked.

* **Bug Fixes**
  * More accurate RPM timing using measured intervals.
  * Temperature-to-PWM mapping changed from stepped to smooth linear interpolation with bounds/validation.
  * Manual PWM = 0 now clears any fan lock state.

* **Chores**
  * Persisted max-temperature setting across restarts; state updates reflect changes immediately.


usermods/PWM_fan/PWM_fan.cpp
```diff
@@ -167,25 +165,23 @@ class PWMFanUsermod : public Usermod {
...
 
-    uint8_t calculatePwmStep(float diffTemp){
-      if ((diffTemp == NAN) || (diffTemp <= -100.0)) {
-        DEBUG_PRINTLN(F("WARNING: no temperature value available. Cannot do temperature control. Will set PWM fan to 255."));
-        return _pwmMaxStepCount;
-      }
-      if(diffTemp <=0){
-        return 0;
+      int pwmMinValue = minPWMValuePct / 100 * _pwmMaxValue;
+      int pwmMaxValue = maxPWMValuePct / 100 * _pwmMaxValue;
```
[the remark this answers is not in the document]
blazoncek: Valid.

usermods/PWM_fan/PWM_fan.cpp
```diff
@@ -167,25 +165,23 @@ class PWMFanUsermod : public Usermod {
...
+
+      if (pwmMaxValue <= pwmMinValue) return;
```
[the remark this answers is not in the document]
blazoncek: Valid.

usermods/PWM_fan/PWM_fan.cpp
```diff
@@ -218,8 +214,11 @@ class PWMFanUsermod : public Usermod {
       unsigned long now = millis();
       if ((now - msLastTachoMeasurement) < (tachoUpdateSec * 1000)) return;
 
-      updateTacho();
+      updateTacho(now - msLastTachoMeasurement);
```
blazoncek, line 229: That's bound to be roughly 30s due to the statement above.

usermods/PWM_fan/PWM_fan.cpp
```diff
@@ -249,15 +248,16 @@ class PWMFanUsermod : public Usermod {
         uiDomString += FPSTR(_name);
         uiDomString += F("':{'");
         uiDomString += FPSTR(_speed);
-        uiDomString += F("':parseInt(this.value)}});\" oninput=\"updateTrail(this);\" max=100 min=0 type=\"range\" value=");
+        uiDomString += F("':parseInt(this.value),'lock':parseInt(this.value)>0}});\" oninput=\"updateTrail(this);\" max=100 min=0 type=\"range\" value=");
```
blazoncek: Why do you need "lock" when "speed" is enough (as both have the same value)?

blazoncek (changes requested):
As an author and maintainer of the usermod I see no *real* added value in the proposed changes.
There is no benefit if fan is running at 33.333% of speed compared to 7-step 28.5% or 42.8%.
There are several other issues I could spot.

usermods/PWM_fan/PWM_fan.cpp
```diff
@@ -54,23 +54,23 @@ class PWMFanUsermod : public Usermod {
     int8_t  tachoPin          = TACHO_PIN;
     int8_t  pwmPin            = PWM_PIN;
     uint8_t tachoUpdateSec    = 30;
-    float   targetTemperature = 35.0;
+    float   targetTemperature = 34.5;
+    float   maxSpeedTemperature = 38;
```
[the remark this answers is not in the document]
qistoph: This is checked before the critical division in `setFanPWMbasedOnTemperature`

usermods/PWM_fan/PWM_fan.cpp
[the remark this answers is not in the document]
qistoph: I must have overlooked this one. Thanks! I'll fix it soon

usermods/PWM_fan/PWM_fan.cpp
qistoph: That assumes that the call is effectively called every 30s. It could throw off the result if this is not the case due to high load.

It's a little improvement to make the code more robust.

usermods/PWM_fan/PWM_fan.cpp
qistoph: "speed" is set to the int value.
"lock" is set to 'true' for values >0, otherwise 'false'
Effectively unlocking (auto mode) when setting to 0.
Before this change it was not possible to unlock after locken from the Config menu in the UI.

qistoph:
> As an author and maintainer of the usermod I see no _real_ added value in the proposed changes. There is no benefit if fan is running at 33.333% of speed compared to 7-step 28.5% or 42.8%. There are several other issues I could spot.

Changing the speed more granularly makes it less noticeable when it happens and becoming less of a disturbance.

What's your suggestion / request on the PR? Will it be accepted after fixing the requested changes? Do you need more explanation or not see it being accepted anyway?

usermods/PWM_fan/PWM_fan.cpp
blazoncek: you can always determine "lock" from "speed". No need to pass it.

blazoncek:
> becoming less of a disturbance.

Apparently you must be using this usermod with different intention than the one used when developing it. The idea is to provide ventilation to a controller enclosure, not living room ventilation.

Adding code to already crowded WLED which yields little real benefit is not really desired. However this is a (not default) usermod and some wiggle room may be exerted. I am willing to accept linear (or non linear for that matter) speed adjustment, but only that. No need to change polling interval beyond the ability for user to select one.

Perhaps @srg74 would also like to be informed.

usermods/PWM_fan/PWM_fan.cpp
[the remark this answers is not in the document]
qistoph: Fixed, pushing soon

usermods/PWM_fan/PWM_fan.cpp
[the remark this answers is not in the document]
qistoph: Fixed, pushing soon

usermods/PWM_fan/PWM_fan.cpp
qistoph: Not being able to return to auto mode from the UI seems like a missing feature.

I see three solutions each with its own associated impact:
1. Introduce a new button/option in the UI to switch back to auto mode - UI looks change
2. Change UI logic - Front-end usage changes slightly
3. Change back-end logic - JSON API usage changes.

Option 2 has the least impact.

Fitting option 3 could be putting the same logic in the back-end: `if (speed<=0) { lockFan=false }`. This would impact users who would want to use `{"speed": 0, "lock": true}` to lock the in fan _off_ state.

Looking forward to your preference and thoughts

qistoph:
The fan I'm using is to used to ventilate the enclosure. The enclosure and LEDs are however in my living area.

I understand and support preventing WLED to become some overcrowded multi-tool. The changes I'm providing reduce the complexity of calculations, reduce the number of parameters used, makes the RPM calculation more reliable and robust and adds a three features.

If some of these are undesirable, could you please be more specific in the desirable changes I'm suggesting?

1. More precise and reliable RPM calculation
2. Show locked state when tacho is used (feature 1)
3. Linear PWM control
  a. Set temperature for fan max speed, instead of 7*0.5=3.5°C above target temperature. (feature 2)
5. Disable lock from UI with speed 0 (feature 3)

usermods/PWM_fan/PWM_fan.cpp
blazoncek: This was not an oversight but the intentional design: once you override fan speed, you did that on purpose (perhaps you want fan to be off, i.e. speed=0).
Adding a separate button to unlock "locked" speed is foreseen in the code but intentionally left out of UI.
Using external control, you can "unlock" fan speed with API provided.

I'm willing to take compromise in the form of separate UI button for Lock/Unlock functionality (2). But no other changes to UI or backend functionality.

usermods/PWM_fan/PWM_fan.cpp
```diff
@@ -167,25 +165,26 @@ class PWMFanUsermod : public Usermod {
...
-    uint8_t calculatePwmStep(float diffTemp){
-      if ((diffTemp == NAN) || (diffTemp <= -100.0)) {
-        DEBUG_PRINTLN(F("WARNING: no temperature value available. Cannot do temperature control. Will set PWM fan to 255."));
-        return _pwmMaxStepCount;
+      int pwmMinValue = minPWMValuePct * _pwmMaxValue / 100;
+      int pwmMaxValue = maxPWMValuePct * _pwmMaxValue / 100;
+
+      if (pwmMaxValue <= pwmMinValue) {
+        updateFanSpeed(pwmMaxValue); // fail safe: run at max speed
+        return;
```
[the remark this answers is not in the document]
qistoph: Agreed

usermods/PWM_fan/PWM_fan.cpp
```diff
@@ -96,17 +96,15 @@ class PWMFanUsermod : public Usermod {
...
+    void updateTacho(unsigned long msInterval) {
       if (tachoPin < 0) return;
 
       // start of tacho measurement
       // detach interrupt while calculating rpm
       detachInterrupt(digitalPinToInterrupt(tachoPin)); 
       // calculate rpm
       last_rpm = (counter_rpm * 60) / numberOfInterrupsInOneSingleRotation;
-      last_rpm /= tachoUpdateSec;
+      last_rpm /= msInterval / 1000;
```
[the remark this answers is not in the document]
qistoph: Agreed

usermods/PWM_fan/PWM_fan.cpp
qistoph: Removed the change for now. I'll try to fix this in a separate PR later.

DedeHai:
these look like solid changes to me. I do not use the UM nor have hardware to test it but if this does not break any existing setups this has my green light

usermods/PWM_fan/PWM_fan.cpp
blazoncek: Usermod's `loop()` is called about 1500-2500 times per second.
`now - msLastTachoMeasurement` will be less than `tachoUpdateSec * 1000` until `tachoUpdateSec` seconds pass which is 30s by default. So no point is storing that value as it will be off at most few ms.

blazoncek (commented):
Original request and URS for PWM fan usermod came from @srg74 and if he approves the changes I'm ok with the update, if not, neither am I.

srg74:
I’m ok with improvement.

blazoncek (approved):

DedeHai (approved):

softhack007:
@DedeHai should we merge this?

qistoph:
Anything I can do to help merge this PR?

softhack007:
@coderabbit re-review please. Create a prioritized list of suggested improvements - major (functional impact, credible robustness issues) only.

softhack007:
> Anything I can do to help merge this PR?

@qistoph sorry for the delay - we were over-busy with release prep for 16.0.0 in the last months, it looks like your PR got pushed down our stack 😉 

I've kicked a final review by the bot, please read through open findings (no need to fix everything), and then we should be ready to merge.

softhack007:
@qistoph please check if any of the [bot findings above](https://github.com/wled/WLED/pull/5015#issuecomment-4471798010) should get a fix.

qistoph:
Thanks for the reminder @softhack007. At first glance they look like solid suggestions. I haven't found the time to dive into them yet. It's on my list though

qistoph:
All done.

It was a bit annoying though that the previous code review suggested to split the calculation and this one noticed that overflows the uint16 in some cases.

PR #5016, opened 2025-10-22
Adding "Complete" mode to Dissolve FX: always fades completely

DedeHai:
This allows for much slower speed setting to not turn into "twinkle" effect


DedeHai:
@benjamw this does what you wanted to achieve in https://github.com/wled/WLED/pull/5004

DedeHai:
@netmindz any objections to merging this? it leaves the current effect as-is, adds a checkmark to trigger the "complete" mode.

Issue #4800, opened 2025-07-31
0.16 crashes when changing from 1D stip to 2D Matrix and clicking on save. 0.15 seams fine


Andro-Marian:
### What happened?

Crashes when changing from 1D to 2D and click on save

### To Reproduce Bug

1. Fresh install the latest beta 0.16.
2. Open settings -> 2D Configuration -> Strip or panel: change to 2D Matrix -> Click: Save

```
The connection was reset

The connection to the server was reset while the page was loading.

    The site could be temporarily unavailable or too busy. Try again in a few moments.
    If you are unable to load any pages, check your computer’s network connection.
    If your computer or network is protected by a firewall or proxy, make sure that Firefox is permitted to access the web.```

### Expected Behavior

To save the settings

### Install Method

Binary from WLED.me

### What version of WLED?

0.16

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell
Backtrace: 0x4008ab44:0x3ffe11f0 0x4008aea1:0x3ffe1210 0x401c0557:0x3ffe1230 0x401680e3:0x3ffe1260 0x4013c1b4:0x3ffe1280 0x40118ba5:0x3ffe12a0 0x4013c7c6:0x3ffe1320 0x4013c827:0x3ffe13b0 0x4008c92a:0x3ffe13e0
  #0  0x4008ab44:0x3ffe11f0 in invoke_abort at /home/cschwinne/esp32-arduino-lib-builder/esp-idf/components/esp32/panic.c:648
  #1  0x4008aea1:0x3ffe1210 in vApplicationStackOverflowHook at /home/cschwinne/esp32-arduino-lib-builder/esp-idf/components/esp32/panic.c:648    
  #2  0x401c0557:0x3ffe1230 in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::__shared_count<WiFiClientRxBuffer*>(WiFiClientRxBuffer*) at C:\Users\DesktopTV\.platformio\packages\framework-arduinoespressif32\libraries\WiFi\src/WiFiClient.cpp:572
      (inlined by) std::__shared_ptr<WiFiClientRxBuffer, (__gnu_cxx::_Lock_policy)2>::__shared_ptr<WiFiClientRxBuffer>(WiFiClientRxBuffer*) at c:\users\desktoptv\.platformio\packages\toolchain-xtensa32\xtensa-esp32-elf\include\c++\5.2.0\bits/shared_ptr_base.h:882
[...]

PR #5023, opened 2025-10-24
fix timing issue when changing 1D <-> 2D credits to blazoncek

### LINKED ISSUE #4911: Saving 2D config with higher LED count than configured leads to crash
### What happened?

I think this was discussed on Discord. When enabling 2D before setting the right amount of LEDs in config WLED just crashes.
I did some investigation on this but was not able to find where exactly this happes

### To Reproduce Bug

Enable 2D and set the panel size larger than the number of configured LEDs.

### Expected Behavior

Save the 2D setting without crashing.

### Install Method

Self-Compiled

### What version of WLED?

0.16

ESP32-C3

### Relevant log/trace output

```shell

```

### Anything else?

_No response_

### Code of Conduct

DedeHai:
this fixes #4911 and #4800

## Release Notes

* **Bug Fixes**
  * Enhanced synchronization during LED matrix geometry reconfiguration to prevent race conditions
  * Improved stability of LED effect drawing during matrix setup and configuration changes

### DIFF
--- a/wled00/set.cpp
+++ b/wled00/set.cpp
@@ -815,8 +815,13 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
       }
     }
     strip.panel.shrink_to_fit();  // release unused memory
+    // we are changing matrix/ledmap geometry which *will* affect existing segments
+    // since we are not in loop() context we must make sure that effects are not running. credit @blazonchek for properly fixing #4911
+    strip.suspend();
+    strip.waitForIt();
     strip.deserializeMap(); // (re)load default ledmap (will also setUpMatrix() if ledmap does not exist)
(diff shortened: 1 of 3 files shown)


wled00/set.cpp
```diff
@@ -815,8 +815,13 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
       }
     }
     strip.panel.shrink_to_fit();  // release unused memory
+    // we are changing matrix/ledmap geometry which *will* affect existing segments
+    // since we are not in loop() context we must make sure that effects are not running. credit @blazonchek for properly fixing #4911
+    strip.suspend();
```
softhack007, line 820: makes sense - we basically have the same logic in WLED-MM for more than a year, and it seems to work well.

PR #5026, opened 2025-10-26
Improvements & fixes for HUB75

DedeHai:
- added proper config parameters to allow multiple panels
- added config checks
- fixed crashes on S3
- changed constant variables to constexpr
- added "wled.h" to bus_manager.cpp and removed local function prototypes (needed for buffer allocations)
- speed optimisations: yields about 10% higher FPS
- proper brightness handling
- updated platformio_override.sample.ini
- some code cleanup

I do not own any quarter-scan HUB75 panels nor a suitable ESP32, so please test if this is working, I tested this on 2x half-scan 64x32 panels using an S3 only.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable HUB75 matrix display, per-pixel accuracy, safer memory handling and cleanup for large panels; consistent brightness to reduce ghosting.

* **New Features**
  * Improved HUB75 wiring/pin handling and UI validation to prevent invalid panel size, wiring conflicts and multiple HUB75 selections; sensible defaults for panel inputs; prompts reboot when required on S3.

* **Documentation**
  * Added brightness guidance to avoid ghosting on some HUB75 panels.

### DIFF
--- a/wled00/bus_manager.cpp
+++ b/wled00/bus_manager.cpp
@@ -1036,7 +1036,7 @@ uint32_t BusHub75Matrix::getPixelColor(unsigned pix) const {
 
 void BusHub75Matrix::setBrightness(uint8_t b) {
   _bri = b;
-  display->setBrightness8(_bri); 
+  display->setBrightness(_bri); 
 }
 
 void BusHub75Matrix::show(void) {
@@ -1063,8 +1063,8 @@ void BusHub75Matrix::show(void) {
 
 void BusHub75Matrix::cleanup() {
   if (display && _valid) display->stopDMAoutput();  // terminate DMA driver (display goes black)
-  delay(30); // give some time to finish DMA
   _valid = false;
+  delay(30); // give some time to finish DMA
   _panelWidth = 0;
   deallocatePins();
   DEBUGBUS_PRINTLN(F("HUB75 output ended."));


wled00/bus_manager.cpp
```diff
@@ -808,38 +774,25 @@ BusHub75Matrix::BusHub75Matrix(const BusConfig &bc) : Bus(bc.type, bc.start, bc.
...
-          virtualDisp->setPhysicalPanelScanRate(FOUR_SCAN_64PX_HIGH);
-          break;
-        default:
-          DEBUGBUS_PRINTLN("Unsupported height");
-          return;
-      }
+      _isVirtual = true;
+      mxconfig.mx_width = min((uint8_t) 64, panelWidth) * 2;
+      mxconfig.mx_height = min((uint8_t) 64, panelHeight) / 2;
   } else {
```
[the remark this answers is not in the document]
DedeHai: correct, this is what the TODO sais.

wled00/bus_manager.cpp
```diff
@@ -1041,59 +1027,54 @@ void __attribute__((hot)) BusHub75Matrix::setPixelColor(unsigned pix, uint32_t c
...
 }
 
 void BusHub75Matrix::setBrightness(uint8_t b) {
   _bri = b;
-  if (display) display->setBrightness(_bri);
 }
 
 void BusHub75Matrix::show(void) {
   if (!_valid) return;
-  display->setBrightness(_bri);
```
softhack007, line 1058: @DedeHai not sure why did you remove the driver built-in panel brightness feature? Is it slower than doing brightness "by hand" in show()?

wled00/bus_manager.cpp
```diff
@@ -1041,59 +1027,54 @@ void __attribute__((hot)) BusHub75Matrix::setPixelColor(unsigned pix, uint32_t c
...
     for (int y=0; y<height; y++) for (int x=0; x<width; x++) {
       if (getBitFromArray(_ledsDirty, pix) == true) {        // only repaint the "dirty"  pixels
-        uint32_t c = uint32_t(_ledBuffer[pix]) & 0x00FFFFFF; // get RGB color, removing FastLED "alpha" component
-        uint8_t r = R(c);
-        uint8_t g = G(c);
-        uint8_t b = B(c);
-        if (isVirtualDisp) virtualDisp->drawPixelRGB888(int16_t(x), int16_t(y), r, g, b);
-        else display->drawPixelRGB888(int16_t(x), int16_t(y), r, g, b);
+        CRGB c = _ledBuffer[pix];
+        c.nscale8_video(_bri); // apply brightness
```
softhack007: see my previous question - is this faster?

wled00/bus_manager.cpp
```diff
@@ -1041,59 +1027,54 @@ void __attribute__((hot)) BusHub75Matrix::setPixelColor(unsigned pix, uint32_t c
...
   _valid = false;
   _panelWidth = 0;
   deallocatePins();
-  DEBUGBUS_PRINTLN("HUB75 output ended.");
-
-  //if (virtualDisp != nullptr) delete virtualDisp;  // warning: deleting object of polymorphic class type 'VirtualMatrixPanel' which has non-virtual destructor might cause undefined behavior
-  delete display;
+  DEBUGBUS_PRINTLN(F("HUB75 output ended."));
+  #ifndef CONFIG_IDF_TARGET_ESP32S3 // on ESP32-S3 deleting display/virtualDisp does not work and leads to crash (DMA issues), request reboot from user instead
+  if (virtualDisp != nullptr) delete virtualDisp; // note: in MM there is a warning to not do this but if using "NO_GFX" this is safe
```
softhack007, line 1072: If I remember correct, I've reported this as a bug to MrCodetastic. It might be that the problem is already solved in the latest hub75 driver.

softhack007:
@DedeHai and @netmindz one thing you might want to try - currently chained hub75 panels are not compatible with the wled 2D layout settings. The reason is that HUB75 is scan line oriented, and adding a second panel means that the physical width (scanline) is increased, for example from width=64 (1panel) to width=128 (2 panels). This is not foreseen in the 2D setup of wled (at least it was not foreseen in the code from 6 months ago 😉 ).

The solution might be to tell the driver to use "vertical" panel layout, which means that chained panels are stacked in one column, instead of putting them into a row next to each other.

This "vertical layout" should make multi-panels (chained panels) compatible with what WLED expects: first pixel = top left, no serpentine, each panel having a separate (not interleaved) pixels range.

wled00/bus_manager.cpp
softhack007, line 772: @DedeHai, @netmindz actually the shift register driver should be user-configurable.
Especially panels with FM6124 will mis-behave with driver=default. In the MM code we made an assumption for simplicity, that "outdoor" panels are always 4-scan FM6124, while "indoor" is always 2-scan with "default" shift register driver -> https://github.com/MoonModules/WLED-MM/blob/d9e542663634812e9136b66bd1d53f187979d33a/wled00/bus_manager.cpp#L684

This assumption works in 90% of all cases, but for a proper solution the user should also have an option to select the shift register chip type.

NB: panel brightness control via display->setbrightness() works better with the correct shift register driver.

wled00/bus_manager.cpp
```diff
@@ -808,38 +774,25 @@ BusHub75Matrix::BusHub75Matrix(const BusConfig &bc) : Bus(bc.type, bc.start, bc.
...
-    // Disable chains of panels for now, incomplete UI changes
-      // if(bc.pins[2] > 1 &&  bc.pins[3] != 0 &&  bc.pins[4] != 0 &&  bc.pins[3] != 255 &&  bc.pins[4] != 255) {
-      //   virtualDisp = new VirtualMatrixPanel((*display), bc.pins[3], bc.pins[4], mxconfig.mx_width, mxconfig.mx_height, CHAIN_BOTTOM_LEFT_UP);
-      // }
+      mxconfig.mx_width = min((uint8_t) 64, panelWidth); // TODO: UI limit is 128, this limits to 64
+      mxconfig.mx_height = min((uint8_t) 64, panelHeight);
   } else if (bc.type == TYPE_HUB75MATRIX_QS) {
-      mxconfig.mx_width = min((uint8_t) 64, bc.pins[0]) * 2;
-      mxconfig.mx_height = min((uint8_t) 64, bc.pins[1]) / 2;
-      virtualDisp = new VirtualMatrixPanel((*display), 1, 1, bc.pins[0], bc.pins[1]);
```
softhack007, line 827: Why did you remove this part? 4-scan panels seem to _require_ using the VirtualMatrixPanel() class, at least that's how I understood the driver documentation.

wled00/bus_manager.cpp
```diff
@@ -944,21 +891,27 @@ BusHub75Matrix::BusHub75Matrix(const BusConfig &bc) : Bus(bc.type, bc.start, bc.
...
+  #define WLED_HUB75_MAX_BRIGHTNESS 255
+  #endif
+  // let's adjust default brightness (128), brightness scaling is handled by WLED
+  display->setBrightness8(WLED_HUB75_MAX_BRIGHTNESS); // range is 0-255, 0 - 0%, 255 - 100%
 
   delay(24); // experimental
   DEBUGBUS_PRINT(F("heap usage: ")); DEBUGBUS_PRINTLN(lastHeap - ESP.getFreeHeap());
   // Allocate memory and start DMA display
-  if( not display->begin() ) {
+  if (!display->begin() ) {
```
softhack007, line 914: Are you sure about this one?
Afaik, driver initially error are indicated with "display->begin() == false" aka "! display->begin()"

softhack007: Sorry my fault. Overlooked the  " ! " in your change. Please ignore this comment.

wled00/bus_manager.cpp
```diff
@@ -988,17 +941,50 @@ BusHub75Matrix::BusHub75Matrix(const BusConfig &bc) : Bus(bc.type, bc.start, bc.
...
+  if (chainLength > 1 && (_rows > 1 || _cols > 1) || bc.type == TYPE_HUB75MATRIX_QS) {
+    _isVirtual = true;
+    chainType = CHAIN_BOTTOM_LEFT_UP; // TODO: is there any need to support other chaining types?
+    DEBUGBUS_PRINTF_P(PSTR("Using virtual matrix: %ux%u panels of %ux%u pixels\n"), _cols, _rows, mxconfig.mx_width, mxconfig.mx_height);
+  }
+  else {
+    _isVirtual = false;
+  }
+
+  if (_isVirtual) {
```
softhack007, line 960: Ah OK, here it is :-)

wled00/bus_manager.cpp
```diff
@@ -808,38 +774,25 @@ BusHub75Matrix::BusHub75Matrix(const BusConfig &bc) : Bus(bc.type, bc.start, bc.
...
+  }
 
   if (bc.type == TYPE_HUB75MATRIX_HS) {
-      mxconfig.mx_width = min((uint8_t) 64, bc.pins[0]);
-      mxconfig.mx_height = min((uint8_t) 64, bc.pins[1]);
-    // Disable chains of panels for now, incomplete UI changes
-      // if(bc.pins[2] > 1 &&  bc.pins[3] != 0 &&  bc.pins[4] != 0 &&  bc.pins[3] != 255 &&  bc.pins[4] != 255) {
-      //   virtualDisp = new VirtualMatrixPanel((*display), bc.pins[3], bc.pins[4], mxconfig.mx_width, mxconfig.mx_height, CHAIN_BOTTOM_LEFT_UP);
-      // }
+      mxconfig.mx_width = min((uint8_t) 64, panelWidth); // TODO: UI limit is 128, this limits to 64
```
softhack007, line 790: This toDo is a bit of a pity ... I have a 128x64 2-scan panel, and it can be used with width=128 height=64... but yes this was never supported in upstream, so ok to address 128x64 panels later in another PR. 
Hint: many 128x64 can also be used as 64x64 chain_lenght=2.

wled00/bus_manager.cpp
softhack007, line 1009: Actually _all_ the fallback code "if (! ledBuffer)" could be removed - settings pixels directly in sPC, instead of doing everything in show() is slow, and it may cause very annoying panel flickering (interference with panel scanline updates). 

In wledMM, I changed the constructor to fail (_valid = false) in case that allocating ledBuffer fails. I removed all the fallback code from sPC(), gPC() and show() which gave a  nice speedup.

wled00/bus_manager.cpp
DedeHai: it is how digital LEDs are handled now so I just did it the same way to not introduce any bugs or future compatibility issues. Actually I had the whole buffer working in 32bit colors but it did not increase speed much and there local scaling is faster. In the curent state it could be handed back to the display driver.

wled00/bus_manager.cpp
DedeHai: I did not check this for speed but it is plenty fast.

wled00/bus_manager.cpp
DedeHai: this is code from @netmindz, I simply added the TODO as I noticed the discrepancy and did not know the reason for this limit.

softhack007: i'm also not sure any more why the limit was added ... maybe just to catch user errors, because 64x64 used to be the max supported panel size. I can confirm that 128x64 works here, too. So it might be safe to increase the "width" limit to 128.

wled00/bus_manager.cpp
softhack007: PS: you might want to keep the fallback code; -S2 is notoriously low on RAM

wled00/bus_manager.cpp
```diff
@@ -1028,8 +1014,8 @@ void __attribute__((hot)) BusHub75Matrix::setPixelColor(unsigned pix, uint32_t c
     uint8_t g = G(c);
     uint8_t b = B(c);
 
-    if(virtualDisp != nullptr) {
-      int x = pix % _panelWidth;
+    if (virtualDisp != nullptr) {
+      int x = pix % _panelWidth; // TODO: check if using & and shift would be faster here, it limits to power-of-2 widths though
```
softhack007, line 1018: comment to your comment: this is in the fallback code, so any optimization will only affect users with "low RAM" problems, like esp32-s2.
"extensa" based chips (esp32, -S3, -S2) have integer division in hardware, which needs ~2 CPU cycles. I'm not sure if "modulo" is a natural result of the division (divmod?), but I'd expect that any optimization here will only win 1 or 2 CPU cycles ... as you say, at the price of dropping support for non-power-of-two panel sizes. Especially for 80x40 which are quite common on ali.

wled00/bus_manager.cpp
DedeHai: S2 does have PSRAM, or at least mine does ;)  not sure you want to run HUB75 on an S2 though...

wled00/bus_manager.cpp
DedeHai: yea, I did notice its in the fallback code. I had it added at first but then removed, this comment is just a leftover from those experiments. In a hot path, every CPU cycle counts, and % is very slow on C3 for example (not that this matters here)

wled00/bus_manager.cpp
softhack007: > not sure you want to run HUB75 on an S2 though

😆 definitely not if you ask me 😜 But there are these "HD-WF2" controllers that have an -S2 with 4MB flash and no PSRAM. These people might still need the low-RAM fallback.

wled00/bus_manager.cpp
```diff
@@ -1041,59 +1027,54 @@ void __attribute__((hot)) BusHub75Matrix::setPixelColor(unsigned pix, uint32_t c
...
-      pix ++;
+      pix++;
     }
     setBitArray(_ledsDirty, _len, false);  // buffer shown - reset all dirty bits
   }
 }
 
 void BusHub75Matrix::cleanup() {
   if (display && _valid) display->stopDMAoutput();  // terminate DMA driver (display goes black)
+  delay(30); // give some time to finish DMA
```
softhack007: (minor) it might be better for thread-safety to change the order of ``_valid=false`` and ``delay(30)`` so that ``_valid=false;`` comes before delay.

softhack007 (commented):
just a minor comment

wled00/bus_manager.cpp
netmindz: Was when i was battling issues with the settings management and so never wanted to use width 255, height 255 or chain length 255 as you were then in a memory based bootloop, so just adding some sanity

netmindz: > This toDo is a bit of a pity ... I have a 128x64 2-scan panel, and it can be used with width=128 height=64... but yes this was never supported in upstream, so ok to address 128x64 panels later in another PR. Hint: many 128x64 can also be used as 64x64 chain_lenght=2.

I believe all 128s are really a pair of 64s. Not sure if was the width of 64 or the height of 64 that needed the extra pin in hub75e, so there might be similar reasons why 128s are built this way

wled00/bus_manager.cpp
netmindz: The issue is the settings UI is a total pain in the ass and needs complete rewrite really. It's a similar issue with the virtual display params, we need the ability for the bus config and UI to allow for an enum type, there is an old issue/pr I started a while back for that, but even getting the basic changes to delegate the led type list to the bus type was slow progress for adoption

wled00/bus_manager.cpp
netmindz: Have you actually tested with suitable panel @DedeHai ?

wled00/bus_manager.cpp
```diff
@@ -853,12 +806,6 @@ BusHub75Matrix::BusHub75Matrix(const BusConfig &bc) : Bus(bc.type, bc.start, bc.
   } else mxconfig.setPixelColorDepthBits(8);
 #endif
 
-  mxconfig.chain_length = max((uint8_t) 1, min(bc.pins[2], (uint8_t) 4)); // prevent bad data preventing boot due to low memory
-
-  if(mxconfig.mx_height >= 64 && (mxconfig.chain_length > 1)) {
-    DEBUGBUS_PRINTLN("WARNING, only single panel can be used of 64 pixel boards due to memory");
-    mxconfig.chain_length = 1;
-  }
```
netmindz, line 861: Not sure it's safe to just remove this, bad for users to be in low memory boot loop

netmindz:
Had a quick test with a basics single panel setup. UI improvements look good. Need to test multiple panels setups.
Need to double check but I think the brightness looked lower, I need to revert to compare

wled00/bus_manager.cpp
DedeHai: the new heap management should take care of that and not allow buffer allocation to below the set threshold.

DedeHai:
> I think the brightness looked lower, I need to revert to compare

brightness should not change, its still set to 255 at driver level. New gamma correction handling may play a role.

blazoncek:
@netmindz @DedeHai & @softhack007 I've backported @netmindz's Hub75 to my fork (0.15) where I made some similar changes as @DedeHai made (no coordination) here. 😄
However, I've also enhanced UI to be aware of GPIO used by Hub75 driver and allowed for dropdown use in LED GPIO.
I do not have any hardware to test so my coding was pure guesswork and logic analysis of flow.

One thing I'd comment here is the removal of HW assisted brightness scaling in this PR. It is a waste of CPU cycles to do any scaling on bus level IMO. For NPB it makes sense as that was purely SW.

The other thing is `getPixelColor()`. Bus level `gPC()` is no longer needed since the introduction of my segment blending. Any and all `getPixelColor()` happen on segment buffers and is not affected by final or segment brightness/opacity. Realtime functions probe WS2812FX buffer instead of bus driver.
This means that drivers no longer need to maintain pixel data and any "double" buffering can be removed as it is done in WS2812FX class.

DedeHai:
> One thing I'd comment here is the removal of HW assisted brightness scaling in this PR. It is a waste of CPU cycles to do any scaling on bus level IMO. For NPB it makes sense as that was purely SW.

I did that to be consistent with NPB handling, as I just learned that HUB75 actually support HW-brightness. I need to check that but if it is true hardware scaling, we should definitely NOT do software scaling. so thanks for pointing this out.

> The other thing is `getPixelColor()`. Bus level `gPC()` is no longer needed since the introduction of my segment blending. 

IIRC there may be some usermods that need bus-level gPC(), at least thats what I remember from the top of my head, because I was close to removing them when I updated the ABL & brightness scaling code.  Not sure this is actually true or just coming from some outdated, lingering inline comment.

blazoncek:
> IIRC there may be some usermods that need bus-level gPC(),

None should do that IMO. I would remove bus level gPC/sPC from public API.

There are benefits of having bus level ABL *and* strip (AKA WS2812FX) level ABL. I would keep both.

wled00/bus_manager.cpp
blazoncek: @softhack007 logic done (including I2S bus frequency). @netmindz not PITA when you know what to do, however it is a hack really. 😉 
See my fork (it even compiles now but I have no way of knowing if it works 😂 ).

softhack007:
> . I need to check that but if it is true hardware scaling,

i think it is "kind of" hardware scaling. the hub75 driver will PWM-modulate the "enable" signal, instead of scaling individual pixels. But this feature depends on having the correct "shift register" driver in HUB75 mxconfig.

softhack007:
> This means that drivers no longer need to maintain pixel data and any "double" buffering can be removed as it is done in WS2812FX class.

Hi @blazoncek, hope you are doing well :-)

About your comments:
* It might be that in general, busses.getPixelColor() could be removed if its no longer needed. I'm not going to interfere with any design decision on this, it has to be decided by the team. However I'd like to keep the scope of PRs clean, meaning that "remove gPC from bus level" should be a separate PR.
*  "drivers no longer need to maintain pixel data and any "double" buffering can be removed" -> this is partially true. One reason that HUB75 has its own double buffer was the need to support getPixelColor on bus level. 
* * There is also a second purpose - buffering all pixel writes until "busses.show()" is necessary to avoid flickering. HUB75 works similar to a TV screen, and the content is continuously refreshed. So if we forward each "setPixelcolor" directly to the HUB75 hardware driver, it results in flickering and tearing. The HUB75 driver offers its own double-buffer solution for this, but it needs a lot more RAM for this buffer - so having a CRGB "FastLED" style buffer is actually removing flickering (all pixels pushed at once), and it reduces the RAM needed by a large amount (like 24Kb saved for one 64x64 panel).
* * An idea for upstream WLED - which I haven't tried yet - is to directly give the busHUB75 driver access to the main "frame buffer" at strip level. It means that hub75.show() can directly push all pixels from the upper level framebuffer (and yes, the "pixel pusher" must be at busHUB75 level !), without needing a buffer of its own. In this scenario, hub75.sPC() will only manage the "dirty bits" for optimization, but we could avoid the full double buffering by accessing the already existing one on upper level.
* Finally - you should try a HUB75 panel, these things are really fun 😀 . 
  I would suggest to start with a 64x64 "indoors" P2 or P3 module - make sure to look at brightness (nits) and contrast range (1:800 or more). These panels are quite affordable (~25€ on ali). Then get the "Moonhub" adapter and the LiliGo "T7-S3" with PSRAM, connect and have fun. I'm sure that @lost-hope would be happy to help you select a panel and get started :-)

Cheers, and take care,
 Frank.

Edit: just for the fun of it - run the below .gif in image effect on HUB75, once you have the hardware
ghostbusters64

DedeHai:
> * There is also a second purpose - buffering all pixel writes until "busses.show()" is necessary to avoid flickering. HUB75 works similar to a TV screen, and the content is continuously refreshed. So if we forward each "setPixelcolor" directly to the HUB75 hardware driver, it results in flickering and tearing. The HUB75 driver offers its own double-buffer solution for this, but it needs a lot more RAM for this buffer - so having a CRGB "FastLED" style buffer is actually removing flickering (all pixels pushed at once), and it reduces the RAM needed by a large amount (like 24Kb saved for one 64x64 panel).

I did not check on how the HUB75 driver uses its internal buffer(s?), the additional CRGB buffer is needed for determining the "dirty" pixels, if we could access the HUB75 driver's internal buffer for that, we would not need it. I ran quite a few experiments when writing this PR, like removing the buffer and the "dirty" bit buffer and makeing the CRGB buffer a uint32_t buffer (slight seed increase) and even using the highest byte of that to mark "dirty" pixels but came back to the way it was i.e. seperate dirty pixel buffer (1 bit per pixel) and CRGB as it yielded highest throughput at lowest RAM cost. Fastest would be: seperate dirty pixel buffer with uint32_t colorbuffer. Best for RAM use is to grant access to the busmanager's _pixels[] buffer and not use dirty.

just FYI: here are my raw notes about those experiments (translated by chatGPT):
```
PS Fire and Black Hole speed test:
With CRGB buffer and dirty: 48 FPS (fire) and 92 FPS (black hole)
Without CRGB buffer and without dirty: 47 FPS and 64 FPS
With uint32_t buffer (non-PSRAM) and without dirty: fire 45 FPS, black hole 63 FPS, palette 45 FPS (huh?) → Probably the loop makes access to the panel much (much) faster.

With uint32_t buffer + dirty (and show(), i.e. inside the loop):

fire: 52 FPS
black hole: 108 FPS
palette: 47 FPS (this is without color_scale!)
with color_scale it’s about 1 FPS lower
DNA: 66 FPS

With dirty stored in the high byte instead of a separate array:

black hole: 58 FPS
DNA: 46 FPS
fire: 44 FPS
→ massively worse.

With dirty in the high byte (not a separate array) and IRAMATTR on SpC: about +1–2 FPS.

With 32-bit buffer, with dirty, with color scale:

fire: 47 FPS
black hole: 102 FPS
palette: 40 FPS
DNA: 89 FPS

With else-if in color fade instead of if, if: 47 FPS / 102 FPS / 88 FPS → no change.

What if you drop all the dirty-tracking stuff and SpC entirely, and just write directly to the array inside the loop?
For that the array must be accessible → you can do it with getPixelRaw.

→ Hack with direct access to _pixels[], including scaling:

black hole: 63 FPS
DNA: 48 FPS
fire: 46 FPS
palette: 45 FPS

Conclusion:
Dirty tracking helps. The fastest is with dirty + draw loop.
So basically exactly like now, just without CRGB.
CRGB uses less RAM, and it’s not that much slower… only a few FPS.
But for sparse scenes it does make a difference (92 FPS → 108 FPS).

Maybe there’s an even better variant, but for now it’s fine as it is.
```

blazoncek:
@softhack007 you are missing my point. :wink:
* With current version of 0.16 *there are no calls* to `BusManager::getPixelColor()`, hence no calls to any of bus' `getPixelColor()`. This doesn't mean it has to be stripped right away (about 1k of code). It's the team's decision to do that, as you say. I'm merely pointing out the fact.
* The same goes to updating actual driver's buffer. It is only updated in the `WS2812FX::show()` and nowhere else. Flickering will only occur if you call `show()` midframe while driver is still updating LEDs.
* IMO giving driver access to frame buffer (`WS2812FX::_pixels[]`) may not work as desired as realtime functions write directly into that buffer and expect `show()` to do proper blending. And the buffer is not properly updated until `blendSegment()` is finished and all realtime calls are done.
* As for the panel, I have started to clean my desk and I do not intend to spend additional money on something I would only use for development (but I would not refuse a gift). I'm purely enjoying logical challenges associated with tough coding choices. However my code is free for anyone to use as a basis for their own improvements. That's why I'm posting here.

The last statement includes improvements to LED settings page where Hub75's GPIO are taken into account (and all LED outputs get GPIO dropdowns) and ability to select shift register via dropdown. I will not make a PR for that due to lack of time, but you are free to cherry pick and adapt as needed.

blazoncek:
For anyone interested: UI code for enabling Hub75 panels has been improved and is now working as expected (limiting dimensions to multiples of 32, allowing maximum of 6 panels, adding ability to select shift register via drop-down (currently a hack but it could be made driver-supplied), allocating GPIO in UI and preventing them to be used for other outputs).

One thing that is not clear to me from the MM or Netmindz's code is if it is allowed to have Hub matrix with dimensions of 16, 48, or 96 (in fact any multiple of 16) and if you can arrange panels in a similar way as 2D setup allows (multiple rows and columns, no UI for that in LED settings page).

softhack007:
> One thing that is not clear to me from the MM or Netmindz's code is if it is allowed to have Hub matrix with dimensions of 16, 48, or 96 (in fact any multiple of 16) and if you can arrange panels in a similar way as 2D setup allows (multiple rows and columns, no UI for that in LED settings page).

@blazoncek Thanks for your effort :-)

About the "multiple of 16" dimensions question: yes there are still 16x16 and 16x32 panels available, especially the "outdoor" ones with high brightness and pixel spacing "P10" or greater. 

* the HUB75 driver does not support mixing panels with different dimensions. ==> total width or height 16/48/80/96/112 is possible, but only when the fixture size is a multiple of one single "base panel size", and the base size is divisible by 8.
* in our original WLED-MM code, we made our (dev) life a bit simpler by just allowing some of these panel sizes (predefined types in a drop-down), and any fixture must be a multiple of these base sizes
* I've also seen panels with "20x40", "80x40", "96x48", and "160x80" (2-scan, 4-scan, or 8-scan) in the wild, but we never tested them - so "may or may not work".
* Another type of panels is 8-scan (in contrast to 2-scan (indoors) and 4-scan (outdoors)). Not sure if these work in the MrCodetastic driver.

* For WLED 2D setup to work, you _*must* arrange your panels vertically_ -- like 1 (wide) x 4 (high) . Other arrangements lead to interleaving within scanlines, and our 2D setup cannot work with panels when they have interleaved (overlapping) pixel ranges.

softhack007:
> Conclusion:
> Dirty tracking helps. The fastest is with dirty + draw loop.
> So basically exactly like now, just without CRGB.
> CRGB uses less RAM, and it’s not that much slower… only a few FPS.
> But for sparse scenes it does make a difference (92 FPS → 108 FPS).
> 
> Maybe there’s an even better variant, but for now it’s fine as it is.


Thanks @DedeHai for this interesting comparison :-)

It matches with my observations. The effect of dirty bits tracing is even stronger in the 0.14.x-based MM codebase (no pixel buffer on bus level) - there it leads to 2x speedup in "show". Maybe i'll do some experiments with ``uint32_t`` pixels, to check if this can speed up things even a bit more. However, it would only be an option for "power users" with S3 and octal PSRAM.

<br/>

I see a few more optimizations, that were not "exploited" in MM yet:

* if memory gets tight, the dirty buffer could also be for a group of pixels, e.g. two (or four) pixels next to each other share one "dirty" flag - example: one pixels changes, a set of 4 marked dirty at once. That's how cache coherency protocols work, which was my original inspiration for _dirty bits_.

* my "dirty bits" code could even be moved up to bus level, where the final framebuffer lives, too. 
  - Dirty bits would optimize sPC calls for any kind of NeoPixelBus output, especially we could avoid pushing _all_ pixels at _each_ "busses.show()". 
  - busHUB75  could remove sPC completely, and 
  - only have a special "show()" function that gets all parameters from BUS level (pixel pushing is fastest on lowest level)
     a) pointer to Framebuffer (start = first HUB75 pixel)
     b) pointer to dirty Bits buffer (start = first HUB75 pixel)
     c) (maybe) logical "width" of a scanline in the Framebuffer.


Edit 1: 
There are no "RGB+White" HUB75 panels, and most likely there never will be, because a 4th color component does not fit into the standard HUB75 16pin connector. So "True white channel" support on HUB75 is for me a purely theoretical exercise 😜 

Edit 2:
>  if we could access the HUB75 driver's internal buffer for that, we would not need it. 

that's the original problem - the MrCodetastic driver directly translates each pixel into and internal DMA buffer, which stores some kind of PWM waveform for output. This is not accessible from outside, and I doubt that we could restore the original pixel even when we had acess to the DMA buffer memory.
 
BTW, this internal DMA buffer is the reason why I called the driver "memory greedy" - the internal buffer for 64x64, 24 bit color is a full 32KB (sic!), about 3x the size of individual pixels.

blazoncek:
> about 3x the size of individual pixels

FYI it is similar with I2S in NPB. Called cadence there.

> * especially we could avoid pushing _all_ pixels at _each_ "busses.show()".

Not sure this would work with layered/overlapping segments. Do not forget that all segments are blended and until upstream has transparency (which should be supported in effects too) _all_ pixels are updated during `show()`.

softhack007:
> _all_ pixels are updated during `show()`.

@blazoncek I see your point. Actually, the dirty bits logic can _possibly_ still help, at least when the framebuffer on lowest level does not get too many "intermediate" updates. 

In a nutshell:
* pixel update: ``if (newColor != framebuffer[pixel]) { dirty[pixel] = true;  framebuffer[pixel]=newColor; }``
* show all (bus pixel pusher):  ``if ( !dirty[pixel] ) continue; // skip if color unchanged``
* after show: all ``dirty[] = false;``

Edit: it could imply that the "blending" and "overlaying" logic needs to be modified, so that the final pixel color is calculated ONCE (loop over all bus pixels, blend/overlay every "upper" pixel;  instead of going segment-by-segment). But the result may be worth it, because you'll always find a lot of pixels that actually did not change between two frames, so its not necessary to push them down to the NPB driver again.

blazoncek:
I'll leave optimizations for more competent people (or people with more time). :smile:

However, having *transparent* pixels would also help (to some extent).

You could have *dirty* flag at WS2812FX level (which would then benefit *any* driver) and it is rather simple to implement. Not much different than what you posted and I don't think blending would need changes (apart from comparing old vs. new). Dirty buffer can be temporary.

blazoncek:
> Finally - you should try a HUB75 panel, these things are really fun 😀 .

I said I won't, but I did. 🤦‍♂️

blazoncek:
FYI I have entirely dynamic Hub75 support ready. No need for custom compiles for different boards or configurations.
It even plays nicely with usermods.

blazoncek:
@softhack007 you seem most knowledgeable about Hub75 implementation in WLED.
Looking at the code I can see that if `_ledBuffer` allocation fails the `BusHub75Matrix` class can work without it. However, `show()` serves no purpose in such case. How is then display synchronised? I.e when does it get information to display modified pixels?

softhack007:
> @softhack007 you seem most knowledgeable about Hub75 implementation in WLED.
> Looking at the code I can see that if `_ledBuffer` allocation fails the `BusHub75Matrix` class can work without it. However, `show()` serves no purpose in such case. How is then display synchronised? I.e when does it get information to display modified pixels?

Hi @blazoncek yes ``_ledBuffer == nullptr`` is a special case, call it the "limb home on a stick` mode:
* no synchronization possible, pixels are directly put into ``virtualDisp->drawPixelRGB888(int16_t(x), int16_t(y), r, g, b)`` as they arrive in ``busses.setPixelColor()`` --> can lead to heavy flickering (i've seen the flicker during my first coding attempts)
* busHUB75 does not know about previous pixels, so getPixelColor() returns GRAY if the "dirty" bit is set
* its true that ``show`` does nothing in this case, ecxept for clearing the dirty pixel bits.

In fact I was thinking about removing this mode and simply set the bus driver ``_valid=false`` when _ledBuffer allocation fails. But then someone reminded me of the HD-WF2 (S3 _without PSRAM_) that might actually still need this fallback mode.

I don't know exactly how the HUB75 hardware driver achieves synchronisation - but when pushing all pixels in busHUB75::show(), I've never seen _any_ sync problems. 🤔 Maybe it synchronizes implicitly on pixel (0,0), and then you just have to go fast enough to stay ahead of the "scan beam" that refreshes the display at 60-120hz.

Currently (with dirty bits) busHUB75:show() takes around 3-8ms for my 192x64 pixels test setup, so we are always ahead of the "beam" = refreshing pixels in the driver fast enough (at 250 to 120 fps) so the beam is behind us.

blazoncek:
Hmm.
@softhack007 have you examined new way WLED (0.16) treats output? AKA segment blending. With this approach `BusXXXX::show()` is needed to:

1. limit current on digital LEDs (and start NPB)
2. schedule LEDC PWM channels
3. transmit network buffer

If what you describe is true, then `show()` is not needed for Hub75 panels as all pixels are drawn immediately prior calling `show()` in `WS2812FX` class. Consquentially `_ledBuffer`is also not needed.

DedeHai:
> Consquentially `_ledBuffer`is also not needed.

on the contrary, it even needs a second buffer to mark "dirty" LEDs which significantly improves frame rate.

blazoncek:
> on the contrary, it even needs a second buffer to mark "dirty"

Even if all pixels are redrawn in `WS2812FX::show()`? The only "drawback" I see in that chain is the conversion from 2D to 1D and then back to 2D (which is skipped with `_ledBuffer`) in `Bus::setPixelColor()`.

However if `Bus::setPixelColor()` would be provided with 2D coordinates then no reverse translation would be needed.
That is easy to accommodate.

blazoncek:
I'll dump this here.

https://github.com/user-attachments/assets/3321661e-0a59-4aaa-acaa-5f9313ddca01

Screenshot 2025-11-27 at 16 22 34

DedeHai:
> I'll dump this here.

awesome! PR when? :)

blazoncek:
> PR when?

I need someone with actual Hub75 panels to test 1st. (Until I get mine)

softhack007:
> I'll dump this here.

@blazoncek settings page looks good :-) all the options I can imagine are there, only two panel sizes still missing: 16 (high or wide), 128 (wide only)
 
I can't test right now, because I'll be "away from HUB75" (on business trip) for the next two weeks.

softhack007:
> Even if all pixels are redrawn in `WS2812FX::show()`? The only "drawback" I see in that chain is the conversion from 2D to 1D and then back to 2D (which is skipped with `_ledBuffer`) in `Bus::setPixelColor()`.

Hi @blazoncek the main problem is speed. And as you already noticed, the last "stone" is the repeated 1D to 2D translation. But there is more:

If we go through ``BusManager::setPixelColor()`` for each pixel, we lose time in the orchestrator ``for`` loop (even with only one bus), in the range check, in the repeated array access and function call indirection, and finally in a few nested function calls. Also, we put stress on the small CPU instruction cache, which may not be able to cope with all these indirections efficiently. 

https://github.com/wled/WLED/blob/ae37f4268cbea8e714281b32321a903c5aadbe65/wled00/bus_manager.cpp#L1331-L1333

It may look like tiny amounts of CPU cycles, but considering that you only have 5ms to push 128x64 pixels out to the driver, it might become a real problem. Also you did not show us yet where to place the dirty bits array - which created a major speedup - in your proposal.

DedeHai:
@softhack007 @blazoncek  since HUB75 is currently pretty unusable in main without this PR: how about merging this and then take it from there with the improvements?

blazoncek:
> If we go through `BusManager::setPixelColor()` for each pixel, we lose time in the orchestrator `for` loop

@softhack007 but you do that in any case. There is no way around it unless you hack and use `bus->setPixelColor()` in `WS2812FX::show()`. I encourage you to sacrifice one of your devices to test 0.16 and its optimised workflow which only paints bus pixels once.

> Also you did not show us yet where to place the dirty bits array

As each bus pixel is only written once (be it black or not and never read) there is no need for dirty array as each pixel is considered dirty on new frame. However if one *really wants* dirty array I have a solution for that too. My fork uses CRGBA instead of (CRGB or uint32_t) which encodes transparency into white channel (unused on Hub75). Alpha channel *can* be used as a dirty flag (question remains what if previous frame had such pixel non-black). If one doesn't want CRGBA you can always encode dirty flag in WW part of uint32_t.

blazoncek:
> then take it from there with the improvements

I'm not willing to spend additional time adjusting the code again. Sorry. If you do merge then you'll need to pick from my fork manually if you want my modifications.
IMO there is no need to rush Hub75.

DedeHai:
RE: speed considerations: I have tested various scenarios, as I mentioned before. The code from this PR is the way to go.

blazoncek:
Got my Adafruit MatrixPortal S3 and two 64x32 panels. We'll see how it goes.
Flashing PIO provided bootloader & partitions seem to work though I can see several reboots on start but then it settles.

softhack007:
> @softhack007 @blazoncek since HUB75 is currently pretty unusable in main without this PR: how about merging this and then take it from there with the improvements?

I agree - let's first concentrate on what we have in this PR, there will be enough time to enhance and extend on it. 

@DedeHai I have the feeling that the PR is kind of "stalled" in mid-air right now 🤔 do you see things that are worth fixing before we merge?

DedeHai:
@softhack007 the main reason I did not yet merge yet is 

>I'm not willing to spend additional time adjusting the code again. Sorry. If you do merge then you'll need to pick from my fork manually if you want my modifications.
IMO there is no need to rush Hub75.

blazoncek:
You don't need to wait for me anymore.

PR #5040, opened 2025-11-06
Improved 1D support for GIF images, bugfixes, blur option

DedeHai:
Instead of showing a scaled, single line of the GIF: map the full image to the strip.

This is needed to be able to load data from GIF's instead of using the "i" JSON command and will allow to store more images/animations for 1D setups. Required for the Pixel Paint Tool and will also make support for the Matrix Tool for 1D setups possible.

edit: 
- also fixes a bug in 2D: need to use virtual dimensions or scaling calculation is wrong
- plus major speedups for drawPixelCallback (@DedeHai, @softhack007)
- improved error handling; and graceful exit at OOM (@softhack007)
- "Blur" option added to image effect (@softhack007)

## Summary by CodeRabbit

* **New Features**
  * Persistent per-image rendering state, three pixel-rendering modes (exact-size, 1D strip, scaled 2D) selected at load time, and optional post-decode blur.

* **Bug Fixes**
  * More robust GIF handling: filename and size validation, safer decoder allocation and decode error reporting, frame-delay compensation, bounds protections, single-active-segment enforcement, and improved cleanup after playback.

* **Chores**
  * Image mode metadata extended to include a "Blur" descriptor and explicit index offset.

### DIFF
--- a/wled00/image_loader.cpp
+++ b/wled00/image_loader.cpp
@@ -10,7 +10,7 @@
  */
 
 static File file;
-static char lastFilename[34] = "/";
+static char lastFilename[WLED_MAX_SEGNAME_LEN+2] = "/"; // enough space for "/" + seg.name + '\0'
 static GifDecoder<320,320,12,true> decoder;  // this creates the basic object; parameter lzwMaxBits is not used; decoder.alloc() always allocated "everything else" = 24Kb 
 static bool gifDecodeFailed = false;
 static unsigned long lastFrameDisplayTime = 0, currentFrameDelay = 0;
@@ -115,10 +115,10 @@ byte renderImageToSegment(Segment &seg) {
   if (activeSeg && activeSeg != &seg) return IMAGE_ERROR_SEG_LIMIT; // only one segment at a time
   activeSeg = &seg;
 
-  if (strncmp(lastFilename +1, seg.name, 32) != 0) { // segment name changed, load new image
+  if (strncmp(lastFilename +1, seg.name, WLED_MAX_SEGNAME_LEN) != 0) { // segment name changed, load new image
     strcpy(lastFilename, "/");  // filename always starts with '/'
-    strncpy(lastFilename +1, seg.name, 32);
-    lastFilename[33] ='\0';     // ensure proper string termination when segment name was truncated
+    strncpy(lastFilename +1, seg.name, WLED_MAX_SEGNAME_LEN);
+    lastFilename[WLED_MAX_SEGNAME_LEN+1] ='\0';     // ensure proper string termination when segment name was truncated
     gifDecodeFailed = false;
     size_t fnameLen = strlen(lastFilename);
     if ((fnameLen < 4) || strcmp(lastFilename + fnameLen - 4, ".gif") != 0) { // empty segment name, name too short, or name not ending in .gif
@@ -222,7 +222,7 @@ void endImagePlayback(Segment *seg) {
   decoder.dealloc();
   gifDecodeFailed = false;
   activeSeg = nullptr;
-  lastFilename[1] = '\0';
+  strcpy(lastFilename, "/");  // reset filename
   DEBUG_PRINTLN(F("Image playback ended"));
 }
 


DedeHai:
@softhack007 you may want to cherry pick this to MM if you want full paint and matrix tool support :)

wled00/image_loader.cpp
```diff
@@ -52,13 +52,25 @@ void screenClearCallback(void) {
...
-  int16_t outY = y * activeSeg->height() / gifHeight;
-  int16_t outX = x * activeSeg->width()  / gifWidth;
-  // set multiple pixels if upscaling
-  for (int16_t i = 0; i < (activeSeg->width()+(gifWidth-1)) / gifWidth; i++) {
-    for (int16_t j = 0; j < (activeSeg->height()+(gifHeight-1)) / gifHeight; j++) {
-      activeSeg->setPixelColorXY(outX + i, outY + j, red, green, blue);
+  if (activeSeg->height() == 1) {
+    // 1D strip: load pixel-by-pixel left to right, top to bottom (0/0 = top-left in gifs), scale if needed
+    int totalImgPix = (int)gifWidth * gifHeight;
+    int stripLen = activeSeg->width();
```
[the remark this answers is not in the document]
DedeHai: good point.

softhack007:
> @softhack007 you may want to cherry pick this to MM if you want full paint and matrix tool support :)

Hi @DedeHai , good point I'll probably pick this one.

Btw, looks like we both found the same scaling bug :-) the bug also causes wrong behaviour with "mirror" at segment level.

 I'm currently working on a PR to improve code quality/performance of image loader (see https://github.com/MoonModules/WLED-MM/pull/279) and I'll PR to upstream once I'm happy with test results

@netmindz this PR needs to be merged soon, to avoid merge conflicts with my upcoming PR that will also touch image_loader.

wled00/image_loader.cpp
```diff
@@ -52,13 +52,25 @@ void screenClearCallback(void) {
...
 
 void drawPixelCallback(int16_t x, int16_t y, uint8_t red, uint8_t green, uint8_t blue) {
-  // simple nearest-neighbor scaling 
-  int16_t outY = y * activeSeg->height() / gifHeight;
-  int16_t outX = x * activeSeg->width()  / gifWidth;
-  // set multiple pixels if upscaling
-  for (int16_t i = 0; i < (activeSeg->width()+(gifWidth-1)) / gifWidth; i++) {
-    for (int16_t j = 0; j < (activeSeg->height()+(gifHeight-1)) / gifHeight; j++) {
-      activeSeg->setPixelColorXY(outX + i, outY + j, red, green, blue);
+  if (activeSeg->height() == 1) {
```
softhack007: @DedeHai it might be good (both for code clarity and for performance) to introduce separate callback functions, one for 1D and one for 2D.
I think that callbacks can be changed on-the-fly in ``renderImageToSegment`` with ``decoder.setDrawPixelCallback();``.

It would also reduce merge conflicts with my upcoming PR. because i'm changing the 2D draw functions by moving calculations out of the function that will not change during a "draw loop".

wled00/image_loader.cpp
```diff
@@ -52,13 +52,25 @@ void screenClearCallback(void) {
...
+    // 1D strip: load pixel-by-pixel left to right, top to bottom (0/0 = top-left in gifs), scale if needed
+    int totalImgPix = (int)gifWidth * gifHeight;
+    int stripLen = activeSeg->vWidth();
+    if (totalImgPix - stripLen == 1) totalImgPix--; // handle off-by-one: skip last pixel instead of first
```
softhack007: the three lines above will not change during a "draw loop", and could be moved into ``renderImageToSegment()``

wled00/image_loader.cpp
DedeHai: will it help performance if they are separated? GIFs are read from flash so I would expect that to be the bottle neck (by far). Or do you have plans to run animations from RAM? anyway, just point me to your code and I can try to merge it in.

wled00/image_loader.cpp
DedeHai: good point.

wled00/image_loader.cpp
softhack007: > will it help performance if they are separated? GIFs are read from flash so I would expect that to be the bottle neck (by far). 

well it clearly depends on image size. starting with ~128x128 animated gifs, I see that improvements in the drawPixelCallback do improve framerates. I still think that keeping the two "variants" as separate functions will make it easier to read the code.

> anyway, just point me to your code and I can try to merge it in.

Thanks for offering - the existing code is simple enough so no problem for me to cross-merge stuff. Once your PR is merged to main, I'll adapt my own PR for upstream and you can review my PR.

<br/>

just for info, my modified draw2DPixelCallback is here:
https://github.com/MoonModules/WLED-MM/blob/d78ea5ae2edd5a4fb03b8ad9947b0b1c64d7165f/wled00/image_loader.cpp#L73-L88


cheers, Frank.

softhack007: > Or do you have plans to run animations from RAM?

nothing yet - however @troyhacks has done some experiments on his P4 branch for pre-caching *large* animations in PSRAM. But he used MJPEG (from SD card), not GIF.

DedeHai:
@softhack007  I have taken a closer look on speed improvements and added my take on it: 
- scaled gif drawing now does not set same pixel multiple times
- no recalculation of static variables (except totalImgPix which is a simple multiplication to save precious RAM)
- only use scaling if actually needed, directly sets pixels if gif size matches ("fastpath")

like this, if you want highest speed, use a gif that does not need "live" scaling. could even remove the static variable pre-calculation to save RAM, making scaling a tiny bit slower (but still faster than it was as double drawing is now omitted)

wled00/image_loader.cpp
```diff
@@ -44,20 +44,42 @@ bool openGif(const char *filename) {
...
-void drawPixelCallback(int16_t x, int16_t y, uint8_t red, uint8_t green, uint8_t blue) {
-  // simple nearest-neighbor scaling 
-  int16_t outY = y * activeSeg->height() / gifHeight;
-  int16_t outX = x * activeSeg->width()  / gifWidth;
+// note: GifDecoder drawing is done top right to bottom left, line by line
+
+// callback to draw a pixel at (x,y) without scaling: used if GIF size matches segment size (faster)
+void drawPixelCallbackNoScale(int16_t x, int16_t y, uint8_t red, uint8_t green, uint8_t blue) {
+  activeSeg->setPixelColor(y * activeSeg->width() + x, red, green, blue);
+}
```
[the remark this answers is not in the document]
DedeHai: no, this is using 1D on purpose to make it universal

wled00/image_loader.cpp
```diff
@@ -44,20 +44,42 @@ bool openGif(const char *filename) {
...
+    activeSeg->setPixelColor(start + i, red, green, blue);
+  }
+}
+
+void drawPixelCallback2D(int16_t x, int16_t y, uint8_t red, uint8_t green, uint8_t blue) {
+  // simple nearest-neighbor scaling
+  int outY = (int)y * activeSeg->vHeight() / gifHeight;
+  int outX = (int)x * activeSeg->vWidth()  / gifWidth;
+  if (outX + outY == lastCoordinate) return; // skip setting same coordinate again
+  lastCoordinate = outX + outY; // since input is a "scanline" this is sufficient to identify a "unique" coordinate
```
softhack007: This is mainly to save 2 "precious bytes" of RAM, right?

DedeHai: no, it does not save ram but its a simple addition as opposed to doing a multiply-add with the exact same result in the hope that the compiler can optimize better. 
you could store both coordinates and add another condition to if but that is even slower.

wled00/image_loader.cpp
```diff
@@ -44,20 +44,42 @@ bool openGif(const char *filename) {
 
 Segment* activeSeg;
 uint16_t gifWidth, gifHeight;
+int lastCoordinate; // last coordinate (x+y) that was set, used to reduce redundant pixel writes
+uint16_t perPixelX, perPixelY; // scaling factors when upscaling
```
softhack007: (style hint) The original code from AC is having all variables as global, which is kind of "polluting" the global namespace and inviting for future trouble ... I would suggest to make all variables "static" when they don't need to be visible outside of this file.

softhack007:
@DedeHai would you mind if I simply add my other changes that I wanted to bring over from MM directly in your PR? (by maintainers superpower). This way we can keep working together directly, and code authors will still be visible.

I'll leave the drawpixel functions as they are, and just add some error handling plus the additional blur slider.

wled00/image_loader.cpp
softhack007: Ah ok, makes sense 👍

wled00/image_loader.cpp
DedeHai: if making them static, there will be duplacates for 1D and 2D. Actually could the image loader be a class that is only instantiated if needed? or is using namespace a better solution (regarding fragmentation)?

DedeHai:
> would you mind if I simply add my other changes that I wanted to bring over from MM directly in your PR?

sure, go ahead.

softhack007:
> sure, go ahead.

Great, i'll do it later today, or tomorrow.

wled00/image_loader.cpp
softhack007: Ah, misunderstanding. You can keep the variables declared where they are, just change ``int lastCoordinate`` into ``static int lastCoordinate``.

This means they are still global on file level, but not accessible from other files like FX.c. Helps the compiler to optimize, and avoids name clashes. Namespaces basically do the same thing, however I've not used them yet.

wled00/image_loader.cpp
```diff
@@ -11,7 +11,11 @@
 
 static File file;
 static char lastFilename[34] = "/";
-static GifDecoder<320,320,12,true> decoder;
+#if !defined(BOARD_HAS_PSRAM)
+  static GifDecoder<256,256,11,true> decoder;  // use less RAM on boards without PSRAM - avoids crashes due to out-of-memory
+#else
+  static GifDecoder<320,320,12,true> decoder;
```
DedeHai: now that I think about it: why is this static? would it not make more sense to make it a pointer only and use new and delete? or does it not use much ram until a gif is actually loaded?

edit: if RAM use of the GifDecoder is dynamic (which I assume) this restriction does not really make sense, it also complicates things with my painting and video tool as there I currently use 320 pixels as the max width for saving a "1D GIF", which this PR is all about.

softhack007: I think the "gifdecoder" is just the basic object with some variables and pointers.
the call to ``decoder.alloc();`` creates the full "Gif" object internally, with line buffers and LZW table that can be big.

softhack007: actually, if we want to support several segments showing gif's in parallel, then we need several GifDecoder objects.

DedeHai: would a better solution be to estimate that heap use, check for available heap and throw a "Not enough RAM" error to the UI?
edit:
or allocate first, then check remaining heap and deallocate if too low (see util.cpp)

softhack007: > there I currently use 320 pixels as the max width for saving a "1D GIF", which this PR is all about.

good point. then lets stay with 320x320 - it was just my "educated guess" that smaller image and smaller LZW tables would require less RAM. But i did not measure this. Any way the exception handler will prevent crashing now if the decoder request too much RAM.

DedeHai: > Any way the exception handler will prevent crashing now if the decoder request too much RAM.

only if it way too much, if the remaining heap is too low, the UI (or WIFI in general) breaks down... happens around 15k of contiguous heap.

DedeHai: the safest solution would be to use the same logic as in `validateFreeHeap()` to make sure we do not run into low heap problems

softhack007: another idea: we could ask AC to modify his library, so it accepts a custom "malloc" callback function. then we can force ``decoder.alloc()`` to utilize your custom malloc() via callback.

Edit: ``decoder.alloc()`` should then return ``bool``, so that image rendering can be disabled at OOM errors.

DedeHai: that would be even better

softhack007: .. and it would be cleaner :-)

wled00/image_loader.cpp
```diff
@@ -35,29 +39,64 @@ int fileSizeCallback(void) {
...
+  if (outX + outY == lastCoordinate) return; // skip setting same coordinate again
+  lastCoordinate = outX + outY; // since input is a "scanline" this is sufficient to identify a "unique" coordinate
   // set multiple pixels if upscaling
-  for (int16_t i = 0; i < (activeSeg->width()+(gifWidth-1)) / gifWidth; i++) {
-    for (int16_t j = 0; j < (activeSeg->height()+(gifHeight-1)) / gifHeight; j++) {
+  for (int i = 0; i < perPixelX; i++) {
+    for (int j = 0; j < perPixelY; j++) {
       activeSeg->setPixelColorXY(outX + i, outY + j, red, green, blue);
     }
   }
```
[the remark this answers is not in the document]
softhack007: @DedeHai do you think this might be problematic? 
The point of "Hasi" is that down-scaling might lead to collisions / wrong rejections 🤔

wled00/image_loader.cpp
```diff
@@ -35,29 +39,64 @@ int fileSizeCallback(void) {
...
+    else activeSeg->blur(blurAmount);                                                              // more blur - slower
+  }
+  lastCoordinate = -1; // invalidate last position
+}
+
+// note: GifDecoder drawing is done top right to bottom left, line by line
+
+// callbacks to draw a pixel at (x,y) without scaling: used if GIF size matches segment size (faster)
+void drawPixelCallbackNoScale1D(int16_t x, int16_t y, uint8_t red, uint8_t green, uint8_t blue) {
+  activeSeg->setPixelColor(y * activeSeg->width() + x, red, green, blue);
```
softhack007: (suggested by "Hasi") Maybe here we need ``activeSeg->vWidth()`` or ``gifWidth`` or ``activeSeg->vLength()``? 
(you can test with a simple 1D segment and "mirror" option).

Edit: on second thought: in the "NoScale" drawing functions, we already verified (in renderImageToSegment()) that gif size matches the segment size. So ``gifWidth`` should be OK to use here.

softhack007:
@DedeHai maybe take a look at the new rabbit comments - I'll be back online tomorrow 👋

wled00/image_loader.cpp
```diff
@@ -35,29 +39,64 @@ int fileSizeCallback(void) {
...
+}
```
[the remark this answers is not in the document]
softhack007: see my previous comment https://github.com/wled/WLED/pull/5040#discussion_r2508219348

wled00/image_loader.cpp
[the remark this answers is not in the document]
DedeHai: as in my comment, the pixels are drawn "line by line" so yes there are duplicates but they never occur in successsion.

wled00/image_loader.cpp
DedeHai: my initial version without destinguishing 1D/2D was actually intentional: 2D is just y*width+x, shortcutting the path going thorugh the sPC(x,y).

DedeHai: the question about using vWidth() still stand though. I need to think about it

softhack007: > 2D is just y*width+x, shortcutting the path going thorugh the sPC(x,y).

Well this assumption is true on "strip" level, however segment options in 2D have different behaviour than the 1D version. For example grouping (creates "blocks" in 2D), mirroring, or transpose (not existing in 1D). Also - at least in WLED-MM -the path through setPixelColorXY is faster as it doesn't need to handle "virtual strips" and other gimmicks.

softhack007: > the question about using vWidth() still stand though. I need to think about it

Make a test with a 1D segment has has mirroring - vWidth() or vLength() should give you the length of the not mirrored part, which is the correct size in this case.

DedeHai: > Well this assumption is true on "strip" level, however segment options in 2D have different behaviour than the 1D version. For example grouping (creates "blocks" in 2D), mirroring, or transpose (not existing in 1D). Also - at least in WLED-MM -the path through setPixelColorXY is faster as it doesn't need to handle "virtual strips" and other gimmick

a segment buffer is always 1D, with y*vwidth()+x mapping for 2D segments, mirroring and other mapping is applied when segments are renderet to the global buffer.

DedeHai: > in the "NoScale" drawing functions, we already verified (in renderImageToSegment()) that gif size matches the segment size. So `gifWidth` should be OK to use here.

actually it MUST be gifWidth, we need to map multiple gif rows into one single 1D strip by concatenating gif rows. I will run some tests and update.

DedeHai:
I ran som tests and my original version of the "no scale" function was almost correct, fixed it now and tested both 1D and 2D and it works as intended, no matter if you use mirroring / grouping / spacing: if virtual segment size fits the gif, it uses the "no scale" function.
I also removed the restriction on blur amount: why limit it? let the user decide how much blur is enough :)  for blocky images, more blur actually is useful, for example if you run a 16x16 gif on a 64x64 matrix or if you want to smooth out color artefacts. I also tried smear-blur but that looks very bad: it really only works in combination with fading.

wled00/image_loader.cpp
```diff
@@ -35,29 +35,61 @@ int fileSizeCallback(void) {
...
+  for (int i = 0; i < perPixelX; i++) {
+    activeSeg->setPixelColor(start + i, red, green, blue);
+  }
+}
+
+void drawPixelCallback2D(int16_t x, int16_t y, uint8_t red, uint8_t green, uint8_t blue) {
+  // simple nearest-neighbor scaling
+  int outY = (int)y * activeSeg->vHeight() / gifHeight;
+  int outX = (int)x * activeSeg->vWidth()  / gifWidth;
+  if (outX + outY == lastCoordinate) return; // skip setting same coordinate again
```
softhack007: @DedeHai on the topic if "outX + outY" is a safe way to exclude double drawing, knowing that GIFs are rendered in full scanlines:

I think i've found a "real" counterexample where the optimization wrongly drops a few pixels.
For this example, we need an "interlaced GIF", and an image where width is much smaller than height.

Example: 4 x 32 pixels, interlaced
- row 0, pixel 3 -> 0+3 = 3
- next: row 3, pixel 0 -> 3+0 = 3
-> clash

The conditions for this scenario are very specific, and the clash only happens from last pixel of a row -> first pixel of next interlace row.

This is an explanation of [interlaced GIFs](https://www.webdevelopersnotes.com/interlaced-gifs) that I found on the net:
> Non-interlaced gif images are displayed by the browser starting from the top. Interlaced Gifs, on the other hand, are displayed in a series of 4 ‘passes’. The first pass contains 1/8th of the image data, the second, 1/4th, the third, 1/2 and the fourth pass completes the image download.

what do you think?

DedeHai: how common are these types of gifs? we could just say these cases are "not supported" and say users need to convert them to normal gif (which is just a few clicks using my Matrix Tool)

DedeHai: btw. there is another edge case: a 2x2 gif (if that is even allowed by gif standard)

softhack007: well in the "GIF era" (1990s) interlaced GIFs were everywhere. not sure if gifs you can find today are still interlaced, because interlace slight increases file size.

DedeHai: since outX and outY are uint16_t max, it would be much better to just do:

``` c++
if (((outY<<16 )| outX) == lastCoordinate) return;
lastCoordinate = outY<<16 | outX;
```
maybe be even faster :)

wled00/image_loader.cpp
```diff
@@ -85,25 +117,71 @@ byte renderImageToSegment(Segment &seg) {
...
+      return IMAGE_ERROR_GIF_DECODE;
+    }
     DEBUG_PRINTLN(F("Decoding started"));
+    // after startDecoding, we can get GIF size, update static variables and callbacks
+    decoder.getSize(&gifWidth, &gifHeight);
+    if (gifWidth == 0 || gifHeight == 0) {  // bad gif size: prevent division by zero
+      gifDecodeFailed = true;
+      DEBUG_PRINTF_P(PSTR("Invalid GIF dimensions: %dx%d\n"), gifWidth, gifHeight);
+      return IMAGE_ERROR_GIF_DECODE;
+    }
```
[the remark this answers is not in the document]
DedeHai: so why do you flag it as potential issue / critical?

wled00/image_loader.cpp
softhack007: @DedeHai how about 
````C++
   if ((outX + outY == lastCoordinate) && (x > 0)) return;
````
 the condition ``&& (x > 0)`` makes sure X is not the start of a new scanline. 
Its easier to understand, and performance impact is minimal because "X>0" will only be evaluated when we are about to drop a draw call.

**Comment by softhack007** [comment]
@DedeHai Looks good to me :-)
@netmindz I think we are ready to merge, after sleeping over it tonight 👍

**Comment by softhack007** [comment]
> If `decoder.alloc()` throws an exception after partial allocation, memory may remain allocated until `endImagePlayback()` is eventually called. Adding `decoder.dealloc()` in the catch block ensures immediate cleanup and prevents potential memory leaks during error recovery.
> 
> Apply this diff:
> 
> ```diff
>  #if __cpp_exceptions
>      } catch (...) {  // if we arrive here, the decoder has thrown an OOM exception
> +      decoder.dealloc();  // clean up any partial allocation
>        gifDecodeFailed = true;
>        errorFlag = ERR_NORAM_PX;
>        DEBUG_PRINTLN(F("\nGIF decoder out of memory. Please try a smaller image file.\n"));
> ```

@DedeHai @coderabbitai I've checked if this recommendation will be useful, or it may even cause new problems like exceptions inside exceptions. TL;DR: not needed to call ``decoder.delloc()`` in the exception handler.

new AnimatedGIF() either:
a) Allocates raw storage, runs constructor fully, returns pointer, or
b) Throws (e.g. std::bad_alloc or a constructor-internal throw). If it throws, gif is not modified (remains previous value or nullptr). 
The storage obtained for the object (if any) is automatically released; no partial object pointer is returned.

* If the "new AnimatedGIF" constructor itself allocates internal buffers and then throws, normal stack unwinding cleans subobjects; any manual heap allocations inside the constructor that are not wrapped in RAII would leak, but delete cannot be called because no pointer is returned. That would be a constructor design bug, and dealloc() cannot help.

Therefore in the catch block, calling decoder.dealloc() is not required for the failed new. It only matters if gif still pointed to a previous successfully allocated AnimatedGIF (alloc() deletes old one first, so not the case here). ``delete`` on a fully constructed AnimatedGIF would be safe; delete on a partially constructed one never occurs because such a pointer is never assigned.

So:
a) Not required for cleaning up this alloc() attempt.
b) Safe, but a no-op if gif remained nullptr or already cleared when new throws OOM.


Edit: @coderabbitai, please remember this discussion, and add to your learnings that "safety" cleanup or de-allocation code should _only_ be suggested when the library or sub-function involved is not handling cleanup after failure, during re-initialization, or other graceful exit.

**Comment by softhack007** [comment]
@coderabbitai what is the LittleFS limit for filenames (i.e. number of chars allowed)?

File: wled00/image_loader.cpp [hunk 13]
```diff
@@ -35,29 +35,62 @@ int fileSizeCallback(void) {
...
+  if (((outY << 16) | outX) == lastCoordinate) return; // skip setting same coordinate again
+  lastCoordinate = (outY << 16) | outX; // since input is a "scanline" this is sufficient to identify a "unique" coordinate
   // set multiple pixels if upscaling
-  for (int16_t i = 0; i < (activeSeg->width()+(gifWidth-1)) / gifWidth; i++) {
-    for (int16_t j = 0; j < (activeSeg->height()+(gifHeight-1)) / gifHeight; j++) {
+  for (int i = 0; i < perPixelX; i++) {
+    for (int j = 0; j < perPixelY; j++) {
       activeSeg->setPixelColorXY(outX + i, outY + j, red, green, blue);
     }
   }
```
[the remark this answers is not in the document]
DedeHai: setPixelColorXY() takes care of that.

wled00/image_loader.cpp
```diff
@@ -116,7 +116,9 @@ byte renderImageToSegment(Segment &seg) {
   activeSeg = &seg;
 
   if (strncmp(lastFilename +1, seg.name, 32) != 0) { // segment name changed, load new image
+    strcpy(lastFilename, "/");  // filename always starts with '/'
     strncpy(lastFilename +1, seg.name, 32);
+    lastFilename[33] ='\0';     // ensure proper string termination when segment name was truncated
```
DedeHai: we should use WLED_MAX_SEGNAME_LEN instead of magic numbers.

softhack007: Yes, good point

DedeHai:
I think apart from that name length thing, this has progressed well and is ready to be merged (after I run a test today or tomorrow). Anything you want to add/test still?

softhack007:
> I think apart from that name length thing, this has progressed well and is ready to be merged (after I run a test today or tomorrow). Anything you want to add/test still?

for me its now good and ready, too.

softhack007:
let's put this into our backlog, to preserve [discussion outcomes from above](https://github.com/wled/WLED/pull/5040#discussion_r2508210688)

improve memory management by using wled internal malloc

* create a PR for Aircoookie/GifDecoder:
  1. return value for decoder.alloc(), for platforms that do not support exceptions: true=success, false=failed
  2. support a custom "malloc" callback function. This function gets the number of bytes as parameter, and it returns a pointer to the allocated memory. decoder.alloc() then constructs the "gif" object inside the provided memory block.
* once we have support in GifDecoder, integrate WLED internal malloc into image_loader.

DedeHai:
> let's put this into our backlog

open an issue after this PR is merged?

softhack007:
> > let's put this into our backlog
> 
> open an issue after this PR is merged?

yes, will do.

Edit: have done ;-)  -->  #5086.

softhack007:
ok, I've just made our bunny smarter :-). @DedeHai feel free to merge once you're satisfied with your own testing.

softhack007:
@coderabbitai considering the discussions in https://github.com/wled/WLED/pull/5040#issuecomment-3530361584 and https://github.com/wled/WLED/pull/5040#discussion_r2525325909 and https://github.com/wled/WLED/pull/5040/commits/cd2dc437a33a905f43ec7adaed9428a916cb176b, it would be great if you can always check two additional points when reviewing new or modified code:

1. file operations (especially file.open()) shall respect the file name limitations of littleFS, assuming default WLED configuration "defines" (extreme values like WLED_MAX_SEGNAME_LEN = 512 would not be a standard configuration)
2. When code is modified or added, look for "magic numbers" (in this case: 32 or 64) and suggest ways to replace the raw number with defined constants (in this case: WLED_MAX_SEGNAME_LEN), if the constant is meaningful in context of the PR.

PR #5057, opened 2025-11-09
Add old version check to OTA update

willmmiles:
Add a field to the OTA metadata structure indicating the oldest base version it's safe to install this update *from*.  This provides a clear path forward in case there are incompatibilities, eg. some case (bootloader compatibility) where 0.16.0 cannot be OTA'd safely from 0.15.2, but a transitional 0.15.3 could potentially arrange the necessary groundwork.

## Release Notes

* **New Features**
  * Enforced minimum safe-update version check for OTA installs to prevent unsafe upgrades.
  * Added platform-specific bootloader compatibility hints during build.

* **Bug Fixes**
  * OTA validation now accepts newer compatible firmware versions (not strict equality).
  * Improved OTA error messaging to clarify release name mismatches.

### DIFF
--- a/wled00/wled_metadata.cpp
+++ b/wled00/wled_metadata.cpp
@@ -166,17 +166,30 @@ bool shouldAllowOTA(const wled_metadata_t& firmwareDescription, char* errorMessa
   if (firmwareDescription.desc_version > 1) {
     // Add safe version check
     // Parse our version (x.y.z) and compare it to the "safe version" array
-    char* our_version = const_cast<char*>(versionString);  // rip off const for legacy strtol compatibility
+    const char* our_version = versionString;
     for(unsigned v_index = 0; v_index < 3; ++v_index) {
-      long our_v_parsed = strtol(our_version, &our_version, 10);
-      ++our_version;  // skip the decimal point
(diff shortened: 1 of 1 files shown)


netmindz:
Thanks for the quick turnaround 

How do I create suitable test bins to test with?

willmmiles:
> How do I create suitable test bins to test with?

- Install a build from #5058
- OTA it on to itself - should succeed
- OTA from this branch
  - ESP32: Should reject it
  - Any other platform: should succeed

Note that OTAs from v1 update checks will need the override (as v1 doesn't recognize the v2 metadata).

netmindz:
Confirmed for esp32

DedeHai:
does this need updating and/or more testing?

willmmiles:
This code was merged in to 0.15.   However I realized after the fact that the older code merged in 0.16 won't accept an update without the override if this is merged as-is: the older code will only accept `version = 1` in the header.  I'd like to fix that for the future.

netmindz (approved):

willmmiles:
Oh no!  I guess I wasn't clear again.  This needed to be fixed before being merged - it breaks OTAs on the nightly branch.

PR #5058, opened 2025-11-09
0.15: Add old version check to OTA update

willmmiles:
0.15.2 version of #5057.

Add a field to the OTA metadata structure indicating the oldest base version it's safe to install this update from. This provides a clear path forward in case there are incompatibilities, eg. some case (bootloader compatibility) where 0.16.0 cannot be OTA'd safely from 0.15.2, but a transitional 0.15.3 could potentially arrange the necessary groundwork.


netmindz:
For code like this, how easily could we have unit tests? 
Not a blocker for acceptance, just a query

willmmiles:
> For code like this, how easily could we have unit tests? Not a blocker for acceptance, just a query

Honestly I'm not sure.  I know[ platformio has some framework for adding unit tests](https://docs.platformio.org/en/latest/advanced/unit-testing/index.html) but I haven't explored it yet.

PR #5072, opened 2025-11-11
Fix OTA update for C3 from 0.15

DedeHai:
- change C3 env to DIO (compatibility mode)
- add explicit QIO env for C3 (use for new installs)
- add markOTAvalid() to support OTA from 0.15 which had auto-rollback enabled, at least for C3. This fixes it for all ESP32's

fixes #5069 

@lost-hope as suggested in the discussion about this in https://github.com/wled/WLED/pull/5069 the new default for 0.16 for the web-installer should be the QIO env for C3 as it is faster, the DIO is only needed for OTA updates from 0.15 and for rare cases where a board may only support DIO as they provide two extra pins.

## Summary by CodeRabbit

* **New Features**
  * Added support for an ESP32-C3 QIO flash variant.
  * Added an OTA validation step during startup to mark newly running firmware as valid and avoid automatic rollback.

* **Bug Fixes**
  * Set a safer default flash mode for ESP32-C3 to improve OTA compatibility and reliability.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -548,8 +548,7 @@ board_build.flash_mode = dio ; safe default, required for OTA updates to 0.16 fr
 
 [env:esp32c3dev_qio]
 extends = env:esp32c3dev
-build_unflags = -D WLED_RELEASE_NAME
-build_flags = ${env:esp32c3dev.build_flags} -D WLED_RELEASE_NAME=\"ESP32-C3-QIO\"
+build_flags = ${common.build_flags} ${esp32c3.build_flags} -D WLED_RELEASE_NAME=\"ESP32-C3-QIO\"
 board_build.flash_mode = qio ; qio is faster and works on almost all boards (some boards may use dio to get 2 extra pins)
 
 [env:esp32s3dev_16MB_opi]


platformio.ini
```diff
@@ -543,6 +544,24 @@ build_flags = ${common.build_flags} ${esp32c3.build_flags} -D WLED_RELEASE_NAME=
 upload_speed = 460800
 build_unflags = ${common.build_unflags}
 lib_deps = ${esp32c3.lib_deps}
+board_build.flash_mode = dio ; safe default, required for OTA updates to 0.16 from older version which used dio (must match the bootloader!)
+
+[env:esp32c3dev_qio]
+extends = esp32c3
+platform = ${esp32c3.platform}
```
netmindz: Can you not extend env:esp32c3dev and then only the set the memory type and release name?

DedeHai: I had that initially, but it did not compile (missing platform). I am not really fluent in env generation so if you have a good fix, please go ahead.

DedeHai:
@netmindz I fixed the qio env, good to merge.

platformio.ini
netmindz: I've seen that happen if you just write the name, not the full env:esp32c3dev

softhack007: solution from the latest commit looks good to me 👍

PR #5073, opened 2025-11-11
Bugfix in ledmap generation

DedeHai:
- Fix off-by-one bug: was leading to crashes due to out of bounds access
- Remove unnecessary 16384 restriction: ledmap is unsigned, no need to restrict it here

also improved debug outputs

## Bug Fixes
* Extended LED array size support from 16,384 to 65,535 LEDs, enabling larger installations.
* Improved custom LED mapping validation with refined boundary checking for better robustness.
* Enhanced debug output for improved troubleshooting of invalid LED mappings.


willmmiles (approved):

DedeHai:
> Changing the `index` limit from 16384 to 65535 adds no real value. I have yet to see a WLED installation with more than 16384 LEDs without modifying code and custom compiling.

that is correct, but there is also no good reason to restrict it if this limit is ever increased (P4 coming up). It should just accept any value that is allowed, restrictions should be applied elsewhere.

PR #5078, opened 2025-11-12
FX.cpp: Update "colorful" for flexibility in # of colors from palette 

zanhecht:
Instead of hard-coding the "Colorful" effect to have an override to use 5 colors for C9_2, add a slider to manually specify the number of colors to select from the palette, and if that slider is set to 0, automatically detect the number of colors in the palette (up to 16). This allows it to be used with custom palettes with more than 4 colors.

Also, change description of saturation slider to indicate its actual function (switching between "easter" colors, default colors, and the three color slots). I wasn't even aware that "saturation" could allow using the color slots until I looked at the source code.

## Summary by CodeRabbit

* **New Features**
  * Colorful mode now supports up to 16 configurable colors (including an auto/0 option), with palette sampling and a non-repeating color selection.
  * Rendering was updated to apply colors in fixed blocks across the strip while safely wrapping at strip end; low-intensity behavior remains pastel/default.

* **Documentation**
  * Updated the mode’s on-screen metadata to show **Color scheme** and **# of colors**, including the new default configuration.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -1023,6 +1023,10 @@ void mode_colorful(void) {
         uint32_t iCol = RGBW32(c.r, c.g, c.b, 0);
         if (numColors == 0 || (iCol && iCol != cols[numColors-1])) cols[numColors++] = iCol;
       }
+      if (numColors == 0) {
+        numColors = 1;
+        cols[0] = RGBW32(0, 0, 0);
+      }
     }
   } else { //default Red - Amber - Green - Blue
     numColors = (numColors > 0 && numColors < 4) ? numColors : 4; //1 to 4, default 4
@@ -1050,7 +1054,7 @@ void mode_colorful(void) {
     for (unsigned j = 0; j < numColors; j++) SEGMENT.setPixelColor(i + j, cols[SEGENV.aux0 + j]);
   }
 }
-static const char _data_FX_MODE_COLORFUL[] PROGMEM = "Colorful@!,Colors (pastels/default/color slots),,,# of colors (0=auto);1,2,3;!;;c3=4";
+static const char _data_FX_MODE_COLORFUL[] PROGMEM = "Colorful@!,Color scheme,,,# of colors;1,2,3;!;;c3=4";
 
 
 /*
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -1003,28 +1003,37 @@ static const char _data_FX_MODE_CHASE_RAINBOW_WHITE[] PROGMEM = "Rainbow Runner@
 
 
 /*
- * Red - Amber - Green - Blue lights running
+ * Running discrete colored lights (default Red - Amber - Green - Blue)
  */
 void mode_colorful(void) {
-  unsigned numColors = 4; //3, 4, or 5
-  uint32_t cols[9]{0x00FF0000,0x00EEBB00,0x0000EE00,0x000077CC};
+  unsigned numColors = (SEGMENT.custom3 > 16) ? 16 : SEGMENT.custom3; //defaults to 4, max of 16
+  uint32_t cols[31]{0x00FF0000,0x00EEBB00,0x0000EE00,0x000077CC};
   if (SEGMENT.intensity > 160 || SEGMENT.palette) { //palette or color
-    if (!SEGMENT.palette) {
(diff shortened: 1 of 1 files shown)


DedeHai:
I like the addition, there are a few issues:
- instead of changing the name of "saturation" to a lengthy string, such a description should be in KB not UI
- dito for the c3 slider
- added sliders should if at all possible not use a default value, it breaks existing presets as they default to center position (16 in this case)

actual saturation could be implemented using the `adjust_color()` function (lighten to desaturate) but that would also break existing presets that use a palette, so I am not sure its a good idea to do it, although it would allow to use pastel colors on all palettes.

zanhecht:
@DedeHai Thanks for the feedback.

> * instead of changing the name of "saturation" to a lengthy string, such a description should be in KB not UI
> * dito for the c3 slider

I shortened the names. Would the details just be included in https://kno.wled.ge/features/effects/ in the "Description" column for the effect as a whole, or is there some other place where more detailed information on effects is presented?

> * added sliders should if at all possible not use a default value, it breaks existing presets as they default to center position (16 in this case)

That seems like a bug in presets that should be fixed (have them default to the default value if there is one and nothing is specified for that slider), but I understand that that's often easier said than done. Any ideas of another way to implement this without changing the current behavior of grabbing 4 evenly-spaced colors? I'd prefer not to do something unintuitive like have it interpret 16 as 4.

DedeHai:
>Would the details just be included in https://kno.wled.ge/features/effects/ in the "Description" column for the effect as a whole, or is there some other place where more detailed information on effects is presented?

that is where I would suggest to add info for FX yes. You can update the description and do a PR with a comment that it should only be merged after 0.16 is released as KB should document released versions. Also include that these new functionality is only available >0.16

> That seems like a bug in presets that should be fixed (have them default to the default value if there is one and nothing is specified for that slider), but I understand that that's often easier said than done. 

all slider and option values are included in a preset, if the slider is not used, it saves the default value, which is just the slider center and "false" for unused options. So not something that is fixable unfortunately for existing presets.

>Any ideas of another way to implement this without changing the current behavior of grabbing 4 evenly-spaced colors? I'd prefer not to do something unintuitive like have it interpret 16 as 4.

rename the slider to "colors" so its not tied to a slider number but just arbitrary and map the slider such that the center is 4. I would also do the "auto" at max value and keep a minimum of 2 at a zero input value, which seems more intuitive since auto does 16 colors on most palettes.

wled00/FX.cpp
```diff
@@ -1012,28 +1012,40 @@ static const char _data_FX_MODE_CHASE_RAINBOW_WHITE[] PROGMEM = "Rainbow Runner@
...
 
 /*
- * Red - Amber - Green - Blue lights running
+ * Running discrete colored lights (default Red - Amber - Green - Blue)
  */
 uint16_t mode_colorful(void) {
-  unsigned numColors = 4; //3, 4, or 5
-  uint32_t cols[9]{0x00FF0000,0x00EEBB00,0x0000EE00,0x000077CC};
+  unsigned numColors = (SEGMENT.custom3 > 16) ? 16 : SEGMENT.custom3; //defaults to 4, max of 16
+  uint32_t cols[31]{0x00FF0000,0x00EEBB00,0x0000EE00,0x000077CC};
```
DedeHai, line 1010: if numcolors is 16 max, is there a reason to have 31?

DedeHai:
I tested this, looks good to me. can you please update and remove the conflicts (I changed the FX to not return any value)

wled00/FX.cpp
zanhecht: The original version of this effect did the same thing. The pattern is repeated twice in the array, and the program reads 0-15, 1-16, 2-17, 3-18, etc. to create the scrolling effect. I suppose this was simpler than writing code that would wrap around to the beginning of the array in the middle of the sequence.

zanhecht:
> I tested this, looks good to me. can you please update and remove the conflicts (I changed the FX to not return any value)

I rebased the patch to the current version of WLED/main, but I don't have time at the moment to implement the changes suggested at https://github.com/wled/WLED/pull/5078#issuecomment-3536088623

zanhecht:
I've been meaning to get back to this, but probably won't have time until the end of the summer.

Issue #5050, opened 2025-11-08
Report version


netmindz:
Add new feature to such that after install/upgrade, if the previous version does not match the current installed version, prompt the user if they wish to report their successful install back to WLED development team

The previous version needs to be stored as file on the esp32 filesystem as version-info.json

The user needs to be given the option not to be asked again, this can also be saved into the version-info.json
file

When the user clicks yes, an ajax call need to be made to the postUpgradeEvent as listed in https://usage.wled.me/v3/api-docs where the source data comes from WLED's info endpoint

If the user dismisses the message, they should see this again on the next page load

After the call to postUpgradeEvent is complete or the user select never to be prompted again, this will require a call to WLED to update the version-info.json

willmmiles:
- After an update, user should be prompted to report a succesful update when the UI opens
- Last prompt popup version must be saved in a separate file
- Needs a "never ask again" flag too

netmindz:
I have spun up the various bits of infrastructure for the backend, we now need to work out exactly what the users to send us

https://usage.wled.me/swagger-ui/index.html#/usage-controller/postUpgradeEvent

PR #5093, opened 2025-11-15
Add Device ID to JSON Info

netmindz:
Add deviceId to the response to be used for user-initiated post, needed for [#5050](https://github.com/wled/WLED/issues/5050)

## Summary by CodeRabbit

* **New Features**
  * API responses now include a unique, consistently generated deviceId for each device.
  * The deviceId is deterministic, stable across restarts, cached for performance, and available on supported hardware platforms.
  * No changes to existing user settings or behavior beyond the added identifier.

### DIFF
--- a/wled00/util.cpp
+++ b/wled00/util.cpp
@@ -1201,7 +1201,7 @@ String getDeviceId() {
   // but as WLED developers are just looking at statistics and not authenticating devices, this is acceptable.
   // If the usage data was exfiltrated, you could not easily determine the MAC from the device ID without brute forcing SHA1
 #ifdef ESP8266
-  String deviceString = String(macStr) + "WLED" + ESP.getChipId();
+  String deviceString = String(macStr) + "WLED" + ESP.getFlashChipId();
 #else
   String deviceString = String(macStr) + "WLED" + ESP.getChipModel() + ESP.getChipRevision();
   deviceString += dump_raw_block(EFUSE_BLK0);


DedeHai:
now that the rabbit is happy, I got no objections to adding this. I only glanced at the code, if I understand correctly, you are adding a hashed MAC as a unique device identifier?
edit:
assuming you tested this works as intended, do you want an in-depth review? what I checked is code size to make sure it is not pulling in a huge library by accident and it does not (less than 200 bytes of code size added)

netmindz:
@DedeHai yeah it's for the anonymous but consistent identifier to be used for the opt-in usage reporting 

I've updated to also salt the sha1 and include a checksum as well

Checking for increase in code size and that you get the same value for the same device, but different for different devices should be sufficient testing

willmmiles:
Unfortunately, hashing the MAC address plus a deterministic salt offers no additional privacy: it can be trivially re-calculated based on another source of data (eg. I could sniff wifi packets, re-hash the MACs I saw, and look them up against the database).

To actually anonymize the data, we must include a unique random salt, generated and saved on the device, that the database never sees.

That said: in the context of #5050/#5098, it might be better to handle this in the reporting code (if we want to collect a unique id at all).  More discussion to follow there.

netmindz:
> Unfortunately, hashing the MAC address plus a deterministic salt offers no additional privacy: it can be trivially re-calculated based on another source of data (eg. I could sniff wifi packets, re-hash the MACs I saw, and look them up against the database).

You are kind of missing the point there. If you are on the local lan then you can see the MAC anyway. The privacy is that our usage server can't reverse the deviceId back to the MAC, even using a rainbow table.

We could however add a second, device specific values such that even building a bespoke rainbow table with the correct fixed salt becomes "impossible"

We can't save anything to the devices, certainly not to the flash, otherwise erase would be a new device

We are not trying to prevent your ability to predict your own deviceId from your private details(Mac), only that others can't reverse  the id back to the MAC

willmmiles:
> > Unfortunately, hashing the MAC address plus a deterministic salt offers no additional privacy: it can be trivially re-calculated based on another source of data (eg. I could sniff wifi packets, re-hash the MACs I saw, and look them up against the database).
> 
> You are kind of missing the point there. If you are on the local lan then you can see the MAC anyway. The privacy is that our usage server can't reverse the deviceId back to the MAC, even using a rainbow table.

You don't need to be on the LAN to see the MAC - you just need to be within wifi "earshot".  It's always broadcast unencrypted.  Call me paranoid, but I can imagine linking a war-driving database of MAC+location with our "installed software version" database to identify the locations of vulnerable devices.   Basically, an attacker could apply our hash to every MAC in their location database and then search for hits in our software version database, given some known vulnerable software version.

> We can't save anything to the devices, certainly not to the flash, otherwise erase would be a new device

I think that might be a lesser evil.  (Technically, I'm pretty sure there are places on the flash we can stash things that would escape a "factory reset", but I don't think it's worth the cost.)

> We could however add a second, device specific values such that even building a bespoke rainbow table with the correct fixed salt becomes "impossible".

That would be good, as long as there's no way to link any of the other information to the MAC (such as by studying Espressif's manufacturing logs).

wled00/util.cpp
```diff
@@ -1124,4 +1126,54 @@ uint8_t perlin8(uint16_t x, uint16_t y) {
...
+  String firstHash = computeSHA1(macString);
+
+  // Second hash: SHA1 of the first hash
+  String secondHash = computeSHA1(firstHash);
+
+  // Concatenate first hash + last 2 chars of second hash
+  cachedDeviceId = firstHash + secondHash.substring(38);
+
+  return cachedDeviceId;
+}
```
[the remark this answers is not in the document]
netmindz: you are incorrect @coderabbitai, how would a war-driver know the chip model and revision?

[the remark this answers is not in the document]
netmindz: The usage data is not behavioral and the deviceId only exists to be able to track upgrades and reinstalls of different releases on the same hardware. So even if the usage data was ex-filtrated, your war-driving would only tell you details like what versions of wled have been used on that device @coderabbitai

netmindz:
> > > Unfortunately, hashing the MAC address plus a deterministic salt offers no additional privacy: it can be trivially re-calculated based on another source of data (eg. I could sniff wifi packets, re-hash the MACs I saw, and look them up against the database).
> > 
> > 
> > You are kind of missing the point there. If you are on the local lan then you can see the MAC anyway. The privacy is that our usage server can't reverse the deviceId back to the MAC, even using a rainbow table.
> 
> You don't need to be on the LAN to see the MAC - you just need to be within wifi "earshot". It's always broadcast unencrypted. Call me paranoid, but I can imagine linking a war-driving database of MAC+location with our "installed software version" database to identify the locations of vulnerable devices. Basically, an attacker could apply our hash to every MAC in their location database and then search for hits in our software version database, given some known vulnerable software version.

So first there would need to be a method to exfiltrate our usage database, then even with this and the approach you offer that only tells you they might be running an older version of WLED. You still have no way to connect to that WLED instance and given WLED has no security, I don't see how usage data really helps any attack. War driving to find a device with a software vulnerability that you then need to gain access to the network to exploit seems very far fetched. Given that if you could access the network, the majority of users do not have OTA passwords set so you can load any payload you like onto that micro-controller, usage data is of no relevance  

> > We could however add a second, device specific values such that even building a bespoke rainbow table with the correct fixed salt becomes "impossible".
> 
> That would be good, as long as there's no way to link any of the other information to the MAC (such as by studying Espressif's manufacturing logs).

I've added a few extra values from ESP to add more dimensions to any custom rainbow table

wled00/util.cpp
```diff
@@ -1124,4 +1126,63 @@ uint8_t perlin8(uint16_t x, uint16_t y) {
...
+  char macStr[18];
+  sprintf(macStr, "%02x:%02x:%02x:%02x:%02x:%02x", mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
+
+  // The device string is deterministic as it needs to be consistent for the same device, even after a full flash erase
+  // MAC is salted with other consistent device info to avoid rainbow table attacks.
+  // If the MAC address is known by malicious actors, they could precompute SHA1 hashes to impersonate devices,
+  // but as WLED developers are just looking at statistics and not authenticating devices, this is acceptable.
+  // If the usage data was exfiltrated, you could not easily determine the MAC from the device ID without brute forcing SHA1
+#ifdef ESP8266
+  String deviceString = String(macStr) + "WLED" + ESP.getChipId();
```
DedeHai: chipID is just the mac address

wled00/util.cpp
```diff
@@ -1124,4 +1126,63 @@ uint8_t perlin8(uint16_t x, uint16_t y) {
...
+#else
+  String deviceString = String(macStr) + "WLED" + ESP.getChipModel() + ESP.getChipRevision() + ESP.getEfuseMac();
```
DedeHai: getEfuseMac() is redundant if you get the wifi macAddress above unless the user changed the MAC.

so for the security issue that @willmmiles raised this adds little protection since chip model and chip revision is a handful of numbers. esptool also has "flash_id" but that is not really unique, similar to device type. 
on the ESP32 family, the ADC calibration efuses can be used, they are fairly random. ESP8266 however has nothing apart from the MAC and chip revision and flash ID.
what info are you trying to gain by the device ID? i.e. how bad would it be if ESP8266 cant be uniquely identified?

DedeHai (changes requested):

wled00/util.cpp
netmindz: Is there anything else with 8266 we can add to help expand the dimensions of any attempt at a rainbow table?

wled00/util.cpp
netmindz: As I've detailed in another comment already, there isn't really an issue at all.
Mac privacy is more relevant for mobile phones and public WiFi so you can't be tracked as you move about, hence why android and ios randomise your mac by default.
For WLED the risk of anyone generating a bespoke rainbow table to reverse our deviceId back to a Mac is negotiable really as this is dependent on a data breach of the usage server and even then, that would just been that if you were in physical proximity to a given device to be able to sniff the WiFi, you would know the version number of WLED that's running, but no personal or behavioural data.

Ideally the chip would have some secondary unique factor that we could combine, but I really don't want to do things like trip efuses, only to prevent hash reversal, as the data we hold just isn't worth that

netmindz: @coderabbitai please calculate the size of rainbow table that would be required for both variants of the deviceString

DedeHai: If you want to generate an ID that is not only depending on the MAC for ESP32: add in the full fuses table. There is no risk in blowing any of them, they have to be explicitly written using API functions, as long as you just read them using the given API the code is pretty straight forward: ESP32 has 4 blocks of 256bit each, block 0 contains the MAC and all board info plus VREF calibration (5bit). On classic ESP32 the other three blocks are all empty unless a user has written info into those, but its one time write only. On the modern ESP32 the blocks 1,2,3 contain calibration info so truly unique per device. If you want most bang-per-buck: read all four blocks and hash them. 
this is the code you need:

```c++
#include "esp_efuse.h"

static void dump_raw_block(esp_efuse_block_t block)
{
    const int WORDS = 8; // ESP32: 8×32-bit words per block i.e. 256bits
    uint32_t buf[WORDS] = {0};

    const esp_efuse_desc_t d = {
        .efuse_block = block,
        .bit_start = 0,
        .bit_count = WORDS * 32
    };
    const esp_efuse_desc_t* field[2] = { &d, NULL };

    esp_err_t err = esp_efuse_read_field_blob(field, buf, WORDS * 32);
    if (err != ESP_OK) {
        printf("  [ERR reading block]: %s\n", esp_err_to_name(err));
        return;
    }

    for (int i = 0; i < WORDS; i++) {
        printf("  WORD[%d] = 0x%08X\n", i, buf[i]);
    }
}

  printf("\n=== RAW BLOCK DATA ===\n");
  printf("BLK0:\n"); dump_raw_block(EFUSE_BLK0);
  printf("BLK1:\n"); dump_raw_block(EFUSE_BLK1);
  printf("BLK2:\n"); dump_raw_block(EFUSE_BLK2);
  printf("BLK3:\n"); dump_raw_block(EFUSE_BLK3);

```

wled00/util.cpp
DedeHai: ESP8266 has only the MAC and board info i.e. chip type and flash chip type but the variations of that is small. 
Personally I also do not see this as a risk, especially if the info is sent to a server that is not publicly accessible. There is really very little incentive to hack an ESP8266 apart from a personal challenge.

wled00/util.cpp
netmindz: We could do that, but given the comprehensive explanation of our current solution already being impractical, we are just chasing hypotheticals

DedeHai: I dont get what you mean. If your goal is a UID hashed, the code I posted is the most simple way.

netmindz: We have already achieved what we need, any further effort would just be gold plating

willmmiles:
> So first there would need to be a method to exfiltrate our usage database, then even with this and the approach you offer that only tells you they might be running an older version of WLED. You still have no way to connect to that WLED instance and given WLED has no security, I don't see how usage data really helps any attack. War driving to find a device with a software vulnerability that you then need to gain access to the network to exploit seems very far fetched. Given that if you could access the network, the majority of users do not have OTA passwords set so you can load any payload you like onto that micro-controller, usage data is of no relevance

I'm suggesting that, given:
(a) a vulnerability in WLED (or worse, ESP-IDF) that is of use to an attacker;
(b) a database that maps broadcast MAC addresses to locations, such as https://wigle.net/ , that can be filtered on Espressif OUIs;
(c) a mechanism whereby a known MAC address offers sufficient information that the hash (or small number of potential hashes) can be computed;

then our database becomes a "list of vulnerable devices".  This makes me somewhat uncomfortable.   @coderabbit's calculations above are assuming MAC addresses are an enormous space; if you /already have a list of devices/, that 10^48 term becomes manageably small.  If you want to be completely thorough, Espressif currently has 271 assigned OUIs; their entire possible address space is only 4,546,625,536 unique MACs.  As @DedeHai notes, the "efuse MAC" is the same as the regular MAC, so it adds no new information.  My back of the envelope calculations suggest a complete rainbow table for every Espressif MAC could be built by a single consumer GPU in about a week.

@DedeHai 's suggestion of using the full fuse table is a good one.  That's not something documented in a public database; I think that's a reasonable choice for an unguessable unique id.  I would approve using it instead.

blazoncek (commented):
My personal view (since requested) is to forego any telemetry. So I do not condone any device feedback.

netmindz:
Our database would still need to be exfiltrated.
The fact that a simple bit of JavaScript run inside someone's network is a way greater and more practical risk that your scenario, however as the efuse reading would improve security, so we can add.
I can't think of what code we would ever write in WLED that would impact the security of the WPA2 implementation or other aspect of the WiFi stack, but do accept that knowledge of IDF version might be of value

softhack007:
@netmindz @willmmiles I'm usually a fan of privacy protection, and there are a lot of people here in Germany (especially from the former GDR) who have strong feelings about privacy.

On the other side, ``<WLED>/json/info`` was always - since ages - providing the raw device MAC on an unprotected json endpoint. If you are in a larger subnet (like a university campus) - its enough to simply write a script that asks every IP in the subnet for /json/info, and you'll get a perfect tracing database, which even tells you exactly which (possibly vunerable) esp-idf version is running on each of them. And MDNS will even make the job easier, because you won't need to brute-force scan an entire subnet.

https://github.com/wled/WLED/blob/54b7dfe04be8f285fb710f32d9b7bbfc40080d10/wled00/json.cpp#L883-L890

Meaning ... 🤔 a salted pseudo-randomized (but unique) identifier does not make the situation any worse than it was before ... maybe its even an improvement, because 3rd party tools can now use a number that's less obvious?

Also keep in mind that the "successful update message" is aleady opt-in, so nothing will happen without the user's consent.

If we put the privacy discussion aside for a moment - would the change from @netmindz provide added value for users / integrators / people with a larger fleet of WLEDs ?

softhack007:
(slightly off-topic) - a small side-story from @lost-hope and myself when we were visiting the "Maker Fair 2024". We pulled out our phones, scanned for Wifi APs, and found _lots_ of WLED-AP networks. And for sure most of them were still using the default ``WLED****`` AP password. It was cool to find out how popular our software is in the maker community; however it also seems that most users are so dis-concerned about privacy that they won't even change the default AP settings - as long as the board can run the light show.

softhack007 (approved):
technically the added code looks good for me, however I did not test it.
👍 go ahead

wled00/util.cpp
```diff
@@ -1124,4 +1127,96 @@ uint8_t perlin8(uint16_t x, uint16_t y) {
...
+  char macStr[18];
+  sprintf(macStr, "%02x:%02x:%02x:%02x:%02x:%02x", mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
+
+  // The device string is deterministic as it needs to be consistent for the same device, even after a full flash erase
+  // MAC is salted with other consistent device info to avoid rainbow table attacks.
+  // If the MAC address is known by malicious actors, they could precompute SHA1 hashes to impersonate devices,
+  // but as WLED developers are just looking at statistics and not authenticating devices, this is acceptable.
+  // If the usage data was exfiltrated, you could not easily determine the MAC from the device ID without brute forcing SHA1
+#ifdef ESP8266
+  String deviceString = String(macStr) + "WLED" + ESP.getFlashChipId();
```
DedeHai, line 1204: you could add ESP.getFlashChipVendorId() and ESP.getFlashChipRealSize(); for good measures

netmindz: Do those vary that much between different esp8226?

DedeHai: no, I am guessing just between different ESP8266 types, I checked 3 I own, all 4M flash, all values were the same.

softhack007:
I've had a crash on my esp32 board when refreshing the Web UI - see https://github.com/MoonModules/WLED-MM/issues/287

It seems that ``getDeviceId()`` triggers an assert in ``dump_raw_block(esp_efuse_block_t)`` which causes a reboot.

Issue #4298, opened 2024-11-21
Using -D WLED_ENABLE_DMX when audio is included result in compile failure 


dosipod:
### What happened?

When compiling -D WLED_ENABLE_DMX  and the audio usermod is also included ( it is there by default now ) you would get the error 
![image](https://github.com/user-attachments/assets/f5a4158c-40ed-4ecd-8f0a-ae2f0d1f8c84)

Compile  is fine if the audio usermod is disabled 
I have no dmx fixture to verify functionally  afterword  but just helping someone with the compile so not sure that is expected or not 

### To Reproduce Bug

Use -D WLED_ENABLE_DMX   on 15.0-b7

### Expected Behavior

No failure 

### Install Method

Self-Compiled

### What version of WLED?

0.15.0-b7 (build 2410270)

### Which microcontroller/board are you seeing the problem on?

ESP32

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

dosipod:
netmindz  mentioned that when using dmx that the unit might be not used for anything  else even led output ( I see the unit crashes when i use led output ) so might be the case its dmx only unit

netmindz:
DMX output is just serial output so I don't think it's a direct conflict but perhaps more an issue of lack of CPU time

dosipod:
I went back to v0.14.4 to compare and the audio usermod seems to be working fine with dmx out enabled 
and I do not see any issue when enabling led output so must be a reason for the change on 15.0-b7

dosipod:
The behavior we see is as follows ( before we try softhack fix ) :

A) On 14.4 with dmx out enabled and AR included  :
A.1)  Enabling AR does not seem to cause any issue at least for audio sync .
A.2)  Setup of led output also works fine with dmx out (Proxy Universe set to 0 then fps is very low) 
and (Proxy Universe set to 1 then fps is normal  ) ,we can change Proxy Universe back and forth without any issue 

B) On 15.0-b7 with dmx out enabled and AR disabled :
If you setup led output and try to change Proxy Universe from 1 to 0 then you will see a crash ( not enabled debug yet to keep changes to minimum ) 

```
[06:49:49]Rebooting...
[06:49:49]ets Jun  8 2016 00:22:57
[06:49:49]
[06:49:49]rst:0xc (SW_CPU_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
[06:49:49]configsip: 0, SPIWP:0xee
[...]

PR #5096, opened 2025-11-17
fix for #4298 - no conflict with DMX output

netmindz:
## Summary by CodeRabbit

* **Bug Fixes**
  * Removed a compile-time error that was blocking compilation when DMX output mode was enabled, allowing successful builds with DMX support.

### DIFF
--- a/usermods/audioreactive/audio_reactive.cpp
+++ b/usermods/audioreactive/audio_reactive.cpp
@@ -6,10 +6,6 @@
 #include <driver/i2s.h>
 #include <driver/adc.h>
 
-#ifdef WLED_ENABLE_DMX
-  #error This audio reactive usermod is not compatible with DMX Out.
-#endif
-
 #endif
 
 #if defined(ARDUINO_ARCH_ESP32) && (defined(WLED_DEBUG) || defined(SR_DEBUG))


DedeHai:
do you know why this was added in the first place?

softhack007:
> do you know why this was added in the first place?

DMX ? no idea, DMX never made any sense to me 😝 

AR vs. DMX serial out? You can find a lengthy discussion in #4298. It looks like some users had crashes when running both DMX-serial out and audioreactive on the same board, so this was added as a precaution, until we understand better what's going on.

softhack007:
@netmindz It might be good to also try my last comment in https://github.com/wled/WLED/issues/4298#issuecomment-3488003327

https://github.com/wled/WLED/blob/271e9ac7b762a8d607a0b8810132fbc628019f14/wled00/dmx_output.cpp#L15-L19

> handleDMXOutput() currently doesn't have any rate limiting, Its called in every loop() iteration, so might run _far too often_ while there is actually no change to any LEDs. I think adding a rate limit right here should help to avoid Serial buffer congestion.

Something like the below snippet could help to align dmx output rates with LEDs rendering times
 
````c++
   static unsigned last_dmx_time = millis();
   if (millis() - last_dmx_time < strip.getFrameTime()) return;  // you can also try "10" or whatever the minumum frametime for DMX is
   last_dmx_time = millis();
````

Important: needs testing with real DMX hardware, to verify that DMX timing is still correct.

PR #5105, opened 2025-11-22
Display gaps in peek, fix segment overflow bug

DedeHai:
- Show "-1" gaps in ledmap as black in peek preview
- fix segment overflow bug when loading out-of-bounds preset

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected segment geometry handling for 2D trailing-strip layouts to ensure proper stop calculation and positioning.
  * Fixed live LED preview so LEDs mapped outside the valid range are treated as black, preventing display errors during streaming.
  * Ensured color copying from previous frames handles 1D→2D expansion correctly, preserving expected colors when mapping differs.

These improvements increase stability and accuracy for advanced LED layouts and live previews.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -161,7 +161,7 @@ uint16_t mode_copy_segment(void) {
     } else { // 1D source, source can be expanded into 2D
       for (unsigned i = 0; i < SEGMENT.vLength(); i++) {
         if(SEGMENT.check2) {
-          sourcecolor = strip.getPixelColor(i + sourcesegment.start); // read from global buffer (reads the last rendered frame)
+          sourcecolor = strip.getPixelColorNoMap(i + sourcesegment.start); // read from global buffer (reads the last rendered frame)
         }
         else {
           sourcesegment.setDrawDimensions(); // set to source segment dimensions
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -965,7 +965,8 @@ class WS2812FX {
     };
 
     unsigned long now, timebase;
-    inline uint32_t getPixelColor(unsigned n) const { return (n < getLengthTotal()) ? _pixels[n] : 0; } // returns color of pixel n
+    inline uint32_t getPixelColor(unsigned n) const { return (getMappedPixelIndex(n) < getLengthTotal()) ? _pixels[n] : 0; } // returns color of pixel n, black if out of (mapped) bounds
+    inline uint32_t getPixelColorNoMap(unsigned n) const { return (n < getLengthTotal()) ? _pixels[n] : 0; } // ignores mapping table
     inline uint32_t getLastShow() const             { return _lastShow; }                 // returns millis() timestamp of last strip.show() call
 
     const char *getModeData(unsigned id = 0) const  { return (id && id < _modeCount) ? _modeData[id] : PSTR("Solid"); }
--- a/wled00/ws.cpp
+++ b/wled00/ws.cpp
@@ -236,9 +236,7 @@ bool sendLiveLedsWs(uint32_t wsClient)
 #ifndef WLED_DISABLE_2D
     if (strip.isMatrix && n>1 && (i/Segment::maxWidth)%n) i += Segment::maxWidth * (n-1);
 #endif
-    uint32_t c = 0;
-    if (strip.getMappedPixelIndex(i) < 0xFFFF) // draw ledmap gaps gaps in black
-      c = strip.getPixelColor(i);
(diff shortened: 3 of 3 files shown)


wled00/ws.cpp
```diff
@@ -236,7 +236,9 @@ bool sendLiveLedsWs(uint32_t wsClient)
 #ifndef WLED_DISABLE_2D
     if (strip.isMatrix && n>1 && (i/Segment::maxWidth)%n) i += Segment::maxWidth * (n-1);
 #endif
-    uint32_t c = strip.getPixelColor(i);
+    uint32_t c = 0;
+    if (strip.getMappedPixelIndex(i) < 0xFFFF) // draw ledmap gaps gaps in black
```
softhack007: Maybe compare with `` < getLengthTotal())`` instead of 0xFFFF ? This would also catch pixels that are mapped to "somewhere outside", which should also be handled as "black" so users know these pixels are not visible.
For better performance, you could cache the value of getLengthTotal() before entering the getPixelColor() loop.

DedeHai: I did use 0xFFFF on purpose: if your map is larger than the set number of pixels, i.e. the hardware setup is wrong, it will not display the map correctly. on the other hand: using `getLengthTotal()` would hint at such an error and may be more useful to the struggeling user ;)

blazoncek:
A much better and universal approach is:
```c++
inline uint32_t getPixelColor(uint16_t n) const     { return (getMappedPixelIndex(n) < getLengthTotal()) ? _pixels[n] : 0; }
```
in WS2812FX class.

@DedeHai you are wrong, HW set-up can be larger than a matrix and is still valid. Also ledmap can never be larger than physical LED count (`getLengthTotal()`).

wled00/FX.h
```diff
@@ -965,7 +965,8 @@ class WS2812FX {
     };
 
     unsigned long now, timebase;
-    inline uint32_t getPixelColor(unsigned n) const { return (n < getLengthTotal()) ? _pixels[n] : 0; } // returns color of pixel n
+    inline uint32_t getPixelColor(unsigned n) const { return (getMappedPixelIndex(n) < getLengthTotal()) ? _pixels[n] : 0; } // returns color of pixel n, black if out of (mapped) bounds
```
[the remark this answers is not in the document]
DedeHai: this is intentional, the mapping happens when the pixel is written to the bus, here we do want unmapped pixels but black if its OOB.

wled00/FX.cpp
```diff
@@ -161,7 +161,7 @@ uint16_t mode_copy_segment(void) {
     } else { // 1D source, source can be expanded into 2D
       for (unsigned i = 0; i < SEGMENT.vLength(); i++) {
         if(SEGMENT.check2) {
-          sourcecolor = strip.getPixelColor(i + sourcesegment.start); // read from global buffer (reads the last rendered frame)
+          sourcecolor = strip.getPixelColorNoMap(i + sourcesegment.start); // read from global buffer (reads the last rendered frame)
```
blazoncek, line 164: I'm sorry to open this old wound, but this is so wrong IMO.
None of the effects should depend on `strip` pixels.

~~Get segment's pixels using `sourcesegment.getPixelColor()` instead.~~

DedeHai: there is currently no other way to get a "full copy" of a segment and I see the copy FX more of a strip function disguised as an FX for UI convenience. Is your concern based  a technical (future) difficulty or rather architectural aesthetic?

willmmiles: In a perfect world, I'd ask for all the options - I can see each of them being useful in different contexts.

1. Re-blend the internal "segment contents" from the source segment, essentially just re-rendering it with different pixel coordinates.  Implementation would be like calling `blendSegment()` with the source segment, but overriding some output properties (start coord, grouping, spacing).
2. Blend the "rendered but not blended" output of the source segment, ie. respecting source segment grouping/skipping.  Implementation would be like calling `blendSegment()` for the source segment in to a temporary buffer with start coords set to 0, then calling `blendSegment()` to blend the contents of that buffer to main buffer.
3. Copy (blit) a block of previously-rendered pixels from one location in the pixel output buffer to another.  Useful to capture a blended output.

DedeHai: @willmmiles if I understand this right, we currently have options 1 and 3 implemented in copy FX. Option2 might be doable if it were implemented as a usermod, did not think it through. I agree that the ideal implementation would allow all three options. The question I am reading from @blazoncek concern is that it may pollute code in a way that hinders future progress as it should stay backwards compatible. 
I dont want to stop discussion about this, but this discussion should be moved to discord or a separate issue as this is bikeshedding.

PR #5108, opened 2025-11-22
(POC) Large effect count management

willmmiles:
Pursuant to the ongoing discussions regarding handling large numbers of effects, this spike targets the core API changes I think set the most practical direction for future effect management.  The design intent is to preserve compatibility with existing APIs (up to the 255 effect limit), and open a way for new APIs to break that limitation with minimal disruption to existing integrations.

First, the internal upgrade:

The key API changes center around supporting "non-contiguous" effect IDs -- be it names, a paging system, a usermod indexing scheme, or whatever.  In all cases, the "effect identifier" is no longer strictly an index in to a vector.  To this end, I refactored the effect metadata to a new `struct Effect`, keeping all per-effect information (name, function, id) in one place, and built an interface that relies on using functions to look up the Effect entry by name or ID.

Internal API changes:
- New:
  - Adds new `struct Effect` to collect effect-related data in one place
  - Moves global list of effects from `class WS2812FX` to new `namespace Effects`
  - Add effect lookup functions `Effects::getEffectById()` or `Effects::getEffectByName()`.  Function implementations can be replaced or extended to support different indexing approaches (larger id type? numeric ID maps?) without breaking API compatibility.
- Changed:
  - Segments now use an `Effect*` to look up the function.
  - Mode-related API on WS2812FX now implemented by calling Effects namespace functions.
  - `Segment::mode` is preserved for API compatibility, but marked as deprecated.
- Removed:
  - `WS2812FX::getModeDataSrc()` was the one existing internal API that could not be trivially supported.  This impacted one usermod.

External API changes:
None.  (Or at least none intentionally; I might have missed a special case re "Solid"'s metadata (or lack thereof)).  Endpoints 'json/fxda' and 'json/eff' continue to return arrays where the index in to the array matches the 'getEffectById()' callback.  Those arrays remain limited to 255 entries to ensure strict compatibility.

---

From there, the external JSON API is expanded:
- New 'json/fxmap' endpoint returns a JSON object, where keys are the effect names, and the values objects of the form `{"info": "...", "id": id}`, where "id" is only present if the effect id is representable in a uint8_t.  This allows an arbitrarily long list of effects to be described to a client without gaps for unused IDs.
- JSON API now accepts `{"ef": "effect name"}` when deserializing a segment, and emits this key when serializing.  If "ef" and "fx" are both supplied, "ef" has precedence.

The following commit adapts the web UI to use this API instead of ids.   

---

The last commit saves some RAM by moving the Effect objects to code space.

---

Implementation notes:
- I have made no effort to optimize effect name/id lookup.  This does not matter to rendering as the Effect pointer is stored directly.  This could be added later to speed up some of the API calls.
- A `strnchr_P` function for ESP8266 would be nice.
  - or alternately, we could choose to store effect name and metadata in separate literals.  That might be a breaking API change though - I don't know if it's possible to `constexpr`-transform the existing literals at compile time.
  - Another alternative might be store the name length (calculated at compile time).
- Automatic "legacy" ID assignment is still performed for effects up to 255.  After that either names will still work; and the architecture permits us to easily expand the numeric ID data type and/or add some expanded ID space as a new API (pages, a configurable mapper, or all of the above).
- Functions to select a random effect, or "step through" effects, could be added to `namespace Effects`.  Current implementations of these features in various endpoints work by mutating IDs, which will be problematic if the ID space becomes sparse.
- The fxmap endpoint implementation either fits a whole effect in to a packet, or defers it for the next one.  This is an inefficient use of bandwidth: keeping a local line buffer would allow it to be more bandwidth friendly at the cost of more RAM usage.
- This implemention still keeps all `Effect*`s in a dynamically allocated vector.  A future design could also offer a compile-time static list, saving a bit more RAM.  The dynamic list could be kept (it's cost is small when empty) for compatibility with older APIs as well as providing a foundation to support runtime-defined effects (ARTI-FX or similar) as fully integrated members.

## Summary by CodeRabbit

* **New Features**
  * Effects are now identified and selected by name, improving consistency across the web interface and device integrations.
  * Added a consolidated effect map endpoint for retrieving effect names, IDs, and parameter information.
  * Effect metadata and default settings are handled more reliably across segments.

* **Bug Fixes**
  * Improved effect synchronization through DMX, UDP, playlists, overlays, and other controls.
  * Preserved compatibility with numeric effect selections while supporting the new name-based format.

### DIFF
--- a/usermods/usermod_v2_rotary_encoder_ui_ALT/usermod_v2_rotary_encoder_ui_ALT.cpp
+++ b/usermods/usermod_v2_rotary_encoder_ui_ALT/usermod_v2_rotary_encoder_ui_ALT.cpp
@@ -178,9 +178,6 @@ class RotaryEncoderUIUsermod : public Usermod {
     void* display;
   #endif
 
-    // Pointers the start of the mode names within JSON_mode_names
-    const char **modes_qstrings;
-
     // Array of mode indexes in alphabetical order.
     byte *modes_alpha_indexes;
 
@@ -265,7 +262,6 @@ class RotaryEncoderUIUsermod : public Usermod {
       , currentSat1(255)
       , currentCCT(128)
       , display(nullptr)
-      , modes_qstrings(nullptr)
       , modes_alpha_indexes(nullptr)
       , palettes_qstrings(nullptr)
       , palettes_alpha_indexes(nullptr)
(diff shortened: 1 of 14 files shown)


Aceftwspades:
Data point in support of this, if it is useful for prioritizing.

I have a cube usermod that registers 33 effects. Against 16.0.1's 220 built-ins that does not fit many more — 255 minus the built-ins leaves ~35 slots for everything on the device, and audioreactive takes some of those. The effects over the line are dropped by addEffect() returning 255, which nothing checks, so they just do not appear.

The part that cost me the most time is that which effects vanish is not stable: usermods register from static initialisers across translation units, so link order decides who gets the remaining slots. It presents as "some of my effects are missing, and it changes between builds". I have opened #5827  for the silent-failure half, since that seems worth fixing regardless of what happens here.

On this PR's design specifically: addressing effects by name rather than by a wider id matches what I independently ended up doing inside my own usermod. I persist per-effect settings keyed on a hash of the effect name, precisely because the numeric ids shift every time an effect is added or removed, and anything I stored against an id silently attached itself to a different effect on the next build. Names were the only stable handle available. {"ef": "effect name"} would have made that a supported thing rather than a workaround.

33 usermod effects and counting. I'm addicted and i don't want to stop creating

softhack007:
@coderabbitai full review

PR #5111, opened 2025-11-25
Safe uid generation

DedeHai:
instead of using Strings, this generates a 64bit unique ID from the MAC address, salted with unique chip information:
- ESP8266 uses flash chip ID, flash chip size and flash vendor ID
- ESP32 variants use full silicon revision plus ADC calibration data

The 64bit fingerprint is then converted to a string and hashed. I tested the `generateDeviceFingerprint()`  function on all platforms with various ESPs and always got consistent, unique outputs.

edit 
forgot the important bit: no more use of "potentially dangerous" direct access of efuses, only standard IDF calls.

## Summary by CodeRabbit

* **Refactor**
  * Device identification now uses a stronger hardware-based fingerprint across supported boards for more consistent unique IDs.
  * Removed legacy low-level hardware dump usage from the ID flow, simplifying and modernizing identification.
  * Expected impact: more reliable, consistent device IDs and reduced dependence on deprecated hardware reads.

### DIFF
--- a/wled00/util.cpp
+++ b/wled00/util.cpp
@@ -1207,6 +1207,7 @@ String generateDeviceFingerprint() {
 // Returns: original SHA1 + last 2 chars of double-hashed SHA1 (42 chars total)
 String getDeviceId() {
   static String cachedDeviceId = "";
+  if (cachedDeviceId.length() > 0) return cachedDeviceId;
   // The device string is deterministic as it needs to be consistent for the same device, even after a full flash erase
   // MAC is salted with other consistent device info to avoid rainbow table attacks.
   // If the MAC address is known by malicious actors, they could precompute SHA1 hashes to impersonate devices,


wled00/util.cpp
```diff
@@ -1159,60 +1159,61 @@ String computeSHA1(const String& input) {
...
+  if (ch.high_curve) {
+    for (int i = 0; i < 8; i++) {
+      fp[1] ^= ch.high_curve[i];
+    }
   }
-  return result;
+  char fp_string[17];  // 16 hex chars + null terminator
+  sprintf(fp_string, "%08X%08X", fp[1], fp[0]);
+  return String(fp_string);
+}
```
DedeHai, line 1193: good call, I wrongly assumed initialization to 0

wled00/util.cpp
```diff
@@ -1159,60 +1159,61 @@ String computeSHA1(const String& input) {
...
```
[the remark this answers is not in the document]
DedeHai: we are still using IDF V4

wled00/util.cpp
```diff
@@ -1159,60 +1159,61 @@ String computeSHA1(const String& input) {
...
 }
 #endif
 
-
-// Generate a device ID based on SHA1 hash of MAC address salted with "WLED"
+// Generate a device ID based on SHA1 hash of MAC address salted with other unique device info
 // Returns: original SHA1 + last 2 chars of double-hashed SHA1 (42 chars total)
 String getDeviceId() {
   static String cachedDeviceId = "";
-  if (cachedDeviceId.length() > 0) return cachedDeviceId;
```
netmindz, line 1194: Why have you removed the caching if this  potentially expensive value?

wled00/util.cpp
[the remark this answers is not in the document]
netmindz: Is this still going to work for 0.15.x that is on V3?

wled00/util.cpp
DedeHai: I forgot to remove the static :)
how often do we need this value? its not particularly expensive to calculate.

wled00/util.cpp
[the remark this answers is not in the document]
netmindz: @coderabbitai Is this still going to work for 0.15.x that is on V3?

wled00/util.cpp
netmindz: The info endpoint is potentially read quite a lot, especially potentially by chatty third party systems. sha hashing is fairly expensive, so the free extra bytes seems like a safer option to me.

Thoughts @willmmiles ?

wled00/util.cpp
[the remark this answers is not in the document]
DedeHai: you are mistaken, SOC_ADC_RTC_MAX_BITWIDTH is defined in esp_adc_cal.h

wled00/util.cpp
softhack007: @netmindz that's a good point. /json/info is reloaded each time the UI is refreshed, and we have 3rd party tools like HA that poll the status more than 10 times per second ... if the value does not change, its sounds reasonable to invest a few bytes for caching.

DedeHai: I forgot about that HA again ;)
On ESP8266 it takes 2.9ms for the hashing, so caching is justified.

edit:
on ESP32 it takes 0.6ms

netmindz:
Has this been tested for 8226, classic, S2, S3 and C3 @DedeHai  or do you need help from other for any varient?

DedeHai:
I tested the underlying function on all platforms but not the final variant.

DedeHai:
If the plan is to put this into 0.15.2 without any beta testing, someone else should confirm this works on all platforms, especially on older silicon revisions which unfortunately I do not own.

netmindz:
> If the plan is to put this into 0.15.2 without any beta testing, someone else should confirm this works on all platforms, especially on older silicon revisions which unfortunately I do not own.

I'd like to get this merged into main, then we will have a few days of nightly builds and then tagging 0.15.2 hopefully this weekend

DedeHai:
ok, I will re-run the code as-is on all platforms

netmindz:
I'm currently ready to move house next Wednesday, so I have limited hardware available, but I did keep out a classic, S3 and C3, think I missed keeping an S2 out of the boxes so might need someone else to do that one and 8226 as we actually as I don't use them

netmindz:
As long as at least one person has seen it running on actual hardware for each of the 5 targets, that's ok for merge

DedeHai:
can confirm it works on all devices: ESP8266, C3, S2, S3 and vanilla ESP32
I even checked on one and the ID is identical after a full flash erase so no bug there.

PR #5113, opened 2025-11-26
FIX for adafruit portal S3: remove extra flash section

DedeHai:
- removed extra flash section in board definition that was causing the FS to be erased upon upload
- renamed the file so it will never use the default board definition in the future

@softhack007 should we also remove the 
```
      "ldscript": "esp32s3_out.ld",
      "partitions": "partitions-8MB-tinyuf2.csv"
```
from the files build section? Reasoning: if a user's env does not specify partitions, it will still compile but with wrong partitioning.

## Summary by CodeRabbit

* **Chores**
  * Renamed the board profile to indicate WLED support.
  * Updated partition configuration for the board.
* **Bug Fixes**
  * Removed an extra flash image specification to prevent unintended filesystem erase during uploads.

### DIFF
--- a/boards/adafruit_matrixportal_esp32s3_wled.json
+++ b/boards/adafruit_matrixportal_esp32s3_wled.json
@@ -2,7 +2,7 @@
   "build": {
     "arduino":{
       "ldscript": "esp32s3_out.ld",
-      "partitions": "partitions-8MB-tinyuf2.csv"
+      "partitions": "default_8MB.csv"
     },
     "core": "esp32",
     "extra_flags": [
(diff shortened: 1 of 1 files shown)


softhack007:
@DedeHai good idea, I was also thinking maybe the "extra image" might be harmful because there is no space reserved for this in the non-UF2 partitions.csv. 

> should we also remove the
> "ldscript": "esp32s3_out.ld", 
> "partitions": "partitions-8MB-tinyuf2.csv" 
> from the files build section?

Maybe we just change it to the "normal 8MB arduino-esp32 partition" file - I would better not remove this part completely. All board manifests have a partitions entry, and usually they are too small or they use FAT32 so wled will not work any way when users forget to set build.partitions.

even the "default_partitions" set in [env] (or was is another section? Not sure atm) will be better than the default used in most board.json manifest files.

boards/adafruit_matrixportal_esp32s3_wled.json
softhack007: change to 
````json
"partitions": "default_8MB.csv"
````

DedeHai:
perfect, thanks. I have been busy working on something else, I will ping you in a moment.

PR #5116, opened 2025-11-26
Add report version feature

netmindz:
New branch to replace the broken co-pilot branch https://github.com/wled/WLED/pull/5098 

## Summary by CodeRabbit

* **New Features**
  * Automatic version check on startup with install/upgrade prompt (Yes / Not Now / Never Ask)
  * Optional reporting flow to send device/version info when users opt in
  * User choice to permanently suppress future upgrade notifications, persisted across loads

* **Style**
  * Button margin styling enabled for improved spacing

### DIFF
--- a/wled00/data/index.js
+++ b/wled00/data/index.js
@@ -3321,65 +3321,65 @@ function checkVersionUpgrade(info) {
 	fetch(getURL('/edit?func=edit&path=/version-info.json'), {
 		method: 'get'
 	})
-	.then(res => {
-		if (res.status === 404) {
-			// File doesn't exist - first install, show install prompt
-			showVersionUpgradePrompt(info, null, info.ver);
-			return null;
-		}
-		if (!res.ok) {
-			throw new Error('Failed to fetch version-info.json');
-		}
-		return res.json();
-	})
-	.then(versionInfo => {
-		if (!versionInfo) return; // 404 case already handled
-		
-		// Check if user opted out
-		if (versionInfo.neverAsk) return;
-		
-		// Check if version has changed
-		const currentVersion = info.ver;
-		const storedVersion = versionInfo.version || '';
-		
-		if (storedVersion && storedVersion !== currentVersion) {
-			// Version has changed, show upgrade prompt
-			showVersionUpgradePrompt(info, storedVersion, currentVersion);
-		} else if (!storedVersion) {
-			// Empty version in file, show install prompt
-			showVersionUpgradePrompt(info, null, currentVersion);
-		}
-	})
-	.catch(e => {
-		console.log('Failed to load version-info.json', e);
-		// On error, save current version for next time
-		if (info && info.ver) {
-			updateVersionInfo(info.ver, false);
-		}
-	});
+		.then(res => {
+			if (res.status === 404) {
+				// File doesn't exist - first install, show install prompt
+				showVersionUpgradePrompt(info, null, info.ver);
+				return null;
+			}
+			if (!res.ok) {
+				throw new Error('Failed to fetch version-info.json');
+			}
+			return res.json();
+		})
+		.then(versionInfo => {
+			if (!versionInfo) return; // 404 case already handled
+
+			// Check if user opted out
+			if (versionInfo.neverAsk) return;
+
+			// Check if version has changed
+			const currentVersion = info.ver;
+			const storedVersion = versionInfo.version || '';
+
+			if (storedVersion && storedVersion !== currentVersion) {
+				// Version has changed, show upgrade prompt
+				showVersionUpgradePrompt(info, storedVersion, currentVersion);
+			} else if (!storedVersion) {
+				// Empty version in file, show install prompt
+				showVersionUpgradePrompt(info, null, currentVersion);
+			}
+		})
+		.catch(e => {
+			console.log('Failed to load version-info.json', e);
+			// On error, save current version for next time
+			if (info && info.ver) {
+				updateVersionInfo(info.ver, false);
+			}
+		});
 }
 
 function showVersionUpgradePrompt(info, oldVersion, newVersion) {
 	// Determine if this is an install or upgrade
 	const isInstall = !oldVersion;
-	
+
 	// Create overlay and dialog
 	const overlay = d.createElement('div');
 	overlay.id = 'versionUpgradeOverlay';
 	overlay.style.cssText = 'position:fixed;top:0;left:0;width:100%;height:100%;background:rgba(0,0,0,0.7);z-index:10000;display:flex;align-items:center;justify-content:center;';
-	
+
 	const dialog = d.createElement('div');
 	dialog.style.cssText = 'background:var(--c-1);border-radius:10px;padding:25px;max-width:500px;margin:20px;box-shadow:0 4px 6px rgba(0,0,0,0.3);';
-	
+
 	// Build contextual message based on install vs upgrade
-	const title = isInstall 
-		? '🎉 Thank you for installing WLED!' 
+	const title = isInstall
+		? '🎉 Thank you for installing WLED!'
 		: '🎉 WLED Upgrade Detected!';
-	
+
 	const description = isInstall
 		? `You are now running WLED <strong style="text-wrap: nowrap">${newVersion}</strong>.`
 		: `Your WLED has been upgraded from <strong style="text-wrap: nowrap">${oldVersion}</strong> to <strong style="text-wrap: nowrap">${newVersion}</strong>.`;
-	
+
 	const question = 'Help make WLED better with a one-time hardware report? It includes only device details like chip type, LED count, etc. — never personal data or your activities.'
 
 	dialog.innerHTML = `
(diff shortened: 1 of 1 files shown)


netmindz:
So the current wording is

```Would you like to help the WLED development team by reporting your installation? This helps us understand what hardware and versions are being used.```


What about 

```Would you like to help the WLED development team by reporting your installation?  This is a one time action that sends details of your hardware configuration, but no personal or behavioral data, e.g which esp32 chip variant, how many LED pixels etc```

**Comment by netmindz** [comment]
Screenshot From 2025-11-26 22-52-56
or the shorter
Screenshot From 2025-11-26 22-56-00

**Comment by netmindz** [comment]
I like shorter version. Does this address your concerns @DedeHai ?

File: wled00/data/index.css [hunk 1]
```diff
@@ -794,7 +794,7 @@ input[type=range]::-moz-range-thumb {
 /* buttons */
 .btn {
 	padding: 8px;
-	/*margin: 10px 4px;*/
+	margin: 10px 4px;
```
DedeHai, line 797: are you sure this wont break anything else in the UI?

DedeHai:
I also like the sort version. Did feed it to an AI to get a few variants, this sounded best to me:
_Help make WLED better with a one-time hardware report?
It includes only device details like chip type, LED count, etc. — never personal data._

in the policy, the last sentence is a bit confusing:
"A single submission occurs only when you click Accept on the post-install/upgrade prompt. No continuous telemetry is sent by default."
why the "by default"? it suggests that there could be continuous telemetry.

DedeHai:
Tested this and works as intended but it suffers from the same issue I posted in the last comment in https://github.com/wled/WLED/pull/5013 - it wont show after an update unless I do a full refresh, at least in chrome.

wled00/data/index.css
netmindz: Buttons for a touch UI should always have a margin as otherwise you risk clicking the wrong item. 
I'll have a check through however for other used of that class to see if this change is an improvement elsewhere too however

netmindz:
> Tested this and works as intended but it suffers from the same issue I posted in the last comment in #5013 - it wont show after an update unless I do a full refresh, at least in chrome.

We might need to check what cache control related headers are being send and confirm the backend supports Not Modified responses. Please open a separate issue for that

netmindz:
> I also like the sort version. Did feed it to an AI to get a few variants, this sounded best to me: _Help make WLED better with a one-time hardware report? It includes only device details like chip type, LED count, etc. — never personal data._

I like that, but I still want to get some kind of reference to the fact that it excludes anything behavioural/activity based

> in the policy, the last sentence is a bit confusing: "A single submission occurs only when you click Accept on the post-install/upgrade prompt. No continuous telemetry is sent by default." why the "by default"? it suggests that there could be continuous telemetry.

Yeah I agree that "automatically" would be better than "by default"

wled00/data/index.js
```diff
@@ -3304,6 +3304,195 @@ function simplifyUI() {
...
+		}
+	})
+	.catch(e => {
+		console.log('Failed to load version-info.json', e);
+		// On error, save current version for next time
+		if (info && info.ver) {
+			updateVersionInfo(info.ver, false);
+		}
+	});
+}
```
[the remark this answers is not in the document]
netmindz: Please confirm that this is resolved now @coderabbitai

wled00/data/index.js
```diff
@@ -3304,6 +3306,195 @@
...
+		body: formData
+	})
+	.then(res => res.text())
+	.then(data => {
+		console.log('Version info updated', data);
+	})
+	.catch(e => {
+		console.log('Failed to update version-info.json', e);
+	});
+}
```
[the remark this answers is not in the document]
netmindz: Please update your comment to reflect the current code @coderabbitai

wled00/data/index.js
```diff
@@ -3304,6 +3306,195 @@
...
+		<p style="color:var(--c-f);">${question}</p>
+		<p style="color:var(--c-f);font-size:0.9em;">
+			<a href="https://kno.wled.ge/about/privacy-policy/" target="_blank" style="color:var(--c-6);">Learn more about what data is collected and why</a>
+		</p>
+		<div style="margin-top:20px;">
+			<button id="versionReportYes" class="btn">Yes</button>
+			<button id="versionReportNo" class="btn">Not Now</button>
+			<button id="versionReportNever" class="btn">Never Ask</button>
+		</div>
+	`;
```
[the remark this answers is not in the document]
netmindz: Incorrect. See the comments from the rabbit

Aircoookie:
> > Tested this and works as intended but it suffers from the same issue I posted in the last comment in #5013 - it wont show after an update unless I do a full refresh, at least in chrome.
> 
> We might need to check what cache control related headers are being send and confirm the backend supports Not Modified responses. Please open a separate issue for that

It doesn't seem to me like `VERSION` in `wled.h` has been updated in this PR (or recently at all, for that matter, it's still on `2506160` in main). UI cache invalidation is tied to this version number, so updating it every time there is a change to UI code should be all that is required for it to work without an explicit refresh.

Aircoookie:
> I like that, but I still want to get some kind of reference to the fact that it excludes anything behavioural/activity based

Then why not: Help make WLED better with a one-time hardware report? It includes only device details like chip type, LED count, etc. — never personal data or your activities.

About the policy, I think we can just do "No continuous telemetry is sent."

wled00/data/index.css
DedeHai: I did not see anything obvious. just wondering why it was commented. Mayb just to save a few bytes.

netmindz:
> It doesn't seem to me like `VERSION` in `wled.h` has been updated in this PR (or recently at all, for that matter, it's still on `2506160` in main). UI cache invalidation is tied to this version number, so updating it every time there is a change to UI code should be all that is required for it to work without an explicit refresh.

It's on my list to kill this field at some point, a value that developers randomly update by hand is not acceptable.

If we need it all, it should be updated by a pipeline that runs after a pr merge, this should never be something a developer is editing by hand on a branch

Aircoookie:
Agreed. It has also caused merge conflicts in the past. I believe this value is very useful, updating it automatically after every push to main would make things a lot easier.

wled00/data/index.js
```diff
@@ -3304,6 +3306,195 @@ function simplifyUI() {
...
+	const versionInfo = {
+		version: version,
+		neverAsk: neverAsk
+	};
+
+	// Create a Blob with JSON content and use /upload endpoint
+	const blob = new Blob([JSON.stringify(versionInfo)], {type: 'application/json'});
+	const formData = new FormData();
+	formData.append('data', blob, 'version-info.json');
+
```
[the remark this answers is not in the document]
DedeHai: this discrepancy is handled by the server code:
```
    String finalname = filename;
    if (finalname.charAt(0) != '/') {
      finalname = '/' + finalname; // prepend slash if missing
    }
```

netmindz:
> > I like that, but I still want to get some kind of reference to the fact that it excludes anything behavioural/activity based
> 
> Then why not: Help make WLED better with a one-time hardware report? It includes only device details like chip type, LED count, etc. — never personal data or your activities.
> 
> About the policy, I think we can just do "No continuous telemetry is sent."

Both updated. Hopefully good to go?

wled00/data/index.js
```diff
@@ -3304,6 +3306,195 @@ function simplifyUI() {
...
+			console.log('Failed to load version-info.json', e);
+			// On error, save current version for next time
+			if (info && info.ver) {
+				updateVersionInfo(info.ver, false);
+			}
+		});
+}
+
+function showVersionUpgradePrompt(info, oldVersion, newVersion) {
+	// Determine if this is an install or upgrade
```
wmiles-sgl: Can we test if the update server is accessible here before showing the prompt?  It might save a lot of "failed to report" cases if the setup device is also offline.

wmiles-sgl (commented):
Nice and clean!  👍

wled00/data/index.js
willmmiles, line 3363: Can we test if the update server is accessible here before showing the prompt? It might save a lot of "failed to report" cases if the setup device is also offline.

wled00/data/index.js
```diff
@@ -3304,6 +3306,195 @@ function simplifyUI() {
...
+				},
+				body: JSON.stringify(upgradeData)
+			});
+		})
+		.then(res => {
+			if (res.ok) {
+				showToast('Thank you for reporting!');
+				updateVersionInfo(info.ver, false);
+			} else {
+				showToast('Report failed. Please try again later.', true);
```
willmmiles, line 3463: Future: we might want to consider saving "user gave permission, but we failed to report" in the version info file. That way we could avoid re-prompting the user with a question they've already answered (and driving more people to click "never again" to just to get it to go away).

wled00/data/index.js
netmindz: Overkill I'd say. It's suppressed already for ap mode, plus to check that connectivity would mean making a connection without consent so definitely not a good idea

wled00/data/index.js
netmindz: Nice idea, but for the first release let's keep this to a strict simple relationship between consent and action. If it's well received we can consider "lasting consent"

willmmiles: Fair enough.  I'm just concerned that being annoying about reporting will exacerbate any privacy concerns.

wled00/data/index.js
willmmiles: Yeah, it's too bad that we can't directly perform DNS resolution from the browser without connecting to anyone. :(  As below I am mostly worried about annoying offline people with repeated prompts.

netmindz: You are correct that we could do a hostname lookup only if we are trying to detect internet access rather than connectivity to the usage platform specifically, which would be a GDPR compliant approach.

If folk are always offline, then they can just click never. 

Let's see how things go and can always improve based on feedback

willmmiles: As far as I could find, only Firefox offers a DNS resolution API.  Sadly other browsers offer no solution for testing connectivity other than opening a connection somewhere.  Maybe someday..

PR #5134, opened 2025-11-29
clear enable bit on unused time macros

DedeHai:
bit 0 of weekday is the "enable" and should not be set by default, it is confusing users if unused macros are enabled by default.

## Summary by CodeRabbit

* **New Features**
  * Added support for day-end timer configuration options
  * Introduced playlist auto-advance control capability
  * Enhanced timer scheduling with updated initialization parameters


blazoncek:
There is no "preset 0" and can never be one, hence any enabled timer will do nothing, even if enabled.
However, I am not objecting to the change but would also like to see the existing timers to be update to reflect default behaviour.

DedeHai:
this is not about existing timers but NON existing timers being enabled by default. Its purely cosmetic.

Issue #5136, opened 2025-11-30
WebUI ERR_EMPTY_RESPONSE after updating esp8266 from 0.15.1 to 0.15.2


popy2k14:
### What happened?

I am running 3x esp8266 with 0.15.1.
HA messaged me today that there is an update tpo 0.15.2.
Done the update to one device over HA (arduino OTA method) and now the WebUI doesnt work.
What i have tried:

- Device is pingable
- Hyperion can send udp led data from my machine to the device and it is displayed (so it's not completely dead)
- Different browsers (brave, chrome & edge) throw ERR_EMPTY_RESPONSE  when accessing: http://<ip>
- `curl http://<ip>` log find attached
- Tried to reach /update/ without successs

- Flashed 0.15.2 with usb from install.wled.me -> it worked and ui was reachable
- Restored json 0.15.1 configuration backup -> webui not reachable anymore
- -> so i assume also the first update flash with HA worked (exact same issue as flashed + restored manually) and thre is an issue with configuration compatibility between 0.15.1 & 0.15.2 (at least for my configuration)

- Flashed 0.15.1 with usb from install.wled.me -> it worked and ui was reachable
- Restored json 0.15.1 configuration backup -> webui is reachable, so my assumption from above seems right

### To Reproduce Bug

Update esp8266 from 0.15.1 to 0.15.2.

### Expected Behavior

WebUI is working

### Install Method

Binary from WLED.me

### What version of WLED?

0.15.1 & 0.15.2

### Which microcontroller/board are you seeing the problem on?

ESP8266

### Relevant log/trace output

```shell
curl http://192.168.0.52 -vvv
16:59:19.342000 [0-x] == Info: [READ] client_reset, clear readers
16:59:19.346000 [0-0] == Info: [SETUP] added
16:59:19.349000 [0-0] == Info:   Trying 192.168.0.52:80...
16:59:19.352000 [0-0] == Info: [SETUP] Curl_conn_connect(block=0) -> 0, done=0
16:59:19.357000 [0-0] == Info: [SETUP] Curl_conn_connect(block=0) -> 0, done=1
16:59:19.361000 [0-0] == Info: Connected to 192.168.0.52 (192.168.0.52) port 80
16:59:19.365000 [0-0] == Info: using HTTP/1.x
16:59:19.367000 [0-0] => Send header, 76 bytes (0x4c)
0000: GET / HTTP/1.1
0010: Host: 192.168.0.52
0024: User-Agent: curl/8.14.1
[...]

PR #5138, opened 2025-12-01
Fix ESP8266 DMA off-by-one

### LINKED ISSUE #4906: ESP8266 DMA method (pin 3) apparently uses all memory
### What happened?

When using pin 3 for digital output on ESP8266, if the length is exactly right such that the buffer size needed for the  3-bit I2S cadence works out to a multiple of 4 bytes, then the system appears to run out of memory.

This seems to be caused by an off-by-one error in the DMA buffer translation introduced in NeoPixelBus v2.8.1.

The overrun typically squashes the "block size" of the next block of heap -- often the majority of the free RAM -- so the system reports no available memory and is unable to respond on the web interface.

### To Reproduce Bug

On an ESP8266, set a WS2812 strip of length 300 on pin 3.

### Expected Behavior

LED output should proceed and the system should respond on the web interface

### Install Method

Self-Compiled

### What version of WLED?

0.15.1, 0.16

ESP8266

### Relevant log/trace output

```shell

```

### Anything else?

Reference discussion in #4614.

### Code of Conduct

willmmiles:
Shim in https://github.com/Makuna/NeoPixelBus/pull/894 until approved by upstream.  Fixes #4906 and #5136.

## Release Notes

* **Bug Fixes**
  * Fixed DMA handling on ESP8266 boards for improved compatibility and performance.

### DIFF
--- a/lib/NeoESP8266DMAFix/include/NeoEsp8266DmaMethodFix.h
+++ b/lib/NeoESP8266DMAFix/include/NeoEsp8266DmaMethodFix.h
@@ -1,0 +1,122 @@
+/*-------------------------------------------------------------------------
+NeoPixel library helper functions for Esp8266.
+
+FIXED VERSION FROM https://github.com/Makuna/NeoPixelBus/pull/894
+This library will overlay/shadow the base version from NeoPixelBus 
+
+Written by Michael C. Miller.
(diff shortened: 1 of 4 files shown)


netmindz:
Thanks for the quick turnaround @willmmiles 

I'll do a merge and refresh today's nightly build to make it easier for people to test

PR #5140, opened 2025-12-01
add more macro/timer slots

benjamw:
This is not a PR in the strict sense, but more a request for conversation.

There have been a few mentions in the Discord channels about increasing the number of macro slots, and the responses given there were not to my satisfaction. So I went ahead and tried it on my controller.

It works just fine, but there are a few concerns/questions, notably, there are some choices that were made prior to this PR that affect how this change affects end-users. Mostly stemming from the indexes the sunrise and sunset macros have. Those macros were placed at the end of the array and the macro addition can be done in a few ways:

1. Move the sunrise/sunset macros to the start of the array, giving them indexes 0 and 1. This will allow future expansion without further affecting users. Although it will affect users who update. This is the method I most agree with, but do agree there are backwards incompatibilities which are bad.
2. Keep them where they are, at indexes 8 and 9, keeping everything backward compatible, but causing weirdness for counting the expanded macros.
3. Move them to the end, keeping with the current method, giving them new indexes 16 and 17. This is arguably the worst method, causing issues now, as well as in the future when more expansion may be added. But this is the choice that was made in this PR because I didn't really think it through before I made the changes and was just going for proof of concept.

Anywho, this PR adds 8 more macro slots, and has been tested and works great. The only issue that I can see is the backward incompatibility for saved configs, and the issue I found of not being able to save the macros in the EEPROM. The macros save just fine in the internal config file, and load from there after startup and run just fine as well on my ESP32.

I don't know how this will affect any other controller as I only have the ESP32, and this does increase the size of the config file, so, it may be too big for smaller controllers.

Please let me know if this is a bad idea for any other reason, but as I see it right now, I can't find any glaring issues against it.

I will also happily modify the code to use any of the above 3 options for adding the macros to the array, option #1 being my preference, but #2 being a good workaround. And I may do that soon anyway.

NOTE: AI was used to help me with this PR.

## Release Notes

* **New Features**
  * Redesigned timer configuration interface with dynamic timer creation
  * Added preset sorting functionality (by ID or name) in timer settings
  * Integrated macro preset selection into timer management
  * Enhanced sunrise/sunset timer support with date range controls

### DIFF
--- a/wled00/data/settings_time.htm
+++ b/wled00/data/settings_time.htm
@@ -42,14 +42,11 @@
 		}
 
 		if (!json) {
-			showPresetWarning();
+			showToast("Please visit the main UI first to load presets", true);
 			return;
 		}
 
 		onPresetsLoaded(json);
-	}
-	function showPresetWarning() {
-		showToast("Please visit the main UI first to load presets", true);
 	}
 	function onPresetsLoaded(json) {
 		presets = json;


wled00/data/settings_time.htm
```diff
@@ -43,41 +43,42 @@
...
 			let wd = gId("W"+i).value;
 			for(j=0;j<8;j++) {
 				gId("W"+i+j).checked=wd>>j&1;
 			}
-			if ((wd&254) != 254 || (i<8 && (gN("M"+i).value != 1 || gN("D"+i).value != 1 || gN("P"+i).value != 12 || gN("E"+i).value != 31))) {
+			// there is a bug in this IF clause that doesn't seem to work as expected 
+			if ((wd&254) != 254 || (i<16 && (gN("M"+i).value != 1 || gN("D"+i).value != 1 || gN("P"+i).value != 12 || gN("E"+i).value != 31))) {
 				expand(gId("CB"+i),i); //expand macros with custom DOW or date range set
 			}
 		}
```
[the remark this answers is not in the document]
benjamw: The bug is due to the macros being initialized with the enable bit set. This was fixed in #5134. I also tried to fix it with a loop that unsets the bit. But the method by DedeHai is cleaner. I can fix that if it's decided to move forward with this change.

DedeHai:
appreciate your effort. without looking too much into this: doesn't your update break sunrise/sunset configurations of existing setups?
The way this should be implemented is the same as with other recent changes, like palettes or buttons: use vectors so its future proof or the next request will just be "I need more than 16 macros, my plan requires 24". Now whether this is feasible without breaking backwards compatibility I am not sure, but if compatibility is broken, it better be the "final" version.

blazoncek:
FYI there was a PR that implemented entirely new approach to *timed presets*. It lacked UI though.
It might be better to revisit that as it opened quite a few new possibilities.

wled00/data/settings_time.htm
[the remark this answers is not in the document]
blazoncek: PR #5134 does not handle existing configurations, only new ones.
*If* there truly is a bug in the code it has to be fixed for existing configurations as well.

FYI even if *timer* is enabled it will have no effect if preset selected is 0 as there can never be a preset with ID 0.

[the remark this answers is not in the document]
blazoncek: @coderabbitai it was intended for your learnings too.

benjamw:
So to clarify... Yes, this current implementation does break sunrise/sunset configs as it currently stands. It can be modified to preserve those, and just skip those 2 for regular macro configs.
If you wish to make this the "final" version with minimal disruption, the proper way to do it IMHO would be to move those sunrise/sunset configs to indexes 0 and 1, and put all the macro configs after it, which would ultimately make those infinite.

I'm not sure what is different about _timed presets_, but I can certainly give those a look.

A truly "final" version of the macro system would be to allow users to select sunrise/sunset for _any_ macro, and allow them to add more macros at will, making them basically unlimited. I know there are issues with strongly typed languages not being happy with open-ended arrays. Although converting to vectors may make this a lot easier as there will only be one open-ended array. But if that seems like a good idea, I can certainly look into doing that.
I can also add a check to see if the macros at index 8 and 9 (the current sun macros) are being used and have settings, and move those as needed when being read from the config file. That may make updating painless for users as they won't have any idea that it was moved, and it will just work as it always has.

I'm going to try converting to vectors tonight and see how that pans out with open-ended macros.

benjamw:
I took a look at #4789 and that's basically the same. I can implement those changes here, or extend that PR, or both.

I'll try my previous change here (vectors in an open-ended array), and pull down that PR and see how it compares locally and see how they are different.

I'll also implement a UI that expands as needed, which I thought of doing, but didn't want to get too crazy in a proof of concept.

blazoncek:
Keep in mind RAM constrained ESP8266.

benjamw:
Ok, I have completely rewritten this using the ideas from that other PR as well as my own ideas.

wled00/data/settings_time.htm
```diff
@@ -20,68 +20,317 @@
...
+		tbody.insertBefore(nextRow1, row1);
+		tbody.insertBefore(nextRow2, row1);
+	}
+
+	function FC()
+	{
+		// This is called after settings are loaded to populate the UI
+		// The timer data comes from the server via settings script
+		// We need to wait for the data to be available
+	}
```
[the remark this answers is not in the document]
benjamw: `addTimerRow()` was called elsewhere before this. But this was implemented in case of existing older formats.

benjamw:
Ok, I've built this and tested on my ESP32 and it seems to be working great after some superficial tests and minor fixes.

I'll run some deeper tests around uploading older config files, adding too many timers, deleting and re-ordering in the same save, etc.

blazoncek:
A hefty update that will take time to digest.
On first glance there are two things I don't like: special code to handle legacy (when will it be removed?) and code size increase.

DedeHai:
I agree on the "dont like" points. Keeping the legacy arrays in RAM is wasteful and not acceptable. Also if there was a way to make it "back and forth" compatible would be great so people dont loose their schedules when downgrading, not sure its feasible. 
Code increase is 3.5k which is not acceptable. Keep it as lightweight as possible.

benjamw:
I agree this is a large update, and I have no issues making edits and getting this right.

I can make edits to ensure the config-stored JSON timer array is the same as the older version of config-stored JSON timer array, which will make the "back and forth" possible. The only issue with that is that the old versions don't handle more than 8 timers (+2 for sun timers). So any extra will probably be silently discarded if anybody ever downgrades or moves their settings between controllers on different versions. I'm not sure anything can be done about that. Also, the new timers have the sun timers intermixed with the "regular" timers, so getting those back into their proper indexes is also not possible, unless we make efforts to do so when storing in the config file, which breaks the ordering.

There are also the EEPROM-stored arrays that are not JSON, and have a different layout. Those can also be back-ported on downgrade by saving a config file and then the older version will take the config file and load those values into the EEPROM as they currently do. So those aren't a huge issue unless I'm missing something.

Regarding keeping legacy code around, that was to allow the migration for EEPROM-stored values. If timers are no longer stored that way, and are stored in config files, then that can be removed.

I can also try to tighten up the code by removing comments and white space and making the code leaner. I can also remove some of the UI improvements and revert those back to number inputs.

I'll try making a config file with more than 8 timers in it, and downgrade my ESP32 and load them in to see what happens, but I don't have high hopes that it will work. I'm pretty sure it will silently discard any extras and the sun timers will break.

DedeHai:
> which will make the "back and forth" possible

thats great!

> So any extra will probably be silently discarded if anybody ever downgrades or moves their settings between controllers on different versions.

unavoidable so not a big deal.

> There are also the EEPROM-stored arrays

I am not sure any device uses EEPROM still, the whole EEPROM stuff may be deprecated. I would not worry about it too much.

Try to get the code increase down as much as possible, it's fine if there are no added bells and whistles.

blazoncek:
EEPROM has been deprecated before 0.15 release and last time it was used was in 0.9.

benjamw:
Yeah, I did some digging and discovered that the EEPROM storage was converted to JSON (with the current format I still use, which makes back and forth within a reasonable version possible) in 2020. I've removed all the code that dealt with converting from old formats, and also removed the old timer arrays that are no longer relevant.

I can also remove the old `wled_eeprom.cpp` file to make room for this if requested. But I totally understand if that's waiting for a deprecation period to end.

I also tightened up the code, removed all the comments, whitespace, etc. and removed all the UI bells and whistles.

I looked for any other dead code regarding timers and macros, and was only able to find a few small spots in the `eeprom` file.

DedeHai:
good progress, the UI does not work right though: I get "time:1 An invalid form control with name='H2' is not focusable." when adding sunrise/sunset. Got the same error before saving it without any changes after uploading.

Also adding a timer and not changing anything (i.e. enabled, 0,0,0) does remove it again when saving. Is that intentional? Not saying this needs changing, just checking for potential bugs.

what I was not able to check: does it allow multiple sunrise/sunset presets?

blazoncek:
Looks much better now. Clean and efficient.
If you still need presets (as in 1st implementation), you do not need to fetch them via HTTP request but can check local storage which keeps a copy.

benjamw:
> good progress, the UI does not work right though: I get "time:1 An invalid form control with name='H2' is not focusable." when adding sunrise/sunset. Got the same error before saving it without any changes after uploading.

I'll take a look at this. I haven't done much more than superficial testing yet.

> Also adding a timer and not changing anything (i.e. enabled, 0,0,0) does remove it again when saving. Is that intentional? Not saying this needs changing, just checking for potential bugs.

Yes. Setting the preset to 0 (and hour and minute) deletes the timer. If this is not wanted, I can remove it, but I feel like there should be a way to remove timers.

> what I was not able to check: does it allow multiple sunrise/sunset presets?

It should... and they follow dates, so all my house timers (Halloween, Christmas, etc) can be set to start at sunset.

There is no "global" sunrise or sunset timer anymore, they all fallow schedules (unless you don't set a date, and then it's enabled everyday) (or should be, I'll do some deeper testing when I can)

benjamw:
> Looks much better now. Clean and efficient. 

Thank you :)

> If you still need presets (as in 1st implementation), you do not need to fetch them via HTTP request but can check local storage which keeps a copy.

Not sure what you mean... Ahh, you mean for the dropdowns? I'll take a look and see what can be done with those. But I tried to make it as small as possible, but if you liked the dropdowns, I can re-implement that...

blazoncek:
> if you liked the dropdowns

They are useful, but keep code as small as possible.

DedeHai:
> Yes. Setting the preset to 0 (and hour and minute) deletes the timer. If this is not wanted, I can remove it, but I feel like there should be a way to remove timers.

good point. setting it all to zero to remove may be a bit unintuitive. check how its done in LED config for buttons: a small x button to the right that sets it all to zero and effectively removing it when saving. small addition for a better UX.

benjamw:
FYI, I'm going out of town this weekend, so I'll work on this when I get back.

Just letting you know so you don't think I bailed.

benjamw:
Ok, I re-added the preset dropdowns and made that as small as possible. I also sorted them to match the main display (pretty closely, some punctuation, like underscore, is sorted differently).

I also fixed the `An invalid form control with name='H2' is not focusable` issue. Let me know if you still get this, because I haven't seen it. And if you do, can you provide your cfg.json file for debugging?

I also fixed an annoyance of mine where when swapping between Sunrise/Sunset/Regular, any hour that was set in Regular would disappear when going back to Regular. Very annoying for users who are just playing around, twiddling all the knobs.

DedeHai:
I like the dropdown approach in general but it does break the layout as the field is too wide with long preset names.
also there is no delete button and the selection to delete does not work. 
there is also a bug when adding mulitple timers: it wont respect the checkmarks.
the sorting of the list is unnecessary IMHO. The main UI can also be sorted by number so it wont match.

please do thorough testing also on mobile.

benjamw:
I see the mobile issue where preset names are too long. Unfortunately this is a result of the Macro Button options being in a table and fixing it would either require reverting those sections back to numerical inputs, or adding some javascript/css that would allow the table to collapse into blocks with the screen size.

I can easily revert the inputs, but converting to collapsible blocks would require quite a bit of HTML/CSS/JS updates that would drastically change the layout and/or quickly eat up the size budget.

I've also adjusted the timer clean up logic to delete any timer with preset = 0, so the delete dropdown option should work now.

I'm not sure what was meant by "there is also a bug when adding mulitple timers: it wont respect the checkmarks."... does that mean that the checkmarks are not showing as expected (on or off), or that timers that are enabled are not working, or timers that are disabled are triggering? I can't replicate any of those states.

I've also added the option to select which sort method to use.

I've updated the HTML/CSS to use flex boxes just to see what that looked like.

There is a bug where the Sunset/Sunrise setting is not saved anymore with the new HTML that I'll look at tomorrow.

blazoncek:
> adding some javascript/css

CSS: `text-overflow:ellipsis;` See: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-overflow

> collapsible blocks would require quite a bit of HTML/CSS/JS

Not really, just a class toggle. I.e. `element.classList.toggle("hide");` CSS does the rest.

Overall, nice progress. 👍

benjamw:
> > collapsible blocks would require quite a bit of HTML/CSS/JS
> 
> Not really, just a class toggle. I.e. `element.classList.toggle("hide");` CSS does the rest.

Sorry I was typing fast and not very clear. I didn't mean collapse and hide, I meant collapse and wrap down the page. Tables can't flow with smaller screens. I switched to use flex boxes and individual cards for each for both buttons and timers, so those can more easily flow down the page on smaller screens.

benjamw:
There is a possible bug in this code. Any timers I have set to "Sunset" are not starting at sunset. I'm am looking into it.

benjamw:
I edited the preset the misbehaving timer was using, as well as changed the dates to be actual dates (instead of just all year), and now I can no longer replicate the issue.

The timers are working as expected. Even the Sunset ones.

I'm still not sure what the issue was, but it's no longer causing an issue on my setup.

I'll continue testing to see if I can setup some kind of debugging and replicate the issue on a separate chip.

I have started another chip with a bunch of debugging to test various combinations of timer settings. It's set to run for a week (to enable all the possible combinations), so I should have some good data after a week. (2026-01-13)

benjamw:
Ok, I've completed the test and everything seems to be running as expected.

Must have been a weird or not-set-up-right timer that was failing for me.

wled00/data/style.css
```diff
@@ -150,9 +150,34 @@ td {
   font-size:1.5rem;
   cursor:pointer
 }
-#TMT table {
-  width: 100%;
-}
+/* Timer layout - tmc=container, tb=block, tm=main, ts=section, tf=field, tfe=enabled field, te=expanded, tw=weekdays, tdr=daterange, tdf=date from/to, tc=calendar, twd=weekday text */
+.tmc{display:flex;flex-direction:column;gap:12px;max-width:800px;margin:0 auto}
```
DedeHai: this CSS code seems excessive to me, is all this code really needed?

wled00/data/settings_time.htm
```diff
@@ -19,13 +24,69 @@
...
+				// Update radio buttons to match saved preference
+				if (sortByName) {
+					var sortByNameRadio = document.getElementById("sortByName");
+					if (sortByNameRadio) sortByNameRadio.checked = true;
+				}
+			}
+		} catch(e) {}
+		buildSortedPresetOptions();
+	}
+	function buildSortedPresetOptions() {
```
DedeHai: sorting presets is guilding the lily as changing these presets is mostly a one-time thing for most configurations

DedeHai:
@benjamw just FYI: #5297 conflicts with this PR, which ever PR makes the finish line first (preferrably this one) will have less merge work to do.

wled00/data/settings_time.htm
benjamw: I have almost a hundred presets and trying to find the one I want can be a bit annoying. I added it as a small usability increaser.

wled00/data/style.css
benjamw: The flexbox allows the boxes to self-collapse when used on a phone vs a computer screen. It may not all be absolutely necessary, but it makes it look better with it than without it. I did shrink it as much as I could.

benjamw:
> I ran a quick test and deleting out-of-order seems to have a bug.

What do you mean by "deleting out-of-order"? What did you do? How can I replicate?

wled00/data/style.css
benjamw: I can try to simplify by removing some of the "fluff", like rounded borders and stuff like that.

DedeHai:
the deletig out of order may have been a fluke, I can not reproduce it now.
I did take a closer look and almost half of the added ~1.6k of code size is styling / sorting / localstorage which is all unnecessary or cosmetics: the main UI does fetch it to local storage already, adding the fallback is a edge case. if you really need to add the fetching of presets.json, do it in common.js and drop the storing to localstorage. If you think sorting by name is really useful, then make that the default and drop the options please, we do not need options for option-settings.
I made some improvemets to the UI readability, which I am about to merge (there are coming up PR's that conflict with it, so might yours, sorry about that but should be  a quick fix).
The way you style the buttons - while loking nicer - is also crowding the settings page. It is totally fine if it looks like it does now:
image


I would rather spend those 700 bytes on a nicer scrolling text font (coming up) than styling a config page.

benjamw:
Ok, I've reverted the HTML (with some minor changes to make it flow better) and removed all the CSS.
I've also moved the preset fetch method into common.js and used it in the settings_time.htm file just as it is used in the index.htm/index.js files.

Hopefully this is much smaller.

I also had to update the deep sleep usermod with the new timer struct. If this is not what I should do, let me know, and I'll revert that.

wled00/ntp.cpp
```diff
@@ -557,4 +544,70 @@ void setTimeFromAPI(uint32_t timein) {
...
+      else if (minute > 120) minute = 120;
+    }
+  } else {
+    if (minute < 0 || minute > 59) {
+      DEBUG_PRINTLN(F("Timer: Invalid minute value"));
+      return;
+    }
+  }
+  if ((monthStart != 0 && (monthStart < 1 || monthStart > 12)) ||
+      (monthEnd != 0 && (monthEnd < 1 || monthEnd > 12))) {
```
DedeHai: this is the same as `monthEnd > 12` , same for other clapms

wled00/data/style.css
```diff
@@ -166,32 +166,32 @@ td {
...
-	bottom: 64px;
-	color: #fff;
-	font-size: 17px;
-	padding: 16px;
-	pointer-events: none;
-	position: fixed;
-	text-align: center;
-	z-index: 5;
-	transform: translateX(-50%);
+  opacity: 0;
```
DedeHai: please revert unnecessary changes of indentation (even though the rabbit complains)

DedeHai:
looks good now apart from the two minor things I commented in code. when compiling this PR and comparing to current main, code still grows by 3k which seems really a lot to me given the changes, I'd expect more like 1k. Maybe the rabbit has an answer to that.

DedeHai:
@coderabbitai the changes from this PR use about 3k of flash. can you please check if there are any external libraries brought in and break down which parts of the changes may cause excessive flash use? take into account (and confirm) that htm pages are minified and gzipped.

wled00/data/index.htm
```diff
@@ -367,6 +367,7 @@
 -->
 <!--  <script src="iro.js"></script> NOTE: iro.js is loaded at the beginning of index.js -->
 <script src="rangetouch.js"></script>
+<script src="common.js"></script>
```
DedeHai: ok found the culprit: you cant add common.js here, it will inline it.

wled00/data/index.js
```diff
@@ -491,22 +491,6 @@ function restore(txt) {
 	return false;
 }
 
-async function loadPresets() {
```
DedeHai, line 494: dont remove it.

DedeHai (changes requested):

wled00/data/common.js
```diff
@@ -90,11 +90,11 @@ function loadJS(FILE_URL, async = true, preGetV = undefined, postGetV = undefine
 	scE.setAttribute("type", "text/javascript");
 	scE.setAttribute("async", async);
 	d.body.appendChild(scE);
-	// success event 
+	// success event
 	scE.addEventListener("load", () => {
 		//console.log("File loaded");
 		if (preGetV) preGetV();
-		GetV();
+		if (GetV) GetV();
```
DedeHai: getV() is always defined, its baked into wled code so this check is futile

wled00/data/common.js
```diff
@@ -126,6 +126,22 @@ function getLoc() {
 	}
 }
 function getURL(path) { return (loc ? locproto + "//" + locip : "") + path; }
+async function loadPresets(callback) {
```
DedeHai: the point of this being async is to await it to avoid multiple connections to the ESP which it does not handle well. if you await it (as you should) there is no point in having a callback.

DedeHai (changes requested):

wled00/data/index.js
benjamw: I'm confused why you asked me to move the loading method to the common.js file, but didn't want me to remove this method?

It's still just duplicating code. So if it's duplicating code, what is the reason for putting it in the common.js file? Why not just keep it together with the other methods in the HTML file?

I'm not pushing back, just wondering what the reason is so I can fit this all together nicely.

DedeHai: > adding the fallback is a edge case

common.js is not used in main UI, it inlines everything it needs so loading happens more smoothly. It is not out of question that it will do so in the future though. I did recently change all "fetches" to be sequential too as there were often 403 errors when free heap is running low and also added sequential loading of .css and .js files to all config pages so that could now also be used in main UI and some stuff could be moved to common.js potentially saving on code duplication and having a more "unified" approach between main UI and config pages.
The reason I asked you to put it in common.js is so that it _may_ be reused in the future if other pages need it. 

I looked at the main UI code and actually the cleanest way to do it is to move the preset sorting that index.js does to before it stores it into localstorage and always use that stored list: this way the presets are sorted the same way in the dropdown as the user chose for the UI. 
In case the local storage is missing, display a short warning to first go back to main UI. It was done this way for custom palettes (if you want to check: need to go back before I redid the cpal.htm as now it's doing it differently because cpal.h itself can change the list).
This completely gets rid of any code duplication and in case all these helper functions are moved into common.js in the future, fetching the list again is a one-line change.

wled00/data/settings_time.htm
```diff
@@ -19,13 +26,66 @@
...
+	function populatePresets(fromls) {
+		let json = null;
+		if (fromls) {
+			try {
+				json = JSON.parse(localStorage.getItem("wledP"));
+			} catch (e) {}
+		}
+
+		if (!json) {
+			showPresetWarning();
```
DedeHai: no need to use a function for a one liner

DedeHai:
the way I understand index.js the presets are stored in localstorage as an array, sorted by ID. The UI explicitly sorts it by name but only for display. So if you want to have it sorted by name as you initially did, that needs some rework. Otherwise I think your current implementation is ok to merge.

benjamw:
I'm fine with ID, the IDs are shown on the main page, so it's not an impossible task to find them

DedeHai:
@netmindz can you please take a quick look on this to check if anything jumps out? looks good to merge.

PR #5147, opened 2025-12-02
Replace #5138 with upstream NeoPixelBus fix

willmmiles:
[Makuna/Neopixelbus#894](https://github.com/Makuna/NeoPixelBus/pull/894) was merged!  Use that instead of a local shim.

## Release Notes

* **Refactor**
  * Updated NeoPixelBus library dependency to source from a specific upstream git commit instead of a fixed version number for improved version control.
  * Simplified ESP8266 hardware-specific DMA support by removing local workaround code and consolidating with upstream library implementations.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -160,7 +160,7 @@ lib_compat_mode = strict
 lib_deps =
     fastled/FastLED @ 3.6.0
     IRremoteESP8266 @ 2.8.2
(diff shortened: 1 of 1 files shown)


willmmiles:
Also the same bug was present on ESP32s using I2S ... now fixed.

PR #5153, opened 2025-12-02
add "awm" JSON command to set temporary auto white mode

DedeHai:
this allows presets or external commands to change the AWM on-the-fly. The value is applied temporary until revoked and not saved in config.

only active if global AWM is not enabled

Same logic as in config:
- 0 = 0 Manual
- 1 = 1 Brighter
- 2 = 2 Accurate
- 3 = 3 Dual
- 4 = Max
- 5 or larger = disabled

## Summary by CodeRabbit

* **New Features**
  * Added temporary auto-white mode override capability that can be controlled via API state updates.
  * Updated auto-white mode selection priority to consider global, temporary, and per-bus settings.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1000,7 +1000,8 @@ void Segment::refreshLightCapabilities() const {
         if (!strip.cctFromRgb && bus->hasCCT())                   capabilities |= SEG_CAPABILITY_CCT;
         if (strip.correctWB && (bus->hasRGB() || bus->hasCCT()))  capabilities |= SEG_CAPABILITY_CCT; //white balance correction (CCT slider)
         if (bus->hasWhite()) {
-          unsigned aWM = Bus::getGlobalAWMode() == AW_GLOBAL_DISABLED ? bus->getAutoWhiteMode() : Bus::getGlobalAWMode();
+          // get auto white mode: priority is global -> temp global -> per bus
+          unsigned aWM = Bus::getGlobalAWMode() == AW_GLOBAL_DISABLED? (Bus::getTempAWMode() == AW_GLOBAL_DISABLED ? bus->getAutoWhiteMode() : Bus::getTempAWMode()) : Bus::getGlobalAWMode();
           bool whiteSlider = (aWM == RGBW_MODE_DUAL || aWM == RGBW_MODE_MANUAL_ONLY); // white slider allowed
           // if auto white calculation from RGB is active (Accurate/Brighter), force RGB controls even if there are no RGB busses
           if (!whiteSlider) capabilities |= SEG_CAPABILITY_RGB;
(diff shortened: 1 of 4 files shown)


blazoncek:
I have read the user request but am not fond of the change. I would have not implemented this.

DedeHai:
why?

blazoncek:
No particular reason, apart from the fact that I do not like the idea of changing AWM using a bus "hack" (which is needed here, no objection). Perhaps it might be better to forgo `_gAWM` from configuration altogether and use it as an override as you tried to implement with temporary AWM. I do not see added value in `_gAWM` as a configuration item but can be used for the purpose requested.

However, I'm not sure that this functionality is really needed (brighter mode will wash out colors due to light-bleed, not make them "crisper and more vivid").

There's also a second aspect I'm considering regarding White channel. No effects use explicit W and none of the palettes employ W which means it is mostly unused except in Solid effect (and a handful of others) with particular color created using Hex input. However people with RGBW strips want W channel to be used and will most likely do so using Brighter or Accurate AWM modes (Dual will confuse them and None may not behave the way they want (except a few that know its purpose)).

Hence I would reduce AWM to Brighter and Accurate only.

DedeHai:
Those are good inputs, thanks. I do not know the history of white modes and only recently got a full WS2805 RGBCCT and it was fun but confusing to figure out the white modes, I do leave it at fully manual but I can understand people wanting auto modes, I agree that two are probably enough.
Regarding RGB and white: I use a solid color as a base set to black and play FX in a overlay segment with ADD, that way I can have white on top of all of the FX if I want to. 
I also noted that the copyFX will add in the white channel into the copy, even if the source segment has no white in it. Although it did not become clear to me from the code on why that is exactly (and yes, I am aware I wrote that code... ). 

If _gAWM has no real purpose except historical reasons, it would be better to remove it. Having the (temporary) global override is something that can be useful for presets, the implementation from this PR however has one flaw IMO: if it is used in one preset, you need to add it to all other presets too to disable it. So not really sure there is not a better implementation.

blazoncek:
> If _gAWM has no real purpose except historical reasons

I had a chat with a few of my users that do use RGBW (most of them since 0.12). Indeed all of them use `_gAWM` instead of bus AWM. It looks like it was passed over from installations without bus AWM and nobody bothered to update bus AWM so their configurations have bus AWM set to none, but global AWM to one of the desired values.

I have sketched the update that will move `_gAWM` into bus AWM if it does not exist, however I suspect those configurations already have bus AWM as it is unreasonable to expect configuration would not change since 0.12.
Since `_gAWM` always takes precedence over bus AWM it may be ok to force update of bus AWM if `_gAWM` exists in configuration.

blazoncek:
It would be prudent to invite @Aircoookie into discussion as global AWM was originally his [addition](https://github.com/wled/WLED/commit/3e6728fedb66383ef9881ba47933c327e95f159c).

softhack007:
I think it's still worth to keep this PR open - changing white mode with a preset sounds like a useful feature.

PR #5150, opened 2025-12-02
Add Gledopto Series With Ethernet

Aogu181:
Add Gledopto Series With Ethernet

## Summary by CodeRabbit

* **New Features**
  * Added support for Gledopto Series Ethernet devices: new Ethernet type, corresponding hardware mapping, and a "Gledopto Series With Ethernet" option in network settings.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #5154, opened 2025-12-02
Add support for the GLEDOPTO GL-C-618WL Ethernet Config

libschmudie-so:
The GLEDOPTO GL-C-618WL seems to have another unique Ethernet Pinout/Config.

This commit adds the support for this ethernet config and therefore makes the board work fully.

The device in question: https://gledopto.com/h-pd-64.html

## Summary by CodeRabbit

* **New Features**
  * Added support for GLEDOPTO GL-C-618WL Ethernet device with updated network configuration options.

### DIFF
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -366,21 +366,22 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
 #define BTN_TYPE_TOUCH_SWITCH     9
 
 //Ethernet board types
-#define WLED_NUM_ETH_TYPES        13
+#define WLED_NUM_ETH_TYPES        14
 
-#define WLED_ETH_NONE              0
-#define WLED_ETH_WT32_ETH01        1
-#define WLED_ETH_ESP32_POE         2
-#define WLED_ETH_WESP32            3
-#define WLED_ETH_QUINLED           4
-#define WLED_ETH_TWILIGHTLORD      5
-#define WLED_ETH_ESP32DEUX         6
-#define WLED_ETH_ESP32ETHKITVE     7
-#define WLED_ETH_QUINLED_OCTA      8
-#define WLED_ETH_ABCWLEDV43ETH     9
-#define WLED_ETH_SERG74           10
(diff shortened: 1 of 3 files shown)


Aogu181:
Hello, regarding the GLEDOPTO Ethernet series, in fact they not only have GL-C-618WL but also 616WL, so you should not choose GLEDOPTO GL-C-618WL as the model. For specific reference: https://github.com/wled/WLED/pull/5150

libschmudie-so:
@Aogu181 are you sure the 616WL is compatible? I tried to flash their firmware for the 616WL to my 618WL and ethernet did not work.

softhack007:
@libschmudie-so @Aogu181 it looks like both of you want to submit the same ethernet pin configs

Please get into direct contact with each other, and create _one_ PR for the new ethernet PIN configuration.

Due to flash size considerations, we do not accept "aliases" for boards where the name is different, but the pin configuration is already available, maybe just with a different brand name.

Aogu181:
> @Aogu181 are you sure the 616WL is compatible? I tried to flash their firmware for the 616WL to my 618WL and ethernet did not work.你确定616WL兼容吗？我试着将他们的616WL固件flash到我的618WL，但以太网不起作用。



> @Aogu181 are you sure the 616WL is compatible? I tried to flash their firmware for the 616WL to my 618WL and ethernet did not work.

I'm pretty sure the 616WL is compatible. The only difference between the 618WL and 616WL is that the 616WL has two outputs, while the 618WL has four. I've confirmed this with the engineers at GLEDOPTO. Have you tried it a few more times?

Aogu181:
> @libschmudie-so @Aogu181 it looks like both of you want to submit the same ethernet pin configs
> 
> Please get into direct contact with each other, and create _one_ PR for the new ethernet PIN configuration.
> 
> Due to flash size considerations, we do not accept "aliases" for boards where the name is different, but the pin configuration is already available, maybe just with a different brand name.

I understand your concerns, but the GLEDOPTO GL-C-618WL is actually part of the Gledopto Ethernet series. I will submit a new PR later.

PR #5161, opened 2025-12-06
Add JSON data handling over ESP-NOW in espNowReceiveCB

spectrenoir06:
I'm building a custom remote and I want to control WLED with it.
The WIZMote packet works great, but it isn’t easily customizable.

Add support for sending JSON API commands over ESP-NOW.
Packet size limits (250B v1 / 1470B v2) are sufficient for most cases, with optional fragmentation for larger commands.
Packets can be broadcast on multiple channels without duplicate processing.

For security, JSON parsing only happens when the sender (the remote) is present in the trusted device list

## Summary by CodeRabbit

* **New Features**
  * Added ESP-NOW JSON Handler Usermod to receive and process fragmented JSON messages over ESP-NOW. Features automatic fragment reassembly, message deduplication, and direct application of JSON payloads to WLED state. Enables remote control via ESP-NOW protocol. Requires ESP-NOW support to be enabled and sender MAC addresses to be linked as remotes.

### DIFF
--- a/usermods/espnow_json_handler/espnow_json_handler.cpp
+++ b/usermods/espnow_json_handler/espnow_json_handler.cpp
@@ -24,7 +24,7 @@ class EspNowJsonHandler : public Usermod {
 private:
   // Fragment reassembly state
     uint8_t lastMsgId = 0;
-    uint8_t lastProcessedMsgId = 0;
+    uint8_t lastProcessedMsgId = 255;
     uint8_t fragmentsReceived = 0;
     uint8_t* reassemblyBuffer = nullptr;
     size_t reassemblySize = 0;


DedeHai:
Sounds like a good addition. What I am not too sure about is to use "{" as an indicator and the lack of suporting longer JSONs. For a simle single command there is already the way to go through a remote.json file which does exactly what you want to achieve, just not in a dynamic way.

spectrenoir06:
remote.json is to remaps a wizmote button to a commande ( on the wled side ) right ?
I like the idea to have different remote who do different thing.
image

DedeHai:
> remote.json is to remaps a wizmote button to a commande ( on the wled side ) right ?

correct, you can have up to 256 button functions to map to any JSON command. 
See https://github.com/DedeHai/WLED-ESPNow-Remote for examples

spectrenoir06:
I already looked at your repo and really like it.
My remote is using the esp32-c6 ( because i want to play with zigbee / thread ), and is using 2 AAA battery ( with a power latch to have good battery ) 
for now both work pretty great the wizmote ( who is comment in my code ) + json
image
 
image

DedeHai:
the advantage of using remote.json is you can change remote functionality without changing the firmware on the remote. Disadvantage is that you need that file on each WLED device.

@blazoncek would such an implementation clash with any of your ESPNow work? What are your thoughts on this in general?

blazoncek:
For purposes like this (home made ESP-NOW remote) you have usermod callback which can be used for handling custom payloads. Write a usermod and it can be included.

If you have found readily available (generic?) remote that is sending JSON payloads (and preferably WLED's JSON API) then I would request to implement split packets as well. Otherwise this is enhancement just for (one) specific use.

Be mindful that 99% of WLED users are not electronics tinkerers and they will prefer to buy off the shelf remote (like Wizmote) which don't benefit from this enhancement.

However this is just my view on the topic.

DedeHai:
I do see some use-cases like writing custom sync sequence for art installations, though that could also be done using the remote.json. I am not opposed to adding this if its done properly. Its not much code for a feature some may find useful.

blazoncek:
> am not opposed to adding this if its done properly

I would require split packet processing as ~250 bytes is really very little for JSON.

> not much code for a feature some may find useful

I do agree.

spectrenoir06:
I'm not fan of the way UsermodManager::onEspNowMessage is done right now, because he don't check if the remote is in the whitelist.

One of the problem with the Wizmote, to send 1 pqt it send a espnow pqt to every channels.
Because the esp only listen on the same channel as the one use for the wifi.

I think be able to send a "small" pqt is important so a remote can do the same. ( spam all channel )

But yes I should also think about a way send the "cmd" in a multiple pqt ( probally only on one channel ), but both should be possible

blazoncek:
> he don't check if the remote is in the whitelist.

You can do that in usermod.

> One of the problem with the Wizmote

That is not a problem. That's a benefit.

spectrenoir06:
- Moved the code into a usermod
- Added support for receiving fragmented packets (removing the 250-byte limit)
- Allows sending the same packet on all channels when the channel is unknown (e.g. WiZ Mote)

usermods/espnow_json_handler/espnow_json_handler.cpp
```diff
@@ -0,0 +1,201 @@
...
+        // Check if we have all fragments
+        if (fragmentsReceived >= totalFragments) {
+            reassemblyBuffer[reassemblySize] = '\0'; // Null terminate
+            DEBUG_PRINTF_P(PSTR("ESP-NOW complete message reassembled (%d bytes): %s\n"), reassemblySize, (char*)reassemblyBuffer);
+
+            // Mark this message as processed for deduplication
+            lastProcessedMsgId = messageId;
+
+            // Process the complete JSON message
+            if (requestJSONBufferLock(18)) {
```
DedeHai, line 169: should be JSON_LOCK_REMOTE

usermods/espnow_json_handler/espnow_json_handler.cpp
```diff
@@ -0,0 +1,199 @@
...
+        // Allocate or reallocate buffer
+        size_t fragmentDataSize = len - ESPNOW_FRAGMENT_HEADER_SIZE;
+        size_t newSize = reassemblySize + fragmentDataSize;
+
+        uint8_t* newBuffer = (uint8_t*)realloc(reassemblyBuffer, newSize + 1); // +1 for null terminator
+        if (!newBuffer) {
+            DEBUG_PRINTLN(F("ESP-NOW fragment reassembly: memory allocation failed"));
+            cleanupReassembly();
+            return true; // Handled by failing gracefully
+        }
```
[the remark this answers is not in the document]
DedeHai: I agree on this, cap the mem usage

usermods/espnow_json_handler/readme.md
```diff
@@ -0,0 +1,42 @@
...
+- Reassembles fragments into complete JSON payloads
+- Deduplication to prevent processing duplicate messages when  broadcasting on multiple channels
+
+## Fragment Protocol
+
+Messages are fragmented with a 3-byte header:
+
+| Byte | Description |
+|------|-------------|
+| 0 | Message ID (unique identifier for reassembly) |
```
DedeHai, line 17: this header is very fragile. if other message types are added in the future, it will break. add at least one or two bytes of unique identifiers.

usermods/espnow_json_handler/espnow_json_handler.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+      // Need at least header size to process
+        if (len < ESPNOW_FRAGMENT_HEADER_SIZE) {
+            return false; // Not handled
+        }
+
+        // Check if this looks like a fragmented JSON message
+        // First byte should be a reasonable message ID (not a WiZ Mote signature or 'W' sync packet)
+        if (payload[0] == 0x91 || payload[0] == 0x81 || payload[0] == 0x80 || payload[0] == 'W') {
```
DedeHai, line 100: what if your ID is one of these?

PR #5170, opened 2025-12-09
Add files via upload

Yannicflight:
Added support for LSC smart connect ledstrip which is sold at Action shops around Europe. This LEDstrip uses dual SM16703 drivers, where the first driver is used for GRB and the second driver drives CW, WW LED and the last channel is not used.

So far there was no support for this in WLED so I made this modification which now introduces SMT16703 RGB+CCT (2X) as an option in the dropdown menu. I build the firmware, tested it and found working with the ledstrip.


Yannicflight:
h

Issue #5171, opened 2025-12-09
WLED 0.14.4 build fails (ESPAsyncWebServer 2.2.1 does not exist)


booherbg:
### What happened?

Cloning the WLED repo and running the build with PlatformIO fails by default when using the v0.14.4 branch:
```
...
git+https://github.com/Aircoookie/ESPAsyncWebServer.git @ 2.2.1 git version 2.41.0 Cloning into '/home/vscode/.platformio/.cache/tmp/pkg-installing-bkgi6sky'... remote: Enumerating objects: 68, done. remote: Counting objects: 100% (68/68), done. remote: Compressing objects: 100% (60/60), done. remote: Total 68 (delta 4), reused 42 (delta 0), pack-reused 0 (from 0) Receiving objects: 100% (68/68), 284.25 KiB | 2.45 MiB/s, done. Resolving deltas: 100% (4/4), done. PackageException: 
Package version 2.4.2+sha.dbb7c33 doesn't satisfy requirements 2.2.1 based on PackageMetadata <type=library name=ESPAsyncWebServerWLED version=2.4.2+sha.dbb7c33 spec={'owner': None, 'id': None, 'name': 'ESPAsyncWebServer', 'requirements': '2.2.1', 'uri': 'git+https://github.com/Aircoookie/ESPAsyncWebServer.git'}

```

Bumping the version from 2.2.1 to 2.4.x fixes the problem. I'll submit a pull request if this is an acceptable fix.

### To Reproduce Bug

Clone the repo, check out the v0.14.4 branch, and build the project using PlatformIO as per the recommended instructions.

### Expected Behavior

I expect the build to work by default.

### Install Method

Self-Compiled

### What version of WLED?

WLED 0.14.4 latest hash

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell
[...]

PR #5173, opened 2025-12-09
Bumping ESPAsyncWebServer.git version to 2.4.2 to fix failing build

booherbg:
Addresses the issue reported in #5171 ; this is based on the `v0.14.4` tag which does not have a corresponding branch, so this PR would need to be cherry-picked by the upstream maintainers I think, or similar.

## Summary by CodeRabbit

* **Chores**
  * Updated how the ESPAsyncWebServer dependency is referenced in build configuration; the resolved version remains 2.2.1. No changes to runtime behavior, API, error handling, or other dependencies.
  * No end-user visible functional changes; builds/configuration bookkeeping only.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -180,7 +180,7 @@ lib_deps =
     fastled/FastLED @ 3.6.0
     IRremoteESP8266 @ 2.8.2
     makuna/NeoPixelBus @ 2.7.5
-    https://github.com/Aircoookie/ESPAsyncWebServer.git @ 2.2.1
+    https://github.com/Aircoookie/ESPAsyncWebServer.git#v2.2.1
   #For use of the TTGO T-Display ESP32 Module with integrated TFT display uncomment the following line
     #TFT_eSPI
   #For compatible OLED display uncomment following


blazoncek:
Use #v2.2.1 tag instead of @2.2.1

blazoncek:
Never force push!

booherbg:
> @ syntax is incompatible with git URLs

I found it to be more specific than that - `@ 2.4.2` actually worked where `@ 2.2.1` must have stopped working at some point. My guess is that something changed relating to how the dep is registered with PlatformIO (v2.4.2 vs v2.2.1)? Either way - config is now dropped back to `#2.2.1` using proper git tag syntax; the commit attached to this PR is updated

Will defer back to maintainers as to potential next steps. 

## Questions about the contribution process

From a process standpoint - and because I'm genuinely curious (and a huge fan of WLED in general) - hypothetically what would be the process to merge this change back into v0.14.x? Would it be a full semantic versioning patch bump to 0.14.5?

To (hypothetically) implement this, I imagine making a branch from that tag, applying this PR as a cherry-picked commit, then updating the `v0.14.4` or `v0.14.5` tag to point to the new branch would be one way to go. 

Is there a different way that does not involve creating new branches? I ask because I don't see version related branches and thus became curious. Thank you!

booherbg:
> Never force push!

Hah good call - regrettable in hindsight sorry about that

blazoncek:
There is no way "to merge this back". And downgrading AWS is out of the question.
If you use 0.14 from the last known commit (or tag) just update local copy. It will not change.

booherbg:
> There is no way "to merge this back". And downgrading AWS is out of the question. If you use 0.14 from the last known commit (or tag) just update local copy. It will not change.

Sounds good, thank you. It builds successfully with v2.4.2 I'm just not sure what side effects this introduces. Hopefully this PR can serve as documentation in case anyone else is interested in compiling 0.14.x locally. The "[WLED Online compiler](https://wled-compile.github.io/)" doesn't seem to have any problems with building this version.

netmindz:
I have created a new tag based off  0.14.4 that includes the fix

booherbg:
Awesome, thank you.

Issue #3771, opened 2024-02-22
AutoSave mod doesn't autosave


markg85:
### What happened?

Installed the AutoSave usermod with these flags:
```
-D USERMOD_AUTO_SAVE -D AUTOSAVE_AFTER_SEC=10 -D AUTOSAVE_PRESET_NUM=100 -D USERMOD_AUTO_SAVE_ON_BOOT=true
```

Relevant screenshots:
(presets section on the main page)
![image](https://github.com/Aircoookie/WLED/assets/49061/4f901690-b76d-4faf-be1c-4f1668d3b600)

(Relevant section from "LED Preferences")
![image](https://github.com/Aircoookie/WLED/assets/49061/19205256-3f00-4fff-a84c-df26456fd60f)

(usermod)
![image](https://github.com/Aircoookie/WLED/assets/49061/fe7222f0-73a9-4536-9a02-594bd29c45d4)

Am i missing a setting somewhere?

### To Reproduce Bug

Install the usermod with these flags:
```
-D USERMOD_AUTO_SAVE -D AUTOSAVE_AFTER_SEC=10 -D AUTOSAVE_PRESET_NUM=100 -D USERMOD_AUTO_SAVE_ON_BOOT=true
```

Create a preset with id 100 and some effect
Change your effect and wait for at least `AUTOSAVE_AFTER_SEC=10` (10 seconds).
Reboot.

You now get your initial effect in id 100 where i was expecting autosave kick in and **save** the change i made automatically.

### Expected Behavior

Preset loads at boot.

### Install Method

Self-Compiled

### What version of WLED?

WLED 0.14.2-b1 (build 2402120)

### Which microcontroller/board are you seeing the problem on?

ESP32-S3

### Anything else?

I did compile from source so this is the current git version as of the moment of this post.

AutoSave is freakingly annoying to setup. It's completely unclear to me if i already need to have a preset to which i "tie" autosave. If i don't have a preset then i get a weirdly named one created by autosave. I did try that path and that only saved my initial settings. It didn't update - autosave - any changes i made.

In the compile flags i explicitly said `-D AUTOSAVE_PRESET_NUM=100` which is ignored in the usermod gui. It's default is 250. Therefore i'm not at all sure if what i set in the compile flags is even properly used in autosave. Clearly the flags are in the right place else the whole usermod wouldn't be visible (compiled in).

[...]

PR #4992, opened 2025-10-08
Missing define examples and new define options

KrX3D:
Hi,

 i added some missing define examples to platformio_override.sample.ini
The defines where present but no info was there how to use them, i also added the location of the option on the webpage like:

Settings -> Wifi Setup -> Network name (SSID, empty to not connect)

for a better understanding what this define is for.

And together with that i added a couple more things that can be defined. Mostly the one i currently use,  since i got an device i bought that keeps resetting to the compiled defaults... ,  but i would like to add more in future when i got time (not in this PR)

The defines can also be added to the my_config file like i did, here is an list of the new and already present once added with the PR.

```
// Device identity / User Interface
#define SERVERNAME                 "ClientXYZ"                     // Settings -> User Interface -> Server description

// Settings -> Wifi Setup
#define CLIENT_SSID                ""                              // Settings -> Wifi Setup -> Network name (SSID, empty to not connect)
#define CLIENT_PASS                ""                              // Settings -> Wifi Setup -> Network password
#define WIFI_DEFAULT_NO_WIFI_SLEEP true                            // Settings -> Wifi Setup -> Experimental -> Disable WiFi sleep
#define MDNS_NAME                  "wled-ClientXYZ"                         // Settings -> Wifi Setup -> mDNS address (leave empty for no mDNS)
#define WLED_AP_SSID               "WLED-ClientXYZ-AP"              // Settings -> Wifi Setup -> Configure Access Point -> AP SSID (leave empty for no AP)
#define WLED_AP_PASS               ""                              // Settings -> Wifi Setup -> Configure Access Point -> AP password (leave empty for open)

// Settings -> Security & Update
#define WLED_ENABLE_AOTA                                        // Settings -> Security & Update -> Software Update -> Enable ArduinoOTA
#define WLED_DISABLE_OTA                                        // Settings -> Security & Update -> Software Update -> OTA
#define OTA_DEFAULT_SAME_SUBNET    false                         // Settings -> Security & Update -> Software Update -> Only allow update from same network/WiFi

// Settings -> Time & Macros
#define WLED_NTP_ENABLED          true                           // Settings -> Time & Macros -> Time setup -> Get time from NTP server
#define WLED_NTP_SERVER           "192.168.178.11"               // Settings -> Time & Macros -> Time setup -> NTP Server
#define WLED_NTP_AMPM             true                           // Settings -> Time & Macros -> Time setup -> Use 24h format
[...]

PR #5175, opened 2025-12-09
Add option to save unmodified presets to autosave usermod

mtin:
The `usermod_v2_auto_save` deliberately will not save the current state if it is a preset without modifications. I think this totally depends on the use case and should be configurable.

My use case:
> Multiple presets that can be cycled through via a button. Whenever power is cut and restored, it should be in the state as it was before.

So I added an option `AUTOSAVE_IGNORE_PRESETS` which defaults to `true` which is the previous behavior.

Screenshot 2025-12-09 at 18 24 40

I'd like to argue that a default of `false` (meaning: also autosave if it is an already saved preset) might make more sense for the majority, but ofc that is an opinionated guess

I personally was confused why it would not save when cycling through my presets, could not find documentation on it - and ended up being educated by @softhack007 at https://github.com/wled/WLED/issues/3771, thanks for that!
> if you're running a preset without changes, then autosave will not save the settings (reason: you're running an unchanged preset and we don't want to make duplicates)

## Summary by CodeRabbit

* **New Features**
  * Added a user-facing setting to control whether auto-save ignores preset changes; default preserves previous behavior. The setting is persisted across saves/loads.
* **Bug Fixes**
  * Auto-save now correctly skips saves in scenarios that previously produced duplicate presets when the ignore-presets option is enabled.

### DIFF
--- a/usermods/usermod_v2_auto_save/usermod_v2_auto_save.cpp
+++ b/usermods/usermod_v2_auto_save/usermod_v2_auto_save.cpp
@@ -18,7 +18,7 @@
 //
 // By default it will not save the state if an unmodified preset
 // is selected (to not duplicate it). You can change this behaviour
-// by setting AUTOSAVE_IGNORE_PRESETS=false
+// by setting autoSaveIgnorePresets=false
 //
 // AutoSaveUsermod is standalone, but if FourLineDisplayUsermod 
 // is installed, it will notify the user of the saved changes.
--- a/usermods/usermod_v2_auto_save/usermod_v2_auto_save.cpp
+++ b/usermods/usermod_v2_auto_save/usermod_v2_auto_save.cpp
@@ -134,6 +134,7 @@ class AutoSaveUsermod : public Usermod {
       static unsigned long lastRun = 0;
       unsigned long now = millis();
       if (!autoSaveAfterSec || !enabled || (autoSaveIgnorePresets && currentPreset>0) || (strip.isUpdating() && now - lastRun < 240)) return;  // setting 0 as autosave seconds disables autosave
+      lastRun = now;
       uint8_t currentMode = strip.getMainSegment().mode;
       uint8_t currentPalette = strip.getMainSegment().palette;
 
--- a/usermods/usermod_v2_auto_save/usermod_v2_auto_save.cpp
+++ b/usermods/usermod_v2_auto_save/usermod_v2_auto_save.cpp
@@ -16,6 +16,10 @@
 // It can be configured to load auto saved preset at startup,
 // during the first `loop()`.
 //
+// By default it will not save the state if an unmodified preset
+// is selected (to not duplicate it). You can change this behaviour
+// by setting AUTOSAVE_IGNORE_PRESETS=false
+//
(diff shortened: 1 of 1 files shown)


usermods/usermod_v2_auto_save/usermod_v2_auto_save.cpp
```diff
@@ -49,6 +53,12 @@ class AutoSaveUsermod : public Usermod {
     bool applyAutoSaveOnBoot = false;     // do we load auto-saved preset on boot?
     #endif
 
+    #ifdef AUTOSAVE_IGNORE_PRESETS
+    bool autoSaveIgnorePresets = AUTOSAVE_IGNORE_PRESETS;
+    #else
+    bool autoSaveIgnorePresets = true;     // ignore by default to not duplicate presets
+    #endif
```
willmmiles: Please remove the #ifdef here -- we're trying to get out of the business of using #defines to set configuration defaults.  See #4992 for more discussion.  If you're doing a custom build and you want to change your own default, edit the code locally in your build environment.

willmmiles (changes requested):
Apart from the build time change, I think this is a reasonable feature to add.

usermods/usermod_v2_auto_save/usermod_v2_auto_save.cpp
```diff
@@ -122,7 +129,8 @@ class AutoSaveUsermod : public Usermod {
     void loop() {
       static unsigned long lastRun = 0;
       unsigned long now = millis();
-      if (!autoSaveAfterSec || !enabled || currentPreset>0 || (strip.isUpdating() && now - lastRun < 240)) return;  // setting 0 as autosave seconds disables autosave
+      if (!autoSaveAfterSec || !enabled || (autoSaveIgnorePresets && currentPreset>0) || (strip.isUpdating() && now - lastRun < 240)) return;  // setting 0 as autosave seconds disables autosave
+      lastRun = now;
```
mtin, line 133: While reviewing this again, I stumbled over `lastRun` never being set!? I think the intention probably was this; fine to add this in this MR as well? This got introduced in https://github.com/wled/WLED/commit/0ff4016250dd8235e6465e62d72ab99ec3df19b6#diff-e57385253fd04c5e758584aeb14c1398a2ec9febfbb230c376b406b4aaa93ffc

And, sorry, rusty cpp: should `lastRun` be a `static` scoped variable here? why not a private member variable?

softhack007: @mtin I think your fix is correct, "lastRun" implements the standard "be nice but not too nice" logic that enforces running the loop() even under high load (`strip.isUdating()`)
A static variable in a function is different from a private class attribute, I think it was chosen here because we needed a persistent value without polluting global namespace.

So not necessary to make it a private member, because we will never have more than one instance of this usermod.

mtin:
heyho @willmmiles, I did the edits you requested, you think we can move forward with this? 😇

willmmiles (approved):

softhack007:
possibly related to
* #5389

PR #5181, opened 2025-12-11
Fix 5168

willmmiles:
Correct the AI code from #5168.

- Remove redundant "psramPresent" property -- if "psramSize" is present, then it must be true
- Actually report if PSRAM is used
- Shorten strings
- Correct PSRAM size estimate

## Summary by CodeRabbit

* **Bug Fixes**
  * PSRAM reporting refined: system now reports free PSRAM in bytes and a rounded total PSRAM size; presence is reported only when PSRAM support is actually available.
  * JSON buffer allocation now falls back to onboard RAM when PSRAM is unavailable, avoiding incorrect assumptions about memory and improving reliability.

### DIFF
--- a/wled00/data/index.js
+++ b/wled00/data/index.js
@@ -3443,10 +3443,7 @@ function reportUpgradeEvent(info, oldVersion) {
 			};
 
 			// Add optional fields if available
-			if (infoData.psrSz !== undefined) {
-				upgradeData.psramSize = infoData.psrSz;  // Total PSRAM size in MB
-				upgradeData.psramUsed = (infoData.psram !== undefined);
-			}
+			if (infoData.psrSz !== undefined) upgradeData.psramSize = infoData.psrSz;  // Total PSRAM size in MB; can be 0
 
 			// Note: partitionSizes not currently available in /json/info endpoint
 
--- a/wled00/json.cpp
+++ b/wled00/json.cpp
@@ -839,15 +839,12 @@ void serializeInfo(JsonObject root)
 #endif
 
   root[F("freeheap")] = getFreeHeapSize();
-  #ifdef ARDUINO_ARCH_ESP32
+  #if defined(ARDUINO_ARCH_ESP32) && defined(BOARD_HAS_PSRAM)
   // Report PSRAM information
-  size_t psramSize = ESP.getPsramSize();  // returns 0 if no psram
-  if (psramSize > 0) {
-    #if defined(BOARD_HAS_PSRAM)
-    root[F("psram")] = ESP.getFreePsram(); // Free PSRAM in bytes (backward compatibility)
-    #endif
-    root[F("psrSz")] = 1 + ((psramSize - 1) / (1024U * 1024U)); // Total PSRAM size in MB, round up to correct for allocator overhead
-  }
+  // Free PSRAM in bytes (backward compatibility)
+  root[F("psram")] = ESP.getFreePsram(); 
+  // Total PSRAM size in MB, round up to correct for allocator overhead
+  root[F("psrSz")] = (ESP.getPsramSize() + (1024U * 1024U - 1)) / (1024U * 1024U); 
   #endif
   root[F("uptime")] = millis()/1000 + rolloverMillis*4294967;
 


wled00/data/index.js
```diff
@@ -3443,8 +3443,11 @@ function reportUpgradeEvent(info, oldVersion) {
 			};
 
 			// Add optional fields if available
-			if (infoData.psramPresent !== undefined) upgradeData.psramPresent = infoData.psramPresent;  // Whether device has PSRAM
-			if (infoData.psramSize !== undefined) upgradeData.psramSize = infoData.psramSize;  // Total PSRAM size in MB
+			if (infoData.psrSz !== undefined) {
+				upgradeData.psramSize = infoData.psrSz;  // Total PSRAM size in MB
+				upgradeData.psramUsed = (infoData.psram !== undefined);
+			}
+
```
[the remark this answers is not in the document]
willmmiles: Reporting that inconsistency is the very purpose of this code.  If a firmware without BOARD_HAS_PSRAM is installed on a device that has PSRAM, we'd like to see that in the report.

wled00/json.cpp
```diff
@@ -841,13 +841,12 @@ void serializeInfo(JsonObject root)
   root[F("freeheap")] = getFreeHeapSize();
   #ifdef ARDUINO_ARCH_ESP32
   // Report PSRAM information
-  bool hasPsram = psramFound();
-  root[F("psramPresent")] = hasPsram;
-  if (hasPsram) {
+  size_t psramSize = ESP.getPsramSize();  // returns 0 if no psram
+  if (psramSize > 0) {
     #if defined(BOARD_HAS_PSRAM)
```
DedeHai: can the BOARD_HAS_PSRAM not be used to comment out this whole block? AFAIK this has to be defined for PSRAM to be used, at least that is what I found at one point but don't remember what the current state of code is with all those pending PRs but I fixed that up at one point.

willmmiles: OK - I've done a more thorough review and I think you're correct about this.   If BOARD_HAS_PSRAM is not set, then the Arduino core shouldn't try to initialize that subsystem.  I had thought that ESP-IDF might be doing so anyways, but after reviewing the Arduino core code, I don't think that's the case.   The upshot is that the code will work as is -- without `BOARD_HAS_PSRAM`, `ESP.getPsramSize()` is guaranteed to return 0 -- but we could lift the conditional compilation around the whole block and save a few bytes of code there.

There is a minor issue in the current code that if you try to use a firmware compiled with BOARD_HAS_PSRAM on a board that doesn't actually have any, it'll crash at boot when it fails to allocate the JSON buffer.  I also have a small patch for that if it isn't already fixed in your memory PRs.  (Hoping to test those soon!)

Also this means that we cannot detect a case where PSRAM is present but unused.  If BOARD_HAS_PSRAM was missing, the hardware is never initialized, and the size always returns 0.  If BOARD_HAS_PSRAM is defined, we will always try to use it.  So the only three cases we can report are:

- BOARD_HAS_PSRAM defined and PSRAM present
- BOARD_HAS_PSRAM defined and PSRAM not present 
- BOARD_HAS_PSRAM was not defined

I will update the code to reflect this.

DedeHai: thanks for confirming, I did dive down the same rabbit hole a few months back and updated some of the code after that but memory is hazy...

wled00/json.cpp
```diff
@@ -841,13 +841,12 @@ void serializeInfo(JsonObject root)
...
     root[F("psram")] = ESP.getFreePsram(); // Free PSRAM in bytes (backward compatibility)
     #endif
-    root[F("psramSize")] = ESP.getPsramSize() / (1024UL * 1024UL); // Total PSRAM size in MB
+    root[F("psrSz")] = 1 + ((psramSize - 1) / (1024U * 1024U)); // Total PSRAM size in MB, round up to correct for allocator overhead
```
netmindz: I really don't think we should be sacrificing readability to save save the handful of bytes

willmmiles: I don't feel strongly about it either way, but it was suggested in the other PR.  Being terse to a fault is definitely consistent with the other API names, but I'd be happy to strike a different direction too.  @DedeHai what do you think?

willmmiles:
Updated logic:

BOARD_HAS_PSRAM defined and PSRAM present: reports "psramSize" > 0
BOARD_HAS_PSRAM defined and PSRAM not present: reports "psramSize" = 0
BOARD_HAS_PSRAM was not defined: does not report "psramSize"

I've added the little patch to allow a psram build to work on hardware with no psram connected -- note that it still does not allow those pins to be re-used, though.  I think that's probably reasonable given that the platform is going to try probing them for psram on every boot.

wled00/json.cpp
DedeHai: > I really don't think we should be sacrificing readability to save save the handful of bytes

each variable string passed between UI and firmware is stored twice (or more), if all of them have long readable names it adds up quickly to several kB. Most others are just two letters for this reason. A comment can clear up the meaning.

wled00/json.cpp
netmindz: If I'm understanding this correctly, are we saying that it's actually impossible to detect if there is PSRAM without the BOARD_HAS_PSRAM included in the build, so as most users are using our generic esp32dev, they will always return false?

However if we fixed the issues where things crash if you build with that flag on a device without then we could have a generic esp32dev build that handled both?

wled00/json.cpp
netmindz: I've not counted them up, but from a quick scan, I would say that most of the values in the info response are correctly spelt words or common abbreviation, e.g repo for repository, which is why this particular abbreviation jars with me

netmindz: Also, as far as the JavaScript goes, as we gzip then trying to abbreviate you might end up using more space if you say use a short version for a variable name but the full form in a message

DedeHai: I would keep strings as short as possible in general, it's also limited JSON buffer size. You are correct about the gzip of js but in C code it is not zipped, nor is it in JSON output.
I do agree that trying to read/modify index.js is a PITA with all those short names but it generally helps: as an experiment I at one point had claude.ai re-write the cpal.htm file with shorter globals and function names, immediately made it 500 bytes smaller after gzip.
Maybe something we can discuss in the upcoming meeting. What bites me in this matter is that we try to keep C code lean and try to save bytes wherever possible and then hand out flash space for strings that are not intended for user readability.

wled00/json.cpp
willmmiles: > If I'm understanding this correctly, are we saying that it's actually impossible to detect if there is PSRAM without the BOARD_HAS_PSRAM included in the build, so as most users are using our generic esp32dev, they will always return false?

That's correct.


> However if we fixed the issues where things crash if you build with that flag on a device without then we could have a generic esp32dev build that handled both?

Yes, though with two caveats:
- Non-psram boards would not be able to use the PSRAM IO pins anymore (16 and 17, IIRC): the platform will always try and check for PSRAM on every boot.  That's probably not safe in the general case, depending on what's connected; and pin 16 happens to be the default LED pin for the current `esp32dev` build.  It would likely break a *lot* of setups.
- The `esp32_wrover` build with PSRAM support uses the `-mfix-esp32-psram-cache-issue` flag to enable a workaround for an MMU bug on the v1 chips.  This comes at a nontrivial flash space cost (and likely a performance hit too, though I haven't benchmarked it).  We'd have to consider how to manage that -- right now we slow all WROVERs down to make sure the firmware is always compatible, but is that an acceptable loss for *every* ESP32?

I did try both ways on my test setup (sending the `esp32dev_wrover` build to a baseline ESP32, and vice-versa) and it does boot, with c86d09c -- that's how I generated the "0mb" report during testing.

Given those caveats, though, I'm not confident that a unified build is worth the cost.

netmindz: Sounds like we need to ditch the Boolean then if we can't actually get the value. 

When I get a moment I'll review the hub75 pin definitions and see if any of them are using 16 or 17 and if not then we can at least ensure PSRAM is enabled for those specialist bins

DedeHai (approved):
I did not run any tests, but code looks good.

wled00/json.cpp
netmindz: Did you see this comment @DedeHai

DedeHai: I did, any action required from my side? I am not too familiar with PSRAM and its implications to builds/pin use, all I remember was that there was some discrepancy between `BOARD_HAS_PSRAM` and how PSRAM was actually checked and IIRC I cleaned that up a little. Not 100% sure anymore but I think `BOARD_HAS_PSRAM` is required by Arduino calls to even compile the logic to check it.

netmindz: Not necessarily action on your part, I was just surprised by your comment approving, while this comment thread was unresolved

willmmiles: I was under the impression this was resolved too - what did I miss?  The boolean has already been removed in 8ec94af.  I was expecting any changes to Hub75 envs to be out of scope for this PR.

netmindz: Yes this code removes the need for a separate Boolean in the JSON info in favour of a undefined check on the size to dictate the value of the Boolean send to the usage server, but if it's always going to be undefined for builds that don't have PSRAM enabled in the software, then this is failing to achieve the key objective, which was to understand what % of device has PSRAM hardware that was potentially going unused due to the choice of firmware.

DedeHai: >I think BOARD_HAS_PSRAM is required by Arduino calls to even compile the logic to check

would first need to confirm this but if that is the case, can't check for PSRAM if it is not compiled in.

willmmiles: > if it's always going to be undefined for builds that don't have PSRAM enabled in the software, then this is failing to achieve the key objective, which was to understand what % of device has PSRAM hardware that was potentially going unused due to the choice of firmware.

Ah, ok - sorry I wasn't clear earlier.  That is impossible.  There is no way to check for PSRAM unless `BOARD_HAS_PSRAM` was compiled in.

PR #5187, opened 2025-12-12
Remove gamma value lower limit to enable inverse gamma correction

flo269:
## Summary
Remove the artificial restriction that prevents gamma values below 1.0, enabling inverse gamma correction.

## Motivation
LEDs can appear too dark at low brightness levels. Inverse gamma correction (gamma < 1.0) is needed to brighten the lower end of the brightness curve and make dimmed LEDs more visible.

Currently, gamma values are restricted to > 1.0, which only allows standard gamma correction. This PR removes that limitation.

## Changes
- Update validation in `set.cpp` and `cfg.cpp` to allow gamma range 0.1-3.0 (was 1.0-3.0)
- Update HTML form validation in `settings_leds.htm` to match (min="0.1")

## Technical Details
The `NeoGammaWLEDMethod::calcGammaTable()` function already supports any gamma value mathematically. This PR simply removes an artificial restriction that prevented values < 1.0.

## Testing
Tested on v0.15.3 and latest main - inverse gamma values (< 1.0) now work correctly and improve visibility at low brightness levels.

## Summary by CodeRabbit

* **Bug Fixes**
  * Expanded valid gamma correction range to accept values from 0.1 to 3.0; values between 0.1 and 1.0 are now preserved rather than being reset.
  * Updated the gamma value input to allow a minimum of 0.1 and adjusted its placeholder to reflect the finer calibration range.

### DIFF
--- a/wled00/data/settings_leds.htm
+++ b/wled00/data/settings_leds.htm
@@ -958,7 +958,7 @@ Swap: <select id="xw${s}" name="XW${s}">
 		Apply preset <input name="BP" type="number" class="m" min="0" max="250" required> at boot (0 uses values from above)<br><br>
 		Use Gamma correction for color: <input type="checkbox" name="GC"> (strongly recommended)<br>
 		Use Gamma correction for brightness: <input type="checkbox" name="GB"> (not recommended)<br>
-		Use Gamma value: <input name="GV" type="number" class="m" placeholder="2.8" min="0.1" max="3" step="0.1" required><br><br>
+		Use Gamma value: <input name="GV" type="number" class="m" placeholder="2.2" min="0.1" max="3" step="0.1" required><br><br>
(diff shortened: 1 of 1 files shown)


DedeHai:
Did you check both gamma functions create correct tables for the new range?

DedeHai:
I checked the functions, they do work still.
you also need to update the other lines in the config so it works after a reboot.

wled00/cfg.cpp
DedeHai: update this

flo269:
I think I have it figured out - updated lines 511 and 513 to check != 1.0f instead of > 1.0f so gamma correction is enabled for any non-linear value (both below and above 1.0). This should fix the persistence issue after reboot.

tablatronix:
Any more testing needed for this? Seems like a low impact change

PR #5190, opened 2025-12-14
platformio sample cleanup

netmindz:
Remove bespoke boards from platformio-override.sample.ini as these are not maintained

## Summary by CodeRabbit

* **Chores**
  * Removed configuration entries for several legacy and specialized microcontroller boards from the sample build configuration, streamlining available setup options.
* **Documentation**
  * Updated wording in a sample environment header to improve clarity and tone without changing its configuration.


netmindz:
I've left Athom for now as I know these are in common use and I don't know the current state of play with regards to their firmware

DedeHai (approved):

platformio_override.sample.ini
```diff
@@ -471,65 +417,6 @@ build_unflags = ${common.build_unflags}
...
-  -D TFT_HEIGHT=240
-  -D CGRAM_OFFSET
-  -D TFT_SDA_READ
-  -D TFT_MOSI=23
-  -D TFT_SCLK=18
-  -D TFT_DC=25
-  -D TFT_RST=26
-  -D SPI_FREQUENCY=40000000
-  -D USER_SETUP_LOADED
-monitor_filters = esp32_exception_decoder
```
willmmiles, line 531: This is a usermod example that's in the wrong place - probably a bad merge moved the header comment.  I think it should get moved to the EleksTube_IPS usermod readme (and we should do the same with the RF433 example too)

willmmiles (changes requested):

platformio_override.sample.ini
```diff
@@ -260,8 +243,7 @@ board_build.f_flash = 80000000L
 board_build.flash_mode = qio
 
 [env:esp32dev_V4_dio80]
-;; experimental ESP32 env using ESP-IDF V4.4.x
-;; Warning: this build environment is not stable!!
+;; ESP32 env using ESP-IDF V4.4.x
 ;; please erase your device before installing.
 extends = esp32_idf_V4  # based on newer "esp-idf V4" platform environment
```
softhack007, line 248: all buildenvs are "V4" now 
-> replace with ``extends = esp32dev``
--> remove everything else in this buildenv, only keep ``board_build.f_flash = 80000000L`` and ``board_build.flash_mode = dio``

DedeHai:
https://github.com/wled/WLED/issues/5643

i.e. remove all `-D USERMOD_x` and replace with correct syntax

PR #5192, opened 2025-12-14
set unused segments to black

DedeHai:
0.15.1 introduced a new behaviour on how "empty space" is treated. While it is technically correct, it is not user friendly and is a breaking change: all previous versions behaved differently and so does 0.16.

What this change does is to set any unused pixels to black instead of keeping the last value.


softhack007 (approved):
looks reasonable. I agree its better to set the previous segment to "black", so to avoid orphaned pixels that stay on the LEDs until the bus driver gets re-initialized.

willmmiles (approved):

PR #3195, opened 2023-05-03
Add support for WPA-Enterprise

Chasethechicken:
Adds options to the WiFi settings, to enable connecting to an enterprise WiFi.

<details>
  <summary>Screenshot WiFi Settings</summary>
  Screenshot WiFi Settings
</details>

WPA-Enterprise on ESP32 needs *[env:esp32dev_V4_dio80]* for a recent Arduino Core, so is disabled for other ESP32 targets.

*Tested using an ESP32. Should also work for ESP8266.*

Currently only supports _Identity_ and _Password_ on ESP8266 and _Anonymous Identity_, _Identity_ and _Password_ on ESP32.

Thanks to 
- https://github.com/martinius96/ESP32-eduroam
- https://github.com/spencer-tb/esp8266-32-eduroam
for example code on how to connect to WPA-Enterprise networks.

Fixes https://github.com/Aircoookie/WLED/issues/2749

PR #4664, opened 2025-04-26
add syslog function to wled

KrX3D:
# WLED Syslog Component

## Overview  
Adds a **SyslogPrinter** to the WLED core that captures all debug-print output and forwards it—over UDP—to a remote syslog server, in any of three standard formats (BSD/RFC3164, RFC5424 or raw text). This makes it easy to integrate WLED logs into your existing centralized logging/monitoring infrastructure.

### 1. Print interception & buffering  
- Capturing all debug-print output.  
- Characters are accumulated in a 128-byte buffer.  
- On newline (or buffer full), `flushBuffer()` trims any trailing `\r`, null-terminates, and sends the complete message over UDP.

### 2. Hostname resolution  
- The first time you send (or after changing host), `resolveHostname()` does a DNS lookup via `WiFi.hostByName()` (or `ETH.hostByName()` on Ethernet builds) and caches the resulting IP.

### 3. Message formatting & UDP send  
- **Priority** (`PRI`) = `(facility << 3) | severity`.  
- Uses `WiFiUDP` to emit one packet per message.  
- Supports three protocols:  
  1. **BSD (RFC 3164)**  
     ```
     <PRI>TIMESTAMP HOST APP-NAME: MESSAGE
     ```  
  2. **RFC 5424**  
     ```
     <PRI>1 TIMESTAMP-ISO HOST APP-NAME – – – MESSAGE
     ```  
  3. **RAW**  
     ```
     MESSAGE
     ```
In **RAW** mode the message payload is same as WLED’s existing NetDebug output—without any PRI, timestamp or header added.

## Example log output  
```text
Apr 26 13:54:59 EleksTube WLED: ===== WLED SYSLOG CONFIGURATION =====
Apr 26 13:54:59 EleksTube WLED:  Hostname:  192.168.1.13
Apr 26 13:54:59 EleksTube WLED:  Cached IP:  192.168.1.13
Apr 26 13:54:59 EleksTube WLED:  Port:       514
Apr 26 13:54:59 EleksTube WLED:  Protocol:   0 (BSD)
Apr 26 13:54:59 EleksTube WLED:  Facility:   20 (LCL4)
Apr 26 13:54:59 EleksTube WLED:  Severity:   7 (DEBUG)
Apr 26 13:54:59 EleksTube WLED: ======================================
```

### Build-time configuration (platform.ini)  
Enable debug output and syslog in your `platform.ini`:

```ini
; Core debug output (required for syslog forwarding)
-D WLED_DEBUG

[...]

PR #5194, opened 2025-12-15
Add support for WPA-Enterprise

### LINKED ISSUE #2749: WPA2 Enterprise Support
I am trying to make WLED work with WPA2 Enterprise, especially Eduroam networks that sometimes requires certificates.

There are example sketches from Arduino. I was able the use that to connect my ESP32 (NodeMCU) to the network of my organisation. But without WLED.

```
#include <WiFi.h> //Wifi library
#include "esp_wpa2.h" //wpa2 library for connections to Enterprise networks
#define EAP_IDENTITY "identitiy" //if connecting from another corporation, use identity@organisation.domain in Eduroam 
#define EAP_USERNAME "username" //oftentimes just a repeat of the identity
#define EAP_PASSWORD "password" //your Eduroam password
const char* ssid = "SSID"; // Eduroam SSID
const char* host = "domain"; //external server domain for HTTP connection after authentification
int counter = 0;

// NOTE: For some systems, various certification keys are required to connect to the wifi system.
//       Usually you are provided these by the IT department of your organization when certs are required
//       and you can't connect with just an identity and password.
//       Most eduroam setups we have seen do not require this level of authentication, but you should contact
//       your IT department to verify.
//       You should uncomment these and populate with the contents of the files if this is required for your scenario (See Example 2 and Example 3 below).

//       HERE YOU CAN JUST INSERT THE CERT CERTIFICATE FROM YOUR ORGANISATION. If it's in .pem format you can just copy-paste 
static const char incommon_ca[] PROGMEM = R"EOF(
-----BEGIN CERTIFICATE-----
MIIDwzCCAqugAwIBAgIBATANBgkqhkiG9w0BAQsFADCBgjELMAkGA1UEBhMCREUx
KzApBgNVBAoMIlQtU3lzdGVtcyBFbnRlcnByaXNlIFNlcnZpY2VzIEdtYkgxHzAd
BgNVBAsMFlQtU3lzdGVtcyBUcnVzdCBDZW50ZXIxJTAjBgNVBAMMHFQtVGVsZVNl
YyBHbG9iYWxSb290IENsYXNzIDIwHhcNMDgxMDAxMTA0MDE0WhcNMzMxMDAxMjM1
OTU5WjCBgjELMAkGA1UEBhMCREUxKzApBgNVBAoMIlQtU3lzdGVtcyBFbnRlcnBy
aXNlIFNlcnZpY2VzIEdtYkgxHzAdBgNVBAsMFlQtU3lzdGVtcyBUcnVzdCBDZW50
ZXIxJTAjBgNVBAMMHFQtVGVsZVNlYyBHbG9iYWxSb290IENsYXNzIDIwggEiMA0G
CSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQCqX9obX+hzkeXaXPSi5kfl82hVYAUd
AqSzm1nzHoqvNK38DcLZSBnuaY/JIPwhqgcZ7bBcrGXHX+0CfHt8LRvWurmAwhiC
FoT6ZrAIxlQjgeTNuUk/9k9uN0goOA/FvudocP05l03Sx5iRUKrERLMjfTlH6VJi
1hKTXrcxlkIF+3anHqP1wvzpesVsqXFP6st4vGCvx9702cu+fjOlbpSD8DT6Iavq
jnKgP6TeMFvvhk1qlVtDRKgQFRzlAVfFmPHmBiiRqiDFt1MmUUOyCxGVWOHAD3bZ
wI18gfNycJ5v/hqO2V81xrJvNHy+SE/iWjnX2J14np+GPgNeGYtEotXHAgMBAAGj
QjBAMA8GA1UdEwEB/wQFMAMBAf8wDgYDVR0PAQH/BAQDAgEGMB0GA1UdDgQWBBS/
WSA2AHmgoCJrjNXyYdK4LMuCSjANBgkqhkiG9w0BAQsFAAOCAQEAMQOiYQsfdOhy
NsZt+U2e+iKo4YFWz827n+qrkRk4r6p8FU3ztqONpfSO9kSpp+ghla0+AGIWiPAC
uvxhI+YzmzB6azZie60EI4RYZeLbK4rnJVM3YlNfvNoBYimipidx5joifsFvHZVw
IEoHNN/q/xWA5brXethbdXwFeilHfkCoMRN3zUA7tFFHei4R40cR3p1m0IvVVGb6
g1XqfMIpiRvpb7PO4gWEyS8+eIVibslfwXhjdFjASBgMmTnrpMwatXlajRWc2BQN
9noHV8cigwUtPJslJj0Ys6lDfMjIq2SPDqO/nBudMNva0Bkuqjzx+zOAduTNrRlP
BSeOE6Fuwg==
-----END CERTIFICATE-----
)EOF";
//const char *client_cert = "insert your client cert from your .crt file here";
//const char *client_key = "insert your client key from your .key file here";

void setup() {
  Serial.begin(115200);
  delay(10);
  Serial.println();
  Serial.print("Connecting to network: ");
  Serial.println(ssid);
  WiFi.disconnect(true);  //disconnect form wifi to set new wifi connection
  WiFi.mode(WIFI_STA); //init wifi mode
  
  // Example1 (most common): a cert-file-free eduroam with PEAP (or TTLS)
  WiFi.begin(ssid, WPA2_AUTH_PEAP, EAP_IDENTITY, EAP_USERNAME, EAP_PASSWORD);

  // Example 2: a cert-file WPA2 Enterprise with PEAP
  //WiFi.begin(ssid, WPA2_AUTH_PEAP, EAP_IDENTITY, EAP_USERNAME, EAP_PASSWORD, ca_pem, client_cert, client_key);
  
  // Example 3: TLS with cert-files and no password
  //WiFi.begin(ssid, WPA2_AUTH_TLS, EAP_IDENTITY, NULL, NULL, ca_pem, client_cert, client_key);
  
  
  while (WiFi.status() != WL_CONNECTED) {
    delay(500);
    Serial.print(".");
    counter++;
    if(counter>=60){ //after 30 seconds timeout - reset board
      ESP.restart();
    }
  }
  Serial.println("");
  Serial.println("WiFi connected");
  Serial.println("IP address set: "); 
  Serial.println(WiFi.localIP()); //print LAN IP
}
void loop() {
  if (WiFi.status() == WL_CONNECTED) { //if we are connected to Eduroam network
    counter = 0; //reset counter
    Serial.println("Wifi is still connected with IP: "); 
    Serial.println(WiFi.localIP());   //inform user about his IP address
  }else if (WiFi.status() != WL_CONNECTED) { //if we lost connection, retry
    WiFi.begin(ssid);      
  }
  while (WiFi.status() != WL_CONNECTED) { //during lost connection, print dots
    delay(500);
    Serial.print(".");
    counter++;
    if(counter>=60){ //30 seconds timeout - reset board
    ESP.restart();
    }
  }
  Serial.print("Connecting to website: ");
  Serial.println(host);
  WiFiClient client;
 if (client.connect(host, 80)) {
    String url = "/rele/rele1.txt";
    client.print(String("GET ") + url + " HTTP/1.1\r\n" + "Host: " + host + "\r\n" + "User-Agent: ESP32\r\n" + "Connection: close\r\n\r\n");

    while (client.connected()) {
      String line = client.readStringUntil('\n');
      if (line == "\r") {
        break;
      }
    }
    String line = client.readStringUntil('\n');
   Serial.println(line);
  }else{
      Serial.println("Connection unsucessful");
    }  
}
```

I tried this solution by @garyd9. But I was not able to implement that. 

Comment by @garyd9 :

> I'll just leave this little block of code I inserted in initConnection() here (mostly taken from various google search results)...  Oh, and it does work.  On my particular network, my username must be "domain\\user" instead of just "user", but it does work fine on PEAP/MSCHAPv2 (coming from an ESP32 device)
> 
> (requires "esp_wpa2.h" to be included to get at the WPA2 type and macro)
> ```
>   const char *username = "MYDOMAIN\\username";
>   const char *password = "mypassword";
> 
>   WiFi.disconnect(true);
>   WiFi.mode(WIFI_STA);  // didn't work without this.  odd.
>   esp_wifi_sta_wpa2_ent_set_username((uint8_t *)username, strlen(username));
>   esp_wifi_sta_wpa2_ent_set_password((uint8_t *)password, strlen(password));
>   esp_wpa2_config_t config = WPA2_CONFIG_INIT_DEFAULT();
>   esp_err_t code = esp_wifi_sta_wpa2_ent_enable(&config);
>   Serial.print ("enable response code: ");
>   Serial.println(code);
>   const char *ssid = clientSSID;
>   Serial.print("wifi Begin connection to: "); Serial.println(ssid);
>   WiFi.begin(ssid);
> 
> ```
> To use this properly, the HTTP UI would need to prompt for an additional piece of info:  a username.   I just hard-coded everything to try it out.  (I'm still exploring this esp32 dev board.  it's fun.)

_Originally posted by @garyd9 in https://github.com/Aircoookie/WLED/issues/1001#issuecomment-656989199_

**Additional context**
I am new to coding and I am just starting to learn through the internet. I asked for help and @blazoncek told me to insert the code in wled.cpp under initConnection() 

>wled.cpp -> initConnection()

_Originally posted by @blazoncek in https://github.com/Aircoookie/WLED/issues/2747#issuecomment-1211998587_

But as you can see I was not able to do that. 

I think this feature would make WLED even better for students and other people that use WPA2-Enterprise.

Thanks

benjaminkraus:
This change adds support for enterprise WiFi requiring a username and anonymous identity.

This change is a revival of this old PR: https://github.com/wled/WLED/pull/3195

The WiFi setup page was augmented with a drop-down menu to choose between "None/WPA/WPA2" and "WPA/WPA2-Enterprise". When you select "WPA/WPA2-Enterprise", two new fields are shown to enter your anonymous identity and identity (username).

I tested using an ESP32.

Fixes https://github.com/wled/WLED/issues/2749

## Release Notes

**New Features**
* Added WPA/WPA2-Enterprise WiFi authentication support with full configuration options
* New WiFi settings fields: encryption type selection, anonymous identity, and identity credentials
* Device now supports enterprise and corporate WiFi network configurations
* Enhanced WiFi setup to handle complex authentication requirements

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -529,7 +529,7 @@ build_flags = ${common.build_flags} ${esp32_idf_V4.build_flags} -D WLED_RELEASE_
   -DBOARD_HAS_PSRAM -mfix-esp32-psram-cache-issue ;; Older ESP32 (rev.<3) need a PSRAM fix (increases static RAM used) https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/external-ram.html
   -D DATA_PINS=25
 lib_deps = ${esp32_idf_V4.lib_deps}
-  
+
 [env:esp32c3dev]
 extends = esp32c3
 platform = ${esp32c3.platform}
@@ -639,7 +639,7 @@ monitor_filters = esp32_exception_decoder
 
 [env:esp32s3_4M_qspi]
 ;; ESP32-S3, with 4MB FLASH and <= 4MB PSRAM (memory_type: qio_qspi)
-board = lolin_s3_mini ;; -S3 mini, 4MB flash 2MB PSRAM 
+board = lolin_s3_mini ;; -S3 mini, 4MB flash 2MB PSRAM
 platform = ${esp32s3.platform}
 platform_packages = ${esp32s3.platform_packages}
 upload_speed = 921600
--- a/platformio.ini
+++ b/platformio.ini
@@ -529,7 +529,7 @@ build_flags = ${common.build_flags} ${esp32_idf_V4.build_flags} -D WLED_RELEASE_
   -DBOARD_HAS_PSRAM -mfix-esp32-psram-cache-issue ;; Older ESP32 (rev.<3) need a PSRAM fix (increases static RAM used) https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/external-ram.html
   -D DATA_PINS=25
 lib_deps = ${esp32_idf_V4.lib_deps}
-
+  
 [env:esp32c3dev]
 extends = esp32c3
 platform = ${esp32c3.platform}
@@ -639,7 +639,7 @@ monitor_filters = esp32_exception_decoder
 
 [env:esp32s3_4M_qspi]
 ;; ESP32-S3, with 4MB FLASH and <= 4MB PSRAM (memory_type: qio_qspi)
-board = lolin_s3_mini ;; -S3 mini, 4MB flash 2MB PSRAM
+board = lolin_s3_mini ;; -S3 mini, 4MB flash 2MB PSRAM 
 platform = ${esp32s3.platform}
 platform_packages = ${esp32s3.platform_packages}
 upload_speed = 921600
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -114,6 +114,15 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
       multiWiFi[n].staticIP = nIP;
       multiWiFi[n].staticGW = nGW;
       multiWiFi[n].staticSN = nSN;
+      byte encType = WIFI_ENCRYPTION_TYPE_PSK;
+      char anonIdent[65] = "";
+      char ident[65] = "";
+      CJSON(encType, wifi[F("enc_type")]);
+      getStringFromJson(anonIdent, wifi["e_anon_ident"], 65);
+      getStringFromJson(ident, wifi["e_ident"], 65);
+      multiWiFi[n].encryptionType = encType;
+      strlcpy(multiWiFi[n].enterpriseAnonIdentity, anonIdent, 65);
+      strlcpy(multiWiFi[n].enterpriseIdentity, ident, 65);
       if (++n >= WLED_MAX_WIFI_COUNT) break;
     }
   }
@@ -869,6 +878,11 @@ void serializeConfig(JsonObject root) {
       wifi_ip.add(multiWiFi[n].staticIP[i]);
       wifi_gw.add(multiWiFi[n].staticGW[i]);
       wifi_sn.add(multiWiFi[n].staticSN[i]);
+    }
+    wifi[F("enc_type")] = multiWiFi[n].encryptionType;
+    if (multiWiFi[n].encryptionType == WIFI_ENCRYPTION_TYPE_ENTERPRISE) {
+      wifi[F("e_anon_ident")] = multiWiFi[n].enterpriseAnonIdentity;
+      wifi[F("e_ident")] = multiWiFi[n].enterpriseIdentity;
     }
   }
 
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -114,6 +114,7 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
       multiWiFi[n].staticIP = nIP;
       multiWiFi[n].staticGW = nGW;
       multiWiFi[n].staticSN = nSN;
+#ifdef WLED_ENABLE_WPA_ENTERPRISE
       byte encType = WIFI_ENCRYPTION_TYPE_PSK;
       char anonIdent[65] = "";
       char ident[65] = "";
@@ -123,6 +124,7 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
       multiWiFi[n].encryptionType = encType;
       strlcpy(multiWiFi[n].enterpriseAnonIdentity, anonIdent, 65);
       strlcpy(multiWiFi[n].enterpriseIdentity, ident, 65);
+#endif
       if (++n >= WLED_MAX_WIFI_COUNT) break;
     }
   }
@@ -879,11 +881,13 @@ void serializeConfig(JsonObject root) {
       wifi_gw.add(multiWiFi[n].staticGW[i]);
       wifi_sn.add(multiWiFi[n].staticSN[i]);
     }
+#ifdef WLED_ENABLE_WPA_ENTERPRISE
     wifi[F("enc_type")] = multiWiFi[n].encryptionType;
     if (multiWiFi[n].encryptionType == WIFI_ENCRYPTION_TYPE_ENTERPRISE) {
       wifi[F("e_anon_ident")] = multiWiFi[n].enterpriseAnonIdentity;
       wifi[F("e_ident")] = multiWiFi[n].enterpriseIdentity;
     }
+#endif
   }
 
   JsonArray dns = nw.createNestedArray(F("dns"));
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -208,6 +208,10 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
 #define USERMOD_ID_BRIGHTNESS_FOLLOW_SUN 57     //Usermod "usermod_v2_brightness_follow_sun.h"
 #define USERMOD_ID_USER_FX               58     //Usermod "user_fx"
 
+//Wifi encryption type
+#define WIFI_ENCRYPTION_TYPE_PSK          0     //None/WPA/WPA2
+#define WIFI_ENCRYPTION_TYPE_ENTERPRISE   1     //WPA/WPA2-Enterprise
+
 //Access point behavior
 #define AP_BEHAVIOR_BOOT_NO_CONN          0     //Open AP when no connection after boot
 #define AP_BEHAVIOR_NO_CONN               1     //Open when no connection (either after boot or if connection is lost)
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -209,8 +209,10 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
 #define USERMOD_ID_USER_FX               58     //Usermod "user_fx"
 
 //Wifi encryption type
-#define WIFI_ENCRYPTION_TYPE_PSK          0     //None/WPA/WPA2
-#define WIFI_ENCRYPTION_TYPE_ENTERPRISE   1     //WPA/WPA2-Enterprise
+#ifdef WLED_ENABLE_WPA_ENTERPRISE
+  #define WIFI_ENCRYPTION_TYPE_PSK          0     //None/WPA/WPA2
+  #define WIFI_ENCRYPTION_TYPE_ENTERPRISE   1     //WPA/WPA2-Enterprise
+#endif
 
 //Access point behavior
 #define AP_BEHAVIOR_BOOT_NO_CONN          0     //Open AP when no connection after boot
--- a/wled00/data/settings_wifi.htm
+++ b/wled00/data/settings_wifi.htm
@@ -89,7 +89,7 @@
 						select.appendChild(option);
 
 						if (input.value === "" || input.value === "Your_Network" || found) input.replaceWith(select);
-						else select.remove(); 
+						else select.remove();
 					}
 				}
 
@@ -118,11 +118,20 @@
 			gId("wifi_add").style.display = (i<maxNetworks) ? "inline":"none";
 			gId("wifi_rem").style.display = (i>1) ? "inline":"none";
 		}
-		function addWiFi(ssid="",pass="",bssid="",ip=0,gw=0,sn=0x00ffffff) { // little endian
+		function addWiFi(type=0,ssid="",anon="",ident="",pass="",bssid="",ip=0,gw=0,sn=0x00ffffff) { // little endian
 			var i = gId("wifi_entries").childNodes.length;
 			if (i >= maxNetworks) return;
 			var b = `<div id="net${i}"><hr class="sml">
 Network name (SSID${i==0?", empty to not connect":""}):<br><input type="text" id="CS${i}" name="CS${i}" maxlength="32" value="${ssid}" ${i>0?"required":""}><br>
+WiFi encryption type:<br>
+<select id="ET${i}" name="ET${i}" onchange="E(${i})">
+<option value="0"${(type==0) ? ' selected':''}>None/WPA/WPA2</option>
+<option value="1"${(type==1) ? ' selected':''}>WPA/WPA2-Enterprise</option>
+</select><br>
+<div id="IDS${i}" style="${(type==0) ? 'display:none;':''}">
+Anonymous identity:<br><input type="text" id="EA${i}" name="EA${i}" maxlength="64" value="${anon}"><br>
+Identity:<br><input type="text" id="EI${i}" name="EI${i}" maxlength="64" value="${ident}"><br>
+</div>
 Network password:<br><input type="password" name="PW${i}" maxlength="64" value="${pass}"><br>
 BSSID (optional):<br><input type="text" id="BS${i}" name="BS${i}" maxlength="12" value="${bssid}"><br>
 Static IP (leave at 0.0.0.0 for DHCP)${i==0?"<br>Also used by Ethernet":""}:<br>
@@ -184,6 +193,14 @@ Static subnet mask:<br>
 			l.appendChild(r);
 			rC++;
 			gId('+').style.display = gId("rml").childElementCount < 10 ? 'inline' : 'none'; // can't append to list anymore, hide button
+		}
+		function E(i) {
+			const select = gId("ET"+i);
+			if (select.value === "0") {
+				gId("IDS"+i).style.display = "none";
+			} else {
+				gId("IDS"+i).style.display = "";
+			}
 		}
 
 	</script>
--- a/wled00/data/settings_wifi.htm
+++ b/wled00/data/settings_wifi.htm
@@ -118,12 +118,12 @@
 			gId("wifi_add").style.display = (i<maxNetworks) ? "inline":"none";
 			gId("wifi_rem").style.display = (i>1) ? "inline":"none";
 		}
-		function addWiFi(type=0,ssid="",anon="",ident="",pass="",bssid="",ip=0,gw=0,sn=0x00ffffff) { // little endian
+		function addWiFi(ssid="",pass="",bssid="",ip=0,gw=0,sn=0x00ffffff,type=-1,anon="",ident="") { // little endian
 			var i = gId("wifi_entries").childNodes.length;
 			if (i >= maxNetworks) return;
-			var b = `<div id="net${i}"><hr class="sml">
-Network name (SSID${i==0?", empty to not connect":""}):<br><input type="text" id="CS${i}" name="CS${i}" maxlength="32" value="${ssid}" ${i>0?"required":""}><br>
-WiFi encryption type:<br>
+			var encryptionTypeField = "";
+			if (type >=0 && type < 2) {
+				encryptionTypeField = `WiFi encryption type:<br>
 <select id="ET${i}" name="ET${i}" onchange="E(${i})">
 <option value="0"${(type==0) ? ' selected':''}>None/WPA/WPA2</option>
 <option value="1"${(type==1) ? ' selected':''}>WPA/WPA2-Enterprise</option>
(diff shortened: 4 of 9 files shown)


DedeHai:
thanks for updating. I am not sure on how to proceed on this one but it can not be in the default builds as this eats up way too much flash, 80k on ESP8266 and over 100k on ESP32 variants.
@willmmiles what are your thoughts on encapsulating it in #ifdefs and what would be the alternative to have this as a custom build option?

willmmiles:
> thanks for updating. I am not sure on how to proceed on this one but it can not be in the default builds as this eats up way too much flash, 80k on ESP8266 and over 100k on ESP32 variants. @willmmiles what are your thoughts on encapsulating it in #ifdefs and what would be the alternative to have this as a custom build option?

You know I'm not a fan of `#ifdef`s ;) but that's probably the least evil approach given that networking is a core feature.

From an implementation standpoint:
- Relevant define should be `WLED_ENABLE_WPA_ENTERPRISE`, ie. the inverse of the current approach
- UI fields, including the type selector, must be hidden if the firmware does not support enterprise auth.  Bonus points if they're omitted entirely at build time to avoid wasting flash.
- New config values (including auth type) should be omitted in `serializeConfig` unless the feature is enabled.  Similarly, `deserializeConfig` should ignore them unless the feature is enabled.

Other notes:
- Should either of the identity fields be stored in the "secure" config?  Is the password the only secret?
- There are a number of unrelated whitespace changes in this PR.  Someone's got a bad editor config (possibly me!) but either way they are cluttering the diffs.  Please avoid staging unrelated changes.

benjaminkraus:
> > thanks for updating. I am not sure on how to proceed on this one but it can not be in the default builds as this eats up way too much flash, 80k on ESP8266 and over 100k on ESP32 variants. @willmmiles what are your thoughts on encapsulating it in #ifdefs and what would be the alternative to have this as a custom build option?
> 
> You know I'm not a fan of `#ifdef`s ;) but that's probably the least evil approach given that networking is a core feature.
> 
> From an implementation standpoint:
> 
> * Relevant define should be `WLED_ENABLE_WPA_ENTERPRISE`, ie. the inverse of the current approach
> 
> * UI fields, including the type selector, must be hidden if the firmware does not support enterprise auth.  Bonus points if they're omitted entirely at build time to avoid wasting flash.
> 
> * New config values (including auth type) should be omitted in `serializeConfig` unless the feature is enabled.  Similarly, `deserializeConfig` should ignore them unless the feature is enabled.

Thanks @willmmiles and @DedeHai for the feedback.

I'll make these changes and do my best to omit the UI fields entirely at built time if the feature is disabled (without overly complicating the code).

> Other notes:
> 
> * Should either of the identity fields be stored in the "secure" config?  Is the password the only secret?

I considered this, but I wasn't sure if there was a cost to putting more things into the secure config. It is an easy enough change if we think the anonymous identity and/or identity should be kept more secure. Based on my (admittedly limited) understanding of WPA enterprise, I believe the "anonymous identity" is presented unencrypted when establishing the connection but the "identity" is only sent over an encrypted channel, so I could see an argument for treating the anonymous identity similarly to the SSID and treating the identity/username similar to the password. As two random data points: Neither my Linux laptop or my Android phone do anything to obscure the identity and anonymous identity on their WiFi configuration pages (they are both visible by default).

> * There are a number of unrelated whitespace changes in this PR.  Someone's got a bad editor config (possibly me!) but either way they are cluttering the diffs.  Please avoid staging unrelated changes.

The whitespace changes were done automatically by Visual Studio Code and it was actually a little hard to undo the changes (Visual Studio Code kept generously fixing them again for me) so I left them in my initial commit. I debated whether it would be more distracting to leave them in or to add a later change reverting them. I'll revert all the unrelated whitespace-only changes.

DedeHai:
> I'll make these changes and do my best to omit the UI fields entirely at built time if the feature is disabled (without overly complicating the code).

IIRC there was a PR that used a script to inject JS code at build time but don't ask me which one ;)

> (Visual Studio Code kept generously fixing them again for me)

I have auto-remove disabled but have white-spaces show in red, I think that is a VSc plugin (I would recommend to get @willmmiles  if you are using VScode ;) )

KrX3D:
> IIRC there was a PR that used a script to inject JS code at build time but don't ask me which one ;)

just if you mean inject_syslog_ui.py : https://github.com/wled/WLED/pull/4664

DedeHai:
thanks @KrX3D, that exactly :)

willmmiles:
Oof, using a python script to manipulate the HTML files is a world of complexity I'd rather not take on.  I was expecting that it should be done via the `npm build` step.  Converting from the "source html" to "this is the binary blob embedded in the firmware" is the whole point of that stage.  Possibly `build_ui.py` needs to do something to explicitly pass key environment variables in, though.

netmindz:
Yeah this definitely needs to be excluded by default given it's increased flash usege and very niche usage

benjaminkraus:
I just pushed an update to address your review feedback.

I couldn't find a clean way to completely remove the JavaScript relating to the settings page when `WLED_ENABLE_WPA_ENTERPRISE` is disabled because `cdata.js` has no knowledge of the compiler flags (as far as I could tell). However, the fields are completely omitted from the resulting HTML when the page is loaded.

I did run into one question when I was debugging: As the code is currently written, if you are running firmware with `WLED_ENABLE_WPA_ENTERPRISE` enabled and have an enterprise WiFi configured, then you switch to firmware that has `WLED_ENABLE_WPA_ENTERPRISE` disabled, the code will attempt to connect to the enterprise WiFi using just the password. I wondered whether I should do something additional to mitigate this, but ultimately decided it was a pretty rare situation and the resulting behavior wasn't much different from accidentally entering the wrong WiFi password, so I left it as-is. Let me know if you disagree.

benjaminkraus:
Another question: I see the notice that "All checks have passed", but I imagine none of those checks sets the `WLED_ENABLE_WPA_ENTERPRISE` compile flag. Should I add a build configuration and/or test that sets `WLED_ENABLE_WPA_ENTERPRISE`? If so, do you have a suggestion for a pattern to follow? It seems like one check for EPS32 and one for ESP8266 makes sense.

willmmiles:
> I couldn't find a clean way to completely remove the JavaScript relating to the settings page when WLED_ENABLE_WPA_ENTERPRISE is disabled because cdata.js has no knowledge of the compiler flags (as far as I could tell). However, the fields are completely omitted from the resulting HTML when the page is loaded.

I think we might have different definitions of "completely omitted".  The DOM elements are never created, yes; but the element code itself is still transmitted from the microcontroller to the browser.  So it's not "omitted" from the persepective of bytes that must be stored in flash and transmitted by the device, which are the senses that matter.  The browser client computer is enormous in terms of memory and storage compared to the microcontrollers; whether an element never constructed or merely hidden is largely immaterial in comparison.

That said: you are correct to note that there's not yet a framework for conditionally including parts of the UI.  I'll admit I was trying to push in the direction of creativity on that problem. ;)

@DedeHai @netmindz are we OK with accepting the extra couple of hundred bytes for this until we can figure out dynamic UI builds?  I'm OK with this for the time being, and it's a good prototype case for a future build-time-configurable UI.

> I did run into one question when I was debugging: As the code is currently written, if you are running firmware with WLED_ENABLE_WPA_ENTERPRISE enabled and have an enterprise WiFi configured, then you switch to firmware that has WLED_ENABLE_WPA_ENTERPRISE disabled, the code will attempt to connect to the enterprise WiFi using just the password. I wondered whether I should do something additional to mitigate this, but ultimately decided it was a pretty rare situation and the resulting behavior wasn't much different from accidentally entering the wrong WiFi password, so I left it as-is. Let me know if you disagree.

I think that's a reasonable behaviour.

> Another question: I see the notice that "All checks have passed", but I imagine none of those checks sets the WLED_ENABLE_WPA_ENTERPRISE compile flag. Should I add a build configuration and/or test that sets WLED_ENABLE_WPA_ENTERPRISE? If so, do you have a suggestion for a pattern to follow? It seems like one check for EPS32 and one for ESP8266 makes sense.

Hm, now this is more tricky.  We don't have currently have any "test only" environments except the usermods build(s).  @netmindz  On the assumption that we don't plan to release pre-built binaries with this feature, do you think this is a good use case of a matrix build with some of our existing environments, vs creating some new envs?  Are there other defines we might want to consider for CI testing in the same way?

benjaminkraus:
@willmmiles what is the status of this pull request? Are we still waiting for a response from either @DedeHai or @netmindz?

netmindz (approved):

netmindz:
The code looks ok, although as noted none of our CI will build with `WLED_ENABLE_WPA_ENTERPRISE`   as we won't be adding this as feature of our default bins due to it's limited usage

willmmiles:
> The code looks ok, although as noted none of our CI will build with `WLED_ENABLE_WPA_ENTERPRISE` as we won't be adding this as feature of our default bins due to it's limited usage

Do we want to consider adding a test build to the CI anyways, even if it won't be published in a release, like the usermods builds?

netmindz:
> > The code looks ok, although as noted none of our CI will build with `WLED_ENABLE_WPA_ENTERPRISE` as we won't be adding this as feature of our default bins due to it's limited usage
> 
> Do we want to consider adding a test build to the CI anyways, even if it won't be published in a release, like the usermods builds?

We don't currently have a way to use a different env lists for PRs vs releases

willmmiles:
> > > The code looks ok, although as noted none of our CI will build with `WLED_ENABLE_WPA_ENTERPRISE` as we won't be adding this as feature of our default bins due to it's limited usage
> > 
> > 
> > Do we want to consider adding a test build to the CI anyways, even if it won't be published in a release, like the usermods builds?
> 
> We don't currently have a way to use a different env lists for PRs vs releases

OK.  Let's take this as-is and open an issue for arranging that some some CI envs are filtered out for releases (like `usermods`); then we can add it to a test build.

damnms:
I tried to get that working with a quinled next dig 2 (esp32), but it fails on compilation as soon as i use that flag. I am not sure if i do something wrong or if there is something broken. is there some CI/CD system that builds those images? 

the error is this: 
```
Compiling .pio/build/esp32dev_8M/FrameworkArduino/wiring_shift.c.o
Retrieved `.pio/build/esp32dev_8M/lib196/libWebServer.a' from cache
Archiving .pio/build/esp32dev_8M/libFrameworkArduino.a
Indexing .pio/build/esp32dev_8M/libFrameworkArduino.a
Linking .pio/build/esp32dev_8M/firmware.elf
/home/oli/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: .pio/build/esp32dev_8M/src/wled.cpp.o:(.literal._ZN4WLED14initConnectionEv+0x24): undefined reference to `WiFiSTAClass::begin(char const*, wpa2_auth_method_t, char const*, char const*, char const*, char const*, char const*, char const*, int, unsigned char const*, bool)'
/home/oli/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: .pio/build/esp32dev_8M/src/wled.cpp.o: in function `WLED::initConnection()':
/home/oli/WLED/wled00/wled.cpp:722: undefined reference to `WiFiSTAClass::begin(char const*, wpa2_auth_method_t, char const*, char const*, char const*, char const*, char const*, char const*, int, unsigned char const*, bool)'
collect2: error: ld returned 1 exit status
*** [.pio/build/esp32dev_8M/firmware.elf] Error 1

```

DedeHai:
@damnms as I stated in discourse, this is uncharted terretory in V5 builds. Use V4 build / 16_x branch.

damnms:
it failed on tag 16.0.1 and 16_x branch

softhack007:
@damnms @DedeHai I can reproduce the build failure on 16_x, but only with the (default) tasmota arduino platform.
If I enable the original espressif platform, the firmware build finishes without error.

This might be another limitation of the tasmota core, similar to the missing WPA3 support (see #5568).

To make the build succeed:
comment out these lines
https://github.com/wled/WLED/blob/a0253c8fdca58b27566bee3f997ccdceced0bbef/platformio.ini#L299-L300

and remove the ";; " at the start of
https://github.com/wled/WLED/blob/a0253c8fdca58b27566bee3f997ccdceced0bbef/platformio.ini#L304-L305


This should be enough for your 8MB esp32; if will not work on 4MB boards though, because the original espressif platform needs more program space.

softhack007:
edit: WPA-Enterprise support also compiles successfully with an older tasmota platform, after fixing an undefined `ADC_ATTEN_DB_12` in audioreactive (change to `ADC_ATTEN_DB_11`)

```ini
platform = https://github.com/tasmota/platform-espressif32/releases/download/2023.06.02/platform-espressif32.zip ;; Tasmota Arduino Core 2.0.9 with IPv6 support, based on IDF 4.4.4
platform_packages =
```

damnms:
thanks for having a look at that!
can you maybe guide me a bit? i am now at the 16_x branch

```
~/WLED$ git diff
diff --git a/platformio.ini b/platformio.ini
index 637cd464..cb9bf8be 100644
--- a/platformio.ini
+++ b/platformio.ini
@@ -109,6 +109,7 @@ debug_flags = -D DEBUG=1 -D WLED_DEBUG
 #  This reduces the OTA size with ~45KB, so it's especially useful on low memory boards (512k/1m).
 build_flags =
+  -D WLED_ENABLE_WPA_ENTERPRISE
   -DMQTT_MAX_PACKET_SIZE=1024
   -DSECURE_CLIENT=SECURE_CLIENT_BEARSSL
   -DBEARSSL_SSL_BASIC
@@ -296,13 +297,13 @@ AR_lib_deps =  ;; for pre-usermod-library platformio_override compatibility
 
 ;; tasmota platform (default)
-platform = https://github.com/tasmota/platform-espressif32/releases/download/2024.06.00/platform-espressif32.zip ;; Tasmota Arduino Core 2.0.18 with IPv6 support, based on IDF 4.4.8
-platform_packages =
+#platform = https://github.com/tasmota/platform-espressif32/releases/download/2024.06.00/platform-espressif32.zip ;; Tasmota Arduino Core 2.0.18 with IPv6 support, based on IDF 4.4.8
+#platform_packages =
 
 ;; espressif platform (optional - needs 300KB extra flash)
 ;;; arduino-esp32 2.0.17 + esp-idf 4.4.7
-;; platform = espressif32@ ~6.13.0
-;; platform_packages =
+platform = espressif32@ ~6.13.0
+platform_packages =
 ;;; arduino-esp32 2.0.14 + esp-idf 4.4.6
 ;; platform = espressif32@ ~6.6.0
 ;; platform_packages = platformio/framework-arduinoespressif32 @ 3.20014.231204

```

when i try to run this with `~/.venv-pio/bin/pio run -e esp32dev_8M`, i get:
```
Compiling .pio/build/esp32dev_8M/src/xml.cpp.o
Building .pio/build/esp32dev_8M/bootloader.bin
Retrieved `.pio/build/esp32dev_8M/partitions.bin' from cache
Traceback (most recent call last):
  File "/home/oli/.platformio/packages/tool-esptoolpy/esptool.py", line 41, in <module>
    import esptool
  File "/home/oli/.platformio/packages/tool-esptoolpy/esptool/__init__.py", line 42, in <module>
    from esptool.bin_image import intel_hex_to_bin
  File "/home/oli/.platformio/packages/tool-esptoolpy/esptool/bin_image.py", line 16, in <module>
    from intelhex import HexRecordError, IntelHex
ModuleNotFoundError: No module named 'intelhex'
Compiling .pio/build/esp32dev_8M/libbde/AsyncTCP/AsyncTCP.cpp.o
*** [.pio/build/esp32dev_8M/bootloader.bin] Error 1
wled00/util.cpp:778:68: note: #pragma message: BOARD_HAS_PSRAM not defined, not using PSRAM.
     #pragma message("BOARD_HAS_PSRAM not defined, not using PSRAM.")
                                                                    ^
======================================================================== [FAILED] Took 47.20 seconds ========================================================================

Environment    Status    Duration
-------------  --------  ------------
esp32dev_8M    FAILED    00:00:47.195
=================================================================== 1 failed, 0 succeeded in 00:00:47.195 ===================================================================

```

softhack007:
> File "/home/oli/.platformio/packages/tool-esptoolpy/esptool/bin_image.py", line 16, in <module>
>     from intelhex import HexRecordError, IntelHex
> ModuleNotFoundError: No module named 'intelhex'

@damnms this sometimes happens when you change platform versions. Usually one of these fixes it:
* install the `intelhex` package: `pip install intelhex`
* clear all build cache folders (`.pio`, `%HOMEPATH%/.buildcache`, `%HOMEPATH%/.platformio`, and maybe `%HOMEPATH%/AppData/Local/npm-cache`)

damnms:
Thanks, that really made it. I had to use platform = espressif32@ ~6.13.0 to be able to use -D WLED_ENABLE_WPA_ENTERPRISE and install intelhex. Can you tell me why is that with the platform? Is there any disadvantage in using espressif32 vs. tasmota? It would be awesome if there would be some sort of CI/CD that creates those images (for platforms that can handle the image size).

awesome, i flashed it. first i got an error because i used the _dev version, but now it works, thanks!

softhack007:
> Can you tell me why is that with the platform? Is there any disadvantage in using espressif32 vs. tasmota?

Yes it's simple. The espressif platform makes wled too big for boards with 4MB flash. So to keep 4MB boards supported, we had to use the tasmota platform, which is smaller than the original from espressif. 

Tasmota has reduced networking support and removed some features of "vanilla" arduino-esp32, in order to reduce firmware size.

PR #5202, opened 2025-12-17
PoC Rotate Segment

Brandon502:
New rotate effect based on copy that rotates a segment's 2D pixel data with optional zoom, wrap, and mirror wrap features. Inspired by @softhack007 spinning matrix https://github.com/MoonModules/WLED-MM/pull/122

This can be implemented as a "utility" effect like copy segment, or @blazoncek mentioned expanding SEGMENT options to include rotate, zoom, etc.

https://github.com/user-attachments/assets/14e90cd8-5857-4f9a-859f-bff2729191f1

https://github.com/user-attachments/assets/55348b9b-3cb6-4839-b987-acc912b3e681

## Summary by CodeRabbit

* **New Features**
  * Added "Rotate Segment" effect: rotate LED segments with adjustable zoom, wrap, mirror wrap and segment selection.
  * Added as a new effect mode (increases total available effects by one).

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -189,6 +189,18 @@ uint16_t mode_copy_segment(void) {
 static const char _data_FX_MODE_COPY[] PROGMEM = "Copy Segment@,Color shift,Lighten,Brighten,ID,Axis(2D),FullStack(last frame);;;12;ix=0,c1=0,c2=0,c3=0";
 
 uint16_t mode_rotate_segment(void) {
+  
+  uint32_t sourceid = SEGMENT.custom3;
+  Segment& sourcesegment = strip.getSegment(sourceid);
+  
+  if (!sourcesegment.isActive() || 
+      !sourcesegment.is2D() ||
+       sourceid >= strip.getSegmentsNum() ||
+       sourceid == strip.getCurrSegmentId()) {
+         SEGMENT.fadeToBlackBy(5); // fade out
+         return FRAMETIME;
+        }
+        
   const int cols = SEG_W, rows = SEG_H;
   const int midX = cols / 2;
   const int midY = rows / 2;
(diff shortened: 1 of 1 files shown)


blazoncek:
To clarify what @Brandon502 referred to when mentioning "segment options":

My personal view is that "effect" (or, if you prefer, function that will paint pixels) should have no idea about other segments (or canvas if you prefer) other than the one it is supposed to paint on (encapsulation). This unfortunately prevents "effects" like segment copy or segment zoom or segment rotate (as it could only copy/zoom/rotate its segment/canvas).

However if we move up in object hierarchy and allow segments to interact (within `WS2812FX` class) then these features (including 1D expansion) no longer belong into pixel painting primitives or "effects" but in canvas blending (and manipulation) part of frame processing. This will speed up processing as no transformations are needed until canvas is ready to be transferred into frame buffer.

We already have segment property for 1D -> 2D expansion (`map1d2d`) and if we enhance that with, let's say, `uint8_t angle, zoom;` and `bool rotate, pulsate;` these could be used in `WS2812FX::show()` to do rotation/zooming (as well as `offset`/panning). Similarly source segment for copy could be added.

This is all theoretical and may not be easily achieved in practice, I agree (I needed 1 whole year to come up with segment blending that actually works). It may also overwhelm MCUs with small amounts of RAM like S2 and ESP8266 as JSON API will increase. However I think idea is worth exploring.

softhack007:
Hi @Brandon502 nice to see how an old idea gets a second chance with your PR 😃 it looks like you used the "shear rotation" from Moonlight, instead of my older "rotate each pixel coordinate" idea. I was always fighting with gaps/holes in the rotated pixels, does shear rotation work better?

I'm (again) very busy with non-wled matters, but I'll take the time to review your code once I find a bit of time ✌️

Brandon502:
> Hi @Brandon502 nice to see how an old idea gets a second chance with your PR 😃 it looks like you used the "shear rotation" from Moonlight, instead of my older "rotate each pixel coordinate" idea. I was always fighting with gaps/holes in the rotated pixels, does shear rotation work better?
> 
> I'm (again) very busy with non-wled matters, but I'll take the time to review your code once I find a bit of time ✌️

Shear rotation shouldn't produce any holes outside of the corners when rotated which wrap can fix. This method is an inverse of Moonlight version. Instead of shearing the pixel to the destination, the destination is inverse sheared to the source location. It's already pretty efficient but I'm sure it can be improved further.

blazoncek:
@Brandon502 @dedehai @softhack007  I've been considering zoom and rotate as a segment property (in a similar way as 1D->2D expansion is) and have come to the following extension proposal:
- introduce 2 new public 4-bit properties: `zoomAmount`and `rotateSpeed`
- introduce 1 new private 16-bit property `rotatedAngle` to record rotation progress (as `aux0` variable in this PR)
- implement zooming and rotating in `WS2812FX::blendSegment()` using the code from this PR (almost no changes needed, but with additional temporary buffer to hold zoomed/rotated content) to transfer segment buffer to frame buffer
- zoom and rotation is processed before transitions take effect

I have prepared POC which is available in my fork (includes UI).

PR #5213, opened 2025-12-20
Add old version check to OTA update

willmmiles:
Replacement for #5057 with compatibility fixes added.  

- We no longer limit the upward OTA metadata structure version.  The expectation is that new structure versions are backwards compatible with old versions; breaking changes should use a new magic number.
- This build reports structure version 1 for compatibility with old nightlies that still have the version number check.  This should be undone before the 0.16 release.

## Summary by CodeRabbit

* **New Features**
  * Added version safety verification for firmware updates to prevent installation of incompatible releases.

* **Bug Fixes**
  * Improved error messaging clarity for firmware update compatibility checks.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #5214, opened 2025-12-20
Remove OTA metadata structure version block

willmmiles:
0.15 backport of the OTA version check fix from #5213, on the off chance we have an OTA version 3 before 0.16 launches.


netmindz:
Good to merge this too then @willmmiles  ?

Issue #5133, opened 2025-11-29
Issue with the WLED frontend


Stijn-Jacobs:
### What happened?

After just having built the master build from source, I'm experiencing a weird issue where the frontend stops working after setting my led stripe length. I have DMX enabled on this build.

After configuring my LED config with the ui it works. This is the relevant part of my working config;

```
    "led": {
      "total": 5,
      "maxpwr": 8000,
      "cct": false,
      "cr": false,
      "ic": false,
      "cb": 0,
      "fps": 60,
      "rgbwm": 255,
      "prl": false,
      "ins": [
        {
          "start": 0,
          "len": 4,
          "pin": [
            16
          ],
          "order": 1,
          "rev": false,
          "skip": 0,
          "type": 22,
          "ref": false,
          "rgbwm": 0,
          "freq": 0,
          "maxpwr": 0,
          "ledma": 55,
          "text": ""
        },
        {
          "start": 4,
          "len": 1,
          "pin": [
            18
          ],
          "order": 1,
          "rev": false,
          "skip": 0,
          "type": 22,
          "ref": false,
          "rgbwm": 0,
          "freq": 0,
          "maxpwr": 0,
          "ledma": 255,
          "text": ""
        }
      ]
    },
```

If I then change the length of the strip from 1 to 144, the frontend throws the following exception:

```
TypeError: Cannot read properties of undefined (reading 'on')
    at readState ((index):7:66326)
    at (index):7:71514
```

This points to the javascript

```
function readState(e, t=!1) {
                if (!e)
                    return !1;
                if (e.success)
                    return !0;
                isOn = e.on,
                gId("sliderBri").value = e.bri,
                nlA = e.nl.on,
                nlDur = e.nl.dur,
                nlTar = e.nl.tbri,
                nlFade = e.nl.fade,
```

Where it seems to be failing to properly read the current state. 
If I revert the led length back to 1, everything functions normally again.

### To Reproduce Bug

1. Build WLED from source with DMX enabled
2. Set led strip length to 144, from the default 30.
3. Frontend no longer functions becouse of javascript exception.

### Expected Behavior

Working frontend after having configured the correct led stripe length.

### Install Method

Self-Compiled

### What version of WLED?

WLED 0.16.0-alpha (2506160)

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

[...]

PR #5216, opened 2025-12-21
DMX improvements and configuration option.

Stijn-Jacobs:
As already reported in my #5133 , I found out that the current DMX implementation was negatively impacting performance. It previously tried to push all of the led strips registered onto the DMX bus. 

With this PR I've added a new option to configure how many DMX fixtures worth of output should be sent.

I've also added rate limiting to the DMX code, to make sure it is updated at 44hz (DMX theoretical maximum speed).

I've tested these changes on my own hardware, which has DMX output added to it.

Update: 
I've changed the implementation from the old sparkfun one, to use esp_dmx instead. The sparkfun implementation created (and destroyed) 2 UART serials for each loop cycle which caused major performance issues. It seems to no longer have any negative impact at all on the frame rate, where previously I've seen about 20 frames per second, now >100 is possible. Also removed the legacy code from sparkfun implementation where it was reading DMX, this has since been moved.

## Summary by CodeRabbit

* **New Features**
  * Added a DMX "Number of fixtures" setting (1–512) in DMX settings and persisted to configuration.
  * Added a configurable "Wired DMX Output Pin" field in sync/settings UI.

* **Improvements**
  * DMX output now limits processing to configured fixtures, skips when inactive/zero brightness, and enforces frame-rate timing for more reliable transmission.
  * UI markup adjusted for clearer inputs and line breaks.

* **Removed**
  * Replaced legacy DMX backend with a new platform-focused DMX output implementation.


DedeHai:
looks good to me, except the rabbit's remark: the break signal should be low.

netmindz (changes requested):
DMX is meant to be a constant stream of data, not only on state change. Fixtures will consider it a failure state of they do not receive updates

netmindz:
If we are doing anything with DMX output we should really be ditching the sparkfun implementation and using the esp_dmx as is used for input as it appears to be more complaint

Stijn-Jacobs:
> DMX is meant to be a constant stream of data, not only on state change. Fixtures will consider it a failure state of they do not receive updates

Good point, bit of an oversight on my part. I've resolved it.


> If we are doing anything with DMX output we should really be ditching the sparkfun implementation and using the esp_dmx as is used for input as it appears to be more complaint

I've been testing a bit with the esp_dmx, and seem to get much better performance out of it. It seems to not influence the frame rate at all, which was the case with the sparkfun implementation. I will push my suggested changes soon

However the library documentation explains that input and output concurrently is not possible. I don't see any code that would prevent anyone from doing so. Wouldn't it be best to prevent compilation with the DMX output + input flag defined. As it most certainly would cause issues, even before the switch over to esp_dmx for output?

wled00/src/dependencies/dmx/EspDmxOutput.cpp
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  if (dmxInstalled) {
+    dmx_set_pin(dmxPort, txPin, rxPin, rtsPin);
+
+    // Ensure NULL start code (slot 0).
+    dmxData[0] = DMX_SC;
+    dmx_write_slot(dmxPort, 0, DMX_SC);
+  }
+}
```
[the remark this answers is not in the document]
Stijn-Jacobs: This doesn't seem the case for the esp_dmx library. As seen in the source code of it, it just returns a boolean.
https://github.com/someweisguy/esp_dmx/blob/9b6a5043add89038388d93ebe77109eb3368bb51/src/dmx/driver.c#L246

wled00/dmx_output.cpp
```diff
@@ -3,21 +3,36 @@
...
 {
   // don't act, when in DMX Proxy mode
   if (e131ProxyUniverse != 0) return;
 
+  // Ensure segments exist before accessing strip data
+  if (strip.getSegmentsNum() == 0) return;
+
+  // Rate limiting
+  static unsigned long last_dmx_time = 0;
+  const unsigned long dmxFrameTime = (1000UL + MAX_DMX_RATE - 1) / MAX_DMX_RATE;
```
softhack007: Wehere is MAX_DMX_RATE defined? (maybe i just overlooked it)

softhack007: The formula ``(1000UL + MAX_DMX_RATE - 1) / MAX_DMX_RATE`` is used to always round up, i guess? 
Instead of 1000/MAX_DMX_RATE which is rounding down.
If yes, then a comment could be good to just state the purpose.

wled00/dmx_output.cpp
```diff
@@ -3,21 +3,36 @@
...
+  if (millis() - last_dmx_time < dmxFrameTime) return;
```
softhack007, line 26: rate limiting looks OK for me.

wled00/dmx_output.cpp
Stijn-Jacobs: Its defined here https://github.com/wled/WLED/blob/d2d37e8135e928eafd91ace45030e571d9e19473/wled00/dmx_output.cpp#L15
top of the file.

wled00/dmx_output.cpp
```diff
@@ -3,21 +3,36 @@
...
 {
   // don't act, when in DMX Proxy mode
   if (e131ProxyUniverse != 0) return;
 
+  // Ensure segments exist before accessing strip data
+  if (strip.getSegmentsNum() == 0) return;
+
+  // Rate limiting
+  static unsigned long last_dmx_time = 0;
+  const unsigned long dmxFrameTime = (1000UL + MAX_DMX_RATE - 1) / MAX_DMX_RATE; // Subtract 1 to round up
```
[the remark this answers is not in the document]
softhack007: a small optimization could be to make it ``constexpr`` instead of ``const``. constexpr guarantees that the formula is calculated by the compiler, zero costs at runtime.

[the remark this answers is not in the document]
Stijn-Jacobs: > a small optimization could be to make it `constexpr` instead of `const`. constexpr guarantees that the formula is calculated by the compiler, zero costs at runtime.

Nice one, just changed it.

netmindz:
Looks like some of our development effort is overlapping @Stijn-Jacobs  https://github.com/wled/WLED/pull/5287

Stijn-Jacobs:
@netmindz Yea we can't really merge them both, would you be willing to add your changes to my PR? The runtime pin change is a nice addition. I've already tested all of my changes with working hardware. Combined would be the best of both.

I also think we should do something when the user enables both DMX input and output at the same time. Since this isn't supported by the underlying library.

wled00/src/dependencies/dmx/EspDmxOutput.cpp
```diff
@@ -0,0 +1,93 @@
...
+#define defaultMax 32
+
+// Some new MCUs (-S2, -C3) don't have HardwareSerial(2)
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(4, 2, 0)
+  #if SOC_UART_NUM < 3
+  #error DMX output is not possible on your MCU, as it does not have HardwareSerial(2)
+  #endif
+#endif
+
+static constexpr dmx_port_t dmxPort = DMX_NUM_2;
```
netmindz, line 30: We can use either HardwareSerial1 or HardwareSerial2  - we just can't use the same one for both DMX input and DMX output (or any other conflicting usage)

netmindz:
Can you please add me as collaborator to your fork so I can push the changes to this PR or you can pull in the changes from the verison i've pushed to my personal repo @Stijn-Jacobs

Stijn-Jacobs:
> Can you please add me as collaborator to your fork so I can push the changes to this PR or you can pull in the changes from the verison i've pushed to my personal repo @Stijn-Jacobs

I've just added you as a collaborator @netmindz

netmindz:
Pushed, can you test your hardware setup @Stijn-Jacobs ?

netmindz:
@coderabbitai review

Stijn-Jacobs:
> Pushed, can you test your hardware setup @Stijn-Jacobs ?

Sorry missed your message, will test on my hardware tomorrow.

wled00/dmx_output.cpp
```diff
@@ -1,23 +1,36 @@
...
+
+  // Rate limiting
+  static unsigned long last_dmx_time = 0;
+  constexpr unsigned long dmxFrameTime = (1000UL + MAX_DMX_RATE - 1) / MAX_DMX_RATE; // Ceiling division to round up
+  if (millis() - last_dmx_time < dmxFrameTime) return;
+
   uint8_t brightness = strip.getBrightness();
+  
+  // Skip DMX entirely if strip is off
+  if (brightness == 0) return;
```
Mdbelen, line 31: Some fixtures have a manual mode in case they don't receive a DMX signal. For example they could jump to a static color, where they should be off. This line should be removed IMO.

wled00/src/dependencies/dmx/EspDmxOutput.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <Arduino.h>
+#if !defined(CONFIG_IDF_TARGET_ESP32C3) && !defined(CONFIG_IDF_TARGET_ESP32S2)
+
+#include "EspDmxOutput.h"
+#include <esp_dmx.h>
+
+#include "wled.h"
+
+#define dmxMaxChannel 512
+#define defaultMax 32
```
Mdbelen, line 20: This is a dead define, as DMXOutput::init calls EspDmxOutput::initWrite always with 512. Should be removed.

Stijn-Jacobs: Well 44hz is the actual hard limit of what the spec can handle. The limiting was there in place to increase performance dramatically before using the old driver, and thus was ported from there. It should have less impact right now, but will still have some negative impact on the actual framerate you will achieve on a microcontroller. 

I think at the moment fixtures with different channel numbers are not really properly supported globally (while you ofcourse can set it to the value of the fixture with the highest fixture count, and setup your channels properly on the fixtures). 

I think its a good point you mention, I'll also do some more testing with 512 channels and see if it impacts the framerate as it did before.

wled00/dmx_output.cpp
Stijn-Jacobs: Hmm this is an interesting one, I agree with you. If you have DMX on in the build, we can assume you intend to use it with DMX and we can always output zero values to the bus, even if this means impacting performance negatively.

wled00/src/dependencies/dmx/EspDmxOutput.cpp
```diff
@@ -0,0 +1,101 @@
...
+#endif
+
+#if SOC_UART_NUM > 2
+static constexpr dmx_port_t dmxPort = DMX_NUM_2;
+#else
+static constexpr dmx_port_t dmxPort = DMX_NUM_1;
+#endif
+
+// DMX value array and size. Entry 0 holds start code, so we need 512+1 elements.
+static uint8_t dmxData[DMX_PACKET_SIZE] = {0};
```
Mdbelen, line 36: Do we need this? This is not read ever, is it?

Stijn-Jacobs: You're right, reminiscence of the old driver. I'll remove it.

wled00/dmx_output.cpp
netmindz: Good spot @Mdbelen I totally agree

wled00/src/dependencies/dmx/EspDmxOutput.cpp
Mdbelen: Previously, I got the impression that SparkFunDMX was only used with HardwareSerial2, i.e. all ESPs with < 3 UARTs like the S2 would fallback to ESPDMX bitbanging, right? Do you know why that was, was serial1 reserved for something like IR, NPB, etc and is this not needed anymore? Just asking out of interest I guess.

Mdbelen:
Tbh, I mistrust the esp_dmx driver a little. With my oscilloscope I measure (some) MAB (mark after break) pulses of as short as 8us, where they should be at least 12, and according to the code _are_ at least 12. While I understand the intention to use the same driver for input and output, I personally feel that this is not really "production ready" and reliable, and there are other way simpler methods of reaching almost similar performance gains. I am not very familiar with the more fancy features of the ++ aspects of C and RTOSes that are being used, so can't really fiddle with that code.
I really appreciate the effort that is going on towards DMX of all of you!
I'll try to finish and link my branch this we for you guys to evaluate if you like.

Stijn-Jacobs:
> Tbh, I mistrust the esp_dmx driver a little. With my oscilloscope I measure (some) MAB (mark after break) pulses of as short as 8us, where they should be at least 12, and according to the code _are_ at least 12. While I understand the intention to use the same driver for input and output, I personally feel that this is not really "production ready" and reliable, and there are other way simpler methods of reaching almost similar performance gains. I am not very familiar with the more fancy features of the ++ aspects of C and RTOSes that are being used, so can't really fiddle with that code. I really appreciate the effort that is going on towards DMX of all of you! I'll try to finish and link my branch this we for you guys to evaluate if you like.

Hmm, I would like to know the accuracy of the previous implentation. I'm guessing that it would be even worse tbh. I unfortunually don't have access to a ossiciscope myself, so am not able to test this deep.

Mdbelen:
The SparkFunDMX implementation is blocking but therefore 100% reproducible. This means that it took 100% processing time for the whole DMX frame of 22ms, but did exact timing every time. Creating and destroying the HardwareSerial object isn't a problem per se as you've mentioned above, but the fact that they use a blocking send method.
Basically you can enable a TX buffer with one line of code, and that fixes it. This is what I did in my approach. The Espressif HAL takes over the buffer handling for us using interrupts. So the break and MAB is still blocking, but the 513 Bytes of DMX data afterwards are non-blocking. This reduces the time spent on the update() function from 22ms to 150us plus some data handling and relieves the main loop, so no more WS28xx lags or network connectivity issues occur. Ofc I encountered some weird behavior as well though, so I still need some improvements ;p

Stijn-Jacobs:
> The SparkFunDMX implementation is blocking but therefore 100% reproducible. This means that it took 100% processing time for the whole DMX frame of 22ms, but did exact timing every time. Creating and destroying the HardwareSerial object isn't a problem per se as you've mentioned above, but the fact that they use a blocking send method. Basically you can enable a TX buffer with one line of code, and that fixes it. This is what I did in my approach. The Espressif HAL takes over the buffer handling for us using interrupts. So the break and MAB is still blocking, but the 513 Bytes of DMX data afterwards are non-blocking. This reduces the time spent on the update() function from 22ms to 150us plus some data handling and relieves the main loop, so no more WS28xx lags or network connectivity issues occur. Ofc I encountered some weird behavior as well though, so I still need some improvements ;p

I can add you as a collaborator on this fork for you to push your changes if you would like, or are you starting over fully?

Mdbelen:
Ugh idk right now. Ofc I also did some unnecessary tidy-up changes etc so far, so I don't think it would go together all too well.
What's the status of @netmindz including his former PR here? I think I didn't see the possibility to config the DMX_TX pin on web UI yet, with the build from your branch that I've made. This was one feature of his branch, wasn't it? Don't want to make him any additional effort. But I see his commits here. Have to look further into that as well.
And rn I'm not saying I have the perfect solution ready. Just that with my measurements I'd feel a little worried about the esp_dmx you're using, but it is definitely already a big win vs previously.
Give me a night or two, I'll go expose to the big yellow halogen bulb in the sky a bit first.

Stijn-Jacobs:
> Ugh idk right now. Ofc I also did some unnecessary tidy-up changes etc so far, so I don't think it would go together all too well. What's the status of @netmindz including his former PR here? I think I didn't see the possibility to config the DMX_TX pin on web UI yet, with the build from your branch that I've made. This was one feature of his branch, wasn't it? Don't want to make him any additional effort. But I see his commits here. Have to look further into that as well. And rn I'm not saying I have the perfect solution ready. Just that with my measurements I'd feel a little worried about the esp_dmx you're using, but it is definitely already a big win vs previously. Give me a night or two, I'll go expose to the big yellow halogen bulb in the sky a bit first.

Hahaha allright I'll add you anyway and you can decide. I haven't compiled this one lately but @netmindz did push his changes of the runtime DMX pin config changes to this branch. @netmindz Do you have more info?

Mdbelen:
Phew, finally opened PR: https://github.com/wled/WLED/pull/5522 . Hope I didn't mess anything up.
Would be great if you could also test that on your hardware & comment your opinion.

PR #5219, opened 2025-12-21
import annotations from __future in validate_modules.py

pyro9:
allows Python version 3.8 to be used on development system
Has no effect on built firmware

## Summary by CodeRabbit

* **Chores**
  * Updated internal validation tooling with minor code improvements. No changes to user-facing functionality or APIs.

### DIFF
--- a/pio-scripts/validate_modules.py
+++ b/pio-scripts/validate_modules.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import re
 from pathlib import Path   # For OS-agnostic path manipulation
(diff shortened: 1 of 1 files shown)


willmmiles:
Python 3.8 has been EOL for more than a year, and [PlatformIO is currently tested only on Python 3.11 and newer.](https://github.com/platformio/platformio-core/blob/3106062c356b5be791f88942c9302b1115497547/.github/workflows/core.yml#L11).  Consider using a development container or cloud environment if your host system cannot be updated.

PR #5242, opened 2025-12-29
use same approach for ledmap as is done in gap file

DedeHai:
- fixes ledmap containing garbage values

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed LED mapping initialization so empty matrix entries are explicitly marked and trailing LEDs default to identity mapping. This prevents uninitialized or incorrect LED mappings after loading a custom map, improving display consistency and preventing unexpected LED behavior when using partial or smaller mapping configurations.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -2005,13 +2005,12 @@ bool WS2812FX::deserializeMap(unsigned n) {
 
   d_free(customMappingTable);
   customMappingTable = static_cast<uint16_t*>(d_malloc(sizeof(uint16_t)*getLengthTotal())); // prefer DRAM for speed
-  // fill with empty in case we don't fill the entire matrix
-  unsigned matrixSize = Segment::maxWidth * Segment::maxHeight;
-  for (unsigned i = 0; i<matrixSize; i++) customMappingTable[i] = 0xFFFFU;
-  for (unsigned i = matrixSize; i<getLengthTotal(); i++) customMappingTable[i] = i; // trailing LEDs for ledmap (after matrix) if it exist
 
   if (customMappingTable) {
     DEBUG_PRINTF_P(PSTR("ledmap allocated: %uB\n"), sizeof(uint16_t)*getLengthTotal());
+    unsigned matrixSize = Segment::maxWidth * Segment::maxHeight;
+    for (unsigned i = 0; i<matrixSize; i++) customMappingTable[i] = 0xFFFFU;          // fill with empty in case we don't fill the entire matrix
+    for (unsigned i = matrixSize; i<getLengthTotal(); i++) customMappingTable[i] = i; // trailing LEDs for ledmap (after matrix) if it exist
     File f = WLED_FS.open(fileName, "r");
     f.find("\"map\":[");
     while (f.available()) { // f.position() < f.size() - 1
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -2005,6 +2005,10 @@ bool WS2812FX::deserializeMap(unsigned n) {
 
   d_free(customMappingTable);
   customMappingTable = static_cast<uint16_t*>(d_malloc(sizeof(uint16_t)*getLengthTotal())); // prefer DRAM for speed
+  // fill with empty in case we don't fill the entire matrix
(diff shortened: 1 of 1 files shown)


DedeHai:
@softhack007 what do you think of this PR? what I am not sure about is whether 
`for (unsigned i = 0; i<matrixSize; i++) customMappingTable[i] = 0xFFFFU;`
should be done or not i.e. if it is a matrix and the size defined is larger than the mapped number of pixels the pixels inside the matrix will be mapped to -1 instead of "1:1" mapping, so same as it is done for gap files.

softhack007:
> should be done or not i.e. if it is a matrix and the size defined is larger than the mapped number of pixels the pixels inside the matrix will be mapped to -1 instead of "1:1" mapping, so same as it is done for gap files.

@DedeHai I think that pixels not covered in the map should be treated like "no map applied", so mapped 1:1 per default.

But 🤔 if the ledmap has "holes" (like 1,4,2,5).... then the mapping strip -> bus is still complete for the range [0...3], "strip" pixel 4 is not included in the map (trailing pixel), but "bus" pixels 0 and 3 have no counterpart of the "strip" level any more. 

🤔 It might confuse users if pixel 4 can be set with either "strip pixel 1" (ledmap) or "strip pixel 4" (unmapped part). Maybe your solution is right for this case, and the "not a trailing strip" pixels in the 2D map range should all have a default of "-1".

@ewowi as the ledmap expert, what's your opinion on this case? 
 in my example above, "strip pixel" are logical pixels, while "bus pixel" are physical pixels.

ewowi:
I am too long gone from WLED mapping to have a good advice. 
I moved over to 'inverse mapping'. I don't provide (index) positions in the led daisy chain but xyz positions of the lights in daisy chain order.
Internally a mappingTable vaguely similar to WLED is created, by definition has the size of the virtual coordinate space and each cell is always one of 3:
1) not mapped (but provided space to store the color of that cell which otherwise was in the LEDs array)
2) mapped to one index in the LEDs array (this is similar to WLED)
3) mapped to multiple indexes in the LEDs array (one to many mappings).

So the mappingTable is always filled with relevant data for [0..x*y*z] so no need to FFFF cells

DedeHai:
> But 🤔 if the ledmap has "holes" (like 1,4,2,5).... then the mapping strip -> bus is still complete for the range [0...3], "strip" pixel 4 is not included in the map (trailing pixel), but "bus" pixels 0 and 3 have no counterpart of the "strip" level any more.
> 
> 🤔 It might confuse users if pixel 4 can be set with either "strip pixel 1" (ledmap) or "strip pixel 4" (unmapped part). Maybe your solution is right for this case, and the "not a trailing strip" pixels in the 2D map range should all have a default of "-1".

@softhack007 this is actually a good example, ultimately there is no right or wrong for the implementation, its just a design choice. Incomplete ledmaps can be handled one way or another, the "setting incomplete to -1" only applies to 2D maps and the updated KB docs (currently in PR IIRC) state that incomplete maps are to be avoided. I have no preference, I simply did it the same as it is done for gap files, which sets incomplete to -1. I would recommend to keep it compatible to MM as much as possible.

edit: I also want to check if the two mapping functions i.e. ledmap and gap files can be unified into a single function as they are almost identical.

DedeHai:
@softhack007 this has been pending for quite some time - I just checked how MM is doing it and it uses 1:1 mapping, the -1 mapping used in this PR is commented out. 

```c++
//memset(customMappingTable, 0xFF, customMappingTableSize * sizeof(uint16_t)); // FFFF = no pixel
for (unsigned i=0; i<customMappingTableSize; i++) customMappingTable[i]=i;     // "neutral" 1:1 mapping
```

so the options for this PR are do it the same as MM or do it the same as gaps file and fill the rest with 1:1

i.e.

```c++
for (unsigned i = 0; i<getLengthTotal(); i++) customMappingTable[i] = i; // fill all with 1:1 mapping in case ledmap is incomplete
```

or

```c++
for (unsigned i = 0; i<matrixSize; i++) customMappingTable[i] = 0xFFFFU;          // fill with empty in case we don't fill the entire matrix
for (unsigned i = matrixSize; i<getLengthTotal(); i++) customMappingTable[i] = i; // trailing LEDs for ledmap (after matrix) if it exist
```

DedeHai:
superseded by #5529

PR #5248, opened 2025-12-31
add json validation to file inputs in UI and minify before upload

### LINKED ISSUE #5246: Restore preset/config
### What happened?

If you accidentally or hastily restore the settings using the JSON presets, it will indicate that it's working OK, but it's actually resetting to factory settings. Therefore, the format isn't verified before application.

### To Reproduce Bug

Restore the settings using the wled_presets.json file

### Expected Behavior

Verify the format and settings before attempting to apply it

### Install Method

Binary from WLED.me

### What version of WLED?

0.15.3

ESP32

### Relevant log/trace output

```shell

```

### Anything else?

Image

### Code of Conduct

DedeHai:
this prevents invalid files even making it to the file system. 
also updated edit.htm to make use of that so even files uploaded through edit.htm benefit from this new functionality. While I was at it I also added sequential file loading in edit.htm like all other htm files already do.

Fixes #5246

## Summary by CodeRabbit

* **New Features**
  * Editor adds a queued request utility for smoother saves and loader behavior.
  * Uploads now accept a callback to refresh and reload the saved file after upload.

* **Bug Fixes**
  * JSON uploads may be auto-minified/validated with an option to continue on parse error.
  * File-type detection and editor mode switching are now case-insensitive.
  * Improved empty-file handling and clearer success/error feedback; live JSON validation updates editor border.


DedeHai:
I tested this with a valid and an invalid json file, works as intended. Also can confirm that pretty-printed json files are minified after upload i.e. when downloaded from the FS. 
Note: if you test this using the new file editor, you will not see it minified as the editor pretty-prints when displaying and minifies again when saving. you need to download the file to see its actual contents as-is.

PR #5261, opened 2026-01-02
FEATURE: WLED Power Consumption in JSON API

hntirgeam:
### Adds approximate Watt-based power reporting to /json/info endpoint for easier integration with Home Assistant. 

New API Fields
- leds.pwr_w - Current power consumption in W
- leds.maxpwr_w - Maximum power budget in W
- leds.voltage - Configured LED strip voltage

```json
  "leds": {
    "count": 60,
    "pwr": 158,
    "pwr_w": 0.79,
    "fps": 41,
    "maxpwr": 0,
    "maxpwr_w": 0,
    "voltage": 5,
    ...
  },
```

### UI: Voltage Configuration
Added voltage configuration in LED settings to support different strip voltages (5V/12V/24V): 

UPD:
- Can be toggled off via UI.

image

### Bug (?) Fix
Power monitoring now works when ABL disabled. Previously, it only reported ESP consumption.

Existing pwr and maxpwr mA fields - unchanged

### Why

I think it's cool that there will soon be an option to add WLED consumption to Home Assistant to track these values over time.

## Summary by CodeRabbit

* **New Features**
  * Optional per-output power monitoring with a configurable supply-voltage setting (persisted in config and UI).
  * Status now reports supply voltage and current power (watts) in the status JSON.

* **Improvements**
  * Automatic Brightness Limiting now accounts for power monitoring for more accurate current/power estimation and per-bus brightness limits.
  * LED settings UI restores and displays per-strip voltage and per-LED current controls.

### DIFF
--- a/wled00/bus_manager.cpp
+++ b/wled00/bus_manager.cpp
@@ -1397,7 +1397,7 @@ void BusManager::initializeABL() {
 }
 
 void BusManager::applyABL() {
-  if (_useABL) {
+  if (_useABL || _usePowerMonitoring) {
     unsigned milliAmpsSum = 0; // use temporary variable to always return a valid _gMilliAmpsUsed to UI
     unsigned totalLEDs = 0;
     for (auto &bus : busses) {
@@ -1411,7 +1411,7 @@ void BusManager::applyABL() {
       }
     }
     // check global current limit and apply global ABL limit, total current is summed above
-    if (_gMilliAmpsMax > 0) {
+    if (_gMilliAmpsMax > 0 && _useABL) {
       uint8_t  newBri = 255;
       uint32_t globalMax = _gMilliAmpsMax > MA_FOR_ESP ? _gMilliAmpsMax - MA_FOR_ESP : 1; // subtract ESP current consumption, fully limit if too low
       if (globalMax > totalLEDs) { // check if budget is larger than standby current
@@ -1432,24 +1432,19 @@ void BusManager::applyABL() {
             busd.applyBriLimit(newBri);
         }
       }
-    }
-    _gMilliAmpsUsed = milliAmpsSum;
-  }
-  else if (_usePowerMonitoring) {
-    // Calculate current for power monitoring without applying brightness limits
-    unsigned milliAmpsSum = 0;
-    for (auto &bus : busses) {
-      if (bus->isDigital() && bus->isOk()) {
-        BusDigital &busd = static_cast<BusDigital&>(*bus);
-        busd.estimateCurrent();
-        busd.applyBriLimit(255); // reset _colorSum without limiting brightness
-        milliAmpsSum += busd.getUsedCurrent();
+    } else if (_usePowerMonitoring) {
+      // reset _colorSum for power monitoring without applying brightness limits
+      for (auto &bus : busses) {
+        if (bus->isDigital() && bus->isOk()) {
+          BusDigital &busd = static_cast<BusDigital&>(*bus);
+          busd.applyBriLimit(255);
+        }
       }
     }
     _gMilliAmpsUsed = milliAmpsSum;
   }
   else
-    _gMilliAmpsUsed = 0; // reset, we have no current estimation without ABL
+    _gMilliAmpsUsed = 0; // reset, we have no current estimation without ABL or power monitoring
 }
 
 float BusManager::currentWatts() {
--- a/wled00/data/settings_leds.htm
+++ b/wled00/data/settings_leds.htm
@@ -138,6 +138,7 @@
 			var en = d.Sf.PM.checked;
 			gId('pwrmon').style.display = (en) ? 'inline':'none';
 			gId('pmwarn').style.display = (en && !d.Sf.ABL.checked) ? 'inline':'none';
+			UI();
 		}
 		function enABL()
 		{
@@ -180,10 +181,11 @@
 		// enable and update LED Amps
 		function enLA(s,n)
 		{
+			const abl = d.Sf.ABL.checked;
 			const t = parseInt(d.Sf["LT"+n].value); // LED type SELECT
 			gId('LAdis'+n).style.display = s.selectedIndex==5 ? "inline" : "none"; // show/hide custom mA field
 			if (s.value!=="0") d.Sf["LA"+n].value = s.value; // set value from select object
-			d.Sf["LA"+n].min = (!isDig(t)) ? 0 : 1; // set minimum value for validation (required for power monitoring)
+			d.Sf["LA"+n].min = (!isDig(t) || !abl) ? 0 : 1; // set minimum value for validation
 		}
 		function setABL()
 		{
@@ -295,11 +297,11 @@
 				memu += getMem(t, n); // calc memory
 				dC += (isDig(t) && !isD2P(t));
 				setPinConfig(n,t);
-				gId("dig"+n+"ma").style.display = isDig(t) ? "inline" : "none"; // show mA/LED for digital LEDs (for power monitoring)
+				gId("dig"+n+"ma").style.display = (isDig(t) && (d.Sf.ABL.checked || d.Sf.PM.checked)) ? "inline" : "none"; // show mA/LED only when ABL or power monitoring enabled
 				if (change) { // did we change LED type?
 					gId("rf"+n).checked = (gId("rf"+n).checked || t == 31); // LEDs require data in off state (mandatory for TM1814)
 					if (isAna(t)) d.Sf["LC"+n].value = 1;                   // for sanity change analog count just to 1 LED
-					d.Sf["LA"+n].min = (!isDig(t)) ? 0 : 1;                 // set minimum value for LED mA (always required for power monitoring)
+					d.Sf["LA"+n].min = (!isDig(t) || !abl) ? 0 : 1;         // set minimum value for LED mA
 					d.Sf["MA"+n].min = (!isDig(t)) ? 0 : 250;               // set minimum value for PSU mA
 				}
 				gId("rf"+n).onclick = mustR(t) ? (()=>{return false}) : (()=>{});           // prevent change change of "Refresh" checkmark when mandatory
@@ -879,7 +881,7 @@ Swap: <select id="xw${s}" name="XW${s}">
 		<br>
 		Enable power monitoring: <input type="checkbox" name="PM" onchange="enPM()"><br>
 		<div id="pwrmon" style="display:none">
-		LED Strip Voltage: <input name="LV" type="number" class="m" step="1" min="1" max="50" value="5" oninput="UI()"> V<br>
+		LED Strip Voltage: <input name="LV" type="number" class="m" step="1" min="1" max="255" value="5" oninput="UI()"> V<br>
 		<i>Exposes power in API. Typical: 5V (WS281x), 12V/24V (others)</i><br>
 		<i id="pmwarn" class="warn">⚠ CPU intensive, may reduce FPS</i><br>
 		</div>
--- a/wled00/set.cpp
+++ b/wled00/set.cpp
@@ -148,7 +148,7 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
     BusManager::setMilliampsMax(ablMilliampsMax);
 
     uint8_t ledVoltage = request->arg(F("LV")).toInt();
-    if (ledVoltage > 0 && ledVoltage <= 50) BusManager::setVoltage(ledVoltage);
+    if (ledVoltage > 0) BusManager::setVoltage(ledVoltage);
 
     BusManager::enablePowerMonitoring(request->hasArg(F("PM")));
 


DedeHai:
thanks for contributing. 
There is a good and simple reason the ABL is not always calculated: it eats up quite a lot of CPU cycles. Enforcing that on every setup is not acceptable.
Also such a feature which has limited usecases should be implemented as a usermod IMHO. The calculated watt values will not be very accurate
Does HA now allow to calculate power from the ABL current estimate?

hntirgeam:
> There is a good and simple reason the ABL is not always calculated: it eats up quite a lot of CPU cycles. Enforcing that on every setup is not acceptable.

Ok, I agree. What do you think if I make these calculations behind a flag (smth like: `ABL_ENABLED || VOLTAGE_METRICS_ENABLED`) in the UI with the "disabled" default state? 

---

> Does HA now allow to calculate power from the ABL current estimate?

No, not yet. I plan to implement it by adding power value to the WLED entity (but based on W, not mA). 

Actually, my first idea was to use pwr and just multiply it by a known voltage in HA and call it a day, but:

1) in order to display pwr in the API, ABL must be enabled (which I don't use since I want max brightness and all my PSUs are powerful). It seems strange to me that I need to have ABL enabled (when I don't need it) in order to see pwr in the API

2) this approach makes it much harder to integrate into HA as it requires quite a lot of manual configuration. 
---

> The calculated watt values will not be very accurate

Yes, they wont, but this rough approximation:
- can be "calibrated" using "custom mA/LED"
- will still be better than nothing in terms of consumed power tracking

DedeHai:
Since HA allows to calculate power from current I do not see the point in doing it in WLED, it just bloats the code and makes setup more complicated for the normal use-case just for the convenience of not having to set it up in HA.
As a usermod why not, as core code I do not approve.

hntirgeam:
@DedeHai, I've addressed the performance concern by making power monitoring user-configurable and disabled by default:

## Changes:
- Added `Enable power monitoring` toggle (OFF by default)
- Only runs calculations when explicitly enabled by the user (no additional CPU usage for users who don't need it)
- Warning in UI about CPU usage (hidden when ABL is already enabled, since calculations run anyway)

image

## **core vs usermod:**

The main issue with a usermod approach is that power calculations in `applyABL()` only run when ABL is enabled. For power monitoring to work without ABL, the usermod would need to:
1. Duplicate the entire `estimateCurrent()` logic, or
2. Force users to enable ABL even when they don't want brightness limiting

Both approaches are more complex and less maintainable than a simple flag in core.

## This feature enables seamless support for power monitoring without:
- Requiring users to enable ABL when they don't need it
- Manual voltage multipliers configured in HA for each WLED device (I've 20 of them)
- Forcing users to understand the relationship between mA and their PSU voltage


## **Setup complexity:**
> makes setup more complicated for the normal use-case

For users who don't need this: nothing changes. Defaults remain the same, no new required fields. 

For users who do need it: one checkbox + voltage value (defaults to 5V).

wled00/bus_manager.cpp
```diff
@@ -293,8 +293,9 @@ void IRAM_ATTR BusDigital::setPixelColor(unsigned pix, uint32_t c) {
   if (Bus::_cct >= 1900) c = colorBalanceFromKelvin(Bus::_cct, c); //color correction from CCT
   c = color_fade(c, _bri, true); // apply brightness
 
-  if (BusManager::_useABL) {
-    // if using ABL, sum all color channels to estimate current and limit brightness in show()
+  // Always sum color channels for power monitoring (used for ABL and power consumption tracking)
```
DedeHai: no. this is slow.

wled00/bus_manager.cpp
```diff
@@ -1393,46 +1398,55 @@ void BusManager::initializeABL() {
 }
 
 void BusManager::applyABL() {
-  if (_useABL) {
```
DedeHai, line 1396: please revert unnecessary changes, it makes it hard to review and spot issues.

wled00/bus_manager.h
```diff
@@ -496,6 +500,12 @@ namespace BusManager {
   //inline uint16_t ablMilliampsMax()             { unsigned sum = 0; for (auto &bus : busses) sum += bus->getMaxCurrent(); return sum; }
   inline uint16_t ablMilliampsMax()             { return _gMilliAmpsMax; }  // used for compatibility reasons (and enabling virtual global ABL)
   inline void     setMilliampsMax(uint16_t max) { _gMilliAmpsMax = max;}
+  inline uint8_t  getVoltage()                  { return _gVoltage; }
+  inline void     setVoltage(uint8_t v)         { _gVoltage = v; }
+  inline bool     getPowerMonitoring()          { return _usePowerMonitoring; }
+  inline void     setPowerMonitoring(bool pm)   { _usePowerMonitoring = pm; }
+  float           currentWatts();               // calculate total power consumption in Watts
+  inline float    ablWattsMax()                 { return (_gMilliAmpsMax * _gVoltage) / 1000.0; }
```
DedeHai: what's this used for?

wled00/bus_manager.h
```diff
@@ -496,6 +500,12 @@ namespace BusManager {
   //inline uint16_t ablMilliampsMax()             { unsigned sum = 0; for (auto &bus : busses) sum += bus->getMaxCurrent(); return sum; }
   inline uint16_t ablMilliampsMax()             { return _gMilliAmpsMax; }  // used for compatibility reasons (and enabling virtual global ABL)
   inline void     setMilliampsMax(uint16_t max) { _gMilliAmpsMax = max;}
+  inline uint8_t  getVoltage()                  { return _gVoltage; }
+  inline void     setVoltage(uint8_t v)         { _gVoltage = v; }
+  inline bool     getPowerMonitoring()          { return _usePowerMonitoring; }
+  inline void     setPowerMonitoring(bool pm)   { _usePowerMonitoring = pm; }
```
DedeHai: why not name it enable?

wled00/json.cpp
```diff
@@ -704,8 +704,11 @@ void serializeInfo(JsonObject root)
   JsonObject leds = root.createNestedObject(F("leds"));
   leds[F("count")] = strip.getLengthTotal();
   leds[F("pwr")] = BusManager::currentMilliamps();
+  leds[F("pwr_w")] = BusManager::currentWatts();
```
DedeHai: name it "watts" to avoid confustion

wled00/set.cpp
```diff
@@ -147,6 +147,11 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
     unsigned ablMilliampsMax = request->arg(F("MA")).toInt();
     BusManager::setMilliampsMax(ablMilliampsMax);
 
+    uint8_t ledVoltage = request->arg(F("LV")).toInt();
+    if (ledVoltage > 0 && ledVoltage <= 50) BusManager::setVoltage(ledVoltage);
```
DedeHai: why restrict to 50?

wled00/data/settings_leds.htm
```diff
@@ -880,6 +895,7 @@ <h2>LED &amp; Hardware setup</h2>
 				<i>Make sure you enter correct value for each LED output.<br>
 				If using multiple outputs with only one PSU, distribute its power proportionally amongst outputs.</i><br>
 			</div>
+			<i class="warn">⚠ Automatic brightness limiter uses CPU resources. This can lower effect FPS on slower devices.</i><br>
```
DedeHai: keep text as brief as possible, it uses flash.

wled00/data/settings_leds.htm
```diff
@@ -478,9 +486,9 @@
 <option value="15">15mA (seed/fairy pixels)</option>
 <option value="0">Custom</option>
 </select><br>
-<div id="LAdis${s}" style="display: none;">max. mA/LED: <input name="LA${s}" type="number" min="1" max="255" oninput="UI()"> mA<br></div>
-<div id="PSU${s}">PSU: <input name="MA${s}" type="number" class="xl" min="250" max="65000" oninput="UI()" value="250"> mA<br></div>
+<div id="LAdis${s}" style="display: none;">Custom mA/LED: <input name="LA${s}" type="number" min="1" max="255" oninput="UI()"> mA<br></div>
```
DedeHai: revert unnecessary change

DedeHai (changes requested):

wled00/bus_manager.cpp
hntirgeam: changed

wled00/bus_manager.cpp
hntirgeam: Reverted changes, added "if", but now there is code duplication and no early return use (ok, I guess)

wled00/bus_manager.h
hntirgeam: removed

wled00/bus_manager.h
hntirgeam: changed

wled00/json.cpp
hntirgeam: changed

wled00/set.cpp
hntirgeam: I could only find 48v leds (UCS2903). So 50V seemed like an appropriate choice.

wled00/data/settings_leds.htm
hntirgeam: changed

wled00/data/settings_leds.htm
hntirgeam: done

wled00/set.cpp
DedeHai: if there is no technical reason, leave it to the user. there may be 120V strips one day, who knows.

wled00/bus_manager.cpp
```diff
@@ -1431,10 +1435,27 @@ void BusManager::applyABL() {
     }
     _gMilliAmpsUsed = milliAmpsSum;
   }
+  else if (_usePowerMonitoring) {
```
DedeHai: there is no need for code duplication, just adjust the logic above:
`if (_useABL || _usePowerMonitoring)`
and
` if (_gMilliAmpsMax > 0 && _useABL)`

wled00/data/settings_leds.htm
```diff
@@ -172,11 +180,10 @@
 		// enable and update LED Amps
 		function enLA(s,n)
 		{
-			const abl = d.Sf.ABL.checked;
 			const t = parseInt(d.Sf["LT"+n].value); // LED type SELECT
 			gId('LAdis'+n).style.display = s.selectedIndex==5 ? "inline" : "none"; // show/hide custom mA field
 			if (s.value!=="0") d.Sf["LA"+n].value = s.value; // set value from select object
-			d.Sf["LA"+n].min = (!isDig(t) || !abl) ? 0 : 1; // set minimum value for validation
+			d.Sf["LA"+n].min = (!isDig(t)) ? 0 : 1; // set minimum value for validation (required for power monitoring)
```
DedeHai: there is a reason the field is hidden if its not used - to not confuse unexperienced users (which there are a lot of)

wled00/bus_manager.cpp
hntirgeam: changed, but still needed to do this:

```
} else if (_usePowerMonitoring) {
      for (auto &bus : busses) {
        if (bus->isDigital() && bus->isOk()) {
          BusDigital &busd = static_cast<BusDigital&>(*bus);
          busd.applyBriLimit(255);
        }
      }
    }
```

because we need to reset _colorSum independently from abl process (it will accumulate over time making wrong calculations)

wled00/data/settings_leds.htm
hntirgeam: made it appear only when the abl/power fields are selected

wled00/set.cpp
hntirgeam: removed this check. In html bumped up to 255.

DedeHai:
@netmindz what is your opinion on this PR? IMHO this adds code that needs to be maintained, current estimation can be pulled from info endpoint already, adding LED voltage to settings purely for the purpose of HA convenience will confuse users as it has no other purpose.

netmindz:
Is there a matching PR for HA to actually use this value? I would say this PR is blocked unless that it going to be accepted, otherwise it serves no purpose. Also how does this differ from using PowerCalc in Home Assistant?

DedeHai:
> Also how does this differ from using PowerCalc in Home Assistant?

that is what I asked in the beginning:

>>Does HA now allow to calculate power from the ABL current estimate?

>>>No, not yet. I plan to implement it by adding power value to the WLED entity (but based on W, not mA).

>>>Actually, my first idea was to use pwr and just multiply it by a known voltage in HA and call it a day, but:

>>>in order to display pwr in the API, ABL must be enabled (which I don't use since I want max brightness and all my PSUs are powerful). It seems strange to me that I need to have ABL enabled (when I don't need it) in order to see pwr in the API

>>>this approach makes it much harder to integrate into HA as it requires quite a lot of manual configuration.

netmindz:
The WLED integration itself can't, but a plugin called PoweCalc can, for many devices, including WLED

DedeHai:
so no need to bloat wled code with something that can be done in HA directly then?

hntirgeam:
@DedeHai, @netmindz 

I see PowerCalc as a workaround, not a replacement for native support in WLED.

If power consumption is exposed natively, Home Assistant can use it via the standard WLED integration out of the box, without extra HACS integration, manual PowerCalc setup, or per device custom configuration.

It also avoids the current odd requirement to enable ABL just to get usable power data, even when you do not want to use this feature at all. Enabling ABL only for telemetry is less intuitive than having an explicit voltage/power report setting.

Goal here was to make native HA integration more capable and user friendly.

I will soon add support for this in the Home Assistant WLED integration, so this PR is not standalone.

netmindz:
Given HA are not responding to any PR requests for the python library that handles the communication with WLED, this PR will be blocked till the dependent PR is progressing

hntirgeam:
OK, I'll prepare a PR for HA WLED python lib and mention it here once there is progress.

PR #5268, opened 2026-01-03
New smooth effect: Color Clouds

JoaDick:
This PR adds a new effect called "Color Clouds", which is a very smooth effect. With its slowly moving clouds of rainbow color, it is very suitable for ambient lights and discreet table lamps.

It offers a variety of configuration options for the user:
- The speed of the cloud movement as well as the color change can be adjusted in a wide range, from almost unnoticeable to quite vibrant.
- The range of the rainbow spectrum on the strip, the size of the clouds and the distance between the clouds can also be widely adjusted.
- And finally there's a checkmark to put more emphasis on the red'ish colors, which can give the effect a "warmer" and cozy appearance.

The algorithm is very simple and uses perlin noise to generate the brightness- and color-gradients.

## Summary by CodeRabbit

* **New Features**
  * Added "Color Clouds" lighting mode: slow, cloud-like color diffusion with per-pixel hue and brightness variation. Includes adjustable speed, intensity, color and palette controls, visible in the effects list for easy selection, and built-in brightness safeguards for stable output.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -14,6 +14,8 @@
 #include "FX.h"
 #include "fcn_declare.h"
 
+#define FX_FALLBACK_STATIC { mode_static(); return; }
+
 #if !(defined(WLED_DISABLE_PARTICLESYSTEM2D) && defined(WLED_DISABLE_PARTICLESYSTEM1D))
   #include "FXparticleSystem.h" // include particle system code only if at least one system is enabled
   #ifdef WLED_DISABLE_PARTICLESYSTEM2D
@@ -80,12 +82,12 @@
 //#define MAX_FREQ_LOG10  3.71f
 
 // effect utility functions
-uint8_t sin_gap(uint16_t in) {
+static uint8_t sin_gap(uint16_t in) {
   if (in & 0x100) return 0;
   return sin8_t(in + 192); // correct phase shift of sine so that it starts and stops at 0
 }
 
-uint16_t triwave16(uint16_t in) {
+static uint16_t triwave16(uint16_t in) {
   if (in < 0x8000) return in *2;
   return 0xFFFF - (in - 0x8000)*2;
 }
@@ -97,7 +99,7 @@ uint16_t triwave16(uint16_t in) {
  * @param attdec attack & decay, max. pulsewidth / 2
  * @returns signed waveform value
  */
-int8_t tristate_square8(uint8_t x, uint8_t pulsewidth, uint8_t attdec) {
+static int8_t tristate_square8(uint8_t x, uint8_t pulsewidth, uint8_t attdec) {
   int8_t a = 127;
   if (x > 127) {
     a = -127;
@@ -131,20 +133,18 @@ static um_data_t* getAudioData() {
 /*
  * No blinking. Just plain old static light.
  */
-uint16_t mode_static(void) {
+void mode_static(void) {
   SEGMENT.fill(SEGCOLOR(0));
-  return strip.isOffRefreshRequired() ? FRAMETIME : 350;
 }
 static const char _data_FX_MODE_STATIC[] PROGMEM = "Solid";
 
 /*
  * Copy a segment and perform (optional) color adjustments
  */
-uint16_t mode_copy_segment(void) {
+void mode_copy_segment(void) {
   uint32_t sourceid = SEGMENT.custom3;
   if (sourceid >= strip.getSegmentsNum() || sourceid == strip.getCurrSegmentId()) { // invalid source
     SEGMENT.fadeToBlackBy(5); // fade out
-    return FRAMETIME;
   }
   Segment& sourcesegment = strip.getSegment(sourceid);
 
@@ -184,7 +184,6 @@ uint16_t mode_copy_segment(void) {
       }
     }
   }
-  return FRAMETIME;
 }
 static const char _data_FX_MODE_COPY[] PROGMEM = "Copy Segment@,Color shift,Lighten,Brighten,ID,Axis(2D),FullStack(last frame);;;12;ix=0,c1=0,c2=0,c3=0";
 
@@ -194,7 +193,7 @@ static const char _data_FX_MODE_COPY[] PROGMEM = "Copy Segment@,Color shift,Ligh
  * Alternate between color1 and color2
  * if(strobe == true) then create a strobe effect
  */
-uint16_t blink(uint32_t color1, uint32_t color2, bool strobe, bool do_palette) {
+void blink(uint32_t color1, uint32_t color2, bool strobe, bool do_palette) {
   uint32_t cycleTime = (255 - SEGMENT.speed)*20;
   uint32_t onTime = FRAMETIME;
   if (!strobe) onTime += ((cycleTime * SEGMENT.intensity) >> 8);
@@ -217,16 +216,14 @@ uint16_t blink(uint32_t color1, uint32_t color2, bool strobe, bool do_palette) {
       SEGMENT.setPixelColor(i, SEGMENT.color_from_palette(i, true, PALETTE_SOLID_WRAP, 0));
     }
   } else SEGMENT.fill(color);
-
-  return FRAMETIME;
 }
 
 
 /*
  * Normal blinking. Intensity sets duty cycle.
  */
-uint16_t mode_blink(void) {
-  return blink(SEGCOLOR(0), SEGCOLOR(1), false, true);
+void mode_blink(void) {
+  blink(SEGCOLOR(0), SEGCOLOR(1), false, true);
 }
 static const char _data_FX_MODE_BLINK[] PROGMEM = "Blink@!,Duty cycle;!,!;!;01";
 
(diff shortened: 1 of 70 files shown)


DedeHai:
thanks for contributing, I quite like the effect. Just a few notes:
- you really should use palettes instead of fixed colors
- do not use inoise, use perlin8 / perlin16 instead, inoise is just an alias for compatibility reasons
- dito for cos8(), use cos8_t()
- there is some overlap in functionality with my "shimmer" effect but I think it has still enoug unique features to be its own effect

DedeHai:
@coderabbitai when reviewing code, please check that the function aliases inoise8, inoise16 and similar fastled functions are not used and the perlin8/perlin16 etc are used instead.
the same goes for sin8(), cos8(), sinf() cosf() etc. which should be replaced by sin8_t / cos8_t  / sin16_t / cos16_t and for the float versions sin_approx and cos_approx. see wled_math.cp

wled00/FX.cpp
```diff
@@ -5031,6 +5031,79 @@ uint16_t mode_aurora(void) {
...
+      pixel = CRGB::Black;
+    }
+
+    SEGMENT.setPixelColor(i, pixel);
+  }
+
+  return FRAMETIME;
+}
+static const char _data_FX_MODE_COLORCLOUDS[] PROGMEM = "Color Clouds@Cloud speed,Color speed,Clouds,Colors,Distance,,,More red;;;;sx=24,ix=32,c1=48,c2=64,c3=12";
+
```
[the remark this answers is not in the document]
JoaDick: SEGMENT.custom3 is the slider with reduced range. Its highest possible value is 31, because it is a bitfield with 5 bits - so volCutoff can never be larger than 46000.
Btw: All these constants were determined experimentally to make the effect look appealing. While doing this, I've also checked the behaviour at the limits of all sliders.

JoaDick:
@DedeHai, thanks a lot for your feedback; glad to hear that you like the effect.

However, this comment astonished me: 
>     * you really should use palettes instead of fixed colors
The effect uses neither palettes nor fixed colors; it is based on HSV color calculations.
I thought that I disabled both with the configuration string - but apparently not!?
Could you give me a hint what is still missing in the string to disable the color sliders?

IIRC, in older versions of WLED (where i started developing this effect), the color sliders were not shown. If you're interested, I'll search in my old playgrounds.

DedeHai:
> The effect uses neither palettes nor fixed colors; it is based on HSV color calculations.

that is what I mean by "fixed colors" you really should use palettes - or use colorwheel() which uses HSV but still supports palettes but you can just make the default the "rainbow" palette and you do have your HSV equivalent in palette form.

JoaDick:
OK, got it and I'll give it a try.
Could you point me to an example where i can peek that usecase? The problem is that I never got my head around how this colorwheel & palette stuff is really working and how to use it properly.

I can remember that there have recently been changes regarding these rainbow colors and the color wheel. Unfortunately I didn't understand all the details from the discussion of that PR, so I got lost again with too many options. Trying to learn from examples of `Segment::color_from_palette()` with its many parameters is quite scary, especially when it is used in combination with these `PALETTE_..._WRAP` macros.

DedeHai:
instead of `CRGB pixel = CHSV(hue, 255, vol);` you either use:
color_wheel(hue) and then scale it down usig color_fade(), if you set "video" in color fade = true it will not scale to 0 if that's beneficial to your FX.
if you want palette colors only (colorwheel will use hsv2rgb if palette 0 is selected) you use 
`uint32_t pixel = SEGMENT.color_from_palette(hue, false, true, 0, vol)`

then adjust the metadata string, see any other effect that uses palette colors, shimmer for example.

JoaDick:
Implementing the usage of color palettes was a bit tricky, because their colors - and the ones from color_from_palette() - are much brighter than the colors from FastLED's CHSV(). That required quite some experimenting again with the constants, but finally the colors and clouds are behaving again as expected.
And the possibility to use palettes is really a great addition; thanks @DedeHai for your hints!

JoaDick:
@dedehai - Is there still anything missing in this PR?
From my point it is good to go.

wled00/FX.cpp
```diff
@@ -5063,10 +5066,9 @@ uint16_t mode_ColorClouds()
   const uint32_t hueSqueeze = SEGMENT.custom2;
 
   // Higher values make larger gaps between the clouds.
-  const long volCutoff = 2000 + SEGMENT.custom3 * 1250;
-
-  // Put more emphasis on the red'ish colors when true.
-  const bool moreRed = SEGMENT.check3;
+  const long volCutoff   = 12500 + SEGMENT.custom3 * 900;
```
DedeHai: avoid long (also in other places), use either int or better yet int32_t

wled00/FX.cpp
```diff
@@ -5075,33 +5077,35 @@ uint16_t mode_ColorClouds()
...
     }
 
-    CRGB pixel = CHSV(hue, 255, vol);
+    uint32_t pixel;
+    if(SEGMENT.palette) { pixel = SEGMENT.color_from_palette(hue, false, true, 0, vol); }
+    else { hsv2rgb(CHSV32(hue, 255, vol), pixel); }
+
     // Suppress extremely dark pixels to avoid flickering of plain r/g/b.
-    // Unfortunately this doesn't work anymore when gamma correction for color is enabled.
+    // Unfortunately this doesn't always work properly when gamma correction for color is enabled.
```
DedeHai: fix it, you have to assume gamma correction is always enabled.

wled00/FX.cpp
```diff
@@ -5075,33 +5077,35 @@ uint16_t mode_ColorClouds()
...
+      pixel = 0;
     }
 
     SEGMENT.setPixelColor(i, pixel);
   }
 
   return FRAMETIME;
 }
-static const char _data_FX_MODE_COLORCLOUDS[] PROGMEM = "Color Clouds@Cloud speed,Color speed,Clouds,Colors,Distance,,,More red;;;;sx=24,ix=32,c1=48,c2=64,c3=12";
+static const char _data_FX_MODE_COLORCLOUDS[] PROGMEM = "Color Clouds@Cloud speed,Color speed,Clouds,Colors,Distance,,,More red;;!;;sx=24,ix=32,c1=48,c2=64,c3=12,pal=0";
```
DedeHai: use "!" instead of Cloud speed. also calling it "more red" makes no sense with palettes.

wled00/FX.cpp
blazoncek: FYI `long` is guaranteed to be at least 32 bits by C and C++ standards. `int` is only guaranteed to be at least 16 bits (most often it is "native" or 32 bit sized).

DedeHai: that is true and the reason why I almost exclusively use uintxx_t as those are well defined.

blazoncek: Both `long` and `int` are *well* defined. However, their actual bit depth allows misuse in certain situations.

If you need a counter with at least 32 bit precision then you need to use `long` to be compatible on all platforms or, as you say, `int32_t` on compilers that support it (`#include <cstdint.h>`).

JoaDick: No problem; I usually also prefer the more descriptive variants. The long probably remainded, because the map() function below is declared with longs.

wled00/FX.cpp
JoaDick: Removed the comment. The problem vanished when the palettes were integrated.

wled00/FX.cpp
JoaDick: Calling it now "cozy" and adjusted the corresponding comment.
However, the "Cloud speed" should remain as is - to express the correlation between the Sliders "Cloud Speed & Clouds" and "Color Speed & Colors".

DedeHai: since the effect is called "Colored Clouds", what else would "Speed" mean (using ! will default to "speed")

JoaDick: I've set both labels to default; "effect intensity" for "Color Speed" also makes sense somehow.

wled00/FX.cpp
```diff
@@ -5031,6 +5031,83 @@ uint16_t mode_aurora(void) {
...
+    uint8_t hue = perlin16(hueX0 + hueX, hueT) >> 7;
+    hue += hueOffset0;
+    hue += hueOffset;
+    if (cozy) {
+      hue = cos8_t(128 + hue / 2);
+    }
+
+    uint32_t pixel;
+    if(SEGMENT.palette) { pixel = SEGMENT.color_from_palette(hue, false, true, 0, vol); }
+    else { hsv2rgb(CHSV32(hue, 255, vol), pixel); }
```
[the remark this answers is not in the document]
JoaDick: The constructor takes 3 uint8_t - see colors.h:88

JoaDick:
@DedeHai Updated this PR to latest main and removed `return FRAMETIME;`. Changing the returnvalue to `void` was a really great idea, btw!
Is there still anything left that should be fixed before merge?

DedeHai:
Still need to test the final version, will get back on this soon

PR #5269, opened 2026-01-03
Add support for TM1815

DedeHai:
This adds custom NPB timings and methods to support TM1815.

@willmmiles this also paves the way to add more custom timings. What do you think of this approach and the file structure I chose?

## Summary by CodeRabbit

* **New Features**
  * Added support for TM1815 RGBW LED chipset across supported output methods.

* **User Interface**
  * TM1815 now appears alongside other LED types in selection lists.

* **Settings**
  * Per-LED “Refresh off” requirement now applies to TM1815 as well as TM1814.

* **Chores**
  * LED interface identifiers and public configuration entries updated to include TM1815 and clarified TM1814/TM1829 naming.


DedeHai:
converting to draft. these timings should still be added but the approach take adds another arm to the kraken, there must be a better way to do this.

Issue #5275, opened 2026-01-05
presets corrupted after saving without "Overwrite with state" (esp32, needs large presets file or slow flash mode)


softhack007:
There is an old bug in the presets saving logic that can lead to presets.json corruption, in case that the user's preset file is relatively big, or flash writing is slow (like 40Mhz dout).

### How to reproduce
* you need a larger presets file; my testcase was a 200kB presets.json
* choose a preset randomly:
  * modify the name, set "overwrite with state", save
  * quickly modify the name again (make it longer), this time leave "overwrite with state" unchecked, save
  * ==> good chance that your presets.json is corrupted now (reload webUI with ``CTRL+F5`` to verify).
  * if your presets file has survived, repeat the above procedure but this time "delete" the preset as the last step.

### preliminary analysis
After a [root cause analysis in the WLED-MM fork](https://github.com/MoonModules/WLED-MM/pull/293#issuecomment-3688014587), it turned out that the code in presets.cpp (called from the preset De-Serializer functions in json.cpp, or from ) can lead to concurrent writing of the presets file on esp32. The same code is still preset here in upstream WLED.

As a secondary root cause, the function writeObjectToFile() can re-assign the global ``File f`` even when the previous ``File`` had writes pending. This leads to "holes" in the presets file.

<br/>

https://github.com/wled/WLED/blob/e95450b318ffc6178ecb07f6215804542fa0f9d6/wled00/presets.cpp#L251-L266

(this even has a comment "often causes presets.json corruption" that's more than 4 years old, but still no solution available for esp32)

https://github.com/wled/WLED/blob/e95450b318ffc6178ecb07f6215804542fa0f9d6/wled00/file.cpp#L276

(missing:``  if (doCloseFile) closeFile(); ``).

### Solution

1/ fix writeObjectToFile() to always close the previous file handle, before re-assigning the file handle with WLED_FS.open()
    -> bugfix in https://github.com/wled/WLED/pull/5276

2/ prevent concurrent file writing from doSaveState(), savePreset() and deletePreset(). The existing mutex code in requestJSONBufferLock() protects the in-memory global JSON buffer, but it cannot protect the prests.json file itself when writes are still pending after releasing the buffer lock, or when a local JSON buffer was used.

[...]

PR #5276, opened 2026-01-05
prevent file data loss due to replacing an open file handle (partial fix for #5275)

softhack007:
The function ``writeObjectToFile()`` was re-assigning ``f`` (global file handle, file.cpp) without first completing any pending write operations on f. 

Reassigning a file pointer thats still in use can lead to loss of data and file corruption.
This PR provides a fix by first checking for pending writes (doCloseFile == true) and calling f.close() when needed.

The same protection code is already preset in `readObjectFromFile()`, it looks like protecting ``writeObjectToFile()`` was forgotten.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file-handling reliability by ensuring files are properly closed when needed.
  * Ensured cached file data is flushed before files are reopened or written, reducing risk of data loss and improving consistency.

### DIFF
--- a/wled00/file.cpp
+++ b/wled00/file.cpp
@@ -38,7 +38,7 @@ void closeFile() {
     DEBUGFS_PRINT(F("Close -> "));
     uint32_t s = millis();
   #endif
-  if (f) f.close(); // only close if we have an open file
+  f.close(); // "if (f)" check is aleady done inside f.close(), and f cannot be nullptr -> no need for double checking before closing the file handle.
   DEBUGFS_PRINTF("took %lu ms\n", millis() - s);
   doCloseFile = false;
 }


wled00/file.cpp
```diff
@@ -38,7 +38,7 @@ void closeFile() {
     DEBUGFS_PRINT(F("Close -> "));
     uint32_t s = millis();
   #endif
-  f.close();
+  if (f) f.close(); // only close if we have an open file
```
DedeHai: I think this check is already done inside of f.close() but wont hurt to double check.

DedeHai: I checked and yes, this is redundant.

softhack007: @DedeHai thanks for checking, then I'll remove this one.

DedeHai: maybe add a comment, the rabbit also complained about files being closed without checking at one point. that is how I knew.

softhack007: @coderabbitAI please remember that in the situation we have here - ``File f`` cannot be `nullptr`, and ``if (f)`` is just checking if the file handle is an open file - there is no need for an explicit "if (f)" check, because the same check is performed again inside the "f.close()" function.

Issue #2546, opened 2022-02-16
[feature request] Apply most recent time-controlled preset at boot


Jzege:
WLED checks every minute for a matching time-controlled preset.
However, I don't leave my LED strip turned on 24/7.
For example, if I set a timer at 6pm, but only power up the LED strip at say 6.30pm, the LED strip will not turn on until the next night, when I really would like it to turn on immediately.

I would like to suggest adding an option in the GUI:
![image](https://user-images.githubusercontent.com/6236352/154263682-6076ad10-70ec-47ef-89e6-8b838de87bc6.png)

And modifying checkTimers() like so:
```
void checkTimers()
{
  if (lastTimerMinute != minute(localTime)) //only check once a new minute begins
  {
    lastTimerMinute = minute(localTime);
    int time_delta = 1440;  // minutes in a day
    int16_t nextPreset = currentPreset;
    for (uint8_t i = 0; i < 8; i++)
    {
      if (timerMacro[i] != 0
          && (timerWeekday[i] & 0x01) //timer is enabled
          && timerWeekday[i] >> weekdayMondayFirst() & 0x01  //timer should activate at current day of week
          && (timerHours[i] <= hour(localTime))
          && ((60*(hour(localTime) - timerHours[i]) + minute(localTime) - timerMinutes[i]) <= time_delta) // find the most recent timer compared to localTime
          )
      {
          time_delta = 60*(hour(localTime) - timerHours[i]) + minute(localTime) - timerMinutes[i];
          nextPreset = timerMacro[i];   
      }
    }
  if (nextPreset != currentPreset) applyPreset(nextPreset);
  }
}

```

I would take a stab at it, except I'm still on v0.12.0 and the main branch doesn't work for me.

Aircoookie:
Thank you for the suggestion!
A similar functionality has been requested in the forum as well recently.
It shouldn't be too hard to add - just figure out the preset that would have most recently been applied once you have a valid time sync (commonly via NTP, but RTC or UI are also possible) and apply it if the option is set.
Not the highest priority for me, but I'll see when I can get it added :)

blazoncek:
Do not forget that, for example, a playlist may be scheduled at 6:00 PM and play only for 15 min. Then stop and not do anything.
In such case starting a playlist at 6:30 PM may be unwanted.

jmerc77:
[...]

PR #5297, opened 2026-01-15
Apply most recent time-controlled preset on boot

kylephillipsau:
Adds an option to apply the most recent scheduled preset when the device boots after NTP sync.

Currently, timed presets only trigger at the exact scheduled minute. If you power on at 9pm with an 8pm preset, nothing happens until 8pm the next day. This checks what preset should be active and applies it.

New checkbox in Time & Macros: "Apply scheduled preset on boot" (enabled by default).

Also refactored checkTimers() to share validation logic with the new code.

My use case: WLED bulbs on a wall switch that I wanted to go red after 8pm to reduce melatonin suppression before sleep.

Addresses #2546

## Summary by CodeRabbit

* **New Features**
  * Added an "Apply scheduled preset on boot" checkbox in Time settings to optionally apply the most recent scheduled preset after a restart (off by default).

* **Improvements**
  * Unified timer evaluation across all timer types; after time sync on boot, the device can detect and apply the most recent missed preset for today (includes hourly and sunrise/sunset presets).

### DIFF
--- a/wled00/ntp.cpp
+++ b/wled00/ntp.cpp
@@ -292,6 +292,8 @@ bool checkNTPResponse()
   // if time changed re-calculate sunrise/sunset
   updateLocalTime();
   calculateSunriseAndSunset();
+  // Apply most recent timer preset on first NTP sync after boot
+  applyBootTimerPreset();
   return true;
 }
 
@@ -376,6 +378,39 @@ bool isTodayInDateRange(byte monthStart, byte dayStart, byte monthEnd, byte dayE
 	return (m == monthStart && d >= dayStart && d <= dayEnd); //just the designated days this month
 }
 
+/*
+ * Returns the minute-of-day (0-1439) for timer i if it's valid for today, or -1 if invalid.
+ * Validates: preset assigned, timer enabled, weekday matches, date range (for timers 0-7).
+ * For "every hour" timers (hour=24), returns the current hour's scheduled minute.
+ * For sunrise/sunset timers (8/9), returns the calculated trigger time.
+ */
+static int getTimerMinuteOfDay(unsigned i)
+{
+  if (i > 9) return -1;
+  if (timerMacro[i] == 0) return -1;
+  if (!(timerWeekday[i] & 0x01)) return -1;  // not enabled
+  if (!((timerWeekday[i] >> weekdayMondayFirst()) & 0x01)) return -1;  // wrong weekday
+
+  if (i < 8) {
+    // Standard timer (0-7)
+    if (!isTodayInDateRange(((timerMonth[i] >> 4) & 0x0F), timerDay[i], timerMonth[i] & 0x0F, timerDayEnd[i])) return -1;
+    if (timerHours[i] == 24) {
+      return hour(localTime) * 60 + timerMinutes[i];  // "every hour" at this minute
+    }
+    return timerHours[i] * 60 + timerMinutes[i];
+  } else if (i == 8) {
+    // Sunrise timer
+    if (!sunrise) return -1;
+    time_t t = sunrise + timerMinutes[8] * 60;
+    return hour(t) * 60 + minute(t);
+  } else {
+    // Sunset timer (i == 9)
+    if (!sunset) return -1;
+    time_t t = sunset + timerMinutes[9] * 60;
+    return hour(t) * 60 + minute(t);
+  }
+}
+
 void checkTimers()
 {
   if (lastTimerMinute != minute(localTime)) //only check once a new minute begins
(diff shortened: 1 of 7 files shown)


wled00/ntp.cpp
```diff
@@ -376,6 +378,39 @@ bool isTodayInDateRange(byte monthStart, byte dayStart, byte monthEnd, byte dayE
 	return (m == monthStart && d >= dayStart && d <= dayEnd); //just the designated days this month
 }
 
+/*
+ * Returns the minute-of-day (0-1439) for timer i if it's valid for today, or -1 if invalid.
+ * Validates: preset assigned, timer enabled, weekday matches, date range (for timers 0-7).
+ * For "every hour" timers (hour=24), returns the current hour's scheduled minute.
+ * For sunrise/sunset timers (8/9), returns the calculated trigger time.
+ */
+static int getTimerMinuteOfDay(unsigned i)
```
softhack007, line 387: This might already be available. Please check if this function works for you

https://github.com/wled/WLED/blob/d1d9dec40234b81ec4ec79da43d4eb36cb1e31cf/wled00/src/dependencies/time/TimeLib.h#L79

wled00/ntp.cpp
```diff
@@ -385,49 +420,61 @@ void checkTimers()
...
+  int currentMinuteOfDay = hour(localTime) * 60 + minute(localTime);
+  int latestTimerMinute = -1;
+  byte latestPreset = 0;
+
+  DEBUG_PRINTLN(F("Checking for boot timer preset..."));
+
+  for (unsigned i = 0; i < 10; i++)
+  {
+    int timerMinute = getTimerMinuteOfDay(i);
+    if (timerMinute < 0) continue;
```
softhack007, line 456: why would "minutes of today" be nagative?

wled00/ntp.cpp
kylephillipsau: Good catch, thank you\! I'll switch to elapsedSecsToday(localTime) / 60.

wled00/ntp.cpp
kylephillipsau: The -1 is a sentinel value from getTimerMinuteOfDay() meaning "timer not valid for today" (disabled, wrong weekday, etc). I'll add a comment to make that clearer.

softhack007:
> > > `453-468`: **Clarify tie-breaking behavior for simultaneous timers.**
> > > When multiple timers are scheduled for the same minute, the current loop selects the last one iterated (highest index). This is a valid approach, but consider whether this matches user expectations or if it should match `checkTimers()` behavior (which would apply all matching timers).
> > > If the current "last wins" behavior is intentional, a brief comment would help future maintainers understand the design choice.

@kylephillipsau please also check if this comment by the rabbit makes sense.

kylephillipsau:
That makes sense, thanks for letting me know. I have updated `>` to `>=` so the highest index wins on ties, which matches the `checkTimers()` behavior. I have included a comment alongside the function to explain why.

DedeHai:
seems like a good addition to have. It should not be enabled by default for backwards compatibility though.
Also know that once https://github.com/wled/WLED/pull/5140 is merged, this will need quite some rework.

kylephillipsau:
That's a fair point, thank you. I just changed the default value to false in b24e7db.
I had a squiz over #5140 . The core logic in `applyBootTimerPreset()` should translate fairly directly to iterating the new Timer vector instead of the current arrays. I'm happy to rebase and adapt this once that's merged.

wled00/ntp.cpp
```diff
@@ -292,6 +292,8 @@ bool checkNTPResponse()
   // if time changed re-calculate sunrise/sunset
   updateLocalTime();
   calculateSunriseAndSunset();
+  // Apply most recent timer preset on first NTP sync after boot
+  applyBootTimerPreset();
```
softhack007, line 296: I'm not sure if this does the right thing. 

The above comment says "if time changed". This is not the same as "on first ntp sync".
Time can also change when
- location has changed
- timezone has changed
- different ntp server selected, or  ntp was disabled by users
- ntp re-sync (after 8 hours) has corrected the local time
- another time source (RTC usermod, or webUI callback with a timetag) have changed the time.

kylephillipsau:
Hey @softhack007,

Thank you for bringing that up. I traced through the codebase to verify the behavior.

`checkNTPResponse()` only executes when a valid NTP UDP packet is received and
  validated. The scenarios you mentioned each go through different code paths:

  - Location/timezone changed - This is handled in `set.cpp`, it forces NTP resync via
  `ntpLastSyncTime = NTP_NEVER`
  - RTC usermod - calls `toki.setTime()` directly in setup, which never touches
  `checkNTPResponse()`
  - JSON/webUI time set - uses `setTimeFromAPI()`, a separate path in `ntp.cpp`
  - NTP re-sync - the `bootTimerApplied` guard flag prevents re-execution, which follows the same pattern as `countdownOverTriggered`

  Since only the NTP packet receipt triggers this path, and the guard flag handles re-syncs, the implementation should behave correctly.

  I updated the comment to read "NTP sync succeeded" which I think captures the intent more clearly.

softhack007 (approved):
Looks good to me - no more questions ;-) I did not find time to test the code on hardware, though.

softhack007:
@coderabbitai review

softhack007:
> @[kylephillipsau](https://github.com/kylephillipsau) kylephillipsau [force-pushed](https://github.com/wled/WLED/compare/5084308733f91d77d340cda358353b4a8591fbd7..32bed6682ca8adb175ec26987a961386bb60fa4a) the feature/apply-timer-preset-on-boot branch from [5084308](https://github.com/wled/WLED/commit/5084308733f91d77d340cda358353b4a8591fbd7) to [32bed66](https://github.com/wled/WLED/commit/32bed6682ca8adb175ec26987a961386bb60fa4a) 
[2 days ago](https://github.com/wled/WLED/pull/5297#event-22067060387)

@kylephillipsau on last thing - for your next PR: [please avoid ``force-push`` while you have a PR open](https://github.com/wled/WLED?tab=contributing-ov-file#updating-your-code) for upstream WLED. We lost the complete review history that came before your force-push (in this PR only), plus coderabbitai just self-disabled for some strange reason and it does not respond to my prompts any more.

<br>

Edit: we can live with the "damage" that was done by accident in this PR, as I think its good for merging. Just for the future please remember "force-push considered harmful" 😉

blazoncek:
What about playlists specifically crafted to apply certain presets at defined times after a trigger time? Those would be broken.

I.e. playlist to apply preset/effect A at 7:00 with duration TA then preset/effect B with duration TB, ending with preset C. One may not want this sequence to play at 8:00, only preset C would be wanted.

kylephillipsau:
Yeah, that's a really good point. As you noted in #2546, playlist content only exists in the `presets.json`, which isn't feasible to read and parse to determine playlist behavior at boot time. If a timer triggers a playlist, it would start from the beginning rather than resume mid-sequence.

Since this feature is disabled by default, users with time-sequenced playlists can leave it off. For the simpler use cases this targets (e.g. warm white after sunset and red at night), it works as expected.

If this behavior is acceptable, I can add a note in the UI description to clarify it. Otherwise if it feels incomplete, I'm happy to defer until a better way to address this can be identified.

blazoncek:
> read and parse to determine playlist behavior at boot time

Why not? My comment was made 3 years ago when I was still fresh with WLED. Playlists are actually simple to parse. Each entry has duration assigned (most often but there may be exceptions) so you sum durations until time is greater than current time.

The only drawback I see is that if a user has boot preset assigned, it will start, only to be replaced a few seconds later by "last timed preset". The "few seconds later" may be very variable as you never know how long will it take to connect to WiFi and receive NTP response. A bit awkward and solvable by removing boot preset, but still. The reality is that user may not *always* want last timed preset to play (emphasis on *always*). I.e. I may not want to play last timed preset in the morning but would want to in the afternoon.

> users with time-sequenced playlists can leave it off

Let's assume users want the behaviour but have mixed presets and playlists scheduled. Life is not simple, unfortunately.

There are of course other arguments "pro et contra" so the choice to enable this behaviour should definitely be in the hands of enduser. In general I am in favour of this PR but would really like to see most aspects to be considered and implemented correctly if possible.

It would also be beneficial if this PR be combined with #5140 .

DedeHai:
maybe instead of just having a checkmark: add a value to that like "apply most recent time-preset if less than xxx minutes ago (0=always)" ?

kylephillipsau:
Thanks for the feedback @blazoncek and @DedeHai.

I'm happy to wait until #5140 is finalised so I can integrate this with the new macro/timer slot implementation. I'll look into implementing playlist handling, since parsing presets.json definitely seems doable.

@DedeHai I really like the idea of adding a configurable time window instead of a simple checkbox. It would address the concern @blazoncek raised about not always wanting the last timed preset to apply, and gives users much more control over the behavior.

Regarding the boot preset interaction, I've had this flashed on my ceiling bulbs for about a month now. With the boot preset set to "off" and a scheduled warm-white/red preset, the latency between power-on and NTP sync applying the timed preset is only a few seconds at most in my environment. It isn't bad, but it's noticeable. I think the configurable time window would help, since users who find the transition awkward can tune the window to match their use case, or just remove the boot preset. I'm open to ideas if anyone has thoughts on making this smoother.

I'll hold off on further changes here until #5140 lands, then rebase and rework this with playlist support and the time window option.

Thank you for your patience, I'm starting a new job soon so there's a lot winding down and ramping up at the moment. And thank you all for your contributions to WLED, being able to programmatically set my lights to red at night is genuinely helping me sleep better :)

PR #5311, opened 2026-01-23
Switch to out of tree AudioReactive Usermod

netmindz:
The AudioReactive usermod is maintained by the MoonModules team, so rather than trying to maintain in one repo and then endless PRs to keep other forks up to date, use the new out-of-tree usermod support to take it out of the tree and WLED-MM can just add the repo to the include path for the v14 style

## Summary by CodeRabbit

* **Refactor**
  * Centralized the audio‑reactive usermod reference in build configs to a common default.
  * Added an extra helper overload in the usermod API for improved logging/printing.

* **Removed**
  * Deleted the Audio Reactive usermod package, its build manifest, build tweak script, and README.

* **New Features**
  * Added two error codes for reboot or power‑cycle required after hardware changes.
  * Enhanced settings UI with new pin/option helper functions and pin list initialization.


netmindz:
aOpt is missing in upstream

netmindz:
Screenshot From 2026-01-31 17-52-43

netmindz:
Screenshot From 2026-01-31 17-54-20

wled00/fcn_declare.h
```diff
@@ -332,6 +332,7 @@ class Usermod {
   protected:
     // Shim for oappend(), which used to exist in utils.cpp
     template<typename T> static inline void oappend(const T& t) { oappend_shim->print(t); };
+    template<typename T> static inline void oappendi(const T& t) { oappend_shim->print(t); };
```
netmindz: @willmmiles  - possibly not needed, just change the usermod

WLED-MM has 

```
bool oappendi(int i)
{
  char s[16];               // WLEDMM max 32bit integer needs 11 chars (sign + 10) not 10
  snprintf(s, 15, "%d", i); // WLEDMM
  return oappend(s);
}
```

platformio.ini
```diff
@@ -391,7 +392,7 @@ extends = env:nodemcuv2
 board_build.f_cpu = 160000000L
 build_flags = ${common.build_flags} ${esp8266.build_flags} -D WLED_RELEASE_NAME=\"ESP8266_160\" #-DWLED_DISABLE_2D
   -D WLED_DISABLE_PARTICLESYSTEM2D
-custom_usermods = audioreactive
+custom_usermods =
```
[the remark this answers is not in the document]
netmindz: It is  not being removed, rather it's using the ability to reference out-of-tree usermods so wled-audioreactive being added as a library instead means that it is still included. We might look to rename custom_usermods to something like custom_usermods_enabled to help clarify this relates to in-tree mods. Do you have any other suggestions for naming?

wled00/fcn_declare.h
willmmiles: Ah, the usual API drift between forks ;)

Yes, in a perfect world I'd suggest a local shim in the usermod.  That said, I wouldn't be opposed to a "MM usermod compatibility" PR, since we wanted to add global name support anyways.

netmindz: To be honest, I'm out of my depth with C when it comes to templates, so I have no idea if what I have written is valid, I was just trying to get it to compile. The comment suggests that oappend is kinda legacy, so is there something better the usermod should use?

platformio.ini
```diff
@@ -280,7 +281,7 @@ extreme_partitions = tools/WLED_ESP32_16MB_9MB_FS.csv
 board_build.partitions = ${esp32.default_partitions}   ;; default partioning for 4MB Flash - can be overridden in build envs
 # additional build flags for audioreactive - must be applied globally
 AR_build_flags = ;; -fsingle-precision-constant ;; forces ArduinoFFT to use float math (2x faster)
-AR_lib_deps =  ;; for pre-usermod-library platformio_override compatibility
+AR_lib_deps =  wled-audioreactive = https://github.com/MoonModules/WLED-AudioReactive-Usermod
```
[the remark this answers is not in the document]
netmindz: @coderabbitai Yes you are correct, this is the pattern we use, however this PR is going side by side with the creation of that repo, so we don't want to keep updating this PR after every commit to the usermod, as the only way to test the usermod is through it's usage here

platformio.ini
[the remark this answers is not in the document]
netmindz: @coderabbitai  unfortunately, I think it may be a platformio restriction that we must name our value with custom_ as the prefix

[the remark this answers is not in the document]
netmindz: What are your thoughts on naming @willmmiles ?

wled00/fcn_declare.h
willmmiles: The intent was to change to a new API, passing an Arduino `Print&` directly to the usermod code so we could completely eliminate buffer size limitations.   [Updated usermods implement `void appendConfigData(Print& uiScript)` instead.](https://github.com/wled/WLED/blob/1ca55e42af6e2a968c28a048e21cc0069f7e9bd1/usermods/audioreactive/audio_reactive.cpp#L1900C10-L1900C26)   The relevant PR was #4152 .  

The quick shim would be to do something like this in `audio_reactive.h`:
```
#ifndef WLEDMM   // or whatever other define is unique to MM
#define oappendi oppend
#endif
```
... since the current oappend shim here is guaranteed to be length safe for all integer values.

platformio.ini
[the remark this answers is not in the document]
willmmiles: Sadly, yes, platformio does require names to start with `custom_` ... I spent entirely too long trying to find some way to hack around that since I though it was so obnoxious.

I have no strong objections to changing the property name: it's easy enough to carry forward support for the old one as well for some period of time.  That said, I'd been experimenting with going the other way: finding some way to make `custom_usermods` accept `lib_deps` syntax so we could just put any or all modules there.  Would that be better?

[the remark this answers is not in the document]
netmindz: That sounds better, it would be much nicer to have a single field that lists all the usermods, with just a different format of name = url for the external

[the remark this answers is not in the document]
willmmiles: OK, I'll get that done, then.

DedeHai:
I gave this a go and it does compile and most importantly, the tested I2S mic works. Did not poke around much though. 

- when switching to this branch then back to main, I need to do a full clean or it won't compile (complains about missing UM's as well as arduinoFFT)
- it adds a hefty 11k to the flash as well as a few hundered bytes of static RAM use - definitely need to slim it down for AC by making a lot of the advanced MM stuff optional (mic filters, FFT windows, ...) @softhack007 do you happen to know which features are the most flash heavy?

netmindz:
Thanks for testing @DedeHai 

We probably need one or more issues opened against the usermod repo itself, discussion on this PR should relate just to the switching to out of tree, rather than about the usermod itself

netmindz:
Why did you mark this as only a draft @softhack007, due what is present in this change itself or more a refection on the out of tree mod being a draft?

netmindz:
> I gave this a go and it does compile and most importantly, the tested I2S mic works. Did not poke around much though.
> 
> * when switching to this branch then back to main, I need to do a full clean or it won't compile (complains about missing UM's as well as arduinoFFT)

Yeah I've seen similar when switching between 0_15 and main due to the platformio dependency management 

> * it adds a hefty 11k to the flash as well as a few hundered bytes of static RAM use - definitely need to slim it down for AC by making a lot of the advanced MM stuff optional (mic filters, FFT windows, ...) @softhack007 do you happen to know which features are the most flash heavy?

Please add you thoughts about which bits you think are advanced features that should be optional in the build to

https://github.com/MoonModules/WLED-AudioReactive-Usermod/issues/2

Or any other topics about the content of the usermod as an issue or PR assist that repo please

DedeHai:
I ran some compile tests: the new mic filter-table use ~3k of flash (no idea why so much but they do)
The FFT windows do not appear to use any flash, looks like they are created on the fly in RAM. 
New audio sources add 1.5k to flash use, could ont find out where the rest (about 5k worth) is coming from, maybe from advanced algorithms. 

edit:
moving this to correct issue

netmindz:
@DedeHai please see my previous comment and comment on the correct issue. This is about the removal of the usermod and any changes the core of WLED required to support AR being out of tree. Any changes to the usermod should be against the usermod or conversation is going to get fragmented

DedeHai:
@netmindz I think a good way foward is to merge this after resolving conflicts and proceed with updating the usermod. Since this is the way to go, why procrastinate. What is your opinion @softhack007 ?

platformio.ini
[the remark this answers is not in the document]
willmmiles: PR with lib_deps syntax in #5403

netmindz:
Possibly after merging https://github.com/wled/WLED/pull/5403 but not a blocker. Can always update after that merge

netmindz:
The only downside I see to swapping before the "striping back" might mean users of the nightly start playing around with the advanced controls, then if we remove some, they can't change back to any other value, so might be stuck with an experimental feature enabled, unless they edit the config by hand

DedeHai:
that's a good point. I do have the ifdefs ready locally, just need to decide on opt in or opt out and the actual define's name. Then I will fork and do a PR

netmindz:
@copilot resolve the merge conflicts in this pull request

wled00/fcn_declare.h
netmindz: that just then moans that oppend is missing if i try and add that to the usermod

netmindz:
I plan to merge this straight after the announcement regarding v17 dev as AR is no longer going to be in-tree for 17 and any changes need to be to the out tree version @DedeHai  @softhack007

wled00/fcn_declare.h
netmindz: was just a typo in your example

willmmiles: > was just a typo in your example

Oops, sorry!  Glad you got it sorted.

platformio.ini
```diff
@@ -91,6 +91,9 @@ build_flags =
   -DWLED_USE_MY_CONFIG
   -D WLED_PS_DONT_REPLACE_FX ; PS replacement FX are purely a flash memory saving feature, do not replace classic FX until we run out of flash
 
+default_usermods =
+  audioreactive = https://github.com/MoonModules/WLED-AudioReactive-Usermod
```
willmmiles, line 95: Do we want to pin a specific version or commit?  Otherwise old builds become non-reproducible.

netmindz: Good shout, we should pin

DedeHai:
I have a working version of the OOT usermod on V5 but it requires a lot more review and testing. I had an AI merge my changes from current main and after a few minor fixes it works, tested on C3 and C6 so good progress. There is one compatibility issue though as MM uses different debug print macros, not sure how we want to tacke those - either change the macros in AC or add a compatibility shim.

`https://github.com/DedeHai/WLED-AudioReactive-Usermod/tree/V5_and_integermath`

use this to include it in AC builds:
`custom_usermods = https://github.com/DedeHai/WLED-AudioReactive-Usermod.git#V5_and_integermath`

What was updated:
- added #ifdefs to include advanced MM features, in AC use `-D MM_AR_NO_ADVANCED_FEATURES` to disable them (saves ~6k of flash)
- add integer math and ESP-DSP FFT support
- port V5 update from AC
- split audio source file into base class, V4 and V5 files

at the time of writing this, it is mostly unreviewd AI code, all I did so far is test with analog mic's on C3 and C6 and confirmed it working.

next steps are:
- fix or shim debug macros
- do a thorough review of the file split to check for AI slop
- test if it still works in MM (@softhack007 please help me out if you can, I do not really use MM)
- add ADC manager for V5 analog mic's once that PR is merged int AC
- PR to OOT AR UM

Issue #5312, opened 2026-01-24
Relay not switching on at power-up with boot-preset


DedeHai:
### What happened?

At bootup, the relay pin is not switched on (or not set as an output) if "power on at boot" is not set even if the boot-up preset uses `"on":true` 
After switching off the LEDs and switching them back on, it works as intended. 
This is confusing as the UI shows everything working and the LEDs being on, even data is sent to the LEDs but they are not powered as the relay stays off.

### To Reproduce Bug

disable "power on at boot" in settings.
set relay pin for example to GPIO25, power on the MCU -> relay does not switch on.

### Expected Behavior

when the UI shows the LEDs are on and data is sent, the relay should also be on.

### Install Method

Self-Compiled

### What version of WLED?

0.16

### Which microcontroller/board are you seeing the problem on?

Other

### Relevant log/trace output

```shell
C3 and ESP32 classic tested.
```

### Anything else?

maybe a comment should be added to the settings that "Turn LEDs on after power up/reset" will be overridden if the boot-preset has "on" set to true.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

blazoncek:
FYI all but a few of my devices are equipped with relay and have "turn on at boot" off and they correctly switch relay either by pressing the power button or by starting a preset.

If yours are not working you must have made changes in 0.16 to break things.

And BTW boot preset always takes precedence over "turn on at boot" setting. It has always been so.
However if you have a boot preset without "on":true or "on":false that toggle is respected. This leaves some creative freedom when creating boot presets.

DedeHai:
Some users reported this and in my hardware tests this bug led to some head-scratching so I checked and the behaviour is indeed different to what it was. I also saw some cases where it would switch but a bit erratic.

DedeHai:
I did a bi-section, this PR breaks it: https://github.com/wled/WLED/pull/4658

blazoncek:
Not in my case. See also #4474 where actual change of relay behaviour was done.

EDIT: Adding `delay(50)` is not a proper way to postpone LED output.

DedeHai:
something weird is going on, it would sometimes work and sometimes not. I suspect some timing issue but in general the code sequence is correct.

softhack007:
@DedeHai I didn't look at the code in detail, but there was some comment (near the relay code) hinting at transitions.

[...]

Issue #3581, opened 2023-12-12
blink random color when turning on with relay (using SSR to control AC power to PSU)


doronazl:
### What happened?

im using a SSR to control my PSU and turn it off when not in use, it works fine but im getting a blink sometimes when turning the led on, been reading about it and found this thread
https://github.com/Aircoookie/WLED/issues/812
seems this was solved there but trying to edit the changes proposed i find things are different in the current code so im stuck

how do i get this done , id like to add a 200ms delay 

### To Reproduce Bug

turning on with a relay

### Expected Behavior

described

### Install Method

Binary from WLED.me

### What version of WLED?

0.14

### Which microcontroller/board are you seeing the problem on?

ESP32

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

doronazl:
ive also tried the multirelay usermod, having problems with it, i set the gpio relay and hit save, it saves but after a refresh the gpio pin return to unused, every numeric one act this way, the only ones that actually save are P0-P7, which i dont think available for my board, esp32wroom

softhack007:
@doronazl could it be you are regularly deleting / changing your questions? I got some questions by e-mail but they are missing here, so hard to help you.

About multi-relay:
* the delay time is an integer (whole number) for seconds. So you can have 0sec, 1sec, 42sec.  But you cannot use fractions like 0.34567 sec.  If you want finer control, either modify the source code, or make a feature request.
* (deleted question) if you enter GPIO pins in the usermod, make sure to not use the same pin anywhere else. So in LED settings, you have to configure the relay pin as unused (-1).
* (deleted question 2) P0...P7 are for the port expander, if you use one. Otherwise don't use P0 ... P7.

doronazl:
> @doronazl could it be you are regularly deleting / changing your questions? I got some questions by e-mail but they are missing here, so hard to help you.
> 
> About multi-relay:
> 
> * the delay time is an integer (whole number) for seconds. So you can have 0sec, 1sec, 42sec.  But you cannot use fractions like 0.34567 sec.  If you want finer control, either modify the source code, or make a feature request.
> * (deleted question 1) if you enter GPIO pins in the usermod, make sure to not use the same pin anywhere else. So in LED settings, you have to configure the relay pin as unused (-1).
> * (deleted question 2) P0...P7 are for the port expander, if you use one. Otherwise don't use P0 ... P7.

My apologies, some issue i got resolved so i just deleted it.
[...]

Issue #812, opened 2020-03-28
Blink on startup,only with relay


Maurotb:
Hi,
i have setup my wemos d1 mini pro with 5v level shifter, I use ws2815b,with relay to enable my power supply when i power on wled.
My led blink on start instead make a soft transition...

https://www.youtube.com/watch?v=jGPSlDlD0Hg

If i remove relay and make led always powered, wled work correctly with soft start of leds.
I think it's a hardware problem, the wled signal arrives before the voltage on the LEDs is sufficient i think...
How i can resolve this?
Thanks

Aircoookie:
Hi,
currently WLED turns the relay on at the same time as it turns on the LEDs (sends data). Having the relay turn on earlier would mean delaying the data signal. I might be able to add support for that in the future, but that is not ideal... What happens if you select a longer transition time in LED settings? (maybe 2 seconds)

Maurotb:
Hi, if i select 2 seconds of transition, problem persist...
I have tried to start power by shelly 200 300ms before send data.
This work.

I think the best solution is to give the option in the options to select how long to activate the power before wle and how long to turn it off after wled. 
I believe time is dependent on power supply.
I'ts possible?
I was hoping I could get rid of shelly with wled, but with this problem I can't do it
Thanks

Aircoookie:
The problem is that the transition and output to LEDs is started immediately once the power-on command is received. Changing that would not be trivial, so while it sounds simple, just turning on the relay immediately, but delaying data would need quite a few adjustments to the logic.
How long does it approx. take for the LEDs to turn on after you press the power on button?

Maurotb:
> How long does it approx. take for the LEDs to turn on after you press the power on button?
I did not understand the question well, the LEDs turn on immediately, but they make an annoying flash at power on  with random colors...
I use wled to light the house as you can see from the video, it's not nice to see that flash every time I turn it on...

I guess this problem exists in general for those who use the relay function,acceptable for effect light but not for lighting... 
Do you think it can be added as a feature in the new release?
Otherwise I build the system using shelly and making a delay via node-red but I'd like to simplify it using the wled relay output ...

huggy-d1:
What size capacitor do you have on the input to the esp and what size capacitor at the power output to the led strip? If the answer is none to either question, try adding some capacitance. You can search online for appropriate sizes if you do not know.

Maurotb:
I think my setup is okay,
i use Mean Well rsp-200-12 power supply for led.
[...]

Issue #346, opened 2019-11-12
Wait some time before turning on LEDs after turning on power/relay


pauldogg:
Would it be possible to add delay in ms for power supply to come online, as well as a pin definition for power supply relay?

Thanks

Aircoookie:
See #345 for where the pin definitions are :)
May I ask what is your use case is for a delay in relay switching? For me, turning the relay on whenever LEDs are on and off when the LEDs are off is the only logical way.

pauldogg:
Sorry I meant delay the signal, allowing time for the power to get to the LEDs.
In my current setup I have it set to 20ms; on another, I need to set it to 1000ms. If i didn't have a delay, the lights would appear off until another signal was sent (change color, brightness or effect etc...).

Aircoookie:
Have you tested that that is an actual issue with APA102?
WLED sends an update to the LEDs at least twice a second.
If the LEDs actually stay off, please confirm whether my hunch that only changing brightness will make them work, changing effects or color should not have an effect.

Fejitatete:
Is it possbile to reopen this issue plz, I have the same need : I power my strip from an ATX PSU and when the power signal is grounded, actual power may take 2s to come. Without a delay, the fade in effect is lost. Another way is to wait for an input <PSU power ok>.
Thanks.

Fejitatete:
Thank you @Aircoookie, keep up the great work :).

Maurotb:
@Aircoookie i have download 0.12.0 , this enanchements (delay output after relay on) is not present?
Thanks

Sangoku:
Ye sadly that code snippet is not working. I corrected it in the #812 (comment)  and made it work. it breaks on color change.

blazoncek:
The delay in powering up can be achieved by using playlist feature introduced in 0.12 and enhanced in 0.13.
Create 2 presets, one with on state and black color, 2nd with the desired effect and color and transition time. From those two create a playlist and use it for powering WLED.
Power off delay has been implemented in code.

up2smoke:
where exactly in the code?  for the people who compile themselves, that would be interesting to know

KlausHans:
Another problem with the playlist solution: If you use a potentiometer for brightness the solution doesn't work because the global brightness via potentiometer has nothing to do with presets. So a custom relais delay setting for switching on would be nice.

KlausHans:
[...]

PR #5315, opened 2026-01-24
fix relay not turning on at boot

DedeHai:
Fixes https://github.com/wled/WLED/issues/5312

edit:
copied from comment below:
offMode is initially set to 0 as an initialized value. if the code runs and handleIO() is not called at the right moment (i.e. when strip.needsUpdate() == false and 600ms have elapsed and offMode is still false) it misses the window, offMode is false but the relay is off, handleIO will never switch it on as it only does so if offMode is true.
Properly initializing offMode and the relay by forcing it "off" in beginStrip() eliminates that race condidtion.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable on/off transitions to prevent relay/LED glitches and ensure clean off behavior, including forced-off initialization and a short stabilization delay before driving LEDs.
  * Preserves LED-off state on boot when no relay is present so strips remain dark until enabled.

* **Chores**
  * Minor comment update in deep-sleep logic.

### DIFF
--- a/wled00/button.cpp
+++ b/wled00/button.cpp
@@ -372,13 +372,14 @@ void handleIO()
 
 void handleOnOff(bool forceOff)
 {
-  if (strip.getBrightness()) {
+  if (strip.getBrightness() && !forceOff) {
     lastOnTime = millis();
     if (offMode) {
       BusManager::on();
       if (rlyPin>=0) {
         // note: pinMode is set in first call to handleOnOff(true) in beginStrip()
         digitalWrite(rlyPin, rlyMde); // set to on state
+        delay(RELAY_DELAY); // let power stabilize before sending LED data (#346 #812 #3581 #3955)
       }
       offMode = false;
     }
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -113,6 +113,8 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
   #endif
 #endif
 
+#define RELAY_DELAY 50 // delay in ms between switching on relay and sending data to LEDs
+
 #if defined(ESP8266) || defined(CONFIG_IDF_TARGET_ESP32S2)
 #define WLED_MAX_COLOR_ORDER_MAPPINGS 5
 #else


DedeHai:
@softhack007 I did a few more changes to the proposed "one-line-of-code" fix (which would have worked, offMode was just not consistently/properly set), this now should be a lot more robust and there is only one place where the relay is initialized and handled.
edit:
to elaborate: 
offMode is initially set to 0 as an initialized value. if the code runs and handleIO() is not called at the right moment  (i.e. when `strip.needsUpdate() == false` and 600ms have elapsed and offMode is still false) it misses the window, offMode is false but the relay is off, handleIO will never switch it on as it only does so if offMode is true.
Properly initializing offMode and the relay by forcing it "off" in `beginStrip()` eliminates that race condidtion.

softhack007:
@DedeHai wow, great catch 😃

Just to be sure, I'll ask the rabbit why "the" mega-PR has exposed the bad behaviour. 

@coderabbitai testing through the commit history shows that the unexpected relay behaviour was first exposed after https://github.com/wled/WLED/pull/4658.
1. Please analyze the PR, and clarify which part of the changes are related to the observed symptom of wrong relay control.
2. Please perform an in-depth data flow analysis, to check if all possible scenarios are now handled correctly.

blazoncek:
Single line change is enough to cover all instances:
```c++
offMode = !bri;
```
Immediately after `turnOnAtBoot` check in `beginStrip()`.

The core issue is in assumption hidden in `wled.h`.

softhack007:
> ```c++
> offMode = !bri;
> ```
> Immediately after turnOnAtBoot check in beginStrip().
> The core issue is in assumption hidden in wled.h.

@blazoncek confirmed :-)

@dedehai I still like your solution, because it creates a single (centralized) function for relay control -> good for understandability and maintainability 👍

DedeHai:
> Single line change is enough to cover all instances:

yes, that is what I meant by "single line change" in my comment.

DedeHai:
> it creates a single (centralized) function for relay control -> good for understandability and maintainability

that is exactly the reason for this, I was struggling with the "distributed logic" already when I did the deep-sleep UM.

blazoncek:
There are a lot of issues with legacy code in WLED. Just try to untangle `handleTransitions()` and `colorUpdated()`.

wled00/button.cpp
```diff
@@ -367,24 +367,28 @@ void handleIO()
...
+void handleOnOff(bool forceOff)
+{
   if (strip.getBrightness()) {
     lastOnTime = millis();
     if (offMode) {
       BusManager::on();
       if (rlyPin>=0) {
-        pinMode(rlyPin, rlyOpenDrain ? OUTPUT_OPEN_DRAIN : OUTPUT);
-        digitalWrite(rlyPin, rlyMde);
-        delay(50); // wait for relay to switch and power to stabilize
```
softhack007, line 377: @DedeHai delay is not necessary any more? I have no problem with a delay() when its needed for the hardware.

wled00/button.cpp
```diff
@@ -367,24 +367,28 @@ void handleIO()
...
+        // note: pinMode is set in first call to handleOnOff(true) in beginStrip()
+        digitalWrite(rlyPin, rlyMde); // set to on state
       }
       offMode = false;
     }
-  } else if (millis() - lastOnTime > 600 && !strip.needsUpdate()) {
+  } else if (millis() - lastOnTime > 600 && !strip.needsUpdate() || forceOff) {
```
softhack007: @coderabbitAI what does operator precedence say for this line? will " || forceOff" override both "and" parts of the condition?

wled00/button.cpp
```diff
@@ -367,24 +367,28 @@ void handleIO()
...
     // for turning LED or relay off we need to wait until strip no longer needs updates (strip.trigger())
     if (!offMode) {
       BusManager::off();
       if (rlyPin>=0) {
-        pinMode(rlyPin, rlyOpenDrain ? OUTPUT_OPEN_DRAIN : OUTPUT);
         digitalWrite(rlyPin, !rlyMde);
+        pinMode(rlyPin, rlyOpenDrain ? OUTPUT_OPEN_DRAIN : OUTPUT);
```
softhack007, line 392: why move the pinMode after digitalWrite?

wled00/button.cpp
blazoncek: But not in the `loop()` context. Or are you serious?

wled00/button.cpp
blazoncek: There are missing brackets.

wled00/button.cpp
DedeHai: The delay did help with one issue, but need to check if I can find out which one. Maybe with mechanical releays and Leds receiving "half a frame" which led to glitching which this fixed. That is the only situation I could imagin it is helpful.

@blazoncek why is a delay in loop context an issue in this particular case?

wled00/button.cpp
DedeHai: to have a defined state when it is set as an output. you will see in an upcoming PR (if tests are successful) why this is important.

wled00/button.cpp
blazoncek: Any delay in loop context is a FPS issue.

wled00/button.cpp
blazoncek: Mode has to be set prior to outputting/inputting.

DedeHai: how sure are you about that?

wled00/button.cpp
DedeHai: how many FPS are acceptable when LEDs are off ;) ?

softhack007: > The delay did help with one issue, but need to check if I can find out which one. Maybe with mechanical releays and Leds receiving "half a frame" which led to glitching which this fixed. That is the only situation I could imagin it is helpful.

I can remember that SSR-type Relays ("solid state" relay to switch AC power to the PSU) were having some specific issues with power stabilization - i.e. you should wait a bit after switching on, otherwise the first few frames are just wild flashes. Maybe that was the reason to add the delay?

* https://github.com/wled/WLED/issues/346
* https://github.com/wled/WLED/issues/812
* https://github.com/wled/WLED/issues/3581

wled00/button.cpp
softhack007: 🤔 its at least ["arduino style"](https://docs.arduino.cc/language-reference/en/functions/digital-io/digitalwrite/) - not sure what espressif made out of that 😆

DedeHai: In general you can set the output state independent of "output driver mode" - Arduino was initially written for AVR MCUs where the output state register is also used to set the "pull-up" resistor of input pins and AFAIK they did try to keep this behaviour wherever possible and given the versatility of the GPIO matrix of ESP's I would be very surprised if they could not achieve it.

DedeHai: ran a quick test on an S3 and my statement (and code) hold true

blazoncek: Do whatever pleases you but logic would say (regardless of hardware capabilities):
1. set GPIO mode (input or output)
2. perform action (read or write) on GPIO

Not the other way around. I.e. read GPIO but set it to output. If that logic is incorrect then don't bother with setting GPIO mode. Read or write however and whenever you want, hardware will swallow up anything..

DedeHai: ideally yes but that is not how outputs work. in this case I need
1. aim
2. shoot

and not
1. shoot
2. aim
3. shoot (right target this time)

wled00/button.cpp
```diff
@@ -367,24 +367,28 @@ void handleIO()
...
-        digitalWrite(rlyPin, rlyMde);
-        delay(50); // wait for relay to switch and power to stabilize
+        // note: pinMode is set in first call to handleOnOff(true) in beginStrip()
+        digitalWrite(rlyPin, rlyMde); // set to on state
       }
       offMode = false;
     }
-  } else if (millis() - lastOnTime > 600 && !strip.needsUpdate()) {
+  } else if ((millis() - lastOnTime > 600 && !strip.needsUpdate()) || forceOff) {
     // for turning LED or relay off we need to wait until strip no longer needs updates (strip.trigger())
```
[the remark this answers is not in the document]
DedeHai: strip brightness is set to zero before calling the function. but you are right, that this is brittle if the init sequence ever were to be changed.

wled00/button.cpp
DedeHai: those are good references, I think a delay makes sense in general as we do not know the user's setups. If using an AC relay, 50ms may not be enough to power up the supply. I will make it a `#define` to get rid of the magic number and make it easily accessible if users want to increase that delay for their specific hardware.
If just powering up LEDs through a FET, a few microseconds would already be enough.

blazoncek:
With a bit of effort it is possible to add runtime selectable and non-blocking delay to LED output after relay is switched on.

Something that seemed impossible not so long ago is now done, without side effects.

softhack007:
> Something that seemed impossible not so long ago is now done, without side effects.

Tequila? With salt?

blazoncek:
> Tequila? With salt?

Apparently. Too much of it as it seems. It also looks like there were side effects after all.

wled00/const.h
```diff
@@ -113,6 +113,8 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
   #endif
 #endif
 
+#define RELAY_DELAY 50 // delay in ms between switching on relay and sending data to LEDs
```
netmindz, line 116: Our of curiosity, why do we need to delay? Surely it's fine to be sending data that nobody is yet listening to and if you delay too much then you risk the LEDs showing random noise until they get data

DedeHai: its not a perfect solution:
- setups that use an AC switch need time until the power is up, probably 50-100ms, at the point where a strip "powers up" it might get half a frame with bit offsets, resulting in a brief flash
- setups that use a FET to switch the LEDs do not need a delay but a "soft on" to avoid the "white flash". I have a solution for that, but needs more testing, waiting on quinled hardware to arrive. It works well on my hardware but I may have tailored it ;)
- ideally the delay would be a parameter to be set in config, I imagine mechanical relays and/or big several hundred watt supplies take longer
- a blocking wait just before LEDs turn on is acceptable IMHO, could make it async but I do not see the point in that, WLED is not a real time system for backgroud tasks

netmindz (approved):

wled00/const.h
softhack007: > * a blocking wait just before LEDs turn on is acceptable IMHO, could make it async but I do not see the point in that, WLED is not a real time system for backgroud tasks

I agree with that. It might cause a short "hickup" for usermods like AR, but that's acceptable as people will not switch their relays repeatedly like 2 times per second and for one minute again and again. It really not worth to create a distributed async timer for this.

softhack007, line 116: Improvements: change to 
````C++
#ifndef WLED_RELAY_DELAY
#define WLED_RELAY_DELAY 50
#endif
````
So users can override it in their custom build_flags

DedeHai: I skipped that intentionally as we want to move away from build flags. Better to make it a parameter in relay config.

softhack007: ok.

PR #5332, opened 2026-01-31
implement async DNS lookup - no more stuck NTP requests

### LINKED ISSUE #3285: Adjustable NTP polling interval

DedeHai:
WiFi.hostByName()  is blocking, leading to freezeing/glitches on NTP requests - what makes it worse is that there is a bug in the IDF that leads to a timeout on every call if the look-up ever fails - if that happens, LEDs freeze for a few seconds every 10 seconds. 
The new class does the look-up async without blocking and also handles that bug with workaround code. IDF V5 should have that bug fixed though.

- adding a wrapper class for DNS lookup (lwIP)
- replaces blockig WiFi.hostByName() in NTP and virtual Bus DNS look-ups

Fixes #3285

## Summary by CodeRabbit

* **New Features**
  * Asynchronous DNS resolution added for NTP and general hostname lookups on ESP32, reducing blocking during network operations.

* **Configuration**
  * Default Wi‑Fi AP channel fallback changed from 1 to 6.

* **Behavior Changes**
  * NTP sync now waits for DNS resolution with retry/backoff; hostname resolution cadence adjusted to avoid unnecessary queries and reduce blocking.

### DIFF
--- a/wled00/asyncDNS.h
+++ b/wled00/asyncDNS.h
@@ -1,7 +1,7 @@
 #pragma once
 /*
   asyncDNS.h - wrapper class for asynchronous DNS lookups using lwIP
-  by @dedehai
+  by @dedehai, C++ improvements & hardening by @willmmiles
 */
 
 #include <Arduino.h>
@@ -9,7 +9,6 @@
 #include <memory>
 #include <lwip/dns.h>
 #include <lwip/err.h>
-
 
 class AsyncDNS {
 
@@ -30,7 +29,7 @@ class AsyncDNS {
   enum class result { Idle, Busy, Success, Error };
 
   // non-blocking query function to start DNS lookup
-  static std::shared_ptr<AsyncDNS> query(const char* hostname, std::shared_ptr<AsyncDNS> current = {}) {    
+  static std::shared_ptr<AsyncDNS> query(const char* hostname, std::shared_ptr<AsyncDNS> current = {}) {
     if (!current || (current->_status == result::Busy)) {
       current.reset(new AsyncDNS());
     }
@@ -49,7 +48,6 @@ class AsyncDNS {
     } else if (err == ERR_INPROGRESS) {
       callback_state.release(); // belongs to the callback now
     } else {
-      Serial.printf("DNS fail: %d\n", err);
       current->_status = result::Error;
       current->_errorcount++;
     }


DedeHai:
I tested this on a C3 and ESP8266 for NTP, I have no hardware setup to thest the virtual bus changes.

wled00/bus_manager.cpp
```diff
@@ -713,12 +716,45 @@ size_t BusNetwork::getPins(uint8_t* pinArray) const {
 #ifdef ARDUINO_ARCH_ESP32
 void BusNetwork::resolveHostname() {
   static unsigned long nextResolve = 0;
```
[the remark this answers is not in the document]
DedeHai: why is only one host name resolved? each bus has its own instance of the asyncDNS class

wled00/bus_manager.cpp
```diff
@@ -713,12 +716,45 @@ size_t BusNetwork::getPins(uint8_t* pinArray) const {
...
+  if (Network.isConnected()) {
+    if (_DNSlookup != nullptr) {
+      if (_DNSlookup->status() == DnsResult::Success) {
+        _client = _DNSlookup->getIP();
+        delete _DNSlookup;
+        _DNSlookup = nullptr;
+      }
+    }
+    if (millis() > nextResolve && _hostname.length() > 0) {
+      nextResolve = millis() + 600000; // resolve only every 10 minutes
```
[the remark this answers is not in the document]
DedeHai: good call. you forgot to add static unsigned long nextResolve = 600000;  for init

wled00/bus_manager.cpp
```diff
@@ -713,12 +716,45 @@ size_t BusNetwork::getPins(uint8_t* pinArray) const {
...
+            } else {
+              forceReconnect = true; // reset network connection as dns is probably stuck
+              delete _DNSlookup;
+              _DNSlookup = nullptr;
+            }
+          }
+          else {
+            delete _DNSlookup; // cleanup if other error just in case (should not happen)
+            _DNSlookup = nullptr;
+          }
```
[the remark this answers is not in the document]
DedeHai: this is only called every 10 minutes, any pending request is resolved long before that.

wled00/bus_manager.cpp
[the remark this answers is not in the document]
DedeHai: got it, so this is a bug that was already present before this PR

wled00/bus_manager.cpp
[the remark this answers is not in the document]
DedeHai: it should also be called periodically

wled00/ntp.cpp
```diff
@@ -189,26 +191,54 @@ void handleNetworkTime()
...
-      sendNTPPacket();
-      ntpPacketSentTime = millis();
+      if (!ntpServerIP.fromString(ntpServerName)) // check if server is IP or domain
+      {
+        if (ntpDNSlookup == nullptr) ntpDNSlookup = new AsyncDNS;
+        DnsResult res = ntpDNSlookup->status();
+        switch (res) {
+          case DnsResult::Idle:
+            //DEBUG_PRINTF_P(PSTR("Resolving NTP server name: %s\n"), ntpServerName);
+            Serial.printf_P(PSTR("Resolving NTP server name: %s\n"), ntpServerName);
```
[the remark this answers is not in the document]
netmindz: Marked as resolved, but it's still in the code

wled00/cfg.cpp
```diff
@@ -131,7 +131,7 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
   getStringFromJson(apPass, ap["psk"] , 65); //normally not present due to security
   //int ap_pskl = ap[F("pskl")];
   CJSON(apChannel, ap[F("chan")]);
-  if (apChannel > 13 || apChannel < 1) apChannel = 1;
+  if (apChannel > 13 || apChannel < 1) apChannel = 6; // reset to default if invalid
```
netmindz, line 134: this seams unrelated to the PR

netmindz:
Great to find this issue, but not sure async dns is the right approach. surely ntp should but it's own task not part of our main loop?

wled00/asyncDNS.h
```diff
@@ -0,0 +1,69 @@
...
+#include <Arduino.h>
+#include <lwip/dns.h>
+#include <lwip/err.h>
+
+enum class DnsResult { Idle, Busy, Success, Error };
+
+class AsyncDNS {
+private:
+  ip_addr_t _raw_addr;
+  volatile DnsResult _status = DnsResult::Idle;
```
willmmiles: To be technically correct for cross-task synchronization, use `std::atomic<>` instead of `volatile`.  

```cpp
std::atomic<DnsResult> _status { DnsResult::Idle };
```

(on my quick test build it's actually 4 bytes smaller.)

wled00/asyncDNS.h
```diff
@@ -0,0 +1,69 @@
+#pragma once
+/*
+  asyncDNS.h - wrapper class for asynchronous DNS lookups using lwIP
+  by @dedehai
+*/
+
+#ifndef ASYNC_DNS_H
+#define ASYNC_DNS_H
```
willmmiles: I recommend using `#pragma once` instead of `#ifdef` guards -- it's guaranteed never to collide with another header.

wled00/asyncDNS.h
```diff
@@ -0,0 +1,69 @@
...
+
+class AsyncDNS {
+private:
+  ip_addr_t _raw_addr;
+  volatile DnsResult _status = DnsResult::Idle;
+  uint16_t _errorcount = 0;
+
+  // callback for dns_gethostbyname(), called when lookup is complete or timed out
+  static void _dns_callback(const char *name, const ip_addr_t *ipaddr, void *arg) {
+    AsyncDNS* instance = static_cast<AsyncDNS*>(arg);
```
willmmiles: This should be `reinterpret_cast`

wled00/asyncDNS.h
```diff
@@ -0,0 +1,69 @@
...
+      _status = DnsResult::Success; // result already in cache
+    } else if (err != ERR_INPROGRESS) {
+      _status = DnsResult::Error;
+      _errorcount++;
+    }
+    return _status;
+  }
+
+  // get the IP once Success is returned
+  IPAddress getIP() {
```
willmmiles: Accessors could be made `const`

wled00/asyncDNS.h
```diff
@@ -0,0 +1,69 @@
...
+*/
+
+#ifndef ASYNC_DNS_H
+#define ASYNC_DNS_H
+
+#include <Arduino.h>
+#include <lwip/dns.h>
+#include <lwip/err.h>
+
+enum class DnsResult { Idle, Busy, Success, Error };
```
willmmiles: This should be scoped inside the class.  (Guarantees there's never a name collision.)

wled00/ntp.cpp
```diff
@@ -180,6 +181,7 @@ void handleTime() {
   }
 }
 
+AsyncDNS* ntpDNSlookup;
```
willmmiles: Use `std::unique_ptr` instead of naked pointers.  Also it should be static inside the function.

wled00/asyncDNS.h
```diff
@@ -0,0 +1,69 @@
...
+
+class AsyncDNS {
+private:
+  ip_addr_t _raw_addr;
+  volatile DnsResult _status = DnsResult::Idle;
+  uint16_t _errorcount = 0;
+
+  // callback for dns_gethostbyname(), called when lookup is complete or timed out
+  static void _dns_callback(const char *name, const ip_addr_t *ipaddr, void *arg) {
```
willmmiles: If this class is destructed while a query is still outstanding, this will result in the `arg` pointer becoming invalid, resulting in memory corruption when the callback is eventually executed.   Disconnecting from the wifi *does not dequeue DNS callbacks* -- they're supposedly guaranteed to be called "eventually" by the lwip stack.

Fixing this is nontrivial -- the state data needs to be stored on the heap separately from the interface class, and scoped with tools like `std::shared_ptr` so it won't be released until the callback is run.  I made a little sketch of this, I can add it to the PR if you'd like.

willmmiles (changes requested):
This is a great idea and definitely the right general approach.  The possible use-after-free when the object is destructed needs to be fixed though.

wled00/cfg.cpp
DedeHai: it is. I should have changed this line in https://github.com/wled/WLED/pull/5115 but was unaware at the time.

DedeHai:
> Great to find this issue, but not sure async dns is the right approach. surely ntp should but it's own task not part of our main loop?

That is a good point actually, we should think about putting the rendering and background work in different tasks but I think the current code structure is not ready for that. Sure could put NTP in a task but that eats up RAM and on single core ESPs the arduino function this PR replaces may still be blocking, I did not look how the espressif "wait for bit set" part is implemented, it may not play nicely with the RTOS as it sounds like a pretty bare-metal thing to wait for a bit to be set.
What I have implemented here is the correct way to use the DNS look-up provided by espressif.

wled00/asyncDNS.h
DedeHai: Good point. I am not opposed to "do it properly" but I also see two more trivial solutions:
- make the class global and instantiate only one, add a "key" or a char array and share the class between callers. this adds the complexity of a shared resource though and probably has implications I am currently not thinking of.
- simplest: do not delete if status is busy (works well for NTP use)
both of these will not solve the issue for virtual buses though - maybe adding another function to the class that is actually blocking would simplify its use for places that "need an IP immediately" - what I have implemented in busmanager is way too complex IMHO but mostly due to the fact that it needs to dance around that bug gracefully.

wled00/ntp.cpp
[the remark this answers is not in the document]
DedeHai: fixed locally but forgot to commit

wled00/ntp.cpp
DedeHai: std::unique_ptr  gives me errors as make_unique is not defined.

wled00/ntp.cpp
```diff
@@ -9,7 +10,7 @@
  * Acquires time from NTP server
  */
 //#define WLED_DEBUG_NTP
-#define NTP_SYNC_INTERVAL 42000UL //Get fresh NTP time about twice per day
+#define NTP_SYNC_INTERVAL 15//42000UL //Get fresh NTP time about twice per day
```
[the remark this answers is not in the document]
DedeHai: that's a debug leftover, will fix

wled00/ntp.cpp
willmmiles: There's a shim in `bus_manager.h`.   (This might be moot re fixing the safety issues though -- shared_ptr might allow a cleaner solution.)

wled00/asyncDNS.h
willmmiles: It's not just deletion: `renew()` and `reset()` are also invalid if `status == Busy` as the results of a new query will be overwritten by the old callback at some point.  If a query is stuck, the state object has to be abandoned -- there's no real alternative.

netmindz:
> > Great to find this issue, but not sure async dns is the right approach. surely ntp should but it's own task not part of our main loop?
> 
> That is a good point actually, we should think about putting the rendering and background work in different tasks but I think the current code structure is not ready for that. Sure could put NTP in a task but that eats up RAM and on single core ESPs the arduino function this PR replaces may still be blocking, I did not look how the espressif "wait for bit set" part is implemented, it may not play nicely with the RTOS as it sounds like a pretty bare-metal thing to wait for a bit to be set. What I have implemented here is the correct way to use the DNS look-up provided by espressif.

NTP sync should never have been anything other than a background task. Rather than adding a load of complex code to handle async DNS lookups, we should be correcting the failure in design

willmmiles:
> NTP sync should never have been anything other than a background task. Rather than adding a load of complex code to handle async DNS lookups, we should be correcting the failure in design

Hard disagree.  Every FreeRTOS task requires >4k of RAM for stack and ancillary data structures.  That would be an enormous waste just for NTP synchronization.

wled00/asyncDNS.h
DedeHai: I wrote it the way I did because I never got it to not run the callback - you are right it is not as robust as it could be but it was not my intention to write it "fool proof" i.e. reset while a request is going on.
So which one do you prefer: 
- using seperate heap for the callback as you propose
- not use a pointer and make it fully static (like I did now in the busmanager) and not go down the road of managing dangling pointers?
to me the option of not using it dynamically makes sense, at least ATM as DNS is only used for NTP and virtual bus. The class only uses a few bytes of RAM.

willmmiles: I'd like to be able to test/treat an AsyncDNS object as a unit, with safety by design -- "it doesn't crash because we don't happen to delete/reuse it while it's active" isn't good enough IMO, static or otherwise.   Eventually that *will* bite us because the class violates the C++ object requirement that a destructor is always safe.

I took a stab at safety, while keeping the current object and re-use semantics:
https://github.com/willmmiles/WLED/commit/26c202aad8445beffc61a7c77d2af000a87426a2

I think it might be done better with `std::future<>` style semantics though; the code would be closer to the current class design.  Eg. the fundamental call is:

```cpp
class AsyncDNS {
   static std::shared_ptr<AsyncDNS> query(const char* hostname, std::shared_ptr<AsyncDNS> previous = {});

   private:
       AsyncDNS();   // Can only be produced by making a query
}
```

willmmiles: I will say that I'm having a *lot* of trouble with DNS hanging on my test system.  Do you have any references on that IDF v4 bug you mention?

willmmiles: Future-like formulation: https://github.com/willmmiles/WLED/commit/efcfc5ed665ce1feb5ecb82c28244967069a69e1

DedeHai: > I will say that I'm having a _lot_ of trouble with DNS hanging on my test system. Do you have any references on that IDF v4 bug you mention?

I do too, that is why I am pretty sure its a bug, it just locks up and does not recover until reconnect. I did not really find concret info about it but the changelog mentions some fix for DNS deadlock: https://github.com/espressif/esp-idf/releases?page=1

DedeHai: > Future-like formulation: [willmmiles@efcfc5e](https://github.com/willmmiles/WLED/commit/efcfc5ed665ce1feb5ecb82c28244967069a69e1)

that is a whole lot of fancy c++ magic I do not (yet) understand. Is it to make it multi-thread safe?

netmindz:
> > NTP sync should never have been anything other than a background task. Rather than adding a load of complex code to handle async DNS lookups, we should be correcting the failure in design
> 
> Hard disagree. Every FreeRTOS task requires >4k of RAM for stack and ancillary data structures. That would be an enormous waste just for NTP synchronization.

I retract my statement, I was unaware that there was such a high penalty for running this way

wled00/asyncDNS.h
willmmiles: > > Future-like formulation: [willmmiles@efcfc5e](https://github.com/willmmiles/WLED/commit/efcfc5ed665ce1feb5ecb82c28244967069a69e1)
> 
> that is a whole lot of fancy c++ magic I do not (yet) understand. Is it to make it multi-thread safe?

Sort of.  The goal is to make sure that the addresses that will be touched by the callback in the future cannot under any circumstances be released until the callback has run.  `std::shared_ptr<>` is a reference-counting pointer: it keeps track of how many `shared_ptr`s address the same storage, so the pointed-to object won't be `delete`d until the last `shared_ptr` pointing at it is removed.  `shared_ptr` guarantees that the reference count is atomic and so it is thread-safe in this context.

The first implementation hides the `shared_ptr` by making the `AsyncDNS` class a wrapper around it: the only member is a `shared_ptr` to the DNS results.  A copy of the `shared_ptr` is "gifted" to the callback, who is then guaranteed to be able to write the state.  If the `AsyncDNS` object is destroyed, or the request is abandoned via `renew()` or `reset()`, that copy keeps the DNS result struct alive until it's finished with it.

The second implementations strips off the wrapper layer, so clients always have to store a `shared_ptr<AsyncDNS>`.  This is intended to make the behaviour to the client more clear, ie. the thing you hold is a handle to the real object (and removes an extra layer of code).  The major change to make that work is the interface to `query()`: it lets a client pass in a `shared_ptr<AsyncDNS>` that's been created before, so it can re-use the context (ie. error count) or storage *if that request has been completed*.  (If the request is still busy, or if you pass in an empty pointer, `query()` makes a new storage object for you.)

I liken the second approach to the use of `std::promise<>`/`std::future<>`.  Promises and futures are part of the C++11 async library.  You can think of them as a thread-safe single result queue: the end holding the `std::promise` can write a single result value in a safe way, such that the `std::future` can `wait()` for it and read it later.  Unfortunately actual `std::promise` and `std::future` are quite heavyweight, particularly with their use of exceptions.

willmmiles:
> > > NTP sync should never have been anything other than a background task. Rather than adding a load of complex code to handle async DNS lookups, we should be correcting the failure in design
> > 
> > 
> > Hard disagree. Every FreeRTOS task requires >4k of RAM for stack and ancillary data structures. That would be an enormous waste just for NTP synchronization.
> 
> I retract my statement, I was unaware that there was such a high penalty for running this way

Yeah, the minimum stack requirement for proper Tasks is really kind of frustrating honestly.  The end result seems to be that every reasonably sized project ends up writing their own async executive, from as simple as just `loop()` callback hooks to more complex systems (#4808's deferred executor) all the way to full on async environments (the latest hotness is [Embassy in Rust](https://embassy.dev/book/index.html)).

wled00/ntp.cpp
```diff
@@ -182,15 +183,50 @@ void handleTime() {
...
+              if (offMode) forceReconnect = true; // do not disturb while LEDs are running
+              delete ntpDNSlookup;
+              ntpDNSlookup = nullptr;
+            }
+            ntpLastSyncTime = millis() - (1000*NTP_SYNC_INTERVAL - 300000); // pause for 5 minutes
+            break;
+        }
+      }
+      else
+        sendNTPPacket();
```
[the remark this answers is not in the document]
DedeHai: I would consider this very much an edge case with no bad consequences to system other than potential heap fragmentation. Chances a user changes to IP during a lookup are slim to none.

wled00/asyncDNS.h
DedeHai: I'll leave this to your judgement. I would keep it as light-weight as possible. Feel free to add to this PR.

DedeHai:
@willmmiles is your implementation here

>(https://github.com/willmmiles/WLED/commit/efcfc5ed665ce1feb5ecb82c28244967069a69e1)

complete? or what is the way forward to get this PR ready?
I think this bug really needs to finally be fixed in 16.0, if we can't get it to work "safely" I would suggest to use a global/persistent implementation rather than not having it at all. And who knows, the core bugs may be fixed in V5.

willmmiles:
IIRC both forms (https://github.com/willmmiles/WLED/commit/efcfc5ed665ce1feb5ecb82c28244967069a69e1) and (https://github.com/willmmiles/WLED/commit/26c202aad8445beffc61a7c77d2af000a87426a2) were functional and safe, but there was an outstanding issue of DNS hanging up after some number of queries.  Looking at it now.

DedeHai:
@willmmiles the DNS hanging up is already happening in the current implementation and with the "wait for it" arduino implementation it causes a full stall every 10 seconds or so. I think it is an IDF bug that can not currently be fixed - at least the new implementation will not lead to any stalls, so even if it eventually may fail to update from NTP its an improvement - and who knows, it might recover afer some time.

willmmiles:
> @willmmiles the DNS hanging up is already happening in the current implementation and with the "wait for it" arduino implementation it causes a full stall every 10 seconds or so. I think it is an IDF bug that can not currently be fixed - at least the new implementation will not lead to any stalls, so even if it eventually may fail to update from NTP its an improvement - and who knows, it might recover afer some time.

@DedeHai I'm in the thick of it now - it doesn't look like an IDF bug, it looks like the DNS client's internal state is getting overwritten by something else.  With any luck the debugger will tell me what...

That debugging aside, if it's OK with you, I'll push the "safest" AsyncDNS version based on https://github.com/willmmiles/WLED/commit/efcfc5ed665ce1feb5ecb82c28244967069a69e1 to this PR?

Also: the current logic here is still re-resolving the DNS with every NTP transaction.  Best practice with NTP is to stick to the same server IP until it drops a couple packets, and only re-resolve then.  I can include that fix too if you'd like.

DedeHai:
@willmmiles sure, go ahead and add to this PR. 
You are right that re-resolving is not strictly needed, but unless there is a good reason not to do that  I would not bother adding code to a function that runs twice a day.

willmmiles:
The DNS freezeups - at least on my network - are being caused by IPv6 router advertisement packets.  Apparently they overwrite the IPv4 DNS servers, and there seems to be no way to turn that off at runtime as far as I can tell.  The result seems to be that DNS becomes completely unavailable.  (I'm not sure why it doesn't send queries via IPv6 instead .. but it doesn't.)

Anyhow, if we want to have working DNS support for NTP, MQTT, etc., we'll have to do something to back up and restore the DNS configuration if it gets overwritten by IPv6 servers.

I'm honestly not sure if this is any better in IDF v5 -- as far as I can tell, the RA packet handling still overwrites any other configured DNS servers on that interface.   Something to look in to at any rate.

DedeHai:
this sounds like something espressif should know about. Is this a regression with this PR or is was this always an/the issue?

willmmiles:
It's unrelated to this PR, and they do know about it.

https://github.com/espressif/arduino-esp32/discussions/9988
https://github.com/espressif/arduino-esp32/issues/3722
https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/lwip.html  (see the "Note" block under "Adapted APIs" -- though it very much undersells the problem.)

It looks like what they did in (later) IDF v5 is extend LwIP to tracking DNS servers per interface -- likely this works around the problem if they consider IPv6 as a unique "interface".

The workaround is to manage the server list ourselves in network event callbacks - it's easy enough to do, I'll make a PR for it.

DedeHai:
tested, works like a charm.
the DNS lookup in busmanager may need further improvements but current implementation expects immediate resolution and can not be done in a non-blocking way easily.

DedeHai:
I have no objections on merging this as-is.

willmmiles:
> the DNS lookup in busmanager may need further improvements but current implementation expects immediate resolution and can not be done in a non-blocking way easily.

I concur.  Also I'd like to think about moving the error counter out to the points of usage - not every client needs it and it's a bit cumbersome in the interface.  Both can be improved later.

PR #5337, opened 2026-02-02
Fix LEDs glitching at startup (ESP32 only)

DedeHai:
This PR fixes the brief halt of LED animations by storing wifi credentials in non volatile memory which makes WiFi connect almost instantly after boot.
If multiple WiFi are configured, only the first one is stored and will benefit from this fix.

ESP8266 does not support this "fast connenct" feature. 

To still prevent flash wear, the stored credentials are compared to config credentials and only updated if they changed. This comparison and updating is done only once at bootup.

## Summary by CodeRabbit

* **Improvements**
  * Fewer flash writes on ESP32 devices to extend device lifespan.
  * Improved detection and synchronization when stored Wi‑Fi credentials differ from current settings.
  * More reliable, faster Wi‑Fi connection behavior after initial setup.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -63,6 +63,8 @@ body:
         - ESP32-S2
         - ESP32-C3
         - Other
+        - ESP32-C6 (experimental)
+        - ESP32-C5 (experimental)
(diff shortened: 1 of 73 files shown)


DedeHai:
Tested working on ESP32, C3, S3

Issue #5334, opened 2026-02-01
Classic ESP32: GPIO17 is not always needed for PSRAM


wled-install:
### What happened?

The following line in pin_manager.cpp:

```
if (gpio == 16 || gpio == 17) return !psramFound(); // PSRAM pins on ESP32-D0WDR2-V3 (these are IO)
```

prevents using GPIO17 together with PSRAM. Modules that have "in-package" PSRAM integrated (like ESP32-WROOM-32E-N8R2, ESP32-WROOM-32E-N4R2 based on ESP32-D0WDR2-V3 / ESP32-D0WDRH2-V3) do not use GPIO17 for PSRAM. In this case GPIO17 can be used for other purposes (for example for Ethernet clock).

GPIO17 is only used for PSRAM on modules with external "off-package" PSRAM (like WROOVER uses).

Please refer to ESP32 Datasheet, pages 20-21. [https://documentation.espressif.com/esp32_datasheet_en.pdf](https://documentation.espressif.com/esp32_datasheet_en.pdf)

### To Reproduce Bug

Compile and install as it is on ESP32-D0WDR2-V3 / ESP32-D0WDRH2-V3 based Controller.

### Expected Behavior

On ESP32-D0WDR2-V3 / ESP32-D0WDRH2-V3 the GPIO17 should not be reserved for PSRAM.

### Install Method

Self-Compiled

### What version of WLED?

0.15.x, 0.16.x

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

softhack007:
Hi @wled-install 👋 

If we just allow the PIN on boards that do need gpio17 for PSRAM, we'll be facing lots of user complaints like "lights freeze after changing the pin, homeassistent cannot connect" 😉 . Similar if we unlock the pin via build flag, because users will often install the wrong firmware as they don't understand the difference - "ethernet broken after updating WLED".

It means we need a reliable way to identify - at runtime - the new package that you mention, while still keeping 16 & 17 protected on the normal "WROVER-B" boards. 

If you have both kinds of boards at hand, can you do a quick check of what ``ESP.getChipModel()`` reports on both boards (arduino-esp32 v2.0.9) ? If the function returns a different string for internal "in-package" PSRAM modules, then it might be possible to change the line you mention into something like

````C++
[...]

PR #5338, opened 2026-02-02
Classic ESP32: GPIO17 is not needed to be reserved for chips with in-package PSRAM

wled-install:
Fix for https://github.com/wled/WLED/issues/5334

Edit (softhack007): The change relies on the framework knowing about the "ESP32-D0WDR2-V3" chip model - available since arduino-esp32 v2.0.7. It means that only "V4" builds will allow using GPIO17. WLED 0.15.3 is still using the older framework, only WLED 0.16 will be able to detect the "ESP32-D0WDR2-V3" chip properly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPIO compatibility checks for ESP32 devices with external memory, enhancing detection of available pins across different module variants and memory configurations. This update better handles device-model-specific cases to reduce false negatives and ensure more reliable pin availability reporting on affected hardware.

### DIFF
--- a/wled00/pin_manager.cpp
+++ b/wled00/pin_manager.cpp
@@ -246,11 +246,14 @@ bool PinManager::isPinOk(byte gpio, bool output)
       // for classic ESP32 (non-mini) modules, these are the SPI flash pins
       if (gpio > 5 && gpio < 12) return false;      //SPI flash pins
     }
-    if ((strncmp_P(PSTR("ESP32-D0WDR2-V3"), ESP.getChipModel(), 15) == 0) || (strncmp_P(PSTR("ESP32-D0WDRH2-V3"), ESP.getChipModel(), 16) == 0)) {  // allow gpio17 on modules with in-package PSRAM
-        if (gpio == 16) return !psramFound();
-        if (gpio == 17) return true;
-    } else if (gpio == 16 || gpio == 17) return !psramFound(); // PSRAM pins on modules with off-package PSRAM
-    
+    if (gpio == 16) return !psramFound(); // PSRAM pins on modules with off-package or in-package PSRAM
+    if (gpio == 17) {
+      if (strncmp_P(PSTR("ESP32-D0WDR2-V3"), ESP.getChipModel(), 15) == 0) {
+        return true;
+      } else {
+        return !psramFound(); // PSRAM pins on modules with in-package PSRAM
+      }
+    }    
   #endif
     if (output) return digitalPinCanOutput(gpio);
     else        return true;


wled00/pin_manager.cpp
```diff
@@ -246,7 +246,11 @@ bool PinManager::isPinOk(byte gpio, bool output)
       // for classic ESP32 (non-mini) modules, these are the SPI flash pins
       if (gpio > 5 && gpio < 12) return false;      //SPI flash pins
     }
-    if (gpio == 16 || gpio == 17) return !psramFound(); // PSRAM pins on ESP32-D0WDR2-V3 (these are IO)
+    if ((strncmp_P(PSTR("ESP32-D0WDR2-V3"), ESP.getChipModel(), 15) == 0) || (strncmp_P(PSTR("ESP32-D0WDRH2-V3"), ESP.getChipModel(), 16) == 0)) {  // allow gpio17 on modules with in-package PSRAM
+        if (gpio == 16) return !psramFound();
+        if (gpio == 17) return true;
+    } else if (gpio == 16 || gpio == 17) return !psramFound(); // PSRAM pins on modules with off-package PSRAM
```
DedeHai: gpio 16 should not be in this conditional if its just for 17

softhack007:
I think we need more testing with classic esp32 that have psram, to confirm that pin17 is only allowed when the board supports it.

The test can be quite simple
1) change led pin to 17 -> if it crashes, the change is not good enough.
2) change led pin to 17, create a large 2D setup where the buffer usually is going into PSRAM -> if it crashes, ...

I have a wrover-b and a lilygo-t7 with psram in my collection, but I'll need a few days before I can test. If anybody else has another classic esp32 with PSRAM, please feel invited to test.

<br/>

Edit:
First tests (but we need more tests with other boards!)
* Lilygo T7 v1.5 identifies as "ESP32-D0WD-V3" (rev 3.0), needs GPIO 16&17 for PSRAM - crashes (bootloop) when assigning GPIO17 to LEDs
* Freenove wrover-E-cam identifies as either "ESP32-D0WDQ5" (older framework) or "ESP32-D0WD-V3" (newer frameworks), needs GPIO 16&17 for PSRAM
* noname WROVER-B: identifies as "ESP32-D0WD" (rev.1 chip, without "V3")

wled00/pin_manager.cpp
```diff
@@ -246,7 +246,11 @@ bool PinManager::isPinOk(byte gpio, bool output)
       // for classic ESP32 (non-mini) modules, these are the SPI flash pins
       if (gpio > 5 && gpio < 12) return false;      //SPI flash pins
     }
-    if (gpio == 16 || gpio == 17) return !psramFound(); // PSRAM pins on ESP32-D0WDR2-V3 (these are IO)
+    if ((strncmp_P(PSTR("ESP32-D0WDR2-V3"), ESP.getChipModel(), 15) == 0) || (strncmp_P(PSTR("ESP32-D0WDRH2-V3"), ESP.getChipModel(), 16) == 0)) {  // allow gpio17 on modules with in-package PSRAM
```
softhack007: The second string "ESP32-D0WDRH2-V3" is never returned by ESP.getChipModel(), so it can be removed.

I've checked both arduino-esp32  V3.x and V2.0.9, the arduino core does not know "RH2-V3"

https://github.com/espressif/arduino-esp32/blob/6cfa93f4b4e2c0e3cf492e4a45780bfbe757fcd4/cores/esp32/Esp.cpp#L306

wled-install: Ok I removed it.

wled00/pin_manager.cpp
wled-install: Yes thank you, I changed it to be more clearly

wled-install:
@softhack007
Thank you for results!!! So, the proposed solution should work as your boards with off-package PSRAM should use GPIO17, so you cannot use it for LEDs. As they are not reported as ESP32-D0WDR2-V3, everything fine. 

Actually in my tests only WLED 0.16.x works reliable with PSRAM, 0.15.x (if not esp-idf4 used) crashes whatever I do.

softhack007:
> Thank you for results!!! So, the proposed solution should work as your boards with off-package PSRAM should use GPIO17, so you cannot use it for LEDs. As they are not reported as ESP32-D0WDR2-V3, everything fine.

@wled-install yes for all the 3 boards in my collection the change is safe. 
I hope that another maintainer also has a classic esp32 with PSRAM. If we get a few more "ok" tests, then the change can be merged.

> Actually, in my tests only WLED 0.16.x works reliable with PSRAM, 0.15.x (if not esp-idf4 used) crashes whatever I do.

same here - I never got any PSRAM boards working with the old "AC" framework. It seems that PSRAM boards always need a V4 build.

DedeHai:
I have one classic WROVER, not sure which model. will check.
btw. I at one point did remove `psramFound()` as all it really does is to check if BOARD_HAS_PSRAM is defined. I either did not commit that or it got stuck in PR limbo.

softhack007:
> `psramFound()` as all it really does is to check if BOARD_HAS_PSRAM is defined.

@DedeHai side-discussion:
I just tried to follow the logic in esp32-hal-psram.c (arduino-esp32 v2.0.9) - the code i found is a bit more complicated.

My understanding is:
* if you don't have BOARD_HAS_PSRAM, psramFound() is always false
*  psramFound() = false can also happen if your firmware was build with BOARD_HAS_PSRAM, but the chip does not have PSRAM.
* if CONFIG_SPIRAM_BOOT_INIT is set (?bootloader aborts if no PSRAM?), all checks in ``psramInit()`` are skipped.
* (!) otherwise the returned value depends on some EFUSEs, plus a short functional test with ``testSPIRAM()``.

<br/>

-> psramFound() == false might also indicate that support is available (BOARD_HAS_PSRAM), but PSRAM is not working

-> For PIN allocation, relying on psramFound() might be dangerous in one case: a broken PSRAM chip does not mean you can use PSRAM pins freely
-> ps_malloc() internally checks psramFound(), and returns NULL when no PSRAM. I think that heap_caps_malloc_prefer(...) can also handle "no PSRAM" gracefully.

<br/>

````C++
bool ARDUINO_ISR_ATTR psramFound(){
    return spiramDetected;
}

void ARDUINO_ISR_ATTR *ps_malloc(size_t size){
    if(!spiramDetected){
        return NULL;
    }
    return heap_caps_malloc(size, MALLOC_CAP_SPIRAM | MALLOC_CAP_8BIT);
}

````

````C++
void initArduino()
{
#if CONFIG_SPIRAM_SUPPORT || CONFIG_SPIRAM
    psramInit();
#endif
````

````C++
static volatile bool spiramDetected = false;

bool psramInit(){
    if (spiramDetected) {
        return true;
    }
#ifndef CONFIG_SPIRAM_BOOT_INIT
    if (spiramFailed) {
        return false;
    }
#if CONFIG_IDF_TARGET_ESP32
    uint32_t chip_ver = REG_GET_FIELD(EFUSE_BLK0_RDATA3_REG, EFUSE_RD_CHIP_VER_PKG);
    uint32_t pkg_ver = chip_ver & 0x7;
    if (pkg_ver == EFUSE_RD_CHIP_VER_PKG_ESP32D2WDQ5 || pkg_ver == EFUSE_RD_CHIP_VER_PKG_ESP32PICOD2) {
        spiramFailed = true;               // <-- HERE  check failed
        log_w("PSRAM not supported!");
        return false;
    }
#elif CONFIG_IDF_TARGET_ESP32S2
    extern void esp_config_data_cache_mode(void);
    esp_config_data_cache_mode();
    Cache_Enable_DCache(0);
#endif
    if (esp_spiram_init() != ESP_OK) {
        spiramFailed = true;               // <-- HERE check failed
        log_w("PSRAM init failed!");
#if CONFIG_IDF_TARGET_ESP32
        if (pkg_ver != EFUSE_RD_CHIP_VER_PKG_ESP32PICOD4) {
            pinMatrixOutDetach(16, false, false);
            pinMatrixOutDetach(17, false, false);
        }
#endif
        return false;
    }
    esp_spiram_init_cache();
    //testSPIRAM() allows user to bypass SPI RAM test routine
    if (!testSPIRAM()) {
        spiramFailed = true;               // <-- HERE  check failed
        log_e("PSRAM test failed!");
        return false;
    }
    if (esp_spiram_add_to_heapalloc() != ESP_OK) {
        spiramFailed = true;              // <-- HERE check failed
        log_e("PSRAM could not be added to the heap!");
        return false;
    }
#if CONFIG_SPIRAM_USE_MALLOC && !CONFIG_ARDUINO_ISR_IRAM
    heap_caps_malloc_extmem_enable(CONFIG_SPIRAM_MALLOC_ALWAYSINTERNAL);
#endif
#endif /* CONFIG_SPIRAM_BOOT_INIT */
    log_i("PSRAM enabled");
    spiramDetected = true;                // <-- HERE all checks passed
    return true;
}

````

DedeHai:
my info may be outdated, at the time (few months ago), I think it was discussed several times but no conclusion reached: if PSRAM is enabled on ESP32, it also requires that compile fix to support the chips with that silicon bug. I am not sure on the best way to solve it in general (not this PR). Ideally we would have one single build for all ESP32 classic variants, no matter if it has PSRAM or not, I am not aware of all pitfalls for that though.

softhack007:
@DedeHai @willmmiles or @netmindz - if you have a classic esp32 with PSRAM, please do a quick check.

softhack007:
Ok - as we did not find any "counter-examples" - let's merge.

intermittech:
Just as a note on this one, is there specific detection for the ESP32-PICO-MINI-02? That one has 8MB flash and 2MB PSRAM.
I believe the PSRAM is connected to GPIO 9 and 10
GPIO 16, 17, 18 and 23 are also not available and are not broken out physically.

In contrast to earlier models GPIO 7 and 8 are available.

softhack007:
> Just as a note on this one, is there specific detection for the ESP32-PICO-MINI-02? That one has 8MB flash and 2MB PSRAM. I believe the PSRAM is connected to GPIO 9 and 10 GPIO 16, 17, 18 and 23 are also not available and are not broken out physically.
> 
> In contrast to earlier models GPIO 7 and 8 are available.

@wled-install @intermittech I don't think we can detect more, and we already have the PICO-D2 in our list.

https://github.com/wled/WLED/blob/ce31d802d530c288989a21698e917c0343e99822/wled00/pin_manager.cpp#L244


All the models that ESP.getChipInfo() knows are here

https://github.com/espressif/arduino-esp32/blob/cb4b0ac769135572c8dd12a1f9329cb5065a4a33/cores/esp32/Esp.cpp#L287-L307

And that's the newest arduino-esp32 3.x, while we are still on 2.0.9.

wled-install:
@softhack007 Danke!!!

PR #5341, opened 2026-02-02
spots effect accuracy improvements

softhack007:
A few improvement for the spots effect, especially relevant for longer strips:
* avoid gaps on left/right side of strip, by using higher accuracy for "zoneLength"
* prevent overflows
* ensure that zones calculations are performed in full 32bit
* fix wrong slider name in "spots fade"

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Spots effect geometry and per-pixel rendering to prevent zero-sized zones, eliminate visual gaps, and ensure stable brightness handling to avoid artifacts.

* **Documentation**
  * Updated Spots Fade parameter label from "Spread" to "Speed" for clearer control and better user understanding.


wled00/FX.cpp
```diff
@@ -2991,7 +2996,7 @@ uint16_t mode_spots_fade()
   unsigned tr = (t >> 1) + (t >> 2);
   return spots_base(tr);
 }
-static const char _data_FX_MODE_SPOTS_FADE[] PROGMEM = "Spots Fade@Spread,Width,,,,,Overlay;!,!;!";
+static const char _data_FX_MODE_SPOTS_FADE[] PROGMEM = "Spots Fade@Speed,Width,,,,,Overlay;!,!;!";
```
DedeHai, line 2959: while you are at it: rename "Width" to "Density" for both variants?

DedeHai (changes requested):
generally an improvement, there are two issues (one introduced, one unsolved):

- on small segments, the threshold can be too high, turning off all spots. something like this fixes that but may introduce "breaking changes" I did only a brief test:
``` c++
  uint16_t maxWave = triwave16(((zoneLen - 1) * 0xFFFF) / zoneLen);
  threshold = min(threshold, uint16_t(maxWave - 1)); // guarantee threshold is always reached (prevent "no spots" on short strips)
```
to check: use a segment with 15 LEDs, turn speed slider down and intensity slider up.

- the newly introduced issue is a classic integer math rounding problem: while the spots now spread better, they distribute unevenly. I faced the same dilemma in "PS chase" and used more elaborate code to get both a full spread distribution as well as an even spacing.

softhack007:
> the newly introduced issue is a classic integer math rounding problem: while the spots now spread better, they distribute unevenly. I faced the same dilemma in "PS chase" and used more elaborate code to get both a full spread distribution as well as an even spacing.

@DedeHai thanks for your comments 😃 Yes unevenly distributed gaps are a new side-effect. At second look the nested loops with offset re-calcution seem unnecessary, we could also just repeat the tri-wave pattern until the end of strip is reached. Just that the effect might not be centered any more for long "spread" values. I'll think about a better implementation.

Edit: maybe this is better
> start with offset = 0.5 * (SEGLEN / pattern_length) // similar to old code, no fixed-point math involved. Could also use float for better accuracy as this is only calculated once per frame
> while (SEGLEN - last_position <  pattern_length) draw_triwave_pattern
 
Edit2: 🤔 actually offset calculation is the only "tricky" thing left, because the effect should be centered no matter how long or short the strip really is...

DedeHai:
I need to check that code but in general for long er strips with 100 LEDs+ you will not notice a 1 pixel offset, unless its coarse and you want it aligned to furniture but I guess that is more of an edge case. What is much more noticeable is an irregular pattern. 

Edit: with 24V strips and "zones" now coming up more and more, low LED-count segments are not uncommon (I have 64 on mine)

softhack007:
@DedeHai I'll make a second attempt on this, you're right that unequally spaced spots look ugly..

softhack007:
@coderabbitai re-review

softhack007:
@coderabbitAI let's take the chance to clean up your learnings:
* compare your learnings against the new coding guidelines in docs/cpp.instructions.md. Remove learnings that are now fully covered by the guidelines that are made applicable to you via .coderabbit.yaml
* remove expired/outdated learnings

softhack007:
@coderabbitai please double-check if your learning f885e51b is correct. It sounds like the pattern is not safe against unsigned wrap-around. Personally i would prefer the "millis() - last_time > wait" pattern that is also in your learnings.

softhack007:
@coderabbitai please also check your learnings against the new .github/copilot-instructions.md (applicable via .coderabbit.yaml) and check for learnings that are fully covered and could be removed now.

PR #5343, opened 2026-02-03
apply inverse gamma to segment brightness for better color preservation

### LINKED ISSUE #5342: Segment brightness resolution
### What happened?

Hello, I have WT32-ETH01 module running on ethernet with latest firmware 0.15.3. I have 8 LED strips connected to 8 hw pins to control them separately. So, I don't want to use global brightness control and I set it to max (255). I would like to use segment brightness controllers to set the brightness individually. Unfortunately, animation of effects is degraded with very poor brightness resolution when segment brightness is set to low. Animations work fine when segments brightness is set to 100% and global brightness to e.g. 50 I found similar issue reported [here](https://www.reddit.com/r/WLED/comments/1pkd0u9/issue_with_segment_brightness_and_effects/).

### To Reproduce Bug

- set global brightness to 20%, set segment brightness to 100% and run a animation (correct behaviour). 
- set global brightness to 100%, set segment brightness to 20% and run a animation (brightness resolution issue).
The result is not same.

### Expected Behavior

- set global brightness to 20%, set segment brightness to 100% and run a animation. 
- set global brightness to 100%, set segment brightness to 20% and run a animation.
The result is same.

### Install Method

Binary from WLED.me

### What version of WLED?

0.15.3

ESP32

### Relevant log/trace output

```shell

```

### Anything else?

_No response_

### Code of Conduct

DedeHai:
to make global brightness slider and segment brightness slider behave the same, the segment brightness needs to be gamma-inverted when using gamma for color correction. 

the reason being (with g = gamma)
global brightness:  `(color)^g * brightness`
segment brightness: `(color * brightness)^g`

using `brightness^(1/g) `results in `(color *  brightness^(1/g) )^g = color^g * brightness` i.e. the same as with global brightness. 

The way WLED handles gamma is not mathematically correct but is an approximation and a good compromise (I ran A LOT of tests on that before setteling for the way it is currently done). The correct way would actually be to do `(color * brightness)^g` but the issue is that that results in color loss. To make it work would require to scale colors to 16bit, apply brightness, then apply gamma, then scale back to 8bit, which is much slower and does only gain "linearity" in brightness scaling which was never really a thing in WLED.

Fixes #5342 

## Release Notes

* **Bug Fixes**
  * Enhanced gamma correction handling to ensure brightness scaling maintains color accuracy and consistency during rendering operations.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1384,7 +1384,7 @@ void WS2812FX::blendSegment(const Segment &topSegment) const {
   const unsigned progInv   = 0xFFFFU - progress;
   uint8_t       opacity    = topSegment.currentBri(); // returns transitioned opacity for style FADE
   uint8_t       cct        = topSegment.currentCCT();
-  if (gammaCorrectCol) opacity = gamma8inv(opacity); // use inverse gamma on brightness for correct color scaling after gamma correction
+  if (gammaCorrectCol) opacity = gamma8inv(opacity); // use inverse gamma on brightness for correct color scaling after gamma correction (see #5343 for details)
 
   Segment::setClippingRect(0, 0);             // disable clipping by default
 


softhack007:
> to make global brightness slider and segment brightness slider behave the same, the segment brightness needs to be gamma-inverted when using gamma for color correction.

Fun fact - I am experimenting in WLEDMM with moving gamma corrections to bus driver level, to exploit HUB75 with 10bits per color. I ran into the same observation that you made: if effects run in "no-gamma space", then segment brightness (segment::opacity) requires inverse gamma to look the same as with previous WLED versions.

blazoncek:
> global brightness slider and segment brightness slider behave the same

FYI these were not meant to behave the same, one is brightness the other is opacity. Opacity with single segment behaves similarly as brightness but is not brightness.

> moving gamma corrections to bus driver

This is the only correct way of handling gamma and I've made that a few days back in my fork.

BTW don't forget that gamma may not always be applied always.

PR #5347, opened 2026-02-05
Introduce comet effect usermod with fire particle system

gustebeast:
A 2D falling comet effect similar to "Matrix" but with a fire particle simulation to enhance the comet trail visuals. Works with custom color palletes, defaulting to "Fire". Supports "small" and "large" comets which are 1px and 3px wide respectively.

Demo: [https://imgur.com/a/i1v5WAy](https://imgur.com/a/i1v5WAy)

## Summary by CodeRabbit

* **New Features**
  * Added PS Comet effect: a 2D falling-comet particle animation with fire-style particles, small/large comet sizes, and configurable speed, frequency, size probability, and length.
  * Supports custom color palettes (default: Fire).

* **Documentation**
  * Added README with installation steps, demo link, and parameter descriptions.

### DIFF
--- a/usermods/PS_Comet/PS_Comet.cpp
+++ b/usermods/PS_Comet/PS_Comet.cpp
@@ -3,32 +3,22 @@
 
 unsigned long nextCometCreationTime = 0;
 
-static bool has_particle_fallen_off_screen(const ParticleSystem2D* PartSys, const PSparticle& particle) {
-  return particle.y < PartSys->maxY * -1;
-}
-
-static bool has_fallen_off_screen(const ParticleSystem2D* PartSys, uint32_t particleIndex) {
-  return has_particle_fallen_off_screen(PartSys, PartSys->sources[particleIndex].source);
-}
-
-static uint16_t mode_static(void) {
-  SEGMENT.fill(SEGCOLOR(0));
-  return strip.isOffRefreshRequired() ? FRAMETIME : 350;
-}
+#define FX_FALLBACK_STATIC { SEGMENT.fill(SEGCOLOR(0)); return; }
+// Use UINT32_MAX - 1 for the "no comet" case so we can add 1 later and not have it overflow
+#define NULL_INDEX UINT32_MAX - 1
 
 ///////////////////////
 //  Effect Function  //
 ///////////////////////
 
-uint16_t mode_pscomet(void) {
+void mode_pscomet() {
   ParticleSystem2D *PartSys = nullptr;
   uint32_t i;
-  // Allocate three comets for every column to allow for large comets with side emitters
-  uint32_t numComets = SEGMENT.vWidth() * 3;
 
   if (SEGMENT.call == 0) { // Initialization
-    if (!initParticleSystem2D(PartSys, numComets)) {
-      return mode_static(); // Allocation failed or not 2D
+    // Try to allocate one comet for every column
+    if (!initParticleSystem2D(PartSys, SEGMENT.vWidth())) {
+      FX_FALLBACK_STATIC; // Allocation failed or not 2D
     }
     PartSys->setMotionBlur(170); // Enable motion blur
     PartSys->setParticleSize(0); // Allow small comets to be a single pixel wide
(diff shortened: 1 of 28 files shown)


DedeHai:
looks good in general, I'll add some comments to consider

usermods/PS_Comet/PS_Comet.cpp
```diff
@@ -0,0 +1,130 @@
+#include "wled.h"
+#include "FXparticleSystem.h"
+
+unsigned long nextCometCreationTime = 0;
+
+static bool has_particle_fallen_off_screen(const ParticleSystem2D* PartSys, const PSparticle& particle) {
```
DedeHai: if you enable gravity and enable killoutofbounds and let the PS handle the movement of your sources, you do not need this function and can just check if TTL=0. you can set gravity to 1 for a very slight acceleration only if that is not disturbing your intentions.
another option would be to set the initial TTL accordingly so your sources die once off-screen. 
or just do this check inside the FX, or add a lambda function.

usermods/PS_Comet/PS_Comet.cpp
```diff
@@ -0,0 +1,130 @@
...
+  PartSys->updateSystem(); // Update system properties (dimensions and data pointers)
+
+  uint32_t sideCometVerticalShift = 2 * PartSys->maxY / (SEGMENT.vHeight() - 1); // Shift emitters on left and right of comet up by 1 pixel
+  uint32_t chosenIndex = hw_random(numComets); // Only allow one column to spawn a comet each frame
+  uint8_t fallingSpeed = 1 + (SEGMENT.speed >> 2);
+  uint16_t cometFrequencyDelay = 2040 - (SEGMENT.intensity << 3);
+
+  // Update the comets
+  for (i = 0; i < numComets; i++) {
+    auto& source = PartSys->sources[i];
```
DedeHai, line 66: do not assume that requested sources = allocated sources. use numsources instead to avoid crashes.

usermods/PS_Comet/PS_Comet.cpp
```diff
@@ -0,0 +1,130 @@
...
+      sourceParticle.ttl = 16;
+    }
+    source.minLife = source.maxLife >> 1;
+    source.vy = 1 + (SEGMENT.speed >> 5); // Emitting speed (upwards)
+
+    bool hasFallenOffScreen = has_fallen_off_screen(PartSys, i);
+
+    if (!hasFallenOffScreen) {
+      // Active comets fall downwards and emit flames
+      sourceParticle.y -= fallingSpeed;
```
DedeHai, line 71: if you want to you could also use vy=-fallingSpeed and let the PS handle the movement, that way you could add gravity as an option. see my PS FX for examples on that.

usermods/PS_Comet/PS_Comet.cpp
```diff
@@ -0,0 +1,130 @@
...
+      }
+      nextCometCreationTime = strip.now + cometFrequencyDelay + hw_random16(cometFrequencyDelay);
+    }
+  }
+
+  // To avoid stretching comets at faster falling speeds, we need to shift the emitted particles as well
+  for (i = 0; i < PartSys->usedParticles; i++) {
+    auto& particle = PartSys->particles[i];
+    if (!has_particle_fallen_off_screen(PartSys, particle)) {
+      particle.y -= fallingSpeed;
```
DedeHai: why not make the emit speed handle this?  the downside would be that the speed would not react immediately to to a slider change.

usermods/PS_Comet/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Installation
+
+To activate the usermod, add the following line to your platformio_override.ini
+```ini
+custom_usermods = ps_comet
+```
+Or if you are already using a usermod, append ps_comet to the list
+```ini
+custom_usermods = audioreactive, ps_comet
```
DedeHai: no comma

DedeHai:
I gave it a quick spin. Parameters may need some tuning: set all sliders except speed to max and it breaks the effect.

gustebeast:
> I gave it a quick spin. Parameters may need some tuning: set all sliders except speed to max and it breaks the effect.

This was due to the code using 3 sources for every column and not cleaning them up properly after falling out of view. With the new implementation it looks normal maxed out. The comet length max makes the large comets appear a bit faint but that's a limitation with the PS system requiring all particle sources to use the same intensity. Ideally the large comets could have their own intensity value since they have their own maxlife/minlife/ttl

DedeHai:
are you confident with this being its own usermod or do you want to add it to the user_fx? The idea behig user_fx is to be a collection of "did not make it into main" effects.

gustebeast:
I'm happy with either. If it were me I'd put all effects that aren't in main into separate usermods so people can choose which ones they want and which ones they don't vs bundling them into a single user_fx where it's all or nothing. I delegate the choice though to more senior contributors.

DedeHai:
either is fine with me. if you give the OK I will merge this in.

gustebeast:
Let's go with the usermod then. By the way I also have this PR approved and ready:
https://github.com/wled/WLED/pull/5320

PR #5351, opened 2026-02-06
Full WiFi scan and apply BSSID if used

### LINKED ISSUE #5345: WLED 0.16.0 alpha on ESP32 can only connect to WiFi access points on channel 1
### What happened?

I have four APs in my home, two are on channel 1, one on channel 6, and one on channel 11. They all share the same ESSID and WPA passphrase. My WLED device is closest to, and gets the strongest signal from, the channel 11 access point. That's the one I would like it to use.

I upgraded my ESP32 (a Gledopto GL-C-017WL-D) from 0.15.3 to 0.16.0 and noticed that WLED now will only connect to the two access points that are on channel 1, resulting in a weaker connection. Interestingly, when I do a network scan, the signal strength of the network that it finds in the scan results is much stronger and what I would expect for the closer AP on channel 11, so I think this just happens when connecting to the network.

Some things I've tried:
- Restarting WLED and doing a full power cycle
- Specifying the BSSID of the close access point in the WLED configuration <- seems to be ignored, does not connect to the specified AP
- Doing a factory reset and setting up WiFi again <- still connects to the weaker channel 1 APs
- Locking the WLED device to the desired access point in the Unifi network controller <- WLED can no longer connect to the network at all until I remove the lock
- Changing WLED's access point channel setting to 11 <- no noticeable effect

I reverted back to 0.15.3 and it immediately connected to the correct access point as it did before.

### To Reproduce Bug

- Install the latest 0.16.0 nightly build of WLED
- Attempt to connect to a wireless network on channel 6 or 11
- Observe that it does not successfully connect

### Expected Behavior

WLED successfully connects to an access point on any channel.

### Install Method

Binary from WLED.me

### What version of WLED?

WLED 0.16.0-alpha (2602030)

ESP32

### Relevant log/trace output

```shell

```

### Anything else?

_No response_

### Code of Conduct

DedeHai:
As found in #5345: 
 - fixes an arduino bug: don't use quick scan (and use the first matching channel) but do a full scan instead
 - BSSID is now passed on to `WiFi.begin()` if set

Fixes #5345 

## Summary by CodeRabbit

* **New Features**
  * Full-channel Wi‑Fi scanning enabled on supported platforms for improved network discovery.
  * BSSID-aware connection logic to prefer a specific access point when configured, improving stability in multi-AP environments.

* **Bug Fixes**
  * Force reconnection when the selected AP (BSSID) changes so the device reliably connects to the intended access point.

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -735,7 +735,7 @@ void WLED::initConnection()
       // check if user BSSID is non zero for current WiFi config
       for (int i = 0; i < sizeof(multiWiFi[selectedWiFi].bssid); i++) {
         if (multiWiFi[selectedWiFi].bssid[i] != 0) {
-        bssid = multiWiFi[selectedWiFi].bssid; // BSSID set, assign pointer and continue
+          bssid = multiWiFi[selectedWiFi].bssid; // BSSID set, assign pointer and continue
           break;
         }
       }


wled00/wled.cpp
```diff
@@ -503,7 +503,9 @@ void WLED::setup()
 
   if (strcmp(multiWiFi[0].clientSSID, DEFAULT_CLIENT_SSID) == 0 && !configBackupExists())
     showWelcomePage = true;
+
   WiFi.persistent(false);
+  WiFi.setScanMethod(WIFI_ALL_CHANNEL_SCAN);
```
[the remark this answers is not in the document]
DedeHai: does the ESP8266 always use a full scan or is there an equivalent call required?

DedeHai:
@coderabbitai the BSSID seems to be applied only after a reboot if changed at runtime through the UI. Can this be fixed?

wled00/wled.cpp
```diff
@@ -727,7 +731,14 @@ void WLED::initConnection()
       wifi_station_clear_enterprise_username();
       wifi_station_clear_enterprise_password();
 #endif
-      WiFi.begin(multiWiFi[selectedWiFi].clientSSID, multiWiFi[selectedWiFi].clientPass);
+      uint8_t *bssid = nullptr;
+      for (int i = 0; i < sizeof(multiWiFi[selectedWiFi].bssid); i++) {
+        if (multiWiFi[selectedWiFi].bssid[i]) {
```
softhack007: is `bssid[i]` correct here? It looks a bit awkward to scan over all array items and then repeatedly copy a pointer.

might be clearer like this
````c++
     bool bssidValid = false;
     for (int i = ...) if (multiWiFi[selectedWiFi].bssid[i] != 0) bssidValid = true;
     if (bssidValid) bssid = multiWiFi[selectedWiFi].bssid;
````

File: wled00/wled.cpp [hunk 3]
```diff
@@ -746,7 +757,14 @@ void WLED::initConnection()
 #endif
     }
 #else // WLED_ENABLE_WPA_ENTERPRISE
-    WiFi.begin(multiWiFi[selectedWiFi].clientSSID, multiWiFi[selectedWiFi].clientPass); // no harm if called multiple times
+    uint8_t *bssid = nullptr;
+    for (int i = 0; i < sizeof(multiWiFi[selectedWiFi].bssid); i++) {
```
softhack007, line 763: see previous comment

wled00/wled.cpp
DedeHai: it only goes through the numbers in the selected wifi once, if one is non zero it assigns the pointer and moves on (break;)

softhack007: ah OK, did not see the ``break;``. 
Im my proposal the logic looks a bit clearer, even when a few cycles get wasted to check all 6 bytes of the BSSID. 

up to you if you want to keep the current code

DedeHai: I can add the ` != 0` and extend the comment.

DedeHai:
@softhack007 btw this change of scanning all WiFi "breaks" the bootup glitch fix - well not really but it needs to be extended. I have a solution: just always update NVM on bootup -  it is quite safe and will not cause flash wear in any realistic use case. Will update that PR once this is merged.

DedeHai:
note to @netmindz this fix needs 0.15 backport

PR #5355, opened 2026-02-07
Add experimental ESP32-C5 support (WiFi 6 dual-band)

jonny190:
## Summary
- Adds ESP32-C5 support to WLED for WiFi 6 dual-band (2.4GHz + 5GHz) capability
- ESP32-C5 is RISC-V based, treated similarly to C3 (2 RMT channels, no parallel I2S for LEDs)
- Modifies 25 files with `CONFIG_IDF_TARGET_ESP32C5` conditionals alongside existing C3/C6 checks
- Uses pioarduino platform since Tasmota framework doesn't include C5 Arduino libs yet

## Key Changes
- **platformio.ini**: New `[esp32c5]` and `[env:esp32c5dev]` build sections modeled after C6
- **bus_manager.cpp**: LEDC register fix for IDF 5.5.0 (`duty_init` vs `duty` on C5)
- **mbedtls_sha1_shim.cpp**: Guard for IDF 5.5.0+ where SHA1 is built-in
- **esp_dmx**: Excluded via `lib_ignore` (doesn't support C5 UART registers)
- **OTA/Web UI**: C5 chip identification (ID: 0x0017)

## Hardware Test Results
- **Board**: ESP32-C5-DevKitC-1 (4MB flash, no PSRAM)
- **Build**: Compiles successfully (89.3% flash, 27.6% RAM)
- **Boot**: Device boots and prints startup message
- **WiFi AP**: WLED-AP visible and functional
- **RMT LED output**: Shows `rmt_tx_wait_all_done: flush timeout` - the IDF 5.x RMT driver_ng API works without the legacy CONFLICT crash, but transmissions time out. This likely needs NeoPixelBus upstream C5 support. No LED strip was connected during testing.

## Known Limitations
- **Experimental**: Requires pioarduino community platform (not Tasmota)
- **RMT flush timeouts**: LED output needs NeoPixelBus C5 support (non-fatal, device runs fine)
- **esp_dmx disabled**: C5 UART register layout differs from other ESP32 variants
- **IDF 5.5.0**: pioarduino uses newer IDF than Tasmota (5.5.0 vs 5.3.3), requiring `duty_init` and SHA1 shim fixes

## Relation to PR #5354
This PR supersedes #5354 (which targeted `main`). Per @softhack007's feedback, this PR targets V5-C6 and follows the V5-C6 build structure.

## Test plan
- [x] Firmware compiles for esp32c5dev environment
- [x] Firmware flashes to ESP32-C5-DevKitC-1 via esptool
- [x] Device boots without crash
- [x] WLED-AP WiFi network appears
- [ ] LED strip output (needs NeoPixelBus C5 RMT support or bit-bang fallback)
- [ ] 5GHz WiFi connection test
- [ ] Web UI functionality test via WLED-AP

## Summary by CodeRabbit

* **New Features**
  - Added support for ESP32-C5 microcontroller boards.
  - WiFi band information (2.4GHz/5GHz) now displayed in device information.

* **Chores**
  - Extended platform configuration and compatibility adjustments for new hardware variants.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -478,7 +478,7 @@ lib_deps =
   ;; the upstream NeoPixelBus with a C5-patched fork; using both causes duplicate compilation.
   ${esp32_all_variants.lib_deps}
   fastled/FastLED @ 3.10.1
-  https://github.com/jonny190/NeoPixelBus.git#d4bbe3f899a9d7c39ff4569827b2cf4665b3b64d ;; C5 support: I2S exclusion + bit-bang CSR/GPIO fixes on top of netmindz CORE3
+  https://github.com/jonny190/NeoPixelBus.git#103cfb585ef8efcdd6645ebb9938bf7012d093fd ;; C5 support: I2S exclusion, bit-bang CSR/GPIO, channel ctor overload
   https://github.com/Aircoookie/ESPAsyncWebServer.git#v2.4.2
   marvinroger/AsyncMqttClient @ 0.9.0
   https://github.com/blazoncek/QuickESPNow.git#optional-debug
@@ -502,7 +502,7 @@ build_flags = ${common.build_flags} ${esp32c5.build_flags} -D WLED_RELEASE_NAME=
   -D WLED_DISABLE_INFRARED       ;; library not compatible with -C5
   -D WLED_DISABLE_ESPNOW         ;; not yet tested
   -D WLED_DISABLE_ALEXA          ;; compile errors
-  -D WLED_DISABLE_WEBSOCKETS     ;; not yet tested
+  -D WLED_DEBUG
 upload_speed = 460800
 lib_deps = ${esp32c5.lib_deps}
 lib_ignore = ${esp32c5.lib_ignore}
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -1,6 +1,9 @@
 #define WLED_DEFINE_GLOBAL_VARS //only in one source file, wled.cpp!
 #include "wled.h"
 #include "wled_ethernet.h"
+#ifdef ARDUINO_ARCH_ESP32
+#include "esp_mac.h"
+#endif
 #include "ota_update.h"
 #ifdef WLED_ENABLE_AOTA
   #define NO_OTA_PORT
@@ -442,6 +445,18 @@ void WLED::setup()
   escapedMac = WiFi.macAddress();
   escapedMac.replace(":", "");
   escapedMac.toLowerCase();
+#ifdef ARDUINO_ARCH_ESP32
+  // WiFi.macAddress() may return all zeros if the WiFi netif is not yet created
+  // (e.g. on ESP32-C5 where WiFi.mode() hasn't been called yet). Fall back to
+  // reading the base MAC directly from eFuse.
+  if (escapedMac == "000000000000") {
+    uint8_t mac[6] = {0};
+    esp_read_mac(mac, ESP_MAC_WIFI_STA);
+    char buf[13];
+    sprintf(buf, "%02x%02x%02x%02x%02x%02x", mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
+    escapedMac = buf;
+  }
+#endif
 
   WLED_SET_AP_SSID(); // otherwise it is empty on first boot until config is saved
   multiWiFi.push_back(WiFiConfig(CLIENT_SSID,CLIENT_PASS)); // initialise vector with default WiFi


netmindz:
As a general comment, I wonder if there is a better way of handling the ever growing list of CONFIG_IDF_TARGET_ESP32* ifdefs, can we invert or define our own values that relate more to the specific feature we are testing for rather than the hardware @softhack007 @willmmiles 

Ifdef WLED_HW_FEATURE_X

Rather than say a long list of or conditions for when we say want to know about Hardware serial support or I2S count?

softhack007:
> CONFIG_IDF_TARGET_ESP32* ifdefs, can we invert or define our own values that relate more to the specific feature we are testing for rather than the hardware @softhack007 @willmmiles
> 
> Ifdef WLED_HW_FEATURE_X

@netmindz you're right, I think we have to go that way - maybe with exception of pinmager.cpp and const.h/wled.h which are meant for board specific parameters.

wled00/NodeStruct.h
```diff
@@ -14,6 +14,8 @@
 #define NODE_TYPE_ID_ESP32S2         33 // etc
 #define NODE_TYPE_ID_ESP32S3         34
 #define NODE_TYPE_ID_ESP32C3         35
+#define NODE_TYPE_ID_ESP32C6         36
+#define NODE_TYPE_ID_ESP32C5         41
```
softhack007: Please check if these values match with nodestruct.h in ``main``. I've recently added new node types there, based on the original types fron espEasy.

wled00/wled.cpp
```diff
@@ -696,7 +696,56 @@ if (multiWiFi.empty()) {                       // guard: handle empty WiFi list
...
+#if defined(CONFIG_IDF_TARGET_ESP32C5)
+    // ESP32-C5: Prefer 5GHz WiFi 6 for better performance
+    DEBUG_PRINTF_P(PSTR("ESP32-C5: Trying 5GHz for SSID: %s\n"), multiWiFi[selectedWiFi].clientSSID);
+
+    WiFi.disconnect(true);
+    WiFi.mode(WIFI_STA);
+
+    // Set country code to enable all 5GHz channels
+    wifi_country_t country = {
+      .cc = "US",
```
softhack007: How do we handle this in other countries? In some countries there are fines/penalty payments if you use reserved channels.

Would the code also work without esp_wifi_set_country()?

platformio.ini
```diff
@@ -454,6 +454,57 @@ build_unflags = ${env:esp32c6dev_8MB.build_unflags} -D WLED_RELEASE_NAME=\"ESP32
...
+build_unflags = ${esp32_idf_V5.build_unflags}
+  -D WLED_ENABLE_DMX_INPUT  ;; esp_dmx library doesn't support C5 UART registers yet
+build_flags = -g
+  -Wshadow=compatible-local
+  -DARDUINO_ARCH_ESP32
+  -DARDUINO_ARCH_ESP32C5
+  -DCONFIG_IDF_TARGET_ESP32C5=1
+  -D CONFIG_ASYNC_TCP_USE_WDT=0
+  -DCO
+  -DARDUINO_USB_MODE=1
```
softhack007: Is this mandatory for all possible -C5 boards?

platformio.ini
```diff
@@ -454,6 +454,57 @@ build_unflags = ${env:esp32c6dev_8MB.build_unflags} -D WLED_RELEASE_NAME=\"ESP32
...
+  ${esp32_idf_V5.lib_deps}
+lib_ignore = ${esp32_idf_V5.lib_ignore}
+  esp_dmx  ;; esp_dmx library doesn't support C5 UART registers yet
+
+[env:esp32c5dev]
+;; ESP32-C5 "devkit C" with 4MB flash - EXPERIMENTAL
+extends = esp32c5
+platform = ${esp32c5.platform}
+platform_packages = ${esp32c5.platform_packages}
+framework = arduino
```
softhack007: ``framework = arduino`` should not be necessary here

platformio.ini
```diff
@@ -454,6 +454,57 @@ build_unflags = ${env:esp32c6dev_8MB.build_unflags} -D WLED_RELEASE_NAME=\"ESP32
...
+upload_speed = 460800
+lib_deps = ${esp32c5.lib_deps}
+lib_ignore = ${esp32c5.lib_ignore}
+  IRremoteESP8266
+  QuickEspNow
+
+board_build.partitions = ${esp32.big_partitions}
+board_build.f_flash = 80000000L
+board_build.flash_mode = qio
+board_build.arduino.memory_type = qio_qspi
```
softhack007: This board has PSRAM?

wled00/util.cpp
```diff
@@ -633,8 +633,8 @@ int32_t hw_random(int32_t lowerlimit, int32_t upperlimit) {
 
 // PSRAM compile time checks to provide info for misconfigured env
 #if defined(BOARD_HAS_PSRAM)
-  #if defined(IDF_TARGET_ESP32C3) || defined(CONFIG_IDF_TARGET_ESP32C6) || defined(CONFIG_IDF_TARGET_ESP32C61) || defined(ESP8266)
-    #error "ESP32-C3 and ESP8266 with PSRAM is not supported, please remove BOARD_HAS_PSRAM definition"
+  #if defined(IDF_TARGET_ESP32C3) || defined(CONFIG_IDF_TARGET_ESP32C5) || defined(CONFIG_IDF_TARGET_ESP32C6) || defined(CONFIG_IDF_TARGET_ESP32C61) || defined(ESP8266)
+    #error "ESP32-C3/C5 and ESP8266 with PSRAM is not supported, please remove BOARD_HAS_PSRAM definition"
   #else
```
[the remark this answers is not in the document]
softhack007: Maybe the check is wrong, too, afaik there are C6 (maybe C61, too?) boards with PSRAM.

@@coderabbitai Please to a quick web research to find out which of the new MCUs (C6,C61,C5,P4) are actually able to support PSRAM.

wled00/wled.cpp
```diff
@@ -696,7 +696,56 @@ if (multiWiFi.empty()) {                       // guard: handle empty WiFi list
...
+    WiFi.setBandMode(WIFI_BAND_MODE_5G_ONLY);
+    DEBUG_PRINTLN(F("ESP32-C5: Band mode set to 5GHz ONLY"));
+
+    wifi_config_t wifi_config = {};
+    strncpy((char*)wifi_config.sta.ssid, multiWiFi[selectedWiFi].clientSSID, sizeof(wifi_config.sta.ssid) - 1);
+    strncpy((char*)wifi_config.sta.password, multiWiFi[selectedWiFi].clientPass, sizeof(wifi_config.sta.password) - 1);
+    wifi_config.sta.scan_method = WIFI_ALL_CHANNEL_SCAN;
+    wifi_config.sta.sort_method = WIFI_CONNECT_AP_BY_SIGNAL;
+    wifi_config.sta.threshold.authmode = WIFI_AUTH_WPA2_PSK;
+    esp_wifi_set_config(WIFI_IF_STA, &wifi_config);
```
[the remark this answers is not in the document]
softhack007: @troyhacks as you already managed to get the C5 connect to 5ghz networks - what would you recommend?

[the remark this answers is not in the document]
softhack007: @@coderabbit is it possible to achieve the same with the Arduino-style WiFi or Network class?

wled00/NodeStruct.h
jonny190: I've updated NodeStruct.h to match the full set from main, including the ESP Easy project node types (C2=37, H2=38, C6=39, C61=40, C5=41, P4=42). The C6 value was incorrectly set to 36 instead of 39. Fixed in the next push.

wled00/wled.cpp
jonny190: You're absolutely right, hardcoding "US" country code is problematic for international users and could cause legal issues with reserved channels. I've removed the entire esp_wifi_set_country() block and the custom C5 WiFi connection logic. The C5 now uses the standard WiFi.begin() path just like all other ESP32 variants. The ESP-IDF will pick up country info from the AP via 802.11d, which is the correct approach. 5GHz band selection can be revisited later as a user-configurable option if needed.

platformio.ini
jonny190: Moved ARDUINO_USB_MODE=1 out of the generic [esp32c5] section and into [env:esp32c5dev] since it may vary per board. Added the same comment pattern as C6 in the generic section noting that ARDUINO_USB_MODE and ARDUINO_USB_CDC_ON_BOOT should be set by the board.json or custom override.

platformio.ini
jonny190: Removed

platformio.ini
jonny190: No, the ESP32-C5-DevKitC-1 has no PSRAM. The qio_qspi memory type refers to the flash configuration only (quad SPI flash), not PSRAM. Added a clarifying comment: ;; flash config only, this board has no PSRAM

jonny190:
@netmindz @softhack007 
Agreed, the growing list of CONFIG_IDF_TARGET_ESP32* conditionals is becoming unwieldy. Introducing feature-based defines like WLED_HW_FEATURE_I2S_PARALLEL, WLED_HW_FEATURE_RMT_CHANNELS, etc. set once in const.h based on the target would be much cleaner. Happy to help with that refactor as a follow-up if desired, but i feel it's a bigger change that probably deserves its own PR?

wled00/wled.cpp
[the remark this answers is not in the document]
softhack007: @coderabbitai could we omit esp_wifi_set_country_code("US,..) ? 

I think that all boards will come with some default country, and the wifi AP (wifi router) will any way provide its supported channel list to clients?

softhack007:
> Happy to help with that refactor as a follow-up if desired, but i feel it's a bigger change that probably deserves its own PR?

I agree with you. This clean up is better in a separate PR.

jonny190:
Suggested changes have been made and tested on my ESP32-C5 and show to be working at least in my limited testing

wled00/wled.cpp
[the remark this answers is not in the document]
softhack007: @coderabbitai could we omit esp_wifi_set_country_code("US,..) ? 

I think that all boards will come with some default country, and the wifi AP (wifi router) will any way provide its supported channel list to clients?

.claude/settings.local.json
```diff
@@ -0,0 +1,7 @@
+{
```
softhack007: Please remove this file before we merge your pr.

jonny190: Apologies, done :)

wled00/idf_component.yml
```diff
@@ -0,0 +1,2 @@
+dependencies:
```
softhack007: @jonny190 do we need these two esp-idf specific files in or source tree? Are they auto-generated?

wled00/js_iro.h
```diff
@@ -0,0 +1,636 @@
+/*
+ * More web UI HTML source arrays.
```
softhack007: @jonny190 please delete this file, its auto generated,

softhack007:
> platformio.ini (1)
> > `457-505`: **Pin the pioarduino platform to a specific release version for consistency and reproducibility.**
> > The `stable` reference can change over time. All other platforms in platformio.ini use explicit version pinning (e.g., Tasmota's `2025.08.30`, ESP32-IDF-V5's `2025.08.30`). Consider pinning to a specific pioarduino release once validated.

@jonny190 this minor finding from the rabbit makes sense, please pin the platform version explicitly

softhack007:
> wled00/bus_wrapper.h (1)
> > `499-509`: _⚠️ Potential issue_ | _🟠 Major_
> > **Exclude ESP32‑C5 from the classic ESP32 channel remap.**
> > Line 506 still applies the classic ESP32 “I2S1 first bus” channel decrement to C5 (because it isn’t excluded). That can collapse RMT channels (e.g., channel 1 → 0) and break multi‑bus output on C5.

@jonny190 another one that should get fixed.

Jerrys-modz:
I would love to help test. I have a few c5s laying around but I don’t see it building any bin files for the c5.

softhack007:
> wled00/wled.cpp (1)
> > `876-878`: **Optional: name the 5 GHz cutoff instead of the raw `36`.**
> > This improves readability for future maintainers

@jonny190 please check - wled actually has "technical debt" on magic numbers, it would be better to have named  constants.

softhack007:
> I would love to help test. I have a few c5s laying around but I don’t see it building any bin files for the c5.

@Jerrys-modz very much appreciated  ❤️ 
As this PR progresses very quickly, it should be part of the ``C6-V5`` branch soon, and you should find bins in the CI  builds for that branch after merge.

@jonny190 please add your C5 buildenv to the default list here

https://github.com/jonny190/WLED/blob/11f5be1861f96e92b24aac6040bbae9c8c552379/platformio.ini#L32

I will try to build your branch locally later tonight, as a final check before merge.

usermods/audioreactive/audio_reactive.cpp
```diff
@@ -672,7 +672,7 @@ class AudioReactive : public Usermod {
     static const char _dynamics[];
     static const char _frequency[];
     static const char _inputLvl[];
-#if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32S2) && !defined(CONFIG_IDF_TARGET_ESP32C3) && !defined(CONFIG_IDF_TARGET_ESP32S3)
+#if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32S2) && !defined(CONFIG_IDF_TARGET_ESP32C3) && !defined(CONFIG_IDF_TARGET_ESP32C5) && !defined(CONFIG_IDF_TARGET_ESP32S3)
     static const char _analogmic[];
 #endif
```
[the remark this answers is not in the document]
softhack007: We can address AR compatibility issues later. There is an upcoming activity to make AR an out-of-tree usermod, where we can also address such details. For now "it compiles" is good enough.

wled00/src/dependencies/dmx/ESPDMX.cpp
```diff
@@ -11,7 +11,7 @@
 // - - - - -
 
 /* ----- LIBRARIES ----- */
-#if defined(ESP8266) || defined(CONFIG_IDF_TARGET_ESP32C3) || defined(CONFIG_IDF_TARGET_ESP32S2)
+#if defined(ESP8266) || defined(CONFIG_IDF_TARGET_ESP32C3) || defined(CONFIG_IDF_TARGET_ESP32C5) || defined(CONFIG_IDF_TARGET_ESP32S2)
```
netmindz, line 14: Do you know if ESPDMX actually supports C5 or just a guess or just a way to make the build compile?

wled00/FXparticleSystem.cpp
```diff
@@ -493,7 +493,7 @@ void ParticleSystem2D::applyGravity(PSparticle &part) {
 // note: a coefficient smaller than 0 will speed them up (this is a feature, not a bug), coefficient larger than 255 inverts the speed, so don't do that
 void ParticleSystem2D::applyFriction(PSparticle &part, const int32_t coefficient) {
   // note: not checking if particle is dead can be done by caller (or can be omitted)
-  #if defined(CONFIG_IDF_TARGET_ESP32C3) || defined(ESP8266) // use bitshifts with rounding instead of division (2x faster)
+  #if defined(CONFIG_IDF_TARGET_ESP32C3) || defined(CONFIG_IDF_TARGET_ESP32C5) || defined(ESP8266) // use bitshifts with rounding instead of division (2x faster)
```
netmindz, line 496: Is there a build flag for RISC-V we can use here @DedeHai ?

wled00/FX.h
```diff
@@ -70,10 +70,10 @@ extern byte realtimeMode;           // used in getMappedPixelIndex()
 #define WLED_FPS         42
 #define FRAMETIME_FIXED  (1000/WLED_FPS)
 #define FRAMETIME        strip.getFrameTime()
-#if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32C3) && !defined(CONFIG_IDF_TARGET_ESP32S2)
-  #define MIN_FRAME_DELAY  2                                              // minimum wait between repaints, to keep other functions like WiFi alive 
-#elif defined(CONFIG_IDF_TARGET_ESP32S2) || defined(CONFIG_IDF_TARGET_ESP32C3)
-  #define MIN_FRAME_DELAY  3                                              // S2/C3 are slower than normal esp32, and only have one core
+#if defined(ARDUINO_ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32C3) && !defined(CONFIG_IDF_TARGET_ESP32C5) && !defined(CONFIG_IDF_TARGET_ESP32S2)
+  #define MIN_FRAME_DELAY  2                                              // minimum wait between repaints, to keep other functions like WiFi alive
+#elif defined(CONFIG_IDF_TARGET_ESP32S2) || defined(CONFIG_IDF_TARGET_ESP32C3) || defined(CONFIG_IDF_TARGET_ESP32C5)
```
netmindz, line 75: Is there a build flag that indicates when we have more than 1 core?

wled00/src/dependencies/dmx/ESPDMX.cpp
jonny190: There are two separate DMX components here:

ESPDMX (DMX output, internal src/dependencies/dmx/ESPDMX.cpp) - This uses Arduino Serial and does compile for C5 (we added C5 to its #if guard). It should work since C5 has standard UART hardware, but I haven't tested DMX output on C5 hardware yet.

esp_dmx (DMX input, external library) - This one genuinely does not support C5. It uses chip-specific UART register access that hasn't been updated for C5, so it was excluded via lib_ignore.

The #error guard in wled.h is a safety net since WLED_ENABLE_DMX is disabled by default anyway (//#define WLED_ENABLE_DMX in wled.h line 43). The ESPDMX compile guard is there to make the build succeed - whether DMX output actually

wled00/json.cpp
```diff
@@ -811,7 +815,7 @@ void serializeInfo(JsonObject root)
     wifi_info[F("txPower")] = (int) WiFi.getTxPower();
     wifi_info[F("sleep")] = (bool) WiFi.getSleep();
   #endif
-  #if !defined(CONFIG_IDF_TARGET_ESP32C2) && !defined(CONFIG_IDF_TARGET_ESP32C3) && !defined(CONFIG_IDF_TARGET_ESP32S2) && !defined(CONFIG_IDF_TARGET_ESP32S3) && !defined(CONFIG_IDF_TARGET_ESP32C6) && !defined(CONFIG_IDF_TARGET_ESP32P4) 
+  #if !defined(CONFIG_IDF_TARGET_ESP32C2) && !defined(CONFIG_IDF_TARGET_ESP32C3) && !defined(CONFIG_IDF_TARGET_ESP32C5) && !defined(CONFIG_IDF_TARGET_ESP32S2) && !defined(CONFIG_IDF_TARGET_ESP32S3) && !defined(CONFIG_IDF_TARGET_ESP32C6) && !defined(CONFIG_IDF_TARGET_ESP32P4)
```
netmindz: Surely there must be a better way to write this line? which targets do not support ESP.getChipModel?

Any suggestions @coderabbitai?

wled00/FX.h
jonny190: Good question - yes, IDF provides SOC_CPU_CORES_NUM in soc_caps.h for each chip (2 for ESP32/S3, 1 for S2/C3/C5/C6/H2). There's also CONFIG_FREERTOS_UNICORE for single-core chips. Either could replace some of the growing CONFIG_IDF_TARGET_* lists where the check is really about core count rather than a specific chip. As discussed above though, that refactor we believe is better suited to a separate PR.

jonny190:
I've pushed the suggested changes to hopefully tie things up

wled00/FXparticleSystem.cpp
DedeHai: there is such a flag, it is used in buswrapper for high priority RMT driver.

DedeHai: however, this is not about risc V per se but slow hardware division. 
edit:see https://github.com/wled/WLED/issues/4206

jonny190:
The CI build for esp32c5dev was failing because the upstream NeoPixelBus (pinned at netmindz/NeoPixelBus#2f05279) doesn't include ESP32-C5 in the exclusion guard for its legacy I2S driver (Esp32_i2s.c). This file includes soc/rtc_cntl_reg.h which doesn't exist on C5, causing a fatal compile error.

This only showed up in CI because local builds had a cached .o file from a previously patched copy. CI downloads a fresh library each time and hits the missing header.

Fix: Added a PlatformIO pre-build script (pio-scripts/patch_neopixelbus_c5.py) that automatically patches the NeoPixelBus Esp32_i2s.c file after download, adding !defined(CONFIG_IDF_TARGET_ESP32C5) to the I2S exclusion guard. This is the same approach used elsewhere in PlatformIO projects to handle upstream library incompatibilities without forking.

The script only runs for C5 builds and is idempotent (skips if already patched). Once upstream NeoPixelBus adds native C5 support, the script can be removed.

Can I push this fix to the branch?

netmindz:
Rather than pio-scripts/patch_neopixelbus_c5.py - please update to point to a fork of NeoPixelBus that has the right bits we need, so this can properly then be merged upstream, hacking the source of a library is really nasty @jonny190

jonny190:
PR's are in for both

wled00/FX.h
softhack007: I think ``#if SOC_CPU_CORES_NUM > 1`` would be good

wled00/FXparticleSystem.cpp
softhack007:
````c++

#ifdef ESP32

#if defined(__riscv) 
// for all riscv based chips: P4,C3, C5, C6
#endif

#if defined(__XTENSA__)
// for S3, S2, and classic esp32
#endif

#endif
````

**Comment by softhack007** [comment]
@netmindz initially my idea was to postpone solving these extra long #ifdef chains for later.
But I like the idea of relying on ``SOC_CPU_...`` , `` __riscv__`` , `` __XTENSA__`` .
Maybe we address a few of the cases here so ideas don't get lost. But no need to solve everything in this one PR.

File: wled00/json.cpp [hunk 16]
```diff
@@ -811,7 +816,7 @@ void serializeInfo(JsonObject root)
     wifi_info[F("txPower")] = (int) WiFi.getTxPower();
     wifi_info[F("sleep")] = (bool) WiFi.getSleep();
   #endif
-  #if !defined(CONFIG_IDF_TARGET_ESP32C2) && !defined(CONFIG_IDF_TARGET_ESP32C3) && !defined(CONFIG_IDF_TARGET_ESP32S2) && !defined(CONFIG_IDF_TARGET_ESP32S3) && !defined(CONFIG_IDF_TARGET_ESP32C6) && !defined(CONFIG_IDF_TARGET_ESP32P4) 
+  #if defined(CONFIG_IDF_TARGET_ESP32)
```
softhack007, line 819: just to explain: ESP.getChipModel() does exists on classic esp32, but the result is just too "wordy" - e.g. "ESP32-D0QF6-V3". We simply wanted "esp32".

I think the fix is OK.

jonny190:
**Latest push (1a025f5):**

**MAC address fix (`wled00/wled.cpp`):**
On ESP32-C5, `WiFi.macAddress()` returns all zeros (`000000000000`) because it's called at line 442 before `WiFi.mode(WIFI_STA)` at line 481. At that point the WiFi netif hasn't been created yet, so `esp_netif_get_mac()` fails internally — and Arduino ESP32 v3.x has no fallback. Added a guard that detects the all-zeros case and falls back to `esp_read_mac(mac, ESP_MAC_WIFI_STA)` to read the base MAC directly from eFuse. This is wrapped in `#ifdef ARDUINO_ARCH_ESP32` so it only applies to ESP32 targets.

**NeoPixelBus fork update (`platformio.ini`):**
Updated the NeoPixelBus fork pin from `d4bbe3f` to `103cfb5`, which includes a channel constructor overload fix needed for C5 builds.

**Build flags (`platformio.ini`):**
- Enabled `WLED_DEBUG` for the C5 build to aid testing/development
- Removed `WLED_DISABLE_WEBSOCKETS` — websockets work fine on C5

Both changes have been tested on hardware with confirmed WiFi 6 connectivity on 5GHz (channel 44).

softhack007:
@jonny190 thanks 👍 
My "it compiles" check also was successful, so I'll merge your PR back into C6-V5.

netmindz:
> @netmindz initially my idea was to postpone solving these extra long #ifdef chains for later. But I like the idea of relying on `SOC_CPU_...` , ` __riscv__` , ` __XTENSA__` . Maybe we address a few of the cases here so ideas don't get lost. But no need to solve everything in this one PR.

I've messaged @willmmiles on discord and he likes the idea too.

In terms of what we do here, yeah the choice would be too either 

* "phase in" by using this new style here for the required changes
* create a new PR that cleans up all the legacy then merge that into this one

jonny190:
@netmindz @softhack007 @willmmiles Agreed - a separate cleanup PR is the way to go. This PR is scoped to C5 support and is ready to merge, so bolting a refactor onto it now would just risk delays and regressions.

For the follow-up PR I'm thinking something like:

- Define feature macros in `const.h` based on target (e.g. `WLED_HW_HAS_I2S_PARALLEL`, `WLED_HW_RMT_CHANNELS`, `WLED_HW_IS_SINGLE_CORE`)
- Could also lean on existing IDF/arch macros where they already express what we need (`SOC_CPU_CORES_NUM`, `__riscv__`, `__XTENSA__`)
- Then sweep through bus_wrapper.h, FX.h, wled.cpp, etc. replacing the long `CONFIG_IDF_TARGET_*` chains

Happy to take that on once this one lands. Should make adding future targets (C61, H4, etc.) much less painful.

jonny190:
The esp32c5dev CI build is failing with:

```
NeoPixelBus@src-.../internal/methods/ESP/ESP32/Core_2_x/Esp32_i2s.c:44:
fatal error: soc/rtc_cntl_reg.h: No such file or directory
```

**Root cause**: PlatformIO always merges the root `[env]` `lib_deps` into every environment. Line 170 of platformio.ini pulls in `netmindz/NeoPixelBus.git#2f05279` globally. The `[esp32c5]` section adds the C5-patched fork (`jonny190/NeoPixelBus.git#103cfb5`), but it can't override the global entry. So PIO downloads **both** copies - the CI dependency graph confirms this:

```
|-- NeoPixelBus @ 2.9.0+sha.2f05279    <-- upstream (from global [env] lib_deps)
|-- NeoPixelBus @ 2.9.0+sha.103cfb5    <-- C5-patched fork
```

The upstream copy gets compiled for C5 and hits the missing `soc/rtc_cntl_reg.h` header (which doesn't exist on C5). Local builds didn't catch this because PIO had a cached `.o` from a previously patched copy.

**Proposed fix**: Add `NeoPixelBus` to `lib_ignore` in the `[esp32c5]` section. The fork installed via git URL gets an internal name of `NeoPixelBus@src-<hash>`, which PIO treats differently from the plain library name, so the ignore filter should only block the upstream copy.

```ini
lib_ignore = ${esp32_idf_V5.lib_ignore}
  esp_dmx
  NeoPixelBus  ;; ignore upstream; C5-patched fork uses NeoPixelBus@src-* name
```

I'll test this locally and push if it works.

jonny190:
@netmindz The esp32c5dev CI failure is because PlatformIO always merges the global `[env]` `lib_deps` into every environment - so the upstream NeoPixelBus (`netmindz/NeoPixelBus#2f05279`) gets pulled in alongside the C5-patched fork, and the upstream copy fails to compile on C5 due to the missing `soc/rtc_cntl_reg.h` in `Esp32_i2s.c`.

I tried using `lib_ignore = NeoPixelBus` but that blocks both copies since they share the same library name.

The cleanest fix would be merging the C5 support PR into your NeoPixelBus fork (https://github.com/netmindz/NeoPixelBus/pull/1), then updating the global commit hash on line 170 of platformio.ini. That way there's only one NeoPixelBus reference and it works for all targets including C5 - no separate fork needed.

Would you be able to take a look at that PR?

PR #5368, opened 2026-02-11
Clean up global variables namespace, save a few 100 bytes of flash

softhack007:
Reduces the scope of some variables and functions to "static" - if not used anywhere else. 

Making them static avoids name conflicts, cleans up the global scope and allows for better optimization by the compiler.
We also save a few bytes, because the compiler does not need to keep copies of previously global functions.

## Summary by CodeRabbit

* **Refactor**
  * Tightened internal scope across multiple modules by restricting many globals and helper functions to file-local visibility and removing a few shared declarations.
  * Cleaned up internal helper declarations and loop variable naming to avoid shadowing.
  * No user-facing behavior, features, or public API changes; runtime behavior and error handling remain unchanged.

### DIFF
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -38,7 +38,7 @@ static constexpr bool validatePinsAndTypes(const unsigned* types, unsigned numTy
 //simple macro for ArduinoJSON's or syntax
 #define CJSON(a,b) a = b | a
 
-void getStringFromJson(char* dest, const char* src, size_t len) {
+static inline void getStringFromJson(char* dest, const char* src, size_t len) {
   if (src != nullptr) strlcpy(dest, src, len);
 }
 
--- a/wled00/e131.cpp
+++ b/wled00/e131.cpp
@@ -5,6 +5,7 @@
 #define MAX_CHANNELS_PER_UNIVERSE 512
 
 // forward declarations
+static void handleDDPPacket(e131_packet_t* p);
 static void handleArtnetPollReply(IPAddress ipAddress);
 static void prepareArtnetPollReply(ArtPollReply *reply);
 static void sendArtnetPollReply(ArtPollReply *reply, IPAddress ipAddress, uint16_t portAddress);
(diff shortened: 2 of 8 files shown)


softhack007:
@coderabbitAI please scan the usermods/ folder for "extern" references to the variables reduced to ``static`` in this PR.

wled00/ntp.cpp
softhack007, line 157: @coderabbitAI is the delete operator used correctly here? or should it be delete[] ?

softhack007:
@netmindz is there a way to trigger a "manual" run of the usermods CI job on this PR branch? Just to be sure that no usermod breaks due to my changes.

softhack007:
@coderabbitai lets repeat your previous analysis to global functions (C style) that could be made ``static``
- function is globally visible, but no other source code file (wled00/ or usermods/) references it
- functions and variables that are declared globally in fcn_declare.h, but not referenced outside of a single translation unit.

softhack007:
@coderabbitAI Please do a second check over your results in https://github.com/wled/WLED/pull/5368#issuecomment-3884260183. I need to be absolutely confident that no usermod (usermods/ folder and subdirs) references these functions.

Keep your answer short, and only list any still in use function you might find.

softhack007:
Flash Size

| board | before | after | delta |
| ------ | ------- | -------| ------|
| esp32dev | 1295769 | 1295657 | -112 |
| nodemcuv2_160 | 910111 | 909731 | -380 |

softhack007:
@coderabbitai please repeat your analysis in https://github.com/wled/WLED/pull/5368#issuecomment-3884260183 for the latest code in this PR, and provide an updated list "Globally visible functions NOT in fcn_declare.h". Exclude functions that are referenced  in more than one source code file.

softhack007:
Flash Size - update

| board | before | after | delta |
| ------ | ------- | -------| ------|
| esp32dev | 1295769 | 1295609 | -160 |
| nodemcuv2_160 | 910111 | 909619 | -492 |

softhack007:
@coderabbitAI please add the following to you regular review checks:

When "C-style" functions or variables are added or modified, check if they are visible globally. If the function is not part of any global header file, suggest to: 
* either make the function "static" (visible only in one source code file) - possibly with a forward declaration,
* or add a comment to clarify that the function is intended to be a global function.

netmindz:
> @netmindz is there a way to trigger a "manual" run of the usermods CI job on this PR branch? Just to be sure that no usermod breaks due to my changes.

The normal build includes the usermods env that triggers every usermod with an esp32 so that should hopefully catch you changes as they don't relate to the compatibility of specific chips

Issue #5288, opened 2026-01-11
LEDs do not turn off when TV is off (ESP32 always powered)


Evan7121:
Hi,

I have a Wled in my esp32 and Hyperion in my tv.

When I turn off my TV, the LEDs do not turn off automatically.
The ESP32 stays powered because it is connected to a wall power outlet.
The LEDs are powered by a Mean Well 5V 20A power supply.
UDP does not automatically disable itself when Hyperion stops sending data.

Is there a recommended way to automatically turn off the LEDs when the TV is off ?

Setup:
- ESP32 powered from wall outlet (always ON)
- LEDs powered by Mean Well 5V 20A
- Common GND
- WLED version: 0.15.3
- LED type: SK6812 RGBW

softhack007:
The root cause of the problem could be the way that the  ``"live": true`` json command is implemented:  When entering realtime mode by json API command, the timeout is "infinite" so the controling device has to explicitly command "end realtime".

https://github.com/wled/WLED/blob/f830ea498cbb4057de4d436c72a5d2aab49e8e50/wled00/json.cpp#L451-L459

``65000`` is a magic number that means "infinite".
Actually it might be better to replace this with ``realtimetimeout``, possibly guarded by a user option.

The rabbit has done some analysis (https://github.com/wled/WLED/pull/5370#issuecomment-3902019155) of protocols we support, and the result was: 
* no protocol demands an infinite stay in realtime mode
* using the json API is not necessary any more before receiving live leds -- if you select "Receive UDP realtime" in sync settings, then WLED will automatically switch to realtime mode when data arrives, and fall back to local effects when the stream ends.
* "infinite timeout" was an implementation decision but nobody knows any more why it was implemented that way.

netmindz:
Do we know if Hyperion send any json calls or just DDP?

CodingGhost:
I have the same problem. I guess removing the infinite timeout and using the configured timeout would fix it.

DedeHai:
> I guess removing the infinite timeout and using the configured timeout would fix it.

the problem (as so often) with that is: will it break something else that does rely on it being infinite.

CodingGhost:
one possible way to rule that out would be to make it configurable and set the default to infinite maybe?

blazoncek:
> make it configurable

There is already an option for that albeti hidden in JAON API.

[...]

PR #5370, opened 2026-02-13
Allow use of presets (physical button) when receiving live data

IsaccoBenedetti:
Hi, first time contributor here!

Recently I ran into a small issue when using physical buttons and HyperHDR. I'm not sure if this was done by design, but currently when WLED is receiving a live data stream physical button inputs are not processed. This makes it impossible to regain local control over the lights once a live source is active, without opening the webUI.

Initially I tried assigning a preset with the API command LO=2 (Live Override) to a physical button, but it didn't work as expected, because the functions that handle preset logic (handlePresets() and handlePlaylist()) are located within the conditional block that is skipped when realtimeMode is true.

This issue has been previously reported by the community on the WLED Discourse forums, but it seems it never got resolved: [Live data override on the physical button](https://www.google.com/url?sa=E&q=https%3A%2F%2Fwled.discourse.group%2Ft%2Flive-data-override-on-the-physical-button%2F12262).

This PR aims to solve the issue, simply by moving the handlePresets() and handlePlaylist() function calls out of the if (!realtimeMode || ...) block in the main loop.

This way button presses and their associated API commands are now processed on every loop cycle, regardless of the live data state. When a button triggers a preset containing LO=2, the realtimeOverride flag is correctly set, allowing WLED to exit the live stream and apply the desired preset.

I was a bit worried these functions were purposely been left inside the check for performance reasons, so I tested it specifically on an ESP8266 instead of an ESP32 with a HyperHDR stream, but I couldn't notice any lag or jitter at all. It seems solid.

What do you think?

## Summary by CodeRabbit

* **New Features**
  * Added a Sync setting to allow presets to be applied while realtime data is being received.
  * When enabled, playlists and presets can continue processing during realtime mode.
  * The setting is available in the web interface and is saved with device configuration.


softhack007:
Hi, thanks for your contribution to wled development 👍 

I think this PR is a bit mis-labeled ... in fact it does not modify the button handling code. Buttons are read and button can be used as always while WLED is in "external control" mode. The PR does address a weakness that presets controlled by button are not effective until "real time mode" has ended.

The PR actually changes the preset and playlist logic, by allowing preset changes and playlist activity during realtime mode. That's a behavioural change and I'm not sure we should go that way.

Which the proposed change, the order in the main loop gets changed and presets are now handled prior to OTA, HUE and NightLight. To protect previous behaviour, there should be at new option in "sync interfaces" settings, something like "presets and playlist always active" (maybe find a better wording).

I think the root cause of the problem that @IsaccoBenedetti wants to address is somewhere else: 
When entering realtime mode, in some protocols the timeout is "infinite" so the controling device has to explicitly command "end realtime".

https://github.com/wled/WLED/blob/f830ea498cbb4057de4d436c72a5d2aab49e8e50/wled00/json.cpp#L451-L459

``65000`` is a magic number that means "infinite".
Actually it might be better to replace this with ``realtimetimeout``, possibly guarded by a user option.

softhack007:
PS the summary from @coderabbitai is rubbish, it completely missed the point 😄

softhack007:
> Would you like me to regenerate the summary with the proper focus on the functional/behavioral implications rather than just the code mechanics?

@coderabbitai yes. And also add that consequences of the behavioural change still need investigation.

softhack007:
> button presets with LO=2

@coderabbitai it's not clear to me what "LO=2" is actually meant for. please inspects utility functions like leds.cpp, set.cpp, cfg.cpp, XML.cpp, json.cpp to find out what this flag is doing.

wled00/wled.cpp
```diff
@@ -119,14 +127,7 @@ void WLED::loop()
     handleHue();
     yield();
     #endif
-
-    if (!presetNeedsSaving()) {
-      handlePlaylist();
-      yield();
-    }
-    handlePresets();
```
softhack007, line 149: Another option would be to create an `else` branch for ``if (!realtimeMode ....``, and put ``handlepresets()`` into the else branch. But I would still prefer this to be guarded by a user-controled settings.

DedeHai:
another option would be to add a checkmark "exit real time mode on button press" in the config and do just that, any button press will exit?

softhack007:
> another option would be to add a checkmark "exit real time mode on button press" in the config and do just that, any button press will exit?

I like your idea 👍 

Edit: it's still a good chance to also  address the "live mode has infinite timeout" problem - the behaviour is different from all other external control modes, which all use ``realtimetimeout`` instead of infinity.

DedeHai:
which one is "live mode"? I only ever used DDP and that one times out after a few seconds.

softhack007:
> which one is "live mode"? I only ever used DDP and that one times out after a few seconds.

Not sure ... its triggered by a json message with ``"live": true`` so it's not directly DDP or art-net.

Some external tools (hyperion? falcon player?) appear to use this command, but then don't send ``"live": false`` when you simply switch off your control box -> wled stuck.

softhack007:
@coderabbitai  please research the internet on protocol specifications related to the "live mode" or "realtime control" supported by WLED.

Which protocols are started by a json message ``"live": true`` (directly or via MQTT) ? 

Do these pprotocols mandate an "infinite timeout" until ``"live": false``, or would a user-defined timeout of a few seconds also satisfy the needs of the protocol?

softhack007:
@IsaccoBenedetti maybe the latest discussion / research is a bit off-topic.

Can you explain your use case of the button solution?

1) wled stuck _after_ live streaming?

2) you want to abort live streaming while leds are still sent from an external app ? I.e. _during_ live streaming?

3) I'm not sure which external app you use - could you also stream leds without explicitly switching WLED to "live forever" via JSON API? (wled is automatically switching to live mode when a UDP stream arrives, and uses a short timeout when the stream ends)

IsaccoBenedetti:
Hi @softhack007, and thank you for your review!

You're right, maybe I should explain my use case better. Basically, it's one of those classic "wife approval" factors: this wled instance drives a led strip in our living room, which I connected both to a HyperHDR instance (which is this fork of Hyperion: https://github.com/awawa-dev/HyperHDR) and to 2 physical buttons on the wall.
The idea is that the led strip should be used both for normal lighting while also be part of an ambilight system when the tv is on and HyperHDR is transmitting live data (so your #2 scenario, basically).
The issue arised when my GF tried to set the led strip to the "warm white" preset while we were watching a film and told me "wait, why doesn't the switch work anymore?"
I kind of agree with her, because it's my opinion that if there's a physical button and I press it, it should take over anything else, even if HyperHDR is still sending data.
So my idea was ok, I'll add LO=2 (or {"lor":2}, it's the same) to the preset, to disable live override with it (basically the same thing as clicking on "Override Until Reboot" in Wled UI) and create another preset to re-enable it, this way I'll be able to switch on and off the ambilight without having to turn off HyperHDR from the app.
Thing is, because preset logic is guarded by that "if (!realtimeMode || realtimeOverride || (realtimeMode && useMainSegmentOnly))", this wouldn't work.
Adjusting the realtimeTimeout wouldn't solve this. Since the movie is still playing, the stream is active and sending packets every few milliseconds. I don't want to wait for a timeout, I want an immediate manual override via physical button.

To be honest, I feel that since api command LO=2 precise purpose is exactly to break out of realtime mode, any user that adds such a command to a preset will probably expect it to work this way, but I also agree that, since the order of the actions in the main loop is slightly changed now, it may introduce some unexpected consequences, though I haven't noticed any so far. Thus, adding a setting in sync interfaces is probably safer. I can look into it if you'd like, do you have any pointers on how you'd do it, since I'm not familiar with the codebase like you?

Let me know what your thoughts are

netmindz:
See also https://github.com/wled/WLED/issues/5288 @softhack007

IsaccoBenedetti:
Hi, small update, I added the option in sync interfaces as suggested, and also refactored the main loop to avoid changing the order of actions.
I tested it on the same ESP8266 and it seems to be working as expected.
Let me know if you have further suggestions

IsaccoBenedetti:
Hi @softhack007, just a quick ping on this.

I've been running the new version (with the opt-in setting) on my setup without issues so far. Let me know if there is anything else you'd like me to change or if you need further tests before the merge. I'd be happy to help if I can.

PS: I saw your reference to #5288. While it may seem related, I don't think it actually is: that issue is related to how the JSON API handles automatic timeouts when the stream ends, whereas my PR focuses on manual override while the stream is still active.

Thanks!

IsaccoBenedetti:
Hi, I had a look at PR #5573 and noticed this very useful suggestion from @smitty078 : https://github.com/wled/WLED/pull/5573#issuecomment-4425214430

I decided to implement it the way he suggested and I can confirm it's working as expected, without any need to interfere with the main loop.

Thanks @smitty078 !

I kept the option in the sync interface section for now, but I feel at this point it should be safe to remove it completely.

wled00/data/settings_sync.htm
```diff
@@ -156,7 +156,8 @@ <h3>Instance List</h3>
 <h3>Realtime</h3>
 Receive UDP realtime: <input type="checkbox" name="RD"><br>
 Use main segment only: <input type="checkbox" name="MO"><br>
-Respect LED Maps: <input type="checkbox" name="RLM"><br><br>
+Respect LED Maps: <input type="checkbox" name="RLM"><br>
+Immediately force presets switching during realtime mode: <input type="checkbox" name="ROP"><br><br>
```
DedeHai, line 160: rename to something shorter. like "Allow presets during realtime:"

wled00/button.cpp
```diff
@@ -365,6 +365,16 @@ void handleIO()
 {
   handleButton();
 
+  // If in realtime mode and presets are explicitly allowed to override (e.g., via buttons),
+  // we process them here to avoid messing with the main loop.
+  if (realtimeMode && !realtimeOverride && !(realtimeMode && useMainSegmentOnly) && realtimeAllowPresets) {
```
DedeHai, line 370: this should not be here but in main loop

DedeHai (changes requested):

DedeHai:
as I just mentioned in #5850 using "main segment only" should do the same thing as this PR.

DedeHai:
@coderabbitai the code added in handleIO() which runs before the guarded presethandling in the main loop will allow presets to always run correct? what does the combination of the two blocks in main loop and handleIO() mean if it is consolidated into a single if statement?

DedeHai:
@coderabbitai what would be the best consolidated solution for the main loop?

wled00/button.cpp
IsaccoBenedetti: Ok... it's how it was done in a previous commit ([009dca7](https://github.com/wled/WLED/pull/5370/commits/009dca704f5d03a6237f9e083bc609451b083b3b)). I moved it to button.cpp because it was my understanding that you guys preferred not to mess with the main loop.
Reverted it back in commit [68e6575](https://github.com/wled/WLED/pull/5370/commits/68e65759d310ff830a8d1db60cb7fb619b340441)

wled00/data/settings_sync.htm
IsaccoBenedetti: Done in commit [68e6575](https://github.com/wled/WLED/pull/5370/commits/68e65759d310ff830a8d1db60cb7fb619b340441)

IsaccoBenedetti:
> as I just mentioned in #5850 using "main segment only" should do the same thing as this PR.

Hi @DedeHai, thank you for your suggestion. I just tried this, and while I get why it may seem to achieve the same result, I don't think it actually does: some presets (the ones using "frz":false) will get garbled: all pixels will be continually switching between displaying the live data and the "normal" preset at the same time.

I asked an LLM to analyse the code and explain what's happening exactly. This was the response:

> - In MO mode the live pixels are written into the main segment's own buffer
  (FX_fcn.cpp, setRealtimePixelColor() with useMainSegmentOnly), while show() rebuilds the frame
  from all segments, so other segments keep animating.
> - When live data starts, MO freezes the main segment so its effect cannot repaint the live picture.
  Applying a preset re-asserts that freeze and at the same time throws away any requested override
  (json.cpp / set.cpp: "freeze = !realtimeOverride; realtimeOverride = NONE"), so LO/lor cannot be
  used in this mode.
> - However, a preset saved from the UI also carries the segment's own freeze flag ("frz"), and
  deserializeSegment() writes it after the MO branch (json.cpp). UI-saved presets have
  "frz": false, so the segment is unfrozen again and the preset's effect keeps drawing into the same
  buffer the live stream draws into - the LEDs then show both at once, changing from frame to frame
  depending on which wrote last. With "frz": true the effect would be invisible instead (only
  brightness/state changes would show).

As I understand it, this workaround also won't work in any multi-section setups, so I'd argue this isn't an equivalent fix

wled00/button.cpp
softhack007: I agree with @DedeHai - things that the main loop handles should go into the main loop. In this case its better to make it explicit in the main loop, instead of "bolting it on" in the button handler.

DedeHai:
Its not equivalent but achieves the same goal differently. Yes it works on multiple segments, just differently. 

Edit: there is only one difference really: calling service()

IsaccoBenedetti:
> Its not equivalent but achieves the same goal differently. Yes it works on multiple segments, just differently.
> Edit: there is only one difference really: calling service()

Unfortunately it doesn't seem to work, at least not for me... Here's how to reproduce:
- turn on "Use main segment only"
- use HyperHDR with effect "Blue Mood Blobs" for example (or something similar), to initiate live data streaming. Keep this on at all time. The strip displays the effect as expected.
- now turn on the following preset for example:
`{"on":true,"lor":0,"transition":7,"mainseg":0,"seg":[{"id":0,"start":0,"stop":118,"grp":1,"spc":0,"of":0,"on":true,"frz":false,"bri":255,"cct":127,"set":0,"col":[[140,0,255,0],[0,0,0,0],[0,0,0,0]],"fx":0,"sx":119,"ix":128,"pal":35,"c1":128,"c2":128,"c3":16,"sel":true,"rev":false,"mi":false,"o1":false,"o2":false,"o3":false,"si":0,"m12":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0}]}`
All good in this case: the strip displays a solid purple. In this case, the workaround does indeed achieve the same result.
- However, now try this other preset instead:
`{"on":true,"lor":0,"transition":7,"mainseg":0,"seg":[{"id":0,"start":0,"stop":118,"grp":1,"spc":0,"of":0,"on":true,"frz":false,"bri":255,"cct":127,"set":0,"n":"","col":[[140,0,255,0],[0,0,0,0],[0,0,0,0]],"fx":79,"sx":2,"ix":255,"pal":35,"c1":128,"c2":128,"c3":16,"sel":true,"rev":false,"mi":false,"o1":false,"o2":false,"o3":false,"si":0,"m12":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0}]}`
In this case, you should get a mix of the two effects and constant flickering.

You can even see this from the preview. This is the HyperHDR effect alone:
image

This is the preset alone:
image

and this is the result when I activate "use main segment only". As you can see, it's a mix of both:
image

So it seems that the function service() you're mentioning is writing in the same frame buffer the live data function is writing, immediately after. But they're both writing, so as long as service() redraws ALL pixels every cycle this works, but if you use an effect that leaves some pixels off, those pixels get the color that the live data is telling them to.

This of course doesn't happen if instead I turn on the setting "Allow presets during realtime" that this PR aims to add.

Hopefully this clarifies my point.

DedeHai:
> This of course doesn't happen if instead I turn on the setting "Allow presets during realtime" that this PR aims to add.

what happens instead with these presets using the PR code?

IsaccoBenedetti:
> what happens instead with these presets using the PR code?

From a user perspective, it behaves exactly like before (with "main segment only" off), with one single difference: if the user adds lor:0, lor:1 or lor:2 to any preset, that instruction will get executed correctly (turning live override on or off), instead of being ignored.
This is why I initially didn't add the option to turn it on or off in the settings: it could even be argued that this isn't even a new feature, it's a bugfix. The feature is the "lor" command itself and was already there, but without this PR it's broken when used in presets.

From an execution perspective, differently from when "main segment only" is on, there's never 2 sources writing at the same time in the frame buffer: if lor:0 you get the live data, if lor:1 or 2 you get the preset effect. And the user can choose when to turn live data on or off in each preset.

I feel like maybe I didn't explain myself well and this PR is getting a bit confusing, so here's how you can reproduce the behaviour and assess the difference:
- flash an esp with my pr code
- create these 3 presets, and connect them to 3 physical buttons. Please notice the presence of lor:2 (meaning turn live data off) in preset 1 and 3, and the presence of lor:0 (meaning turn live data on) in preset 2:
Preset 1: `{"on":true,"lor":2,"transition":7,"mainseg":0,"seg":[{"id":0,"start":0,"stop":118,"grp":1,"spc":0,"of":0,"on":true,"frz":false,"bri":255,"cct":127,"set":0,"n":"","col":[[140,0,255,0],[0,0,0,0],[0,0,0,0]],"fx":79,"sx":2,"ix":255,"pal":35,"c1":128,"c2":128,"c3":16,"sel":true,"rev":false,"mi":false,"o1":false,"o2":false,"o3":false,"si":0,"m12":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0}]}`
Preset 2: `{"lor":0}`
Preset 3: `{"on":true,"lor":2,"transition":7,"mainseg":0,"seg":[{"id":0,"start":0,"stop":118,"grp":1,"spc":0,"of":0,"on":true,"frz":false,"bri":255,"cct":127,"set":0,"col":[[140,0,255,0],[0,0,0,0],[0,0,0,0]],"fx":0,"sx":119,"ix":128,"pal":35,"c1":128,"c2":128,"c3":16,"sel":true,"rev":false,"mi":false,"o1":false,"o2":false,"o3":false,"si":0,"m12":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0}]}`
- use hyperion/hyperhdr, or similar software to send any effect to the esp. Never turn this source off during the whole test.
- ensure the settings "Use main segment only" and "Allow presets during realtime" are off. Use the buttons to switch to preset 1, then preset2, then 3. You shouldn't be able to. If you manually disable live data from the UI, preset 1 will work. Preset 2 will turn on live data again, and from this moment onward the esp will be stuck displaying live data. There's no way to switch to the other presets using physical buttons.
- now turn the setting "Allow presets during realtime" on. You should now be able to switch between each preset correctly. Preset 2 will display whatever the external source is transmitting, the other two presets will ignore it and display each their own effect.

Let me know if this makes the goal clearer

DedeHai:
the goal is not unclear to me, what is unclear is why we need it in the first place when "use main segment only" does almost exactly this. So my question is: which part of the "use main segment only" is not adequate? The "both live & FX overlapping" part? I think that is actually a bug.

IsaccoBenedetti:
Well, there's also the fact that it confines the live picture to a single segment, but in my case there's only one, so yes, to me the reason it's not adequate is exactly the fact that both live data and FX are overlapping. But if you put it like that, and we assume that's an unintended behavior/bug, then yes, fixing it would make it acceptable for my use case.

However, this feature was introduced in this PR: https://github.com/wled/WLED/pull/2601. And from what I gather, the intended behavior behind the toggle was to display the live data on a single section instead of the whole strip, and the wording of the toggle reflects this. This doesn't seem like it was intended to be used in this context.

Still, if you prefer, I can have a look and try to make the "use main segment only" toggle behave like this PR does instead of overlapping effects, without adding a new checkbox in the settings. Let me know if I should proceed this way.

PR #5374, opened 2026-02-16
Add initialization logic to switch type inputs for correct boot-up behaviour

KuglicsL:
Added boot-up state initialization logic to switch type inputs.
This ensures switch type inputs can be in any state at boot-up without erroneously calling Off->On state change macros.

The initialization logic debounces the input state, initializes button state based on the filtered value and prints debug information.

The changes were tested on an ESP32.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved button input debouncing and stability through enhanced state handling logic.

### DIFF
--- a/wled00/button.cpp
+++ b/wled00/button.cpp
@@ -123,8 +123,39 @@ bool isButtonPressed(uint8_t b)
 
 void handleSwitch(uint8_t b)
 {
+  // extract the current input state check one; makes code faster and less prone to glitches if input state changes mid-function
+  uint8_t currentSwitchState = isButtonPressed(b);
+
+  // SWITCH type input on-boot state should not matter; the software should determine the initial state before processing state changes
+  switch (buttons[b].waitTime) {  // we are hijacking the waitTime variable to keep track of initialized and uninitialized switches
+    case 0: // reading initial state
+      buttons[b].pressedBefore = currentSwitchState;      // save current switch state
+      buttons[b].pressedTime = millis();                  // save debounce start time
+      ++buttons[b].waitTime;                              // change initialization stage to debouncing
+      return;                                             // will have to debounce later, might as well return now
+      break;                                              // redundant, left as guard 
+    case 1: // debouncing
+      if (buttons[b].pressedBefore == currentSwitchState) { // if state is not changing
+        if (millis() - buttons[b].pressedTime > WLED_DEBOUNCE_THRESHOLD) {  // and debounce time is reached
+          buttons[b].longPressed = currentSwitchState;                                    // set longterm state
+          buttons[b].waitTime = 2;                                                        // set switch as initialized
+          DEBUG_PRINTF_P(PSTR("Switch: State init to %u %u\n"), currentSwitchState, b);   // debug initialization success
+        }
+      }
+      else {    // if state is changing, save state and reset timer
+        buttons[b].pressedBefore = currentSwitchState;
+        buttons[b].pressedTime = millis();
+      }
+      return;
+      break;    // redundant, left as guard 
+    case 2:   // the switch is initialized
+      break;
+    default:  // only jumps here if something else messes with waitTime
+      break;
+  }
+
   // isButtonPressed() handles inverted/noninverted logic
-  if (buttons[b].pressedBefore != isButtonPressed(b)) {
+  if (buttons[b].pressedBefore != currentSwitchState) {
     DEBUG_PRINTF_P(PSTR("Switch: State changed %u\n"), b);
     buttons[b].pressedTime = millis();
     buttons[b].pressedBefore = !buttons[b].pressedBefore; // toggle pressed state


DedeHai:
thanks for contributing. 
IMHO this is overcomplicating it, this would do the same:
```c++
if (buttons[b].waitTime == 0) {
  buttons[b].waitTime++;
  buttons[b].pressedBefore = isButtonPressed(b);;
}
```

KuglicsL:
> thanks for contributing. IMHO this is overcomplicating it, this would do the same:
> 
> ```c++
> if (buttons[b].waitTime == 0) {
>   buttons[b].waitTime++;
>   buttons[b].pressedBefore = isButtonPressed(b);;
> }
> ``` 

This would skip the debouncing.
I think it is necessary, as my solution elimimates any uncertainity about system voltage dV/dt at startup.

My example. The board is supplied from 12V -> SMPS -> MCU 3.3V. It also has optocoupler inputs triggered by >=10V, driven from the 12V supply. If the 12V ramp up is slow, the MCU can start up, read switch input and save it as bootup state before 12V reaches the trigger level for the optocoupler stage.

Having the initial debounce also helps if your input has a large external filter cap, only pulled up by internal pullup - it gives the cap time to be charged by the large MCU pullup resistance.

Also debounce in general is used against noise spikes from bad wiring and unstable supply practices. E.g. WLED hardware installed in cars could suffer the most from the lack of it.

Overall I think the slightly more complicated solution is more foolproof and has minimal impact on the handleSwitch() function's speed and size.

DedeHai:
debounce as the name sais is to "debounce" a switch, it should not be used to compensate for all sorts of bad hardware design. At boot a switch will be stable and can not bounce. if power is unstable at boot, there is a compile flag for such setups.

KuglicsL:
I agree about the bad HW design part, but 99% of WLED users have no knowledge what a complier even is. If they encounter an issue like this, they will just find a workaround or downgrade their project. Increasing stability is always welcome especially in unpredictable end user environments.

Debouncing buttons in all states (even initial) keeps industry best practices in sight and gives button handling the most stability. It has low added size (based on amap: button.cpp 256873 -> 267622) and very low added runtime cycles after initialization is done (maybe two extra comparsions in switch - case).

I'd like if others chipped in with their ideas, other than that I rest my case.

DedeHai:
if you want to have debounce for each state change that is fine with me but it then should be done properly without code duplication.

DedeHai:
startup delay logic aside:
won't this change break users setups that rely on current behaviour? like if I have a switch where on is leds on and off is leds off: if the switch is on and a reboot happens it turns back on, after this PR change it wont. correct?

KuglicsL:
_On the code duplication note: I think the whole button handling could use a makeover to move debouncing from individual handleButton/handleSwitch debouncing to a combined debouncer. The short/long press differentiation then could be handled in their own separate state machine. Right now, the debounce threshold is referenced in three different places, and it is not possible to implement initial state debouncing without duplication._

> won't this change break users setups that rely on current behaviour?

That is correct - I presumed the suggested behaviour should be the default based on the comments in the source code.
I believe the best option would be to introduce a new input type: a switch, but it only activates on edge transitions; the actual logic level does not matter. This would maximalize user options for interfacing with the controller.

We would have two types:
1. Switch (level) - old
2. Switch (edge) - new

If this sounds like a useful feature, I might sift through the source and try implementing it. In the meantime, this PR is a dead end.

DedeHai:
> Right now, the debounce threshold is referenced in three different places, and it is not possible to implement initial state debouncing without duplication.

yea, that is what I was hinting at. piling fixes upon a bad base makes it worse over time (and is probably how it got the way it is). We try to improve things in baby steps. 

> * Switch (level) - old
> * Switch (edge) - new

I am not sure this is the best way but it might be, I am not using toggle switches except for tests, all I can say that to me the current implementation is a bit confusing on when it actually switches a preset. Also PIR sensor uses the same code if I read that right, current implementation is already edge driven(?). You are right that the whole code could use a better, unified approach.

KuglicsL:
Current implementation based on my understanding:
1. Initializes switches to default false value
2. If the actual switch state on boot-up is true, it will call the assigned Off->On preset (macroButton) immediately after WLED_DEBOUNCE_THRESHOLD time, which might interrupt the boot-up playlist if one is configured
3. After that, the switch works normally.

After looking at the source some more, there's a few more weird behaviours in there. If a long strip is currently updating, both analog and digital button refresh time is tied to ANALOG_BTN_READ_CYCLE (250ms). Why not have a dedicated define for that frequency, and why is it this large?

The only difference between a PIR switch and a normal SWITCH is inverted reading and a unique MQTT topic (/motion) for the PIR sensor.

I believe the overall cleanest solution would be to rework the button handling to individually configurable entities, rather than pre-configured input types and a global pullup toggle checkbox.
Example:

```
[COMMON GLOBAL OPTIONS]
Global debounce time.
Global touch threshold.

[COMMON INDIVIDUAL OPTIONS]
Inputs can be either normal or inverted.
Inputs can have their own MQTT topic configured for state change reporting. (with limited length topic names)
Inputs can have internal pulls configured. Possible options are pullup/pulldown/none.
Inputs have a short press/positive edge and a long press/negative edge macro slot.
An input can be a momentary switch, a touch switch, or a toggle switch.

[MOMENTARY SWITCH GLOBAL OPTIONS]
Global button long press threshold.
Global button double press interval.

[TOGGLE SWITCH ONLY OPTIONS]
Activation mode can be either level sensitive (current behaviour) or edge sensitive (boot-up state does not matter).
```

Obviously this approach introduces a lot of new possibilities for interfacing on the user side. It is totally feasible to make this change backwards compatible, and selectable presets (BUTTON, SWITCH, PIR) would allow users to easily select old button types in the new system.

If you think this is not over the top, I could start working on a large, *backwards compatible* overhaul in my free time.
The possible downsides would be a more cluttered setup page (perhaps move the buttons to a new "Button preferences" page?), more memory usage, and larger config files.

Let me know what you think.

DedeHai:
without checking code: the list you have there should be pretty much how it is implemented right now. or which are the differences (apart from the code being messy)

KuglicsL:
[COMMON GLOBAL OPTIONS]
Global debounce time.  - **Currently not changeable from UI, set at compile time.**
Global touch threshold. - **Same as current operation, but could also be on a button by button basis**

[COMMON INDIVIDUAL OPTIONS]
Inputs can be either normal or inverted. - **Inputs can't be selected as inverted, inversion is coded into the input type itself.** 
Inputs can have their own MQTT topic configured for state change reporting. (with limited length topic names). - **Right now all events on inputs go to the same MQTT channel (_/button_), except PIR**
Inputs can have internal pulls configured. Possible options are pullup/pulldown/none. - **Currently a global option, and it only support pullups, except for the inverted pushbutton which is pulldown**
Inputs have a short press/positive edge and a long press/negative edge macro slot. - **Same as current operation**
An input can be a momentary switch, a touch switch, or a toggle switch. - **Same as current, but choosing the input type does not lock you into the default values assigned to that input in code**

[MOMENTARY SWITCH GLOBAL OPTIONS]
Global button long press threshold. - **Currently not changeable from UI, set at compile time.**
Global button double press interval. - **Currently not changeable from UI, set at compile time.**

[TOGGLE SWITCH INDIVIDUAL OPTIONS]
Activation mode can be either level sensitive (current behaviour) or edge sensitive (boot-up state does not matter). - **No option for this mode right now**

**Example structure:**
```c++
//shared variables between all configurable buttons
class ConfigurableInput {
  int8_t pin;                   // -1 if uninitialized, positive otherwise
  uint8_t type;                 // cfginput type
  bool inverted;                // if true, input logic level is inverted
  String MQTTEventTopic;        // individual configurable MQTT topic for each input
  uint8_t pullDirection;        // none/up/down
  
  uint8_t prevState;            // previous value for debouncing
  unsigned long lastTransition; // last transition time for debouncing
  unsigned long debounceTime;   // debouncing time threshold

  uint8_t presetEventPrimary;   // preset to call on primary event (short press for button, positive edge for simple switch, any edge for toggle switch)
  uint8_t presetEventSecondary; // preset to call on secondary event (long press for button, negative edge for simple switch)
};

class MomentarySwitch: public ConfigurableInput {
  unsigned long longPressThreshold;   // minimum duration for long press
  unsigned long doublePressInterval;  // maximum interval for double press
};

class ToggleSwitch: public ConfigurableInput  {
  bool edgeMode;                // if true, boot-up state does not matter
};

class TouchSwitch: public ConfigurableInput  {
  uint8_t touchThreshold;       // touch button threshold level
};

class AnalogSwitch: public ConfigurableInput  {
  uint16_t turnOffThreshold;    // global brightness mode turnoff value
  uint16_t filteredValue;       // IIR filtered analog value  
  uint8_t filterCoeff;          // 0 - 100 IIR filter coefficient
};
``` 
So basically this would move lots of compile time options into the runtime configurable territory.
By allowing the user to select every property of each individual input, while also offering "quick setup input presets" (AKA the legacy input types), input configuration flexibility increases by a huge margin, and backwards compatibility is maximized.

DedeHai:
I think the general direction is good. Making everything run-time configurable may be overkill. Like debounce time for example, there may be others. 
Weeding out the current structure and making it more clear is good. On the other hand adding complexity by supporting legacy things (which is a must), ESP8266 (no pulldowns) and keeping the UI simpel for the average user may be a challenge. Also all communication protocols must be backwards compatible.
If you think this is doable please feel free to submit a draft.

KuglicsL:
I started working out the button structure, making it as transparent and flexible as possible. I already have some ideas about implementing total backwards compatibility, but it will take quite some time to finish redoing the whole button handling logic while also adding all the configuration options.

Currently, button configuration is done in two places: in LED preferences for input declaration, and Time & Macros for event action setups.

The best way to keep the UI simple is to move the button setup into it's own configuration tab along with the macro/preset setups for buttons (also the IR remote should be here too IMO).

The "Input setup" configuration tab should have a similar complexity to the current input configuration section. However, having an "Advanced button configuration" checkbox or "Simple <-> Advanced" toggle in there would allow advanced users to fine-tune the software to their specific needs while keeping new users out of deep waters.

I will submit a draft once I have a solid foundation planned out (and we have something concrete we can discuss).

DedeHai:
thanks for taking the effort.
since this adds no new functionality and is a refactor, make sure binary size does not increase, or at least no more than a few hundred bytes.
currently the config of buttons is in hardware config and "what they do" is in macros. It is a bit convoluted but I'd need to think about if adding a new tab is the way to go, probably not for a draft. If the config is to be restructured we need to make sure it makes sense as a whole concept and not just start moving things around.

softhack007:
@DedeHai @KuglicsL if you are still planning to re-design input handling, maybe consider that we might even want "keys" that are using a completely different method for input -see https://github.com/wled/WLED/issues/5387#issuecomment-3937748659 for a few examples.

To me it looks like several features are needed
* "input registration" (for example: usermod for reading from a "port expander" aka shift register). 
* The UI (buttons & macros) must be able to assign actions for any kind of registered input
* Input reading
* Input processing and handling

PR #5379, opened 2026-02-18
Add slow transition FX

### LINKED ISSUE #5375: Long Transitions are currently not possible
I have seen users request longer transition times more than once to be able to make slow changing light transitions without the use of very long playlists or external software.

While increasing transition time in general is a non-starter, this could be done for basic transitions using an FX.

The proposal is this:
- speed slider controls the number of minutes for the transition, 0 means instant which is good for testing

all the FX does is display the currently selected palette on the segment. this way it can be used as a mask.

the effect stores three states in segment data:
- initial state
- current state
- target state

if the palette (or the selected color if used) changes, the current state becomes the initial state and the new state becomes the target state. states should update at 10Hz but rendering should take place each frame. 

using palette blending / color blending this allows for very slow brightness (black to white in overlay for example) or color transitions.
could even use another slider to select a "target preset" that is triggered once the time has run out. that is a hack though as FX should probably not be triggering presets. it would be a simple way to generate "very slow playlists" by chaining such slow transition effects without having to use the scheduler.

DedeHai:
A new effect that is purely made for very slow transitions, up to 255 minutes.
It also allows to call a preset after the transition is done by setting intensity slider to the preset number to be applied after the transition has finished.
Using segment layering, this also allows it to be a "mask" although that takes a bit of tinkering with the preset json.

One scenario possible with this effect would be this:
- start with a preset using this FX, say fully black, set the next preset to one with the same FX but that one at full brightness at a warm white with a transition time of 60 minutes
- the LEDs will transition from black to warm white over 60 mintues, at the end it sets another preset "cold white"
- the "cold white" preset transitions to full bright daylight white in another 60 minutes

The FX only transitions palettes (colors if palette 0 or any color derived palette is used), brightness slider transitions normally i.e. fast. If a color/palette is changed mid transition, it starts a new transition towards that color with the current state as a starting point so transitions never jump.

p.s.
I also remvoed the return value from `Segment::loadPalette`()` as the target palette is passed by reference and the return value is never even used anymore.

Fixes #5375

## Summary by CodeRabbit

* **New Features**
  * Added "Slow Transition" LED effect for smooth palette blending with real-time timing, adjustable speed, and optional sweep mode that updates palette entries sequentially.
  * Preserves start/current/end palettes per segment to ensure seamless transitions and automatic reinitialization when targets change.

* **Bug Fixes / Behavior**
  * Improved palette fallback handling for an existing effect to ensure consistent visuals when a primary palette check is disabled.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -10790,8 +10790,6 @@ static const char _data_FX_MODE_PS_SPRINGY[] PROGMEM = "PS Springy@Stiffness,Dam
  * Slow Transition effect
  * Displays the currently selected palette/color with a very slow transition
  * speed slider controls the number of minutes for the transition (0 = 10s)
- * after transition completes, applies preset set by intensity slider (if any)
- * allows for chains of slow transitions just like a playlist
  * by DedeHai
  */
 typedef struct SlowTransitionData {
@@ -10815,7 +10813,6 @@ void mode_slow_transition(void) {
   // (re) init
   if (changed || SEGMENT.call == 0) {
     if (SEGMENT.call == 0) {
-      //data->presetapplied = false;
       data->startPalette = SEGPALETTE;
       data->currentPalette = SEGPALETTE;
       data->endPalette = SEGPALETTE;
@@ -10827,12 +10824,12 @@ void mode_slow_transition(void) {
       *stepsDone = 0; // reset counter
     }
     *startSpeed = SEGMENT.speed;
-    *startTime = strip.now; // set start time
+    *startTime = millis(); // set start time note: intentionally not using strip.now as this is based on real time
   }
 
   uint32_t totalSteps = SEGMENT.check2 ? 16 * 255 : 255;
   uint32_t duration = (SEGMENT.speed == 0) ? 10000 : (uint32_t)SEGMENT.speed * 60000; // 10s if zero (good for testing), otherwise map 1-255 to 1-255 minutes
-  uint32_t elapsed = strip.now - *startTime; // note: will overflow after ~50 days if just left alone (edge case unhandled)
+  uint32_t elapsed = millis() - *startTime; // note: will overflow after ~50 days if just left alone (edge case unhandled)
   uint32_t expectedSteps = (uint64_t)elapsed * totalSteps / duration;
   expectedSteps = min(expectedSteps, totalSteps); // limit to total steps
 
@@ -10854,12 +10851,8 @@ void mode_slow_transition(void) {
       }
     }
     if (*stepsDone >= totalSteps) {
-      // transition complete, apply preset set by intensity slider (if any)
+      // transition complete, apply end palette
       data->currentPalette = data->endPalette; // set to end palette (sweep may not have set all entries)
-      if (SEGMENT.intensity > 0) {
-        uint8_t targetPreset = SEGMENT.intensity;
-        applyPreset(targetPreset, CALL_MODE_DIRECT_CHANGE); // apply preset (if it exists)
-      }
     }
   }
   // display current palette over segment
@@ -10869,7 +10862,7 @@ void mode_slow_transition(void) {
     SEGMENT.setPixelColor(i, palcol.color32);
   }
 }
-static const char _data_FX_MODE_SLOW_TRANSITION[] PROGMEM = "Slow Transition@Time (min), End reset,,,,,Sweep;3;!;0;pal=2,sx=0,ix=0";
+static const char _data_FX_MODE_SLOW_TRANSITION[] PROGMEM = "Slow Transition@Time (min),,,,,,Sweep;3;!;0;pal=2,sx=0,ix=0";
 
 //////////////////////////////////////////////////////////////////////////////////////////
 // mode data


netmindz:
Not tried it, but sounds exactly what one of my customers wanted from my lamps

DedeHai:
it also was requested _somewhere_ before, I saw it more than once but could not find the references. 
Its not the perfect solution and the "call preset after fade" is a bit of a hack or at least not strictly future proof as an FX basically has no business calling presets but I am not aware of anything breaking if it does it. 
this probably needs some more testing on systems that use mqtt or sync or other means of external communication. I tested only on a confined device and there it works quite well.

In a nutshell this is the "sunrise FX" but instead of a fixed sunrise pattern, it allows to use any palette/color as a starting point and as a destination. I am not sure it is very easy to use though: if you just call it once it does nothing but initialize, so you have to use it either in a playlist or in a "chain of presets"

wled00/FX.cpp
```diff
@@ -10787,6 +10786,91 @@ static const char _data_FX_MODE_PS_SPRINGY[] PROGMEM = "PS Springy@Stiffness,Dam
...
+      for (uint8_t i = 0; i < 16; i++) {
+        data->currentPalette[i] = CRGB(color_blend(CRGBW(data->startPalette[i]), CRGBW(data->endPalette[i]), blendAmount));
+      }
+    }
+    if (*stepsDone >= totalSteps) {
+      // transition complete, apply preset set by intensity slider (if any)
+      data->currentPalette = data->endPalette; // set to end palette (sweep may not have set all entries)
+      if (SEGMENT.intensity > 0) {
+        uint8_t targetPreset = SEGMENT.intensity;
+        applyPreset(targetPreset, CALL_MODE_DIRECT_CHANGE); // apply preset (if it exists)
```
DedeHai: @willmmiles what do you think of this line? i.e. applying a preset from an FX? this is the only thing I find questionable about this PR but it works. If you see no bigger issues with this, I will fix it up and merge.
if there is potential issues in the future, we should do it differently: as soon as this one is out, people will use it and it will need to be maintained - or broken.

willmmiles: Technically it is safe in this context.  Contrary to the name, `applyPreset()` doesn't actually apply the preset, it just schedules it to be applied on the next pass through the main loop.  There's no queue, though, so it may interact strangely if there's a button press or http call in flight at the same time.  (That's not new though.)

Architecturally, I don't love the interface design -- adding another system for time-based preset firing in parallel with playlists and timer slots feels like we're asking for subsystems to trip over each other.  I also recognize that there's value in ensuring synchronicity between overlay effects and play state (ie. without having to make sure that the playlist entry duration exactly matches the effect ramp time).

Although now that I've typed it out, it occurs to me that maybe having an effect that reads the remaining playlist step duration might solve this problem just as well -- and then the "next preset to select" can be left to the playlist logic, with all the expanded power that has (random preset selection, etc.).   What do you think?

DedeHai: the point of this PR is to enable VERY slow transitions, playlist have a 999s "per entry" time limit where as this PR aims for 255min. It suffers from the same discrepancies noted in the copy FX of an FX stepping over its (architectural) boundaries, which I also do not like.
While the copy FX could be moved to the segment, the slow transition can't, at least I see no simple way to do it.

willmmiles: > the point of this PR is to enable VERY slow transitions, playlist have a 999s "per entry" time limit

I think the documentation might be out of date on that - as far as I can tell from the playlist code, the current limitation is 6553 seconds (109 minutes, 1.82 hours).   I'd support removing that limitation - using a 32-bit integer for playlist times isn't a huge memory cost.

> It suffers from the same discrepancies noted in the copy FX of an FX stepping over its (architectural) boundaries, which I also do not like.
> While the copy FX could be moved to the segment, the slow transition can't, at least I see no simple way to do it.

I am actually a fan of the copy FX implementation!  I agree it's a different "kind" of effect, and I would be entirely on board with putting it in a different category in the UI, but I think the ability of our engine to support meta-FX like that is a sign of a good design.  I like the idea of a fully composable rendering pipeline.  (So many ideas, so little time...)

If we lifted the playlist time limitation, do you think it would be better use "playlist progress" instead of directly applying the next effect?

DedeHai: > If we lifted the playlist time limitation, do you think it would be better use "playlist progress" instead of directly applying the next effect?

Absoluteley. The playlist has another limit applied, the 999s is real. Will check where that is applied and see if it can be lifted. If playlists allow for longer durations, there is no need for this FX to be able to trigger presets.
edit: turns out if you enter 9999 it truncates to 999 and not the max allowed value 6553

> I am actually a fan of the copy FX implementation! I agree it's a different "kind" of effect, and I would be entirely on board with putting it in a different category in the UI, but I think the ability of our engine to support meta-FX like that is a sign of a good design. I like the idea of a fully composable rendering pipeline. (So many ideas, so little time...)

I also like the concept of the copy FX, its a very useful tool to have and I would not want to remove it. Just the fact that it accesses other segment's buffers might become an issue at one point.

DedeHai:
@coderabbitai please trace through the code from UI to preset handling and check where limits to playlist item duration are applied. Check if removing the restriction and making the duration a uint32 is possible or if there are potential side-effects. Provide a suggestion on how to implement longer playlist item duration.

DedeHai:
@coderabbitai in playlists there is also a transition between entries. explain how the transition is applied and what time limits are set and how they are restricted.

wled00/FX.cpp
```diff
@@ -10787,6 +10786,84 @@ static const char _data_FX_MODE_PS_SPRINGY[] PROGMEM = "PS Springy@Stiffness,Dam
...
+    }
+  }
+  // display current palette over segment
+  for (unsigned i = 0; i < SEGLEN; i++) {
+    uint8_t paletteIndex = (i * 255) / SEGLEN;
+    CRGBW palcol = ColorFromPaletteWLED(data->currentPalette, paletteIndex, 255, LINEARBLEND_NOWRAP);
+    SEGMENT.setPixelColor(i, palcol.color32);
+  }
+}
+static const char _data_FX_MODE_SLOW_TRANSITION[] PROGMEM = "Slow Transition@Time (min),,,,,,Sweep;3;!;0;pal=2,sx=0,ix=0";
```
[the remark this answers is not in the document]
DedeHai: this is intentional and left to the user to discover

wled00/FX.cpp
```diff
@@ -10787,6 +10786,91 @@ static const char _data_FX_MODE_PS_SPRINGY[] PROGMEM = "PS Springy@Stiffness,Dam
...
+    *stepsDone = expectedSteps;// in case sweep was disabled mid transition
+
+  if (*stepsDone < expectedSteps) {
+    *stepsDone = expectedSteps; // jump to expected steps to make sure timing is correct (need up to 4080 frames, at 20fps that is ~200 seconds)
+    if (SEGMENT.check2) {
+        // sweep: one palette entry at a time
+        uint8_t i = *stepsDone % 16;
+        uint8_t blendAmount  = *stepsDone / 16;
+        data->currentPalette[i] = CRGB(color_blend(CRGBW(data->startPalette[i]), CRGBW(data->endPalette[i]), blendAmount));
+    } else {
```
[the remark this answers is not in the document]
DedeHai: please elaborate on how this manifests and suggest a procedure to test when this becomes an issue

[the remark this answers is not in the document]
DedeHai: so this is an issue when setting to "test the transition" i.e. speed=0 only? I have noticed it and its a non issue as this PR is about long transitions i.e. longer than 1 minute because 1 minute is already doable by normal transitions.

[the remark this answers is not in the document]
DedeHai: If this is not an issue if the transition is >1min I do not see the point, my code handles it as intended and "jumps" are really minor and intended to handle mid-transition enabling/disabling of the sweep option

PR #5389, opened 2026-02-20
Implement lastPreset tracking for auto-save usermod

ampledashes:
Added lastPreset variable to track the last preset used for auto-saving functionality so that the last preset active will be called upon reboot if the user has those settings configured

### DIFF
--- a/usermods/usermod_v2_auto_save/usermod_v2_auto_save.cpp
+++ b/usermods/usermod_v2_auto_save/usermod_v2_auto_save.cpp
@@ -63,6 +63,7 @@ class AutoSaveUsermod : public Usermod {
     uint8_t knownEffectIntensity = 0;
     uint8_t knownMode = 0;
     uint8_t knownPalette = 0;
+    uint8_t lastPreset = 0;
 
     #ifdef USERMOD_FOUR_LINE_DISPLAY
     FourLineDisplayUsermod* display;
@@ -117,6 +118,7 @@ class AutoSaveUsermod : public Usermod {
       knownEffectIntensity = effectIntensity;
       knownMode = strip.getMainSegment().mode;
       knownPalette = strip.getMainSegment().palette;
+      lastPreset = currentPreset;
     }
 
     // gets called every time WiFi is (re-)connected. Initialize own network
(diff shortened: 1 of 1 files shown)


softhack007:
@ampledashes I'm still struggling to understand the use case for this PR. Can you explain a bit more in which scenarios this will be helpful?

The usermod already tracks the current effect, palette, effect sliders, colors, brightness. Afaik the purpose is not to create a copy of the last active preset, but to help users who tinker with effects and want the same effect to be active on next start.

> the last preset active will be called upon reboot

This is exactly something that will not happen: If you activate preset 37, then switch WLED off after a minute, WLED will not re-activate preset 37. It will boot into the autosave preset, let's say preset 250.

Maybe I missed something?

ampledashes:
The specific gap we’re addressing is that recalling a preset and then powering off before making any value changes results in nothing being saved — the autosave never triggered because no tracked values changed. On reboot, the device loads stale autosave state rather than the recently recalled preset’s state. This PR ensures preset recalls also schedule a save, so the autosave snapshot stays current.

DedeHai:
```
// By default it will not save the state if an unmodified preset
// is selected (to not duplicate it). You can change this behaviour
// by setting autoSaveIgnorePresets=false
```
does that not work?

softhack007:
> recalling a preset and then powering off before making any value changes results in nothing being saved

Ah OK, so technically the issue is that this UM is not multi-segment aware. It tracks changes to the global "currenteffect", "currentPalette" etc. and reacts on changes in the main segment, but changing the effect in the second, third, etc segment will not be recognised. 

The same happens when changing to a preset that doesn't  modify the first segment, but only affects other segments.

https://github.com/ampledashes/WLED/blob/5d1e8e921f5cef4f2857b3ac2291b029719b6ca0/usermods/usermod_v2_auto_save/usermod_v2_auto_save.cpp#L116-L121

Adding "currentPreset" to the watched state information will be an improvement, but it's still not a solution when users change the effect in the second segment without using a preset.

Plus it will add flash writes (wearing out flash) that are just copying the current preset state into the dedicated autosave preset.

softhack007:
1) I think we need a new UM settings checkbox, like "autosave when activating existing presets".
2) do we also want to update the autosave preset slot when  any kind of JSON preset - for example a button or timer controlled preset with  ``LO=2`` to end realtime mode?
3) what happens when I run a playlist that changes to another preset every 10 seconds? Should autosave update on each playlist step?

softhack007:
@coderabbitai please re-review this PR.
Summarize good practice, changes requested by reviewers, and improvements (prioritized).

softhack007 (changes requested):
@ampledashes please address/ clarify the points raised by reviewers.

Summary is [here](https://github.com/wled/WLED/pull/5389#issuecomment-4215020213)

Issue #5391, opened 2026-02-21
Conflict between Ethernet and mic settings on main


kilrah:
### What happened?

Tested with iotorero ethernet controller and dig-uno with ethernet on current main, and all the way back to October.
Sometimes enabling mic kills ethernet, particularly PDM.

### To Reproduce Bug

Erase flash, flash esp32dev_eth
Enable ethernet, power cycle, works
Enable I2S PDM mic, select the pins (35/15/-1/-1 for the iotorero), power cycle, AR works but the ethernet ACT led blinks for a couple of seconds after boot and then freezes and Ethernet stops responding, only Wi-Fi.

I tried on the quinled esp and same with PDM and 2 pins. I selected I2S and 3 pins because that's the config I have on this unit and the problem wasn't present. I however tried to unassign the 3rd pin and then it locked up ethernet again.
In case it's related to the 3rd pin I tried assigning a dummy SCK pin in I2S PDM mode but that didn't help.

### Expected Behavior

Ethernet working

### Install Method

Self-Compiled

### What version of WLED?

439d4a0e98c947b1778a826c2a5cc3742a976eca

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DedeHai:
@coderabbitai please check the codebase for any explanation of such a behaviour.

DedeHai:
@coderabbitai you assessment is incorrect. this combination works on earlier releases like 0.15.3 so it can not be hardware related unless this was something introduced in IDF V4. Please check again, also check if this is a known bug in IDF V4. Check changes to initialization sequence between 0.15.3 and current code.

softhack007:
> , the audio reactive code was changed to enable **`use_apll = true`** for PDM microphones with the comment "experimental - use aPLL clock source to improve sampling quality". This change occurred **after v0.15.3 was released**, which explains why it worked in that version.

While the timeline suggested by the rabbit is wrong (the was no 0.15.x even in sight in Nov 2022), the APLL incompatibility makes sense, maybe ``use_apll = true`` is even in the idf 4.4.x part only. 

For a quick check, it should be enough to change this line into ``use_apll = false``:

https://github.com/wled/WLED/blob/822e07d9f5c4c3eaa705f696697d172ea7e961bb/usermods/audioreactive/audio_source.h#L228

[...]

Issue #4703, opened 2025-05-30
ESP-NOW must be disabled on ESP32 Ethernet boards that use GPIO17 as clock source


Arcitec:
### What happened?

During the WLED 0.15.1 testing, I found an extreme slowdown of the ESP32 when using ESP-NOW based remote controls to control WLED. My original mention of it (with a video example) was here:

https://github.com/wled/WLED/issues/4701#issuecomment-2917949281

I have now found out why ESP-NOW is such a bad thing on Ethernet boards. It's a hardware bug that was discovered by ESP32 staff in August 2024:

- When WiFi is active, the PLL clock output on GPIO17 becomes unstable.
- Many ethernet boards use GPIO17 as the clock for the Ethernet chip to avoid buying external and possibly broken crystals, to simplify the board design and end up with a very reliable board (or so we thought). I do that. Even Quindor who "never spares any expense" does exactly that. Lots of other boards do it.
- The hardware bug is triggered when ESP-NOW is enabled on an Ethernet board that uses GPIO17 as clock for any purpose whatsoever (usually people tap it for Ethernet). The ESP-NOW protocol uses WiFi. Since WiFi is now enabled, it then throws off timing of all the clock signals for the Ethernet chip. In turn, it creates havoc in the decoding of Ethernet data and the operation of the Ethernet chip. Literally, the Ethernet port becomes unstable and the ESP32 chip itself can get hung for a while as it tries to make sense of the chaos.
- The ESP32 hanging/slowdown is random since it only happens if the ESP-NOW (WiFi) protocol transmits at the exact same time that Ethernet packets are being decoded.

In short, this hardware bug means that Ethernet boards which use the ESP32 as clock source cannot use ESP-NOW (the remote control protocol based on WiFi).

I found this via the following ESP32 hardware errata page, and there's a discussion where it was originally discovered:

- https://docs.espressif.com/projects/esp-chip-errata/en/latest/esp32/03-errata-description/esp32/clock-rmii-clock-when-using-ethernet-and-wifi.html
- https://esp32.com/viewtopic.php?t=37757

## So how does this affect WLED:

- In the past, in WLED 0.14.x and earlier, some WLEE developer had discovered issued but wasn't fully sure what was going wrong, so they added a flag to disable ESP-NOW (WiFi) in all Ethernet builds.
[...]

PR #5394, opened 2026-02-23
Add compile-time WiFi hard-off mode (WLED_FORCE_WIFI_OFF)

Stargazer2026:
### Motivation

- Provide a compile-time "flight mode" option to completely disable the WiFi radio (WIFI_OFF) so builds for Ethernet-only or radio-restricted environments do not start STA or AP interfaces.
- This saves 5kb for the resulting firmware.bin binary.

### Description

- Add a commented compile-time flag `WLED_FORCE_WIFI_OFF` to `wled.h` and document it in `my_config_sample.h` so users can enable it from `my_config.h` or build flags.
- In `wled.cpp`, change startup networking to call `WiFi.mode(WIFI_OFF)` when `WLED_FORCE_WIFI_OFF` is defined instead of starting STA scanning and `findWiFi(true)`.
- Short-circuit `initAP()` to immediately return when `WLED_FORCE_WIFI_OFF` is defined so no softAP is created.
- Make `initConnection()` early-exit when `WLED_FORCE_WIFI_OFF` is defined after calling `WiFi.disconnect(true)` and forcing `WiFi.mode(WIFI_OFF)`, and add debug prints to indicate the forced-off state.

## Summary by CodeRabbit

* **New Features**
  * Optional build-time WiFi radio disable (requires Ethernet); disabled by default.
  * Emergency access-point activation via a configured momentary button remains available.

* **Bug Fixes / Behavior**
  * Startup, scan, and reconnect behavior now respect the optional WiFi-off mode, preventing normal WiFi scans/connections when enabled.

* **Chores**
  * Added build-time validation warnings to highlight incompatible or missing button/ethernet configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wled/WLED/pull/5394?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


wled00/wled.cpp
```diff
@@ -673,6 +682,13 @@ void WLED::initAP(bool resetAP)
 void WLED::initConnection()
 {
   DEBUG_PRINTF_P(PSTR("initConnection() called @ %lus.\n"), millis()/1000);
+#ifdef WLED_FORCE_WIFI_OFF
+  DEBUG_PRINTLN(F("WLED_FORCE_WIFI_OFF active. Skipping WiFi/AP init."));
+  WiFi.disconnect(true);
+  WiFi.mode(WIFI_OFF);
+  return;
+#endif
```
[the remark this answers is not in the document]
Stargazer2026: Good catch, I fixed it.

softhack007:
If we add this PR, then there should be a HUGE warning that "no wifi" means "no wifi connection at all" 

* no way to configure ethernet via WLED-AP
* if ethernet settings is are bad, the only recovery option is to flash wled via USB
* if ethernet stop working (for example bug #5391, or users accidentally configure usermods to use ethernet pins) then there is no easy recovery (see above)
* you can't use the "long button press" method to open an "emergency" access point (WLED-AP)

softhack007:
(joke) maybe rename the build flag to WLED_FORCE_WIFI_OFF_AND_SHOOT_YOURSELF_IN_THE_FOOT 😉 

I'm not saying NO to this PR - there might be special use cases where its helpful.

But I would prefer a less "hard off" approach, maybe by adding a "wifi mode" user setting next to "AP mode", and keep at least "AP mode by button" operational for emergency cases.

DedeHai:
can you please elaborate on the advantages?

Stargazer2026:
Thank you for reviewing this.

The intention of this PR is not to change default behavior. WiFi remains enabled unless explicitly excluded via a compile time flag.

The primary motivation is deterministic Ethernet only deployments. In certain installations, especially fixed architectural or PoE based setups, it is desirable to completely exclude the WiFi stack at build time rather than simply not configuring it.

This approach has several technical advantages:

1. Reduced attack surface, the WiFi stack is not activated at all  
2. Deterministic network behavior, no AP fallback and no unintended RF activity  
3. Slight reduction in firmware size (measured delta: 5 kB)  
4. No runtime configuration complexity  

The decision to implement this as a compile time exclusion instead of a runtime toggle was intentional.

A runtime option would require:

- additional UI configuration  
- persistence handling  
- conditional logic in multiple code paths  
- interaction with AP fallback logic  
- expanded testing surface  

By contrast, this implementation:

- introduces no additional runtime branches  
- does not extend the configuration UI  
- does not increase support surface  
- keeps the code delta small and isolated  
- leaves default builds completely unchanged  

The feature is strictly opt in for specialized builds and does not affect existing users or standard firmware releases.

If needed, I am happy to provide exact firmware size comparisons or adjust naming and placement of the build flag to better match existing conventions.

DedeHai:
The only point I am getting is the "radio restricted area" thing, but that is very niche. All other points can be addressed in a different way and the firmware size reduction is not really helpful unless you pack it full of usermods.
On the other hand this make it really unflexible and if anything in your network changes you need to recompile - so I agree with @softhack007 that the AP fallback should still be there and if it is, the rest does not really make sense.

Stargazer2026:
Thanks for the clarification.

I understand that this is a niche use case. I am not trying to position it as something broadly needed for typical installations.

My main point is that the implementation cost for the project is extremely low:

- 19 lines of code
- isolated behind a compile-time flag
- no changes to default builds
- no runtime branches added
- no UI changes
- no impact on the existing AP fallback logic in standard firmware

This does not introduce additional behavioral complexity in the normal code path. If the flag is not set, the firmware behaves exactly as before.

Regarding flexibility: yes, a recompile is required if the network topology changes. But since this is a compile-time exclusion, it is explicitly targeting fixed-purpose deployments where that trade-off is intentional.

In other words, this is not meant as a user-facing feature but as a specialization hook for constrained or regulated environments.

Given the very small and well-contained delta, I would argue that the long-term maintenance overhead should be close to zero, especially compared to runtime configuration features.

If you still feel this increases project surface unnecessarily, I completely understand. I just wanted to clarify that the implementation was intentionally minimal and non-invasive.

DedeHai:
> In other words, this is not meant as a user-facing feature but as a specialization hook for constrained or regulated environments.

I do understand your intend and I can see its usefulness for some cases, however my concern is this: 
many users do "random stuff without understanding consequences" and if they think this feature is something they want but it is more of a "know what you do" thing there will be complaints and questions on why its not working right. Just one example: it assumes you have DHCP. If you don't it wont even show up in the network without compiling that in too.

edit:
a solution would be to have this flag produce a build error with this big fat warning included and saying that this build error line must be deleted, therefore forcing a user to aknowledge and bear the consequences. It also hinders online compilers from using it, making sure only advanced users get access to this flag.

edit2:
this still wont keep users from posting bins and using those though.

wled00/wled.h
```diff
@@ -29,6 +29,7 @@
 #endif
 
 // You can choose some of these features to disable:
+//#define WLED_FORCE_WIFI_OFF      // saves 5kb, hard-disable WiFi radio entirely (WIFI_OFF). AP/STA will not start.
```
softhack007: Its better to not add the flag here. It will be too tempting for "cool lets try this" users to brick their boards.

For documentation, i'd prefer a PR in [WLED-Docs](https://github.com/wled/WLED-Docs) and describe the new flag as an expert option.

wled00/my_config_sample.h
```diff
@@ -22,5 +22,6 @@
 #define CLIENT_PASS "Your_Password"
 */
 
+//#define WLED_FORCE_WIFI_OFF // Hard disable WiFi radio (WIFI_OFF): no STA and no AP.
```
softhack007: please remove the flag here, too (see my comment on wled.h)

softhack007:
A few use case questions that came to my mind: 

* if you take a freshly wiped board (esptool erase_flash), then install WLED compiled with ``-D WLED_USE_ETHERNET -D WLED_FORCE_WIFI_OFF``, how can people configure ethernet?
  Ethernet type, static/dynamic IP, gateway, DNS server, etc?
* similar scenario, but with a fully configured ethernet: someone presses button #0 for longer than 10 seconds => factory reset => device lost.
* Radio-restricted environments (security?):  ``-D WLED_FORCE_WIFI_OFF`` should also include ``-D WLED_DISABLE_OTA`` and ``-D WLED_DISABLE_ESPNOW``, right? 
  OTA would be the easiest way to re-enable wifi radio if i want to break out of the restricted zone.
* Should we add be a compile time ``#error`` when ``-D WLED_FORCE_WIFI_OFF`` is used _without_ ``-D WLED_USE_ETHERNET`` ?

softhack007:
@Stargazer2026 please don't get me wrong - I'm not trying to corner your PR. This is just an attempt to find out what could be the consequences of having WLED without wifi option. See it as a free "safety analysis", that can be used to improve the safety and availability in non-normal cases.

Actually, this kind of analysis is part of my "day time job", so don't take my questions personally :-)

netmindz:
I can completely understand that outside of the home setting, that to have devices that might run on ethernet 99% of the time, the suddenly swap to AP mode in an office, nightclub etc just because someone rebooted the router is a valid concern, so having the option to select ethernet only makes sense. 

I'm less clear on the use case of when you would use WLED in a setting where you couldn't (legally?) have an active radio and if this change would be robust enough to meet that compliance 100% - it's certainly a feature I wouldn't want the team to take responsibility for ensuring every release never activates the radio at all

softhack007:
> it's certainly a feature I wouldn't want the team to take responsibility for ensuring every release never activates the radio at all

I agree, especially we know that many "non-expert admins" will simply use OTA to refresh their device with the latest software. Like "I trust HomeAssistent to pick the right update for me". After that point, the WLED device will start offering its WLED-AP against all "radio off" assumptions. 

As a counter-measure, removing OTA and ESP-NOW could work. But if hard "radio off" requirements need to be met, it also means to put the board into a hard-to-break enclosure with intruder alert, to prevent updating WLED by UART pins.

Stargazer2026:
First of all, thank you all for the detailed analysis! I genuinely appreciate the “safety review” perspective. These are exactly the right questions to ask :-)

Let me clarify one important point:

This feature is **not** intended as a regulatory compliance guarantee and should not be marketed or understood as “certified radio-off firmware”. It is simply a compile-time exclusion of the WiFi stack for controlled, fixed-purpose builds where the integrator fully owns the deployment.

I completely agree that default WLED behavior must remain flexible, recoverable and safe for normal users.

Given the concerns raised, I would propose tightening the scope further:

- Require `-D WLED_USE_ETHERNET` together with `-D WLED_FORCE_WIFI_OFF`
- Add a compile-time `#error` if `WLED_FORCE_WIFI_OFF` is used without Ethernet
- Add a prominent compile-time warning message explaining the consequences
- Clearly document this as an expert-only option in WLED-Docs
- Explicitly state that OTA, ESP-NOW and recovery expectations change under this flag

Regarding responsibility:  
The integrator who compiles with this flag accepts that recovery requires physical access (UART, rebuild, etc). This is intentional and aligned with the “expert-only” nature of the option.

From an implementation perspective, the delta remains small, isolated and fully opt-in. Default builds and OTA users remain completely unaffected.

If the concern is mainly about inexperienced users experimenting with it, I am fully in favor of making access intentionally inconvenient (for example via forced acknowledgement in code), so it does not become a “cool let’s try this” toggle.

My goal here is not to expand WLED’s general feature surface, but to provide a safe specialization hook for advanced deployments without affecting mainstream users.

Happy to adjust the PR accordingly.

softhack007:
Thanks for your response :-) I think "make access inconvenient" is the way to go 👍 

> My goal here is not to expand WLED’s general feature surface, but to provide a safe specialization hook for advanced deployments without affecting mainstream users.

What if we could have both? If we lower our expectations on "radio off", we could still allow "WLEP-AP by button" for dummy user recovery. The 5Kb firmware shrink is nothing we absolutely need, so keeping the AP code active but enforcing "AP_BEHAVIOR_BUTTON_ONLY" could be an option.

Not saving 5Kb flash is really a small price, compared to a few "stuck users" who need help to get out of their "no Wi-Fi" problems, especially if the "inaccessible" board is sealed in epoxy and fixed to the roof ;-)

Stargazer2026:
Thanks again for the constructive feedback. I’ve reworked the implementation based on the discussion.

## Updated Behavior

When `WLED_FORCE_WIFI_OFF` is enabled, WLED now follows a **soft-off WiFi model**:

- Regular STA startup is disabled
- AP fallback behavior is disabled
- WiFi remains OFF during normal operation
- The long-press emergency AP via button 0 remains available

In other words, WiFi is no longer part of normal network behavior, but intentional recovery via button-triggered AP is still possible.

This addresses the main concern about “stuck” devices while still preventing unintended RF activity in Ethernet-first deployments.

## Compile-Time Safeguards

To make this feature safer and clearly expert-only, I added the following safeguards:

- `#error` if `WLED_FORCE_WIFI_OFF` is used without `WLED_USE_ETHERNET`
- `#warning` if `BTNPIN < 0` (no hardware button configured)
- `#warning` if `BTNTYPE` is not momentary by default

This shifts prerequisite validation to compile time, where it is most useful for this type of feature.

## Scope Clarification

This is not intended to provide regulatory “radio-off compliance”.  
It is an expert build option for controlled Ethernet deployments.

Default firmware behavior remains completely unchanged.

The implementation remains small and isolated:
- No impact on standard builds
- No change to default AP behavior
- Emergency recovery path preserved
- No additional runtime configuration surface

If the naming should better reflect the soft-off semantics instead of “FORCE_WIFI_OFF”, I’m happy to adjust that as well.

wled00/wled.cpp
```diff
@@ -673,6 +679,17 @@ void WLED::initAP(bool resetAP)
...
+#ifdef WLED_FORCE_WIFI_OFF
+  DEBUG_PRINTLN(F("WLED_FORCE_WIFI_OFF active. WiFi stays OFF unless AP emergency mode is opened by button."));
+  apBehavior = AP_BEHAVIOR_BUTTON_ONLY;
+  lastReconnectAttempt = millis();
+  if (!apActive) {
+    WiFi.disconnect(true);
+    WiFi.mode(WIFI_OFF);
+  }
+  return;
+#endif
```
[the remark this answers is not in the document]
Stargazer2026: Good catch, I've hardened this part now as follows:

`wled00/wled.cpp` (`handleConnection()`):
- All WiFi scan/reselect logic (`WiFi.scanComplete()` / `findWiFi(...)`) is now skipped when `WLED_FORCE_WIFI_OFF` is enabled.
- This prevents reconnect loops from triggering WiFi scan paths while still allowing normal interface initialization flow for Ethernet.

And then also added the following:

`wled00/network.cpp` (`WiFiEvent()`):
- On `WIFI_STA_DISCONNECTED` and `ETH_DISCONNECTED`, scan re-init calls are now also skipped under `WLED_FORCE_WIFI_OFF`.
- This ensures event handlers do not restart WiFi scanning in forced-off builds.

Result: in forced-off builds, regular WiFi scan/reconnect activity stays disabled, while the intended non-WiFi flow remains intact.

Please review again.

wled00/wled.h
```diff
@@ -291,6 +292,18 @@ WLED_GLOBAL char otaPass[33] _INIT(DEFAULT_OTA_PASS);
 #ifndef BTNTYPE
   #define BTNTYPE BTN_TYPE_PUSH
 #endif
+
+#ifdef WLED_FORCE_WIFI_OFF
+  #ifndef WLED_USE_ETHERNET
+    #error "WLED_FORCE_WIFI_OFF requires WLED_USE_ETHERNET."
+  #endif
+  #if (BTNPIN < 0)
+    #warning "WLED_FORCE_WIFI_OFF: BTNPIN is disabled. Emergency AP long-press on button 0 will not be available unless configured in cfg.json."
```
DedeHai, line 301: both button defines should be mendatory IMHO

Stargazer2026:
I refactored the implementation now to use less #ifdef blocks. Please have a look at commit d14643bae8d8e257631695145f0de857602245bb and let me know if you prefer this implementation or the former implementation, with more explicit (inline) #ifdef blocks.

wled00/wled.h
Stargazer2026: Thanks for the feedback. How do you feel about a change as outlined below?

It moves the default definitions of BTNPIN and BTNTYPE below the checks for the WLED_FORCE_WIFI_OFF use case. This results in a hard enforcement that the button is explicitly configured by the user in the intended way. It also changes from warning to error in case it's missing.

```c++
diff --git a/wled00/wled.h b/wled00/wled.h
index 207848d2671fda94e14b671d2d61fb30c7e178e8..7e29a4b231df4fe9baeb652fa4ee85c38a432b0f 100644
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -263,69 +263,75 @@ using PSRAMDynamicJsonDocument = BasicJsonDocument<PSRAM_Allocator>;
   #define _INIT(x)
   #define _INIT_N(x)
   #define _INIT_PROGMEM(x)
 #else
   #define WLED_GLOBAL
   #define _INIT(x) = x
   //needed to ignore commas in array definitions
   #define UNPACK( ... ) __VA_ARGS__
   #define _INIT_N(x) UNPACK x
   #define _INIT_PROGMEM(x) PROGMEM = x
 #endif
 
 #define STRINGIFY(X) #X
 #define TOSTRING(X) STRINGIFY(X)
 
 #define WLED_CODENAME "Niji"
 
 // AP and OTA default passwords (for maximum security change them!)
 WLED_GLOBAL char apPass[65]  _INIT(WLED_AP_PASS);
 #ifdef WLED_OTA_PASS
 WLED_GLOBAL char otaPass[33] _INIT(WLED_OTA_PASS);
 #else
 WLED_GLOBAL char otaPass[33] _INIT(DEFAULT_OTA_PASS);
 #endif
 
-// Hardware and pin config
-#ifndef BTNPIN
-  #define BTNPIN 0
-#endif
-#ifndef BTNTYPE
-  #define BTNTYPE BTN_TYPE_PUSH
-#endif
-
 #ifdef WLED_FORCE_WIFI_OFF
   #ifndef WLED_USE_ETHERNET
     #error "WLED_FORCE_WIFI_OFF requires WLED_USE_ETHERNET."
   #endif
+  #ifndef BTNPIN
+    #error "WLED_FORCE_WIFI_OFF requires BTNPIN to be explicitly defined."
+  #endif
+  #ifndef BTNTYPE
+    #error "WLED_FORCE_WIFI_OFF requires BTNTYPE to be explicitly defined."
+  #endif
   #if (BTNPIN < 0)
-    #warning "WLED_FORCE_WIFI_OFF: BTNPIN is disabled. Emergency AP long-press on button 0 will not be available unless configured in cfg.json."
+    #error "WLED_FORCE_WIFI_OFF requires BTNPIN >= 0 so emergency AP long-press can be triggered on button 0."
   #endif
   #if !((BTNTYPE == BTN_TYPE_PUSH) || (BTNTYPE == BTN_TYPE_PUSH_ACT_HIGH) || (BTNTYPE == BTN_TYPE_TOUCH))
-    #warning "WLED_FORCE_WIFI_OFF: BTNTYPE is not momentary by default. Emergency AP long-press on button 0 requires a momentary button type."
+    #error "WLED_FORCE_WIFI_OFF requires a momentary BTNTYPE (BTN_TYPE_PUSH / BTN_TYPE_PUSH_ACT_HIGH / BTN_TYPE_TOUCH)."
   #endif
 #endif
+
+// Hardware and pin config
+#ifndef BTNPIN
+  #define BTNPIN 0
+#endif
+#ifndef BTNTYPE
+  #define BTNTYPE BTN_TYPE_PUSH
+#endif
 #ifndef RLYPIN
 WLED_GLOBAL int8_t rlyPin _INIT(-1);
 #else
 WLED_GLOBAL int8_t rlyPin _INIT(RLYPIN);
 #endif
 //Relay mode (1 = active high, 0 = active low, flipped in cfg.json)
 #ifndef RLYMDE
 WLED_GLOBAL bool rlyMde _INIT(true);
 #else
 WLED_GLOBAL bool rlyMde _INIT(RLYMDE);
 #endif
 //Use open drain (floating pin) when relay should be off
 #ifndef RLYODRAIN
 WLED_GLOBAL bool rlyOpenDrain _INIT(false);
 #else
 WLED_GLOBAL bool rlyOpenDrain _INIT(RLYODRAIN);
 #endif
 #ifndef IRPIN
   #define IRPIN -1
 #endif
 #ifndef IRTYPE
   #define IRTYPE 0
 #endif
 
 #if defined(CONFIG_IDF_TARGET_ESP32S3) || defined(CONFIG_IDF_TARGET_ESP32C3) || defined(CONFIG_IDF_TARGET_ESP32S2) || (defined(RX) && defined(TX))
```

Stargazer2026:
@coderabbitai resume

DedeHai:
how do the two versions compare regarding code size?

Stargazer2026:
> how do the two versions compare regarding code size?

They're almost identical:

```markdown
| Code Base / Build Target | WIFI ON | WIFI OFF |
|--------------------------|---------|----------|
| Before Refactoring       | 1301777 | 1297105  |
| After Refactoring        | 1301789 | 1297077  |
```

Stargazer2026:
Just a quick follow-up in case anything is still open on my side.

If there are remaining concerns, adjustments, or preferred refinements to the current approach, I’m happy to work on them.

Otherwise I’ll wait for further feedback :-)

wled00/FX.cpp
```diff
@@ -6880,6 +6880,92 @@ void mode_midnoise(void) {                  // Midnoise. By Andrew Tuline.
 static const char _data_FX_MODE_MIDNOISE[] PROGMEM = "Midnoise@Fade rate,Max. length;!,!;!;1v;ix=128,m12=1,si=0"; // Bar, Beatsin
 
 
+/////////////////////////////
+//   * MIDNOISE ONSET      //
```
DedeHai: please put this in a new PR

Stargazer2026: Oops, that's what happens when you use "main" as source for a PR. Sorry for that.

I removed the accidental commit that got looped into this PR.

softhack007:
@coderabbitai full review

softhack007:
> `WLED_FORCE_WIFI_OFF` breaks the project's feature-toggle convention, so it adds a one-off compile-time API to maintain across headers, docs, and build flags. Please rename it consistently here and in the sample config/docs before merge.

I agree with the bunny here - maybe better to rename the feature flag to `WLED_DISABLE_WIFI`.

wled00/network.cpp
```diff
@@ -2,6 +2,25 @@
 #include "fcn_declare.h"
 #include "wled_ethernet.h"
 
+namespace {
+
+bool restartWiFiScanIfEnabled(bool requireMultiWiFi)
```
willmmiles, line 7: Factoring the common logic out to a function is a good idea, but conditionals in function names are an antipattern.

In a feature-toggle case like this, it's better to #ifdef gate the call sites instead.   Blanking out function bodies makes it hard to reason about what's actually happening when reading at the call sites.

wled00/wled.cpp
```diff
@@ -14,6 +14,58 @@
...
+namespace {
+
+bool regularWiFiEnabled()
+{
+#ifdef WLED_FORCE_WIFI_OFF
+  return false;
+#else
+  return true;
+#endif
+}
```
willmmiles, line 26: If you must expose this state as a variable, use a `static constexpr` variable to communicate clearly that it is a compile-time constant.

wled00/wled.cpp
```diff
@@ -14,6 +14,58 @@
...
+  if (!regularWiFiEnabled()) {
+    DEBUG_PRINTLN(F("WLED_FORCE_WIFI_OFF: disabling regular WiFi startup (AP button emergency remains available)."));
+    apBehavior = AP_BEHAVIOR_BUTTON_ONLY;
+    WiFi.mode(WIFI_OFF);
+    return;
+  }
+
+  WiFi.mode(WIFI_STA); // enable scanning
+  findWiFi(true);      // start scanning for available WiFi-s
+}
```
willmmiles, line 39: As for `restartWifiScan`: gate the feature at the call site, not inside the implementing function.   If a function is named `initRegularWifiStartup()`, then it should init the wifi any time it is called, not just sometimes.

wled00/wled.cpp
```diff
@@ -14,6 +14,58 @@
...
+
+bool handleForceWiFiOffConnection()
+{
+  if (regularWiFiEnabled()) return false;
+
+  DEBUG_PRINTLN(F("WLED_FORCE_WIFI_OFF active. WiFi stays OFF unless AP emergency mode is opened by button."));
+  apBehavior = AP_BEHAVIOR_BUTTON_ONLY;
```
willmmiles, line 46: It's not appropriate to mysteriously overwrite system config values in the main loop.  (I'm not saying the code we have is clean: just that we shouldn't repeat existing mistakes.)

For handling the configuration value, instead set an appropriate bootup default and dike out the reader sites that set it (`cfg.cpp`, `set.cpp`).

wled00/wled.cpp
```diff
@@ -14,6 +14,58 @@
...
+    return;
+  }
+
+  WiFi.mode(WIFI_STA); // enable scanning
+  findWiFi(true);      // start scanning for available WiFi-s
+}
+
+bool handleForceWiFiOffConnection()
+{
+  if (regularWiFiEnabled()) return false;
```
willmmiles, line 43: This call path is unnecessarily convoluted.
- First, we unconditionally call this function - expressly named as being handling a specific build-time configuration value, regardless of its setting
- Then we call some other function to test the build time value
- ...and return, cancelling out the call, if it's not set.

This needlessly obfuscates what's being done.   Please instead #ifdef the call at the call site, so the function (a) do exactly what it says, and (b) clearly get called only when needed.

wled00/wled.cpp
```diff
@@ -14,6 +14,58 @@
...
+  if (!apActive) {
+    WiFi.disconnect(true);
+    WiFi.mode(WIFI_OFF);
+  }
+  return true;
+}
+
+bool wifiScanRunning(bool wifiConfigured)
+{
+  if (!regularWiFiEnabled()) return false;
```
willmmiles, line 57: Is this check necessary?

wled00/wled.cpp
```diff
@@ -14,6 +14,58 @@
...
+  return wifiConfigured && multiWiFi.size() > 1 && WiFi.scanComplete() < 0;
+}
+
+int findWiFiIfEnabled(bool doScan = false)
```
willmmiles, line 61: Conditional in function name, again.

wled00/wled.cpp
```diff
@@ -686,6 +737,8 @@ void WLED::initAP(bool resetAP)
 void WLED::initConnection()
 {
   DEBUG_PRINTF_P(PSTR("initConnection() called @ %lus.\n"), millis()/1000);
+  if (handleForceWiFiOffConnection()) return;
```
willmmiles, line 740: The websocket line below is important for Ethernet too.

willmmiles (changes requested):
Thanks for submitting this.  My colleagues' concerns about safety are not ill-founded, but I also agree that that there are valuable use cases for being able to isolate the Wifi network interface from the rest of the system.   ESP32-P4 support is the top of that list - we still need V5 and ESP-IDF support but we can make some steps like this to get ready.

I concur that we should be using `WLED_DISABLE_WIFI` as the flag name.

This dovetails a bit with #5650 .

At the limit, I think there's a bigger "network management" refactor waiting in the wings -- both with the "dependencies/network" name collision with V5, as well as issues like #4703 which points out that the two interfaces can't be treated as truly independent in some hardware configurations.   The particular goal of treating wifi client support as a compile-time option would definitely be cleaner in that wider context.

PR #5395, opened 2026-02-23
Add FSEQ usermod (SD playback + FPP)

Danit2:
This is a fork of the original usermod by @Andrejchrcek.

This usermod has been updated to be compatible with the new usermod scheme introduced in WLED 16.0-alpha.

It introduces a new FSEQ usermod that enables playback of .fseq animations from an SD card, including a web UI and FPP/UDP synchronization support.

The PR adds the following core components:

- FSEQ player (fseq_player.cpp / fseq_player.h)
- SD manager (sd_manager.*)
- Web UI, upload handling, and FPP integration (usermod_fpp.h, web_ui_manager.*)
- Usermod registration (register_usermod.cpp)
- Library metadata and auto-build script (library.json, auto_fseq_sd.py)
- README documentation

Additional helper functionality includes:
- Buffered file uploads
- FPP ping and synchronization packet handling

Small modifications in existing wled00 files (const.h, json.cpp) are included to properly integrate the usermod into the WLED core.

## Summary by CodeRabbit

* **New Features**
  * Play .fseq animations from SD with selectable LED range, loop/single-play and realtime sync (HTTP/UDP).
  * Web UI at /fsequi for uploading, listing, deleting and controlling FSEQ files; status polling and upload progress.
  * REST endpoints for SD and FSEQ management (list/upload/delete/start/startloop/stop/status).
  * Automatic .xlz archive unpacking of FSEQ files and enhanced SD backend compatibility.

* **Documentation**
  * Added README detailing installation, configuration flags, endpoints and usage.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -2345,7 +2345,7 @@ void mode_colortwinkle() {
           unsigned index = i >> 3;
           unsigned  bitNum = i & 0x07;
           bitWrite(SEGENV.data[index], bitNum, true);
-          SEGMENT.setPixelColor(i, ColorFromPalette(SEGPALETTE, hw_random8(), 64, NOBLEND));
+          SEGMENT.setPixelColor(i, ColorFromPalette(SEGPALETTE, hw_random8(), gamma8inv(64), NOBLEND));  // note on gamma8inv: inverting results in non-linear brightness fade as originally designed
           break; //only spawn 1 new pixel per frame per 50 LEDs
         }
       }
@@ -2616,7 +2616,7 @@ static CRGBW twinklefox_one_twinkle(uint32_t ms, uint8_t salt, bool cat)
   unsigned hue = slowcycle8 - salt;
   CRGBW c;
   if (bright > 0) {
-    c = ColorFromPalette(SEGPALETTE, hue, bright, NOBLEND);
+    c = ColorFromPalette(SEGPALETTE, hue, gamma8inv(bright), NOBLEND); // note on gamma8inv: inverting results in non-linear brightness fade as originally designed
     if (!SEGMENT.check1) {
       // This code takes a pixel, and if its in the 'fading down'
       // part of the cycle, it adjusts the color a little bit like the
@@ -2651,19 +2651,18 @@ static void twinklefox_base(bool cat)
   if (SEGMENT.speed > 100) SEGENV.aux0 = 3 + ((255 - SEGMENT.speed) >> 3);
   else SEGENV.aux0 = 22 + ((100 - SEGMENT.speed) >> 1);
 
-  // Set up the background color, "bg".
+  // Set up the background color, "bg". Note: using gamma invert for brightness as the FX was written without any gamma correction, it will dim down too much now
   CRGBW bg = SEGCOLOR(1);
-  unsigned bglight = bg.getAverageLight();
+  unsigned bglight = bg.getRGBaverage();
   if (bglight > 64) {
-    bg = color_fade(bg, 16, true); // very bright, so scale to 1/16th
+    bg = color_fade(bg, gamma8inv(16), true); // very bright, so scale to 1/16th
   } else if (bglight > 16) {
-    bg = color_fade(bg, 64, true); // not that bright, so scale to 1/4th
+    bg = color_fade(bg, gamma8inv(64), true); // not that bright, so scale to 1/4
   } else {
-    bg = color_fade(bg, 86, true); // dim, scale to 1/3rd.
+    bg = color_fade(bg, gamma8inv(86), true); // dim, scale to 1/3rd
   }
-  bg = gamma32inv(bg); // need to invert gamma as the FX was written without any gamma correction and it will dim down too much otherwise
 
-  unsigned backgroundBrightness = bg.getAverageLight();
+  bglight = bg.getRGBaverage(); // update after scaling
 
   for (unsigned i = 0; i < SEGLEN; i++) {
 
@@ -2680,8 +2679,8 @@ static void twinklefox_base(bool cat)
     // on the "brightness = f( time )" idea.
     CRGBW c = twinklefox_one_twinkle(myclock30, myunique8, cat);
 
-    unsigned cbright = c.getAverageLight();
-    int deltabright = cbright - backgroundBrightness;
+    unsigned cbright = c.getRGBaverage();
+    int deltabright = cbright - bglight;
     if (deltabright >= 32 || (bg==0)) {
       // If the new pixel is significantly brighter than the background color,
       // use the new color.
--- a/wled00/colors.h
+++ b/wled00/colors.h
@@ -198,6 +198,11 @@ struct CRGBW {
   uint8_t getAverageLight() const {
     return (r + g + b + w) >> 2;
   }
+
+  // get the average of the R, G, B values
+  uint8_t getRGBaverage() const {
+    return ((r + g + b) * 21846) >> 16; // x*21846>>16 is equal to "divide by 3"
+  }
 };
 
 inline CHSV32::CHSV32(const CRGBW& rgb) {


usermods/FSEQ/web_ui_manager.bak
softhack007, line 1: This is a backup file, normally not needed. Please remove from your PR.

DedeHai:
the rabbit seems not happy at all about this

DedeHai:
I do not really get the purpose of this PR:
- the original one is not closed - did you reach out and can confirm it was abandoned?
- this PR does not address any of the points discussed in the original PR (see https://github.com/wled/WLED/pull/4570)

Danit2:
@DedeHai 

Yes, I know. The old PR is not closed. 
But the PR is over a year old and no longer compatible with the new usermoe schema. 

It also had a few bugs. 
- File upload from xlights was not completed.
- Usermode did not work on the esp32-s3.
- Files with special characters did not work.
- The detection of the ESP was incorrect and incomplete in the FPP Player.

I tried to get in touch, but unfortunately did not receive any replies.
The PR can be closed and work can continue on the old one.

I know that the PR does not address any of the points discussed in the old PR. 
However, I am not in a position to implement all of this. 

Several PRs have been made for the support of FSEQ files in recent years, but none have ever been approved. 
This is now simply a working usermode that also works with the new usermode schema.

You are welcome to do whatever you want with it.

netmindz:
Your PR must be from a branch, not just the main branch

Danit2:
Sorry, I didn't know. 
But why? 
How can I change this without closing the PR? Or should I create a new PR?

netmindz:
I've send you a PR with some draft *UNTESTED* changes  https://github.com/Danit2/WLED/pull/1/changes @Danit2

Danit2:
@netmindz 

I tested your PR.

I think the idea is good in theory, but I see a few problems.

1. The fpp player works. However, real-time mode is no longer available. As a result, when the LEDs are turned off, the effect is not played at all. Of course, the brightness also remains at the set value. This means that if the brightness is only set to 128, the entire sequence is played back with a brightness of 128. So it is no longer the fseq file that has the upper hand, but the wled settings.

2. I can no longer play the sequences directly. I have an effect called “FSEQ Player,” but there I only have a checkbox for “loop”; I don't see any files there.

image

4. My extra script no longer works. You now have to set the “-D WLED_USE_SD_SPI” flag in platform.ini and also the usermode “sd_card.” However, this can certainly be easily corrected. Theoretically, the user mode “sd_card” should be registered automatically when FSEQ is registered. Without an SD card, the user mode does not work anyway. 

I basically think it's better if the FSEQ Player works in real-time mode. That way, the FSEQ file is always the master.

I understand the benefit of the FSEQ Player as an effect. This allows the file to be played independently and offers greater possibilities, including for playing sequences during boot, etc. 
However, if the FSEQ Player is to be used to function synchronously with other instances from a higher-level system, and the higher-level system is to be the master, I find real-time mode to be preferable. 

Or it should simply be possible to do both. All fseq files as effects for independent playback.
And when the fseq files are accessed externally in real-time mode.

I think the switch from the proprietary SD adapter to user mode “sd_card” is fine.

netmindz:
> @netmindz
> 
> I tested your PR.
> 
> I think the idea is good in theory, but I see a few problems.
> 
> 1. The fpp player works. However, real-time mode is no longer available. As a result, when the LEDs are turned off, the effect is not played at all. 

Surely that's exactly what you would expect - off is off

> Of course, the brightness also remains at the set value. This means that if the brightness is only set to 128, the entire sequence is played back with a brightness of 128. So it is no longer the fseq file that has the upper hand, but the wled settings.

I'm not familiar with the file format, does the file format have an explicit brightness value? 

> 2. I can no longer play the sequences directly. I have an effect called “FSEQ Player,” but there I only have a checkbox for “loop”; I don't see any files there.

So this is where there is still work, as per earlier comments the choice of what to playback through the UI would be in the matrix tool, to be consistent with how gif playback works

> image

> 4. My extra script no longer works. You now have to set the “-D WLED_USE_SD_SPI” flag in platform.ini and also the usermode “sd_card.” However, this can certainly be easily corrected. Theoretically, the user mode “sd_card” should be registered automatically when FSEQ is registered. Without an SD card, the user mode does not work anyway.

Not sure which "script" you mean sorry. As for the dependency for sd_card, we can possibly define that in the library.json

> I basically think it's better if the FSEQ Player works in real-time mode. That way, the FSEQ file is always the master.

Hopefully we can address what you concerns around brightness as that seems to the be the only place where it's not in control

> I understand the benefit of the FSEQ Player as an effect. This allows the file to be played independently and offers greater possibilities, including for playing sequences during boot, etc. However, if the FSEQ Player is to be used to function synchronously with other instances from a higher-level system, and the higher-level system is to be the master, I find real-time mode to be preferable.

Let's continue exploring where you feel it falls short so we get the best of both (hopefully)

> Or it should simply be possible to do both. All fseq files as effects for independent playback. And when the fseq files are accessed externally in real-time mode.

I think that would actually be less capable as you would loose all the benefits of it being an effect and I'm unsure what we gain (other than brightness)

> I think the switch from the proprietary SD adapter to user mode “sd_card” is fine.

Cool

Danit2:
The FSEQ file contains the exact pixel values for each color  in the range 0-255.
This includes brightness. 
The file then contains all data for all pixels per frame plus the frame intervals, e.g., 25 ms or 50 ms.

The FSEQ player works in exactly the same way as when the data comes via DDP or E1.31. The only difference is that here the pixel data is stored locally and only needs to be retrieved. But actually, the external player is the master here.

With DDP or E1.31, you don't have the option of turning off the LED or changing the brightness either. 

What are the “advantages” of the FPP player playing the sequence as an effect? 

I would find it better if the effect player were used for local playback of the sequence, but as soon as a higher-level system comes in, real-time mode should come into play. 

Because in that case, the FPP player must have control over the system and also control the LEDs regardless of what settings are currently made in wled.

In my case, my LEDs and ESPs have no power, and I then switch them all on.
After that, they have to do exactly what the FPP player says. The LEDs should also remain dark until the sequence starts. 
If I want this, I would have to create a boot preset with the status “on,” brightness “255,” and color “black.”

DedeHai:
@netmindz this PR is just an update to the original PR with some improvemetns. As discussed there this is not the way to do it if we want to merge it into main: it should be done as an FX like the gif player and SD card support should be added as its own UM for other things - like the gif player - to benefit from that as well.  

edit: this concept of SD card + FX may need some refinement

netmindz:
> @netmindz this PR is just an update to the original PR with some improvemetns. As discussed there this is not the way to do it if we want to merge it into main: it should be done as an FX like the gif player and SD card support should be added as its own UM for other things - like the gif player - to benefit from that as well.
> 
> edit: this concept of SD card + FX may need some refinement

You clearly missed this  @DedeHai 

> I've send you a PR with some draft UNTESTED changes https://github.com/Danit2/WLED/pull/1/changes @Danit2

willmmiles:
@netmindz  +1 for the direction you've taken with the FSEQ player in your sub-PR.

Regarding the semantics of live modes: in the broadest sense, when playing live data, there's a number of decisions about how much of the rendering framework features get applied.
 - Have transitions in to/out of live mode?
 - Apply live data to which segment(s)?
 - Apply per-segment or global brightness adjustments?
 - Apply gamma correction or other color balancing to RGB data?
 - Apply LED maps?
 - Apply ABL or just YOLO the pixel data?

My thought is that an ideal solution would offer end users control over all of these.  In practice, I think the way to get there is to offer "enter live mode"/"exit live mode" preset slots, like the boot preset slot; this lets users customize the settings applied in live mode however they like -- assigning specific segments to live input or whatever.   If no preset is selected, thjen we generate a default (make 1 segment that fills all pixels with received live data, sets the segment and global brightness to maximum, and execute the currently configured transition).

With that in mind, in the near term in the context of FPP, I'd probably suggest expanding your `activateFseqEffect()` to behave more like existing live modes, including recognizing `useMainSegmentOnly` and `realtimeRespectLedMaps`.   Later on we can co-opt most of that code to become a general-purpose `enterLiveMode()` when other live inputs can also be mapped as FX; and then finally we can implement a preset hook for it.

DedeHai:
> You clearly missed this @DedeHai
> 
> > I've send you a PR with some draft UNTESTED changes https://github.com/Danit2/WLED/pull/1/changes @Danit2

Indeed I did, I only looked at the diff of this PR... did not yet look at your changes in detail but that is definitely a good step forward.

Danit2:
> With that in mind, in the near term in the context of FPP, I'd probably suggest expanding your `activateFseqEffect()` to behave more like existing live modes, including recognizing `useMainSegmentOnly` and `realtimeRespectLedMaps`. Later on we can co-opt most of that code to become a general-purpose `enterLiveMode()` when other live inputs can also be mapped as FX; and then finally we can implement a preset hook for it.


I like this way. 
@netmindz Can you adjust your PR accordingly?

netmindz:
I'm a bit confused now @DedeHai 

You say you like my pr to make it an effect, but yet you also say it should behave like existing live modes - surely these statements are contradictory?

Or do you mean that the local playback works like gif as an effect, but when FPP we instead behave like realtime modes, as this original PR does? (Just a real realtime mode where you already have the data on the SD card, a bit like replaying recorded ArtNet or DDP streams)

DedeHai:
> it should behave like existing live modes

I do not remember stating that. I don't see how live modes and FSEQ player are related, or does live mode already use this protocol? I did not dive into the topic as I do not use FSEQ nor xlights nor dmx, I just can offer insight into how the gif player / pixelforge handles it. If this is something that goes from file -> decode -> LED it should not go through any live mode but do as the gif player does.

Danit2:
@DedeHai @netmindz 

The FSEQ Player is actually a player controlled by the FPP player and is used to show live shows such as DMX, DDP, or ArtNet.
The bonus is that the LED data is available locally on site and does not have to be sent live over the network. 

The FSEQ files contain pixel data for all LEDs. 
The FPP Player then specifies which file is to be played when and synchronizes all FSEQ Players with each other to enable synchronous playback on all devices. 

However, since the data is now on the device with this player, it can also be started and played locally and independently of the FPP player. 
But this is only an addition and not the actual use.

Therefore, it would make sense to me if, as long as the FSEQ player is controlled at a higher level, this functions as live mode and only when using it as a local player should it be treated as an effect.

DedeHai:
> The FSEQ Player is actually a player controlled by the FPP player and is used to show live shows such as DMX, DDP, or ArtNet.
> The bonus is that the LED data is available locally on site and does not have to be sent live over the network.

I don't quite understand. if its a player of FSEQ files, which I understand as sort of video files, the whole rest makes not much sense to me. Live mode is for live data streaming, if that is not used, it should be an FX. If the goal is to turn WLED into a FPP "host" of sorts that manages these animations, it still should be an FX on the slaves as you could trigger sequences of lights very well using and animated file format, hence the similarity to gif. Maybe I am missing the bigger picture?

willmmiles:
> I don't quite understand. if its a player of FSEQ files, which I understand as sort of video files, the whole rest makes not much sense to me. Live mode is for live data streaming, if that is not used, it should be an FX. If the goal is to turn WLED into a FPP "host" of sorts that manages these animations, it still should be an FX on the slaves as you could trigger sequences of lights very well using and animated file format, hence the similarity to gif. Maybe I am missing the bigger picture?

The FPP system consists of two parts:
- A "Player" component that hosts FSEQ animation files, and can render them to LED outputs;
- A "Controller" component that sends a stream of "play this file" commands to the Player devices.

As I see things, the objective is to add support to allow WLED to act as a "Player" device to the FPP "Controller".  When in use, it should work like a "live" mode, though instead of "set the pixels to this array of colors", the command stream is "play back such-and-such FSEQ file".  I believe the expectation is that the WLED device follows the commands from the "Controller" when it sends them, overriding any local state, but then falls back to the original display output if the Controller stops sending commands; and also our UI shows that the device is being driven by external commands; just like any other live mode.

From a code perspective, whether or not FSEQ rendering is performed by the FX renderer when commanded by the FPP Controller is to some extent an implementation detail.  But by implementing it that way we take on some additional complexity:
- How do we back up/restore the old segment state during FPP "live" playback?
- How do we ensure that FPP live FSEQ playback has the same behaviour as other "live" modes, in terms of blending and color correction and brightness control and "first segment only" and whatnot?

I believe what @Danit2 is trying to convey is that all of that can be avoided if "live FPP" mode runs the FSEQ renderer directly overtop the global buffer/first segment, using the same output path as other live inputs today.  And also that 99% of users requesting FPP integration don't care about playing back FSEQ files outside of the live mode; the key feature is being able to replace a RPi-based "Player" with an ESP32-based "Player".

So you can kind of blame me for this -- I've been pushing to ask for the FPP integration to do something a little more complex (an FX-based "live mode" with the full power of our rendering engine available) because I'm hoping to leverage that development to add those same features to the existing live data integrations.  @Danit2, I do recognize that it is asking for extra work here, though; that said, I think it's only a little bit further along past the expansion @netmindz sent you in his PR.

DedeHai:
so generating animations as FSEQ files and playing them back similar to the gif player is not a use case? 

edit:
as I am unfamiliar with the tools used: would playing FSEQ files locally as an FX make generating custom "effects" much easier? as in: use xlights or similar to generate the animation(s) then just stuff them in a playlist as opposed to "live" control?

edit2:
I am not opposed to using live mode if that is the only real use case.

netmindz:
> > it should behave like existing live modes
> 
> I do not remember stating that. I don't see how live modes and FSEQ player are related, or does live mode already use this protocol? I did not dive into the topic as I do not use FSEQ nor xlights nor dmx, I just can offer insight into how the gif player / pixelforge handles it. If this is something that goes from file -> decode -> LED it should not go through any live mode but do as the gif player does.

Sorry. My mistake. It was @willmmiles

willmmiles:
> so generating animations as FSEQ files and playing them back similar to the gif player is not a use case?
> 
> edit: as I am unfamiliar with the tools used: would playing FSEQ files locally as an FX make generating custom "effects" much easier? as in: use xlights or similar to generate the animation(s) then just stuff them in a playlist as opposed to "live" control?
> 
> edit2: I am not opposed to using live mode if that is the only real use case.

Local FSEQ playback is not a common use case for people who are using the FPP Controller software. (*Maybe* they would find it useful for testing that an animation has been uploaded and works correctly?)

For everyone else, though, yes, I think FX-style FSEQ playback would be a useful feature for us to have, just like the GIF player.  It would allow users to download animations and put them in a playlist, or control them with whatever other WLED integration they already have.  I don't know what tools are used for making FSEQs - if it's easier than GIFs or whatnot - but whatever they are, letting people use them if they like them seems like a clear win.

Danit2:
@willmmiles So what are the next steps?
I am not in a position to program this the way you want it. I don't know the program well enough for that. 

However, I am very willing to test your changes.

Danit2:
@coderabbitai review

Danit2:
Hello everyone,

I have now created a new version of usermode.
I have taken a completely new approach here. 

Please take a look at it and let me know whether this approach is feasible or not?

https://github.com/Danit2/WLED_FSEQ/tree/FSEQ_Effect_player

https://github.com/Danit2/WLED_FSEQ/blob/FSEQ_Effect_player/usermods/FSEQ/README.md

softhack007:
@Danit2 esp32 builds are failing, please check

````shell
usermods/FSEQ/fseq_player.cpp: In static member function 'static void FSEQPlayer::syncPlayback(float)':
usermods/FSEQ/fseq_player.cpp:266:51: warning: declaration of 'secondsElapsed' shadows a previous local [-Wshadow=compatible-local]
 void FSEQPlayer::syncPlayback(float secondsElapsed) {
                                                   ^
usermods/FSEQ/fseq_player.cpp:19:7: note: shadowed declaration is here
 float FSEQPlayer::secondsElapsed = 0;
       ^~~~~~~~~~
````
--> please use unique names, shadowing a global variable is bad for maintainability of your code.


````shell
usermods/FSEQ/usermod_fpp.h: At global scope:
usermods/FSEQ/usermod_fpp.h:739:37: warning: inline variables are only available with -std=c++17 or -std=gnu++17
 inline const char UsermodFPP::_name[] PROGMEM = "FPP Connect";
````

--> there are several of these warnings. I would suggest that you move constants into global scope, and change ``inline const `` into ``static const ``. Inlining strings is generally not a good idea, because each inlined copy will consume flash space.

````shell
usermods/FSEQ/web_ui_manager.cpp: In lambda function:
usermods/FSEQ/web_ui_manager.cpp:553:16: error: 'SD_ADAPTER' was not declared in this scope
     bool res = SD_ADAPTER.remove(path.c_str());
                ^~~~~~~~~~
`````

--> it seems like SD_ADAPTER is unknown on esp32 builds, please fix.

Danit2:
@softhack007 
I've tried to fix all the bugs.

The problem with `SD_ADAPTER` is that my usermod requires the usermod `sd_card.` That's why the build of the standalone usermod fails. I've tried to fix this.

PR #5396, opened 2026-02-24
avoid I2S resource conflict (aPLL) with ethernet - V4 builds only (fixes #5391)

### LINKED ISSUE #5391: Conflict between Ethernet and mic settings on main

softhack007:
When AR ties to initialize a PDM microphone, this stops ethernet from working because the aPLL unit is needed for ethernet.
Solution: don't use aPLL clock for I2S PDM microphones.

Backport to 0.15.x not needed - the problem only occurs on V4 builds.

Ethernet with internal RMII clock (typically on GPIO 0/16/17) uses the Audio PLL (APLL) to generate the 50 MHz reference clock for RMII. I2S audio input cannot use the same aPLL in this case.

see https://github.com/wled/WLED/issues/5391

## Summary by CodeRabbit

* **New Features**
  * Enhanced audio source type detection to differentiate between PDM and I2S digital microphones in the interface.

* **Bug Fixes**
  * Corrected PDM microphone labeling from "I2S PDM" to "Generic PDM" for improved accuracy and clarity in audio source selection.

### DIFF
--- a/usermods/audioreactive/audio_reactive.cpp
+++ b/usermods/audioreactive/audio_reactive.cpp
@@ -1206,7 +1206,7 @@ class AudioReactive : public Usermod {
           break;
         #if  !defined(CONFIG_IDF_TARGET_ESP32S2) && !defined(CONFIG_IDF_TARGET_ESP32C3)
         case 5:
-          DEBUGSR_PRINT(F("AR: I2S PDM Microphone - ")); DEBUGSR_PRINTLN(F(I2S_PDM_MIC_CHANNEL_TEXT));
+          DEBUGSR_PRINT(F("AR: Generic PDM Microphone - ")); DEBUGSR_PRINTLN(F(I2S_PDM_MIC_CHANNEL_TEXT));
           audioSource = new I2SSource(SAMPLE_RATE, BLOCK_SIZE, 1.0f/4.0f);
           useBandPassFilter = true;  // this reduces the noise floor on SPM1423 from 5% Vpp (~380) down to 0.05% Vpp (~5)
           delay(100);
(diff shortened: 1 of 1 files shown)


netmindz:
Do we not have any ethernet builds for S3? Aren't we using V4 for the S3 on 0.15? @softhack007 ?

netmindz:
Do we have the same bug in MM too?

softhack007:
> Do we not have any ethernet builds for S3? Aren't we using V4 for the S3 on 0.15? @softhack007 ?

Good point - right now WLED only support ethernet on classic esp32. 
WLED 0.15 supports custom "V4" builds, so we could also bring the patch back to 0.15.x. V3 builds are not affected by my patch, they don't have the problem.

We might need to look at S3 (and P4) later, once that the W5500 chip supports from @troyhacks comes.

> Do we have the same bug in MM too?

yep, and I've added the same fix to ``mdev`` today. There is still an uncertainty about "I2S slave mode" (PR from @troyhacks) and "I2S with Mclk" - not sure if these modes work well without the aPLL timer.

softhack007:
ok, ready to merge. I'll also cross-merge into 0.15.x, to help ethernet users with "V4" builds.

PR #4893, opened 2025-08-31
Hostname unification 2: The Force Push Strikes Back

milhnl:
Hi, as promised, the cleaner version. See https://github.com/wled/WLED/pull/4751 and https://github.com/wled/WLED/pull/4888.

One thing: there's still one `cmDNS` reference in `wled00/wled_eeprom.cpp:83:29:`. This did not break compilation for me, probably due to `WLED_ADD_EEPROM_SUPPORT` not being defined. I'm not sure what the correct fix would be.

## Summary by CodeRabbit

* New Features
  * Revamped Network Settings: host name field, mDNS enable toggle with live URL preview, Ethernet type selector, ESP-NOW controls, and linked remotes management.
  * Unified device hostname across services (mDNS, OTA, MQTT) with sensible defaults.
* Bug Fixes
  * More reliable hostname/IP resolution and MQTT topic/client defaults; AP IP shown only when AP is active.
* Style
  * Settings labels updated (“Network Setup”, “Hardware Setup”, “Time & Scheduler”).
  * “Device name” replaces “Server description”; invalid inputs are clearly highlighted.

PR #5397, opened 2026-02-24
 Fix: ESP32-S3 DHCP Hostname timing issue (Option 12) for OpenWrt/dnsmasq

gsieben:
Description:
On ESP32-S3 (and other fast S3/C3 targets), the WiFi stack starts the DHCP handshake (DHCPDISCOVER) almost immediately after WiFi.begin(). In the current WLED implementation, WiFi.setHostname() is often called too late, after the first discovery packet has already been sent with the default esp32s3-xxxxxx identifier.
This causes issues in strict network environments like OpenWrt (dnsmasq), where the router caches the first hostname it sees during the lease process and ignores subsequent updates.

Changes:
I 'moved' or better set another WiFi.setHostname() call to the very beginning of WLED::initConnection(). By doing this, the hostname (Option 12) is injected into the LwIP stack before the radio starts transmitting.
The fix uses the dynamic cmDNS buffer but falls back to the hardcoded DEVICE_NAME if the config hasn't been loaded yet (first boot).

I successfully tested the fix with my OpenWrt server (logread -f | grep -i dhcp) and an ESP32-S3-DevKitC-N16R8 (ESP32-S3 (QFN56) (revision v0.2)) as well as an ESP32-DevKitC (ESP32-D0WDQ6 (revision v1.0)).

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced WiFi hostname assignment on ESP32 devices, now prioritizing custom network identification when configured.

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -865,6 +865,17 @@ void WLED::handleConnection()
   #endif
   const bool wifiConfigured = WLED_WIFI_CONFIGURED;
 
+  // The DNS name of the ESP must be set before the first connection to the DHCP server. Otherwise, the esp default name, such as “esp32s3-267D0C,” will be used. GG
+  #ifdef ARDUINO_ARCH_ESP32
+    // Use cmDNS (this is the buffer that holds your name)
+  if (cmDNS && cmDNS[0] != '\0') {
+    WiFi.setHostname(cmDNS);
+  } else {
+    // Fallback to DEVICE_NAME if cmDNS is still empty
+    WiFi.setHostname(WLED_HOST_NAME);
+  }
+  #endif
+  
   // ignore connection handling if WiFi is configured and scan still running
   // or within first 2s if WiFi is not configured or AP is always active
   if ((wifiConfigured && multiWiFi.size() > 1 && WiFi.scanComplete() < 0) || (now < 2000 && (!wifiConfigured || apBehavior == AP_BEHAVIOR_ALWAYS)))


softhack007:
Build failure

````log
wled00/wled.cpp: In member function 'void WLED::handleConnection()':
wled00/wled.cpp:875:22: error: 'WLED_HOST_NAME' was not declared in this scope
     WiFi.setHostname(WLED_HOST_NAME);
                      ^~~~~~~~~~~~~~
````

blazoncek:
This has already been (properly) addressed in #4893 quite some time ago but unfortunately project isn't very well maintained recently.
Or at least priorities are all mixed up.

softhack007:
> project isn't very well maintained recently. Or at least priorities are all mixed up.

says who? I thought that you asked not be involved in these topics any more?

blazoncek:
> says who?

Someone who's taken care of the project for about 3 years and contributed hefty amount of code. To say the least.

> you asked not be involved in these topics any more?

I still have the right to comment, don't I? Or is this now a dictatorship and opinions are frowned upon? In such case **block** me otherwise I'll keep commenting.
Anyway, if you'd be taking care of the project (instead of being AWOL for over a year) you'd known that the base issue was already being taken care of and the PR is waiting.

@gsieben please check #4893 and report back in that PR. However I won't be able to assist you in case of issues but hopefully there still is someone willing to process old PRs.

softhack007:
included in https://github.com/wled/WLED/pull/5424 (merged into main)

PR #5400, opened 2026-02-27
feat(eps-now): add 3 scenes for GLEDOPTO ESP-NOW WLED Remote Control

GLEDOPTO:
This PR adds three new scenes to the EPS-NOW remote controller to support the GLEDOPTO ESP-NOW WLED Remote Control.

**Product link:**   https://gledopto.com/h-pd-146.html

## Summary by CodeRabbit

* **New Features**
  * Added support for three new WizMote remote buttons (5, 6, 7) for one-press access to lighting presets: Candle, Random Color, and Fade for quicker mode changes.

### DIFF
--- a/wled00/remote.cpp
+++ b/wled00/remote.cpp
@@ -220,8 +220,8 @@ void handleRemote() {
       case WIZMOTE_BUTTON_TWO            : presetWithFallback(2, FX_MODE_BREATH,        0); break;
       case WIZMOTE_BUTTON_THREE          : presetWithFallback(3, FX_MODE_FIRE_FLICKER,  0); break;
       case WIZMOTE_BUTTON_FOUR           : presetWithFallback(4, FX_MODE_RAINBOW,       0); break;
-      case WIZMOTE_BUTTON_FIVE           : presetWithFallback(5, FX_MODE_DISSOLVE,      0); break;
-      case WIZMOTE_BUTTON_SIX            : presetWithFallback(6, FX_MODE_BLINK,         0); break;
+      case WIZMOTE_BUTTON_FIVE           : presetWithFallback(5, FX_MODE_CANDLE,        0); break;
+      case WIZMOTE_BUTTON_SIX            : presetWithFallback(6, FX_MODE_RANDOM_COLOR,  0); break;
       case WIZMOTE_BUTTON_SEVEN          : presetWithFallback(7, FX_MODE_FADE,          0); break;
       case WIZMOTE_BUTTON_NIGHT          : activateNightMode();                             break;
       case WIZMOTE_BUTTON_BRIGHT_UP      : brightnessUp();                                  break;


wled00/remote.cpp
```diff
@@ -217,6 +220,9 @@ void handleRemote() {
       case WIZMOTE_BUTTON_TWO            : presetWithFallback(2, FX_MODE_BREATH,        0); break;
       case WIZMOTE_BUTTON_THREE          : presetWithFallback(3, FX_MODE_FIRE_FLICKER,  0); break;
       case WIZMOTE_BUTTON_FOUR           : presetWithFallback(4, FX_MODE_RAINBOW,       0); break;
+      case WIZMOTE_BUTTON_FIVE           : presetWithFallback(5, FX_MODE_SCAN,          0); break;
```
DedeHai: better use a single led compatible FX as a fallback. like fireflicker

GLEDOPTO: OK, thanks

DedeHai: why did you fix it and then unfix it?

GLEDOPTO: > why did you fix it and then unfix it?

The fireflicker effect is already working. The code is: 
`case WIZMOTE_BUTTON_THREE : presetWithFallback(3, FX_MODE_FIRE_FLICKER, 0); break;`

DedeHai: good point, but my comment was about support for non-strip setups to have a fallback that works.

GLEDOPTO: > good point, but my comment was about support for non-strip setups to have a fallback that works.

Okay, I understand what you mean.
I think the following effects also work well on a single LED, for example:
**1. Color change effects:**
- FX_MODE_RAINBOW
- FX_MODE_TRICOLOR_FADE
- FX_MODE_COLORWAVES

**2. Brightness change effects:**

- FX_MODE_BREATH
- FX_MODE_STROBE/FX_MODE_MULTI_STROBE

**3. Random blinking effects:**

- FX_MODE_COLORTWINKLE

However, users usually reset these presets, so I don't think what's set here is very important. Perhaps setting it back to FX_MODE_FIRE_FLICKER? Do you have any better suggestions? Thank you!

DedeHai: users will very likely test the remote before setting up presets, if a fallback fails it will default to "static". The FX with single pixel support can be filtered in the FX list, the default should not be a too aggressive one like strobe (blink is also not really a good choice).  Can you check for other FX that have single pixel support and replace blink and dissolve? I think candle may be one candidate.

GLEDOPTO: > users will very likely test the remote before setting up presets, if a fallback fails it will default to "static". The FX with single pixel support can be filtered in the FX list, the default should not be a too aggressive one like strobe (blink is also not really a good choice). Can you check for other FX that have single pixel support and replace blink and dissolve? I think candle may be one candidate.

Thank you for the feedback! Based on your requirements for gentle, single-LED compatible effects, I've updated the fallbacks:

- BUTTON_FIVE: FX_MODE_SCAN → FX_MODE_CANDLE (as you suggested)
- BUTTON_SIX: FX_MODE_BLINK → FX_MODE_RANDOM_COLOR (smooth color transitions)
- BUTTON_SEVEN: Keep FX_MODE_FADE (already gentle and works well)

What do you think?

DedeHai: thats a good choice

GLEDOPTO: OK, I have resubmitted

PR #5423, opened 2026-03-13
Create usermod phillips hue v2

netmindz:
## Summary by CodeRabbit

* **New Features**
  * Philips Hue V2 integration for ESP32: syncs on/off, brightness, color, temperature and effects; SSE streaming with polling fallback, reconnect/backoff, and configurable polling/apply options
  * HTTPS by default with optional plain-HTTP override, link-button authentication, mDNS bridge discovery, and light discovery/selection

* **Documentation**
  * Added comprehensive user guide with setup, TLS options, settings and troubleshooting

* **Chores**
  * Registered new usermod and added library manifest and usermod ID


netmindz:
@coderabbitai review  - can you pay particular attention to if the code handles self-signed certificates please

netmindz:
It appears to be working for me, either if i use HTTP or build with the official ESP Platform  (not tasmota as that lacks TLS)

softhack007:
@coderabbitAI How does the flash (program size) increase, based on the additional components (TLS, HTTPS, certificate bundle, etc) needed by this PR?

DedeHai:
> TLS in insecure mode carries an **~48 KB SRAM cost on ESP32** at connection time

oof, thats a heavy toll if its correct. It means that anything over 1000LEDs will probably not be possible without PSRAM on an ESP32, the S3 has more RAM. Also only a single connection would be possible if this is per connection.

netmindz:
I'm not too worried about pixel count as this is a single pixel as far is hue is concerned, so this feature is most suitable non-addressible LEDs 

The use case would say be turning on the lights under the kitchen counter at the same time as the main light bulb in the kitchen

Issue #5427, opened 2026-03-16
Random flickers after updating to 0.15.4 (8266 bit-bang driver)


coltography:
### What happened?

I have an esp8266 (nodemcu) with 350 WS281B on GPIO15. Powered by 10A 5V PSU. It's worked without issue for over 2~ years and I've updated it any time there's one available. Updating to 0.15.4 made it randomly flicker with no real cause, nothing else changed. I originally had it on GPIO2 and thought maybe there was a problem with this, so changed to GPIO15 and the issue persisted. 

Downgraded to 0.15.3 and the random flickers are gone. I checked Home Assistant to see if the device was rebooting or going unavailable during these flickers but it wasn't. 

### To Reproduce Bug

Not sure. Power on device and allow it to run.

### Expected Behavior

No flickers

### Install Method

Binary from WLED.me

### What version of WLED?

0.15.4

### Which microcontroller/board are you seeing the problem on?

ESP8266

### Relevant log/trace output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DedeHai:
That is very strange as there was no change to code that touches anything regarding LED outputs on the ESP8266.
How often do you see the flickering and what does it look like? 
What levelshifter are you using and how long and how tidy are the data wires?

kbickar:
I'm also seeing this issue. I don't have a level shifter, the controller is about 1 inch from the first LED, and it didn't happen before the update

DedeHai:
WLED can not compensate for insufficient hardware (yet) so your best option is to just stay on 0.15.3

kbickar:
That's definitely an odd way to state that the recent minor update made it more bloated and less efficient.

Everyone knows the esp8266 is a very capable board that can handle most usages (as it has for years) . Saying it's "insufficient hardware" is a lame excuse. If support is being dropped, the project should be forked for a new major version while the compatible version continues

DedeHai:
like I said, there was no change to code that affects LED output and by insufficient hardware I did not mean the ESP8266.
I have done extensive investigation on flickering and only very few instances are related to software bugs, the vast majority is systems that are not built properly and are just at the brink of being unstable and the slightest change in timing (we are talking a handful of nanoseconds) can tip them over into the flickering side.
It is not something that can currently be solved.

kbickar:
I have around 150 LEDs so nothing major (I have another one that's not updated with over 600 LEDs).

[...]

PR #5435, opened 2026-03-21
(0.15.x) wait for LEDs output to finish before OS calls that potentially suspend interrupts

softhack007:
Add a short wait time before calling file.close() and ESP.getFreeHeap(), to prevent LED glitches caused by interrupt stalls.

potential fix for #5427 

## Summary by CodeRabbit

* **New Features**
  * Added LED synchronization methods to improve system coordination during operations

* **Bug Fixes**
  * Enhanced LED timing synchronization during file handling and uploads
  * Improved LED output coordination in firmware update and OTA processes
  * Added LED synchronization before heap memory status checks
  * Better coordination of LED transmission with system task execution

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1477,7 +1477,7 @@ bool WS2812FX::isUpdating() const {
 }
 
 /**
- * Be nice, but not too nice - wait util LEDs are idle, or maxWaitMS milliseconds have passed
+ * Be nice, but not too nice - wait until LEDs are idle, or maxWaitMS milliseconds have passed
  * On 8266 this call will _not_ wait outside the main loop context
  *   Function returns isUpdating() status after waiting
  */


wled00/FX_fcn.cpp
```diff
@@ -1476,6 +1476,27 @@ bool WS2812FX::isUpdating() const {
...
+    if (can_yield()) {
+      // If we are in a yieldable context (main loop), wait until the LEDs output finishes
+      yield();
+      unsigned long waitStart = millis();
+      while (strip.isUpdating() && (millis() - waitStart < maxWaitMS)) delay(1);
+      yield();
+    }
+  #endif
+  return isUpdating();
+}
```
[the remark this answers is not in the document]
softhack007: pairing the wait with strip.suspend() / strip.resume() has to be done by the caller, because strip.resume() at the end of the wait function would not protect the critical call.

Implemented in the next commit.

wled00/file.cpp
```diff
@@ -34,13 +34,26 @@ static File f; // don't export to other cpp files
...
+  bool haveSuspended = false;
+  #if defined(WLED_USE_SHARED_RMT) || defined(__riscv) || !defined(ARDUINO_ARCH_ESP32)
+  if (!strip.isSuspended()) { strip.suspend(); haveSuspended = true; }// prevent that a new strip.show() starts after waiting
+  strip.waitForLEDs(15); // be nice, but not too nice. Waits up to 15ms
+  #endif
+
+  // In 8266 async webserver / lwIP interrupt context can_yield() is false and we cannot wait,
+  // so we proceed immediately — the close is still required to prevent file corruption.
   f.close(); // "if (f)" check is aleady done inside f.close(), and f cannot be nullptr -> no need for double checking before closing the file handle.
+  if (haveSuspended) strip.resume();  // end of critical section - new LEDs updates are allowed again
```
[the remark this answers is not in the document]
softhack007: @coderabbitAI please perform an in-depth controlflow analysis to justify your recommendation. consuming doCloseFile early in closeFile() should prevent unnecessary attempts to double-close the same file from parallel tasks.

wled00/file.cpp
```diff
@@ -34,13 +34,26 @@ static File f; // don't export to other cpp files
...
   DEBUGFS_PRINTF("took %d ms\n", millis() - s);
-  doCloseFile = false;
```
softhack007, line 43: @coderabbitAI i would like a second check to be sure that setting ``doCloseFile=false`` before actually performing ``f.close()`` does not create new failure possibilities in the WLED architecture
* find all call sites of `writeObjectToFile()`,  `writeObjectToFileUsingId()`, `readObjectFromFile()`,  `readObjectFromFileUsingId()`. Find out if each calls belong to the main task or webserver context.
* The feared scenario: one call is inside `closeFile()` - with `doCloseFile` already set to `false` but `f.close()` not completed. Can a parallel task now replace ``f`` by performing a new ``f = open(...)`` in parallel?
* also compare the C++ documented behaviour in our case (static FILE f) to the assumed behaviour in any hazard scenario you may find.
* Do you consider the cases you found credible and likely?

wled00/file.cpp
```diff
@@ -34,13 +34,26 @@ static File f; // don't export to other cpp files
...
 void closeFile() {
+  if (!doCloseFile || !f) { doCloseFile = false; return; }  // file not open, or no request to close -> nothing to do, nothing to wait
   #ifdef WLED_DEBUG_FS
     DEBUGFS_PRINT(F("Close -> "));
     uint32_t s = millis();
   #endif
+  doCloseFile = false; // consume flag early, to reduce the time window for concurrent closing attempts from several tasks.
+
+  // f.close() flushes to flash and briefly disables cache-related interrupts on ESP8266,
+  // which can corrupt WS281x LED timing (bit-bang / UART) if it coincides with strip.service().
```
willmmiles: This can never happen on ESP8266.  It's impossible to be handling file IO during bus operations because the system is using cooperative multitasking -- it cannot be pre-empted.  The strip operation must complete before any web handling/file IO can proceed.

softhack007: @willmmiles i thought that LEDs output is interrupt driven, also on 8266?

Edit: I agree that "if f.close() coincides with strip.service()." is misleading (AI wording....). Can the 8266 LEDs driver continue to send data in the background (ISR) ?

willmmiles: I'm just triple-checking this now..

For digital LEDs:
  - Bigbang buffer is blocking (obviously)
  - DMA methods (pin 3) uses full up-front conversion instead; it's memory intensive
  - The UART method (pins 1 or 2) do use an ISR, but they're configured to operate entirely out of SRAM, so it will work with flash disabled

For analog outputs, the system implements soft PWM using an NMI; it's also entirely SRAM and safe to use during flash IO.

I have to take a break for a bit but I've got my test board up and running, I'll see what I can find later today.

willmmiles: There is a possible issue if someone's trying to use soft-PWM analog and bit-banging simultaneously, but that's firmly in the category of "has never and probably will never work".

softhack007: @willmmiles I just made claude.AI munch through all the related source code (8266 SDK, arduino-8266, NeopixelBus UART method for 8266).

It's still "stupid AI day", but does this statement make sense for you? It looks like f.close() is very unlikely to cause glitches on 8266 (but its still a good idea to wait on -C3 and other risc-v chips).

> In ESP8266 builds, f.close() on LittleFS is not just a handle close; it performs sync/flush work that can trigger SPI flash write/erase in the SDK.
> During those SPI flash operations, the SDK can mask interrupts and disable flash cache access for a short or sometimes noticeable window.
> Running ISR code from SRAM/IRAM is relevant for surviving flash-cache-off periods without crashing or illegal instruction fetches. It does not solve interrupt masking. If the SDK raises interrupt level (for flash write/erase), the ISR will not run regardless of where it is stored.
> For your current 8266 WLED setup (synchronous UART method), this ISR-from-SRAM point is mostly not the deciding factor, because that path is not relying on a background UART refill ISR during show.
> So the specific starvation mechanism we discussed earlier applies to the async UART ISR path (present in NeoPixelBus, but not what WLED instantiates by default): NeoEsp8266UartMethod.cpp:144
> 
> Use this one-liner for the comment:
> // On ESP8266, f.close() can enter SDK flash critical sections (interrupts/cache paused), so we wait for LED output to finish first to avoid WS281x glitches, mainly on ISR-fed UART paths.

willmmiles: It's my understanding that the ESP8266 SDK can be configured whether or not to mask interrupts during flash IO, as a way of managing the fact that the flash unit is shared with the instruction cache.  If you have non-fully-IRAM ISRs, then hitting a non-cached function will require a flash read, which will then crash if the flash unit is in use.  If all your ISRs are put in static IRAM (as the Arduino core does), then it doesn't need to mask interrupts.   As far as I'm aware, the platformio Arduino core build is configured for IRAM ISRs, so flash IO doesn't mask.  (We'd be in big trouble doing soft-PWM if it did!)

That said, Claude is correct that we are currently using synchronous UART ops, so there are no ISRs involved anyways - any flash IO has to wait until the main process gets around to it regardless.

If it was up to me, I wouldn't mention ESP8266 at all, as none of our current code paths are actually vulnerable.  Adding comments about inapplicable conditions will only serve to mislead future developers (and AIs) -- we have enough problems as it is with irrelevant or outdated information propagating about the hardware and drivers.

That said: I'd still accept the PR in general, in the context of being a dirty hack workaround for the C3 we'll want to revert as soon as we can get properly stable output drivers.

softhack007: 👍 I fully agree

wled00/wled_server.cpp
```diff
@@ -380,12 +387,18 @@ void initServer()
 
   server.on(_update, HTTP_POST, [](AsyncWebServerRequest *request){
     if (request->_tempObject) {
+      // prevent random LEDs flashing during OTA
```
softhack007: i'll revert this one - suspend/resume is already handled in ota_update.cpp

wled00/file.cpp
softhack007, line 452: @DedeHai question: are presets.json file requests still coinciding with flicker on the -C3? Like full WebUI refresh (shift-F5) or frequent polling by HA? 

If yes, then i could also add ``strip.waitForLEDs(15);`` here 🤔 

Could you test if waiting before ``request->send`` makes a difference on C3? Maybe we also need ``strip.suspend()`` before the send call.

DedeHai: yes, UI access is still an issue. In my experience ANY file system access causes a glitch.

softhack007: I'm still wondering if ``strip.waitForLEDs(15);`` would be enough here - too many strip.suspend() / strip.resume() calls increase the risk of effects stuttering 🤔

willmmiles: When RMT is being used as the output driver on the C3, the FreeRTOS scheduler switch to the LwIP/AsyncTCP tasks can be enough interrupt latency to cause glitching.  By the time any of our code runs it can already be too late.  Flash IO makes it worse, but deferring it at this level won't cure the problem.

DedeHai: the cure is high priority interrupts. In my driver even using 4 blocks of RMT memory leads to frequent glitches when loading presets or other large files. No such glitches using parallel SPI out as that is DMA driven (even works at DMA ISR priority set to level1). So proper cure in sight but for now, lets mitigate what we can.

wled00/FX_fcn.cpp
```diff
@@ -1420,7 +1420,7 @@ void WS2812FX::service() {
   }
   _virtualSegmentLength = 0;
   _isServicing = false;
-  _triggered = false;
+  if (!_suspend) _triggered = false; // avoid losing "trigger" events if suspend requested during effect service()
```
DedeHai, line 1422: this should also be in 16.0. 
In fact, service() still has some left overs in 16.0 from before I removed the `return FRAMETIME;`, at least I think some of the conditionals can be removed there.

wled00/json.cpp
```diff
@@ -771,6 +771,13 @@ void serializeInfo(JsonObject root)
   root[F("lwip")] = LWIP_VERSION_MAJOR;
 #endif
 
+  bool haveSuspended = false;
+#if defined(WLED_USE_SHARED_RMT) || defined(__riscv) || !defined(ARDUINO_ARCH_ESP32)
+  // calling ESP.getFreeHeap() during led update causes glitches on C3 and possibly on 8266, too
+  if (!strip.isSuspended()) { strip.suspend(); haveSuspended = true; } // prevent that a new strip.show() starts after waiting
```
DedeHai: since this waiting is on single core MCU's: is it even possible that service() can continue to run? it can only do so if it is in a seperate task that has higher priority, which I don't think it is (?).

softhack007: Well, you have a point  here - freeRTOS scheduling is pre-emptitive, but not fair. Task switching happens each 1ms, however the task with highest priority will usually "win" , unless its blocked or waiting.

Let's say the probability is low that a lower-prio task can progress while a higher-prio one is running. But it's not completely excluded.

softhack007: On second thought - ESP.getFreeHeap() does not take much time, feels like overkill to suspend effects for this.

wled00/ota_update.cpp
```diff
@@ -137,6 +137,7 @@ static bool beginOTA(AsyncWebServerRequest *request, UpdateContext* context)
   UsermodManager::onUpdateBegin(true); // notify usermods that update is about to begin (some may require task de-init)
   
   strip.suspend();
+  strip.waitForLEDs(25);  // wait max 25 ms for LED transmissions to finish
   strip.resetSegments();  // free as much memory as you can
```
softhack007: @DedeHai slightly off-topic: in ``main`` the order is different - backupConfig() runs before strip.resetSegments(). This provides a small "grace period" before all segments are deleted. Do you remember if there is a reason why it is done in a different order in 0_15_x ?

wled00/FX_fcn.cpp
softhack007: 👍 good point, I can take a look at 16.0 ``service()`` today.

wled00/ota_update.cpp
DedeHai: I dont recall paying attention to that - so no particular reason.

softhack007: ok, then I'll change the order so its the same as in ``main``.

PR #5438, opened 2026-03-22
Fix DNS hangs on ESP32

willmmiles:
Discard IPv6 RA packets, which cause LwIP to overwrite the IPv4 DNS servers, resulting in DNS queries hanging up because there are no accessible servers.

Original issue discussed in https://github.com/wled/WLED/pull/5332#issuecomment-4097691858 .   The detect-and-recover method did not work well because there is no event notification on processing unsolicited packets.

## Summary by CodeRabbit

* **New Features**
  * ESP32 devices with IPv6 support now block unwanted IPv6 router advertisements during startup, improving network stability and IPv6 connectivity management on compatible hardware.

### DIFF
--- a/wled00/network.cpp
+++ b/wled00/network.cpp
@@ -374,18 +374,17 @@ bool isWiFiConfigured() {
 // References:
 // https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/lwip.html (see the "Note" block under "Adapted APIs" -- though it very much undersells the problem.)
 // https://github.com/espressif/arduino-esp32/discussions/9988 - links to older discussions
-static u8_t blockRouterAdvertisements(void* arg, struct raw_pcb* pcb,
-                     struct pbuf* p, const ip_addr_t* addr) {
-    // ICMPv6 type is the first byte of the payload, so we skip the header
-    if (p->len > 0 && (pbuf_get_at(p, sizeof(struct ip6_hdr)) == ICMP6_TYPE_RA)) {
-        return 1; // claim the packet — lwIP will not pass it further
-    }
-    return 0; // not consumed, pass it on
+static u8_t blockRouterAdvertisements(void* arg, struct raw_pcb* pcb, struct pbuf* p, const ip_addr_t* addr) {
+  // ICMPv6 type is the first byte of the payload, so we skip the header
+  if (p->len > 0 && (pbuf_get_at(p, sizeof(struct ip6_hdr)) == ICMP6_TYPE_RA)) {
+    return 1; // claim the packet — lwIP will not pass it further
+  }
+  return 0; // not consumed, pass it on
 }
 
 void installIPv6RABlocker() {
-    static struct raw_pcb* ra_blocker = raw_new_ip_type(IPADDR_TYPE_V6, IP6_NEXTH_ICMP6);
-    raw_recv(ra_blocker, blockRouterAdvertisements, NULL);
+  struct raw_pcb* ra_blocker = raw_new_ip_type(IPADDR_TYPE_V6, IP6_NEXTH_ICMP6);
+  raw_recv(ra_blocker, blockRouterAdvertisements, NULL);
 }
 #endif
 


wled00/network.cpp
```diff
@@ -360,6 +360,35 @@ bool isWiFiConfigured() {
...
+    if (p->len > 0 && (pbuf_get_at(p, sizeof(struct ip6_hdr)) == ICMP6_TYPE_RA)) {
+        return 1; // claim the packet — lwIP will not pass it further
+    }
+    return 0; // not consumed, pass it on
+}
+
+void installIPv6RABlocker() {
+    static struct raw_pcb* ra_blocker = raw_new_ip_type(IPADDR_TYPE_V6, IP6_NEXTH_ICMP6);
+    raw_recv(ra_blocker, blockRouterAdvertisements, NULL);
+}
```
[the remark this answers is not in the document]
willmmiles: This is called once during program setup.  If we're out of heap then, nothing will ever work.

wled00/network.cpp
```diff
@@ -360,6 +360,34 @@ bool isWiFiConfigured() {
...
+// This may have been improved in IDF v5 -- Espressif has added a feature to store DNS servers
+// on a per interface basis in their LwIP fork.
+//
+// References:
+// https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/lwip.html (see the "Note" block under "Adapted APIs" -- though it very much undersells the problem.)
+// https://github.com/espressif/arduino-esp32/discussions/9988 - links to older discussions
+static u8_t blockRouterAdvertisements(void* arg, struct raw_pcb* pcb, struct pbuf* p, const ip_addr_t* addr) {
+  // ICMPv6 type is the first byte of the payload, so we skip the header
+  if (p->len > 0 && (pbuf_get_at(p, sizeof(struct ip6_hdr)) == ICMP6_TYPE_RA)) {
+    return 1; // claim the packet — lwIP will not pass it further
```
softhack007, line 380: it might be interesting to have a DEBUG_PRINT here. Both for debugging and for testing.

willmmiles: I tried that during testing and it didn't work well - this is called directly from the LwIP task on the network core.  I ended up pipelining the test data to the main thread for printing.  (The exploration code was quite messy, with a bunch of atomics and static packet buffers, not really appropriate for even most debug builds.)

softhack007: ah ok, understood. Then let's not add DEBUG_PRINT here.

netmindz:
Is this just in V4 ?
Should we be backporting to 0-15 ?

OberfragGER:
Okay.
Im using a Wemos D1 Mini Pro and i have reboots with all current firmwares with my new AP from Ubiquity. Im using the U7 Pro. The UK Ultra works fine.
With FritzBox Wifi it runs fine.

With this fix, the network connection is still up but i got a 503 Error. It seems that i run out of RAM. Ping is working the whole time but no web frontend. I have tested the WLED_16.0.0-alpha_ESP02.bin. The compat.bin didnt work on that device.

When i switch back to the old wifi - everything works fine.
Free heap goes from 7kb to 17kb.

Changing Force 802.11g mode (ESP8266 only) or Disable WiFi sleep doesnt help with the problem.

Tested a NodeMCU. Same problem.

softhack007:
@OberfragGER can you test with an esp32 based board? 

This fix does not change anything on esp8266 (esp01, esp02, esp12, nodemcu, etc), its only for esp32 (classic, S3, S2, C3) with the "V4" framework.

OberfragGER:
> @OberfragGER can you test with an esp32 based board?
> 
> This fix does not change anything on esp8266 (esp01, esp02, esp12, nodemcu, etc), its only for esp32 (classic, S3, S2, C3) with the "V4" framework.

Im not sure if i have any ESP32 V4. Is that one?
grafik

This Device is working fine. No disconnects or something like this. Everything is as fine as expected.

softhack007:
I don't have any DNS problems on my local wifi - but I can say that nothing is "obviously broken" in a few tests I did on one of my S3 boards after uploading this PR.

willmmiles:
> Is this just in V4 ? Should we be backporting to 0-15 ?

Honestly, I don't know - depends on how much IPv6 integration was in the v3 core.  If it's a problem there, it's been broken for a very long time.  I'm inclined to focus more on shipping 16.0 for now.

Issue #4716, opened 2025-06-09
H-bridge on ESP8266


bsvdoom:
As I see the H-bridge option is available on the ESP32, however not on the ESP8266.
Long time ago I have made a proof-of-concept replacing the CCT with H_BRIDGE in the following issue:

https://github.com/wled/WLED/pull/2987/files

Can someone implement this, so we can use this feature on the ESP8266?
Thanks!

DedeHai:
ESP8266 may not support it. @willmmiles is the timing of phase shifts precise enough?

bsvdoom:
I was maybe not clear.
My code is for driving an external H-bridge which has A and B inputs.
With it we can drive simple charlieplexed Christmas/outdoor lights just fine.

I do not really know how the WLED implemented H-bridge driver works on the ESP32, I just assumed it works like how I described it above.

willmmiles:
The phase management design doesn't *guarantee* safety for H-bridge usage; it was intended as a low effort mimicry of the ESP32 design for reducing peak load on ARGB setups.  I did a fair amount of testing to reduce the chances of crossover behaviour, and I think it performed pretty well, but I wouldn't stake my life on it.

The key limitation is that the PWM engine still processes each output individually, so they can be subject to different interrupt latencies.  A purpose-built engine with native internal phase relationship grouping could offer stronger guarantees.

DedeHai:
Thanks for confirming. I do not think its worth the effort given that it already works on ESP32 (including S2, S3, C3)

bsvdoom:
Now I understand the concerns, and the hardware limitations, and the unworthiness of implementing software prevention of shoot-trough.  Closing.

PR #5442, opened 2026-03-23
Experimental: H-bridge safe mode for TYPE_ANALOG_2CH (cheap hardware?)

bsvdoom:
Add H-bridge compatible mode for TYPE_ANALOG_2CH_HBRIDGE 
(mutually exclusive PWM outputs)

### Summary

This PR modifies the behavior of TYPE_ANALOG_2CH in WLED to better support 2-wire CCT LED strips driven via an H-bridge (reverse polarity).

Instead of driving both channels simultaneously (as in standard WW/CW blending), this change ensures that only one channel is active at a time, based on the CCT value.

### Motivation

When using an H-bridge for polarity-based CCT control:

Driving both outputs simultaneously can lead to short circuits (shoot-through) inside the H-bridge
The current implementation (Bus::calculateCCT) assumes independent LED channels, which is unsafe for this use case

### This PR introduces a safer mapping:

CCT range 0–127 → Channel A active
CCT range 128–255 → Channel B active
Never both at the same time
Implementation

### Adds the new TYPE_ANALOG_2CH_HBRIDGE handling with:

case TYPE_ANALOG_2CH_HBRIDGE:
{
  uint8_t level = (Bus::_cct < 0 || Bus::_cct > 255) ? 127 : Bus::_cct;
  uint8_t brightness = w;

  if (level <= 127) {
    _data[0] = (map(level, 0, 127, 255, 0) * brightness) / 255;
    _data[1] = 0;
  } else {
    _data[0] = 0;
    _data[1] = (map(level, 128, 255, 0, 255) * brightness) / 255;
  }
}
break;

### Behavior
Ensures mutually exclusive outputs
Maps CCT slider directly to polarity direction
Preserves brightness via w scaling

### Current Status  (Needs Review)
This implementation works in principle, but is not perfect:

Output switching is not hardware-synchronized
GPIO/PWM updates happen sequentially
There may still be very short overlap periods (µs range) between channels
Behavior near the midpoint (127/128) may not be perfectly smooth

Because of this, I consider this:
Experimental / partial solution

### Request for Review 
I would really appreciate feedback from maintainers or contributors on:

Whether this approach fits WLED’s design philosophy
If there is a better way to integrate H-bridge support
Possibility of:
synchronized PWM updates
using ESP32 LEDC features for safer switching
or introducing a dedicated “H-bridge mode” instead of modifying default behavior

### Safety Notes 
This change reduces the risk of shoot-through by avoiding simultaneous activation of both outputs in software
However, perfect synchronization between GPIO updates is not guaranteed
On ESP32, PWM channels are updated sequentially, which may still introduce very short overlap periods (µs range)

### Users should still:

Prefer H-bridge drivers with built-in shoot-through protection / dead-time
Avoid high-power configurations without proper hardware safeguards

### Scope / Limitations
Applies only to TYPE_ANALOG_2CH_HBRIDGE
Intended for H-bridge-driven 2-wire CCT setups
Not suitable for motor control or general bidirectional loads

### Notes
Ignores Bus::calculateCCT() intentionally
Uses Bus::_cct directly for deterministic channel selection

This aligns with the existing warning in LED settings regarding H-bridge usage and enforces safer behavior at the firmware level.

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted warm/cold white channel mapping for 2-channel analog LED configurations. Brightness levels now independently control warm and cold channels for improved lighting behavior.

### DIFF
--- a/wled00/bus_manager.cpp
+++ b/wled00/bus_manager.cpp
@@ -532,7 +532,7 @@ void BusPwm::show() {
     unsigned duty = (_data[i] * pwmBri) / 255;
     unsigned deadTime = 0;
 
-    if ((_type == TYPE_ANALOG_2CH) && Bus::_cctBlend == 0) {
+    if (_type == TYPE_ANALOG_2CH && Bus::_cctBlend == 0) {
       // add dead time between signals (when using dithering, two full 8bit pulses are required)
       deadTime = (1+dithering) << bitShift;
       // we only need to take care of shortening the signal at (almost) full brightness otherwise pulses may overlap


DedeHai:
did you model this off the current main implementation where I already had implemented this?

softhack007:
@bsvdoom what's the reason to make this PR against our ``V5-C6`` branch? Maybe i overlooked something, but I did not find special code that would only run on new MCU's like C6, C5, or P4 👀 

btw, analog output currently is totally broken in the V5 branch ... not sure how that works for you?

wled00/bus_manager.cpp
```diff
@@ -454,6 +454,19 @@ void BusPwm::setPixelColor(unsigned pix, uint32_t c) {
         Bus::calculateCCT(c, _data[0], _data[1]);
       }
       break;
+    case TYPE_ANALOG_2CH_HBRIDGE: //warm white + cold white, relies on auto white calculation, uses H-bridge for better dimming performance at low brightness
+    {
+      uint8_t level = (Bus::_cct < 0 || Bus::_cct > 255) ? 127 : Bus::_cct;
+      uint8_t brightness = w;
+
+      if (level <= 127) {
+        _data[0] = (map(level, 0, 127, 255, 0) * brightness) / 255;
```
softhack007: The use of ``map()`` is not ideal here, because it leads to jumps in the final brightness result that can be avoided:

````c++
   if (level <= 127) { 
        data[0] = (127-level) * brightness) / 127;   // ramp down, brightest at 0
[...]
   } else {
        data[1] = ((level-128) * brightness) / 127;  // ramp up, brightest at 255
[...]
   }
````

EDIT:  I've initially overlooked the direction of your "map". I've corrected my proposal accordingly.

File: wled00/bus_manager.cpp [hunk 2]
```diff
@@ -454,6 +454,19 @@ void BusPwm::setPixelColor(unsigned pix, uint32_t c) {
         Bus::calculateCCT(c, _data[0], _data[1]);
       }
       break;
+    case TYPE_ANALOG_2CH_HBRIDGE: //warm white + cold white, relies on auto white calculation, uses H-bridge for better dimming performance at low brightness
+    {
+      uint8_t level = (Bus::_cct < 0 || Bus::_cct > 255) ? 127 : Bus::_cct;
```
softhack007:
````c++
int16_t Bus::_cct = -1;     // -1 means use approximateKelvinFromRGB(), 0-255 is standard, >1900 use colorBalanceFromKelvin()
````

The ``white`` from auto-white calculation is in ``w``; ``Bus::_cct`` has the color temperature, either in kelvin or directly from the "cct" slider.

**Comment by softhack007** [comment]
> I would really appreciate feedback from maintainers or contributors on:
> 
> Whether this approach fits WLED’s design philosophy

To be honest with you, I don't think the code that I see is usable as it is, sorry. It doesn't support the same CCT modes as the normal "2channel WW+CW" mode, however its not clear to me if this can be achieved at all with a H-Bridge; @DedeHai might be able to better comment on this aspect.

In general, adding a new mode to the switch-case is OK, just that the mode number must be aligned with the range reserved for PWM output (see comments from coderabbitAI).

File: wled00/bus_manager.cpp [hunk 3]
```diff
@@ -454,6 +454,19 @@ void BusPwm::setPixelColor(unsigned pix, uint32_t c) {
...
+      uint8_t brightness = w;
+
+      if (level <= 127) {
+        _data[0] = (map(level, 0, 127, 255, 0) * brightness) / 255;
+        _data[1] = 0;
+      } else {
+        _data[0] = 0;
+        _data[1] = (map(level, 128, 255, 0, 255) * brightness) / 255;
+      }
```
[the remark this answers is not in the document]
softhack007: @coderabbitAI you've overlooked the direction of the "map()" blending: WW brightest at level=0, CW brightest at level=255. The "dead zone" seems to be intentional, because WW and CW are both driven to 0 in the middle ranges.

bsvdoom:
> @bsvdoom what's the reason to make this PR against our `V5-C6` branch? Maybe i overlooked something, but I did not find special code that would only run on new MCU's like C6, C5, or P4 👀
> 
> btw, analog output currently is totally broken in the V5 branch ... not sure how that works for you?

No, I just happened to be using the XIAO C6 for my project, that's why I made it against V5-C6, no special code here.

I am not familiar with WLED code in general, that's why i could not add this as a new mode, just improvised the new LED_TYPE from the current code.

Btw it works just fine if i replace the old TYPE_ANALOG_2CH with my new TYPE_ANALOG_2CH_HBRIDGE case contents.

I just finished the project, and it drives my custom made h-bridge (witch lights up the 2-pin charlieplexed Christmas lights) as expected.

I just use this setup in a couple of projects, and it would be nice if someone could help me implement this, so i don't have to custom build modify the code every time. (And it could be helpful to others too)

I also used this method to drive the same circuit on esp8266, but [that PR](https://github.com/wled/WLED/issues/4716) was declined for technical reasons.

bsvdoom:
> did you model this off the current main implementation where I already had implemented this?

No, I did not see that. Can you provide me a link?

bsvdoom:
> > I would really appreciate feedback from maintainers or contributors on:
> > Whether this approach fits WLED’s design philosophy
> 
> To be honest with you, I don't think the code that I see is usable as it is, sorry. It doesn't support the same CCT modes as the normal "2channel WW+CW" mode, however its not clear to me if this can be achieved at all with a H-Bridge; @DedeHai might be able to better comment on this aspect.
> 
> In general, adding a new mode to the switch-case is OK, just that the mode number must be aligned with the range reserved for PWM output (see comments from coderabbitAI).

I understand, this was just my way to solve this problem, and I am more than happy to help if someone can show me a way to achieve this in a proper way.
OR should I just open a feature request and link this PR in?

wled00/bus_manager.cpp
```diff
@@ -525,7 +539,7 @@ void BusPwm::show() {
     unsigned duty = (_data[i] * pwmBri) / 255;
     unsigned deadTime = 0;
 
-    if (_type == TYPE_ANALOG_2CH && Bus::_cctBlend == 0) {
+    if ((_type == TYPE_ANALOG_2CH || _type == TYPE_ANALOG_2CH_HBRIDGE) && Bus::_cctBlend == 0) {
```
DedeHai: this is what I meant, you add a new type that is already supported by simply setting cctblend=0.

DedeHai:
There is really no need to add a new type as h-bridge is already supported.

wled00/bus_manager.cpp
```diff
@@ -454,6 +454,19 @@ void BusPwm::setPixelColor(unsigned pix, uint32_t c) {
...
+
+      if (level <= 127) {
+        _data[0] = (map(level, 0, 127, 255, 0) * brightness) / 255;
+        _data[1] = 0;
+      } else {
+        _data[0] = 0;
+        _data[1] = (map(level, 128, 255, 0, 255) * brightness) / 255;
+      }
+    }
     case TYPE_ANALOG_5CH: //RGB + warm white + cold white
```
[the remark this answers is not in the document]
softhack007: @bsvdoom this is clearly a bug, i'd suggest you fix it. 

The ``break;`` before the next ``case`` is missing, which means that _both_ cases will be executed.

wled00/bus_manager.cpp
bsvdoom: Regardless of setting cctblend=0, the signal on the tow outputs (A and B) are not a complimentary inverted signal that could drive a H-bridge, the truth table is not satisfied.
kép

DedeHai: they are complementary, I spent quite a few hours writing that code. If they are not complementary for the C6, changing setpixelcolor() will not make it so, you need to implement/change the driver.

softhack007: > Regardless of setting cctblend=0

@bsvdoom did you test the settings suggested by @DedeHai? Did you measure (logic analyzer needed) how it behaves with respect to avoiding shot-through?

@DedeHai I'd say lets short the discussion - my understanding is
* PWM output on C6 (and C5, too) needs a complete re-write due to major changes in esp-idf. Building anything on top of the broken "V5-C6" analog driver code is like building a house in quicksand - wasted effort.
* I'm not sure about the specifics of H-Bridges, but the code in this PR just changes the color temperature slider into a left-or-right slider for two LEDs.
* The output will be BLACK in the middle values, Left = 100% WW, right = 100% CW.
* Advanced white handling is not supported.

I don't see added value for WLED users. Especially when we already have a working solution for H-Bridge in ``main``.

DedeHai: agreed, as with all C6 / V5 code, it should be done properly to avoid a patchwork of fixes that then have to be re-adjusted later. By properly I mean write a driver that does the same as current main.

softhack007:
Looking back at the discussion flow  ... I still get the feeling that multiple discussion paths were mixed together, and that makes it hard for us (as maintainers) to judge added value. It also does not feel satisfying to me personally 😉 

Maybe lets put source code quality aspects aside for the moment. If we align first on common reference behavior and measurable evidence, that might help to make a decision that is maintainable and ensures added value for users.

From maintainers perspective, I would like to separate overlapping discussions into a clear sequence:

0. Baseline
    * @bsvdoom Thanks for sharing your work and for testing it on real hardware, that is always valuable.
    * Existing H-bridge support in ``main`` stays the reference behavior, we should not redefine CCT semantics before confirming a real gap.

1. feature parity check - `main` vs this PR:
    Let us verify whether PWM output (latest source code in ``main``) matches the electrical needs. Then If there is a mismatch, that is first a driver problem, not a new mode problem. 

2. Evidence for decision
    @bsvdoom We need a small, reproducible test set:
    * board + exact settings,
    * signal captures for CCT 0 / 63 / 127 / 128 / 255 at low and high brightness (logic analyzer would be perfect),
    * expected vs observed A/B truth table, and dead-time/overlap observations from measurements.

3. Decision path
    * If this PR code behavior matches ``main`` (baseline) or we can't confirm a real problem, we close this PR.
    * If behavior does not match baseline, we focus on a possible driver issue first (might be better to do it in an issue ticket instead of a PR).
    * Only after that, we can discuss whether additional mode behavior provides broader user value.


Edit: The ``V5-C6`` branch is behind ``main`` by about 5-6 weeks. This means any decision evidence should come from comparing with ``main``, maybe on a C3 board.

DedeHai:
image
current main, on C3. PWM CCT with blending set to 0.

bsvdoom:
I removed new type as requested, and updated the code to work as I originally desired.
Now brightness controls the H-bridge (instead of color temperature (so now ct slider does nothing)), so the PWM animations can run properly on the charlieplexed leds.

Maybe I overlooked somewhere, but I could not achieve such a behavior in WLED, without modifying the code.
So. Can this behavior be implemented in some way to `main`?

@softhack007 Since the signals are PWM, a logic analyzer won't do. And I do not have a scope.
I can only make the requested measurements with two multimeters if desired.

I can make a new PR aginst `main` if desired, as the `TYPE_ANALOG_2CH` case is the same in `V5-C6` and `main`.

Thank you for your attention.

DedeHai:
> Maybe I overlooked somewhere, but I could not achieve such a behavior in WLED, without modifying the code. So. Can this behavior be implemented in some way to `main`?
> 
> @softhack007 Since the signals are PWM, a logic analyzer won't do. And I do not have a scope. I can only make the requested measurements with two multimeters if desired.

which ESP does not exhibit this behavior and how do you know without a scope?

softhack007:
> @softhack007 Since the signals are PWM, a logic analyzer won't do.

Why? I have one of these cheap 8-channel logic analyzers here, and it's no problem to measure two output signals in parallel at high frequency (even 1MHz is no problem). The logic analyzer is definitely fast enough to capture two PWM output signals (digital signal) at the neccessarily time resolution. You don't need to attach you XMas lights, measuring the esp32 output pins is sufficient. 

Edit: @bsvdoom just to be sure - you know how PWM works, right? It's actually a digital on-off signal at high frequency, nothing analogue. To control a H-Bridge, the only thing to do is make sure that two PWM outputs (channels) are interlocked so they are never "digital high" at the same point in time.

softhack007:
> I can make a new PR aginst `main` if desired, as the `TYPE_ANALOG_2CH` case is the same in `V5-C6` and `main`.
> Thank you for your attention.

I am not sure if we need a new PR - we are still at step 1 "feature parity check", to understand if there is a problem / gap in the existing behaviour of WLED.

bsvdoom:
@softhack007 Yes, my bad. 
The measurements with this PR code in `V5-C6` with the `Seeed Studio XIAO ESP32C6` at brightness:

0:
Képernyőkép 2026-03-27 13-17-46

~63:
Képernyőkép 2026-03-27 13-18-19

127:
Képernyőkép 2026-03-27 13-18-43

~192:
Képernyőkép 2026-03-27 13-18-54

255:
Képernyőkép 2026-03-27 13-19-01


This is the expected/desired behavior for my x-mas lights.

DedeHai:
The correct way to do it is to implement the driver as it is in main. This is just a hack so closing.

bsvdoom:
@DedeHai Thanks for the clarification — that makes sense.

I realize this implementation is more of a workaround than a proper integration with the current driver architecture. My goal was mainly to explore a way to support H-bridge–driven 2-wire setups without having to maintain a custom build.

I’m not very familiar with the internal driver structure in WLED yet, so I wasn’t sure how to implement this in the “correct” way.

Would you recommend opening a feature request for this use case, or is there already a proper extension point where this kind of behavior could be implemented?

I’d be happy to rework this in a cleaner way if you can point me in the right direction.

DedeHai:
> I realize this implementation is more of a workaround than a proper integration with the current driver architecture. My goal was mainly to explore a way to support H-bridge–driven 2-wire setups without having to maintain a custom build.

the help is appreciated, its good explore the C6 and the way forward. 

If you want to push on: the implementation of "the correct way" is just a few lines below the ones you edited, the PWM driver is quite simple (compared to the digital drivers) but in IDF V5 I hear everything changed and its a new API.
what you need to do if you want to implement: read the datasheed (user manual) about what the LEDC driver can (and can not) do, then explor how the API is implemented and how it was implemented in V4 (i.e. in main). In the end I expect it to be a change of 5-10 lines of code but its a tough nut to crack.

DedeHai:
the whole "magic" happens in `BusPwm::show()`
https://github.com/bsvdoom/WLED/blob/3c7fbabd23013c566327041b8085807a064d44c2/wled00/bus_manager.cpp#L531

anything before that are just pre-calculations (CIE brightness formula approximation math, determining bit depth from frequency etc.) For starts a fixed frequency would be acceptable if that eases things but since the hardware is still the same and just the API changed, its just a question of finding out how to make the new V5 API do the same as the V4.

if you want to take a step back: implement the driver as an Arduino sketch as that is much quicker to compile and test and once that works, port it into WLED. Unfortunatel AI are not really helpful with driver development, sometimes they come up with a working framework but mostly they lead you down a wrong path.

PR #5443, opened 2026-03-24
service cleanup after removing return FRAMETIME

softhack007:
Clean-up and simplification of strip.service() after individual effect FRAMETIME was removed.

* Activation "_frametime"  is now the same for all effects, so effects scheduling based on individual frametime is not needed any more
* Special handling for mode_static is obsolete, because the solid effect does not have a different timing any more

## Summary by CodeRabbit

* **Refactor**
  * Unified frame/timing so effects refresh on a single global schedule for more consistent visuals.
  * Only active segments now determine when a frame is shown, reducing irregular timing.
  * Triggered updates and unlimited‑FPS still force immediate refreshes for snappier responses.
  * Early timing gate skips work when no update is due, lowering CPU use.
  * Ensures a safe fallback segment after servicing to avoid stale state.
  * Brightness changes use a capped refresh threshold to avoid excessive redraws.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1288,7 +1288,7 @@ void WS2812FX::service() {
 
     if (seg.isActive()) {
       // current segment is active -> re-run effect, and remember that show() call is necessary
-      //if (timeToShow) // removed - if we arrive here, its always showtime
+      // if we arrive here, its always showtime (timeToShow == true)
       doShow = true;
       if (!seg.freeze) { //only run effect function if not frozen
         // Effect blending


wled00/FX_fcn.cpp
```diff
@@ -1287,11 +1288,9 @@ void WS2812FX::service() {
 
     if (!seg.isActive()) continue;
 
-    // last condition ensures all solid segments are updated at the same time
-    if (nowUp > _lastServiceShow + _frametime || _triggered || (doShow && seg.mode == FX_MODE_STATIC))
-    {
+    // current segment is active -> re-run effect if _frametime has passed or update was triggered externally
+    if (timeToShow) {
```
softhack007: Maybe this line can be removed entirely - if we arrive here, running the effect is always necessary.

wled00/FX_fcn.cpp
softhack007, line 1305: @DedeHai @willmmiles any idea what is the intention behind the condition? 
It looks like 
* only run (blend) the old effect if different from the current one (ok)
* or if both "old" and "current" segment have a name, or they have different names (??? wtf ???)
--> not sure what the second part achieves?

wled00/FX_fcn.cpp
```diff
@@ -1266,17 +1266,17 @@ void WS2812FX::finalizeInit() {
...
-  if (!_triggered && (_targetFps != FPS_UNLIMITED)) {   // unlimited mode = no frametime
-    if (elapsed < _frametime) return;                   // too early for service
-  }
-
-  bool doShow = false;
 
   _isServicing = true;
-  _segment_index = 0; // current segment index for getCurrSegmentId()
+  now = nowUp + timebase; // common time base for all effects
+  bool doShow = false;    // book keeping: if at least one effect was run, we need to call show()
```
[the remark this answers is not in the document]
softhack007: I need to think about this ... actually the only change in the "needs rendering" logic is the removal of ``|| (doShow && seg.mode == FX_MODE_STATIC)`` - this did not exclude FX_MODE_STATIC, but actually included it in case that any other segment was rendered already.

The comment for doShow (“at least one effect was run”) is actually incomplete, because it really means "at least one effect and not suspended while processing segments".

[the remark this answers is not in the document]
softhack007: keeping unchanged "frozen" segments off the ``doShow`` logic is not possible, because "freeze" exactly means "do not run effect". A frozen segment can still receive new pixels (external control, or single pixel via JSON API). That's the main use case for "trigger".

wled00/FX_fcn.cpp
willmmiles: Fade blending can be special-cased by smoothly changing the effect parameters from the old to new states instead of computing twice and overlaying the results.  This cannot be applied to scrolling text / gifplayer if the name has changed, though, so the optimization is disabled on name changes.

wled00/FX_fcn.cpp
```diff
@@ -1266,17 +1266,17 @@ void WS2812FX::finalizeInit() {
...
+  if (_triggered || _targetFps == FPS_UNLIMITED) timeToShow = true; // unlimited mode = no frametime; strip.trigger() can overrule timing
 
-  bool doShow = false;
+  now = nowUp + timebase; // common time base for all effects
+  if (!timeToShow) return;                              // too early for service
+  if (_suspend || elapsed <= MIN_FRAME_DELAY) return;   // keep wifi alive - no matter if triggered or unlimited
 
   _isServicing = true;
-  _segment_index = 0; // current segment index for getCurrSegmentId()
+  bool doShow = false;    // true if ≥1 active segment was processed (and strip was not suspended mid-loop) → triggers show()
```
softhack007: possible bug in a corner case: if _triggered but we have _zero_ active segments, then the trigger call does not lead to show().

-> could be solved with ``doShow = _triggered``
-> need to investigate if "strip.trigger() but no active segments exist" is a realistic case.

wled00/FX_fcn.cpp
```diff
@@ -1287,11 +1287,10 @@ void WS2812FX::service() {
 
     if (!seg.isActive()) continue;
```
softhack007: possible bug (need to investigate):

``continue`` might be wrong here - it implies that the last statement in the loop body gets skipped (``_segment_index++``) for existing but not active segments. 
This means that segment IDs returned by ``strip.getCurrSegmentId()`` can change if one of the segment in the middle of the list changes from active to inactive. This could confuse the copy segment effect.

@willmmiles do you know if the vector class has a method like "get my index" ? This might be more reliable than counting up our own array index in the ``for(Segment &seg, ..`` loop.

wled00/FX_fcn.cpp
```diff
@@ -1266,17 +1266,17 @@ void WS2812FX::finalizeInit() {
...
 
-  bool doShow = false;
+  now = nowUp + timebase; // common time base for all effects
+  if (!timeToShow) return;                              // too early for service
+  if (_suspend || elapsed <= MIN_FRAME_DELAY) return;   // keep wifi alive - no matter if triggered or unlimited
 
   _isServicing = true;
-  _segment_index = 0; // current segment index for getCurrSegmentId()
+  bool doShow = false;    // true if ≥1 active segment was processed (and strip was not suspended mid-loop) → triggers show()
+  _segment_index = 0;     // current segment index for getCurrSegmentId()
```
softhack007: @coderabbitAI currently the loop is counting up  `_segment_index` while iterating over the segments vector. Please check if the vector class provides a method that returns the real "array index" of the current vector element? Something like ``_segment_index = seg.getOwnArrayIndex()`` would be more reliable here, if it exists.

softhack007: I would prefer to not rely on pointer arithmetic here. Maybe moving "_segment_index ++" out of the if (seg.isActive()) body is the better solution.

@coderabbitAI Please analyze all places in the codebase where ``strip.getCurrSegmentId()`` is used, to confirm if the result is actually only used to index the segments vector.

DedeHai: why overcomplicate it? this should work:
```c++
for (int i = 0; i < _segments.size(); i++) {
    Segment &seg = _segments[i];
   _segment_index = i;
    // rest of code as it was
}
```

softhack007: @coderabbitAI are there any other call sites of getCurrSegmentId() where this behavioral change might cause new regressions?

softhack007: > why overcomplicate it? this should work:

good point, I think that's how the code always looked in previous releases. OK lets's do it 😃

wled00/FX_fcn.cpp
softhack007: edit: fixed. We go back to the older code where the for loop simply iterated over the index, avoiding any guesses of what the real array index might be.

softhack007:
@DedeHai @willmmiles @netmindz I'm happy now with the cleaned-up result, but not sure if this should go into 16.0 or into 16.1. The ``service()`` function is central for effects, and its still kind of a hack in how it adresses corner cases. 

Maybe better to keep the code as-is in 16.0 (bugfixes already backported to main) and wait until 16.1 with the full cleanup 🤔 or maybe by "quality manager mindset" is too cautious? What do you think?

wled00/FX_fcn.cpp
DedeHai: Its more old-school. Personally I do not like the "implicit" things in the newer C++ standards - some call it "easier to read" but I would call it the opposite. Just like python with its "iterate through anything automatically" obfuscation. I'd like to keep my off-by-1 explicit ;)

wled00/FX_fcn.cpp
softhack007: solved - the cleanest solution is to always run show() when a strip.trigger() was received.
strip.show() will handle the "empty strip" case properly.

DedeHai:
IMHO 16.0 will be a leap forward from 0.15, better pack whatever changes we have into the beta and get it tested instead of waiting and breaking things again in 16.1

wled00/FX_fcn.cpp
softhack007: ok i'll leave that part as-is for now. Its still somewhat "hacky" because I have the feeling that what we really want to know here has little to do with segment naming. But it works, and I don't see a better way right now.

DedeHai:
@coderabbitai please do a thorough check if there are any bugs or missing special case handling in service(). Also compile a list of tests to be performed on real hardware to check if all cases are handled properly.

DedeHai:
I ran most of the tests suggested by the rabbit, except:
- API calls
- explicit check of "suspend"
- CCT transitions (doubt there is anything noteworthy)
- the last two  points, i.e. platform specifics and edge cases.

I have found one bug, but that is probably not related to this PR:
to reproduce: set transition to 5s, choose any transition type except fade, fairy dust for example. power leds off and while in transition power on again, then off again before the transition completes -> now the on off transition is broken and wont do fairy dust.

other than that, all tests passed, including live streaming (used the video tool over WS at 2FPS starget -> still runs at 50FPS).
I found some odd things we may want to look at (or just leave it be, they are edge cases): when setting FPS very low, like 1 or 2 fps, brightness change still causes an FX speedup (due to trigger probably). When changing the color, it respects the FPS but there is a weird "double fps" just before a frame update (tested with akemi on palette 0)

softhack007:
Thanks for the review and tests 🥇 

> brightness change still causes an FX speedup (due to trigger probably). When changing the color, it respects the FPS but there is a weird "double fps" just before a frame update (tested with akemi on palette 0)

Looks like both FPS jumps are due to brightness handling, and especially it comes from "responsiveness improvement" hacks from the times when the SOLID effect was running at 3fps fixed, causing brightness/color changes to stutter

https://github.com/wled/WLED/blob/d8cb20a9e70be8721d5ac673b1ad90b55e2b59e1/wled00/FX_fcn.cpp#L1755-L1758

MIN_SHOW_DELAY is 3ms, so effectively the strip would jump into "unlimited FPS" mode whenever brightness changes.

--> solution: ``MIN_SHOW_DELAY`` --> ``min(FRAMETIME, uint16_t(FRAMETIME_FIXED))`` will still improve responsiveness, without making all effects run wild

https://github.com/wled/WLED/blob/d8cb20a9e70be8721d5ac673b1ad90b55e2b59e1/wled00/led.cpp#L76-L82

--> solution: this is ancient legacy code, don't touch 😉 


@coderabbitai I intend to implement the first solution - please check if this would cause new side-effects on other features like transitions, blending, cross-fading, and palette transitions.

softhack007:
> I have found one bug, but that is probably not related to this PR:
> to reproduce: set transition to 5s, choose any transition type except fade, fairy dust for example. power leds off and while in transition power on again, then off again before the transition completes -> now the on off transition is broken and wont do fairy dust.

@coderabbitai please analyze the chain of events and state changes that lead to this problem. Do you see a simple and minimal-impact fix?

DedeHai:
> Looks like both FPS jumps are due to brightness handling, and especially it comes from "responsiveness improvement" hacks from the times when the SOLID effect was running at 3fps fixed, causing brightness/color changes to stutter

the hack was to not use frametime in transitions, updating brightness through "trigger" is correct - otherwise the brightness of frozen segments is unresponsive. The "issue" I pointed at really only matters if you go below 10FPS deliberately, otherwise its not really noticeable. If it is, we will hear about it ;)

DedeHai:
RE: turn off transition issue - I would be cautios with any changes to transition code as it handles many edge cases and changing one thing can break another. I can check again on when this happens (i.e. only on turn-off/on or also on normal transitions) and then open a seperate issue to properly track it.

softhack007:
> RE: turn off transition issue - I would be cautios with any changes to transition code as it handles many edge cases and changing one thing can break another. I can check again on when this happens (i.e. only on turn-off/on or also on normal transitions) and then open a seperate issue to properly track it.

makes sense - I think its a bigger topic and the risk to break stuff is high

netmindz:
> @DedeHai @willmmiles @netmindz I'm happy now with the cleaned-up result, but not sure if this should go into 16.0 or into 16.1. The `service()` function is central for effects, and its still kind of a hack in how it adresses corner cases.
> 
> Maybe better to keep the code as-is in 16.0 (bugfixes already backported to main) and wait until 16.1 with the full cleanup 🤔 or maybe by "quality manager mindset" is too cautious? What do you think?

I'm happy to follow @DedeHai opinion on this as I feel he is best placed to understand the risk

softhack007:
alright, then I'll merge this into 16.0

PR #5445, opened 2026-03-25
presets-reset

Satto-ship-it:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where animations would not properly restart when loading presets, ensuring segments reset to their initial state and animations begin from the start.


DedeHai:
no comment as to why and breaking change. closing.

PR #5457, opened 2026-03-27
Zigbee rgb light usermod

netmindz:
## Summary by CodeRabbit

* **New Features**
  * Added Zigbee RGB Light support for ESP32-C6 with bidirectional state synchronization to Zigbee coordinators

* **Bug Fixes**
  * Fixed WiFi softAP startup station deauthentication behavior
  * Improved upgrade report persistence on network errors

* **Documentation**
  * Added Zigbee RGB Light integration guides and Philips Hue pairing troubleshooting
  * Added integration test suite for Zigbee/Hue compatibility validation

* **Chores**
  * Updated build configuration and partition tables for Zigbee support

PR #5466, opened 2026-03-30
8MB partition: 2.3MB program, 3.4MB LittleFS (for discussion)

softhack007:
* increase program partition by 132 KB (to be discussed)
* additional zigbee compatible 8MB partition file: 2.5MB program, 3MB LittleFS (to be discussed)

Actually the esp32-c5 "full debug" build is the best reason for increasing the 8MB program partition now (before massive rollout) - it barely fits into the current partitions' layout.

### V4

esp32-S3 8MB "opi" PSRAM (**tasmota** framework)
* old:  ``Flash: [======    ]  58.4% (used 1224085 bytes from 2097152 bytes)`` with 4MB LittleFS
* new: ``Flash: [=====     ]  54.9% (used 1224085 bytes from 2228224 bytes)`` with 3.75MB LittleFS

### V5

esp32-C5 8M "qspi" PSRAM (**pioarduino** framework)

* old  ``Flash: [========= ]  85.4% (used 1790459 bytes from 2097152 bytes)``
* new ``Flash: [========  ]  80.4% (used 1790459 bytes from 2228224 bytes)``

C5 full debug build (build_type = debug, -fno-lto, -D WLED_DEBUG -D CORE_DEBUG_LEVEL=5)
* old: 96% used (and > 100% before we replaced fastled ;-) )
* new: ``Flash: [========= ]  91.6% (used 2040809 bytes from 2228224 bytes)``

C5 build with -O2 (optimize for speed), no debug
* new ``Flash: [========= ]  87.4% (used 1946845 bytes from 2228224 bytes)``

### question
This is mainly a discussion input for finding a good balance between LittleFS and program size on 8MB boards.
The partition boundaries can be changed in 64KB steps. We need two OTA partitions, that's why each program increase counts 2x against LittleFS.

What do you think - Should we increase the 8MB program partition even more? Or better to stay with the previous 2MB/4MB layout, risking that some ``V5`` debug builds will not fit onto an 8MB board in the future?

### DIFF
--- a/tools/WLED_ESP32_8MB.csv
+++ b/tools/WLED_ESP32_8MB.csv
@@ -1,7 +1,7 @@
 # Name,   Type, SubType, Offset,  Size, Flags
 nvs,      data, nvs,     0x9000,  0x5000,
 otadata,  data, ota,     0xe000,  0x2000,
-app0,     app,  ota_0,   0x10000, 0x200000,
-app1,     app,  ota_1,   0x210000,0x200000,
-spiffs,   data, spiffs,  0x410000,0x3E0000,
+app0,     app,  ota_0,   0x10000, 0x220000,
+app1,     app,  ota_1,   0x230000,0x220000,
+spiffs,   data, spiffs,  0x450000,0x3A0000,
 coredump, data, coredump,,64K
--- a/tools/WLED_zigbee_8MB.csv
+++ b/tools/WLED_zigbee_8MB.csv
@@ -1,9 +1,9 @@
 # Name,     Type, SubType, Offset,  Size, Flags
 nvs,        data, nvs,     0x9000,  0x5000,
 otadata,    data, ota,     0xe000,  0x2000,
-app0,       app,  ota_0,   0x10000, 0x220000,
-app1,       app,  ota_1,   0x230000,0x220000,
-spiffs,     data, spiffs,  0x450000,0x39B000,
+app0,       app,  ota_0,   0x10000, 0x270000,
+app1,       app,  ota_1,   0x280000,0x270000,
+spiffs,     data, spiffs,  0x4F0000,0x2FB000,
 zb_storage, data, fat,     0x7EB000,0x4000,
 zb_fct,     data, fat,     0x7EF000,0x1000,
 coredump,   data, coredump,0x7F0000,0x10000,
--- a/tools/WLED_zigbee_8MB.csv
+++ b/tools/WLED_zigbee_8MB.csv
@@ -1,0 +1,9 @@
+# Name,     Type, SubType, Offset,  Size, Flags
+nvs,        data, nvs,     0x9000,  0x5000,
+otadata,    data, ota,     0xe000,  0x2000,
+app0,       app,  ota_0,   0x10000, 0x220000,
+app1,       app,  ota_1,   0x230000,0x220000,
+spiffs,     data, spiffs,  0x450000,0x39B000,
+zb_storage, data, fat,     0x7EB000,0x4000,
+zb_fct,     data, fat,     0x7EF000,0x1000,
+coredump,   data, coredump,0x7F0000,0x10000,


softhack007:
@coderabbitai please review the two .csv partition layouts modified by this PR, and check that
* no gaps exist
* they are fully compliant with espressif specifications

softhack007:
It seems that the espressif Zigbee stack has "Estimated flash increase: ~300–500 KB of additional program code."
So the zigbee partition should actually reserve ~2.5MB for program, leaving only 3MB for LittleFS. I'll update WLED_zigbee_8MB.csv accordingly.

softhack007:
@coderabbitAI please check if the modified zigbee partions file is still correct and aligned with espressif specs. Short answer is sufficient.

netmindz:
So would we set the app size to be the same for both the regular 8 and the ZigBee variant, but the littlefs smaller on the ZigBee to allow for the ZigBee partitions?

End users might be less likely to be swapping back and forth, but as Devs it with would be nice to be able to OTA between a regular and a ZigBee C6 build

DedeHai:
In general I would be careful in changing partitions for default ESP32 builds as it is a "breaking change" to existing setups in the sense that a build will not fail if it grows too large but it is unusable for OTA to any existing setup. Should that point of "need those extra 200k bytes" arrive, we need to provide a way to change partition tables OTA for existing setups. Changing the default now is asking for trouble down the line IMHO.
As for zigbee: the feature has been requested several times - and the "standard" C6 are 4M so I would expect most people to have that version (me included). If for the moment a 8M C6 with huge app partitions is required then so be it.

softhack007:
> So would we set the app size to be the same for both the regular 8 and the ZigBee variant, but the littlefs smaller on the ZigBee to allow for the ZigBee partitions?

Actually the zigbee partitions are small, around 20KB. We could even add that to the "standard" partitioning. The main driver is program size +350KB for zigbee support.


A bit of comparing / [tinkering with the zigbee PR](https://github.com/wled/WLED/pull/5457#issuecomment-4169307043):
* minimum program size for zigbee is 1.9MB (tight fit -> no OTA on 4MB boards; unless we change to use the tasmota "safeboot" system)
* with Zigbee, the "comfort zone" for development is 2.4MB to 2.5 MB program, leaving 3MB for LittleFS (settings, gifs, ...)
* without Zigbee, 2.2 MB program is still very comfortable, leaving 3.5MB for files
* Typical program size - without zigbee - is ~1.3MB (C6 tasmota) or 1.8MB (C5 pioarduino)

Edit: for development, we could even go for the "big enough for MATTER" solution => ~2MB Filesystem, 3MB program

softhack007:
> In general I would be careful in changing partitions for default ESP32 builds as it is a "breaking change" to existing setups

I agree that breaking existing setups is bad. Currently there are only very few esp32 installs with 8MB, more for esp32-S3 as this MCU was always 8MB flash except for the cheap "s3 tiny/zero" boards.

Looking at WLED history, we already "broke" program size limits at least two times
* very old versions had ~1MB for program
* that was "slightly lifted" to ~1.3MB (WLED 0.12/0.13)
* Currently our 4MB setups are 1.5MB program, or (WLED-MM partitioning) 1.75 MB program. And its a tight fit, again...

From these experiences - every late increase was a big cry-out for users - I would say its better to break stuff early, and start BIGGER on 8MB. OTA updates for old installs will continue to work, until we actually hit the limit of previous partioning (inevitable). If we have a bigger 8MB default with 16.0, it means that more first-install users will be "prepared for the future" when program size grows.

softhack007:
What do you think about this:

* We adjust the 8M default partitioning so its enough for "basic zigbee", plus some headroom for non-zigbee debug
* ~~all 8MB partitions get the extra zigbee filesystems (maybe the 16/32mb ones, too) - cheap 20KB extra~~ (maybe not, see edit2)
* A second 8M partitions.csv for developers, big enough for matter (matter+wled needs at least 2.5MB program).

--> 8MB default = 2.3 MB program, 3.4 MB files - with coredump
--> 8MB zigbee = 2.3 MB program, 3.4 MB files - with zigbee and coredump
--> 8MB dev = 3MB program, 2MB files (slightly less than the arduino "big program" 3.2MB / 1.6MB files layout) - with zigbee and coredump; big enough for matter

Is 3.4 MB for media files still OK? or too tight already?

<br/>

Edit: I guess that "2.3 MB program" will also have enough space for adding HTTPS or bluetooth (maybe not enough when combining them with zigbee, though).
Edit2: the 20KB zigbee extras might actually create broken LittleFS filesystems on existing setups - the space is cut out at the end of the LittleFS partition. Better to keep a special partitions file for ZigBee.

PR #4989, opened 2025-10-06
Add GC9A01 round TFT display usermod

srg74:
## Overview
Adds comprehensive support for GC9A01 240x240 round TFT displays with a circular UI layout optimized for round displays.

## Features
- Circular UI design optimized for 240x240 round displays
- Real-time status display: WiFi signal, power state, brightness, current effect
- Clock mode with 12/24-hour format options
- Configurable sleep/clock modes (5-300s timeout)
- Color preview buttons (FX/BG/CS) with automatic synchronization from web UI
- Effect and palette names with automatic display
- Custom startup logo
- Full web UI configuration for all settings
- Compatible with rotary encoder usermod for visual feedback

## Hardware Requirements
- ESP32 microcontroller
- GC9A01 240x240 round TFT display
- TFT_eSPI library (v2.5.43+)

## Additional Changes
This PR also includes a bug fix for the rotary encoder ALT usermod that prevents palette navigation wraparound when decrementing from index 0.

## Testing
Thoroughly tested on ESP32-D0WDQ6 with GC9A01 display. All features verified including power management, color synchronization, and rotary encoder integration.

## Files Changed
- New usermod: `usermods/usermod_v2_gc9a01_display/` (5 files)
- Bug fix: `usermods/usermod_v2_rotary_encoder_ui_ALT/usermod_v2_rotary_encoder_ui_ALT.cpp` (line 895)

## Summary by CodeRabbit

- **New Features**
  - GC9A01 240x240 round TFT support with full on-device UI: clock, WiFi/power indicators, brightness arc, effect/palette labels, overlays, encoder-driven controls, sleep/wake and backlight/timeout handling; integrates with rotary-encoder UI.

- **Documentation**
  - Comprehensive GC9A01 display usermod guide: wiring, setup, usage, build flags, configuration examples, and changelog.

- **Chores**
  - Added library manifest for the new usermod and updated ignore rules for development artifacts.

PR #5479, opened 2026-04-05
TTGO-T-Display usermod fixup

willmmiles:
Import all the latest fixes for this usermod, and upgrade it to v2.  Also includes the ability to specify the display setup header as a custom property in `platformio_override.ini`.

Replaces #4600.

- Draft because I haven't updated the documentation yet
- Realistically, there's probably a common "lcd_display" usermod that could underpin this, #4989, and possibly others (#5475?, #5653)

## Summary by CodeRabbit

* **New Features**
  * Added a new TTGO T-Display build workflow with a preconfigured PlatformIO environment.
  * Display setup is now selected automatically, with easier customization for other boards.

* **Documentation**
  * Updated setup instructions for building and flashing the TTGO T-Display usermod.
  * Added clearer guidance on what appears on the screen, power-saving behavior, and config options.

* **Bug Fixes**
  * Improved display compatibility and startup handling for TFT-based builds.
  * Added safer behavior when display-related pins conflict or the display is disabled.


softhack007:
> In file included from .pio/libdeps/usermods_esp32/TFT_eSPI/TFT_eSPI.h:101,
>                 from usermods/TTGO-T-Display/usermod.cpp:22:
> .pio/libdeps/usermods_esp32/TFT_eSPI/Processors/TFT_eSPI_ESP32.h:139:12: fatal error: SPIFFS.h: No such file or directory

I think the reason for the build failure is crazy simple: tasmota has removed SPIFFS 😢 , so the display driver lib fails to compile (even when SPIFFs is not used, the lib needs SPIFFS.h)

You can still build this usermod with the original arduino-esp32 framework from espressif.
https://github.com/wled/WLED/blob/bf948fadb5eb95eb6ff0c81604d2d3eb4d687642/platformio.ini#L305-L308


btw, same build failure in another usermods that use `TFT_eSPI`
* #4989

willmmiles:
Re [18d4a41](https://github.com/wled/WLED/pull/5479/commits/18d4a4158245489829b23c52fb4e159dad5d9380):  Arg, I had unpushed commits in this branch, now I have to play rebase games.  :(    Would you mind not doing this in the future?

> I think the reason for the build failure is crazy simple: tasmota has removed SPIFFS 😢 , so the display driver lib fails to compile (even when SPIFFs is not used, the lib needs SPIFFS.h)

Yes, I know.  There are better workarounds via the custom script.  (This branch is labeled draft for a reason!)

softhack007:
> Re [18d4a41](https://github.com/wled/WLED/pull/5479/commits/18d4a4158245489829b23c52fb4e159dad5d9380): Arg, I had unpushed commits in this branch, now I have to play rebase games. :( Would you mind not doing this in the future?

@willmmiles sorry and your're right - I should have thought for a moment about the "why DRAFT?" thing 😅 


Good to know that you see a way via custom script, i was a bit over-nervous maybe when I found out that tasmota simply does not have "SPIFFS.h", so all our TFT_eSPI based usermods will suffer from the same problem.

willmmiles:
> @willmmiles sorry and your're right - I should have thought for a moment about the "why DRAFT?" thing 😅
> 
> Good to know that you see a way via custom script, i was a bit over-nervous maybe when I found out that tasmota simply does not have "SPIFFS.h", so all our TFT_eSPI based usermods will suffer from the same problem.

Yes.  It's complicated by the fact that the TFT_eSPI library seemed to be unmaintained for a year or so there.  I think it's faced stiff competition with Espressif's well optimized `esp_lcd` library.   When V5 lands.... ;)

I've shimmed it in my WIP build with `lib_ldf_mode=deep+` to dig up an external `SPIFFS.h` (I've got a project due Friday that needed a quick fix) but that's also a terrible solution long-term.  I could fork the lib - might be easier than shimming in place.  I'm also looking at factoring out a framework-y "TFT_eSPI usermod" that can share the common bits for this library (pin management, build shims, etc.) for all of the local use cases.

willmmiles:
@coderabbitai review

PR #5492, opened 2026-04-09
Fireworks only explode in a single color

RussStringham:
I have made several enhancements to the Fireworks 1D effect. The primary enhancement is to make the exposion look more like a real firework. When a firework explodes, it explodes as a single color, not a rainbow. I have also added a new Firewworks palette that consists of the colors most commonly used for real fireworks, with their distribution approximately equal to how common they are (white and yellow are more common than in my palette, but they don't look as impressive in LEDs as they do in real life). Other changes will be described in the code diff.

## Summary by CodeRabbit

* **New Features**
  * Added a "Fireworks" gradient palette for enhanced color effects
  * Added a "Single Color" option to the exploding fireworks effect

* **Improvements**
  * Refined exploding fireworks behavior for more consistent flare and spark color handling
  * Improved palette cache validation to detect palette count changes
  * Increased built-in gradient palette count (slightly raising max custom palette capacity)

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -10,30 +10,28 @@
 # ------------------------------------------------------------------------------
 
 # CI/release binaries
-;default_envs = nodemcuv2
-;               esp8266_2m
-;               esp01_1m_full
-;               nodemcuv2_160
-;               esp8266_2m_160
-;               esp01_1m_full_160
-;               nodemcuv2_compat
-;               esp8266_2m_compat
-;               esp01_1m_full_compat
-;               esp32dev
-;               esp32dev_debug
-;               esp32_eth
-;               esp32_wrover
-;               lolin_s2_mini
-;               esp32c3dev
-;               esp32c3dev_qio
-;               esp32S3_wroom2
-;               esp32s3dev_16MB_opi
-;               esp32s3dev_8MB_opi
-;               esp32s3dev_8MB_qspi
-;               esp32s3_4M_qspi
-;               usermods
-
-default_envs = esp32dev
+default_envs = nodemcuv2
+               esp8266_2m
+               esp01_1m_full
+               nodemcuv2_160
+               esp8266_2m_160
+               esp01_1m_full_160
+               nodemcuv2_compat
+               esp8266_2m_compat
+               esp01_1m_full_compat
+               esp32dev
+               esp32dev_debug
+               esp32_eth
+               esp32_wrover
+               lolin_s2_mini
+               esp32c3dev
+               esp32c3dev_qio
+               esp32S3_wroom2
+               esp32s3dev_16MB_opi
+               esp32s3dev_8MB_opi
+               esp32s3dev_8MB_qspi
+               esp32s3_4M_qspi
+               usermods
 
 src_dir  = ./wled00
 data_dir = ./wled00/data
@@ -536,7 +534,7 @@ build_flags = ${common.build_flags} ${esp32_idf_V4.build_flags} -D WLED_RELEASE_
   -DBOARD_HAS_PSRAM -mfix-esp32-psram-cache-issue ;; Older ESP32 (rev.<3) need a PSRAM fix (increases static RAM used) https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/external-ram.html
   -D DATA_PINS=25
 lib_deps = ${esp32_idf_V4.lib_deps}
-
+  
 [env:esp32c3dev]
 extends = esp32c3
 platform = ${esp32c3.platform}
@@ -660,7 +658,7 @@ monitor_filters = esp32_exception_decoder
 
 [env:esp32s3_4M_qspi]
 ;; ESP32-S3, with 4MB FLASH and <= 4MB PSRAM (memory_type: qio_qspi)
-board = lolin_s3_mini ;; -S3 mini, 4MB flash 2MB PSRAM
+board = lolin_s3_mini ;; -S3 mini, 4MB flash 2MB PSRAM 
 platform = ${esp32s3.platform}
 platform_packages = ${esp32s3.platform_packages}
 upload_speed = 921600
--- a/platformio.ini
+++ b/platformio.ini
@@ -10,28 +10,30 @@
 # ------------------------------------------------------------------------------
 
 # CI/release binaries
-default_envs = nodemcuv2
-               esp8266_2m
-               esp01_1m_full
-               nodemcuv2_160
-               esp8266_2m_160
-               esp01_1m_full_160
-               nodemcuv2_compat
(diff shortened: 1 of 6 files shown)


wled00/FX.cpp
```diff
@@ -3763,7 +3763,7 @@ void mode_exploding_fireworks(void)
     if (SEGENV.aux0 == 0) { //init flare
       flare->pos = 0;
       flare->posX = SEGMENT.is2D() ? hw_random16(2,cols-3) : (SEGMENT.intensity > hw_random8()); // will enable random firing side on 1D
-      unsigned peakHeight = 75 + hw_random8(180); //0-255
+      unsigned peakHeight = 84 + hw_random8(146); // 33% to 90% of 256
```
RussStringham, line 3766: If the firework goes to the very top, it only explodes down, rather than in both directions. If it is too low, it mostly explodes up.

wled00/FX.cpp
```diff
@@ -3835,9 +3836,6 @@ void mode_exploding_fireworks(void)
             c = color_blend(spColor, WHITE, uint8_t((prog - 300)*5));
           } else if (prog > 45) { //fade from spark color to black
             c = color_blend(BLACK, spColor, uint8_t(prog - 45));
-            unsigned cooling = (300 - prog) >> 5;
-            c.g = qsub8(c.g, cooling);
-            c.b = qsub8(c.b, cooling * 2);
```
RussStringham, line 3840: The code comes from the original algorithm code. In that code the firework always explodes as white, then fades to yellow, before fading to black. Where the firework here can be any color. If it is already yellow it doesn't fade at all. If it is blue or green it fades completely before the fade to black even starts.

platformio.ini
```diff
@@ -10,28 +10,30 @@
 # ------------------------------------------------------------------------------
 
 # CI/release binaries
-default_envs = nodemcuv2
```
softhack007, line 13: @RussStringham please revert modifications in platformio.ini. A usermod should _never_ modify the global platformio.ini - this breaks nightly builds.

platformio.ini
```diff
@@ -10,28 +10,30 @@
...
+;               esp32c3dev
+;               esp32c3dev_qio
+;               esp32S3_wroom2
+;               esp32s3dev_16MB_opi
+;               esp32s3dev_8MB_opi
+;               esp32s3dev_8MB_qspi
+;               esp32s3_4M_qspi
+;               usermods
+
+default_envs = esp32dev
```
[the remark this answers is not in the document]
softhack007: @coderabbitAI correct. please remember that usermods are _never_ allowed to modify the main platformio.ini. This should be a RED FLAG when reviewing any usermod.

Usermod authors can create their custom buildenvs using ``platformio_override.ini``. It should be sufficient to add the usermods to ``custom_usermods`` for testing.

platformio.ini
softhack007: edit: just saw that your PR is not a usermod - sorry if I caused confusion. However the main rule still stands - don't modify platformio.ini.

wled00/FX.cpp
```diff
@@ -4976,13 +4974,13 @@ void mode_ColorClouds()
 
   // Higher values make the clouds move faster.
   const uint32_t volSpeed = 1 + SEGMENT.speed;
-
```
softhack007, line 4979: @RussStringham please revert all "whitespace only" changes. They are not necessary, and will cause problems in case we want to cherry-pick your PR from one branch to another.

wled00/FX.cpp
```diff
@@ -3835,9 +3836,6 @@ void mode_exploding_fireworks(void)
             c = color_blend(spColor, WHITE, uint8_t((prog - 300)*5));
           } else if (prog > 45) { //fade from spark color to black
             c = color_blend(BLACK, spColor, uint8_t(prog - 45));
-            unsigned cooling = (300 - prog) >> 5;
```
softhack007, line 3838: @coderabbitAI is removing this block a user-visible change, or just refactoring?

wled00/data/index.js
```diff
@@ -2812,7 +2812,7 @@ function loadPalettesData() {
 		if (lsPalData) {
 			try {
 				var d = JSON.parse(lsPalData);
-				if (d && d.vid == lastinfo.vid) {
+				if (d && d.vid == lastinfo.vid && d.pcount == lJson.length) {
```
softhack007, line 2815: @RussStringham what is the purpose of this change? is it a bugfix?

wled00/data/index.js
```diff
@@ -2824,7 +2824,8 @@ function loadPalettesData() {
 		getPalettesData(0, () => {
 			localStorage.setItem("wledPalx", JSON.stringify({
 				p: palettesData,
-				vid: lastinfo.vid
+				vid: lastinfo.vid,
```
softhack007, line 2827: @RussStringham please explain

softhack007:
@coderabbitai please remember to point out trailing-whitespace "optimizations" that were intermixed with real code changes. 

In this repo, we prefer to address "whitespace optimization" in a separate PR. This helps reviewers to easily see real changes (logic of the code). Additionally, diffs with lots of unrelated whitespace removal regularly case trouble when cherry-picking.

wled00/data/index.js
RussStringham: The new Fireworks palette didn't display the palette at the bottom of the selector, because the cache wasn't regenerated since the vid didn't change. This forced a cache update for the new palette.

DedeHai:
you'd need to make your new options a checkmark, the way it is now it is an unacceptable breaking change.
That being said: the PS version of the fireworks already does similar things and much more.

RussStringham:
> you'd need to make your new options a checkmark, the way it is now it is an unacceptable breaking change. That being said: the PS version of the fireworks already does similar things and much more.

I have added a "Single Color" checkmark option to the fireworks effect, which is off by default. When off it behaves as it always has. When checked, it uses my code.

I'm not familiar with this code base. What is the "PS version of fireworks"?

DedeHai:
> I'm not familiar with this code base. What is the "PS version of fireworks"?

the FX called fireworks that have PS in front, PS standing for particle system

wled00/data/index.js
RussStringham: See the comment for line 2815

DedeHai:
The palette needs to be updated so it is useable for other things.. I have a proposal for an improved one but need to test it.

wled00/data/index.js
DedeHai: this is actually a fix for a long standing bug.

DedeHai: this fix is actually incomplete. I will add the correct version in a seperate PR

PR #5493, opened 2026-04-10
Refactor wled-tools discover_devices for deduplication and clarity

LordMike:
I've discovered that when we have multiple network interfaces, possible on different VLANs, which see the same mDNS entries, avahi will emit multiple entries (since it also keys on interface).

This change will make awk deduplicate, by only emitting unique name/ip/port pairs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Device discovery now removes duplicate entries so each discovered device appears only once in scan results.
  * Discovery output remains compatible with existing parsing — format and emitted triplets are preserved to avoid breaking downstream consumers.

### DIFF
--- a/tools/wled-tools
+++ b/tools/wled-tools
@@ -135,11 +135,11 @@ EOF
 }
 
 # Discover devices using mDNS
-discover_devices() {
-    if ! command -v avahi-browse &> /dev/null; then
+discover_devices() {  
+    if ! command -v avahi-browse &> /dev/null; then  
         log "ERROR" "$RED" "'avahi-browse' is required but not installed, please install avahi-utils using your preferred package manager."
-        exit 1
-    fi
+        exit 1  
+    fi  
 
     # Map avahi responses to strings separated by 0x1F (unit separator), deduplicated
     mapfile -t raw_devices < <(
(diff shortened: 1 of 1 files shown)


LordMike:
I am unsure why the original wled-tools has so many spaces after each line.. it makes the diff weird.

softhack007:
> I am unsure why the original wled-tools has so many spaces after each line.. it makes the diff weird.

@LordMike yeah probably the reason is "it doesn't matter". Normally we recommend "don't try to optimize out trailing whitespaces". As you say, it makes the diff hard to read and it makes cherry-picking between branches a nightmare.

If you have any VSCode add-on that "optimizes" out trailing spaces, just remove them. Also don't use clang-format on the complete file before committing a PR.

LordMike:
I cleaned it up

softhack007:
> I cleaned it up

Thanks, that's _much_ clearer 👍

softhack007:
@coderabbitAI re-review.

softhack007 (approved):
looks fine for me. 
I can't test it, but the only functional difference is ``print key`` => ``if (!seen[key]++) print key``. everything else is just styling.

PR #5497, opened 2026-04-12
0.16 - Backport dynarray fixes from V5 WIP

### LINKED ISSUE #5495: Single configured usermod crashes on ESP32-C3
### What happened?

The default i2ssdPin is hardcoded to 32, but ESP32-C3 only has GPIOs 0–21. This causes pin allocation to fail silently, leaves GPIO 15 stuck as allocated, and ends up with disableSoundProcessing = false      
  despite having no working audio source (line 1475 overrides the earlier check at 1472). Result: watchdog crash loop, no WiFi AP.
                                                                                                                                                                                                                  
  Fix is straightforward — define valid default pins for C3, either in the board env or in the code:                                                                                                              
  
  -D I2S_SDPIN=2                                                                                                                                                                                                  
  -D I2S_WSPIN=3                                            
  -D I2S_CKPIN=4

  Tested on ESP32-C3-DevKitM-1, WLED 0.17.0-dev. Without the fix it bootloops, with it everything works fine.

### To Reproduce Bug

flash audio reactive version to esp32-c3

### Expected Behavior

The board boots properly, there’s no boot loop, and the WLED SSID is visible.

### Install Method

Self-Compiled

### What version of WLED?

0.16.x

ESP32-C3

### Relevant log/trace output

```shell

```

### Anything else?

A propsed PR to solve the issue #5496 

### Code of Conduct

willmmiles:
Bring in the dynarray/usermod improvements from the V5 work in progress.  This improves the system to support LTO and `framework = arduino, esp-idf`, as well as fixes some issues with dynarray linkage on RISCV platforms.

fadc75d in particular should fix #5495 

## Summary by CodeRabbit

* **Refactor**
  * Improved linker script injection mechanism for dynamic arrays to use inlined fragments instead of platform-specific files.
  * Enhanced module validation to use compilation-unit analysis for more reliable module confirmation across platforms.
  * Unified dynamic array section naming across all targets for consistency.


DedeHai (approved):
can confirm it fixes the bootloop on C3

DedeHai:
@willmmiles this also needs to go upstream

PR #5503, opened 2026-04-13
fix for serial port on esp32 (#5501)

### LINKED ISSUE #5501: No ADAlight response on esp32 with wled 16 or 17
### What happened?

No serial console on esp32, tested with two QuinLED-ESP32

### To Reproduce Bug

flash wled 16 or 17, then open the serial console and try any command, for example "v"

### Expected Behavior

serial console should work

### Install Method

Binary from WLED.me

### What version of WLED?

WLED 16.0.0-beta (2604110)

ESP32

### Relevant log/trace output

```shell

```

### Anything else?

tested also `WLED 17.0.0-dev (2604120)`

### Code of Conduct

RobertoD91:
Possible fix for #5501
Tested on a real ESP32, and CI passes. It has not been tested on a real ESP8266.

This is my first pull request, so please feel free to reject it if anything is wrong.

## Summary by CodeRabbit

* **Chores**
  * Refined ESP32 serial communication initialization during startup for improved stability and hardware compatibility.

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -11,6 +11,9 @@
 #include "soc/soc.h"
 #include "soc/rtc_cntl_reg.h"
 #endif
+#ifdef ARDUINO_ARCH_ESP32
+#include "driver/gpio.h"
+#endif
 
 extern "C" void usePWMFixedNMI();
 
@@ -368,13 +371,16 @@ void WLED::setup()
   WRITE_PERI_REG(RTC_CNTL_BROWN_OUT_REG, 0); //disable brownout detection
   #endif
 
-  #ifdef ARDUINO_ARCH_ESP32
-  pinMode(hardwareRX, INPUT_PULLDOWN); delay(1);        // suppress noise in case RX pin is floating (at low noise energy) - see issue #3128
-  #endif
   #ifdef WLED_BOOTUPDELAY
   delay(WLED_BOOTUPDELAY); // delay to let voltage stabilize, helps with boot issues on some setups
   #endif
   Serial.begin(115200);
+  #ifdef ARDUINO_ARCH_ESP32
+  // Pull down RX pin to suppress noise when floating, without breaking UART0 IOMUX mapping.
+  // pinMode() routes GPIO through the GPIO matrix and detaches UART0 RX - use gpio_pulldown_en() instead.
+  // See issue #3128.
+  gpio_pulldown_en((gpio_num_t)hardwareRX);
+  #endif
   #if !ARDUINO_USB_CDC_ON_BOOT
   Serial.setTimeout(50);  // this causes troubles on new MCUs that have a "virtual" USB Serial (HWCDC)
   #else


wled00/wled.cpp
```diff
@@ -368,13 +371,16 @@ void WLED::setup()
...
-  #endif
   #ifdef WLED_BOOTUPDELAY
   delay(WLED_BOOTUPDELAY); // delay to let voltage stabilize, helps with boot issues on some setups
   #endif
   Serial.begin(115200);
+  #ifdef ARDUINO_ARCH_ESP32
+  // Pull down RX pin to suppress noise when floating, without breaking UART0 IOMUX mapping.
+  // pinMode() routes GPIO through the GPIO matrix and detaches UART0 RX - use gpio_pulldown_en() instead.
+  // See issue #3128.
+  gpio_pulldown_en((gpio_num_t)hardwareRX);
```
DedeHai, line 382: all that is needed is to replace pinMode(hardwareRX, INPUT_PULLDOWN); with this line, nothing else.

wled00/wled.cpp
```diff
@@ -368,13 +371,16 @@ void WLED::setup()
...
+  #endif
```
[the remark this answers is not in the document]
softhack007: @coderabbitai your learning is outdated - setRxTimeout(1) is not necessary, the default applied by Serial.Begin() is rxTimeout=2.

softhack007:
@coderabbitai as it turns out, this PR fixes the problem in #5501 - pinMode(hardwareRX, INPUt_PULLDOWN) has an unexpected side effect in arduino-esp32 v2.0.x. It leaves the RX pin in state where it is detached from the uart unit.

Please clean up your learnings collected during discussions of #5501.

wled00/wled.cpp
softhack007: @DedeHai originally the pinMode() call was located before Serial.Begin(), to avoid side-effects when Serial initializes its pins. 🤔 Do you think gpio_pulldown_en() should happen before Serial.Begin(), or keep it after? I don't have much experience with the under-the-hood pin config tools in esp-idf.

DedeHai: it should stay where it was, the API calls are very low level and will directly write the GPIO registers, enabling the pulldown with no fuss.

DedeHai: I can confirm the fix working, all that is needed is this:
replace 
`pinMode(hardwareRX, INPUT_PULLDOWN);`
with 
`gpio_pulldown_en((gpio_num_t)hardwareRX);`
and maybe add a comment that pinMode() sets the gpio mux and breaks UART.

no additional includes, nothing.

DedeHai:
This merge needs cleanup, the proper fix is a 1 line change, the other things changed are unnecessary. I will make a cleanup commit

netmindz:
ah, sorry about that. I miss-understood your comment as confirmation this fixed the issue, so thought ok to merge

wled00/wled.cpp
RobertoD91: Hi, sorry for the delay. I see that you already submitted a clear PR.
Thanks for your work :)

PR #5504, opened 2026-04-14
Add moving skip to Chase 2

RussStringham:
I have added a new slider to Chase 2 that generates values between 0 and 15. When this slider is at its default value of 0, it behaves exactly like the current version of Chase 2. When the slider is greater than 0, it is treated as a skip value. At any instance in time, the result looks the same as when a skip is specified in the segment definition. However, when the stripe moves by one LED, it moves to the next physical LED, rather than the next lit LED. This results is the skips rotatings as well.

## Summary by CodeRabbit

* **New Features**
  * Added Skip control parameter to the running color effect, enabling customization of striped rendering patterns with variable gaps between stripes.

* **Improvements**
  * Enhanced the running color effect with improved animation progression and refined LED color assignment logic for better visual stripe pattern rendering.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -543,26 +543,40 @@ static const char _data_FX_MODE_RAINBOW_CYCLE[] PROGMEM = "Rainbow@!,Size;;!";
 /*
  * Alternating pixels running function.
  */
-static void running(uint32_t color1, uint32_t color2, bool theatre = false) {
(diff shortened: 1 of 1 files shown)


RussStringham:
[PR-5507](https://github.com/wled/WLED/pull/5507) is an alternate implementation of this same functionality, but adds the ability to change the skip (gap) color.

PR #5507, opened 2026-04-15
Add moving gap to Chase 2 version 2

RussStringham:
This PR is an alternate implementation for [PR-5504](https://github.com/wled/WLED/pull/5504). Only one or the other should be merged. Here is the description:

I have added a new slider to Chase 2 that generates values between 0 and 15. When this slider is at its default value of 0, it behaves exactly like the current version of Chase 2. When the slider is greater than 0, it is treated as a gap value. At any instance in time, the result looks the same as when a gap is specified in the segment definition. However, when the stripe moves by one LED, it moves to the next physical LED, rather than the next lit LED. This results is the gaps rotating as well.

In this PR, the gap color is controled by Cs (color 3) instead of always being black. I have also simplified the running() function so that the theater modes only differ from Chase 2 by the parameters that are passed. There is no special logic in the function to handle theater differently than chase. In this case, the theater BG color is the gap color and the theater width is the gap width.

In the previous PR the gap is called the skip value.

## Summary by CodeRabbit

* **New Features**
  * Added a Gap parameter to control stripe spacing.
  * Added a third color for off/filled regions to expand visual options.
  * Exposed Gap and Width in the Running Color effect metadata.

* **Refactor**
  * Reworked chase/running effects to a stripe-based model for finer control of width, spacing, motion, and color handling.
  * Safer parameter clamping and smoother, more predictable motion.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -543,34 +543,31 @@ static const char _data_FX_MODE_RAINBOW_CYCLE[] PROGMEM = "Rainbow@!,Size;;!";
 /*
  * Alternating pixels running function.
  */
-static void running(uint32_t color1, uint32_t color2, bool theatre = false, int skip = 0) {
-  int width = (theatre ? 3 : 1) + (SEGMENT.intensity >> 4);  // window
+static void running(uint32_t color1, uint32_t color2, uint32_t color3, int skip, int width) {
   uint32_t cycleTime = 50 + (255 - SEGMENT.speed);
   uint32_t it = strip.now / cycleTime;
   bool usePalette = color1 == SEGCOLOR(0);
+  bool usePalette2 = usePalette && color2 == color1;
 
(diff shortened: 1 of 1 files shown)


DedeHai:
I think version2 is better as it has the additional color. 
To me "skip" is not the right term, it better be called "gap". also you need to add a checkmark as it breaks existing presets otherwise: to test, create a preset in the current version, then update to the PR and see if it changes.

RussStringham:
@DedeHai I have renamed "skip" to "gap" in the code and the UI.

There is no need for a checkmark. The theater modes work exactly as before. Chase 2 defaults to a gap value of zero. When it is zero, it behaves exactly as the previous version, so previously saved presets will be unchanged.

DedeHai:
> There is no need for a checkmark. 

there is. an existing preset does not use the defaults. please test it.

RussStringham:
I have added a checkmark so that the gap slider is only functional when checked and verified that previously defined chase2 presets work as before.

PR #5513, opened 2026-04-18
Fix blending style options list filter for iOS

### LINKED ISSUE #5498: Safari/webkit browsers show 2D transitions for 1D segments ( correctly hidden on PC)
### What happened?

You get an amended version of transitions when using the desktop browser (chrome) UI vs. the mobile UI on an apple device.

### To Reproduce Bug

Check the transitions list on a mobile device (viewing the mobile/device toolbar in chrome doesn't work) vs the list provided on a computer.

### Expected Behavior

The full list or transitions should be available on the desktop version of the UI.

### Install Method

Binary from WLED.me

### What version of WLED?

16.0.0-beta

ESP32

### Relevant log/trace output

```shell

```

### Anything else?

_No response_

### Code of Conduct

DedeHai:
fixes #5498

tested and it fixes the issue on safari.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved blending style handling between matrix and non-matrix modes so invalid or missing selections are reliably reset to a valid default.

* **Refactor**
  * Reworked option update logic to rebuild and filter blending choices more cleanly, improving reliability and maintainability.

### DIFF
--- a/wled00/data/index.js
+++ b/wled00/data/index.js
@@ -662,10 +662,9 @@ function parseInfo(i) {
 	mh = i.leds.matrix ? i.leds.matrix.h : 0;
 	isM = mw>0 && mh>0;
 	if (!bsOpts) bsOpts = Array.from(gId('bs').options).map(o => o.cloneNode(true)); // snapshot all options on first call
-	const bsSel = gId('bs'), bsCur = bsSel.value;
+	const bsSel = gId('bs');
 	// note: style.display='none' for option elements is not supported on all browsers (notably iOS)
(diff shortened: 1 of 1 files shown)


remixmark:
Hello @DedeHai may I test this fix or do I just wait for the next beta to be released and test from that?

DedeHai:
sure, each PR always has bins built, you can find them here: https://github.com/wled/WLED/actions/runs/24599924765?pr=5513#artifacts

the merged fix is also in the nightly release tomorrow.

blazoncek:
Overcomplicated solution. Adding `hidden` attribute would be sufficient.

Issue #5290, opened 2026-01-12
Pin owner for out-of-tree usermod, and inter-usermod method calls


blazoncek:
**Is your feature request related to a problem? Please describe.**
Currently if one needs GPIO for usermod a new entry is added to `pin_manager.h`  to add a new `PinOwner`. The same goes for usermod ID constant defined in `const.h`.

If one wants to create out-of-tree (i.e. not included in WLED source code, privately maintained) usermod that requires GPIO allocation there is no way to extend `PinOwner` enum (except in local/private repository by modifying `pin_manager.h`) and that defeats usability of such usermod.

The same is true for usermod ID constant defined in `const.h`.

**Describe the solution you'd like**
Add extensibility to `PinManager` to allow addition of custom `PinOwner`. Remove the need for usermod ID constants or replace it with something more versatile.

**Describe alternatives you've considered**
Reuse one of the existing `PinOwner` codes but that is not a safe approach as there might be a clash. Or, provide a patch for upstream to add new owner, but that is no better than the old approach with in-tree usermods.

**Additional context**
Since usermods were modified to behave like libraries (i.e. no longer included in `usermod_list.cpp`) it is expected that WLED API would allow extensibility without modifying WLED's source code.

DedeHai:
that's a good point. IMHO the UM numbering should be taken care of by the usermod manager and dynamically assign the ID's. AFAIK the usermod ID's are not used in any communication protocols and is purely internal. If that is feasible, usermods can just use their ID as a pin owner without conflict.

blazoncek:
> usermod ID's are not used in any communication protocols

That is not entirely true. 4LD, Rotary encoder and Autosave usermods use IDs to identify if usermods are accessible (they were all written by the same author). Then they communicate with each other directly.

DedeHai:
maybe that can be resolved by a function to get the ID through the usermod name? see ongoing work in https://github.com/wled/WLED/pull/5234

blazoncek:
That will require usermod to store other usermod's name, which is IMO unfortunate waste of memory.
There is also a necessity to know other usermod's methods (as seen in 4LD, Rotary encoder and Autosave).
API should be enhaced to support that as well.

softhack007:
"Direct calling of other usermod methods" is a very special case, I haven't seen that in other usermods than the three created by @blazoncek.

> That will require usermod to store other usermod's name, which is IMO unfortunate waste of memory.
[...]

PR #5521, opened 2026-04-21
Integrate um_data structure to battery usermod

PSi86:
Enable other usermods to access the battery voltage and percentage using the um_data interface.

## Summary by CodeRabbit

* **Refactor**
  * Improved battery state caching and handling for more responsive, consistent voltage and level readings; exported battery data now remains stable across restarts and recalibrations.
  * Battery level is recalculated immediately after calibration and on every measurement cycle for up-to-date reporting.

* **Bug Fixes**
  * MQTT battery publishing now sends a consistent integer percentage payload for clearer downstream display.


willmmiles:
We're planning to eliminate the `um_data` interface in favor of direct module accesses or a framework for sensor channels (see discussion in #5290, #5344, and #5319).  The `um_data` API design requires both double indirection and extra memory, costing performance and complexity for little benefit.

PSi86:
Thanks for your reply. I have a completed usermod relying on this. What is the timeline for the new interface? I suggest to merge this as long as the new interface is not ready. https://github.com/PSi86/RaceLink_WLED

DedeHai:
the new interface is on the horizon for 17.0, certainly not for 16.0.
As for your changes: are all the bat->getLevel() replacements with getBatteryLevel() needed? why not update the values for the UM data in the main loop and keep the rest of the code as-is?

usermods/Battery/Battery.cpp
```diff
@@ -16,10 +16,14 @@ class UsermodBattery : public Usermod
   private:
     // battery pin can be defined in my_config.h
     int8_t batteryPin = USERMOD_BATTERY_MEASUREMENT_PIN;
-    
+
     UMBattery* bat = new UnkownUMBattery();
     batteryConfig cfg;
 
+    int8_t batteryLevel = 0; // current battery level in %
+    float batteryVoltage = 0.0f; // current battery voltage in V
```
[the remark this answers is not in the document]
PSi86: In UMBattery.h the battery level is also defined as int8_t and its range is 0 to 100. Please confirm if it should be changed to uint8_t as suggested by coderabbit.

usermods/Battery/Battery.cpp
```diff
@@ -277,13 +305,14 @@ class UsermodBattery : public Usermod
...
       bat->calculateAndSetLevel(filteredVoltage);
+      refreshExportedBatteryData();
 
       // Auto off -- Master power off
       if (autoOffEnabled && (autoOffThreshold >= bat->getLevel()))
         turnOff();
 
 #ifndef WLED_DISABLE_MQTT
-      publishMqtt("battery", String(bat->getLevel(), 0).c_str());
+      publishMqtt("battery", String(static_cast<int>(bat->getLevel())).c_str());
```
DedeHai: why? this is worse.

DedeHai (changes requested):

usermods/Battery/Battery.cpp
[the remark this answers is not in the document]
PSi86: good analysis. what about the change of "String(bat->getLevel(), 0).c_str()" to "String(static_cast<int>(bat->getLevel())).c_str()", what are the advantages and disadvantages?

usermods/Battery/Battery.cpp
PSi86: This change is based on a note from coderabbitai on my first proposed version of the code.

The original `String(bat->getLevel(), 0)` was problematic because for integer values the second `String` argument is the numeric base, not the number of decimal places. So `0` is not “zero decimals” here.

The `static_cast<int>` was intended to force normal decimal integer formatting and avoid small-integer / char-like overload ambiguity. However, I agree that it is more verbose than necessary here. I will change this to `String(bat->getLevel())`, or alternatively `String(bat->getLevel(), DEC)` if you prefer the decimal base to be explicit.

DedeHai: I am not too deep into this usermod. I suspect that a float was once changed to an int (if it is an int now) but doing a static cast of an int into an int only makes sense if String class requires that (I dont think it does) so its more confusing than helping here.

PSi86: I agree with you. But the old notation with 0 as numeric base is not right either. Which of the two proposed fixes should I commit? String(bat->getLevel()), or alternatively String(bat->getLevel(), DEC)?

DedeHai: I would expect String(bat->getLevel()) to use base 10 by default? the `String(bat->getLevel()), 0)` may be a left-over from the time where it was a float and this was required. WLED code has a lot of contributors and history and not everything in it makes perfect sense or is well aligned :)

PSi86: Absolutely. To my knowledge you are right about the base 10 default. Ok, then lets use the simplest form of the code: ```publishMqtt("battery", String(bat->getLevel()).c_str());```

softhack007:
is this PR still compatible with the changes made in
* #5399 ?

softhack007:
@coderabbitai full review. Additionally, check if changes from this PR have conflicts with PR #5399 or #5236

PR #5522, opened 2026-04-21
Non-blocking, unified, dependency-less DMXOutput

### LINKED ISSUE #5133: Issue with the WLED frontend

Mdbelen:
tl;dr:
Related to https://github.com/wled/WLED/pull/5216 also see previous comments there. This is an alternative to make dmx.update() non-blocking. With less dependencies, more unified between ESP8266&32 and I claim more stable.
Fixes https://github.com/wled/WLED/issues/5133

The core essence of this PR is dmx_output.cpp:57, the rest is cosmetics.

Long:
Independent from https://github.com/Stijn-Jacobs I also encountered the problem that once DMX output gets enabled in v0.15.3, WLED becomes super laggy regarding LED output via ArtNet, and almost unreachable via network. Short digging showed, that the DMX output is implemented in a blocking way, leading to 23ms of CPU time per loop run just for DMX.
This is unacceptable.
Stijn kindly prepared a fix in 5216 and I made mine in parallel locally. While I evaluated Stijns solution, I saw on oscilloscope, that the DMX frame initial MAB pulse, which is supposed to be at least 12us, sometimes is only 8us. I consider this suboptimal and would feel a little unsure using this. The used esp_dmx dependency is comparatively super complicated for what DMX needs, so I decided to post my solution as an alternative here.

The DMX part is quite trivial and the same for DMXESPOutput and SparkFunDMX. Now that DMX Input has been moved to another module altogether, I saw no more reason to keep those two separate dependencies and merged both in this new DMXOutput module. ESP8266 is still mostly blocking (due to limitations in HAL/Arduino), while ESP32 is almost completely non-blocking and takes less than 1ms per run.

The module is prepared to have an update rate element added to the web UI in future. Setting update rate to 0 ensures the max possible rate. Maximum with all 512/513 DMX channels is 43Hz, this is currently the default.

_[outdated, see comments: I decided to also change the #defines to better differentiate between the new DMXInput and DMXOutput modules. The new way to enable DMXOutput is to set WLED_ENABLE_DMX_OUTPUT instead of WLED_ENABLE_DMX, which I found confusing.]_
Also, there's the possibility to set the initial TX pin using DMX_TXPIN_DEFAULT=1337. [kno.wled.ge](https://kno.wled.ge/interfaces/dmx-output/) would need an update here.

In this branch, I also included some commits from netmindz https://github.com/wled/WLED/pull/5287 adding TX pin setting on web UI, and the update rate limiting from Stijn (sadly couldn't cherry-pick as was mixed with too many other changes).
I did not use the number of DMX-fixtures UI element, because I consider it irrelevant for ESP32 and found it more confusing. This would lower the processor and RAM load marginally. It could be possibly helpful for ESP8266, but I think this is not the focus here.

This is tested on my ESP32 hardware with an oscilloscope. Didn't get the chance to connect my DMX fixtures, yet, but I'm pretty sure that will work. Untested on ESP8266, no hardware available.

## Summary by CodeRabbit

* **New Features**
  * DMX output pin is now configurable in Sync settings, allowing flexible pin assignment instead of hardcoded defaults

* **Improvements**
  * Updated DMX settings UI with clearer labeling ("DMX Output Settings") and instructional guidance
  * Enhanced DMX output transmission with improved efficiency and rate limiting capabilities


wled00/dmx_output.cpp
```diff
@@ -1,33 +1,186 @@
+#ifdef WLED_ENABLE_DMX_OUTPUT
+
 #include "wled.h"
+#include "dmx_output.h"
+#ifdef ESP8266
+#include "uart.cpp"
+#else
+#include "hal/uart_ll.h"
+#endif
```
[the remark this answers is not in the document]
Mdbelen: Sadly, uart_tx_fifo_available is not available thru the header, so I must go this dirty way. Acceptable?
I didn't find out how HardwareSerial does it (this includes only uart.h), for me it failed.

wled00/dmx_output.cpp
```diff
@@ -1,33 +1,186 @@
...
+  #else //not ESP8266
+  #if SOC_UART_NUM <= 1
+  #error DMX output is not possible on your MCU, as it does not have HardwareSerial(1)
+  #endif
+
+  if(uartNo == -1) {
+    uartNo = SOC_UART_NUM - 1;    // use last UART as default
+  }
+  if(uartNo == 0) {
+    DEBUG_PRINTF_P(PSTR("DMXOutput: Error: Cannot run on chips with <=1 hardware UART, or with UART0."));
```
DedeHai, line 31: why is this restriction necessary?

netmindz:
I'll try and have a proper look through, but a quick bit of feedback

Please don't rename the define for the output feature, it's just an unnecessary breaking change. Now we are on V4 for all builds I'll probably be removing the DMX_INPUT as it's supposed on all esp32s and has minimal size impact

wled00/dmx_output.cpp
Mdbelen: Good question. I'm inclined to say I only updated some module here and these restrictions applied before too, even more restrictive. SparkFunDMX would throw an error if HardwareSerial(2) didn't exist.
This restriction was eased by netmindz in his commit to allow 2-UART SoCs.
The problem IMO is that hardware resources don't have any means to be "reserved/dedicated" to a specific functionality, like PinManager does for pins. I don't have much clue what modules and built-in functionality exists, so what conflicts can or will arise. All I can say is Serial0 would probably be a bad choice, as at least in the beginning we have some debug output there, even if WLED_DEBUG is disabled. If this is enabled, then it gets ofc so much worse.
And even if hw is only used in an ordered, sequential manner, you'd need to pay attention to fully reset all registers, as depending on the HAL, you'd assume reset condition when initializing, thus don't reset bits that were set by other users before.

netmindz:
Thanks for the PR @Mdbelen , can you clarify which MCUs you have tested with and have you used a DMX tester or logic analyzer to validate the output of your new driver?

wled00/dmx_output.cpp
DedeHai: Fair point. We really need a resource manager, there is a growing list of features that have hardware resource conflicts. This is something that came up in discussions several times, it is only a question of when and how.

Mdbelen:
I've changed back to WLED_ENABLE_DMX without _OUTPUT as requested. Note that I kept the DMX_OUTPUT naming with PinManager. Hope that's okay.
I've flashed this to an ESP32 on a custom board and made some oscilloscope measurements. I have no DMX fixtures here, but at another location. Could probably test that on the weekend.

DedeHai:
As this moves away from any library (which I am not at all opposed to), it requires thorough testing on all supported platforms.

Mdbelen:
True in general. As I said in the initial message, the principle is the same between SparkFunDMX, DMXESPOutput (used for ESP8266) and still very similar to this module.

```
void SparkFunDMX::update() {
  if (_READWRITE == _WRITE)
  {
    //Send DMX break
    digitalWrite(txPin, HIGH);
    DMXSerial.begin(BREAKSPEED, BREAKFORMAT, rxPin, txPin);//Begin the Serial port
    DMXSerial.write(0);
    DMXSerial.flush();
    delay(1);
    DMXSerial.end();

    //Send DMX data
    DMXSerial.begin(DMXSPEED, DMXFORMAT, rxPin, txPin);//Begin the Serial port
    DMXSerial.write(dmxData, chanSize);
    DMXSerial.flush();
    DMXSerial.end();//clear our DMX array, end the Hardware Serial port
  }
```
```
void DMXESPSerial::update() {
  if (dmxStarted == false) init();

  //Send break
  digitalWrite(sendPin, HIGH);
  Serial1.begin(BREAKSPEED, BREAKFORMAT);
  Serial1.write(0);
  Serial1.flush();
  delay(1);
  Serial1.end();

  //send data
  Serial1.begin(DMXSPEED, DMXFORMAT);
  digitalWrite(sendPin, LOW);
  Serial1.write(dmxDataStore, channelSize);
  Serial1.flush();
  delay(1);
  Serial1.end();
}
```

```
  // Send DMX break
  // Cannot change UART format while running. End and reinit takes much longer than the additional stopbit here.
  _dmxSerial->updateBaudRate(BREAKSPEED); //change to DMX break settings
  _dmxSerial->write(0);
  _dmxSerial->flush();

  // Send DMX data
  _dmxSerial->updateBaudRate(DMXSPEED);   //change to regular DMX speed
  _dmxSerial->write(_dmxData, DMX_CHANNELS);
```

The difference in mine is that i set `_dmxSerial->setTxBufferSize(DMX_CHANNELS + SOC_UART_FIFO_LEN);` and don't flush the buffer after writing the DMX data. Then I use low level API access to check whether the transfer has finished before allowing the next DMX frame to be sent and reject calls to update() in between, if _updateRate is set to 0 (max speed). Or I use the _updateRate setting to throttle the frames.
I would hope that the HardwareSerial and Arduino libs are the same for all ESP32xx variants(?) so they'd all be checked if mine is running. ESP8266 is a different story.
I've just build esp32S3_wroom2, lolin_s2_mini and esp32c3dev variants to check whether I'd get any incompatible API errors or warnings. This is not the case.

Maybe if I find time, I'll add more UI improvements to allow the TX pin to be changed without rebooting. And add an input field to set the DMXOutput serial port (like with DMXInput) and update rate.

DedeHai:
> The difference in mine is that i set `_dmxSerial->setTxBufferSize(DMX_CHANNELS + SOC_UART_FIFO_LEN);` and don't flush the buffer after writing the DMX data.

In my experience each and every ESP can act differently as the hardware implementations are not the same and they do have quirks, so testing is absolutely required.

wled00/cfg.cpp
```diff
@@ -607,6 +607,9 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
   tdd = if_live[F("timeout")] | -1;
   if (tdd >= 0) realtimeTimeoutMs = tdd * 100;
 
+  #ifdef WLED_ENABLE_DMX_OUTPUT
+    CJSON(dmxOutputPin, if_live_dmx[F("dmxOutputPin")]);
+  #endif
```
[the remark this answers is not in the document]
Mdbelen: init() changed from uint to int. This was indeed a problem and funnily still is the exact same case with DMXInput.

Mdbelen:
Successfully tested WLED as Art-Net to DMX with ESP32WROVER and Cameo and Eurolight fixtures.

nmeurer:
Successfully tested this on esp32c3dev (Pin 5) using a SN75176B IC and a Eurolite fixture

softhack007:
@coderabbitAI please compare the new implementation in this PR with the previous code in `main`, especially ESPDMX and SparkfunDMX (send mode). Is the previous behaviour fully included in the new implementation, including special handling of channel0?

softhack007:
@coderabbitai the previous implementation in `main` had a problem due to missing rate limiting - `handleDmxOutput()` was called on each main loop iteration, regardless of the fact that DMX fixtures usually cannot handle more than ~40Hz refresh rate. This lead to serial buffer congestion and very bad overall response time, see https://github.com/wled/WLED/issues/5133#issuecomment-3657734389. The suggested solution was to add a simple rate limiter in `handleDmxOutput()`.

Is the refresh rate limit problem addressed properly in this PR?

softhack007:
@Mdbelen please check if any of the differences listed by the rabbit in https://github.com/wled/WLED/pull/5522#issuecomment-4351904509 would need a fix in your new code.

softhack007:
@coderabbitai please check if this PR implies other behavioral changes (user-visible and not yet covered by the PR description) or ripple effects that should be considered by reviewers.

wled00/e131.cpp
```diff
@@ -117,9 +117,8 @@ void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol){
   #ifdef WLED_ENABLE_DMX
   // does not act on out-of-order packets yet
   if (e131ProxyUniverse > 0 && uni == e131ProxyUniverse) {
-    for (uint16_t i = 1; i <= dmxChannels; i++)
-      dmx.write(i, e131_data[i]);
-    dmx.update();
+    dmxOutput.writeBytes(1, &e131_data[1], dmxChannels);
```
softhack007, line 120: Should this be `e131_data[0]` in Art-Net mode? Afaik the first channel offset is 0 for Art-Net, but 1 for other E131 protocols?

softhack007: @coderabbitai please verify if my concern is correct here.

softhack007: > However, **this is a pre-existing bug**, not a regression introduced by this PR.

As the PR is a major modernization/overhaul of previous code, I would like to see this old problem addressed, too.

Mdbelen:
This btw also fixes the bug covered in #5316 .

@softhack007 Thanks for chasing the rabbit though this PR! Here's some comments to the findings:

> 1\. Break byte serial format — protocol compliance concern

According to wikipedia, DMX defines only minimum times for break and MAB. As noted in comment on dmx_output.cpp:114
"Cannot change UART format while running. End and reinit takes much longer than the additional stopbit here." keeping the UART format at 8N2 looked to be faster at oscilloscope captures than calling end() and init() which is needed to change the format.

> 2\. No `digitalWrite(txPin, HIGH)` before break
> Old `SparkFunDMX::update()` explicitly drove the TX pin HIGH before reconfiguring the UART for the break. The new code omits this. This could affect the Mark Before Break (MBB) period on some hardware configurations.

As the new implementation keeps the UART "alive" all the time and never releases the pin to GPIO, this is unnecessary in all cases.
Besides, I unnecessarily do write the pin to high in init().^^

> 5\. Break speed: 83333 → 83000 baud

I tried to do a real pessimistic implementation, so I did a 0,5% increase in break time here to account for possible crystal drifts 🙃 (this is calculated from the wikipedia's 120us, as before).

> 1\. dmxOutputPin defaults to 0 on fresh install (GPIO0 bug)

I believe this bug is untrue, as dmxOutputPin is initialized to -1 in wled.h and the CJSON macro will do the suggested handling already. Can you confirm this @softhack007 @coderabbitai ?

> 2\. DEFAULT_LED_PIN migration for ESP8266 + DMX users

I didn't remember changing anything about LED pins haha. Turns out this comes from @netmindz cherry-picked commit. AI comment makes sense to me, so I'll revert those lines, k? ESP8266 can only TX DMX (Serial1) on pin 2.

> 3\. TX pin change requires reboot but no reboot warning is shown
> 4\. Update rate hardcoded at 43 Hz — no UI to change it

True, planned to add to a future release. Will probably add to this PR very soon

IMO that covers all important AI findings.

wled00/e131.cpp
Mdbelen: Is there any bug report about this? I kept everything as-is because I figured this would lead to a misalignment by one channel for Art-Net (or E1.31) and somebody would have reported for sure. So I thought e131.cpp handles this somewhere internally by biasing the channels for me already.
Will look into it.

PR #5523, opened 2026-04-22
GoL missing return.

Brandon502:
Return likely was deleted when FRAMETIME was removed. This fixes speed slider and games restarting correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Game of Life effect to properly handle paused states and redraw cycles, preventing unnecessary processing that could impact performance during paused animations.


DedeHai:
good call, I accidentally removed that return.

PR #5524, opened 2026-04-22
make per-segment on/off and brightness behave the same as global

### LINKED ISSUE #5520: New transitions also for individual segments.
Good morning. The new features in version 16 are great. I especially like the new transitions when turning bars on and off. If we have a single strip and turn it on and off globally, the transitions work great. However, if the bar is divided into segments and we only want to turn a single segment on and off, the transitions don't work. I'm not looking for a workaround using presets, because in my case, when I want to control two segments with a single button (single-click and double-click), presets won't work. It would be nice if the transitions also worked when turning a single segment on and off.

DedeHai:
fix for #5520

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable segment transitions when another transition is already active.
  * Smoother, more consistent brightness blending (fixes on/off brightness glitches).
  * Opacity changes now start predictable transitions.
  * Per-segment off pixels no longer show residual colors.
  * Global brightness transitions now start only on actual brightness changes; one‑shot transition delays supported.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -287,17 +287,19 @@ void Segment::startTransition(uint16_t dur, bool segmentCopy) {
     return;
   }
   if (isInTransition()) {
+    const uint8_t visibleBri = currentBri();
     if (segmentCopy && !_t->_oldSegment) {
       if (blendingStyle != TRANSITION_FADE) {
         // transition without a copy means we are in a segment brightness transition
-        // need to cancel the transition to not interfere with the next one (or global brightness change)
+        // need to cancel the transition to not interfere with the new one (or global brightness change)
         stopTransition();
       } else {
-        // already in FADE transition but segment copy requested  and not yet created (for example color change during brightness transition)
+        // already in FADE transition but segment copy requested  and not yet created (for example a SWIPE color change during brightness transition)
         // create segment copy and restart timer so currentBri() can interpolate correctly
         _t->_oldSegment = new(std::nothrow) Segment(*this); // store/copy current segment settings
         _t->_start = millis();                              // restart countdown
         _t->_dur   = dur;
+        _t->_bri   = visibleBri;
         _t->_prevPaletteBlends = 0;
         if (_t->_oldSegment) {
           _t->_oldSegment->palette = _t->_palette;          // restore original palette and colors (from start of transition)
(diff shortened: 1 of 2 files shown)


DedeHai:
@coderabbitai please compile a test plan to verify the changes, focus on edge cases that might behave incorrectly. also check if any externally applied changes by sync or home assistent etc. work

DedeHai:
performed all tests except the ones involving HA and MQTT and all behave correctly

blazoncek:
If you want my opinion: This incorrectly solves OP issue. It also changes behaviour of non-segment-copy transitions.
The real OP culprit lies elsewhere.

DedeHai:
yea, I am not too convinced this is the proper let alone best solution. If you find something better, please keep me posted.

blazoncek:
I think I've solved the issue (all of them). Watch my fork for changes after I test them.

DedeHai:
@coderabbitai in `button.cpp` there is this condition to delay off mode:
`(millis() - lastOnTime > 600`
what purpose does it serve? please check reported issues as well as the code.

softhack007:
> 3. `strip.service()` loses the queued final black frame because the hardware power is already off.
> 
> The 600 ms delay gives the output path a grace interval. The `!strip.needsUpdate()` check ensures that no final update remains queued.

Not sure this is the primary purpose. The timer check is implemented in button.cpp, and it appears to be very old code.
* Maybe this was initially meant as a button/relay de-bouncer. If the relay was directly controlled by button, this would prevent relay on/off flickering when the button bounces.
* (slightly less likely) It could be a guard for the output path, but different than the rabbit explains. WS2812b is output-only, so normally there is no need to wait for "output queue draining" before power-off. However, some LED strips (especially "fairy light" ones) keep the last light status even when cutting the 5V supply. In case the last "black frame" is cut by power-off, its possible that these strips produce a short flash at next power-on.

softhack007:
> ```c++
> } else if ((millis() - lastOnTime > 600 && !strip.needsUpdate()) || forceOff) {
> ```

@coderabbitai could this line cause problems when the last "on" command has arrived less than 600ms ago, for example from HyperHDR? Is it possible that the power-off is silently lost?

DedeHai:
@softhack007 my hunch is the 600ms delay was added to caputre a "long press" from a button/switch as that takes 600ms to register. If that is actually the case, the delay could be skipped if no buttons are currently pressed.

PR #5535, opened 2026-04-25
Allow us to use loadPalette() by moving from protected to public

BobLoeffler68:
We currently cannot use loadPalette() to change palettes via the code because it's a protected member, not a public member.  I want to use it to load random palettes in an FX, so I have moved loadPalette() to the Public section in FX.h   I tested the change and it works.

## Summary by CodeRabbit

* **Refactor**
  * Palette loading accessibility changed to be publicly available for direct use.

* **Style**
  * Whitespace normalization performed in one effect (formatting only; no behavioral changes).


DedeHai:
looks good, please revert the unrelated whitespace changes, also move the changed line up by one not leaving a blank line before it.

BobLoeffler68:
> looks good, please revert the unrelated whitespace changes, also move the changed line up by one not leaving a blank line before it.

Done.  Thx.

blazoncek:
FYI there are other places in WLED code that could use public method.

BobLoeffler68:
> FYI there are other places in WLED code that could use public method.

For loadPalette()?  Or other methods?

PR #5544, opened 2026-04-26
refactor platformio.ini to eliminate copy-paste duplication

netmindz:
Centralise platform, platform_packages, build_unflags, lib_deps and monitor_filters into the per-chipset abstract sections ([esp8266], [esp32], [esp32s2], [esp32c3], [esp32s3]) and have concrete envs inherit via extends rather than repeating the same values.

- [esp8266]: add platform, platform_packages, monitor_filters
- [esp32]: add monitor_filters
- [esp32s2]: add monitor_filters
- [esp32c3]: add monitor_filters
- [esp32s3]: add upload_speed, monitor_filters
- esp8266 envs (nodemcuv2, esp8266_2m, esp01_1m_full): extends = esp8266
- esp32dev: extends = esp32
- esp32dev_8M, esp32dev_16M: extends = env:esp32dev
- esp32_eth, usermods: extends = esp32 / env:esp32dev
- esp32_wrover: drop redundant build_unflags, lib_deps (already extends esp32_idf_V4)
- esp32c3dev: drop redundant platform, platform_packages, framework, build_unflags, lib_deps, board_build.partitions (already extends esp32c3)
- esp32s3 envs: extends = esp32s3
- lolin_s2_mini: extends = esp32s2

No functional changes. Verified with pio run across all five chipset families (ESP8266, ESP32, C3, S3, S2).

## Summary by CodeRabbit

* **Refactor**
  * Consolidated build environment configs to use shared platform groups for ESP8266, ESP32, ESP32-S2, ESP32-S3, and ESP32-C3, reducing duplication.
  * Centralized monitor filter settings and standardized inheritance across environments.

* **Chores**
  * Increased default upload speed for ESP32-S3 devices.
  * Updated sample override templates to explicitly reference ESP8266 platform/package variables.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -41,40 +41,6 @@ extra_configs =
 
 [common]
 # ------------------------------------------------------------------------------
-# PLATFORM:
-#   !! DO NOT confuse platformio's ESP8266 development platform with Arduino core for ESP8266
-#
-#   arduino core 2.6.3 = platformIO 2.3.2
-#   arduino core 2.7.0 = platformIO 2.5.0
-# ------------------------------------------------------------------------------
-arduino_core_2_6_3 = espressif8266@2.3.3
-arduino_core_2_7_4 = espressif8266@2.6.2
-arduino_core_3_0_0 = espressif8266@3.0.0
-arduino_core_3_0_2 = espressif8266@3.2.0
-arduino_core_3_1_0 = espressif8266@4.1.0
-arduino_core_3_1_2 = espressif8266@4.2.1
-
-# Development platforms
-arduino_core_develop = https://github.com/platformio/platform-espressif8266#develop
-arduino_core_git = https://github.com/platformio/platform-espressif8266#feature/stage
-
-# Platform to use for ESP8266
-platform_wled_default = ${common.arduino_core_3_1_2}
-# We use 2.7.4.7 for all, includes PWM flicker fix and Wstring optimization
-#platform_packages = tasmota/framework-arduinoespressif8266 @ 3.20704.7
-platform_packages = platformio/toolchain-xtensa @ ~2.100300.220621 #2.40802.200502
-                    platformio/tool-esptool #@ ~1.413.0
-                    platformio/tool-esptoolpy #@ ~1.30000.0
-
-## previous platform for 8266, in case of problems with the new one
-## you'll need  makuna/NeoPixelBus@ 2.6.9 for arduino_core_3_0_2, which does not support Ucs890x
-;; platform_wled_default = ${common.arduino_core_3_0_2}
-;; platform_packages = tasmota/framework-arduinoespressif8266 @ 3.20704.7
-;;                    platformio/toolchain-xtensa @ ~2.40802.200502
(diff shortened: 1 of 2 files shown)


platformio.ini
```diff
@@ -359,22 +366,20 @@ build_flags = -g
 lib_deps =
   ${esp32_idf_V4.lib_deps}
 board_build.partitions = ${esp32.large_partitions}   ;; default partioning for 8MB flash - can be overridden in build envs
+upload_speed = 921600
```
softhack007, line 368: maybe comment out the upload speed flag. 920kbaud might be too much for boards where the UART-to-USB module gets connected by flimsy dupond wires only.

softhack007:
@netmindz I generally like the idea to use "extends = " for reducing the number of copied entries in buildenvs. Actually, I've proposed a similar approach some time ago, but then there was user feedback that people prefer to see "full entries", and "extends = " also created new pitfalls and makes it harder to understand which flags are actually passed to the compiler. 

Not sure if this has changed in the meantime?

platformio.ini
```diff
@@ -192,6 +192,8 @@ lib_deps =
 extra_scripts = ${scripts_defaults.extra_scripts}
 
 [esp8266]
+platform = ${common.platform_wled_default}
+platform_packages = ${common.platform_packages}
```
willmmiles: We should move all the ESP8266 platform definitions in to this section instead of `common`, since they're not `common` for all environments.  Yes, this will potentially be a breaking change for user `platformio_override.ini` files for anyone  building custom ESP8266 environments; I think that's reasonable for 17.0, to keep our definitions clean (and prepare for a proper HAL).

platformio.ini
willmmiles: My $0.02, but I'd prefer to default to the higher speed -- it's much more convenient and failures are easily caught.  If a user is doing a custom build and needs to lower the speed, it's straightforward to do in the override file.  Try the fast speed first and lower if you have to, instead of the other way around.

(AFAIK these settings don't affect users using the web installer or the like -- just people doing custom builds.)

netmindz:
> @netmindz I generally like the idea to use "extends = " for reducing the number of copied entries in buildenvs. Actually, I've proposed a similar approach some time ago, but then there was user feedback that people prefer to see "full entries", and "extends = " also created new pitfalls and makes it harder to understand which flags are actually passed to the compiler.
> 
> Not sure if this has changed in the meantime?

I think it's fine so long as we keep it to just one (2 at max) levels of extends. Most things defined at the chipset level, the env is then only setting things like the `custom_usermods`, filesystem size and not much else really. We no longer need to mess about with `lib_deps` and `build_flags` for usermods, so if you really are a power user enough to want to pass custom compiler flags, you should be able to follow some fairly basic ini file settings

platformio.ini
netmindz: now is the time to break things, at the very start of the development cycle

platformio.ini
netmindz: What happens if you have no value?

Issue #5412, opened 2026-03-09
DDP "skip out-of sequence packets" assumes that 5 packets are enough for a complete frame


softhack007:
https://github.com/wled/WLED/blob/53d6c4b059b60578cb2bdb7ebdb9c38b8dcf1dd4/wled00/e131.cpp#L24-L38

We need to verify if the assumption stated in the source code - 4 packets max before push - is correct w.r.t to the protocol definitions, and analyze if packets that are more than 4 out-of-sequnce can lead to a corrupted LEDs display.

With HUB75 panels, users have succesfully streamed 64x64 pixels, so it seem that larger frame sizes  generally do work.

Reference
https://github.com/wled/WLED/pull/5390#discussion_r2878786110

softhack007:
@coderabbitai please check the DDP protocol specifications. Is  the assumption "4 packets max before push" aligned with the DDP specs? How many DDP packets would be needed for a full frame of 64x64 RGB pixels?

softhack007:
@coderabbitai extended question - how would our code handle frames that use more than 15 packets for a full frame? Would some packets be discarded as "duplicate", or would a full frame still be treated correctly, at least without "skip out of sequence"?

Is the 4bit sequence field  part of the protocol specifications, or was it a wled design choice to only evaluate the lowest 4 bits?

DedeHai:
@softhack007 I think this can be closed, IMHO the current implementation is a good compromise

softhack007:
@DedeHai you're right, it's probably the best that can be done with only 15 sequence numbers. And the "drop threshold" is actually after 10 packets, not 5. So 64x64 RGB is still on the edge of "work with drop late packets". 

Maybe we can (later?) add a workaround that disables out-of-sequnece skipping when more than 11 DDP packets were used in the last "push frame".

I'll leave it up to you, I'm fine with closing this ticket now.

PR #5547, opened 2026-04-26
less restrictive DDP header acceptance, add safety checks to all protocols

### LINKED ISSUE #5532: ddp from openRGB to wled stopped working in some later v16 nightly version
EDIT:  updating typos and misstates
### What happened?

one of the later v16 nightly versions broke ability to control wled device using openRGB ddp.  i had been updating nightly v16 versions over the last few months without issue and i only recently observed this problem, so it must be in one of the later v16 nightly but unfortunately i cant tell which one.
rolling back to v0.15.4 resolves the problem.
switching from DDP protocol to E1.31 in openRGB is a workaround unless you exceed max leds supported by E1.31

very basic config with 2 outputs of same type, attached configs

[wled_cfg_stripscase.json](https://github.com/user-attachments/files/27039215/wled_cfg_stripscase.json)

[wled_cfg_stripscase.json](https://github.com/user-attachments/files/27039219/wled_cfg_stripscase.json)

### To Reproduce Bug

install latest pawnio version of openrgb and add manual ddp device and point it to wled esp32 controller.   this works in v0.15.4 but doesnt work in v16 although it did work in v16 one of the earlier nightly versions.

### Expected Behavior

ddp should continue working

### Install Method

Binary from WLED.me

### What version of WLED?

v16 beta

ESP32

### Relevant log/trace output

```shell

```

### Anything else?

_No response_

### Code of Conduct

DedeHai:
instead of being restrictive about the DDP protocol WLED now accepts malformed headers again and does packet size checking instead to prevent crashes.
also adding packet size checking for other protocols.

fix for #5532

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved length and bounds validation for E1.31/Art‑Net/DDP packets to prevent crashes, out‑of‑bounds reads, and misaligned proxy‑universe data.

* **New Features**
  * Packet handlers and callbacks now receive actual payload lengths for safer processing (including websocket paths).
  * DDP filtering tightened to ignore control/status/query frames; storage frames accepted only when explicitly pushed.
  * DDP color handling defaults to RGB, using RGBW only when indicated.

* **Chores**
  * Clarified inline comments and added a TODO about header/sequence semantics.

### DIFF
--- a/wled00/e131.cpp
+++ b/wled00/e131.cpp
@@ -33,6 +33,12 @@ static void handleDDPPacket(e131_packet_t* p, size_t packetLen) {
   // reject control, status and config packets (not implemented)
   if (p->destination == DDP_ID_CONTROL || p->destination == DDP_ID_STATUS || p->destination == DDP_ID_CONFIG) return;
 
+  // reject query and response packets (not implemented)
+  if (p->flags & (DDP_FLAGS_QUERY | DDP_FLAGS_REPLY)) return;
+
+  bool push = p->flags & DDP_FLAGS_PUSH; // push flag means "render now"
+  if (!push && (p->flags & DDP_FLAGS_STORAGE)) return; // reject "from storage" flag but still let the push flag pass if set along with it
+
   //reject late packets belonging to previous frame (assuming 4 packets max. before push, if more are used and packets are very late, they are still accepted)
   if (e131SkipOutOfSequence && lastPushSeq) {
     int sn = p->sequenceNum & 0xF; // sequence number is 4 bits, 1-15, 0 means unused
@@ -78,7 +84,6 @@ static void handleDDPPacket(e131_packet_t* p, size_t packetLen) {
     }
   }
 
-  bool push = p->flags & DDP_FLAGS_PUSH;
   ddpSeenPush |= push;
   if (!ddpSeenPush || push) { // if we've never seen a push, or this is one, render display
     e131NewData = true;
--- a/wled00/e131.cpp
+++ b/wled00/e131.cpp
@@ -5,7 +5,7 @@
 #define MAX_CHANNELS_PER_UNIVERSE 512
 
 // forward declarations
-static void handleDDPPacket(e131_packet_t* p);
+static void handleDDPPacket(e131_packet_t* p, size_t packetLen);
 static void handleArtnetPollReply(IPAddress ipAddress);
 static void prepareArtnetPollReply(ArtPollReply *reply);
 static void sendArtnetPollReply(ArtPollReply *reply, IPAddress ipAddress, uint16_t portAddress);
@@ -17,20 +17,25 @@ static void sendArtnetPollReply(ArtPollReply *reply, IPAddress ipAddress, uint16
 
 //DDP protocol support, called by handleE131Packet
 //handles RGB data only
-static void handleDDPPacket(e131_packet_t* p) {
+static void handleDDPPacket(e131_packet_t* p, size_t packetLen) {
   static bool ddpSeenPush = false;  // have we seen a push yet?
   int lastPushSeq = e131LastSequenceNumber[0];
 
+  if (packetLen < DDP_HEADER_LEN) return; // too short to safely read any DDP header fields
+
   // reject unsupported color data types (only RGB and RGBW are supported)
-  uint8_t maskedType = p->dataType & 0x3F; // mask out custom and reserved flags, only type bits are relevant
-  if (maskedType != DDP_TYPE_RGB24 && maskedType != DDP_TYPE_RGBW32) return;
+  //uint8_t maskedType = p->dataType & 0x3F; // mask out custom and reserved flags, only type bits are relevant
+  //if (maskedType != DDP_TYPE_RGB24 && maskedType != DDP_TYPE_RGBW32) return;
 
-  // reject status and config packets (not implemented)
-  if (p->destination == DDP_ID_STATUS || p->destination == DDP_ID_CONFIG) return;
+  // note: for maximum compatibility we do not reject unknonw or malformed data types but simply default to RGB24 and check there is enough data available in the packet to do so
+  //       also we assume 8bit per channel and currently do not support other bit depths
 
-  //reject late packets belonging to previous frame (assuming 4 packets max. before push)
+  // reject control, status and config packets (not implemented)
+  if (p->destination == DDP_ID_CONTROL || p->destination == DDP_ID_STATUS || p->destination == DDP_ID_CONFIG) return;
+
+  //reject late packets belonging to previous frame (assuming 4 packets max. before push, if more are used and packets are very late, they are still accepted)
   if (e131SkipOutOfSequence && lastPushSeq) {
-    int sn = p->sequenceNum & 0xF;
+    int sn = p->sequenceNum & 0xF; // sequence number is 4 bits, 1-15, 0 means unused
     if (sn) {
       if (lastPushSeq > 5) {
         if (sn > (lastPushSeq -5) && sn < lastPushSeq) return;
@@ -40,7 +45,8 @@ static void handleDDPPacket(e131_packet_t* p) {
     }
   }
 
-  unsigned ddpChannelsPerLed = ((p->dataType & 0b00111000)>>3 == 0b011) ? 4 : 3; // data type 0x1B (formerly 0x1A) is RGBW (type 3, 8 bit/channel)
+  unsigned ddpChannelsPerLed = 3; // default to RGB
+  if ((p->dataType & 0b00111000)>>3 == 0b011) ddpChannelsPerLed = 4; // RGBW data type (see DDP protocol definition)
 
   uint32_t start =  htonl(p->channelOffset) / ddpChannelsPerLed;
   start += DMXAddress / ddpChannelsPerLed;
@@ -49,6 +55,12 @@ static void handleDDPPacket(e131_packet_t* p) {
   uint8_t* data = p->data;
   unsigned c = 0;
   if (p->flags & DDP_FLAGS_TIME) c = 4; //packet has timecode flag, we do not support it, but data starts 4 bytes later
+
+  // ensure the received packet is at least as long as the header claims
+  if (packetLen < DDP_HEADER_LEN + c + dataLen) {
+    DEBUG_PRINTLN(F("DDP packet incomplete"));
+    return;
+  }
 
   unsigned numLeds = stop - start; // stop >= start is guaranteed
   unsigned maxDataIndex = c + numLeds * ddpChannelsPerLed; // validate bounds before accessing data array
@@ -76,7 +88,7 @@ static void handleDDPPacket(e131_packet_t* p) {
 }
 
 //E1.31 and Art-Net protocol support
-void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol){
+void handleE131Packet(e131_packet_t* p, IPAddress clientIP, byte protocol, size_t packetLen){
 
   int uni = 0, dmxChannels = 0;
   uint8_t* e131_data = nullptr;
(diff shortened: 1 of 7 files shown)


DedeHai:
@netmindz the rabbit seems content with the changes, I did test DDP and artnet from ESP to ESP and it is working but I don't know if there are any edge cases. In theory, the size checks are safe.

wled00/src/dependencies/e131/ESPAsyncE131.h
```diff
@@ -64,10 +64,14 @@ typedef struct ip_addr ip4_addr_t;
 
 #define DDP_TYPE_RGB24  0x0B // 00 001 011 (RGB , 8 bits per channel, 3 channels)
 #define DDP_TYPE_RGBW32 0x1B // 00 011 011 (RGBW, 8 bits per channel, 4 channels)
+#define DDP_TYPE_LEGACY 0x01 // 00 000 001 legacy RGB 8bit definition
```
softhack007, line 67: maybe add ``DDP_TYPE_UNDEF 0x00`` (type / bit depth undefined)

DedeHai: can do that, its more documentation than anything else: these values are not used.

softhack007:
@DedeHai we might also need to reject "query" packets - it seems that some frontends have a kind of "auto-discovery" feature that might cause LEDs flicker if we interpret the content as pixels.

https://github.com/xLightsSequencer/xLights/blob/62ed59baa1a1dc555c6c36803f3372d01882d3a8/src-core/outputs/DDPOutput.cpp#L164

http://www.3waylabs.com/ddp/#Discovery

Like this (over-restrictive) example https://github.com/MoonModules/WLED-MM/commit/579477c0bae227927e612f8c8c5a8ce833decf84

Edit: its not clear to me (from reading specs) if these QUERY/REPLY/STORAGE packets will always have a non-display destination.

DedeHai:
`#define DDP_DISCOVERPACKET_LEN 10` (from xLights) so this will be rejected as "contains no data" already but maybe be explicit about it is better.

softhack007:
> `#define DDP_DISCOVERPACKET_LEN 10` (from xLights) so this will be rejected as "contains no data" already but maybe be explicit about it is better.

good point - maybe better to reject explicitly, unless we are 100% sure (from the specs) that these packets will always be 10 bytes only.

softhack007:
@DedeHai I found another 10 bytes "message" - DDP sync. Its basically a "push" with dataLen=0. 

Do you think we already handle that one properly? If we miss it, that could lead to dropped frames.

````c++
// sent data to all devices, now broadcast PUSH flag so they all display sync'd
if (NDEVICES > 1)  // if 1 device, already sent PUSH above
{
   dh.flags1 = DDP_FLAGS1_VER1 | DDP_FLAGS1_PUSH;
   dh.id     = DDP_ID_DISPLAY;
   dh.offset = 0;
   dh.len    = 0;  // <--- !!
   UDP_SEND(local-IP-broadcast-address,DDP_PORT,dh,databuf);         
}
````

**Comment by DedeHai** [comment]
> maybe better to reject explicitly, 

agreed.

> unless we are 100% sure (from the specs) that these packets will always be 10 bytes only.

if a software sends a query packet that contains LED data it also needs to specify the data length. I am not sure ther eis a mechanism that is a non-config packet that contains data AND is valid according to the protocol. It's certainly possible to construct such a bogus message - and if you do, you have to live with the consequences ;)

**Comment by DedeHai** [comment]
> @DedeHai I found another 10 bytes "message" - DDP sync. Its basically a "push" with dataLen=0.
> 
> Do you think we already handle that one properly? If we miss it, that could lead to dropped frames.
> 
> ```c++
> // sent data to all devices, now broadcast PUSH flag so they all display sync'd
> if (NDEVICES > 1)  // if 1 device, already sent PUSH above
> {
>    dh.flags1 = DDP_FLAGS1_VER1 | DDP_FLAGS1_PUSH;
>    dh.id     = DDP_ID_DISPLAY;
>    dh.offset = 0;
>    dh.len    = 0;  // <--- !!
>    UDP_SEND(local-IP-broadcast-address,DDP_PORT,dh,databuf);         
> }
> ```

@coderabbitai please check if such a "push only" message is handled correctly (I suspect it is not). Also do a cross-check on known projects that use DDP like hyperion, xlights, signalRGB and check if the code from this PR handles all messages including the actual RGB data correctly.

File: wled00/e131.cpp [hunk 2]
```diff
@@ -17,20 +17,31 @@ static void sendArtnetPollReply(ArtPollReply *reply, IPAddress ipAddress, uint16
...
+  // reject control, status and config packets (not implemented)
+  if (p->destination == DDP_ID_CONTROL || p->destination == DDP_ID_STATUS || p->destination == DDP_ID_CONFIG) return;
+
+  // reject query and response packets (not implemented)
+  if (p->flags & (DDP_FLAGS_QUERY | DDP_FLAGS_REPLY)) return;
 
-  // reject status and config packets (not implemented)
-  if (p->destination == DDP_ID_STATUS || p->destination == DDP_ID_CONFIG) return;
+  bool push = p->flags & DDP_FLAGS_PUSH; // push flag means "render now"
+  if (!push && (p->flags & DDP_FLAGS_STORAGE)) return; // reject "from storage" flag but still let the push flag pass if set along with it
```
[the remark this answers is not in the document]
softhack007: this is ensured at line 34.

wled00/e131.cpp
```diff
@@ -76,24 +93,29 @@ static void handleDDPPacket(e131_packet_t* p) {
...
+    if (packetLen < 10) return; // need at least art_opcode (offset 8, 2 bytes)
     if (p->art_opcode == ARTNET_OPCODE_OPPOLL) {
       handleArtnetPollReply(clientIP);
       return;
     }
+    if (packetLen < 18) return; // need art_length (offset 16, 2 bytes) for DMX data
     uni = p->art_universe;
     dmxChannels = htons(p->art_length);
+    const int artNetMaxData = (packetLen > 19) ? (int)(packetLen - 19) : 0; // art_data at offset 18; clamp so e131_data[dmxChannels] stays in bounds
+    if (dmxChannels > artNetMaxData) dmxChannels = artNetMaxData;
```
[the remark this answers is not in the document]
DedeHai: @softhack007 do you know anything about artnet protocol? I am not too sure the implementation of packet restrictions in this PR is 100% correct and I have no means to test any of it.

wled00/e131.cpp
```diff
@@ -17,20 +17,31 @@ static void sendArtnetPollReply(ArtPollReply *reply, IPAddress ipAddress, uint16
...
-  // reject status and config packets (not implemented)
-  if (p->destination == DDP_ID_STATUS || p->destination == DDP_ID_CONFIG) return;
+  bool push = p->flags & DDP_FLAGS_PUSH; // push flag means "render now"
+  if (!push && (p->flags & DDP_FLAGS_STORAGE)) return; // reject "from storage" flag but still let the push flag pass if set along with it
 
-  //reject late packets belonging to previous frame (assuming 4 packets max. before push)
+  //reject late packets belonging to previous frame (assuming 4 packets max. before push, if more are used and packets are very late, they are still accepted)
   if (e131SkipOutOfSequence && lastPushSeq) {
-    int sn = p->sequenceNum & 0xF;
+    int sn = p->sequenceNum & 0xF; // sequence number is 4 bits, 1-15, 0 means unused
```
softhack007, line 44: @coderabbitAI please check if WLED handling of sequence numbers is compatible with DDP protocol specs. Accepting an "outdated" packet may be acceptable, but a wrong rejection should be avoided, even when more than 15 packets are needed to transmit a full "frame". Also check against the real implementation of known DDP senders like LedFx, hyperion, xlights, OpenRGB, FPP.

DedeHai: @softhack007 I looked at this code in detail, it is not perfect but rejecting "late packets" seems reasonable. if large setups are needed, just disable this option in the sync settings (it is disabled by default), it is not possible to have both, late packet rejection AND a large setup as the sequence number for DDP is limited to 1-15 and to check for late packets, a max window of 7 would be possible.

softhack007: @DedeHai @coderabbitai I was actually wondering if we could make improvements for https://github.com/wled/WLED/issues/5412 while working on the code. 

The DDP spec reads like they wanted to support packet re-transmit within one "push frame", so the sequence counter could be used to identify duplicates (in a history window of 15 packets max). I think dups (same frame) are not a problem in WLED, because they simply cause a re-painting of some pixels. You're right that rejecting late packets - from the previous "push frame" - might be useful in some setups - especially with high latency / unstable WiFi and with lower pixel counts. My fear is that users will simply enable "Skip out-of-sequence packets" (which makes sense for other protocols like Art-Net), and later complain that a bigger setup always misses pixel updates when sequence checks are enabled.

I'm not sure atm where is the "breaking limit" (matrix size) when sequence checking will start to mis-behave.  But as this is a legacy "buggy feature" I could live with a use-as-is decision 🤔, so we still support the previous "proper use" cases.

What do you think?

softhack007: Ok, so with RGB (not RGBW which is totally uncommon in large displays, and not supported for HUB75), false rejections start above ~ 82x64 pixels.

DedeHai: It's not perfect and the fix is to uncheck that option, its more a question of documentation. Until looking into this code for this PR I did not even know that option existed :)
We can always refine at a later point if this becomes an issue with growing number of HUB75 users. The key point of this PR is to make DDP as compatible as we dare to at the cost of accepting wrong packets that lead to a screwed up display (which will be rare).

What I am most concerned about in this PR is rejecting valid packets in non DDP protocols, which may be the case if the rabbit is correct in its statement above on artnet packets.

DedeHai: so if we can not verify that to be working in a real setup, I'd rather comment out those packet size checks for now as this PR will block 16.0 release until done.

wled00/e131.cpp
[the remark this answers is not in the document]
softhack007: @DedeHai i don't know much, but @troyhacks is a real power-user of Art-Net, maybe he can help.

There is some info here https://mm.kno.wled.ge/advanced/ddp/#art-net

Besides that, there is also limited testing possibilities for me, as I don't have any Art-Net controller hardware.

wled00/e131.cpp
softhack007: Yes, good point. As testing is very hard to do, especially we don't know much about all the possible senders - maybe this PR should strictly focus on "prevent array bounds violations" and on DDP. Which means, any clamping that is not driven by array/buffer bounds should be commented out.

DedeHai:
@coderabbitai is this code also used for artnet? if so is it correct? if its incorrect, is this a long standing bug?
`for (uint16_t i = 1; i <= dmxChannels; i++)
    dmx.write(i, e131_data[i]);`

DedeHai:
@netmindz  please check the off by one bug I found, are you aware of any reports on that? I do not use dmx so don't know what it means really or how the bug would manifest.
cc @troyhacks

softhack007:
> ```c++
> if (e131ProxyUniverse > 0 && uni == e131ProxyUniverse) {
> ```

@coderabbitai can this condition be true in Art-Net mode? Please also check the Art-Net specs to clarify if the protocol supports DMX proxy operation.

softhack007:
@DedeHai I would lean towards fixing the Art-Net off-by-one in DMX proxy mode, as suggested by the rabbit. 

To test this, we would need to have a UART DMX fixture that is driven by an Art-Net control desk over network, WLED acting as a proxy between them - so its almost impossible to test without buying semi-professional DMX equipment, plus understanding the logic of the control desk.

I still think that using WLED as a WiFi-to-UART router is an abuse, but hey why not ;-)

softhack007:
> Is it a long-standing bug?
> Yes. The loop itself was not introduced by this PR

@coderabbitAI please trace back the commit history and identify the commit (and related PR) that introduced the DMX proxy loop. Also check for old issue tickets related to the change. Was Art-Net support already implemented when the DMX proxy output code was added?

softhack007:
> @DedeHai I would lean towards fixing the Art-Net off-by-one in DMX proxy mode, as suggested by the rabbit.

@DedeHai I changed my mind (not literally 😉) - fixing the bug now would add a new incompatibility with previous versions, as Art-Net output channels would move one to the left. Let's keep this as-is, and fix it later in 
* https://github.com/wled/WLED/pull/5522

which is a complete re-write of the DMX output driver.

DedeHai:
@softhack007 how about keeping this PR for main and I'll open a DDP changes only version for the 16 branch?

wled00/e131.cpp
```diff
@@ -50,6 +62,12 @@ static void handleDDPPacket(e131_packet_t* p) {
   unsigned c = 0;
   if (p->flags & DDP_FLAGS_TIME) c = 4; //packet has timecode flag, we do not support it, but data starts 4 bytes later
 
+  // ensure the received packet is at least as long as the header claims
+  if (packetLen < DDP_HEADER_LEN + c + dataLen) {
+    DEBUG_PRINTLN(F("DDP packet incomplete"));
+    return;
+  }
```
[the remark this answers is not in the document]
DedeHai: please explain your conclusion. one checks against packetLen i.e. the actual length of the packet, the other checks agains "claimed" length from the header. how are they equivalent?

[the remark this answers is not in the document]
DedeHai: if DDP_FLAGS_TIME=true but actual time data is not sent, omitting that `+c`  could still result in an OOB read.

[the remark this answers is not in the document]
DedeHai: I think the +c in the second check is actually a bug. `dataLen` is the length of color data in bytes, excluding all header bytes ("if T flag, header extended 4 bytes for timecode field (not counted in data length)" from DDP protocol definition)

softhack007:
@DedeHai good idea 👍

DedeHai:
merged to get more feedback. need to fix things as they emerge.

netmindz:
Also cheery picked to 16_x ? @DedeHai

DedeHai:
Not yet, it may be too restrictive and break edge cases

DedeHai:
@netmindz I forgot to mention 16.0 had a slim version of this already.

PR #5553, opened 2026-04-30
Add Ethernet profiles for QuinLED Dig-Uno/Quad v4 and Dig-Octa 32-8L v4

intermittech:
I will soon be updating the QuinLED Dig-Uno and Dig-Quad to version v4 which uses a new ESP32 module and with that a new Ethernet configuration including external crystal and doing the same for the Dig-Octa brainboard also updating it to v4 with it's own configuration (sadly, since I'm using a different module there).

Sadly none of the existing profiles match my configuration used.

I'd like for these profiles to be included within WLED so that when the new boards are introduced they are immediately compatible with v16 of WLED!

I have tested the proposed configurations on prototype hardware and they are working correctly.

If there are any questions, please let me know, I'm always available on Discord user the user "Quindor".

## Summary by CodeRabbit

* **New Features**
  * Added support for two new Ethernet board types: QuinLED v4 Uno/Quad and QuinLED v4 Octa.
  * Ethernet device list updated so the new QuinLED v4 options appear in network settings.
  * Each new option includes matching presets for PHY, clock mode and GPIO pin configuration to simplify setup.

### DIFF
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -399,8 +399,8 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
 #define WLED_ETH_ESP32_POE_WROVER 11
 #define WLED_ETH_LILYGO_T_POE_PRO 12
 #define WLED_ETH_GLEDOPTO         13
-`#define` WLED_ETH_QUINLED_V4_UNOQUAD  14
-`#define` WLED_ETH_QUINLED_V4_OCTA     15
+#define WLED_ETH_QUINLED_V4_UNOQUAD  14
+#define WLED_ETH_QUINLED_V4_OCTA     15
 
 
 //Hue error codes
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -399,8 +399,8 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
 #define WLED_ETH_ESP32_POE_WROVER 11
 #define WLED_ETH_LILYGO_T_POE_PRO 12
 #define WLED_ETH_GLEDOPTO         13
-#define WLED_ETH_QUINLED-v4-UnoQuad  14
-#define WLED_ETH_QUINLED-v4-Octa     15
+`#define` WLED_ETH_QUINLED_V4_UNOQUAD  14
+`#define` WLED_ETH_QUINLED_V4_OCTA     15
 
 
 //Hue error codes
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -382,7 +382,7 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
 #define BTN_TYPE_TOUCH_SWITCH     9
 
 //Ethernet board types
-#define WLED_NUM_ETH_TYPES        14
+#define WLED_NUM_ETH_TYPES        16
 
 
 #define WLED_ETH_NONE              0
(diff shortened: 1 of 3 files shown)


wled00/const.h
```diff
@@ -399,6 +399,9 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
 #define WLED_ETH_ESP32_POE_WROVER 11
 #define WLED_ETH_LILYGO_T_POE_PRO 12
 #define WLED_ETH_GLEDOPTO         13
+`#define` WLED_ETH_QUINLED_V4_UNOQUAD  14
+`#define` WLED_ETH_QUINLED_V4_OCTA     15
```
softhack007: should be ``#define`` not ``` `#define` ```

softhack007: @coderabbitai why didn't you spot this during your own review?

wled00/data/settings_wifi.htm
```diff
@@ -243,6 +243,9 @@ <h3>Ethernet Type</h3>
 			<option value="3">WESP32</option>
 			<option value="1">WT32-ETH01</option>
 			<option value="13">Gledopto</option>
+			<option value="14">QuinLED v4 Uno/Quad</option>
+			<option value="15">QuinLED v4 Octa</option>
```
softhack007: This list does not need to be ordered by value.
You could move the new entries up, so they come directly after the other two "QuinLED" entries.

wled00/data/settings_wifi.htm
```diff
@@ -243,6 +243,9 @@ <h3>Ethernet Type</h3>
...
+
```
softhack007: (minor) please remove this extra blank line.

DedeHai:
just out of curiosity: why did you not use the same ETH config for all hardware? is there an issue with the dig octa pinout for eth to not reuse that?

wled00/const.h
```diff
@@ -399,6 +399,9 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
 #define WLED_ETH_ESP32_POE_WROVER 11
 #define WLED_ETH_LILYGO_T_POE_PRO 12
 #define WLED_ETH_GLEDOPTO         13
+#define WLED_ETH_QUINLED-v4-UnoQuad  14
+#define WLED_ETH_QUINLED-v4-Octa     15
```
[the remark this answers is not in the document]
softhack007: @coderabbitai your "Commitable suggestion" contains invalid code: ``` `#define` ``` has stray backticks, so this suggestion will not compile.

intermittech:
> just out of curiosity: why did you not use the same ETH config for all hardware? is there an issue with the dig octa pinout for eth to not reuse that?

The Dig-Uno/Quad v4 are using the same new ESP32 WROOM module (ESP32-PICO-MINI-02) with 8MB flash and 2MB PSRAM as the recently introduced Dig-Next-2 and those have different pins available. I also made the switch to use an external crystal instead of internal out allowing the use of WiFi (or ESP-now) and such while Ethernet is active.

The Dig-Octa Brainboard 32-8L v4 still uses the "older" ESP32 WROOM but already had one with 16MB flash so it keeps that one but also moves to the external crystal hence still needing a new profile.

Wish I could have kept it the same but the IO pins available between the two ESP modules differ so much that that sadly wasn't possible.

wled00/data/settings_wifi.htm
intermittech: Ok, trying and also renamed them slightly!

wled00/data/settings_wifi.htm
intermittech: Ack!

wled00/const.h
netmindz: Can you remove the backtick please @intermittech

PR #5565, opened 2026-05-04
Track cmDNS-default state to restore Device-Name-based hostname

### LINKED ISSUE #5563: DHCP request supplied hostname changed from 0.15.4 -> 16.0.0
### What happened?

After upgrading to WLED 16.0.0 a ESP32 based WLED device the hostname changed back to `wled-<last-6-hex-mac>`.

It appears that the DHCP supplied hostname used to be derived from the "Device Name" (under User Interface) which I had set.

It probably does not matter, but for completeness: I did upgrade through Home Assistant update entity. After the update the device was no longer reachable since Home Assistant was no longer to resolve the WLED device by the same hostname.

### To Reproduce Bug

1. Install WLED 0.15.4
2. Set "Device Name" to `something` in Settings → User Interface, labeled "Device Name"
3. DHCP request uses hostname `wled-something` as preferred hostname. Typical routers will resolve based on their default DNS suffix, e.g. `wled-something.lan` for OpenWrt routers
4. Upgrade to WLED 16.0.0
5. Observe different hostname is used in DHCP request (based on MAC)

### Expected Behavior

The same hostname is used in DHCP requests.

### Install Method

Binary from WLED.me

### What version of WLED?

WLED 16.0.0 (2605030)

ESP32

### Relevant log/trace output

```shell

```

### Anything else?

_No response_

### Code of Conduct

agners:
PR #5424 made `getWLEDhostname()` prefer `cmDNS` over the legacy `serverDescription`-based hostname, but `cmDNS` is still overwritten at boot with "wled-<mac6>" when it holds the "x" default sentinel. The new hostname check (strcmp_P against `DEFAULT_MDNS_NAME`) sees the auto-filled value and treats it as user-configured, so users who never customized the mDNS field — and previously got ``wled-<DeviceName>`` as their network hostname — now get ``wled-<mac6>`` instead.

Track whether the auto-fill was applied via a new cmDNSisDefault flag, and use that in `getWLEDhostname()`. The flag is set at boot from the comparison against `DEFAULT_MDNS_NAME` and cleared when the user saves the WiFi settings form. All other consumers (MDNS.begin, ArduinoOTA.setHostname, captive portal, MQTT, bus_manager, improv) keep using the auto-filled `cmDNS` unchanged.

Fixes: #5563

## Summary by CodeRabbit

Improved internal tracking of mDNS hostname configuration to better distinguish between auto-generated and user-configured values, enabling more accurate settings management going forward.

* **Refactor**
  * Enhanced hostname configuration state tracking for improved settings management

### DIFF
--- a/wled00/set.cpp
+++ b/wled00/set.cpp
@@ -101,6 +101,7 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
     }
 
     strlcpy(cmDNS, request->arg(F("CM")).c_str(), 33);
+    cmDNSisDefault = false; // any explicit save means user-set (form has no way to re-enter the "x" sentinel)
 
     apBehavior = request->arg(F("AB")).toInt();
     char oldSSID[33]; strcpy(oldSSID, apSSID);
--- a/wled00/util.cpp
+++ b/wled00/util.cpp
@@ -156,7 +156,7 @@ static void sanitizeHostname(char* hostname, size_t maxLen) {
  */
 void getWLEDhostname(char* hostname, size_t maxLen, bool preferMDNS) {
   if (maxLen <= 6) { strlcpy(hostname, "wled", maxLen); return; } // buffer too small (should not happen)
-  if (preferMDNS && (strlen(cmDNS) > 0) && (strcmp_P(cmDNS, PSTR(DEFAULT_MDNS_NAME)) != 0)) {     // avoid "x" = not set (use wled-MAC)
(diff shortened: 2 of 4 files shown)


softhack007:
@agners before we look further into your PR, we need to clarify if what you experienced is actually a bug, or intended (new) behaviour.

Please see my comments in #5563 and check the description in #5424 for expected behaviour.

softhack007:
Closing - this PR tries to improve backwards compatibility; however it also adds new problems:

* errors in the persistence path (see comments by the rabbit)
* failure to detect DEFAULT_MDNS_NAME on fresh installs
* still leads to different names used by mDNS vs. DHCP/DNS.

While the backwards compatibility problem should receive further investigation, I think this PR does not provide a future-proof solution, sorry.

PR #5569, opened 2026-05-04
Add boot_lor usermod - enforces a configured realtime override at startup

smitty078:
Applies a configured realtime override (lor) value after initial network activity.

Useful for API-driven setups (HomeKit, Home Assistant, etc.) where realtime streaming (DDP/E1.31) should not take control at boot.

- Optional additional delay before applying
- Reasserts value for a configurable duration to handle startup races

## Summary by CodeRabbit

* **New Features**
  * Boot LOR usermod: waits for network, optional delay, applies a configured realtime override at boot, reasserts for a set duration, then stops.

* **Documentation**
  * README with configuration schema, runtime status indicators, example config, tested hardware notes, and behavioral details.

* **Chores**
  * Packaging manifest added for installation.
  * New usermod identifier added for platform integration.

### DIFF
--- a/usermods/boot_lor/library.json
+++ b/usermods/boot_lor/library.json
@@ -1,0 +1,11 @@
+{
+    "name": "boot_lor",
+    "version": "1.0.0",
+    "description": "Set WLED realtime override at boot",
+    "frameworks": "arduino",
+    "platforms": "espressif32",
+    "build": {
+        "libArchive": false,
+        "srcDir": "."
+    }
+}
--- a/usermods/boot_lor/readme.md
+++ b/usermods/boot_lor/readme.md
@@ -49,7 +49,6 @@ WiFi connected → (optional delay) → apply lor → assert for N seconds → s
 ```
 
 - Waits for network connectivity
-- Optionally waits for the first UDP packet (recommended for DDP setups)
 - Waits an additional configurable delay (if set)
 - Applies the configured `lor` value
 - Reasserts it for a short period to allow system "settling"
@@ -64,7 +63,6 @@ The following options are available under `"boot_lor"` in the WLED config:
 ```json
 "boot_lor": {
   "bootLor": 2,
-  "waitForUdpPacket": true,
   "additionalWaitSec": 0,
   "assertForSec": 10
 }
@@ -75,8 +73,7 @@ The following options are available under `"boot_lor"` in the WLED config:
 | Name                | Type | Default | Description |
 |---------------------|------|---------|-------------|
 | `bootLor`           | int  | `2`     | Realtime override mode to apply. Valid values: `-1` (disabled), `0`, `1`, `2` |
-| `waitForUdpPacket`  | bool | `true`  | Wait for first UDP packet before starting the delay timer |
-| `additionalWaitSec` | int  | `0`     | Additional delay (in seconds) after trigger (connection or UDP) before applying |
+| `additionalWaitSec` | int  | `0`     | Additional delay (in seconds) after trigger (connection) before applying |
 | `assertForSec`      | int  | `10`    | Duration (in seconds) to reassert the value after first application |
 
 ---
@@ -88,7 +85,6 @@ For most DDP / API-first setups:
 ```json
 {
   "bootLor": 2,
-  "waitForUdpPacket": true,
   "additionalWaitSec": 0,
   "assertForSec": 10
 }
@@ -104,7 +100,6 @@ This ensures:
 
 ## Notes
 
-- This usermod does **not** block or modify UDP traffic
 - It does **not** interfere with realtime streaming once `lor` is manually changed
 - It simply ensures a predictable startup state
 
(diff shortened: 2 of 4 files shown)


smitty078:
@coderabbitai resume

softhack007:
@coderabbitAI review

netmindz:
Thanks for the submission, but it feels like we should fix the underlying issue of not being able to set that value at boot than a usermod messing about with state

netmindz:
@coderabbitai please investigate why setting lor as part of a boot preset does not work 100% of the time

smitty078:
> Thanks for the submission, but it feels like we should fix the underlying issue of not being able to set that value at boot than a usermod messing about with state

Thanks for the feedback. It’s encouraging that you’re interested in going after a fix in core so that presets can work reliably for this use case. As a (new) user looking for this use-case to work for me, the usermod was the quick fix I needed and thought I’d submit as it may help others looking for the same behavior. As I investigated initially I found a fair number of “setting lor in a preset doesn’t work” type reports around the web that corroborated my own experience. Having no prior knowledge of the codebase, I didn’t initially attempt to look for a fix in core. I’m happy to take a stab at it and/or test a proposed fix as this is my primary use-case currently, so I’m invested in it working. Is the way forward to work towards implementing a proposed fix myself, or await a fix from you/the team?

willmmiles:
> ### Why the Usermod Works (and the Core Fix)
> 
> The `boot_lor` usermod addresses this by continuously re-asserting `realtimeOverride` in `loop()` for a configurable duration after connectivity — working around all three issues. But the **proper fix in core WLED** would be one or more of:
> 
>     1. **Persist `realtimeOverride` in `cfg.json`** — it's currently a runtime-only global (`wled00/wled.h:834`, `_INIT(REALTIME_OVERRIDE_NONE)`). If it were loaded from config during `deserializeConfig()`, it would be set before the main loop ever starts, eliminating the race entirely.
> 
>     2. **Reorder the main loop** — run `handlePresets()` before `handleNotifications()`. This is a minimal change that ensures the boot preset `lor` is applied before any streaming packet can lock the strip.
> 
>     3. **Don't reset `realtimeOverride` in `exitRealtime()` unconditionally** — distinguish between "set by a persistent/config source" vs. "one-shot API call". If `lor` came from a boot preset or config, it should survive stream timeouts.
> 
> 
> Option 1 is probably the cleanest long-term fix, as it removes the dependency on loop ordering entirely.

@coderabbitai Option 1 would be flat out wrong.  Live override is meant to be a dynamic property of the system, so that it can be enabled or disabled along with other all the other effect state.  Persisting the value every time it's changed would cause excessive flash wear as the cfg would be constantly rewritten as that feature is turned on and off; the obvious alternative, naiively adding it to the cfg serialization, would have it be persisted only when some other cfg setting is saved, causing the at-boot behaviour to mysteriously change depending on what value it happened to have when some other setting was saved.

Probably the correct fix here is to actually apply the boot preset JSON before starting the main loop (instead of using the defer-execution approach we use to manage resource contention).  This would also have the benefit of being able to absolutely guarantee user-controlled initialization of the FX state before any frames are rendered.

smitty078:
I did my own analysis today and I’ve concluded that I believe the best and correct option is to call handlePresets() immediately after applyPreset() in startStrip(), where the boot preset is initially added to the queue. This totally resolves the issue with the only side effect being a single use case: a boot preset that does not set lor:1 or lor:2 is set AND realtime data is received immediate or soon after boot. In the case, the user would observe a preset-> realtime transition in state, which might be undesirable, but should not be unexpected. The solution would be to avoid a boot preset that sets state that interferes with potential realtime data. This is easily enough explained in documentation and is not a behavioral bug.

smitty078:
Would it be appropriate to test this solution and submit a PR for it? Is your preference to have a new PR or to revert and amend this one for sake of discussion that has already occurred?

willmmiles:
> I did my own analysis today and I’ve concluded that I believe the best and correct option is to call handlePresets() immediately after applyPreset() in startStrip(), where the boot preset is initially added to the queue. This totally resolves the issue with the only side effect being a single use case: a boot preset that does not set lor:1 or lor:2 is set AND realtime data is received immediate or soon after boot. In the case, the user would observe a preset-> realtime transition in state, which might be undesirable, but should not be unexpected. The solution would be to avoid a boot preset that sets state that interferes with potential realtime data. This is easily enough explained in documentation and is not a behavioral bug.

Almost.  `handlePresets()` should not be called until at least after `UsermodManager::setup()` to avoid confusing usermods by calling their `readFromJsonState()` functions prior to calling `setup()`.  There's also a risk that `serializeConfigToFS()` gets polluted by usermods interacting with preset state.   (Presets can be really powerful!)

I'd sketched this patch myself tonight, but I hadn't tested it thoroughly.  I'd be happy to accept a better tested contribution. :)

smitty078:
I'll take a look and should have something to submit in the next few days. I tested my initial plan as described above and it works flawlessly for my particular use case buy I hadn't considered the implications you described for cases other than my own. I'm pretty confident I can knock out something pretty well vetted in the next couple of days.

netmindz:
Let's close this PR and we look forward to your bugfix for correct boot loading @smitty078

PR #5573, opened 2026-05-07
Improve Boot preset handling

smitty078:
## Problem

Previously, boot preset application depended on the normal loop path. `handlePresets()` in the main loop is gated behind realtime mode checks:

```cpp
if (!realtimeMode || realtimeOverride || (realtimeMode && useMainSegmentOnly))
```

If realtime mode became active before the first normal preset processing pass, the boot preset could be skipped or only partially applied. This was especially problematic for boot presets intended to set `LiveDataOverride` (`lor`).

---

## Solution

This PR adds `handlePresets()` and `handlePlaylist()` calls during `setup()`.

The startup-time calls are intentionally placed:

- after `UsermodManager::setup()`
- after any required config migration save (`serializeConfigToFS()`)
- before normal network and realtime activity begins

This placement avoids calling usermod JSON handlers before setup while ensuring boot presets are processed deterministically during startup.

---

### Earlier and More Consistent “LEDs On” Timing

For users utilizing a boot preset with `on:true`, LEDs may now turn on earlier during boot.

Because the boot preset is now applied earlier in the startup sequence, lights may illuminate faster than before. During testing, this range varied greatly with a range of 100ms faster to almost 2s faster.

This is particularly beneficial for users powering controllers from a physical wall switch who expect lights to turn on immediately when power is applied.

Users who intentionally relied on the previous delayed or inconsistent timing — particularly setups sensitive to brownouts or power fluctuations — may need to adjust any related delays accordingly.

While the previous behavior could vary depending on timing and realtime state, the new behavior should provide more consistent startup timing across reboots and power-on events on the controller.

### Improvements to Boot Preset Capabilities

- The boot preset is now applied deterministically during setup(), rather than during the main loop.
- The boot preset or playlist can now reliably set `lor` during startup.
- The boot preset can now request another preset using JSON API `ps`

Example:

```json
{ "lor": 2, "ps": 250 }
```
This is useful for a boot preset to establish realtime override behavior first and then immediately advance to another preset.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable preset and playlist processing at startup to ensure expected state after boot.
  * Preset application behavior refined to avoid unintended removal of minimal preset markers and to prevent boot-initialization mode from affecting preset selection.
  * Improved handling when selecting presets that only trigger cycling markers for consistent results.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wled/WLED/pull/5573)

### DIFF
--- a/wled00/json.cpp
+++ b/wled00/json.cpp
@@ -520,7 +520,7 @@ bool deserializeState(JsonObject root, byte callMode, byte presetId)
     if (root["win"].isNull() && getVal(root["ps"], presetCycCurr, 1, 250) && presetCycCurr > 0 && presetCycCurr < 251 && presetCycCurr != currentPreset) {
       DEBUG_PRINTF_P(PSTR("Preset select: %d\n"), presetCycCurr);
       // b) preset ID only or preset that does not change state (use embedded cycling limits if they exist in getVal())
-      if (callMode == CALL_MODE_INIT) callMode = CALL_MODE_DIRECT_CHANGE // avoid propogating CALL_MODE_INIT, which may cause accidental recursion.
+      if (callMode == CALL_MODE_INIT) callMode = CALL_MODE_DIRECT_CHANGE; // avoid propogating CALL_MODE_INIT, which may cause accidental recursion.
       applyPreset(presetCycCurr, callMode); // async load from file system (only preset ID was specified)
       return stateResponse;
     } else presetCycCurr = currentPreset; // restore presetCycCurr
--- a/wled00/json.cpp
+++ b/wled00/json.cpp
@@ -520,8 +520,9 @@ bool deserializeState(JsonObject root, byte callMode, byte presetId)
     if (root["win"].isNull() && getVal(root["ps"], presetCycCurr, 1, 250) && presetCycCurr > 0 && presetCycCurr < 251 && presetCycCurr != currentPreset) {
       DEBUG_PRINTF_P(PSTR("Preset select: %d\n"), presetCycCurr);
       // b) preset ID only or preset that does not change state (use embedded cycling limits if they exist in getVal())
-      if (callMode == CALL_MODE_INIT) callMode = CALL_MODE_DIRECT_CHANGE; // avoid propogating CALL_MODE_INIT, which may cause accidental recursion.
-      applyPreset(presetCycCurr, callMode); // async load from file system (only preset ID was specified)
+      // async load from file system (only preset ID was specified)
+      // avoid propogating CALL_MODE_INIT, which may cause accidental recursion
+      applyPreset(presetCycCurr, callMode == CALL_MODE_INIT ? CALL_MODE_DIRECT_CHANGE : callMode);
       return stateResponse;
     } else presetCycCurr = currentPreset; // restore presetCycCurr
   }
--- a/wled00/json.cpp
+++ b/wled00/json.cpp
@@ -520,6 +520,7 @@ bool deserializeState(JsonObject root, byte callMode, byte presetId)
     if (root["win"].isNull() && getVal(root["ps"], presetCycCurr, 1, 250) && presetCycCurr > 0 && presetCycCurr < 251 && presetCycCurr != currentPreset) {
       DEBUG_PRINTF_P(PSTR("Preset select: %d\n"), presetCycCurr);
       // b) preset ID only or preset that does not change state (use embedded cycling limits if they exist in getVal())
+      if (callMode == CALL_MODE_INIT) callMode = CALL_MODE_DIRECT_CHANGE // avoid propogating CALL_MODE_INIT, which may cause accidental recursion.
       applyPreset(presetCycCurr, callMode); // async load from file system (only preset ID was specified)
       return stateResponse;
     } else presetCycCurr = currentPreset; // restore presetCycCurr
--- a/wled00/presets.cpp
+++ b/wled00/presets.cpp
@@ -195,6 +195,10 @@ void handlePresets()
   } else {
     if (!fdo["seg"].isNull() || !fdo["on"].isNull() || !fdo["bri"].isNull() || !fdo["nl"].isNull() || !fdo["ps"].isNull() || !fdo[F("playlist")].isNull()) changePreset = true;
 
+    // TODO:
+    // Scenario 1: boot preset (CALL_MODE_INIT AND sets lor>0) - do not strip ps, allowing boot preset to chain 1 single preset
+    // Scenario 2: boot playlist (CALL_MODE_DIRECT_CHANGE AND sets lor>0) - continue, else re-queue. is CALL_MODE_DIRECT_CHANGE correct or do I need to pass through CALL_MODE_INIT or create a new special-case call mode
+    
     if (!(tmpMode == CALL_MODE_INIT || (tmpMode == CALL_MODE_BUTTON_PRESET && fdo["ps"].is<const char *>() && strchr(fdo["ps"].as<const char *>(),'~') != strrchr(fdo["ps"].as<const char *>(),'~')))) // exception for CALL_MODE_INIT and button preset
       fdo.remove("ps"); // remove load request for presets to prevent recursive crash (if not called by button and contains preset cycling string "1~5~")
     deserializeState(fdo, CALL_MODE_NO_NOTIFY, tmpPreset); // may change presetToApply by calling applyPreset()
--- a/wled00/presets.cpp
+++ b/wled00/presets.cpp
@@ -173,7 +173,7 @@ void handlePresets()
   } else
   #endif
   {
-    presetErrFlag = readObjectFromFileUsingId(getPresetsFileName(tmpPreset < 255), tmpPreset, pDoc) ? ERR_NONE : ERR_FS_PLOAD;
+  presetErrFlag = readObjectFromFileUsingId(getPresetsFileName(tmpPreset < 255), tmpPreset, pDoc) ? ERR_NONE : ERR_FS_PLOAD;
   }
   fdo = pDoc->as<JsonObject>();
 
--- a/wled00/presets.cpp
+++ b/wled00/presets.cpp
@@ -176,12 +176,24 @@ void handlePresets()
   } else
   #endif
   {
-  presetErrFlag = readObjectFromFileUsingId(getPresetsFileName(tmpPreset < 255), tmpPreset, pDoc) ? ERR_NONE : ERR_FS_PLOAD;
+    presetErrFlag = readObjectFromFileUsingId(getPresetsFileName(tmpPreset < 255), tmpPreset, pDoc) ? ERR_NONE : ERR_FS_PLOAD;
   }
   fdo = pDoc->as<JsonObject>();
 
-  // only reset errorflag if previous error was preset-related
+  // only reset error flag if previous error was preset-related
   if ((errorFlag == ERR_NONE) || (errorFlag == ERR_FS_PLOAD)) errorFlag = presetErrFlag;
+
+  // is this the boot preset and will the preset set lor
+  const bool isBootPreset = (tmpMode==CALL_MODE_INIT || tmpPreset==bootPreset);
+  const bool presetWillSetLor = (!fdo["lor"].isNull() && fdo["lor"].as<int>() > REALTIME_OVERRIDE_NONE);
+  
+  // During setup, only allow the boot preset itself or safe boot-preset chains.
+  const bool shouldAllowPresetApply = (
+                                       setupComplete || isBootPreset ||
+                                       (currentPreset == bootPreset && realtimeOverride > REALTIME_OVERRIDE_NONE) ||
+                                       (currentPreset == bootPreset && currentPlaylist > 0 && presetWillSetLor)
+                                       );
+  if (!shouldAllowPresetApply) return;
 
   //HTTP API commands
   const char* httpwin = fdo["win"];
@@ -194,13 +206,12 @@ void handlePresets()
     changePreset = true;
   } else {
     if (!fdo["seg"].isNull() || !fdo["on"].isNull() || !fdo["bri"].isNull() || !fdo["nl"].isNull() || !fdo["ps"].isNull() || !fdo[F("playlist")].isNull()) changePreset = true;
-
-    // TODO:
-    // Scenario 1: boot preset (CALL_MODE_INIT AND sets lor>0) - do not strip ps, allowing boot preset to chain 1 single preset
-    // Scenario 2: boot playlist (CALL_MODE_DIRECT_CHANGE AND sets lor>0) - continue, else re-queue. is CALL_MODE_DIRECT_CHANGE correct or do I need to pass through CALL_MODE_INIT or create a new special-case call mode
     
-    if (!(tmpMode == CALL_MODE_INIT || (tmpMode == CALL_MODE_BUTTON_PRESET && fdo["ps"].is<const char *>() && strchr(fdo["ps"].as<const char *>(),'~') != strrchr(fdo["ps"].as<const char *>(),'~')))) // exception for CALL_MODE_INIT and button preset
-      fdo.remove("ps"); // remove load request for presets to prevent recursive crash (if not called by button and contains preset cycling string "1~5~")
+    // only allow load requests from boot presets that set lor or button calls
+    const bool isButtonException = (tmpMode == CALL_MODE_BUTTON_PRESET && fdo["ps"].is<const char *>() && strchr(fdo["ps"].as<const char *>(),'~') != strrchr(fdo["ps"].as<const char *>(),'~'))
+    
+    const bool shouldAllowLoadRequest = (isBootPreset && presetWillSetLor) || (tmpMode == CALL_MODE_BUTTON_PRESET && fdo["ps"].is<const char *>() && strchr(fdo["ps"].as<const char *>(),'~') != strrchr(fdo["ps"].as<const char *>(),'~'))
+    if (!shouldAllowLoadRequest) fdo.remove("ps"); // remove load request for presets to prevent recursive crash
     deserializeState(fdo, CALL_MODE_NO_NOTIFY, tmpPreset); // may change presetToApply by calling applyPreset()
   }
   if (!errorFlag && tmpPreset < 255 && changePreset) currentPreset = tmpPreset;
--- a/wled00/presets.cpp
+++ b/wled00/presets.cpp
@@ -162,7 +162,7 @@ void handlePresets()
 
   presetToApply = 0; //clear request for preset
   callModeToApply = 0;
-  
+
   DEBUG_PRINTF_P(PSTR("Applying preset: %u\n"), (unsigned)tmpPreset);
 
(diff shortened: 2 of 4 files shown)


smitty078:
@coderabbitai review

DedeHai:
would a playlist not be able to do the same?
I do not see a significant speed increase, maybe 100ms.

smitty078:
> would a playlist not be able to do the same? I do not see a significant speed increase, maybe 100ms.

Thanks for the feedback! Hence why I’m still in draft mode - I’m not ready to say this is review or merge ready just yet.

1) In my initial test I thought the same thing but was unable to get a playlist to do the same thing. There may very well be a similar improvement needed for a boot playlist - I have not yet investigated that codepath but will.

2) I wonder if the speed increase I see is specific to my hardware and/or environment. My controller is at the edge of WiFi range, so maybe that impacts startup time for WiFi fairly significantly for me. I’ll update my statement on performance before submitting for review.

wled00/wled.cpp
```diff
@@ -503,6 +503,20 @@ void WLED::setup()
 
   if (needsCfgSave) serializeConfigToFS(); // usermods required new parameters; need to wait for strip to be initialised #4752
 
+  if (bootPreset > 0) {
+    if (!presetNeedsSaving()) {
```
willmmiles: How could this be true during `setup()`?

wled00/wled.cpp
```diff
@@ -503,6 +503,20 @@ void WLED::setup()
...
+      handlePlaylist(); // handle boot playlist at setup
+      yield();
```
willmmiles: Why do we need to `yield()` from `setup()`?  That seems quite strange to me -- we haven't even turned the network on yet at this point.

wled00/wled.cpp
```diff
@@ -503,6 +503,20 @@ void WLED::setup()
 
   if (needsCfgSave) serializeConfigToFS(); // usermods required new parameters; need to wait for strip to be initialised #4752
 
+  if (bootPreset > 0) {
+    if (!presetNeedsSaving()) {
+      handlePlaylist(); // handle boot playlist at setup
```
willmmiles: This will never execute because the boot preset hasn't been parsed yet, and thus no playlist can be in flight.

willmmiles (changes requested):
Re performance changes: I do expect that the latency of how long it takes the main loop to get around to applying the boot preset will be highly dependent on environment and wifi setup.  In fact, one down side of the proposed new approach is that it may lead to more devices showing visible stutters and hiccups during the connection sequence, as the light "starts up" earlier.   There's undoubtedly room for improvement there -- I suspect some of the execution order details in the current code are workarounds for issues with older platforms that may not still apply.  I don't think we should spend too much time worrying about this (either way) at the moment.

Re ps vs playlists:  there's a very real sense in which the functionality of "ps" and "playlist" overlap: "ps" could (almost) be implemented as a single-entry "playlist" with 0 delay.  In this context, both approaches require extra work at `setup()` time -- but realistically we probably should handle both, if only so that we don't have to document a bunch of special cases.

In practice I think this works out to something like:

```cpp
if (bootPreset > 0) {
   handlePreset();   // execute boot preset
   handlePlaylist();  // if boot preset set a playlist, process it
   handlePreset();   // if either of the above set a preset for deferred execution, execute it now
}
```

DedeHai:
> Re performance changes: I do expect that the latency of how long it takes the main loop to get around to applying the boot preset will be highly dependent on environment and wifi setup.  In fact, one down side of the proposed new approach is that it may lead to more devices showing visible stutters and hiccups during the connection sequence, as the light "starts up" earlier.   There's undoubtedly room for improvement there -- I suspect some of the execution order details in the current code are workarounds for issues with older platforms that may not still apply.  I don't think we should spend too much time worrying about this (either way) at the moment.
> 
> Re ps vs playlists:  there's a very real sense in which the functionality of "ps" and "playlist" overlap: "ps" could (almost) be implemented as a single-entry "playlist" with 0 delay.  In this context, both approaches require extra work at `setup()` time -- but realistically we probably should handle both, if only so that we don't have to document a bunch of special cases.
> 
> In practice I think this works out to something like:
> 
> ```cpp
> if (bootPreset > 0) {
>    handlePreset();   // execute boot preset
>    handlePlaylist();  // if boot preset set a playlist, process it
>    handlePreset();   // if either of the above set a preset for deferred execution, execute it now
> }
> ```
> 

With the reports about brownouts which may be related to faster LED turnon time (I measured roughly 1.2s vs 2.5s in 0.15) making it more consistent would be a welcome improvement, albeit this may cause some transitional friction with existing setups that have insufficient power stability

smitty078:
@coderabbitai help

smitty078:
@coderabbitai full review

wled00/presets.cpp
```diff
@@ -194,8 +206,12 @@ void handlePresets()
     changePreset = true;
   } else {
     if (!fdo["seg"].isNull() || !fdo["on"].isNull() || !fdo["bri"].isNull() || !fdo["nl"].isNull() || !fdo["ps"].isNull() || !fdo[F("playlist")].isNull()) changePreset = true;
-    if (!(tmpMode == CALL_MODE_BUTTON_PRESET && fdo["ps"].is<const char *>() && strchr(fdo["ps"].as<const char *>(),'~') != strrchr(fdo["ps"].as<const char *>(),'~')))
-      fdo.remove("ps"); // remove load request for presets to prevent recursive crash (if not called by button and contains preset cycling string "1~5~")
+    
+    // only allow load requests from boot presets that set lor or button calls
+    const bool isButtonException = (tmpMode == CALL_MODE_BUTTON_PRESET && fdo["ps"].is<const char *>() && strchr(fdo["ps"].as<const char *>(),'~') != strrchr(fdo["ps"].as<const char *>(),'~'))
+    
+    const bool shouldAllowLoadRequest = (isBootPreset && presetWillSetLor) || (tmpMode == CALL_MODE_BUTTON_PRESET && fdo["ps"].is<const char *>() && strchr(fdo["ps"].as<const char *>(),'~') != strrchr(fdo["ps"].as<const char *>(),'~'))
```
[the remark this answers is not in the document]
smitty078: this was resolved in the latest commit

wled00/presets.cpp
```diff
@@ -176,13 +176,25 @@ void handlePresets()
...
+  const bool isBootPreset = (tmpMode==CALL_MODE_INIT || tmpPreset==bootPreset);
+  const bool presetWillSetLor = (!fdo["lor"].isNull() && fdo["lor"].as<int>() > REALTIME_OVERRIDE_NONE);
+  
+  // During setup, only allow the boot preset itself or safe boot-preset chains.
+  const bool shouldAllowPresetApply = (
+                                       setupComplete || isBootPreset ||
+                                       (currentPreset == bootPreset && realtimeOverride > REALTIME_OVERRIDE_NONE) ||
+                                       (currentPreset == bootPreset && currentPlaylist > 0 && presetWillSetLor)
+                                       );
+  if (!shouldAllowPresetApply) return;
```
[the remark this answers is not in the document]
smitty078: resolved in the latest commit

smitty078:
@coderabbitai review

smitty078:
@coderabbitai review

softhack007:
> Previously, boot preset application depended on the normal loop path. `handlePresets()` in the main loop is gated behind realtime mode checks:
> 
> ```c++
> if (!realtimeMode || realtimeOverride || (realtimeMode && useMainSegmentOnly))
> ```

reminds me of another PR that is still on our stack:
* https://github.com/wled/WLED/pull/5370

PR 5370 suggests to keep presets code active while in realtime mode, so that users are able to use presets assigned to physicals buttons. I think these two PR's should be considered together, or maybe merged into one PR.

wled00/presets.cpp
```diff
@@ -176,13 +173,31 @@ void handlePresets()
   } else
   #endif
   {
-  presetErrFlag = readObjectFromFileUsingId(getPresetsFileName(tmpPreset < 255), tmpPreset, pDoc) ? ERR_NONE : ERR_FS_PLOAD;
+    presetErrFlag = readObjectFromFileUsingId(getPresetsFileName(tmpPreset < 255), tmpPreset, pDoc) ? ERR_NONE : ERR_FS_PLOAD;
```
softhack007: please undo this whitespace correction - I agree its visually unpleasing, however we have other PR in the queue that would show merge conflicts due to such indentation "fixes".

smitty078:
> > Previously, boot preset application depended on the normal loop path. `handlePresets()` in the main loop is gated behind realtime mode checks:
> > ```c++
> > if (!realtimeMode || realtimeOverride || (realtimeMode && useMainSegmentOnly))
> > ```
> 
> reminds me of another PR that is still on our stack:
> 
> * [Allow use of presets (physical button) when receiving live data #5370](https://github.com/wled/WLED/pull/5370)
> 
> PR 5370 suggests to keep presets code active while in realtime mode, so that users are able to use presets assigned to physicals buttons. I think these two PR's should be considered together, or maybe merged into one PR.

I took a brief look at this PR. I do think mine and this one go well together but I have some concerns about implementation of the other PR. I understand exactly what the author did and why - and in fact my very first instinct on this PR was to move handlePreset() and handlePlaylist() outside the realtime gate in the main loop, but ultimately I decided messing with the main loop flow was fairly dangerous, which is why I opted to add calls to setup() and some safety logic inside of handlePreset(). I think the same could be applied to the other PR by adding handlePreset(), handlePlaylist(), handlePreset() inside of handleButton() or, probably more appropriately, in the handle IO loop (which I have not looked at yet) then the same safety logic inside handlePreset() that is added with this PR can be applied to IO calls. Since I’m not the author of the other PR, I didn’t deep dive any further than this and I also don’t want to step on his toes. TL;DR I agree that other PR makes sense along side mine. I’m not certain the implementations are compatible and think there’s an opportunity to share logic. The main concern is the potential effects of altering the main loop behavior.

wled00/wled.h
```diff
@@ -890,6 +890,9 @@ WLED_GLOBAL byte optionType;
 WLED_GLOBAL bool configNeedsWrite  _INIT(false);        // flag to initiate saving of config
 WLED_GLOBAL bool doReboot          _INIT(false);        // flag to initiate reboot from async handlers
 
+// setup status
+WLED_GLOBAL bool setupComplete _INIT(false); // flag for preset loading safety
```
willmmiles: This permanently wastes runtime memory for a setup time concern.  Remove.

wled00/presets.cpp
```diff
@@ -180,9 +177,27 @@ void handlePresets()
...
+                                       (currentPreset == bootPreset && currentPlaylist > 0 && presetWillSetLor)
+                                       );
+  if (shouldAllowPresetApply) {
+    presetToApply = 0; //clear request for preset
+    callModeToApply = 0;
+  } else {
+    releaseJSONBufferLock();
+    return;
+  }
+
```
willmmiles: This check is inappropriate and should be removed.  We should not limit which presets are applied at boot.   Supporting setting the boot preset to something like `{"ps": "1~5r"}`, so it essentially randomly selects a preset between 1 and 5 at boot, would be a useful feature.

wled00/presets.cpp
```diff
@@ -194,8 +209,11 @@ void handlePresets()
     changePreset = true;
   } else {
     if (!fdo["seg"].isNull() || !fdo["on"].isNull() || !fdo["bri"].isNull() || !fdo["nl"].isNull() || !fdo["ps"].isNull() || !fdo[F("playlist")].isNull()) changePreset = true;
-    if (!(tmpMode == CALL_MODE_BUTTON_PRESET && fdo["ps"].is<const char *>() && strchr(fdo["ps"].as<const char *>(),'~') != strrchr(fdo["ps"].as<const char *>(),'~')))
-      fdo.remove("ps"); // remove load request for presets to prevent recursive crash (if not called by button and contains preset cycling string "1~5~")
+    
+    // only allow load requests from boot presets that set lor or button calls
+    const bool isButtonException = (tmpMode == CALL_MODE_BUTTON_PRESET && fdo["ps"].is<const char *>() && strchr(fdo["ps"].as<const char *>(),'~') != strrchr(fdo["ps"].as<const char *>(),'~'));
+    const bool shouldAllowLoadRequest = (isBootPreset && presetWillSetLor) || (tmpMode == CALL_MODE_BUTTON_PRESET && fdo["ps"].is<const char *>() && strchr(fdo["ps"].as<const char *>(),'~') != strrchr(fdo["ps"].as<const char *>(),'~'));
+    if (!shouldAllowLoadRequest) fdo.remove("ps"); // remove load request for presets to prevent recursive crash
```
willmmiles: Your previous commit's logic of checking "CALL_MODE_INIT" was simpler and cleaner.  Also the `strchr` tests are should only appear in the code once - please avoid copy-and-pasting complex constructs.

To avoid unbounded recusion, the better solution is to patch the site `json.cpp` where `applyPreset()` is called to avoid propagating `CALL_MODE_INIT` and instead switch to `CALL_MODE_DIRECT_CHANGE`.  (Playlists, and the sites in for the legacy syntax `set.cpp`, will already change the call mode.)

willmmiles (changes requested):

wled00/presets.cpp
smitty078: Existing path:
boot preset is set -> realtime mode starts -> main loop queues boot preset -> realtime mode ends -> boot preset is applied

New path:
boot preset is set -> if boot preset overrides realtime mode, boot preset is applied immediately, if not, the behavior above applies (this is why it is gated like this)

Proposed change:
Same as "New path" if boot preset sets realtime override
If boot preset does not set realtime override:
boot preset is applied -> realtime starts -> transition from boot preset to realtime

This could make sense, but it introduces an unexpected transition from the boot preset to the realtime data. For a user that wants to boot into realtime data, there is no longer a way to achieve this without turning off boot preset, which means that there is no preset state in the queue for when/if realtime ends.

This is specifically why I gated it the way that it did. Perhaps the documentation is a little unclear - it's not that the boot preset is never getting applied when it does not set lor, it's that it remains queued until realtime is ended, which preserves the previously existing behavior.

wled00/wled.h
smitty078: Based on your other comments, I'm going to investigate using CALL_MODE_INIT as the gate for everything. I believe I can remove this if I can also ensure that CALL_MODE_INIT vs CALL_MODE_DIRECT_CHANGE is used properly everywhere else. Updates to come.

wled00/presets.cpp
smitty078: Thanks for pointing me at `json.cpp` and `set.cpp`. Updates to follow.

wled00/presets.cpp
willmmiles: > boot preset is set -> realtime mode starts -> main loop queues boot preset 

This is a nondeterministic race, and this outcome is not and cannot be guaranteed -- whether the selected boot preset is applied by the main loop before or after processing the first event that triggers real-time mode will be sensitive to local network timing and subtle code timing changes in WLED or the platform toolkit we depend on.  I don't think we should encourage users to bet on races, regardless of how often they might win in their particular configurations.
 
Always applying the boot preset in `setup()` has the major advantage of being strictly deterministic about the starting state.  (As well as offering a guarantee that the "first light" state will be exactly the user preset state instead of only "off" or "yellow" -- this has been a source of confusion in the past!)

If someone wants the behavior of "start up and wait for a little bit to allow real-time mode to start, then fall back to preset N", I think this could be implemented by setting a boot preset with a little playlist, eg. `{"playlist":{"ps":[n_waiting, n_fallback], "dur": [ wait_time, 1]}}`, where `n_waiting` points to the waiting state preset, and `n_fallback` is the "fallback after live mode ends" preset.

smitty078: I tend to agree with you. I believe I overthought this in an attempt to preserve existing behavior as much as possible rather than go all in on new behavior. The reality is that the new behavior is better, and that's the whole point. Check my latest commit.

wled00/presets.cpp
smitty078: Latest commit addresses this. The doubling was an error to begin with, but the latest commit reverts to the simpler CALL_MODE_INIT method anyway.

wled00/wled.h
smitty078: check latest commit

wled00/json.cpp
```diff
@@ -520,6 +520,7 @@ bool deserializeState(JsonObject root, byte callMode, byte presetId)
     if (root["win"].isNull() && getVal(root["ps"], presetCycCurr, 1, 250) && presetCycCurr > 0 && presetCycCurr < 251 && presetCycCurr != currentPreset) {
       DEBUG_PRINTF_P(PSTR("Preset select: %d\n"), presetCycCurr);
       // b) preset ID only or preset that does not change state (use embedded cycling limits if they exist in getVal())
+      if (callMode == CALL_MODE_INIT) callMode = CALL_MODE_DIRECT_CHANGE // avoid propogating CALL_MODE_INIT, which may cause accidental recursion.
```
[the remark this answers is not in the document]
smitty078: fixed in latest commit

smitty078:
@coderabbitai resume

wled00/json.cpp
```diff
@@ -520,6 +520,7 @@ bool deserializeState(JsonObject root, byte callMode, byte presetId)
     if (root["win"].isNull() && getVal(root["ps"], presetCycCurr, 1, 250) && presetCycCurr > 0 && presetCycCurr < 251 && presetCycCurr != currentPreset) {
       DEBUG_PRINTF_P(PSTR("Preset select: %d\n"), presetCycCurr);
       // b) preset ID only or preset that does not change state (use embedded cycling limits if they exist in getVal())
+      if (callMode == CALL_MODE_INIT) callMode = CALL_MODE_DIRECT_CHANGE; // avoid propogating CALL_MODE_INIT, which may cause accidental recursion.
```
willmmiles: Instead of altering the global state here, pass the updated value to `applyPreset`, and let the function decide how to store it.  It'll make it easier to manage when we get to refactoring the notification system.

wled00/presets.cpp
```diff
@@ -282,4 +282,5 @@ void deletePreset(byte index) {
   writeObjectToFileUsingId(getPresetsFileName(), index, &empty);
   presetsModifiedTime = toki.second(); //unix time
   updateFSInfo();
-}
\ No newline at end of file
+}
+
```
willmmiles: Looks like an unexpected whitespace change slipped in here?

willmmiles (changes requested):
It always warms my heart when the necessary patch turns out to be so minimal after all is said and done.  Nice job.

wled00/json.cpp
smitty078: done.

smitty078:
> It always warms my heart when the necessary patch turns out to be so minimal after all is said and done. Nice job.

Thanks, and thank you for being an attentive, helpful maintainer. Too often maintainers are simply jerks or don't care or both. The team and the project here make me want to continue to contribute.

wled00/presets.cpp
smitty078: fixed.

willmmiles (approved):

PR #5580, opened 2026-05-07
Fix custom tooltip showing "null" on sliders with child elements

### LINKED ISSUE #5578: Tooltip on top slider shows "null"
### What happened?

The top slider tooltip shows "null", no matter what I do. This happens in latest Firefox on macOS.

Image

### To Reproduce Bug

1. Open the Web interface
2. Hover over the top slider

### Expected Behavior

Should probably show something like "Value/Brightness"

### Install Method

Binary from WLED.me

### What version of WLED?

0.16.0

ESP8266

### Relevant log/trace output

```shell

```

### Anything else?

_No response_

### Code of Conduct

pReya:
closes #5578 

**Summary**

- Switch the custom tooltip handler in index.js from pointerover/pointerout to pointerenter/pointerleave so it fires exactly once per real entry/exit.

**Problem**

The custom tooltip implementation at `wled00/data/index.js` saves an element's title attribute to data-title on hover, removes title (to suppress the native browser tooltip), and restores it from data-title on exit.

Because pointerover bubbles, the handler fired again whenever the cursor crossed onto a child element (e.g. moving from the sliderwrap div onto its <input> or .sliderdisplay child). On that second invocation getAttribute("title") returned null (it had already been removed), and setAttribute("data-title", null) stored the literal string "null". Subsequent hovers then displayed a tooltip reading "null".

This was reproducible on the Value/Brightness slider on the Colors tab, and could affect any element with a title attribute that contains child nodes the cursor can land on.

**Fix**

pointerenter/pointerleave do not bubble — they fire only on the actual entry to and exit from the listening element, regardless of internal cursor movement. This makes the handler idempotent for a single hover and matches the intended semantics.

**AI Disclaimer:** Claude Code was used to find and fix this bug and create this PR description. However it's not "vibe coded" as I understand the code completely and have reviewed and tested it manually.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tooltip event handling to prevent unintended tooltip triggers and dismissals in edge cases.

### DIFF
--- a/wled00/data/index.js
+++ b/wled00/data/index.js
@@ -3181,7 +3181,7 @@ function mergeDeep(target, ...sources)
 function tooltip(cont=null)
 {
 	d.querySelectorAll((cont?cont+" ":"")+"[title]").forEach((element)=>{
-		element.addEventListener("pointerover", ()=>{
+		element.addEventListener("pointerenter", ()=>{
 			// save title
 			element.setAttribute("data-title", element.getAttribute("title"));
 			const tooltip = d.createElement("span");
@@ -3206,7 +3206,7 @@ function tooltip(cont=null)
 			tooltip.classList.add("visible");
 		});
 
-		element.addEventListener("pointerout", ()=>{
+		element.addEventListener("pointerleave", ()=>{
 			d.querySelectorAll('.tooltip').forEach((tooltip)=>{
 				tooltip.classList.remove("visible");
 				d.body.removeChild(tooltip);


DedeHai:
thanks for spotting and fixing this. The original code works for me on chrome and safari. On what setup did it not work?

pReya:
> thanks for spotting and fixing this. The original code works for me on chrome and safari. On what setup did it not work?

The original bug appeared in Firefox for me.

DedeHai:
Not seeing it in firefox either...

DedeHai:
Looks like a good update in general, I can't judge if it would have any negative side effects. Maybe the rabbit can dig a little.

DedeHai:
@coderabbitai does this change have any unforseen consequences? Please evaluate for all major browsers and operating systems.

pReya:
I can't recreate the initial problem anymore, myself. I only cared to validate the fix (which obviously works, when the initial problem isn't even there). I don't know how I managed to create the initial error state in the first place (it just happened during regular usage). Since I can't recreate it, I don't think anything should be fixed, until we know what the initial problem was. So I'm closing this PR, but I'll keep the Issue open if that's okay. Maybe other people will come encounter this same error state.

Issue #4334, opened 2024-12-01
Drip doesn't use palette


djdevin:
### What happened?

Drip is marked as using a palette but only solid colors can be used

### To Reproduce Bug

Set drip effect, select palette

### Expected Behavior

Effect uses selected palette

### Install Method

Binary from WLED.me

### What version of WLED?

0.14

### Which microcontroller/board are you seeing the problem on?

ESP8266, ESP32

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

DedeHai:
easy fix: remove palette setting

djdevin:
It would be nice if it could use palette! Could make a winter/ice setting with blues and whites.

Should be marked as a feature then.

softhack007:
> I have a new, much enhanced version of the effect that supports palette. It may come into WLED soon.

@DedeHai maybe I can merge my MoonModules changes with your version? I've also added a few extras - mainly to make the effect look better on large panels (slowed it down, added palette support, made drops slightly more random).

https://github.com/MoonModules/WLED/blob/796eb6e23e493cfce9b942fc950afdfebd83deac/wled00/FX.cpp#L3747

DedeHai:
@softhack007 my version uses the particle system... so not really compatible

softhack007:
> the question is, should the current 1D FX be replaced with the PS versions I made or should both coexist.

Can your effect also support virtual strips ("Bars" mapping where each bar is treated like an effect instance) ? 
![image](https://github.com/user-attachments/assets/f7edf286-27af-4110-8479-9bc49eab0ca8)

If yes, then maybe we just need one effect based on the particle system.

DedeHai:
currently, bar mapping/virtual strips is not possible, that is something I need to look into how to solve once the rest of the code is done

PR #5583, opened 2026-05-09
Log buffer

netmindz:
This pull request introduces a PSRAM-backed diagnostic log buffer for WLED, enabling users to view, copy, and clear runtime logs from the web UI on devices with PSRAM, without requiring serial access or debug builds. It also updates usermod ID documentation, improves debug macro routing.

**Diagnostic Log Buffer and Web UI Integration:**

* Added a PSRAM-backed ring buffer (`log_buffer.h`) for capturing log output via new `wledLogBuffer` and `wledLog` objects; logs are exposed through a new `/json/log` HTTP endpoint and a `/log` web UI page, available only on devices with PSRAM. [[1]](diffhunk://#diff-e2e356edc33aea4b696ecd3c477faed89f01f4d13e5aaa0a87faa73b744c61a7R1-R114) [[2]](diffhunk://#diff-c921278adaae7a31dfc233a0045f050bf53cfd6a794e3e3802431ca4699bfab0R977-R988) [[3]](diffhunk://#diff-c921278adaae7a31dfc233a0045f050bf53cfd6a794e3e3802431ca4699bfab0L993-R1021) [[4]](diffhunk://#diff-c921278adaae7a31dfc233a0045f050bf53cfd6a794e3e3802431ca4699bfab0R1030) [[5]](diffhunk://#diff-81fb86539a4966e6e5a2df3e463988478732991e5db8bdb763dc340ab0e2462bL427-R433) [[6]](diffhunk://#diff-6b6413edc70eb5a8b8d037f55e51154ae77788f431721701e949d2f00cd99820R1317-R1319) [[7]](diffhunk://#diff-6b6413edc70eb5a8b8d037f55e51154ae77788f431721701e949d2f00cd99820R1386-R1413) [[8]](diffhunk://#diff-5b6c03a5c1680ec3273bde4934ce9adb4342e0a0314b9c2c607c8099be66d4ecR23-R26) [[9]](diffhunk://#diff-5b6c03a5c1680ec3273bde4934ce9adb4342e0a0314b9c2c607c8099be66d4ecR625-R636) [[10]](diffhunk://#diff-4b583ec520b3838172923ce22617255549632d45275fd138d1dff716231e8365L29-R29) [[11]](diffhunk://#diff-4b583ec520b3838172923ce22617255549632d45275fd138d1dff716231e8365R481-R495) [[12]](diffhunk://#diff-583cf41c06362dc471369257b8001b1227273c0f23a108394242e5b041a3d940R1-R164)

* Added a log viewer page (`log.htm`) with refresh, auto-refresh, copy, and clear functionality, and integrated it into the settings UI (hidden unless PSRAM is present). [[1]](diffhunk://#diff-856865b40c2d120fe12084dbfb95eab86ddaf95dcec190e56ad0d8e9d042f333R54) [[2]](diffhunk://#diff-ea99264d317ef01ab535954ced6325bb08ff41b206092908bcaebbcd3487731bR219-R221) [[3]](diffhunk://#diff-583cf41c06362dc471369257b8001b1227273c0f23a108394242e5b041a3d940R1-R164)

**Debug Output Routing:**

* Updated debug macros so that, on PSRAM builds, all debug output is captured in the ring buffer (and also output to Serial/NetDebug if `WLED_DEBUG` is enabled); on non-PSRAM builds, behavior is unchanged or disabled as appropriate. [[1]](diffhunk://#diff-c921278adaae7a31dfc233a0045f050bf53cfd6a794e3e3802431ca4699bfab0L993-R1021) [[2]](diffhunk://#diff-c921278adaae7a31dfc233a0045f050bf53cfd6a794e3e3802431ca4699bfab0R1030)

**Usermod ID Documentation:**

* Clarified in `AGENTS.md` and `const.h` that usermod IDs are only required for inter-usermod communication, pin ownership, or JSON info identification, reducing unnecessary ID assignments. [[1]](diffhunk://#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9L162-R172) [[2]](diffhunk://#diff-749825f808f7ce1ce1108fffc7ed037085160a63cad32397135dfc0063494373L169-R175)

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated log viewer page accessible from the settings interface
  * Log viewer displays device diagnostic logs with manual refresh, auto-refresh (5-second intervals), clear, and copy-to-clipboard functions
  * Live log display shows current buffer size in kilobytes
  * Feature available on devices with PSRAM support

[Review Change Stack](https://app.coderabbit.ai/change-stack/wled/WLED/pull/5583?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


willmmiles:
From an architecture review level:
- The log buffer feature should be available on all platforms instead of gated on PSRAM presence.  We should instead make the log ring buffer size configurable, and default to 0/off on memory constrained platforms.  This will allow users to selectively enable logging when required for troubleshooting.
- `a.printf(...); b.printf(...)` is expensive.  It's better to multiplex inside the `Print` virtual functions.  Particularly in the context of the above, 
ideally we want to route logging output (for debugging or otherwise) as a first-class feature of the logging subsystem, instead of network debugging being a compile-time switch.  (There's a PR for `syslog` style log output IIRC too, and I guarantee someone's going to ask for MQTT at some point....)
- Use C++ mutexes with safer constructs like guard objects instead of platform-specific C calls.  There's no excuse for using C synchronization primitives; that's just asking for more footguns.  (Also we don't need interrupt-safe spinlocks; Arduino Print infrastructure is not really a good fit for ISRs even under ideal circumstances.)

netmindz:
Sounds like you are in favour of the idea @willmmiles , I think it was actually something you or @DedeHai suggested on a previous call

Do you have any interest/capacity to help take this poc into final code @willmmiles ?

netmindz:
It would also be nice if we can implement an SSE endpoint, rather than polling. That would make this useful for devices with very small ringbufders

willmmiles:
> Sounds like you are in favour of the idea @willmmiles , I think it was actually something you or @DedeHai suggested on a previous call
> 
> Do you have any interest/capacity to help take this poc into final code @willmmiles ?

Yeah, I've talked about it a couple times.  So interest, yes; capacity, hard to say.

> It would also be nice if we can implement an SSE endpoint, rather than polling. That would make this useful for devices with very small ringbufders

That's an interesting idea.  AWS has an SSE implementation, but it operates by copying every message in to a temporary outbound buffer for every client -- or in other words, it'd be fairly memory hungry itself.   The upstream AWS version is a bit cleaner in that it at least uses `shared_ptr` to keep only one copy of any given message at a time.  It could probably be hacked to do something `weak_ptr`-like addressing the ring buffer contents directly, though how to account for outstanding pointers would be a trick.  (Maybe a second, small message-identity ring buffer with a generation counter.)

netmindz:
What do we want to do with this then @willmmiles ?

Are you building something yourself based on this idea and I close this one off? Should I transfer the branch to the WLED org repo so we can both work on it?

willmmiles:
I started building something bigger based on this branch, and it spun off in enough interesting features that I'd wanted to break it into smaller chunks.  I can push some of the updated draft code to a common branch, but it's also still very rough-AI-draft-y and needs more attention to get the API right.  I'm targeting this after the V5 merge and the usermod UI work -- the WIP framework is extensible so the various outputs (serial, network socket, syslog #4334, and the ring buffer web ui) can be modularized.

PR #5590, opened 2026-05-10
fix(Fix_unreachable_netservices_v2): v16 compatibility fixes

Joeboyc2:
## Summary

Two issues prevent `Fix_unreachable_netservices_v2` from building under WLED v16:

- **Missing `libArchive: false` in `library.json`** — the v16 build system requires `"build": {"libArchive": false}` in every usermod's `library.json`. Without it the build fails immediately with `ERROR: libArchive=false is missing on usermod(s)`.

- **`serializeConfig()` no longer has a no-argument overload** — v16 changed `serializeConfig()` to require a `JsonObject` parameter. The correct no-argument replacement for saving config to the filesystem is `serializeConfigToFS()`.

## Test plan

- [x] `Fix_unreachable_netservices_v2` compiles cleanly in an ESP8266 build environment with both fixes applied
- [x] Verified `serializeConfigToFS()` is the correct equivalent — it allocates the JSON document internally and calls `serializeConfig(root)` + writes to LittleFS

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where network service configuration updates were not being properly persisted to storage.

* **Chores**
  * Updated library build configuration settings.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wled/WLED/pull/5590)

### DIFF
--- a/usermods/Fix_unreachable_netservices_v2/library.json
+++ b/usermods/Fix_unreachable_netservices_v2/library.json
@@ -1,4 +1,5 @@
 {
   "name": "Fix_unreachable_netservices_v2",
+  "build": { "libArchive": false },
   "platforms": ["espressif8266"]
 }
--- a/usermods/Fix_unreachable_netservices_v2/usermod_Fix_unreachable_netservices.cpp
+++ b/usermods/Fix_unreachable_netservices_v2/usermod_Fix_unreachable_netservices.cpp
@@ -77,7 +77,7 @@ public:
     }
     if (m_updateConfig)
     {
-      serializeConfig();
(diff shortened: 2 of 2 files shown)


softhack007 (approved):
Looks good for me - I don't have any 8266 for testing, however i've checked that serializeConfigToFS() is a proper replacement. In fact, `configNeedsWrite = true;` would have been sufficient to trigger config writing in the main loop, but it should also work as proposed in this PR.

Joeboyc2:
Thanks Frank,
Ive deployed this to 3 of my esp8266s devices and all seems to working well

Issue #5562, opened 2026-05-04
"Power button preset override" for on/off are retained after Factory Reset


Yersi88:
### What happened?

"Power button preset override for On:" setting is retained between Factory Resets. 

### To Reproduce Bug

1. Change "Power button preset override for On:" to value other than default 0. 
2. Apply settings. 
3. Choose 'Factory Reset' in Update Settings Menu. 
4. Apply settings. 
5. "Power button preset override for On:" remains at value you set it to in step 1.

### Expected Behavior

Setting resets to default value of 0 after factory reset. 

### Install Method

Binary from WLED.me

### What version of WLED?

WLED version: 16.0.0 (2605030)

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DedeHai:
both factory reset path's call
` WLED_FS.format();
doReboot = true;`

so either this is a bug in littleFS or it is async and gets halted by the reboot. I have no other explanation that would not wipe the file system flash. 
@willmmiles do you see any other reason?

softhack007:
@DedeHai @willmmiles just a wild guess - could it be that writing to cfg.json is needed (asyncWebserver task is in ``serializeConfig()``), then the filesystem is formated by the main task (cfg.json write still pending/preparing JSON), and then the pending write happens directly after formatting but before the reboot executes (next turn in main loop) ?

DedeHai:
@coderabbitai please trace the factory reset and confirm / disprove our theories.

softhack007:
lessons learned: global state is a nightmare 😓 

We could try the "fix covers both paths" from the rabbit - it does not make much sense to keep a config write pending after FS re-formating.

DedeHai:
@coderabbitai in the webUI path, is the `configNeedsWrite` always set when confirming the factory reset through the UI or do conditions apply? since this is the path the issue is about it would mean that the race condition is met every time - or it is pretty much an edge case hit by bad luck.

DedeHai:
given that it is almost impossible to have configNeedsWrite set prior to doing a factory reset through the UI I would call the rabbits analysis either incomplete or wrong. The finding of playing it save in the main loop is still a good idea.  If it is in fact the root cause for the issue raised this is an extreme edge case.

softhack007:
[...]

PR #5591, opened 2026-05-10
save global on / off button override in cfg

DedeHai:
as discussed in https://github.com/wled/WLED/issues/5562 this adds saving of the UI setting in config in addition to localstorage.

@softhack007 I am not too sure this is a good idea - at least as is: it prevents users from using different on/off presets on different devices. This could be solved with a checkmark to use "local override only", ignorie the cfg setting and use localstorage only.

the other way to solve #5562 would be to not store them in config but add a mechanism to tell the browser to clear relevant local storage after a factory reset, I am not sure how to do that though as it requires a variable the persists a reboot so it would need to go into the config or use RTC memory. Clearing browser cache for a UI induced factory reset would be possible by adding some code to settings_sec.htm but doing it for a button called factory reset seems a bit more tricky. 

## Summary by CodeRabbit

* **New Features**
  * Power button preset configuration – users can now configure custom preset actions that automatically execute when the power button is pressed, with separate settings for on and off states enabling flexible personalized automation
  * Enhanced configuration synchronization between device and web interface ensuring settings persist consistently across restarts

[Review Change Stack](https://app.coderabbit.ai/change-stack/wled/WLED/pull/5591)

### DIFF
--- a/wled00/cfg.cpp
+++ b/wled00/cfg.cpp
@@ -58,6 +58,8 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
   getStringFromJson(alexaInvocationName, id[F("inv")], 33);
 #endif
   CJSON(simplifiedUI, id[F("sui")]);
+  CJSON(powerButtonOnPreset, id[F("pon")]);
+  CJSON(powerButtonOffPreset, id[F("pof")]);
 
   JsonObject nw = doc["nw"];
 #ifndef WLED_DISABLE_ESPNOW
@@ -853,6 +855,8 @@ void serializeConfig(JsonObject root) {
   id[F("inv")] = alexaInvocationName;
 #endif
   id[F("sui")] = simplifiedUI;
+  id[F("pon")] = powerButtonOnPreset;
+  id[F("pof")] = powerButtonOffPreset;
 
   JsonObject nw = root.createNestedObject("nw");
 #ifndef WLED_DISABLE_ESPNOW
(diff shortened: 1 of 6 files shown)


softhack007:
> @softhack007 I am not too sure this is a good idea - at least as is: it prevents users from using different on/off presets on different devices. This could be solved with a checkmark to use "local override only", ignorie the cfg setting and use localstorage only.

@DedeHai yes I see your concern, too. If configuring different on/off presets per UI instance (=different on your Pad, on your MacBook, on your son's PC) is a feature that people use, we would create a regression. I also agree that a "complete solution" - that also clears the browser's local storage on reset-by-button, plus clears _all_ connected browsers - would be an overkill in terms of complexity.

Maybe let's ask users (discord) how they use the feature 🤔 
* Are our users aware that the setting is "per UI" ? is this useful? or maybe its just a PITA?
* Maybe we should simply document that "UI settings" are not reset on factory reset? This could be added to the reset confirmation page, and written down in wled-docs.
* Generally I think that "UI settings" is an unusual place for overriding the "on/off virtual button" behaviour. It might be more logical to have this in LEDs settings, next to "Apply preset on boot". Or maybe next to "button macros" in time&button settings.
* Finally, why not simply assign "transition styles" to the on/off button, instead of using presets? (like these "banlanx" non-wled controllers usually do)


@netmindz @lost-hope @willmmiles @Moustachauve what do you think?

PR #5612, opened 2026-05-13
rename CW LED descriptions to more commonly used CCT

### LINKED ISSUE #5597: Request direct IC configuration for WS2811 RGBCCT LED strip
WLED supports WS2811 RGBCCT, but it needs to be set to FW1906 in the app, which is not very intuitive. 
Most users will simply set it to WS2811 RGB, causing the LED strip to malfunction and leading them to believe that WLED does not support the LED strip.

WS2811 RGBCCT led strip link:
https://www.aliexpress.com/item/1005009812073953.html

DedeHai:
this is a suggestion to rename "CW" into "CCT"

@intermittech what do you think? Open for discussion

## Summary by CodeRabbit

* **Documentation**
  * Updated LED type display names returned by the LED types JSON API to more accurately reflect supported color-channel layouts, with clearer RGBCCT/RGBW wording for relevant bus types.
  * Revised inline channel labels and descriptive text for several supported LED bus formats so the displayed names align consistently with actual capabilities.

### DIFF
--- a/wled00/bus_manager.cpp
+++ b/wled00/bus_manager.cpp
@@ -358,19 +358,19 @@ std::vector<LEDType> BusDigital::getLEDTypes() {
     {TYPE_FW1906,        "D",  PSTR("RGBCCT FW1906/WS2811")},
     {TYPE_WS2805,        "D",  PSTR("RGBCCT WS2805")},
     {TYPE_SM16825,       "D",  PSTR("RGBCCT SM16825")},
-    {TYPE_WS2812_1CH_X3, "D",  PSTR("WS2811 White")},
-    {TYPE_TM1814,        "D",  PSTR("TM1814")},
-    {TYPE_TM1829,        "D",  PSTR("TM1829")},
-    {TYPE_UCS8903,       "D",  PSTR("UCS8903")},
-    {TYPE_APA106,        "D",  PSTR("APA106/PL9823")},
-    {TYPE_TM1914,        "D",  PSTR("TM1914")},
+    {TYPE_WS2812_1CH_X3, "D",  PSTR("White WS2811")},
+    {TYPE_TM1814,        "D",  PSTR("RGBW TM1814")},
+    {TYPE_TM1829,        "D",  PSTR("RGB TM1829")},
+    {TYPE_UCS8903,       "D",  PSTR("RGB UCS8903")},
+    {TYPE_APA106,        "D",  PSTR("RGB APA106/PL9823")},
+    {TYPE_TM1914,        "D",  PSTR("RGB TM1914")},
     //{TYPE_WS2812_2CH_X3, "D",  PSTR("WS281x CCT")}, // not implemented
-    {TYPE_WS2812_WWA,    "D",  PSTR("WS281x WWA")}, // amber ignored
-    {TYPE_WS2801,        "2P", PSTR("WS2801")},
-    {TYPE_APA102,        "2P", PSTR("APA102")},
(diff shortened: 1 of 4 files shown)


intermittech:
Oh, I am 100% for that!

Any term other than RGBCCT or RGB+CCT is confusing in my opinion. WW is generally used for "Warm White" and CW is often used for "Cold White", that's fine. Things we're okay up until RGBW but then people starting calling it RGBWW or RGBCW sometimes, denoting the particular color temperature of that strip but there things went wrong, sometimes using RGBCW as referring to dual white also since the C and W is in there, very confusing. I've also seen it written as RGBWWCW, less confusing but still a bit meh in my opinion.

CCT is clear what it means, dual white color temperature
RGBCCT is also clear what it means, R+G+B with dual white

RGBW stays as is, R+G+B with a single white (whatever temperature)

I sometimes use the distinction of RGBCCT and RGB+CCT if the white is integrated into the same package, or external, but for the software side, that really doesn't matter, CCT and RGBCCT is the correct way to do it in my opinion.

wled00/bus_manager.cpp
```diff
@@ -359,10 +359,10 @@ std::vector<LEDType> BusDigital::getLEDTypes() {
     {TYPE_UCS8903,       "D",  PSTR("UCS8903")},
     {TYPE_APA106,        "D",  PSTR("APA106/PL9823")},
     {TYPE_TM1914,        "D",  PSTR("TM1914")},
-    {TYPE_FW1906,        "D",  PSTR("FW1906 GRBCW")},
+    {TYPE_FW1906,        "D",  PSTR("RGBCCT FW1906/WS2811")},
```
softhack007: (Minor) this entry uses "RGBCCT" as prefix to the chip name, but the other entries have that as postfix (after the chip name). Was this intended?

softhack007 (approved):
Makes sense for me 👍
Same as @intermittech says, I remember there were different namings around when RGB strips got white as an extra LED. CCT seems to be the term that "survived".

**To be discussed**: maybe we could always use the RGB+" naming?
* RGB (no white)
* RGB+W (single white - either cold or warm white)
* RGB+CCT (both warm and cold white)
* White (single white)
* CCT (cold+warm)

Not sure what to do with "CCT + Amber" strips that have _three_ white channels?

wled00/bus_manager.cpp
DedeHai: yes, as it applies to multiple strips, could also add this to the others that are frequently used, i.e.  

- RGB WS281x
- RGB 400kHz
- RGBW SK6812/WS2814
- RGBW UCS8904
- RGBCCT WS2805 (note: uses slightly faster timing)
- RGBCCT SM16825

etc.

intermittech:
Oh that is a good one @softhack007, there is a SK6812 WWA which has Warm, Cold and Amber White, I'd say it's so much of an exception (and the only one of that sort as far as I am aware) that it shouldn't define the naming for the rest. So maybe we just keep it as an exception to the rule, I think that's fine.

It seems the PWM naming was done by another developer and that it's 5 channel mode is called RGB+CCT, maybe just clean it all up and make it the same and I'd personally choose for RGBCCT then.

softhack007:
> It seems the PWM naming was done by another developer and that it's 5 channel mode is called RGB+CCT, maybe just clean it all up and make it the same and I'd personally choose for RGBCCT then

@DedeHai while we are at this, should we also change the "analogue" naming scheme?

DedeHai:
cherry picked into 16_x

PR #5625, opened 2026-05-17
Fix brighness not being applied properly in nightlight mode

### LINKED ISSUE #5620: nightlight problem with brightness transitions (for small changes)
### What happened?

The brightness transition in `handleTransitions()` was changed from float to integer math between v0.15.x and v16.0.0. The integer version uses C's truncation-toward-zero division, which causes small brightness deltas to produce zero change, stalling the transition indefinitely.

This is most visible during nightlight fade at low starting brightness, where the brightness appears stuck and then snaps off at the end.

### Root cause

v0.15.x used float math:
```cpp
float tper = (millis() - transitionStartTime) / (float)strip.getTransition();
briT = briOld + ((bri - briOld) * tper);
```

v16.0.0 uses integer math:
```cpp
int deltaBri = (int)bri - (int)briOld;
briT = briOld + (deltaBri * ti / tr);
```

For a brightness change of -1 with default transition time of 700ms, at 100ms into the transition:
- **Float (v0.15.x):** `4 + (-1 * 0.143) = 3.86` -> truncated to `briT = 3`. Works.
- **Integer (v16.0.0):** `4 + (-1 * 100 / 700) = 4 + (0) = 4`. C truncates `-100/700` toward zero, giving 0. **briT never changes.**

This interacts badly with nightlight, which calls `colorUpdated()` every 100ms (via `handleNightlight`). Each call restarts the transition timer in `stateUpdated()` (`transitionStartTime = now`). Since the integer math produces zero change in 100ms for small deltas, and the transition is restarted before it can complete, `briT` is permanently stuck at its starting value.

The transition completes only when `ti >= tr` (700ms elapsed), but nightlight restarts it every 100ms, so this condition is never reached. The brightness remains stuck until the nightlight period expires and forces bri=0.

### Fix

Use floor division instead of C's default truncation-toward-zero for the interpolation. This matches the v0.15.x float behavior where fractional results truncate downward:

```diff
     int deltaBri = (int)bri - (int)briOld;
-    briT = briOld + (deltaBri * ti / tr);
+    long interp = (long)deltaBri * ti;
+    if (interp < 0) interp -= (tr - 1); // floor division instead of truncation toward zero
+    briT = briOld + interp / tr;
```

With this fix, `deltaBri=-1` at `ti=1ms` gives: `(-1 - 699) / 700 = -1`, so `briT` immediately steps to the target value — matching v0.15.x behavior where the float result (e.g., 3.99) truncated to 3 on the first frame.

### Environment

- WLED v16.0.0
- Default transition time: 700ms
- Nightlight mode: fade or color fade

### To Reproduce Bug

1. Set any effect, brightness to 4-10.
2. Enable nightlight fade to 0 over 1 minute
3. **Expected:** Brightness steps down gradually (4, 3, 2, 1, 0)
4. **Actual:** Brightness stays at starting value, then snaps to 0 when nightlight ends

Note the reported `bri` brightness and UI does drop, just the observed LED brightness does not.

This affects any brightness transition where `|deltaBri| * updateInterval < transitionTime`. With the default 700ms transition and 100ms nightlight update interval, any brightness change of less than 8 levels stalls.

### Expected Behavior

LEDs reduce to very low brightness before finally turning off.

### Install Method

Binary from WLED.me

### What version of WLED?

16.0

ESP32

### Relevant log/trace output

```shell

```

### Anything else?

_No response_

### Code of Conduct

DedeHai:
fix for #5620 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved nightlight fade and colorfade behavior by temporarily pausing strip transitions during the fade, applying the interpolated color without notifications, then restoring the prior transition timing.
  * Fixed brightness finalization when transitions are inactive by clearing pending transition-related state and ensuring the final brightness is applied reliably after related updates complete.


DedeHai:
in order to properly fix it, transitions should be disabled in nighlight mode. @blazoncek has fixed it but the change is complex and I did not yet fully work through that change to port it upstream.

DedeHai:
I changed this into a hack-fix which is good for now but needs proper rework of how nightlight mode works.
ready to merge for the bugfix release.

DedeHai:
cherry picked to 16_x

PR #5630, opened 2026-05-19
support for building inside docker

ondras:
Added a `Dockerfile` to create an image with building infrastructure.
Added a `compose.yaml` to build+run that image in order to create a binary in a clean environment.

## Summary by CodeRabbit

* **Chores**
  * Added a containerized build environment to provide consistent, reproducible builds across machines.
  * Added a compose-based service to run the build and output artifacts to a local folder, with a configurable build target.
  * Excluded common VCS/editor files, logs, temporary files, dependencies, and build output from the build context to speed up image builds and reduce image size.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wled/WLED/pull/5630?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /workdir
+
+COPY . .
+
+RUN npm ci
+RUN pip install --break-system-packages --ignore-installed -r requirements.txt
+
+ENV WLED_RELEASE=True
```
softhack007: what is `WLED_RELEASE` needed for? 

Its not checked by platformio or pio-scripts, neither referenced anywhere in our nightly CI jobs. SO maybe just remove it, to avoid future problems.

.dockerignore
```diff
@@ -0,0 +1,8 @@
+.git
+.github
+.vscode
+.devcontainer
+*.log
+*.tmp
+node_modules/
+build_output/
```
softhack007, line 8: does this mean docker will ignore any firmware.bin files created by pio?

Dockerfile
ondras: Will remove it then. This was suggested in the intially drafted Dockerfile in the Discord conversation, I just copypasted it.

.dockerignore
ondras: Well, they will not be copied inside the image (during the image build process).

But they will happen to be available (accidentally) during the actual firmware build phase, because the `compose.yaml` file bind-mounts the local `build_output` into the container (because we want to be able to retrieve those `.bin` files afterwards).

willmmiles:
Why do we need a second container definition?  There's already a Dockerfile in the `.devcontainer` folder, where it belongs.

netmindz:
This is redundant duplication, we already have a devcontainer


https://github.com/wled/WLED/tree/main/.devcontainer

ondras:
Uhm... may I ask for some explanation here? The [Dockerfile](https://github.com/wled/WLED/blob/main/.devcontainer/Dockerfile) in question is basically empty, not suitable for building and/or running via the `docker` CLI.

Perhaps let me re-iterate the motivation for the PR: in my opinion, it is useful to have a way to build the project (the binary firmware file) without having to clutter the system with necessary prerequisities -- namely NodeJS, all npm dependencies, Python, PlatformIO and more. Docker is an ideal way to do so. So I would expect a straightforward way to use Docker in order to build the firmware; my Dockerfile (+compose.yaml) achieves that goal. The currently existing one does not.

Also, the `.devcontainer` directory -- is this a standardized place for Dockerfiles? As far as I can tell, this is something specific to one editor (VS Code).

netmindz:
The devcontainer is a vscode extension, but I should give everything needed to run all of your pio setup without the need for any extra tooling. Please try your compose with the existing Dockerfile to confirm. Maybe VSCode calls something you bootstrap the container?

ondras:
> The devcontainer is a vscode extension, but I should give everything needed to run all of your pio setup without the need for any extra tooling. Please try your compose with the existing Dockerfile to confirm. Maybe VSCode calls something you bootstrap the container?

Trying to build & run the `.devcontainer/Dockerfile`:

```
$ docker build -f .devcontainer/Dockerfile -t wled --progress=plain .
#0 building with "default" instance using docker driver

#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 953B done
#1 DONE 0.0s

#2 [internal] load .dockerignore
#2 transferring context: 2B done
#2 DONE 0.0s

#3 [internal] load metadata for mcr.microsoft.com/devcontainers/python:0-3
#3 DONE 0.1s

#4 [1/1] FROM mcr.microsoft.com/devcontainers/python:0-3@sha256:452e43dd2381de67191c53f7692433baba156ee9138ccfaed5124ff26c63eaed
#4 CACHED

#5 exporting to image
#5 exporting layers done
#5 writing image sha256:49772e83e0c2edab5f45780ea53be7012cd55731ccabf5d6bd57ca7a6dc1dbff done
#5 naming to docker.io/library/wled done
#5 DONE 0.0s

$ docker run -it wled
Python 3.11.4 (main, Jun  7 2023, 18:32:58) [GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>
```

Apparently, this is just a plain image with a Python REPL interpreter. This corresponds to the fact that the Dockerfile is mostly empty. Also, it is somewhat suspicious that the image itself comes from Microsoft -- I guess that this particular file is strongly bound to some particlar VS Code extension (a piece of software that I do not use).

All in all, I do not think that this Dockerfile is sufficient/suitable for independent firmware building. Perhaps someone with a deeper knowledge of that extension can explain how is this supposed to work?

netmindz:
Yeah it looks like vscode must be doing some bootstrapping to install the dependencies automatically. 

We should still try and standardise on a single Dockerfile as it's the same requirement - an environment to do your builds in

I think putting a docker-compose.yml into the root of the project will confuse more users than it would help

The approach of baking the source into the build image it also flawed, adding .gitignore doesn't resolve the issue - the source should be mounted into the builder image

ondras:
> I think putting a docker-compose.yml into the root of the project will confuse more users than it would help

I have no problem with moving that file to other location.

> The approach of baking the source into the build image it also flawed, adding .gitignore doesn't resolve the issue - the source should be mounted into the builder image

Okay then -- what do you propose *should* be in the image? We can only copy the definition of dependencies (e.g. `package.json` and `requirements.txt`) and install those during the build phase. The whole source tree will be then mounted (along with the output build dir) at runtime via bind-mounts specified in the `compose.yaml`.

I am not sure if this is a suitable approach, though: bind-mounting the source makes the image "always up to date", but changing a dependency breaks the build completely (as the container will attempt to build the mounted source with out-of-the-date dependencies).

ondras:
Giving this some more thought: apparently for the purpose of "building a firmware binary in an isolated environment", the difference between a *build phase* and a *runtime phase* is very small. To summarize our options, we can:

1) Build the firmware during image build. This means only specifying stuff in the Dockerfile and this approach has been actually suggested by the very first draft posted in a [Discord discussion](https://discord.com/channels/473448917040758787/473448917543944193/1506185323833593936) by Sören.

2) Do nothing in the image build (just use a suitable baseimage) and perform the whole shebang at runtime, via a script mounted (and executed) inside of the image. This approach basically ignores the existence of a docker image, as we always start on a clean slate.

3) Perform a "setup phase" (build deps) when building and continue with the compilation at runtime.

From the point of view of a "one-shot build", I do not see any differences between these approaches. But if we consider the lifecycle of a container (data created at runtime are stored in a temporary writable FS layer that overlays the image), IMHO it is useful to think of the container as of something easily disposable. And for repeated builds (e. g. for multiple targets), we probably do not want to install the deps every time we build. So from my point of view, the approach **3** makes the most sense (as it allows persistent caching of build dependencies -- something that does not change often when compared to the source code itself).

willmmiles:
> Also, the .devcontainer directory -- is this a standardized place for Dockerfiles? As far as I can tell, this is something specific to one editor (VS Code).

Yes, it's an attempt to create a standard place for containerized software development with a composable featureset of tools.  The tooling is homed here: https://containers.dev/  -- while it was created to smooth the experience for VSCode, using VSCode is not required for using the devcontainer standard.   The [`devcontainer` command line tool](https://containers.dev/implementors/reference/) can be used to assemble a container from the specification.

> Yeah it looks like vscode must be doing some bootstrapping to install the dependencies automatically.

That's also correct - `devcontainer.json` instructs VSCode to install the `platformio.platformio-ide` extension, which in turn bootstraps PlatformIO.   If we wanted to have PlatformIO pre-installed we could switch to a standard image and adopt the PlatformIO feature:

```json
"image": "mcr.microsoft.com/devcontainers/python:3",
"features": {
    "ghcr.io/ar90n/devcontainer-features/platformio:1": {}
}
```
(Note that the PlatformIO VSCode extension will pretty much never use a pre-installed version -- it always wants to install its own captive version.   The option to enable using an external PlatformIO install seems generally broken.)

Same could be done for our npm install I think.

----

To sum up: I think it's definitely true that our `devcontainer.json` could use some attention to produce something that works out-of-the-box for development with other tooling.  I would prefer to stay within the devcontainer framework for managing container specifications, if in no small part because it integrates directly with Github Codespaces, allowing people to make custom builds without having to deploy a local container environment.

ondras:
> To sum up: I think it's definitely true that our devcontainer.json could use some attention to produce something that works out-of-the-box for development with other tooling. I would prefer to stay within the devcontainer framework for managing container specifications, if in no small part because it integrates directly with Github Codespaces, allowing people to make custom builds without having to deploy a local container environment.

I think that we are in full agreement: the ability to build binary firmware in an isolated environment is the goal; individual tools that enable it are not that important.

I am okay with using the `devcontainer` feature as long as it is available in my non-IDE/non-GUI/non-VSCode setup. Is this something I can use *right now* to build the firmware?

ondras:
I tried reading more about the whole devcontainer concept and honestly I am really puzzled about it. The official website does not seem to answer the most basic questions -- why is this abstraction necessary, why they decided not to use regular Dockerfiles, why they decided not to use (and/or extend) `docker compose`. I've been working with Docker-based products for ~10 years and it seems that a container+portforward+bindmounts (orchestrated via compose) was always sufficient for all development needs.

willmmiles:
> I tried reading more about the whole devcontainer concept and honestly I am really puzzled about it. The official website does not seem to answer the most basic questions -- why is this abstraction necessary, why they decided not to use regular Dockerfiles, why they decided not to use (and/or extend) `docker compose`. I've been working with Docker-based products for ~10 years and it seems that a container+portforward+bindmounts (orchestrated via compose) was always sufficient for all development needs.

The key limitation with pure Docker tooling is composability -- you can't trivially mix+match layers to construct a final environment.   The devcontainer tooling brings a registry of "features" -- essentially, publicly maintained script snippets to install a certain tool -- that can be stacked arbitrarily to compose the final image.  It makes it easy to have this project include (nodejs + platformio), that project (nodejs + aws), the next one (aws + conda), and so forth.  And best of all, I don't have to maintain copies of those snippets in every one of my projects; they're implemented as references to public repositories.

Essentially the devcontainer spec is a way of constructing a unique project-specific dockerfile from publicly available fragments that do not have a rigid layer ordering.

I sketched a more-complete devcontainer at https://github.com/willmmiles/WLED/tree/devcontainer-completeness -- you're welcome to give it a try.

ondras:
> I sketched a more-complete devcontainer at https://github.com/willmmiles/WLED/tree/devcontainer-completeness -- you're welcome to give it a try.

Well, I tried -- it *did* something:

```
$ devcontainer up
[3 ms] @devcontainers/cli 0.87.0. Node.js v22.22.2. linux 7.0.0-14-generic x64.
{"outcome":"success","containerId":"04f84fabbc2dae6ccb61da10df89ee3e85fe918d67ab2723eba153b5803ba7b1","remoteUser":"vscode","remoteWorkspaceFolder":"/workspaces/WLED"}
```

What exactly is the next step? I tried `exec`ing into the running container:

```
$ devcontainer exec bash
vscode ➜ /workspaces/WLED (main) $ pio
bash: pio: command not found
```

willmmiles:
> > I sketched a more-complete devcontainer at https://github.com/willmmiles/WLED/tree/devcontainer-completeness -- you're welcome to give it a try.
> 
> Well, I tried -- it _did_ something:
> What exactly is the next step? I tried `exec`ing into the running container:
> 
> ```
> $ devcontainer exec bash
> vscode ➜ /workspaces/WLED (main) $ pio
> bash: pio: command not found
> ```

Hm - that's pretty much exactly what I did?

```bash
wmiles@Alcyone:~/Projects/WLED4$ git checkout devcontainer-completeness
branch 'devcontainer-completeness' set up to track 'origin/devcontainer-completeness'.
Switched to a new branch 'devcontainer-completeness'
wmiles@Alcyone:~/Projects/WLED4$ devcontainer up
[1 ms] @devcontainers/cli 0.87.0. Node.js v22.18.0. linux 6.6.87.2-microsoft-standard-WSL2 x64.
[865 ms] Resolving Feature dependencies for 'ghcr.io/ar90n/devcontainer-features/platformio:1'...
[1513 ms] Resolving Feature dependencies for 'ghcr.io/devcontainers/features/node:2'...
<output snipped>
wmiles@Alcyone:~/Projects/WLED4$ devcontainer exec bash
vscode ➜ /workspaces/WLED4 (devcontainer-completeness) $ pio run -e nodemcuv2
Processing nodemcuv2 (board: nodemcuv2; platform: espressif8266@4.2.1; framework: arduino)
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
<output snipped>
```

The CLI tool doesn't seem clever enough to notice that the `devcontainer.json` file has been updated when you switch branches.  You might need to run `devcontainer up --remove-existing-container` to get it to rebuild the container with the new branch, otherwise it will continue with the container built for the previous branch.

ondras:
Sorry, my bad! I forgot that your fix lives in a branch and tried the whole stuff in `main`. It works now -- the firmware is compiled.

The only small non-fatal problem I encountered when running `pio run -e esp32dev` is this red-colored log fragment:

```
Installing collected packages: zopfli, wheel, pyyaml, tasmota-metrics
ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied: '/usr/local/py-utils/venvs/platformio/lib/python3.14/site-packages/zopfli'
Check the permissions.


*** Error 1
```

Nevertheless, the goal of having a working isolated build environment has been achieved 👍 and I would welcome merging the updated devcontainer definition into the mainline WLED. Now that my PR has been obsoleted, I can at least provide a new PR with updated documentation for those devs (me included) who would benefit from the new build env. Is this something desirable?

willmmiles:
Superseded by #5638.

PR #5637, opened 2026-05-21
Adding ESPNOW connection to sync multiple hardware with sound reactive.

Gallosauro:
Hello, first of all, I would like to thank everyone for the Wled. I have been using it for many years, and it is simply amazing. With the help of the AI I made this change that adds an option to the menu to connect multiple sound reactive devices with ESP now instead of UDP since I was having problems with the connection. 
I built a custom-made hardware with an ESP32 Ue , PCM1808 and a Bluetooth receiver. The file compiles, and it works nicely so far. On the INFO menu, also display the frame loss and the sync status.

### Recommended setup: 
Master on send, Slave on receive, ESPNOW, and in case of different pinout : CheckboxSettingBrightness✅Color✅Effects✅Palette / Effect arguments✅Main segment only✅Segment options / Segment bounds / Segment data❌ untickedApply incoming sync✅

### Known issue: 
Every ESP32 has one radio. That radio can only be on one 2.4 GHz channel at a time. In a normal single-AP Wi-Fi setup, every WLED board associates to the same AP, so every radio ends up on the same channel — ESP-NOW broadcasts reach everyone. In a mesh, each mesh node (satellite) operates on its own channel, and your WLED boards roam to whichever node is closest. So board #1 ends up on channel 1, board #2 on channel 6, board #3 on channel 11 — and they literally can't hear each other's ESP-NOW packets even though they're on the same Wi-Fi network. WLED's state sync still works (it goes through the AP and onto the IP network), but ESP-NOW is peer-to-peer and channel-pinned.
Easiest fix: Lock the mesh to a single 2.4 GHz channel. 

<details>
<summary>Audio Reactive ESP-NOW Patch — Complete Edit Recipe</summary>

This is the full explanation of the code from the AI:

Audio Reactive ESP-NOW Patch — Complete Edit Recipe
Canonical single-doc version. Applies to a clean copy of WLED v16.0.0's usermods/audioreactive/audio_reactive.cpp (2307 lines). Eleven edits total — overlapping ones from the original two-doc workflow are now merged.
Drop espnow_audio_sync.h into the same folder before starting.
How to apply each edit: in VSCode, Ctrl+F → click the small arrow to open Find & Replace → paste the FIND block → paste the REPLACE block → hit Replace once. Each FIND block is unique in the file, so there's no ambiguity.

________________________________________
Edit 0 — include the transport module
What it does: makes the ESP-NOW transport API available to the rest of the file.
FIND: (in the top include block, around line 12 — already done in your current file if you see this line)
#include <esp_timer.h>
REPLACE WITH:
#include <esp_timer.h>

#include "espnow_audio_sync.h"
________________________________________
Edit 1 — class members for the two GUI settings
What it does: adds two uint8_t members that hold the user's choices: which transport (UDP or ESP-NOW) and which channel (auto or forced 1–13).
FIND:
    uint16_t audioSyncPort= 11988;// default port for UDP sound sync
REPLACE WITH:
    uint16_t audioSyncPort= 11988;// default port for UDP sound sync
    uint8_t  audioSyncTransport = 0;  // NEW: 0 = UDP (stock), 1 = ESP-NOW (low-latency)
    uint8_t  espnowChannel      = 0;  // NEW: 0 = auto (follow Wi-Fi), 1-13 = force channel
________________________________________
Edit 2 — config-key string declarations
What it does: declares the names that will appear as JSON keys in cfg.json under the sync section.
FIND:
    static const char UDP_SYNC_HEADER[];
    static const char UDP_SYNC_HEADER_v1[];
REPLACE WITH:
    static const char UDP_SYNC_HEADER[];
    static const char UDP_SYNC_HEADER_v1[];
    static const char _syncTransport[];
    static const char _espnowChannel[];
________________________________________
Edit 3 — config-key string definitions
What it does: assigns the actual JSON-key strings ("transport" and "espnow-channel") to the labels declared above. These are stored in PROGMEM to save RAM.
FIND:
const char AudioReactive::UDP_SYNC_HEADER_v1[] PROGMEM = "00001"; // old sync header version - need to add backwards-compatibility feature
REPLACE WITH:
const char AudioReactive::UDP_SYNC_HEADER_v1[] PROGMEM = "00001"; // old sync header version - need to add backwards-compatibility feature
const char AudioReactive::_syncTransport[]     PROGMEM = "transport";
const char AudioReactive::_espnowChannel[]     PROGMEM = "espnow-channel";
________________________________________
Edit 4 — initialise ESP-NOW when Wi-Fi connects
What it does: every time WLED's connected() fires (boot or Wi-Fi reconnect), this brings up the ESP-NOW transport. The configured channel overrides the auto-detected one if the user has chosen a fixed channel in the GUI. espnowAudioSync::begin() auto-detects whether to run hosted (alongside WLED's own ESP-NOW for WiZmote/state-sync) or standalone (we own the radio).
FIND:
    void connected() override
    {
      if (udpSyncConnected) {   // clean-up: if open, close old UDP sync connection
        udpSyncConnected = false;
        fftUdp.stop();
      }
      
      if (audioSyncPort > 0 && (audioSyncEnabled & 0x03)) {
      #ifdef ARDUINO_ARCH_ESP32
        udpSyncConnected = fftUdp.beginMulticast(IPAddress(239, 0, 0, 1), audioSyncPort);
      #else
        udpSyncConnected = fftUdp.beginMulticast(WiFi.localIP(), IPAddress(239, 0, 0, 1), audioSyncPort);
      #endif
      }
    }
REPLACE WITH:
    void connected() override
    {
      if (udpSyncConnected) {   // clean-up: if open, close old UDP sync connection
        udpSyncConnected = false;
        fftUdp.stop();
      }
      
      if (audioSyncPort > 0 && (audioSyncEnabled & 0x03)) {
      #ifdef ARDUINO_ARCH_ESP32
        udpSyncConnected = fftUdp.beginMulticast(IPAddress(239, 0, 0, 1), audioSyncPort);
      #else
        udpSyncConnected = fftUdp.beginMulticast(WiFi.localIP(), IPAddress(239, 0, 0, 1), audioSyncPort);
      #endif
      }
      // NEW: bring up ESP-NOW transport (auto-detects hosted vs standalone)
      // espnowChannel: 0 = auto (follow Wi-Fi), 1-13 = force channel
      const uint8_t syncChannel = (espnowChannel >= 1 && espnowChannel <= 13)
                                  ? espnowChannel
                                  : WiFi.channel();
      const bool    amSender    = (audioSyncEnabled & 0x01) != 0;
      if (!espnowAudioSync::begin(syncChannel, amSender)) {
        DEBUG_PRINTLN(F("[ENaudio] espnowAudioSync::begin() failed - falling back to UDP."));
      }
    }
________________________________________
Edit 5 — send over ESP-NOW (master)
What it does: in transmitAudioData(), if the user has selected the ESP-NOW transport, hand the packet to ESP-NOW. Otherwise keep doing the stock UDP multicast. Either path uses the same 44-byte V2 packet.
FIND:
      if (fftUdp.beginMulticastPacket() != 0) { // beginMulticastPacket returns 0 in case of error
        fftUdp.write(reinterpret_cast<uint8_t *>(&transmitData), sizeof(transmitData));
        fftUdp.endPacket();
      }
REPLACE WITH:
      if (audioSyncTransport == 1) {
        // NEW: ESP-NOW transport
        espnowAudioSync::send(&transmitData, sizeof(transmitData));
      } else {
        // UDP transport (stock behaviour)
        if (fftUdp.beginMulticastPacket() != 0) { // beginMulticastPacket returns 0 in case of error
          fftUdp.write(reinterpret_cast<uint8_t *>(&transmitData), sizeof(transmitData));
          fftUdp.endPacket();
        }
      }
________________________________________
Edit 6 — receive over ESP-NOW (slave)
What it does: in receiveAudioData(), if the user has selected the ESP-NOW transport, poll the addon for the latest packet and decode it through the same decodeAudioData() the UDP path uses. Otherwise the function falls through to the stock UDP code unchanged.
FIND:
    bool receiveAudioData()   // check & process new data. return TRUE in case that new audio data was received. 
    {
      if (!udpSyncConnected) return false;
      bool haveFreshData = false;
REPLACE WITH:
    bool receiveAudioData()   // check & process new data. return TRUE in case that new audio data was received. 
    {
      if (audioSyncTransport == 1) {
        // NEW: ESP-NOW transport
        audioSyncPacket enFrame;
        uint32_t        enDeadline = 0;
        if (espnowAudioSync::poll(&enFrame, sizeof(enFrame), &enDeadline, /*renderDelayUs=*/15000)) {
          decodeAudioData(sizeof(enFrame), reinterpret_cast<uint8_t*>(&enFrame));
          receivedFormat = 2;
          last_UDPTime   = millis();
          return true;
        }
        return false;
      }
      if (!udpSyncConnected) return false;
      bool haveFreshData = false;
________________________________________
Edit 7 — persist the new settings (save side)
What it does: adds both new settings to the sync JSON object that WLED writes to cfg.json whenever the user saves.
FIND:
      JsonObject sync = top.createNestedObject("sync");
      sync["port"] = audioSyncPort;
      sync["mode"] = audioSyncEnabled;
    }
REPLACE WITH:
      JsonObject sync = top.createNestedObject("sync");
      sync["port"] = audioSyncPort;
      sync["mode"] = audioSyncEnabled;
      sync[FPSTR(_syncTransport)] = audioSyncTransport;
      sync[FPSTR(_espnowChannel)] = espnowChannel;
    }
________________________________________
Edit 8 — persist the new settings (load side)
What it does: reads the two new settings back from cfg.json on boot. The (uint8_t)0 is the default used if the field is missing (first boot after flashing the patch, before the user has saved anything).
FIND:
      configComplete &= getJsonValue(top["sync"]["port"], audioSyncPort);
      configComplete &= getJsonValue(top["sync"]["mode"], audioSyncEnabled);
REPLACE WITH:
      configComplete &= getJsonValue(top["sync"]["port"], audioSyncPort);
      configComplete &= getJsonValue(top["sync"]["mode"], audioSyncEnabled);
      configComplete &= getJsonValue(top["sync"][FPSTR(_syncTransport)], audioSyncTransport, (uint8_t)0);
      configComplete &= getJsonValue(top["sync"][FPSTR(_espnowChannel)], espnowChannel,      (uint8_t)0);
________________________________________
Edit 9 — add the two dropdowns to the Sound Settings page
What it does: emits the JavaScript that builds the Audio Sync Transport and espnow-channel dropdowns in WLED's web UI under Config → Sound Settings. Both appear right after the existing Sync (Off/Send/Receive) dropdown.
FIND:
      uiScript.print(F("addOption(dd,'Receive',2);"));
REPLACE WITH:
      uiScript.print(F("addOption(dd,'Receive',2);"));
      // NEW: Audio Sync Transport selector (UDP vs ESP-NOW)
      uiScript.print(F("dd=addDropdown(ux,'sync:transport');"));
      uiScript.print(F("addOption(dd,'UDP (default)',0);"));
      uiScript.print(F("addOption(dd,'ESP-NOW (low-latency)',1);"));
      // NEW: ESP-NOW channel selector (Auto + 1..13)
      uiScript.print(F("dd=addDropdown(ux,'sync:espnow-channel');"));
      uiScript.print(F("addOption(dd,'Auto (follow Wi-Fi)',0);"));
      uiScript.print(F("addOption(dd,'1',1);"));
      uiScript.print(F("addOption(dd,'2',2);"));
      uiScript.print(F("addOption(dd,'3',3);"));
      uiScript.print(F("addOption(dd,'4',4);"));
      uiScript.print(F("addOption(dd,'5',5);"));
      uiScript.print(F("addOption(dd,'6',6);"));
      uiScript.print(F("addOption(dd,'7',7);"));
      uiScript.print(F("addOption(dd,'8',8);"));
      uiScript.print(F("addOption(dd,'9',9);"));
      uiScript.print(F("addOption(dd,'10',10);"));
      uiScript.print(F("addOption(dd,'11',11);"));
      uiScript.print(F("addOption(dd,'12',12);"));
      uiScript.print(F("addOption(dd,'13',13);"));
________________________________________
Edit 10 — show ESP-NOW stats on the Info pane
What it does: when the user has switched to the ESP-NOW transport, the WLED home page's Info pane shows live counters: how many frames have been received, the rolling frame-loss percentage, how many packets the master has queued for transmit, any TX errors, and which mode the addon is running in (hosted / standalone). Healthy slaves show RX climbing fast and Frame Loss under 1 %.
FIND:
        if (audioSyncEnabled && udpSyncConnected && (millis() - last_UDPTime < 2500)) {
            if (receivedFormat == 1) infoArr.add(F(" v1"));
            if (receivedFormat == 2) infoArr.add(F(" v2"));
        }
REPLACE WITH:
        if (audioSyncEnabled && udpSyncConnected && (millis() - last_UDPTime < 2500)) {
            if (receivedFormat == 1) infoArr.add(F(" v1"));
            if (receivedFormat == 2) infoArr.add(F(" v2"));
        }

        // NEW: ESP-NOW audio sync stats (only shown when transport is set to ESP-NOW)
        if (audioSyncTransport == 1) {
          auto s = espnowAudioSync::stats();

          infoArr = user.createNestedArray(F("ESP-NOW Audio RX"));
          infoArr.add(s.rx);
          infoArr.add(F(" frames"));

          infoArr = user.createNestedArray(F("ESP-NOW Frame Loss"));
          uint32_t denom = s.rx + s.gaps;
          if (denom > 0) {
            char buf[16];
            snprintf(buf, sizeof(buf), "%.2f %%", 100.0f * (float)s.gaps / (float)denom);
            infoArr.add(buf);
          } else {
            infoArr.add(F("--"));
          }

          infoArr = user.createNestedArray(F("ESP-NOW Audio TX"));
          infoArr.add(s.tx);
          infoArr.add(F(" packets"));

          if (s.tx_err > 0) {
            infoArr = user.createNestedArray(F("ESP-NOW TX Errors"));
            infoArr.add(s.tx_err);
          }

          infoArr = user.createNestedArray(F("ESP-NOW Mode"));
          infoArr.add(espnowAudioSync::isHosted() ? F("hosted (sharing WLED stack)") : F("standalone"));
        }
________________________________________
Edit 11 — coexistence hook with WLED's own ESP-NOW
What it does: WLED's core ESP-NOW dispatcher (in wled00/udp.cpp) calls UsermodManager::onEspNowMessage(...) on every incoming ESP-NOW packet before it does its own WiZmote / state-sync handling. By overriding this method on the AudioReactive class, we get first look at every packet. We claim audio-sync packets (their magic bytes match 0xA53E) and return true → WLED skips its own handling for that packet. For anything else (WiZmote button presses, ESP-NOW state sync, etc.) we return false and WLED proceeds normally. This is what lets the WiZmote remote and ESP-NOW state-sync features keep working alongside the audio sync. The whole thing is wrapped in #ifndef WLED_DISABLE_ESPNOW so the build still compiles if you ever disable ESP-NOW in build flags.
FIND: (this multi-line block is unique — copy it exactly, including the two blank lines between onStateChange and the addToConfig comment)
        createAudioPalettes();
      }
    }

    /*
     * addToConfig() can be used to add custom persistent settings to the cfg.json file in the "um" (usermod) object.
REPLACE WITH:
        createAudioPalettes();
      }
    }

#ifndef WLED_DISABLE_ESPNOW
    // ESP-NOW audio-sync coexistence hook. WLED's core ESP-NOW dispatcher
    // (espNowReceiveCB in wled00/udp.cpp) calls this on every received
    // packet BEFORE its own WiZmote/state-sync handling. We grab audio-sync
    // packets and return true; WLED then skips the rest of its dispatch.
    bool onEspNowMessage(uint8_t* sender, uint8_t* data, uint8_t len) override {
      return espnowAudioSync::handleIncomingPacket(data, len);
    }
#endif
</details>

## Summary by CodeRabbit

* **New Features**
  * ESP-NOW added as an alternative transport for audioreactive sync
  * Transport selection exposed in UI/config; info page shows ESP-NOW RX/TX stats, loss, and TX errors

* **Improvements**
  * Optimized FFT buffer allocation and cleanup for float and integer DSP paths
  * Improved UDP receive buffering, realtime-mode suspension logic, and ESP-NOW/Wi‑Fi coexistence for more reliable sync

[Review Change Stack](https://app.coderabbit.ai/change-stack/wled/WLED/pull/5637?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


DedeHai:
Thanks for contributing

I did not yet look at the code but just from my experience with ESPNow I know it is a bit of a "diva" and unless you know what you are doing and testing all possible scenarios it can run into a lot of issues. So my question is: how much of the code is AI and did you review and try tounderstand every line of code? 
This question is just to know where this is at in terms of "ripeness"

Gallosauro:
Hello, it is mostly AI-generated. I tested on the hardware, and it is working fine, though.

DedeHai:
I just now read the first few lines of changes and I have to say that this wont fly at all in its current state, all I see is AI slop.

DedeHai:
Are you experienced enough in C++ to clean this up?

usermods/audioreactive/audio_reactive.cpp
```diff
@@ -296,17 +298,17 @@ void FFTcode(void * parameter)
   ArduinoFFT<float> FFT = ArduinoFFT<float>(valFFT, vImag, samplesFFT, SAMPLE_RATE, true);
 #elif !defined(UM_AUDIOREACTIVE_USE_INTEGER_FFT)
   // allocate and initialize FFT buffers on first call
-  // note: free() is never used on these pointers. If it ever is implemented, this implementation can cause memory leaks (need to free raw pointers)
   if (valFFT == nullptr) {
-    float* raw_buffer = (float*)heap_caps_malloc((2 * samplesFFT * sizeof(float)) + 16, MALLOC_CAP_8BIT);
-    if ((raw_buffer == nullptr)) return; // something went wrong
-    valFFT = (float*)(((uintptr_t)raw_buffer + 15) & ~15);  // SIMD requires aligned memory to 16-byte boundary. note in IDF5 there is MALLOC_CAP_SIMD available
+    valFFT = (float*)heap_caps_aligned_calloc(16, 2 * samplesFFT, sizeof(float), MALLOC_CAP_8BIT); // SIMD requires aligned memory to 16-byte boundary. note in IDF5 there is MALLOC_CAP_SIMD available
```
softhack007, line 302: Why did you change this part? It has nothing to to with sound sync,

softhack007 (changes requested):
Hi @Gallosauro thanks for your contribution.

Before we can continue with code reviews, please align your code with the latest source code from `main`. 

Currently your PR contains lots of deletions that revert some of our latest improvements. We cannot integrate your PR like this, and I don't want to spam with "please revert this change" comments 😉

softhack007:
> Hello, it is mostly AI-generated. I tested on the hardware, and it is working fine, though.

oh-oh, this sounds like lots of work for reviewers 😉

softhack007:
> Recommended setup:
> Master on send, Slave on receive, ESPNOW, and in case of different pinout : CheckboxSettingBrightness✅Color✅Effects✅Palette / Effect arguments✅Main segment only✅Segment options / Segment bounds / Segment data❌ untickedApply incoming sync✅

Aehm ... this has nothing to do with Sound sync. Are you sure? And what does the "different pinout" remark want to say?

softhack007:
> This is the full explanation of the code from the AI:
> 
> Audio Reactive ESP-NOW Patch — Complete Edit Recipe Canonical single-doc version. Applies to a clean copy of WLED v16.0.0's usermods/audioreactive/audio_reactive.cpp (2307 lines). Eleven edits total — overlapping ones from the original two-doc workflow are now merged.

Please clean up the PR description. Parts like above text are just confusing, at least for me.

usermods/audioreactive/audio_reactive.cpp
```diff
@@ -11,6 +11,8 @@
 #if defined(ARDUINO_ARCH_ESP32) && (defined(WLED_DEBUG) || defined(SR_DEBUG))
 #include <esp_timer.h>
 #endif
+#define ESPNOW_AUDIO_SYNC_SKIP_PACKET_DECL  // audio_reactive.cpp already declares audioSyncPacket
```
softhack007, line 14: Please don't re-define the audiosync packet structure.

softhack007: edit: looks like this define is not referenced, please remove it.

usermods/audioreactive/audio_reactive.cpp
```diff
@@ -1903,6 +1933,37 @@ class AudioReactive : public Usermod {
...
+          infoArr.add(F(" packets"));
+
+          if (s.tx_err > 0) {
+            infoArr = user.createNestedArray(F("ESP-NOW TX Errors"));
+            infoArr.add(s.tx_err);
+          }
+
+          infoArr = user.createNestedArray(F("ESP-NOW Mode"));
+          infoArr.add(espnowAudioSync::isHosted() ? F("hosted (sharing WLED stack)") : F("standalone"));
+        }
```
softhack007, line 1966: please reduce the statistics lines added here - add only 2-3 lines that are absolutely needed. If you want more infos, put them behind `#ifdef WLED_DEBUG` ... `#endif`

usermods/audioreactive/audio_reactive.cpp
```diff
@@ -2164,6 +2240,10 @@ class AudioReactive : public Usermod {
       uiScript.print(F("addOption(dd,'Send',1);"));
 #endif
       uiScript.print(F("addOption(dd,'Receive',2);"));
+      // NEW: Audio Sync Transport selector (UDP vs ESP-NOW)
+      uiScript.print(F("dd=addDropdown(ux,'sync:transport');"));
+      uiScript.print(F("addOption(dd,'UDP (default)',0);"));
+      uiScript.print(F("addOption(dd,'ESP-NOW (low-latency)',1);"));
```
softhack007, line 2247: why do you think ESP-NOW is low-latency?

usermods/audioreactive/espnow_audio_sync.h
```diff
@@ -0,0 +1,414 @@
...
+
+#pragma once
+
+#include <Arduino.h>
+#include <WiFi.h>
+#include <esp_now.h>
+#include <esp_wifi.h>
+#include <esp_timer.h>
+#include <esp_idf_version.h>
+#include <string.h>
```
softhack007, line 44: too many includes, please clean up

usermods/audioreactive/espnow_audio_sync.h
```diff
@@ -0,0 +1,414 @@
...
+ * @brief  On-the-wire ESP-NOW packet layout.
+ *
+ * Packed to ensure byte-identical layout across senders/receivers.
+ * Carries a 12-byte preamble followed by the raw 44-byte
+ * audioSyncPacket payload from WLED's audioreactive usermod.
+ */
+struct __attribute__((packed)) ENPacket {
+    uint16_t magic;                ///< Equals EN_MAGIC (0xA53E).
+    uint8_t  version;              ///< Equals EN_VERSION (1).
+    uint8_t  flags;                ///< Bit 0: this is the duplicate copy.
```
softhack007, line 83: ?

usermods/audioreactive/espnow_audio_sync.h
```diff
@@ -0,0 +1,414 @@
...
+    uint16_t seq;                  ///< Monotonically increasing per sender.
+    uint16_t pad;                  ///< Alignment pad (set to 0).
```
softhack007, line 85: padding in a packed struct? why?

usermods/audioreactive/espnow_audio_sync.h
```diff
@@ -0,0 +1,414 @@
...
+
+/** @brief True once begin() has succeeded. Subsequent begin() calls no-op. */
+static bool      _initialised = false;
+
+/**
+ * @brief True when WLED's QuickEspNow already owned ESP-NOW at begin()
+ *        time (HOSTED mode); false when this module owns the radio
+ *        (STANDALONE mode).
+ */
+static bool      _hosted = false;
```
softhack007, line 105: can you explain why these two modes (hosted/self-hosted) are needed?

usermods/audioreactive/espnow_audio_sync.h
```diff
@@ -0,0 +1,414 @@
...
+ *         frame is available since the last call.
+ */
+inline bool poll(void * out, size_t outLen,
+                 uint32_t * deadline_us = nullptr,
+                 uint32_t renderDelayUs = 15000) {
+    if (!_rxReady) return false;
+    if (outLen < EN_PAYLOAD) return false;
+
+    uint32_t arrLocalUs;
+    portENTER_CRITICAL(&_rxMux);
```
softhack007, line 303: (major problem) 
portENTER_CRITICAL() stalls everything, including LEDs drivers. If you need a mutex, please use a regular FreeRTOS semaphore instead.

usermods/audioreactive/espnow_audio_sync.h
```diff
@@ -0,0 +1,414 @@
...
+    if (ep->magic != EN_MAGIC || ep->version != EN_VERSION) return false;
+    if (_isSender) return true;  // ours but we ignore -- we sent it
+
+    if (_hasLastRxSeq) {
+        int16_t diff = (int16_t)((int16_t)ep->seq - (int16_t)_lastRxSeq);
+        if (diff <= 0) { _stat_dupes++; return true; }
+        if (diff > 1)  { _stat_gaps += (diff - 1); }
+    }
+
+    portENTER_CRITICAL(&_rxMux);
```
softhack007, line 352: same as above - no  portENTER_CRITICAL() please, use a semaphore if needed.

usermods/audioreactive/espnow_audio_sync.h
```diff
@@ -0,0 +1,414 @@
...
+ * straight to handleIncomingPacket().
+ *
+ * @note In HOSTED mode this function is never registered; WLED's
+ *       QuickEspNow owns the receive callback and the usermod's
+ *       onEspNowMessage() override is what invokes
+ *       handleIncomingPacket() instead.
+ */
+static void _onRecv(ESPNOW_RECV_CB_SIG) {
+    ESPNOW_RECV_CB_ARGS_UNUSED
+    handleIncomingPacket(data, len);
```
softhack007, line 411: What if an incoming packet is not for audio sync, but for esp-now remote or for WLED sync? Will these be droped, or not?

softhack007:
Generic comment: please make sure that esp-now sound sync is only added when WLED_DISABLE_ESPNOW is not defined. This also implies the "self-hosted" part of your PR can be removed. 

If users build with `-D WLED_DISABLE_ESPNOW` the usermod should respect that.

usermods/audioreactive/audio_reactive.cpp
softhack007, line 1: @coderabbitAI compilation of this file fails, maybe there is a missing or superfluous `}` ?

softhack007, line 1: @coderabbitAI please scan `espnow_audio_sync.h‎` for unused variables, unused struct members, unneeded getter/setter functions (=only called from the main object), and dead or unreachable code.

usermods/audioreactive/espnow_audio_sync.h
softhack007, line 1: @coderabbitai please scan `espnow_audio_sync.h`‎ for unused variables, unused struct members, unneeded getter/setter functions (=only called from the main object), and dead or unreachable code.

usermods/audioreactive/espnow_audio_sync.h
```diff
@@ -0,0 +1,414 @@
...
+ * @file    espnow_audio_sync.h
+ * @brief   Low-latency ESP-NOW transport for WLED audioreactive sync.
+ *
+ * Drop-in replacement for the UDP multicast path used by WLED's
+ * audioreactive usermod. Keeps the exact same V2 audioSyncPacket
+ * wire-payload (44 bytes), so receivers process audio identically to
+ * before.
+ *
+ * Features:
+ *   - ESP-NOW broadcast (~2-5 ms one-way vs ~8-30 ms UDP/Wi-Fi)
```
softhack007, line 11: Just for explaination: the observed delays are only a minor problem. sound sync sends one packet each 22millis, and the listed network delays (wifi included) are both the same magnitude, so not going to add noticeable lagging. 

"Wifi sleep" - if enabled - is actually the relevant "realtime killer". If just one client on your wifi allows sleep mode, the AP/router is allowed to send udp data in batches, with up to 500ms between batches.

usermods/audioreactive/espnow_audio_sync.h
```diff
@@ -0,0 +1,414 @@
...
+    if (payloadLen != EN_PAYLOAD) return false;
+
+    ENPacket ep;
+    memset(&ep, 0, sizeof(ep));
+    ep.magic   = EN_MAGIC;
+    ep.version = EN_VERSION;
+    ep.flags   = 0;
+    ep.seq     = ++_txSeq;
+    ep.pad     = 0;
+    ep.send_us = (uint32_t)(esp_timer_get_time() & 0xFFFFFFFFULL);
```
softhack007, line 261: This is a microsecond timer, it will roll-over about once per hour. Are you sure?

usermods/audioreactive/espnow_audio_sync.h
```diff
@@ -0,0 +1,414 @@
...
+        if (diff > 1)  { _stat_gaps += (diff - 1); }
+    }
+
+    portENTER_CRITICAL(&_rxMux);
+    _lastRxSeq        = ep->seq;
+    _hasLastRxSeq     = true;
+    _stat_rx++;
+    memcpy(_rxPayload, ep->payload, EN_PAYLOAD);
+    _rxSendUs         = ep->send_us;
+    _rxArrivedLocalUs = (uint32_t)(esp_timer_get_time() & 0xFFFFFFFFULL);
```
softhack007, line 358: What do you want to measure here? Getting two esp32 into time sync on microseconds level is an impossible mission....

DedeHai: not impossible but futile ;)

DedeHai:
@softhack007 I did not really look into this code: is the general approach viable? From glancing over the code it looks like its a full re-implementation of ESPnow code and duplicates what is already implemented in the main code - I do know there is a hook for receiving messages, I do not know if there is one for sending them but that would be the proper way instead of making AR fully incompatible with any other ESPNow functionality.

softhack007:
@DedeHai yes I agree, I also have doubts about the general approach.
* Adding a new "transport" settings option is a good idea - there were also suggestions in the past for using "UDP boadcast" instead of "UDP multicast" as the transport protocol, so esp-now would fit into this picture.
* 8266 not supported - receiving should work on 8266 with the existing callback hooks. Only esp32 can be a sender.
* Receiving should _only_ rely on the `onEspNowMessage()` callback hook. The hooked function can check the message format and return false in case of "not meant for AR", otherwise directly unpack with ` decodeAudioData()` - no need to pass this on to the main loop.
* I'm not sure if "sending" actually needs a dedicated implementation - it would be better if AR calls the same functions that ESP-NOW based WLED Sync already provides.
* De-duplication might be a specific topic for esp-now, however the timestamp / sequence ID approach in this PR might cause more troubles than it brings benefit. WLED-MM AR already has an optional sequence-check implemented, so adding another one on esp-now level might become a technical debt for the future.
* UDP sound sync must be inactive while running on ESP-NOW - this is not addressed in the PR.
* A lot more testing is needed - including corner cases and different WiFi / ethernet / AP-less setups, multiple senders, etc. 

My Conclusion
* Some good new ideas
* Avoid a full re-implementation of ESPnow within audioreactive. Better re-use existing esp-now sending code and receiving callbacks. 
* 8266 support is needed for esp-now receiving.
* This PR only works for one user setup, and I already see lots of troubles and forgotten cases. The "critical section" that may cause LEDs flickering on -C3 is just one example. 
* I'm not sure if @Gallosauro has good enough experience with C++ (and in working with github) to continue adjusting the AI code in this PR. We cannot rely on an AI agent alone - the topic is much more complicated than "works for me" and "let's vibe-code it".
* (minor) we cannot merge this PR as it stands - @Gallosauro started with "add files via upload", and this unfortunately reverts changes we already made after 16.0.0 was released.

softhack007:
For completeness - I've found two PR that also suggest a way for adding esp-now in AR. They all need rework but can be examples for learning what's needed.

* https://github.com/MoonModules/WLED-MM/pull/148
* https://github.com/wled/WLED/pull/4066

Gallosauro:
Thanks, everyone, for the comments. I will work from that. No, I do not have enough experience in C++, but I will try to see from these comments if I can do anything. In my case solved the issue of connecting multiple cards, which failed with the UDP connection. This code implementation also works with the espnow normal function, and if a command is sent will receive it on the slave card. In the menu, by selecting ESPnow instead of UDP, it automatically deselects it to answer the question. Regarding the C3, no testing has been done yet. 
Again, thanks everyone for the amazing job with Wled.

netmindz:
Based on my POC, there is no advantage to using ESPnow over just changing the sender to use UDP broadcast rather than multicast.
In fact it's better as you don't get clashing WiFi and ESPnow on the same frequency 

ESPnow also had no visibly better latency or consistency

So even if a festival setup, a mobile WiFi hotspot would be better than trying to use ESPnow really

PR #5646, opened 2026-05-23
fix(FX): restore palette wrap-around in the Palette effect (regression since 0.15.x)

kaibae19:
### Summary

The **Palette** effect (`mode_palette`, FX 65) no longer blends across the palette's end→start seam under the default palette-blending setting. Because the effect scrolls the palette along the strip, the unblended seam shows up as a hard color step that visibly travels pixel-by-pixel along the strip. Blending *between* adjacent palette entries still works; only the wrap-around is broken. This is a regression introduced in 0.16 — the effect was smooth in 0.15.x.

It is most obvious with palettes whose first and last colors differ strongly (e.g. *Retro Clown*).

### Root cause

`mode_palette` colors each pixel via `Segment::color_wheel()`. In commit 9bfd34e6 ("Rework color_from_palette() - respect paletteBlend") the third parameter of `color_from_palette()` was repurposed from `wrap` to `moving`, and `color_wheel()` was changed to pass `false`:

```cpp
// 0.15.x
  if (palette) return color_from_palette(pos, false, true, 0);  // wrap = true
  ...
// 0.16
  if (palette) return color_from_palette(pos, false, false, 0); // moving = false -> "only wrap if always wrap is set"
  ...
```

With the default setting (`paletteBlend == 0`, *"Linear (wrap if moving)"*), `color_from_palette()` resolves `moving == false` to `LINEARBLEND_NOWRAP`, so the end→start interpolation is dropped. In 0.15.x the same call passed `wrap == true`, giving full `LINEARBLEND` with wrap.

Note that the `color_wheel()` *non-palette* branch generates a continuous CHSV hue wheel that already wraps seamlessly (hue 255→0), so the palette branch not wrapping is also internally inconsistent for a function named "wheel".

### Fix

Add an optional `moving` parameter to `color_wheel()` defaulting to `false` — so every existing caller is byte-for-byte unchanged and the maintainer's "wheel defaults to no-wrap" intent is preserved — and have `mode_palette`, which is by definition a *moving* palette, request wrap.

```cpp
uint32_t Segment::color_wheel(uint8_t pos, bool moving) const {
  if (palette) return color_from_palette(pos, false, moving, 0);
  ...
}
// mode_palette:
const uint32_t color = SEGMENT.color_wheel((uint8_t)colorIndex, true);
```

### Steps to reproduce

1. Effect: **Palette**; Palette: **Retro Clown** (or any palette with very different first/last colors); enable **Animate Shift** (or set Speed > 0).
2. LED Preferences → Palette wrapping left at the default *"Linear (wrap if moving)"*.
3. A hard color seam travels along the strip at the wrap point. Setting Palette wrapping to *"Linear (always wrap)"* hides it (workaround); with this patch the default is smooth again.

### Risk / blast radius

Only `mode_palette` changes behavior. The new parameter defaults to `false`, so all other `color_wheel()` callers (≈30, mostly discrete random-color picks) are unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved color-wheel palette handling for smoother, continuous animations and better palette interpolation, resulting in more fluid color transitions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wled/WLED/pull/5646?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1117,7 +1117,7 @@ void Segment::blur(uint8_t blur_amount, bool smear) const {
  * Rotates the color in HSV space, where pos is H. (0=0deg, 256=360deg)
  */
 uint32_t Segment::color_wheel(uint8_t pos) const {
-  if (palette) return color_from_palette(pos, false, false, 0); // only wrap if "always wrap" is set
+  if (palette) return color_from_palette(pos, false, true, 0); // color_wheel is a continuous (moving) wheel, so wrap end->start (restores pre-0.16 behaviour)
   uint8_t w = W(getCurrentColor(0));
   CRGBW rgb;
   rgb = CHSV32(static_cast<uint16_t>(pos << 8), 255, 255);


wled00/FX.h
```diff
@@ -714,7 +714,7 @@ class Segment {
     inline void addPixelColor(int n, CRGB c, bool preserveCR = true) const         { addPixelColor(n, RGBW32(c.r,c.g,c.b,0), preserveCR); }
     inline void fadePixelColor(uint16_t n, uint8_t fade) const                     { setPixelColor(n, color_fade(getPixelColor(n), fade, true)); }
     [[gnu::hot]] uint32_t color_from_palette(uint16_t, bool mapping, bool moving, uint8_t mcol, uint8_t pbri = 255) const;
-    [[gnu::hot]] uint32_t color_wheel(uint8_t pos) const;
+    [[gnu::hot]] uint32_t color_wheel(uint8_t pos, bool moving = false) const;
```
softhack007: Why did you add a new parameter?

wled00/FX_fcn.cpp
```diff
@@ -1116,8 +1116,8 @@ void Segment::blur(uint8_t blur_amount, bool smear) const {
  * The colours are a transition r -> g -> b -> back to r
  * Rotates the color in HSV space, where pos is H. (0=0deg, 256=360deg)
  */
-uint32_t Segment::color_wheel(uint8_t pos) const {
-  if (palette) return color_from_palette(pos, false, false, 0); // only wrap if "always wrap" is set
+uint32_t Segment::color_wheel(uint8_t pos, bool moving) const {
```
softhack007: Ah ok, understood. Forget about my previous question 👍

blazoncek:
@kaibae19 about [a year ago](https://github.com/wled/WLED/commit/9bfd34e6b201c111b107396eae7b66c2f1ee2eba) I started replacing/modifying (confusing) `color_wheel()` and `color_from_palette()` but was interrupted midway. Unfortunately not all changes made it through and upstream remained in an even more confusing state.

What I have done (privately) since is replace `color_wheel()` with an inline version that calls `color_from_palette()` and embedded rainbow color selection if `mcol`==255 in `color_from_palette()`. This allows all palette selection to happen inside one function and inline version of `color_wheel()` just calls `color_from_palette()` with `moving` parameter set to `true`. BTW using "Default" palette is confusing and not very obvious to users (but that has a very long history).

I would recommend you to revert adding a parameter (even one with default value) to `color_wheel()` and perhaps move rainbow palette color selection into color_from_palette() or (even simpler solution) to just replace `false` with `true` in `color_from_palette()` call as it was obvious mistake when modifying code. Historically `color_wheel()` called `color_from_palette()` with `moving` parameter set to `true`.

kaibae19:
Thanks — that matches what I found: `color_wheel()` called `color_from_palette()` with `moving=true` until `ee9ac947` flipped it to `false`, so this restores the original behaviour rather than adding new surface.

I've dropped the added parameter and just changed the `false` back to `true` in `color_wheel()`. The PR is now a single-line change.

Tested on a classic ESP32 (141 LEDs): the Palette effect with a moving palette (Retro Clown, blend mode "wrap if moving") wraps the end→start seam smoothly again, and the other `color_wheel()` callers just get the same wrap restored.

Happy to defer to your larger `color_wheel()`/`color_from_palette()` refactor (inlining + `mcol==255` rainbow path) whenever it lands — this is just the minimal regression fix in the meantime.

softhack007:
> kaibae19 force-pushed the fix/palette-wrap-color-wheel branch from 9a8c639 to  b5d0a90 26 minutes ago

@kaibae19 I appreciate your contribution, and happy that you found an even simpler patch that restores backward compatibility 👍

 For the next time you contribute, please remember that `force-push`ing an open PR is bad practice on github. You can see that the review bot already lost track, and it will sit there with `review skipped` until I kick it back into live. 

https://github.com/wled/WLED?tab=contributing-ov-file#updating-your-code

softhack007:
@coderabbitai full review. Please follow back the commit history to confirm that the suggested fix restores backward compatibility, avoiding negative side-effects for other call sites.

softhack007 (approved):
Looks like the right solution 👍 

Cherry-pick to 16_x after merge.

DedeHai:
Looks good

Issue #5628, opened 2026-05-18
HUB75 not supporting FM6124 and FM6126A chips


ondras:
### What happened?

My 64x32 HUB75 panel uses the FM6126A driver chip. It currently does not work with WLED. By studying the source code of the WLED/MM fork, I discovered that it is because the chip is not correctly initialized (my panel works with WLED/MM). Apparently, this can be already done in WLED: https://github.com/wled/WLED/blob/main/wled00/bus_manager.cpp#L815

### To Reproduce Bug

Connect the Waveshare 64x32 panel.

### Expected Behavior

LEDs should turn on.

### Install Method

Binary from WLED.me

### What version of WLED?

WLED 16.0

### Which microcontroller/board are you seeing the problem on?

ESP32-C3

### Relevant log/trace output

```shell

```

### Anything else?

May I submit a PR that enables this explicit driver assignment, available as another HUB75 panel type? I can add both new options ("new half scan", "new quarter scan") if that is desirable.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

softhack007:
This is actually a known limitation - the WLED variant of the HUB75 bus driver currently does not support chipsets that don't work with the standard driver.

The proper solution will be to add a dropdown in LEDs & Harware settings, so users can select the proper chipset driver.

WLED-MM currently uses a simple rule-of-thump: most "outdoor" and 4-scan panels are FM6124 compatible (like FM6126A), while "indoor" 2-scan usually work better with the default driver.

https://github.com/MoonModules/WLED-MM/blob/1f2919e7abf91d045eebb4b8c2499f79f3b94d94/wled00/bus_manager.cpp#L707


Edit: actually we need _two_ driver fields - one for the "shiftregister" chip, and one for the "line" driver chip.

https://github.com/mrcodetastic/ESP32-HUB75-MatrixPanel-DMA/blob/3171232e660e4c2dd43293e8fb2c48e160318032/src/ESP32-HUB75-MatrixPanel-I2S-DMA.h#L232-L253

Edit2: afaik the reason we did not add these yet is that we barely managed to clean to up mess in settings_leds.htm. But we still need a good concept for adding drop-downs for specific led output types.

netmindz:
That's annoying, it looks like my comment didn't save 😔 

[...]

PR #5647, opened 2026-05-23
HUB75: use FM6124 driver for 4-scan panels, remove 64x64 limitation if board has PSRAM (partial solution for #5628)

softhack007:
* some 4-scan "outdoor" panels need the FM6124 driver. Tests on my hardware showed that "shiftreg" panels seem to work with this driver, so the risk of regressions is low
* partial solution for #5628
* Remove 1-panel limit if the board has PSRAM.
* Remove an unnecessary mxconfig.double_buff == false check (we never use the driver's internal double buffer)

## Bug Fixes
* HUB75 LED matrix driver now validates hardware compatibility more thoroughly and automatically adjusts configuration settings when needed for improved reliability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wled/WLED/pull/5647?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/wled00/bus_manager.cpp
+++ b/wled00/bus_manager.cpp
@@ -997,10 +997,8 @@ BusHub75Matrix::BusHub75Matrix(const BusConfig &bc) : Bus(bc.type, bc.start, bc.
     }
     setBitArray(_ledsDirty, _len, false);             // reset dirty bits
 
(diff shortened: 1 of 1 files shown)


softhack007:
tested, works for me.

Issue #5247, opened 2025-12-31
0.16.0 - Ethernet not initialising properly


brettbear:
### What happened?

After compiling WLED 0.16.0-alpha 2506160 a number of different ways and getting the same behaviour, I discovered that when WLED boots with ethernet configured it doesn't connect using the defined static IP Address. This presents in two different ways.

If DHCP is enabled on the ethernet network, 0.16.0 seems to pickup a DHCP address and not use the static one definied in WLED. If the ethernet is unplugged and reconnected, then it uses the static IP Address definied in WLED.

If DHCP is disabled on the network, WLED seems to pick up the static IP Address, however, it doesn't "connect" to it. If the ethernet is unplugged and reconnected, then it uses AND connects via the static IP Address definied in WLED.

Platform: Quinled dig-octa

[WLED 0.16.0-alpha 2506160 debug log - DHCP enabled on router.txt](https://github.com/user-attachments/files/24390600/WLED.0.16.0-alpha.2506160.debug.log.-.DHCP.enabled.on.router.txt)

[WLED 0.16.0-alpha 2506160 debug log - no DHCP on router.txt](https://github.com/user-attachments/files/24390602/WLED.0.16.0-alpha.2506160.debug.log.-.no.DHCP.on.router.txt)

[Online compiler output.txt](https://github.com/user-attachments/files/24390630/Online.compiler.output.txt)

[cfg.json](https://github.com/user-attachments/files/24390647/cfg.json)

### To Reproduce Bug

Compile and install 0.16.0 Nightly (esp32dev) with ethernet.
Via WIFI, configure the ethernet to use a static IP Address.
Connect Ethernet, reboot the module and try to connect to the defined static IP address. Connection will fail.
Unplug ethernet for 5 seconds, reconnect. You'll be able to connect to WLED over ethernet.

### Expected Behavior

If ethernet is connected when WLED boots, it should connect immediately via ethernet using the defined static IP Address.

### Install Method

Self-Compiled

### What version of WLED?

WLED 0.16.0-alpha 2506160

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell
*** WLED 0.16.0-alpha 2506160 debug log - DHCP enabled on router ***
Using 'COM7' as serial port.
Unexpected error: could not open port 'COM7': PermissionError(13, 'Access is denied.', None, 5)
[13:53:01]ets Jul 29 2019 12:21:46
[13:53:01]
[...]

PR #4838, opened 2025-08-14
🚀 ESP-IDF V5

netmindz:
Update the code to handle at least being able to compile against V5

<br/>

Edit (softhack007): espressif has "silently" changed a lot of APIs, and - in many cases - replacements are not behaving the same way as before. We are lucky that a few "legacy APIs" are still available, like the "V4" I2S driver. So even when "it compiles", there may be a lot more adaptations needed before "it really works".

<br/> 

## Helpful information and migration guidance
* https://docs.espressif.com/projects/esp-idf/en/stable/esp32/migration-guides/release-5.x/5.0/peripherals.html
* https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference
* https://docs.espressif.com/projects/arduino-esp32/en/latest/migration_guides/2.x_to_3.0.html
* https://randomnerdtutorials.com/esp32-migrating-version-2-to-3-arduino/#adc-removed-apis

## coding guides

We should make sure that the "V5" branch will still compile for "V4" as a backup solution.
When adding code that only works in the new framework, it should be conditionally compiled, like in the examples below

* "V5" code that is not specific to new boards
````C++
      #if defined(ESP_IDF_VERSION) && (ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 0))
      ledcAttach(pwmPin, 25000, 8);  // New API: ledcAttach(pin, freq, resolution)
      #else
      ledcSetup(pwmChannel, 25000, 8);
      // attach the channel to the GPIO to be controlled
      ledcAttachPin(pwmPin, pwmChannel);
      #endif
````

* "V5" code only for esp32-P4 or esp32-C6:
````c++
 #if defined(CONFIG_IDF_TARGET_ESP32C6)
    // https://docs.espressif.com/projects/esp-idf/en/latest/esp32c6/api-reference/peripherals/gpio.html
    // strapping pins: 4, 5, 8, 9
    if (gpio > 11 && gpio < 14) return false;     // 12-13 USB-JTAG
    if (gpio > 23 && gpio < 31) return false;     // 24-30 SPI FLASH
#elif defined(CONFIG_IDF_TARGET_ESP32P4)
    // strapping pins: 34,35,36,37,38
    if (             gpio <   2) return false;     // NC
    if (gpio > 13 && gpio <  20) return false;     // ESP-Hosted WiFi pins
    if (gpio > 23 && gpio <  26) return false;     // USB Pins
[...]

PR #5650, opened 2026-05-24
Add independent Ethernet/WiFi IP config and primary network interface selection

kelvtech-co-uk:
## Caution

The majority of this code was generated using claude.ai, I don't have the skills or experience in langages to do this myself unfortunately.  However I'm keen to learn so feedback welcomed on how I can improve all of this.

## Motivation

I needed a low latency network connection between my Gledopto GL-C-618WL and my media player running Hyperion.  However I was unable to provide a wired LAN connection from my network switch due to both devices location.  I hacked a solution by modifying my media players CoreELEC network configuration and directly cabling the devices.  In this configuration WLED was only useable whilst my media player was turned on as it became the router between my main LAN and the stub network I hacked into CoreELEC.  I wanted a better solution and modifying WLED's firmware was easier than modifying my house to run a cable.

## Core Changes

- Add separate static IP/gateway/subnet configuration for Ethernet interface independent of WiFi static IP configuration
- Add primary network interface selection to control which interface lwIP uses for outbound connections (MQTT, NTP, internet access)
- Implement setPrimaryNetworkInterface() using netif_set_default() to set lwIP default netif based on user selection
- Keep both interfaces active simultaneously when Ethernet connects
- Show both interface IPs in UI when both are active
- Update WiFi & Network settings UI with Ethernet IP config section and Primary Network Interface selector
- Ethernet IP config section hidden when Ethernet Type is None

## Resolves

Issue #5247 - ethernet static IP not applying independently of WiFi config

## Related

Issue #4703 - ESP-NOW + Ethernet GPIO17 PLL instability (referenced in known limitations)

## File Changes

- `wled00/wled.h` — new global variables
- `wled00/cfg.cpp` — serialisation/deserialisation of new ethernet IP config
- `wled00/set.cpp` — form POST handler for new UI fields
- `wled00/xml.cpp` — s.js population of new UI fields, dual IP display
- `wled00/network.cpp` — ETH.config() with new variables, setPrimaryNetworkInterface(), WiFi kept active on ETH connect
- `wled00/wled.cpp` — WiFi config() update, setPrimaryNetworkInterface() call in handleConnection()
- `wled00/fcn_declare.h` — declaration of setPrimaryNetworkInterface()
- `wled00/data/settings_wifi.htm` — Ethernet IP Configuration section, Primary Network Interface selector

## Known limitations

- mDNS resolves to primary interface IP only; per-interface mDNS hostname registration is a candidate for follow-on work
- Dual interface operation has been validated on boards using ETH_CLOCK_GPIO0_IN (external clock source). Boards using ETH_CLOCK_GPIO17_OUT may experience instability when WiFi and Ethernet operate simultaneously due to ESP32 PLL instability described in #4703. A future enhancement could detect the board clock mode and conditionally disable WiFi on affected boards, preserving existing single-interface behaviour for those users.

## Testing

Tested on:
- **WT32-ETH01** (ETH_CLOCK_GPIO0_IN) — dev board, dual interface operation validated, both IPs accessible simultaneously from clients on both subnets
- **Gledopto GL-C-618WL** (ETH_CLOCK_GPIO0_IN) — production hardware, OTA updated, ambilight use case validated

Test scenarios covered:
- Static IP persistence across reboot
- Both interfaces accessible from clients on respective subnets
- Primary network interface selection verified via serial debug output confirming correct netif selection (setPNI: log lines) and practical network reachability tests from clients on both of my subnets
- WiFi-only builds unaffected (all new code inside `WLED_USE_ETHERNET` guard)
- Existing WiFi static IP configuration unaffected

## Summary by CodeRabbit

* **New Features**
  * Independent Ethernet static-IP (address, gateway, subnet) with UI controls (shown when Ethernet enabled); saved/loaded from configuration.
  * “Primary Network Interface” selector (WiFi vs Ethernet) with live application and “Active:” display plus conditional fallback warning.

* **Bug Fixes / Reliability**
  * Improved primary-interface selection and re-evaluation on IP/link events; Ethernet connect no longer disables WiFi and selection is reapplied as needed.

* **Behavior**
  * Network status can show both active interfaces; DNS/mDNS label renamed to “Active IP(s)”.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wled/WLED/pull/5650?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


softhack007:
> @kelvtech-co-uk force-pushed the feature/eth-dual-interface branch from 9513309 to 015e2e2

@kelvtech-co-uk one important rule - even when its non-technical - is to not `force-push` to an open PR on github. Reviewer comments sometimes vanish, and the review bot often loses track and stays in "nothing to review" forever.

https://github.com/wled/WLED?tab=contributing-ov-file#during-review


I'll restart the bot for this PR to fix it - you may receive some duplicate comments from coderabbit.

softhack007:
@coderabbitai full review

please create a prioritized list of suggested improvements and bugfixes after your review (major items only).

softhack007:
> **P2 — Clean up generated-code comments before merge**
> 
> * There are many `AI:` / “generated by an AI” comments in production files. They add noise and may not match project style.
> * Suggested fix: replace with concise technical comments only where they explain non-obvious behavior.

@coderabbitai clarification: `// AI: below section was generated by an AI` comments are the project-specific convention to highlight AI-generated code blocks, these can be kept inside webUI files as they will be stripped out during the `uglify/minify` process during firmware builds. I agree that documentation inside these blocks needs to be understandable and relevant, both for maintainability and for a meaningful PR review.

kelvtech-co-uk:
Fixes pushed for 3 of the 4 issues flagged by CodeRabbit.

- `xml.cpp`: replaced `sprintf` with `snprintf_P` for IP formatting
- `cfg.cpp`: added `ARDUINO_ARCH_ESP32` guard around ESP32-only ethernet symbols  
- `set.cpp`: ethernet IP config now applied immediately on save without reboot, bringing parity with WiFi live-update behaviour

@coderabbitai review

kelvtech-co-uk:
@softhack007 I've draft code changes to address the last priority item that CodeRabbit has flagged but I'm stuck hunting a bug from testing on my gledopto so it might be a couple more days.  Do I need to be concerned about a release window you work to i.e. ideally get this PR final by X date?

kelvtech-co-uk:
I don't know whether I'm Claude's puppet or he's mine but sharing verbaitm our current stubling block in the hope someone might have relevant experience to share...

During testing of setPrimaryNetworkInterface() we observed different netif_is_up() behaviour between our two test boards:

WT32-ETH01 — netif_is_up() correctly returns true when ethernet is connected and has a valid IP. The function works as expected and setPrimaryNetworkInterface() correctly switches the default netif to ethernet.

Gledopto GL-C-618WL — netif_is_up() consistently returned false (up=0) even when ethernet was physically connected and passing traffic (Hyperion/E1.31 ambilight working correctly over the same interface). We were unable to determine conclusively whether this is a hardware-specific behaviour of the Gledopto's LAN8720 driver, a timing issue in our test setup, or an artefact of our unusual network configuration described below.  

Our Gledopto test environment involves a non-standard stub network topology — an isolated 192.168.10.0/24 subnet between the Gledopto and a CoreELEC media player, with ConnMan blacklisted from managing the ethernet interface and IP assignment handled manually via ifconfig as part of a systemd ExecStartPre script. This causes the ethernet interface to flap briefly during CoreELEC's boot sequence, and our UDP debug logging (which requires WiFi to be up) made it impossible to observe a clean stable ethernet state.

We believe this fringe case is clouding the netif_is_up() question. The Gledopto's inaccessibility at 192.168.10.2 from other subnets is a routing/asymmetric path issue specific to this network topology rather than a bug in our code — the same address was accessible when running the original Gledopto firmware with no WiFi and a direct gateway configured.

We are deferring the Gledopto-specific investigation and leaving netif_is_up() unchanged. If anyone has experience with netif_is_up() behaviour on other ESP32 ethernet boards we'd welcome input.

@GLEDOPTO

kelvtech-co-uk:
The following changes have been made since the last review:

- `network.cpp`: Replaced fragile lwIP netif name prefix matching ('en'/'st') with esp-netif API (`esp_netif_next` / `esp_netif_get_desc`) to identify interfaces by their official description strings ("eth" for Ethernet, "sta" for WiFi STA) as defined in `esp_netif_defaults.h`. This addresses the P1 finding from the previous review. `LOCK_TCPIP_CORE()` is retained around `netif_set_default()`. Fallback logic preserved — if preferred interface is unavailable any ready interface is used instead.

Expect at least 1 further commit for this PR adding a UI indicator showing active vs configured primary network interface.

@coderabbitai review

kelvtech-co-uk:
Okay I think this is enough for this PR.

@coderabbitai review

kelvtech-co-uk:
> I don't know whether I'm Claude's puppet or he's mine but sharing verbaitm our current stubling block in the hope someone might have relevant experience to share...
> 
> During testing of setPrimaryNetworkInterface() we observed different netif_is_up() behaviour between our two test boards:
> 
> WT32-ETH01 — netif_is_up() correctly returns true when ethernet is connected and has a valid IP. The function works as expected and setPrimaryNetworkInterface() correctly switches the default netif to ethernet.
> 
> Gledopto GL-C-618WL — netif_is_up() consistently returned false (up=0) even when ethernet was physically connected and passing traffic (Hyperion/E1.31 ambilight working correctly over the same interface). We were unable to determine conclusively whether this is a hardware-specific behaviour of the Gledopto's LAN8720 driver, a timing issue in our test setup, or an artefact of our unusual network configuration described below.
> 
> Our Gledopto test environment involves a non-standard stub network topology — an isolated 192.168.10.0/24 subnet between the Gledopto and a CoreELEC media player, with ConnMan blacklisted from managing the ethernet interface and IP assignment handled manually via ifconfig as part of a systemd ExecStartPre script. This causes the ethernet interface to flap briefly during CoreELEC's boot sequence, and our UDP debug logging (which requires WiFi to be up) made it impossible to observe a clean stable ethernet state.
> 
> We believe this fringe case is clouding the netif_is_up() question. The Gledopto's inaccessibility at 192.168.10.2 from other subnets is a routing/asymmetric path issue specific to this network topology rather than a bug in our code — the same address was accessible when running the original Gledopto firmware with no WiFi and a direct gateway configured.
> 
> We are deferring the Gledopto-specific investigation and leaving netif_is_up() unchanged. If anyone has experience with netif_is_up() behaviour on other ESP32 ethernet boards we'd welcome input.
> 
> @GLEDOPTO

**Solved...**

In single interface ethernet builds (WiFi disabled), lwIP's default netif is always ethernet — replies to inbound connections naturally return via the same interface. In dual interface configurations (WiFi + Ethernet), WLED may use WiFi as the default netif depending on PNI configuration and interface availability. This means that if your ethernet interface is on an isolated subnet and you want it reachable from other subnets via static routing, you must configure a gateway IP on the ethernet interface pointing to your router — leaving it as 0.0.0.0 will cause reply packets to route via WiFi instead, making the ethernet IP unreachable from remote subnets even with correct static routes in place.

In my case a Gledopto GL-C-618WL on an isolated 192.168.10.0/24 stub network became reachable at 192.168.10.2 from my main LAN only after configuring 192.168.10.1 (the stub network router) as the ethernet gateway. With 0.0.0.0 the device was unreachable despite correct static routing on the Unifi gateway.

willmmiles:
> In dual interface configurations (WiFi + Ethernet), WLED may use WiFi as the default netif depending on PNI configuration and interface availability. This means that if your ethernet interface is on an isolated subnet and you want it reachable from other subnets via static routing, you must configure a gateway IP on the ethernet interface pointing to your router — leaving it as 0.0.0.0 will cause reply packets to route via WiFi instead, making the ethernet IP unreachable from remote subnets even with correct static routes in place.

This is the expected behaviour of an IP stack: it uses the only route it knows about to route reply packets back to the source address, which leaves only the wifi if you configure no routes on the ethernet.  IP packets do not contain the addresses of the routers that passed them through -- the device has no idea the packet was routed by "192.168.10.1" on its way to your WLED device, all it knows is that it's supposed to reply to "192.168.x.100" or whatever your client IP is.  So it is not possible to "reply by the interface you received a packet on" unless you tell your device that there's a router on that subnet that can reach the source network.

In the general case, you can of course also add explicit static routes, informing the stack that "IP ranges in the 192.168.0.0/16 block can be reached by sending packets to 192.168.10.1" without saying "any address can be reached by sending packets to 192.168.10.1" (ie. setting a default route aka "gateway" on that interface) -- though honestly I think that supplying interfaces to such advanced networking features is beyond the scope of WLED.

kelvtech-co-uk:
> This is the expected behaviour of an IP stack: it uses the only route it knows about to route reply packets back to the source address, which leaves only the wifi if you configure no routes on the ethernet.  IP packets do not contain the addresses of the routers that passed them through -- the device has no idea the packet was routed by "192.168.10.1" on its way to your WLED device, all it knows is that it's supposed to reply to "192.168.x.100" or whatever your client IP is.  So it is not possible to "reply by the interface you received a packet on" unless you tell your device that there's a router on that subnet that can reach the source network.
> 

Thanks, I was caught out by this not being the case with v0.15.4. I have a static route on my router for the 192.168.10.0/24 range sending it to my coreelec player, this is then onward routing that subnet to the directly connected gledopto. With only ethernet enabled and statically configured with no gateway I could browse to WLEDs ethernet stub address from remote subnets. This wasnt the case with my code changes and I stupidly lost time to an obvious check i should have done upfront.

> In the general case, you can of course also add explicit static routes, informing the stack that "IP ranges in the 192.168.0.0/16 block can be reached by sending packets to 192.168.10.1" without saying "any address can be reached by sending packets to 192.168.10.1" (ie. setting a default route aka "gateway" on that interface) -- though honestly I think that supplying interfaces to such advanced networking features is beyond the scope of WLED.

100% agee, static routing or supporting topologies like mine are niche and shouldn't be a focus for the project. Caveat that Id love this PR to be merged 🤞🏻

kelvtech-co-uk:
Merged latest main into this branch to resolve a conflict in wled00/wled.cpp.

The conflict was between this PR's change to the static IP condition in initConnection() (dropping the staticGW != 0 requirement, since gateway 0.0.0.0 is explicitly supported for stub-interface configurations) and the recent uint32_t cast fix on main (8a67495) for the ambiguous-cast compiler warning.

Resolved by keeping this PR's single-condition check with the explicit cast applied:
`if (uint32_t(multiWiFi[selectedWiFi].staticIP) != 0U) {`
So the warning fix from main is retained. Verified with a clean local build of esp32_eth.

willmmiles:
@kelvtech-co-uk  Don't worry too much about keeping up with main for our sakes at the moment.  I'd like to hold this PR until we land V5 (#4838), and there's some internal network API cleanup that I'd like to do as part of the platform/HAL work that's planned; this task will go in cleaner once that's done.  I've tagged this PR to keep it around until we're ready for it.  Thank you!

kelvtech-co-uk:
@willmmiles no problem thanks for letting me know.  On a related note I've been holding off the mDNS updates necessary as a result of this PR pending feedback on it.  Should I go ahead and start adding these or would you prefer I hold for V5 also?

kelvtech-co-uk:
@willmmiles I had some spare time and spare tokens so I decided to explore the V5 [#4838](https://github.com/wled/WLED/pull/4838) works...

Rather than rebase this PR, I've prototyped it against the V5 branch on a side branch to see how it lands on the new stack: [feature/eth-dual-interface-v5](https://github.com/kelvtech-co-uk/WLED/tree/feature/eth-dual-interface-v5). Short version: it ports very cleanly, and hardware testing surfaced a few findings that may be useful for the V5/HAL work.

**The port:** on arduino-esp32 3.x, NetworkInterface::setDefault()/isDefault() (via ETH/WiFi.STA) replace this PR's manual esp-netif iteration and lwIP locking entirely — the V5 selection path is ~12 lines vs ~55, and the deprecated esp_netif_next() disappears. The V4 path is retained under ESP_IDF_VERSION guards; both esp32_eth (IDF 5.3.4) and esp32dev_V4 compile. Hardware-tested on a WT32-ETH01: live PNI switching in both directions without reboot, full failover matrix (eth unplug → WiFi fallback → replug → restore, 6+ cycles), outbound routing verified via DNS + NTP in both states.

**Findings from hardware testing:**

1. Comparing IPAddress against (uint32_t)0 crashes on the 3.x core: under C++20 operator rewriting, != resolves through operator==(const uint8_t*) with the constant zero binding as a null pointer → memcmp(NULL, addr, 4) LoadProhibited (hit in serveSettings). Fixed with explicit uint32_t(ip) != 0U casts — possibly worth grepping for as more code migrates to V5.
2. NetworkInterface::hasIP() alone is insufficient as a readiness check for statically-configured interfaces — the static IP persists on the netif when the link drops, so a dead interface still reports hasIP(). Needs && linkUp().
3. A pre-existing V5 fragility this feature makes reachable: the WIFI_MODE_NULL teardown in initConnection() while the IDF mDNS task is live causes a use-after-free in _udp_join_group / esp_netif_is_netif_up (two distinct crash signatures captured — LoadProhibited @ offset 0xef and InstrFetchProhibited through a garbage function pointer — reproducible within 2–3 ethernet unplug cycles). Calling MDNS.end() before the teardown closes the window; verified clean over 6+ cycles with stable heap. Stock ethernet builds rarely execute this path since WiFi is disabled while eth is up, but on V5 it's reachable whenever WiFi reconnect churn coincides with mDNS running — happy to split this into a small standalone PR against V5 if useful, and it may be relevant to my mDNS question above.

No urgency from my side — the config/UI/JSON layers are unaffected by any of this, so I'm happy to rearrange the couple of functions that touch the interface layer onto whatever API shape emerges from the platform/HAL cleanup.

softhack007:
> 1. Comparing IPAddress against (uint32_t)0 crashes on the 3.x core: under C++20 operator rewriting, != resolves through operator==(const uint8_t*) with the constant zero binding as a null pointer → memcmp(NULL, addr, 4) LoadProhibited (hit in serveSettings). Fixed with explicit uint32_t(ip) != 0U casts — possibly worth grepping f

This aligns with a recent Bugfix in `main`: 8a67495d07e8a67cdd0ce2f48205b4a8205b4c97

kelvtech-co-uk:
> > 1. Comparing IPAddress against (uint32_t)0 crashes on the 3.x core: under C++20 operator rewriting, != resolves through operator==(const uint8_t*) with the constant zero binding as a null pointer → memcmp(NULL, addr, 4) LoadProhibited (hit in serveSettings). Fixed with explicit uint32_t(ip) != 0U casts — possibly worth grepping f
> 
> This aligns with a recent Bugfix in `main`: [8a67495](https://github.com/wled/WLED/commit/8a67495d07e8a67cdd0ce2f48205b4a8205b4c97)

Yes agreed, but I'm just checking as maybe my claude techno babble hid the real message.  In V4 the problem was the compiler "ambiguous cast" warning but in V5 it didn't warn, it compiled clean without the casting and then crashed WLED at runtime when I tried to open the wifi settings webpage.  Apols if this isn't useful, just sharing given the V5 works is still in-progress.

**Claude techno babble -->** One nuance worth capturing for the V5 migration though: they fail oppositely. On V4 this announces itself at build time (the "ambiguous cast" warning) and the resolved behaviour is a benign integer compare. On the 3.x core under C++20, the same pattern compiles silently — operator rewriting brings operator==(const uint8_t*) into candidacy for !=, the constant zero binds as a null pointer, and it fails only at runtime as memcmp(NULL, addr, 4) when the path first executes. So a warning-clean V4 build isn't evidence of V5 safety here, which is why I'd suggest IPAddress-vs-bare-integer comparisons as an explicit V5-migration grep item rather than relying on compiler diagnostics to surface them.

PR #5654, opened 2026-05-29
Add per-bus white-LED color temperature for accurate auto-white

NerdyGriffin:
## What this does

Adds an opt-in, per-bus "white LED color temperature" setting that makes the
auto-white calculation account for the actual color of a strip's W LED.

## Why

WLED's "Accurate" auto-white mode subtracts the white channel value equally
from R, G and B. That implicitly assumes the physical W LED emits neutral
white — RGB(255,255,255), i.e. the sRGB white point near D65 / ~6500 K. Real "RGBW"
strips often use 3000 K warm-white or 4000-5000 K "natural-white" or 5000 K+ cool-white LEDs, so the equal
subtraction shifts the resulting color visibly. This lets the calculation use
the W LED's true per-channel contribution instead.

## How it works

- New per-bus checkbox + Kelvin input below "Auto-calculate W channel from
  RGB" (shown for Accurate mode). Disabled by default.
- When enabled, the configured Kelvin is converted once (via the existing
  `colorKtoRGB()`) to the W LED's RGB equivalent and cached on the bus.
- In `autoWhiteCalc`, instead of `w = min(r,g,b)` and an equal subtract, it
  picks the largest W whose per-channel RGB contribution won't overflow any
  channel, then (ACCURATE mode) subtracts that correct per-channel amount.
  Floor division composes back through the subtract so it can't underflow;
  per-channel zero guards handle channels that are 0 (blue is 0 at/below
  1900 K).
- Opt-in via the checkbox: when off (default, `wk = 0`) `autoWhiteCalc` takes
  a fast path that is identical to current behavior, so existing configs
  render unchanged and there's no added per-pixel cost.
- Persisted per-bus as `wk` in the LED config; range 1000–10000 K.

## Testing

- Built `esp32dev`; flashed via OTA to a QuinLED Dig-Uno (ESP32 + RGBW).
- Verified: UI show/hide across all AW modes; value persists across a full
  power-cycle; Accurate-mode saturation with smooth color transitions and no
  regressions; low end down to the 1000 K floor with no underflow artifacts.

## Notes / limitations

- `colorKtoRGB()` is a blackbody-curve approximation; real phosphor LEDs
  differ from a true blackbody, so the temperature is a user-tunable value
  rather than a fixed per-LED-type table. (In my own testing a 3000 K-rated
  strip looked when set to 2700 K or 2500 K instead.)
- Happy to take maintainer guidance on the per-channel-cap approach in the
  hot path.

## AI assistance

This change was developed with AI assistance (Claude). AI-generated sections
are marked with `// AI:` comments per the contribution guidelines. All code
and this description were reviewed and proof-read by @NerdyGriffin, who takes
responsibility for the contribution.

## Summary by CodeRabbit

* **New Features**
  * Added per-bus white-channel color temperature control from 1000–10000 K, with 0 preserving legacy neutral behavior.
  * Added controls to enable and configure W-channel Kelvin correction; enabling defaults to 6500 K.

* **Improvements**
  * Kelvin settings are validated, saved, restored, and applied per bus.
  * Improved auto-white calculations with faster legacy handling and more accurate, safer W extraction for configured color temperatures.

## Changes since review

Addressing review feedback from @DedeHai, @softhack007 and CodeRabbit
(2026-09-09 to 09-11). Behaviour for buses with the feature off is unchanged
throughout.

- **Hot path is branch-free and division-free** (5d4b69fa). Reciprocals are
  now of `(_wX + 1)` so they can never be zero, which removes the three
  per-channel guard ternaries; the `/255` became `>> 8` to match the file's
  existing brightness scaling. Verified exhaustively on the host over all
  2^24 colours at every 50 K from 1000 to 10000 K: no underflow, and output
  within a few units of the previous arithmetic except below ~2350 K where a
  near-zero blue coefficient now allows slightly more W for zero-blue inputs.
- **Reciprocals shrunk to `uint16_t`** and members ordered by size to avoid
  padding (5d4b69fa, 2f56fde0).
- **Bus eligibility enforced once in `setWhiteKelvin()`** instead of per pixel
  (33817dc3). CCT, non-RGB and non-W buses store 0.
- **`WKE` flag removed** (90096412). `WK = 0` now means off; the checkbox is
  UI-only and derived from `WK` at load. Config-template import normalises
  `wk` the same way the backend does.
- **Correction applies in Accurate mode only** (011b035a). Brighter, and Dual
  with W = 0, never subtract W from RGB, so there was nothing to correct;
  they are now byte-for-byte upstream behaviour. Settings control is shown
  only when the per-bus or global auto-white mode is Accurate.
- **`wk` validated before narrowing at both ingress points** (f010e005).
  1000..10000 K accepted, anything else stored as 0.
- **Q15 constants named** (fba65094) and **source attribution** added to the
  AI-marked blocks (fbe93c16).

All changes flashed and verified on a QuinLED Dig-Uno (ESP32, SK6812 RGBW)
including the settings save round trip.


NerdyGriffin:
I plan to submit a PR to `WLED-Docs` if/when this gets reviewed. Just holding off in case there are any requested changes.

wled00/bus_manager.cpp
```diff
@@ -98,6 +98,22 @@ void Bus::calculateCCT(uint32_t c, uint8_t &ww, uint8_t &cw) {
...
+// where autoWhiteCalc subtracted the same value from R, G, B.
+void Bus::setWhiteKelvin(uint16_t k) {
+  _whiteKelvin = k;
+  if (k == 0) {
+    _wR = _wG = _wB = 255; // legacy: treat W as neutral
+  } else {
+    byte rgb[4];
+    colorKtoRGB(k, rgb);
+    _wR = rgb[0]; _wG = rgb[1]; _wB = rgb[2];
+  }
```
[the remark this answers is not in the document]
NerdyGriffin: I don't think the function _needs_ to be contrained to only accept 1000-10000, since those limits are just "quality of life" ranges for the sake of the UI. I guess it's a question of should it be constrained to color temperatures that are realistic/probably in the real world, or leave it open to all theoretical values. 
Technically there is no reason to limit the low end to 1000, it just seemed like a sensible lower bound.

wled00/data/settings_leds.htm
```diff
@@ -205,6 +205,22 @@
...
+		// (legacy fast path). Seed the field to 6500 K when re-enabling from
+		// a blank or sub-min value so the UI default matches the sRGB white
+		// point.
+		function wkChk(n)
+		{
+			const wke = d.Sf["WKE"+n], wk = d.Sf["WK"+n];
+			if (!wke || !wk) return;
+			if (wke.checked && parseInt(wk.value)<1000) wk.value = 6500;
+			UI();
+		}
```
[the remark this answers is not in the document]
NerdyGriffin: When tested on the WebUI, it appears to reseed the value as intended, so I wonder if there is a mismatch of the value in the UI vs the value in memory, or if there is a misunderstanding of the behavior of `parseInt("")` or `wk.value` in this review comment. I will re-test this to see if I can produce this theoretical bug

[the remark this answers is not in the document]
NerdyGriffin: This was indeed a valid bug. Fixed in 5f61870

DedeHai:
Looks good in general, the hot path needs speed optimization.
Adding config values needs careful consideration: adding per bus white balance/ color correction is on my todo list. Need to think about if that collides with the approach used, the config values especially.

NerdyGriffin:
> Looks good in general, the hot path needs speed optimization. Adding config values needs careful consideration: adding per bus white balance/ color correction is on my todo list. Need to think about if that collides with the approach used, the config values especially.

Thanks for the update. I see how this would overlap with general-purpose color correction, though depending on the implementation they might not achieve the same thing. I implemented this specifically for the use-case of an RGBW strip with `Auto-calculate white channel from RGB` set to `Accurate`, where the resulting colors are very wrong for anything orange/yellow if the hardware's white channel is not 6500K. Whether or not a general color correction can replace my implementation might depend on where the color correction goes in the pipeline, before or after `autoWhiteCalc()`.

In the meantime, I will continue daily-driving my build with this feature on my DigUno and dig2go until something like this or a compatible color correction is implemented in a future release.

NerdyGriffin:
Pushed the hot-path optimization. The per-pixel divisions by `_wR/_wG/_wB` in `autoWhiteCalc`'s enabled path are now precomputed Q15 reciprocals (multiply + shift) cached in `setWhiteKelvin`. The reciprocals are floor-biased, so the white cap can only come out ≤ the exact value — the underflow guarantee is preserved — and I verified the output is equivalent to the previous code across the full input space (≤1/255 delta, no unsafe over-cap).

I also checked the generated Xtensa assembly to be sure: the per-pixel path now has **zero divide instructions**. The `/255` in the subtraction step was already strength-reduced by the compiler to a `muluh` + shift, so I left it alone and only replaced the runtime-variable `/_wX` divisions (which the compiler can't reduce). Re-tested on ESP32 (QuinLED Dig-Uno + dig2go).

Separately — relevant to your point about where general color correction should sit in the pipeline: gamma is applied *before* `autoWhiteCalc` (`FX_fcn.cpp:1742` → `:1743`), so with gamma enabled a per-bus correction subtracts perceptual-space coefficients from already-gamma-corrected channels. I prototyped compensating for that and it didn't cleanly help in practice, so it's not actionable here — just flagging the ordering in case it's useful for the color-correction design.

— Claude (Anthropic's Claude Code; this work was reviewed and hardware-tested by @NerdyGriffin)

wled00/bus_manager.h
```diff
@@ -221,6 +228,13 @@ class Bus {
     uint8_t  _autoWhiteMode; // global Auto White Calculation override
     uint16_t _start;
     uint16_t _len;
+    uint16_t _whiteKelvin; // physical W channel color temperature in Kelvin (0 = neutral/legacy behavior)
+    uint8_t  _wR;          // cached W LED RGB equivalent (255,255,255 when _whiteKelvin==0)
+    uint8_t  _wG;
+    uint8_t  _wB;
+    uint32_t _rwR;         // Q15 reciprocal of _wR (floor((255<<15)/_wR), 0 if _wR==0) for autoWhiteCalc hot path
```
DedeHai, line 238: uint16_t should be accurate enough I think, no need to waste RAM
also I am not sure how the compiler packs this, it's best practice to try and align values to 4byte boundaries to avoid padding.

wled00/set.cpp
```diff
@@ -230,6 +231,9 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
         break;  // no parameter
       }
       awmode = request->arg(aw).toInt();
+      uint16_t whiteK = request->hasArg(wk) ? (uint16_t)request->arg(wk).toInt() : 0;
+      // Reject out-of-range or sub-1000K Kelvin values; 0 means "neutral/legacy"
+      if (whiteK != 0 && (whiteK < 1000 || whiteK > 10000)) whiteK = 0;
```
DedeHai, line 236: the first condition is redundant

wled00/xml.cpp
```diff
@@ -371,6 +371,8 @@ void getSettingsJS(byte subPage, Print& settingsScript)
       char sl[4] = "SL"; sl[2] = offset+s; sl[3] = 0; //skip 1st LED
       char rf[4] = "RF"; rf[2] = offset+s; rf[3] = 0; //off refresh
       char aw[4] = "AW"; aw[2] = offset+s; aw[3] = 0; //auto white mode
+      char wke[5] = "WKE"; wke[3] = offset+s; wke[4] = 0; //W channel color temperature enabled (UI checkbox)
```
DedeHai, line 374: I don't like the use of an enabled flag, could use 0 (or out of range) means disabled (the code in set.cpp already treats it that way)

wled00/bus_manager.cpp
```diff
@@ -112,9 +138,43 @@ uint32_t Bus::autoWhiteCalc(uint32_t c, uint8_t &ww, uint8_t &cw) const {
       //ignore auto-white calculation if w>0 and mode DUAL (DUAL behaves as BRIGHTER if w==0)
     } else if (aWM == RGBW_MODE_MAX) {
       w = r > g ? (r > b ? r : b) : (g > b ? g : b); // brightest RGB channel
+    } else if (_whiteKelvin == 0 || _hasCCT || !_hasRgb) {
```
DedeHai, line 141: this should be done in the UI, avoid conditionals in the hot path if possible

wled00/bus_manager.cpp
```diff
@@ -112,9 +138,43 @@ uint32_t Bus::autoWhiteCalc(uint32_t c, uint8_t &ww, uint8_t &cw) const {
...
+      // floor((r*255)/_wR) * _wR <= r*255 — so the subtraction is safe.
+      // Hot path: the (channel*255)/_wX divisions are replaced by the Q15
+      // reciprocals precomputed in setWhiteKelvin (multiply + shift, no
+      // per-pixel divide). _rwX is floor-biased so wMax never over-estimates,
+      // keeping the cap underflow-safe. _rwX==0 means _wX==0 (channel doesn't
+      // constrain w — _wB is 0 at/below 1900 K, _wG only at extreme lows),
+      // matching the previous per-channel zero guards. The /255 in the
+      // subtraction stays: 255 is a compile-time constant the compiler already
+      // strength-reduces, so it isn't an actual division.
+      unsigned wMaxR = _rwR ? (r * _rwR) >> 15 : 255U;
```
DedeHai, line 166: remove conditionals, _rwR is never zero, _rwG is also never zero and the multiplication + shift is usually faster than a branch as it lets the compiler optimize more.

wled00/bus_manager.cpp
```diff
@@ -112,9 +138,43 @@ uint32_t Bus::autoWhiteCalc(uint32_t c, uint8_t &ww, uint8_t &cw) const {
...
+      unsigned wMaxG = _rwG ? (g * _rwG) >> 15 : 255U;
+      unsigned wMaxB = _rwB ? (b * _rwB) >> 15 : 255U;
+      unsigned wCap = wMaxR < wMaxG ? (wMaxR < wMaxB ? wMaxR : wMaxB) : (wMaxG < wMaxB ? wMaxG : wMaxB);
+      if (wCap > 255U) wCap = 255U;
+      w = wCap;
```
DedeHai, line 171: this changes the MODE_MIN behaviour. intentional? if so why only min and not MODE_MAX?

wled00/bus_manager.cpp
```diff
@@ -112,9 +138,43 @@ uint32_t Bus::autoWhiteCalc(uint32_t c, uint8_t &ww, uint8_t &cw) const {
...
+      if (aWM == RGBW_MODE_AUTO_ACCURATE) {
+        r -= (w * _wR) / 255; // subtract W LED's R contribution
```
DedeHai, line 173: still uses division, why not right shift?

DedeHai (changes requested):

NerdyGriffin:
Worked through the review findings — two fixed, two declined with reasoning:

**Fixed (d9881b83):**
- **Uninitialized reciprocal cache** — `_rwR/_rwG/_rwB` are now initialized in the `Bus` constructor init list alongside `_wR/_wG/_wB`. (Latent only, since they're read solely when `_whiteKelvin != 0`, which is set together with them in `setWhiteKelvin()` — but worth making defensive.)
- **Global auto-white override** — the per-bus Kelvin control now shows when *either* the per-bus AW mode *or* the global override (`Bus::_gAWM`) is Brighter/Accurate/Dual, with `onchange="UI()"` on the global select so it updates live. The per-bus AW selector stays visible under a global override, so the Kelvin sub-control now stays consistent with it instead of vanishing. (Hardware-verified the show/hide on an RGBW bus across the override modes.)

**Declined (with reasoning):**
- **Clamp `wk` at config-load ingress (`cfg.cpp`)** — `colorKtoRGB()` clamps its output to 0–255 for any `uint16_t`, and the per-channel-cap/reciprocal math is well-defined for any `_wX ∈ [0,255]`, so an out-of-range stored `wk` yields a defined color rather than a fault. Consistent with keeping the range a UI convenience rather than a hard protocol contract (per the earlier thread on `setWhiteKelvin`). Happy to add it if you'd prefer strict symmetry with the `set.cpp` web-form clamp.
- **Range-gate the loadCfg checkbox to `[1000,10000]`** — the runtime enable-semantic is literally `wk != 0` (`_whiteKelvin == 0` in `autoWhiteCalc`), which the current `wk > 0` check mirrors. Gating the checkbox on the range would *mis*-represent an active out-of-range config as disabled; and for web-saved configs (already clamped by `set.cpp`) the two are equivalent.

— Claude (Anthropic's Claude Code; reviewed and hardware-tested by @NerdyGriffin)

wled00/bus_manager.cpp
NerdyGriffin: Since it divided by a constant, the compiler already converts that to a shift at compile time. But it can just as easily be explicitly written as a right shift

wled00/bus_manager.cpp
NerdyGriffin: Last time I worked through the math, I thought there was a case where _rwR can be zero, but I also want to simply this. I will look into this some more

wled00/xml.cpp
NerdyGriffin: Agreed, I will change that

wled00/bus_manager.cpp
NerdyGriffin: It was primarily intended for MODE_AUTO_ACCURATE and DUAL because they advertise themselves as an "accurate" translation of RGB into RGBW (and ACCURATE is the mode I use on my daily-driver WLED controllers). But it may be extended to include MODE_MAX if desired, although the `RGB + whiteKelvin -> RGBW` math might need to be different to match the intent of MODE_MAX.
I originally assumed MODE_MAX didn't need it because it isn't concerned about color accuracy, but I am open to corrections/suggestions.

Please let me know if I am misunderstanding what you are referring to as MODE_MIN and MODE_MAX

softhack007:
@coderabbitai full review

wled00/bus_manager.cpp
DedeHai: the compiler will not convert it, how could it?

NerdyGriffin: When compiled in an Ubuntu VM, the resulting machine-level code output by the compiled did use a shift in place of the multiple/divide operation. However, when I compiled it in Windows (after the merge from main, and with a new/different install of the build deps), the resulting machine-level code output by the compiler used a multiply-by-magic-constant sequence rather than a shift, so the result is provably inconsistent. So you are correct that it should have been coded to use shift explicitly.

Switched to `(w * (_wX + 1)) >> 8` in 5d4b69fa, matching the `(x + 1) * bri >> 8` form already used further down in this file. The cap calculation uses the same `_wX + 1` scaling so the two stay consistent and the subtraction stays underflow-safe. (Assisted by Claude)

wled00/bus_manager.cpp
NerdyGriffin: You're right that `_wR` and `_wG` can't be zero in practice (`_wG` only hits 0 below the 1000 K floor), but `_wB` is 0 at or below 1900 K, which is what the guards were for. I've removed them a different way in 5d4b69fa: the cached reciprocal is now `32768 / (_wX + 1)`, so the divisor can't be zero and the sentinel is gone. The cap becomes `(((x + 1) << 8) - 1) * _rwX >> 15`, which is the largest w whose contribution `(w * (_wX + 1)) >> 8` still fits the channel. Three multiply-and-shift ops, no branches.

I verified this on the host over every color at every 50 K from 1000 to 10000 K: no underflow anywhere, and the results match the previous arithmetic within a few units except below ~2350 K, where a near-zero blue coefficient now allows slightly more W for zero-blue inputs. The old cap dropped straight to 0 there, so this also removes a hard step at 1900 K.
(Written by Claude, proofread by @NerdyGriffin)

wled00/bus_manager.h
NerdyGriffin: Done in 5d4b69fa. With the `(_wX + 1)` divisor the reciprocal ranges from 128 to 32768, so it fits `uint16_t` exactly with no precision loss. I'll take a pass at grouping `_whiteKelvin`, the three coefficients and the three reciprocals with the existing small members in a follow-up commit so they don't introduce padding.

wled00/bus_manager.cpp
NerdyGriffin: Done in 33817dc

wled00/xml.cpp
NerdyGriffin: Done in 90096412b14ddf84ce37611fb2c1a1e85f1a9f44

wled00/bus_manager.cpp
NerdyGriffin: You are correct, I was mistaken. I originally designed that under the belief that `DUAL behaves as ACCURATE if w==0`, when in reality the code is `DUAL behaves as BRIGHTER if w==0`. I will try gating this to only the `RGBW_MODE_AUTO_ACCURATE` path.

NerdyGriffin: Change applied in 011b035aa627c5eca69b06340f8b1c266c364375, but I am still undecided which approach is the right call. Let me know if you would prefer the opposite, i.e. applying the auto-white behavior to all of `Accurate`, `Brighter`, `Dual`, and `Max`, rather than apply it only to `Accurate`.

wled00/bus_manager.h
```diff
@@ -118,6 +119,13 @@ class Bus {
     , _NPBbri(255)
     , _start(start)
     , _len(std::max(len,(uint16_t)1))
+    , _whiteKelvin(0)
+    , _rwR(32768 / 256) // Q15 reciprocal of (_wR+1) for _wR=255 (kept consistent though unused while _whiteKelvin==0)
```
softhack007, line 123: 32768/256 = 128. Is the special meaning of "32768/256" already documented in the code?

softhack007:
@coderabbitai full review. Please pay particular attention to typical AI mistakes, and to architectural design aspects that may create ripple effects or maintenance burden in the future.

wled00/set.cpp
NerdyGriffin: Fixed in f010e005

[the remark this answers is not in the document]
NerdyGriffin: Fixed in f010e005: parsed as `long`, range-checked, then cast. An oversized value now lands as 0 instead of wrapping into range.

wled00/cfg.cpp
```diff
@@ -238,6 +238,7 @@ bool deserializeConfig(JsonObject doc, bool fromFS) {
       bool refresh = elm["ref"] | false;
       uint16_t freqkHz = elm[F("freq")] | 0;  // will be in kHz for DotStar and Hz for PWM
       uint8_t AWmode = elm[F("rgbwm")] | RGBW_MODE_MANUAL_ONLY;
+      uint16_t whiteK = elm[F("wk")] | 0; // physical W channel color temperature in K (0 = neutral/legacy)
```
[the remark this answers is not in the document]
NerdyGriffin: Fixed in f010e005 with the same rule as the settings form: 1000..10000 is accepted, anything else becomes 0 (off). Went with reset-to-off rather than clamping to the nearest bound, since clamping would silently turn a typo into an active correction.

wled00/bus_manager.h
NerdyGriffin: Good catch, I missed that. The "32768/256" was derived from the `_rwR = 32768U / (_wR + 1U);` of `bus_manager.cpp`, but blindly substituting a "default" value of 255 for `_wR`. It should be simplified to the constant, with the comment explaining why it is that value

wled00/bus_manager.cpp
```diff
@@ -98,6 +98,39 @@ void Bus::calculateCCT(uint32_t c, uint8_t &ww, uint8_t &cw) {
...
+// recompute cached W-LED RGB equivalent when the configured Kelvin changes;
+// 0 means "treat the W LED as neutral white" which preserves legacy behavior
+// where autoWhiteCalc subtracted the same value from R, G, B.
+void Bus::setWhiteKelvin(uint16_t k) {
+  // The correction only makes sense for RGB+W buses with a single fixed white:
+  // dual-white CCT buses have a variable white point set via the CCT control,
+  // and buses without RGB or without W have nothing to correct. Force the
+  // feature off here (once, at configuration) so autoWhiteCalc's per-pixel
+  // path only has to test _whiteKelvin == 0.
+  if (_hasCCT || !_hasRgb || !_hasWhite) k = 0;
```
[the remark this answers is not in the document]
NerdyGriffin: Good catch, that was a real regression from moving the eligibility check into `setWhiteKelvin()`. Fixed in eb60a82c: the base `Bus` constructor now initialises `_hasRgb`, `_hasWhite` and `_hasCCT` to false, and derived constructors override as before.

wled00/data/settings_leds.htm
```diff
@@ -779,6 +841,17 @@
 							d.getElementsByName("RF"+i)[0].checked = v.ref;
 							d.getElementsByName("CV"+i)[0].checked = v.rev;
 							d.getElementsByName("AW"+i)[0].value   = v.rgbwm;
+							// AI: below section was generated by an AI
+							// stored wk: 0 = feature off; normalise like the backend (1000..10000 else 0)
+							{
+								const wkChkEl = gId("wke"+i);
+								const wkEl = d.getElementsByName("WK"+i)[0];
+								const raw = parseInt(v.wk, 10) | 0;
```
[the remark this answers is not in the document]
NerdyGriffin: Fixed in eb60a82c using `Number()` and `Number.isInteger()` as suggested, so the range check sees the real value. Fractional inputs are rejected too, matching the firmware's integer contract.

wled00/data/settings_leds.htm
```diff
@@ -368,6 +394,42 @@
...
+					// global override, so keep this control consistent with it: show when EITHER
+					// the per-bus mode OR the global override is Accurate.
+					const gAWel = d.Sf["AW"];
+					const gAW = gAWel ? parseInt(gAWel.value) : 255;
+					const isCap = (m) => (m === 2);
+					const wkBox = gId("dig"+n+"wk");
+					// only true single-white RGBW types: a fixed W-LED color temperature is
+					// meaningless for dual-white CCT buses (variable white point) and for
+					// non-RGB buses (nothing to derive the correction from)
+					if (wkBox) wkBox.style.display = (hasW(t) && hasRGB(t) && !hasCCT(t) && (isCap(awv) || isCap(gAW))) ? "inline" : "none";
```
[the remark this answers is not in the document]
NerdyGriffin: I think I agree, but I would like confirmation from @softhack007 or @DedeHai before I make that change, since it is essentially a UX preference. 

I would argue that the same is true for the per-bus mode itself: When the global mode is set, the UI still displays `AW${s}` even though the runtime ignores it. To keep the UI behavior consistent, I would argue that the UI should either show both (current behavior) or hide both.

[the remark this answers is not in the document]
NerdyGriffin: By "both", I mean `AW<n>` and `WK<n>`

PR #5656, opened 2026-05-30
Migrated AnimatedStaircase usermod to VL53L0X TOF sensors

Hoverman1977:
## Usermod
AnimatedStaircase_VL53L0X
A modification of the AnimatedStaircase  mod to be able to use VL53L0X I2C sensors instead of the ultrasonic sensors.

## Features
- Dual VL53L0X sensors
- Stairway lighting automation
- Configurable thresholds and timing

## Design rationale
I found that without a worker task the UI and WLED were extremely unresponsive as it appears as though the (continuous) reading of the sensors blocks other code from running. Implementing a worker tasks solved this issue.

I had found that 10ms would allow for large staircase's with many segments to light up smoothly (60 segments at 10ms per segment is 600ms) for the whole staircase. 10ms would also allow for smaller staircases as well as large ones.

When testing at 100ms it was taking 6 seconds or so to light up the entire staircase and I found the most effective lower limit was 10ms where the usermod would not affect other WLED functions yet still work effectively on a large staircase

## Tested On
- ESP32 DevKit and custom ESP32 LED controller board
- WLED 17.0.0-dev

## Build Status
- [x] Compiles cleanly
- [x] Tested on ESP32 DevKit and custom ESP32 LED controller board

## Summary by CodeRabbit

* **New Features**
  * Added Animated Staircase usermod using two VL53L0X distance sensors to drive segmented staircase animations with bidirectional swipe detection.
  * Includes per-sensor one-shot controls, JSON state exposure with an info UI toggle, configurable XSHUT pins, thresholds, timing, optional auto power-off, and MQTT motion updates.
* **Documentation**
  * Added README with wiring guidance, WLED/user setup steps, API/MQTT examples, and a sample JSON configuration.
* **Chores**
  * Added a usermod library manifest and a PlatformIO override sample for building the module.


usermods/AnimatedStaircase_VL53L0X/AnimatedStaircase_VL53L0X.cpp
```diff
@@ -0,0 +1,562 @@
...
+      xshutBottomPin
+    );
+
+    return true;
+  }
+
+  void ensureTask() {
+    if (taskHandle != nullptr) return;
+
+    xTaskCreatePinnedToCore(
```
softhack007: Why `xTaskCreatePinnedToCore()` and not just `xTaskCreate()`?
- there is nothing realtime-critical in your task, so no need to pin the task.
- pinning to core 1 means your usermod will crash on single core (-S2 and -C3)

platformio.ini
```diff
@@ -10,29 +10,29 @@
...
+               #lolin_s2_mini
+               #esp32c3dev
+               #esp32c3dev_qio
+               #esp32S3_wroom2
+               #esp32s3dev_16MB_opi
+               #esp32s3dev_8MB_opi
+               #esp32s3dev_8MB_qspi
+               #esp32s3dev_8MB_none
+               #esp32s3_4M_qspi
+               #usermods
```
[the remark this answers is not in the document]
softhack007: This change is a no-go for merging, sorry. Please revert, and leave platformio.ini untouched.

softhack007 (changes requested):

usermods/AnimatedStaircase_VL53L0X/AnimatedStaircase_VL53L0X.cpp
```diff
@@ -0,0 +1,562 @@
...
+      [](void* param) {
+        const TickType_t xFrequency = 10 / portTICK_PERIOD_MS;
```
softhack007, line 260: Are you sure this task must run at 100Hz?

usermods/AnimatedStaircase_VL53L0X/AnimatedStaircase_VL53L0X.cpp
```diff
@@ -0,0 +1,562 @@
...
+    for (int i = minSegmentId; i <= lastSeg && i < maxSegmentId; i++) {
+      Segment& seg = strip.getSegment(i);
+      if (!seg.isActive()) continue;
+
+      seg.setOption(SEG_OPTION_ON, i >= onIndex && i < offIndex);
+    }
+
+    strip.trigger();
+    stateChanged = true;
+    colorUpdated(CALL_MODE_DIRECT_CHANGE);
```
softhack007: Don't call this function in a background task. Please use a volatile synchronisation flag (or mutex, queue), and move colorUpdated() into the main usermod loop().

usermods/AnimatedStaircase_VL53L0X/AnimatedStaircase_VL53L0X.cpp
```diff
@@ -0,0 +1,562 @@
...
+    if (bottomVL53.timeoutOccurred()) return false;
+    if (topVL53.timeoutOccurred()) return false;
+
+    bottomSensorRead = bottomSensorWrite || (bottomReading < bottomThresholdMM);
+    topSensorRead = topSensorWrite || (topReading < topThresholdMM);
+
+    if (bottomSensorRead != bottomSensorState) {
+      bottomSensorState = bottomSensorRead;
+      sensorChanged = true;
+      publishMqtt(true, bottomSensorState ? "on" : "off");
```
softhack007, line 317: Don't call this function in a background task. Please use a volatile synchronisation flag (or mutex, queue), and move publishMqtt() into the main usermod loop().

softhack007:
@coderabbitai full review

usermods/AnimatedStaircase_VL53L0X/AnimatedStaircase_VL53L0X.cpp
```diff
@@ -0,0 +1,562 @@
...
+            continue;
+          }
+
+          if (!self->initDone || strip.getMaxSegments() == 0) continue;
+          if (strip.getSegment(0).stop == 0) continue;
+
+          self->minSegmentId = strip.getMainSegmentId();
+          self->maxSegmentId = strip.getLastActiveSegmentId() + 1;
+
+          self->checkSensors();
```
softhack007: Your design of using a separate task is good to ensure swift responses to sensor changes, but I'm still wondering if its "overkill" for the purpose of this usermod. The core of the task function is that it runs self->checkSensors() every 10 millis. Calling self->autoPowerOff() and self->updateSwipe() is not safe inside the task function, because this changes segment state, triggers UDP and MQTT etc - all without being synchronized with the main wled loop.

So you'd need to add semaphores and move the function call that interact with the wled core back into the usermod::loop() context, otherwise it won't work reliably.

softhack007:
General comment: this PR lacks a description, please add it.

The description should also provide some design rationale, especially why you chose to implement a worker task for polling the sensors.

https://github.com/wled/WLED?tab=contributing-ov-file#describing-your-pr

usermods/AnimatedStaircase_VL53L0X/platformio_override.ini.sample
softhack007, line 1: usually this file is named `platformio_override.ini.sample`. consider renaming.

usermods/AnimatedStaircase_VL53L0X/platformio_override_example.ini
```diff
@@ -0,0 +1,14 @@
...
+default_envs = esp32dev_usermod_AnimatedStaircase_VL53L0X
+
+# Animated Staircase VL53L0X sensors
+[env:esp32dev_usermod_AnimatedStaircase_VL53L0X]
+extends = env:esp32dev
+custom_usermods =
+  AnimatedStaircase_VL53L0X
+lib_deps = 
+  ${env:esp32dev.lib_deps}
+  pololu/VL53L0X @ ^1.3.0
```
softhack007: this line should not be necessary => move dependencies into `library.json`

usermods/AnimatedStaircase_VL53L0X/platformio_override_example.ini
```diff
@@ -0,0 +1,14 @@
...
+build_flags =
+  ${env:esp32dev.build_flags}
+  -D USERMOD_ANIMATED_STAIRCASE_VL53L0X
```
softhack007: normally not needed, either.

usermods/AnimatedStaircase_VL53L0X/AnimatedStaircase_VL53L0X.cpp
Hoverman1977: I had found that 10ms would allow for large staircase's with many segments to light up smoothly (60 segments at 10ms per segment is 600ms) for the whole staircase. Before replying now, I did test it at 100 and 250 and they are both very slow.

The minimum segment time is dictated by the tick rate, so if I set it to 100, for example, it would take 6 seconds for a 60 segment staircase to light up fully, as the minimum segment delay would be 100ms. I'm not sure there is a better way to do it?

usermods/AnimatedStaircase_VL53L0X/AnimatedStaircase_VL53L0X.cpp
Hoverman1977: I think that has been done now in the latest commit - this is my first usermod - and first WLED PR - my testing had been successful locally, but I can only test so much. I did find that it took too long to turn on all the segments when I set the tickrate to a slower iteration which is why I settled on 10.

platformio.ini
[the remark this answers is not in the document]
joshua-dixon-doe: This file has been reverted and removed from the PR

DedeHai:
PR is a good addtion but IMHO too niche to merge, should be out of tree.

Issue #5613, opened 2026-05-14
WLED - buttons label incorrect


xplosv1:
### What happened?

In the latest WLED version (16.0.0) in the action button presets each button is labeled wrong and or erroneously

<img width="607" height="161" alt="Image" src="https://github.com/user-attachments/assets/fe41dae6-41da-4c85-a067-9170d8c95a63" />

push/switch is single press
short is long press
long is double press

### To Reproduce Bug

Set presets, push button 

### Expected Behavior

push/switch is single press -> should be labelled as press
short is long press -> should be labelled as long press
long is double press -> should be labelled as double press 

### Install Method

Binary from WLED.me

### What version of WLED?

WLED 16.0.0 (2605030)

### Which microcontroller/board are you seeing the problem on?

ESP32-C3

### Relevant log/trace output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

softhack007:
@coderabbitai please compare the related source code of the `main` and `0_15_x` branch. Is the reported behaviour a regression since release 0.15.4? If yes, what is the root cause?

softhack007:
@coderabbitai which commit introduced this regression?

DedeHai:
@benjamw are you able to provide a fix?

mkdzabel:
This Info just saved my night. Was debugging endlessly when trying to trigger a playlist via PIR. Can confirm the bug. Changing to short switches correctly from off to on when motion is detected.

benjamw:
I'll get a fix in soon.

DedeHai:
@benjamw it was mentioned that analog buttons no longer work with the dropdowns, see https://kno.wled.ge/features/macros/#analog-button 

can you please check if this is the case and open an issue (or a fix PR)?

DedeHai:
@benjamw in the macros I see this:

<img width="737" height="325" alt="Image" src="https://github.com/user-attachments/assets/56c4c267-c256-49d0-bf61-7aae16164f54" />

but I think it shold be "short press" , "long press", "double press". can you check?

benjamw:
I've added more to the PR. Try again. This is what I'm seeing:

<img width="550" height="653" alt="Image" src="https://github.com/user-attachments/assets/8e4bf7d1-8623-4bde-bce3-36103e59e387" />

benjamw:
What branch are you looking at? My fix PR for this issue went into `main`.

My new PR for the updated analog buttons also is based off of `main`.

[...]

PR #5659, opened 2026-06-01
Analog button fix

benjamw:
This fixes the issues noted in https://github.com/wled/WLED/issues/5613#issuecomment-4590027231

This also adds clarification for the differences between analog and digital inputs. Showing and hiding dropdowns as needed to make those selections clearer.

## Release Notes

* **New Features**
  * Enhanced button action preset configuration with type-specific UI rendering for analog, switch, and standard pushbutton types.
  * Updated preset labeling to display "Default Action (0)" for improved clarity.
  * Simplified section heading for button action presets.

### DIFF
--- a/wled00/data/settings_time.htm
+++ b/wled00/data/settings_time.htm
@@ -219,16 +219,37 @@
 			rPS(sel, presetOpts, "data-preset");
 		}
 	}
+	function bAO() { // buildAnalogOptions: analog functions + per-segment opacity (segment 0 included; MD=0 => segment 0)
+		var o = '<optgroup label="Analog Functions"><option value="250">Global brightness (250)</option><option value="249">Effect speed (249)</option><option value="248">Effect intensity (248)</option><option value="247">Palette (247)</option><option value="200">Primary color hue (200)</option></optgroup><optgroup label="Analog Segment Opacity">';
+		for (var j=0; j<=32; j++) o += `<option value="${j}">Segment ${j} (${j})</option>`;
+		o += '</optgroup>';
+		return o;
+	}
+	function isAnalogBtn(t) { return t==7 || t==8; } // BTN_TYPE_ANALOG / BTN_TYPE_ANALOG_INVERTED
+	function btnTypeName(t) { // mirrors the button type dropdown on the LED settings page
+		switch (+t) {
+			case 2: return 'Pushbutton';
+			case 3: return 'Push inverted';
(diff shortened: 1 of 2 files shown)


wled00/data/settings_time.htm
```diff
@@ -246,33 +270,54 @@
...
-				<div class="ba">
-					<label>Long (off → on)</label>
-					<select name="ML${b}" class="s" required>${presetOpts}</select>
+		buttonBlock.setAttribute('data-btype', t); // read back by rBPO() to rebuild selects correctly
+		if (analog) {
+			// analog buttons: MD holds the function/segment; short/long press are unused and forced to 0
+			buttonBlock.innerHTML = `
+				<div class="bh">Button ${i} - ${typeName}</div>
+				<div class="bs">
+					<input type="hidden" name="MP${b}" value="0">
```
DedeHai: do we require the hidden values? i.e. they are set to default zero in firmware if missing, see set.cpp

wled00/data/settings_time.htm
```diff
@@ -246,33 +270,54 @@
 		if (d.Sf.LTR.value==="S") { d.Sf.LT.value = -1*parseFloat(d.Sf.LT.value); }
 		if (d.Sf.LNR.value==="W") { d.Sf.LN.value = -1*parseFloat(d.Sf.LN.value); }
 	}
-	function addRow(i,p,l,d) {
+	function addRow(i,p,l,d,t) {
+		if (t===undefined) t = 0;
 		var b = String.fromCharCode((i<10?48:55)+i);
-		var presetOpts = '<option value="0">Default Action</option>' + sortedPresetOptions;
+		var analog = isAnalogBtn(t);
```
DedeHai: single use variable adds code for no benefit

DedeHai:
thanks, looks good to me, just two minor things.

edit:
the rabbit has a point about the button vs. switch: since now the page knows the button type and you already went through the effort of labelling the buttons, the function labels could be type specific. Is that something that can be done by adding a few more lines without duplicating the whole block?

benjamw:
I agree. I didn't want to scope creep the PR, but had thought of making the labels dynamic.

benjamw:
I've added functionality to make each button type have its own options available as well as remove anything that needs to default to 0.

Image

DedeHai:
cherry picked into 16_x

PR #5662, opened 2026-06-03
HUB75 bugfixes for 4-scan and chained panels + new "Seengreat" pinout

softhack007:
Improvements:
* new -S3 pin-out for Seengreat RGB Matrix Adapter Board (https://seengreat.com/wiki/186)
* allow up to 128 pixels wide panels, prevent uint8 overflow for 128px panels

Bugfixes for 4-scan (aka QS) panels
* prevent panels going flatter each time that cfg.json is saved (32x32->16x64->8x128->4x128)
* correct VirtualMatrixPanel setup: need to use real panel dimensions, not modified `mxconfig` dimensions
* only set chainType when chain length > 1
* use a chaintype that does not flip the display upside-down

## Summary

* **New Features**
  * Added support for an additional ESP32-S3 HUB75 pinout configuration for broader hardware compatibility.

* **Bug Fixes**
  * Improved HUB75 panel initialization for HS vs QS types, including corrected virtual/quad-scan sizing and chaining behavior.
  * Updated reported pin/matrix dimensions for quad-scan layouts.
  * Fixed pixel color reads to return consistent RGB values.
  * Improved robustness by handling display/driver allocation failures gracefully with fallback.
  * Added clearer handling for unsupported/unknown HUB75 panel configurations.


wled00/bus_manager.cpp
```diff
@@ -1011,15 +1018,15 @@ BusHub75Matrix::BusHub75Matrix(const BusConfig &bc) : Bus(bc.type, bc.start, bc.
   // chained panels with cols and rows define need the virtual display driver, so do quarter-scan panels
   if (chainLength > 1 && (_rows > 1 || _cols > 1) || bc.type == TYPE_HUB75MATRIX_QS) {
     _isVirtual = true;
-    chainType = CHAIN_BOTTOM_LEFT_UP; // TODO: is there any need to support other chaining types?
-    DEBUGBUS_PRINTF_P(PSTR("Using virtual matrix: %ux%u panels of %ux%u pixels\n"), _cols, _rows, mxconfig.mx_width, mxconfig.mx_height);
+    if (chainLength > 1 && (_rows > 1 || _cols > 1)) chainType = CHAIN_TOP_RIGHT_DOWN; // we need to use a _DOWN chainType, otherwise the display is upside-down
```
DedeHai: the condition is a duplicate, better to change the initial condition `bc.type == TYPE_HUB75MATRIX_QS`

also what effect does chainType have? CHAIN_BOTTOM_LEFT_UP works fine on my chained panel.

netmindz: The chain type is typically used for multiple physical panels and is a common technique of physically instal panels upside down to keep cables shorter, see the driver docs for details.

I've yet to see a true 128 wide panel myself, the ones I've seen have been two 64x64 mounted side by side, so you need a chain length of 2 per panel, but that could get tricky once you then try and use multiple physical panels to get both halves of the panel the right way up

wled00/bus_manager.cpp
```diff
@@ -1155,8 +1162,8 @@ std::vector<LEDType> BusHub75Matrix::getLEDTypes() {
 
 size_t BusHub75Matrix::getPins(uint8_t* pinArray) const {
   if (pinArray) {
-    pinArray[0] = mxconfig.mx_width;
-    pinArray[1] = mxconfig.mx_height;
+    pinArray[0] = _isVirtual ? mxconfig.mx_width  /2 : mxconfig.mx_width;
```
netmindz: Originally I was tracking QS virtual panels separately from regular use of virtual panels, we possibly need to use separate flag not the isVirtual

wled00/bus_manager.cpp
```diff
@@ -887,6 +888,12 @@ BusHub75Matrix::BusHub75Matrix(const BusConfig &bc) : Bus(bc.type, bc.start, bc.
   // HUB75_I2S_CFG::i2s_pins _pins={R1_PIN, G1_PIN, B1_PIN, R2_PIN, G2_PIN, B2_PIN, A_PIN, B_PIN, C_PIN, D_PIN, E_PIN, LAT_PIN, OE_PIN, CLK_PIN};
   mxconfig.gpio = {4, 5, 6, 7, 15, 16, 18, 8, 3, 42, 9, 40, 2, 41};
   
+#elif defined(SEENGREAT_V2_PINOUT)
```
netmindz: Probably worth adding the S3 check to that define

wled00/bus_manager.cpp
softhack007: > also what effect does chainType have? CHAIN_BOTTOM_LEFT_UP works fine on my chained panel.

This should - normally - only be relevant when you have chained panels. Similar our 2D setup, chained panels can be connected from top down, left to right or even "sepentine" to save wire length. 

For my 4-scan panel, CHAIN_BOTTOM_LEFT_UP always results in 180° rotated display where everything is upside-down, even without chaining panels. TOP_RIGHT_DOWN is the only one where I still had the "correct" orientation. I've  compared to CHAIN_NONE and to 2-scan mode, just to be sure I'm not holding the panel wrongly.

wled00/bus_manager.cpp
softhack007: More to come - the board has a second pinout for esp32 classic, and it even comes in "V1" and "V2" pinout variants. 4 new printouts in total 😅

DedeHai:
@softhack007 offtopic: what is up with the rabbit adding random labels?

softhack007:
> @softhack007 offtopic: what is up with the rabbit adding random labels?

That was an experiment, I've allowed it to directly add the labels it suggests. Initially it looked meaningful, but it's true  we see to many "random" labels. I'll switch that feature off.

wled00/bus_manager.cpp
softhack007: I agree, will add a separate flag `_isQuadScan` to de-couple them

wled00/bus_manager.cpp
```diff
@@ -1007,18 +1052,18 @@ BusHub75Matrix::BusHub75Matrix(const BusConfig &bc) : Bus(bc.type, bc.start, bc.
...
     _isVirtual = true;
-    chainType = CHAIN_BOTTOM_LEFT_UP; // TODO: is there any need to support other chaining types?
-    DEBUGBUS_PRINTF_P(PSTR("Using virtual matrix: %ux%u panels of %ux%u pixels\n"), _cols, _rows, mxconfig.mx_width, mxconfig.mx_height);
+    DEBUGBUS_PRINTF_P(PSTR("Using virtual matrix: %ux%u panels of %ux%u pixels\n"), _cols, _rows, mxconfig.mx_width/2, mxconfig.mx_height*2);
   }
   else {
     _isVirtual = false;
   }
 
   if (_isVirtual) {
```
softhack007, line 1067: ToDO: _isVirtual logic is still too tangled - cleanup, and make sure that non-QS panels can be _isVirtual, too

> **Split “uses virtual panel” from “quarter-scan geometry was rescaled.”**
> 
> This is the same `_isVirtual` conflation called out earlier: regular chained panels now enter the virtual path too, but `VirtualMatrixPanel(...)` and `getPins()` always undo QS scaling with `/2` and `*2`. For non-QS chains that corrupts both runtime mapping and the dimensions written back to config.

softhack007:
toDO

> wled00/bus_manager.cpp (1)
> > `828-845`: _⚠️ Potential issue_ | _🟠 Major_ | _⚡ Quick win_
> > **Move the 64px chain-memory clamp below the real height assignment.**
> > This guard now reads `mxconfig.mx_height` before the constructor copies `panelHeight` into it, so the single-panel fallback is based on stale/default state instead of the current bus config. That can let 64px chained panels bypass the safeguard, or clamp smaller panels unnecessarily depending on the library default.

wled00/bus_manager.cpp
```diff
@@ -821,7 +823,7 @@ BusHub75Matrix::BusHub75Matrix(const BusConfig &bc) : Bus(bc.type, bc.start, bc.
 
   mxconfig.clkphase = bc.reversed;
   // allow chain length up to 4, limit to prevent bad data from preventing boot due to low memory
-  mxconfig.chain_length = max((uint8_t) 1, min(chainLength, (uint8_t) 4));
+  mxconfig.chain_length = max(1U, min(chainLength, 4U));
 
   if (mxconfig.mx_height >= 64 && (mxconfig.chain_length > 1)) {
```
softhack007: toDO

> wled00/bus_manager.cpp (1)
> > `828-845`: _⚠️ Potential issue_ | _🟠 Major_ | _⚡ Quick win_
> > **Move the 64px chain-memory clamp below the real height assignment.**
> > This guard now reads `mxconfig.mx_height` before the constructor copies `panelHeight` into it, so the single-panel fallback is based on stale/default state instead of the current bus config. That can let 64px chained panels bypass the safeguard, or clamp smaller panels unnecessarily depending on the library default.

softhack007:
@coderabbitai full review

softhack007:
To-Do list for testing:
* [x] Quad-scan 32x32x2, horizontal
* [x] Quad-scan 32x32x2, vertical
* [x] Quad-scan 32x32
* [x] Dual-Scan 64x64
* [x] Dual-Scan 64x64x2, horizontal

softhack007:
@coderabbitAI please cross-check if this PR causes any regression in the "DS" DualScan path (single panel and chained panels). The improvement that chained DS setups can now use "VirtualMatrixPanel" is not a regression, however the panel/overall dimensions should still be handled correctly.

softhack007:
🤔 The wrong panel dimensions problem came back for double-scan panels. but only with a "-O2" build. Everything works as it should in a normal "-Os" build.

This might be a pre-existing read/write race / memory error / use-after-free issue that only surfaces with -O2.  WLED is built with -Os, so this problem is not blocking, but worth to chase this down later.


Edit: my fault, not a bug. I've accidentally booted an old firmware partition 😅

wled00/bus_manager.cpp
```diff
@@ -1089,7 +1141,7 @@ void IRAM_ATTR BusHub75Matrix::setPixelColor(unsigned pix, uint32_t c) {
 uint32_t BusHub75Matrix::getPixelColor(unsigned pix) const {
   if (!_valid) return IS_BLACK; // note: no need to check pix >= _len as that is checked in containsPixel()
   if (_ledBuffer)
-    return uint32_t(_ledBuffer[pix]);
+    return uint32_t(_ledBuffer[pix]) & 0x00FFFFFF;  // FastLED 32bit is RGBA, we need RGBW
```
DedeHai: no need for the mask, see fastled_slim
  // converts a CRGB to a 32-bit color with white = 0
  inline explicit operator uint32_t() const {
    return (uint32_t{r} << 16) |
           (uint32_t{g} << 8)  |
            uint32_t{b};
  }

DedeHai (approved):
I have no hardware but changes look good to me besides the mask

wled00/bus_manager.cpp
softhack007: thanks, good point. I still have too much original FastLED in my head ;)

PR #5663, opened 2026-06-03
Add clocked playlist scheduling

aeoa:
## What

Adds an optional `clocked` mode for playlists. A clocked playlist derives its active entry from synchronized Toki time, so multiple WLED controllers with matching playlists and presets can stay aligned without continuously communicating with each other.

This is useful for setups where controllers may have spotty network connectivity after initial time sync, where relying on frequent live sync packets is undesirable, or where controllers are not on the same local network but can sync to the same time source.

## How

When `playlist.clocked` is enabled, playlist handling uses absolute Toki time modulo the total finite playlist duration to select the active entry. The selected entry offset is also applied to `strip.timebase`, so effects using `strip.now` can start from the same relative effect time.

Shuffle remains deterministic per wall-clock playlist cycle, so matching controllers derive the same shuffled order.

The playlist editor now includes a `Clock sync` checkbox. Manual advance is disabled for clocked playlists because zero-duration entries cannot be placed on a wall-clock timeline.

## Limitations

Clocked playlists require an existing time source, such as NTP enabled in WLED Time settings. They do not enable or change NTP settings automatically.

Zero-duration/manual-advance entries are unsupported. Nested playlists keep existing WLED behavior; child playlist duration wins once loaded.

## Testing

Hardware sanity-tested on two ESP32-C3 during development

## Summary by CodeRabbit

* **New Features**
  * Added per-playlist **“Clock sync”** to drive playlist progression from wall-clock time (saved with the playlist).
  * Added per-playlist **deterministic shuffle** for repeatable shuffle order.
  * Updated the playlist editor with a **Clock sync** control and unified wiring so **shuffle/deterministic/clock-sync/manual/repeat** settings reflect the active timing mode.

* **Bug Fixes / Improvements**
  * Enforced **mutual exclusivity** between **Clock sync** and **Manual advance**; improved duration normalization and made duration inputs behave more clearly.
  * Older playlist presets now **backfill missing fields**, improving compatibility and UI consistency (including repeat/end behavior and visibility).


softhack007:
> A clocked playlist derives its active entry from synchronized Toki time, so multiple WLED controllers with matching playlists and presets can stay aligned without continuously communicating with each other.

@aeoa can you explain how this feature behaves different from time-controlled presets (settings -> time & macros -> Time-Controlled Presets) ?

aeoa:
> > A clocked playlist derives its active entry from synchronized Toki time, so multiple WLED controllers with matching playlists and presets can stay aligned without continuously communicating with each other.
> 
> @aeoa can you explain how this feature behaves different from time-controlled presets (settings -> time & macros -> Time-Controlled Presets) ?

Time-Controlled Presets trigger a specific preset / playlist at a specific time of day.
Clocked playlists are continuous wall-time-derived playback. The intended use case is multiple controllers with the same playlist: each controller derives the active playlist entry and effect offset from synced time, so they select the same preset within the playlist at the same wall-clock moments, regardless of when the playlist was loaded.

DedeHai:
do I understand correctly that what this tries to solve is time drift of different controllers over long periods of time where a playlist plays? like several hours? how much drift did you observe without this PR?

aeoa:
> do I understand correctly that what this tries to solve is time drift of different controllers over long periods of time where a playlist plays? like several hours? how much drift did you observe without this PR?

The actual problem I am trying to solve is this: I have several separate mobile LED installations, each with their own controller but with the same playlist preset set to apply on boot. I want those installation to show the same effect at the same time such that they appear as one connected unit. We cannot rely on the WLED sync feature, because there is no reliable WIFI connection, and the controllers may not all be on the same network.

The feature of this PR is that clocked playlists schedule their entries in a deterministic manner relative to wall clock. Say we have a playlist with three presets A, B, and C, each with a duration of 10s. Each controller is started at a different point in time. Until first NTP time fix, the first effect of the playlist is shown as a fallback. Once wall time is available, the playlist determines the correct effect for the current point in time and stays locked to wall time.

```
Wall clock:      0s        10s        20s        30s        40s
Timeline:        A----------B----------C----------A----------B---- ...

Controller 1:
Before NTP:                   A---
After NTP:                        B----C----------A----------B---- ...

Controller 2:
Before NTP:                                A---
After NTP:                                     C--A----------B---- ...
```

This also works for shuffled playlists, because the seed for the randomization is also derived from the wall clock.

The actual time synchronization is left to the Toki implementation already present in WLED. With NTP, time is only updated about twice a day. Within that time frame, the controller clocks may drift apart by a couple of seconds. Perhaps another PR could make that NTP update interval configurable.

DedeHai:
ok got it.
while this seems like a useful feature it is very specific - multiple controllers, need sync, controllers on different wifi's - not something that has broad use IMHO. This should be implemented as a usermod.

softhack007:
> This also works for shuffled playlists, because the seed for the randomization is also derived from the wall clock.

Did your AI tell you? Because the playlist shuffle code uses the hardware random generator, which is *not* pseudo-random but a true random number based on entropy. `random()` does not use any seed that is based on time.

https://github.com/wled/WLED/blob/bd0ebf90c074cd54141a93790ebf4213177c5101/wled00/playlist.cpp#L35

https://github.com/wled/WLED/blob/bd0ebf90c074cd54141a93790ebf4213177c5101/wled00/fcn_declare.h#L484-L485

aeoa:
> > This also works for shuffled playlists, because the seed for the randomization is also derived from the wall clock.
> 
> Did your AI tell you? Because the playlist shuffle code uses the hardware random generator, which is _not_ pseudo-random but a true random number based on entropy. `random()` does not use any seed that is based on time.
> 
> https://github.com/wled/WLED/blob/bd0ebf90c074cd54141a93790ebf4213177c5101/wled00/playlist.cpp#L35
> 
> https://github.com/wled/WLED/blob/bd0ebf90c074cd54141a93790ebf4213177c5101/wled00/fcn_declare.h#L484-L485

Normal playlists still use true randomness through `shufflePlaylist()`. This function shuffles the `playlistEntries` in-place. This is stateful and thus cannot be used for clocked playlists.

Instead, a separate set of RNG functions are used for clocked playlists:
https://github.com/aeoa/WLED/blob/5d3853872181f79dd1fb792cb18234abc04d6f8a/wled00/playlist.cpp#L53:L67

The seed is computed from the playlist cycle, which is computed from the wall time and playlist duration:
https://github.com/aeoa/WLED/blob/5d3853872181f79dd1fb792cb18234abc04d6f8a/wled00/playlist.cpp#L97:L100

The seed is then used shuffle an array of playlist entry indices:
https://github.com/aeoa/WLED/blob/5d3853872181f79dd1fb792cb18234abc04d6f8a/wled00/playlist.cpp#L104:L113

This is then used in `handlePlaylist()`.

wled00/playlist.cpp
```diff
@@ -41,6 +42,31 @@ void shufflePlaylist() {
...
+  if (durMs > 0) return durMs;
+  return fallbackDur > 0 ? fallbackDur : CLOCKED_PLAYLIST_DEFAULT_DUR;
+}
+
+static uint32_t playlistShuffleRand(uint32_t &state) {
+  state ^= state << 13;
+  state ^= state >> 17;
+  state ^= state << 5;
+  return state;
+}
```
softhack007: Do you have a reference on the origin of this PRNG function? (wikipedia, research papers, etc?) 
Is it more than a "random guess" by an AI? How does it compare to other pseudo-random number generators?

Just for info, WLED also has a PRNG class, maybe you could use it - see `wled00/prng.h`

softhack007: Edit: looks like your PRNG is a simple "XorShift" https://en.wikipedia.org/wiki/Xorshift. 
Fast for gaming, but not very good randomness, unfortunately. The wikipedia article has some suggestions on how to improve the random number quality ([`xorshitf*`](https://en.wikipedia.org/wiki/Xorshift#xorshift*) is an easy improvement).

For a list of other possible PRNG codes, check https://en.wikipedia.org/wiki/List_of_random_number_generators#Pseudorandom_number_generators_(PRNGs)

aeoa: I found the PRNG on wikipedia, it's a xorshift32 variant from a book by Marsaglia: https://en.wikipedia.org/wiki/Xorshift

I'll take a look at the existing WLED PRNG, agree we should use that one instead.

aeoa: I changed the implementation to use the existing 16-bit PRNG. To me it looks random enough for the purpose.

DedeHai: the existing PRNG is tuned for simplicity. I wrote a brute-force function to check parameters for randomness and repetition and found a few, the one I picked for PRNG shows decent randomness in human standards (i.e. no obvious patterns and no loops).

softhack007:
> while this seems like a useful feature it is very specific - multiple controllers, need sync, controllers on different wifi's - not something that has broad use IMHO. This should be implemented as a usermod.

@DedeHai i share your view about this feature is "just" for a very specific usecase. However it might be very hard to implement it as a usermod, without re-implementing playlists at usermod level. Maybe its better to have the "clocked playlist steps" encapsulated with a feature flag (like WLED_ENABLE_TIME_IN_PLAYLIST), disabled by default? 

Any other feature flag name that is not too long should be ok...

As a side-effect, we'd need to test that a "timed playlist step" will be ignored by a normal build.

willmmiles:
I think there's a valid use case for a feature like this, but I really don't like the implementation by creating a second, independent code path in the playlist logic.  This approach will create a maintenance burden for any future playlist features: they'll have to be implemented twice.  The fact that you had to consider `nightLightActive`, handle zero-duration entries and repeats as special cases is an architecture smell.   The existence of `applyPlaylistEntry` as a function is another: if there are multiple paths that need to call it, something has gone wrong with the code design.

(FWIW, this is a common problem with AI-generated code architectures -- it's biased towards copy-and-paste and bolting on new logic as an aside, "minimizing changes to the existing code", vs. rethinking the design to accommodate new goals properly.)

Instead I suggest breaking down the feature into its primitive components:
- An option to use a synchronizable PRNG for shuffling playlists.  I think this is a useful feature in isolation and an easy extension to `shufflePlaylist()`.  
- An option to choose the clock type for playlist calculations (millis() vs Toki::Time -- or CLOCK_MONOTONIC vs CLOCK_REALTIME for my unix friends).   
- "Skip ahead" handling for the CLOCK_REALTIME case.   An explicit resync on a change of `toki::getTimeSource()` might be necessary.

The latter two could be implemented in the context of the existing `handlePlaylist()` function by augmenting the test of whether or not to change the preset, while keeping the apply handling as-is.

Zero-duration entries should be treated with the existing semantics, ie. treated as terminal when applied.   The fallback logic isn't worth the code complexity and weird special cases -- documenting exceptions to the usual behaviour is a huge pain in the butt and makes it hard for people to understand.

On the bikeshed front: I don't love "clocked" as a name, either -- the past-tense form makes me think of hardware devices with clock strobe lines.  Maybe "real time" or "wall time" or "absolute time".   Thoughts?

DedeHai:
fully agree with @willmmiles 
also: the static array is a no no.

re: naming 
I also do not like clocked as a name, the UI should be as intuitive as possible, few do RTFM so using a name that has intrinsic meaning is key.  I would throw in "clock mode" - it's a pity english does have double meanings for both clock and watch ;)

aeoa:
Thank you for the good feedback @willmmiles. I have reworked the implementation according to your suggestions.

Changes:
- I have renamed clocked to clock sync. What do you think?
- Deterministic shuffle and wall clock sync are now separate features that work independently and have their own checkboxes. 
    * Only enabling deterministic shuffle will produce the same pseudo-random order every time the playlist is triggered.
    * Only enabling clock sync (and normal shuffle) causes effects to be switched at the same time, but different ones are chosen on each controller.
- Finite repetitions and end preset now also work for clock sync mode. The initial sync with wall time will probably cause one repetition to be considered done.
- Zero-durations are no longer normalized. In clock sync mode, they are currently implicitly skipped / ignored since the function `mapCycleTimeToPlaylistSlot` will never select them. To me this seems fine, I cannot come up with a good reason to combine infinite entry durations with clock synched playlists. Doing this properly seems complicated.
- There is only one shuffle function that supports both PRNG and hardware random.
- There is a now `playlistOrder` index array that is allocated together with `playlistEntries`. Shuffling now only changes the order of `playlistOrder`. This has the advantage that the order can be easily reset, which is needed for deterministic shuffle. Normal shuffle no longer stacks on previous shuffles, the order is always reset, then shuffled. Should not make a difference for truly random shuffles.
- The `handlePlaylist` function now has one shared path instead of a completely separate function for clock sync. It is still a bit more complex than I'd like it to be, but there is now only one place for handling rollovers, shuffle, and applying presets.
- A shared UI handler for Shuffle, Deterministic, Clock sync, Manual advance, Repeat (indefinitely) and End preset to enforce constraints.

I have tested this a bit on two ESP32C3, and it seems to do the right thing. I have not looked into the nightlight.

Does this go in the right direction?

willmmiles:
Looking much better!  I don't have time to do a deep review today, unfortunately - I'll give more detailed feedback next week.  Some quick thoughts:

- The memory cost of storing the shuffle order separately from the playlist order is not insignificant, and we all pay the cost, needed or not.  I'd like to think about alternatives before committing to that approach.  (I understand your objective is to support late joiners.)
- Avoid running `mapCycleTimeToPlaylistSlot()` on every pass -- it's expensive to run an O(n) check every `loop()`.   Instead check the time against the current playlist item duration (or end time), like the "regular" playlist logic.  The initial mapping can be done at playlist load, and an explicit remap performed only if `toki::getTimeSource()` changes (a much faster check).  Otherwise it should be safe to assume that any further clock adjustments are small, and we don't need to worry about jumping ahead or backwards very far.

softhack007:
@coderabbitai full review

softhack007:
This might also be something to look into:

> 🧹 Nitpick comments (1)
> > wled00/playlist.cpp (1)> `292-294`: _🚀 Performance & Scalability_ | _🔵 Trivial_ | _⚡ Quick win_
> > > **Avoid scanning the playlist on every main-loop iteration.**
> > > `handlePlaylist()` runs from the main loop, and this calls `mapCycleTimeToPlaylistSlot()` every time clock sync is valid. Cache the current slot window and only remap when the cycle changes, the current entry window no longer contains `clockSyncCycleTime`, or a deliberate resync interval expires.

wled00/playlist.cpp
```diff
@@ -126,9 +204,15 @@ int16_t loadPlaylist(JsonObject playlistObj, byte presetId) {
...
   if (playlistEndPreset > 250) playlistEndPreset = 0;
-  shuffle = shuffle || playlistObj["r"];
+  shuffle = shuffle || playlistObj["r"].as<bool>();
   if (shuffle) playlistOptions |= PL_OPTION_SHUFFLE;
 
+  bool clockSync = playlistObj[F("clockSync")].as<bool>();
+  if (clockSync) playlistOptions |= PL_OPTION_CLOCK_SYNC;
+
+  bool deterministicShuffle = playlistObj[F("deterministic")].as<bool>();
+  if (shuffle && deterministicShuffle) playlistOptions |= PL_OPTION_DETERMINISTIC_SHUFFLE;
```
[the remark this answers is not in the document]
willmmiles: Sorry @coderabbitai, we went through this already -- please make sure to read the PR when doing a review.  We do not want to exclude or special-case 0-length entries.  The intent is explicitly that they should be skipped.

Issue #5661, opened 2026-06-03
Live data override breaks gamma correction


SirDella:
### What happened?

I updated from 0.15.4 to 0.16 and noticed that my presets looked different. WLED wasn't applying gamma correction while Live Data Override was on.

### To Reproduce Bug

- Select a solid color
- Start a live data stream (in my case, Hyperion)
- Override the live data
- Close the data stream

### Expected Behavior

Colours should become dull when applying the override, looking the same way they would if gamma correction was off. Closing the data stream should make gamma correction work again.

### Install Method

Binary from WLED.me

### What version of WLED?

0.16.0

### Which microcontroller/board are you seeing the problem on?

ESP32

### Relevant log/trace output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DedeHai:
@coderabbitai please analyze the regression.

DedeHai:
@coderabbitai how is `realtimeOverride` ended? and what is the difference between "override once" and "override until reboot" ?

softhack007:
@coderabbitai `arlsDisableGammaCorrection` is a user option for realtime streaming - it means "Disable realtime gamma correction". When users activate this option, it is expected that no gamma correction is applied for streamed pixels. 🤔 Actually I'm not sure if "no gamma" should also be applied for other segments when "Use main segment only" is selected. 

Please check if your initial analysis is still sound.

softhack007:
> ```c++
> // Before (broken):
> if (c > 0 && !(realtimeMode && arlsDisableGammaCorrection))
>     c = gamma32(c);
> 
> // After (fixed):
> if (c > 0 && !(realtimeMode && !realtimeOverride && arlsDisableGammaCorrection))
>     c = gamma32(c);
> ```

I think this fix is correct, we always use the pattern `realtimeMode && !realtimeOverride` when checking if realtime LEDs are received.
The other problem - `useMainSegmentOnly` means that only one segment does not need gamma adjustment - seems to be a heavier lift, should be adressed separately.

DedeHai:
> The other problem - `useMainSegmentOnly` means that only one segment does not need gamma adjustment - seems to be a heavier lift, should be adressed separately.

there is a good reason gamma is not applied on a segment level - they can overlap which would apply it multiple times for those pixels. Not sure that can be solved cleanly.

softhack007:
[...]

PR #5666, opened 2026-06-04
Apply gamma during realtime override

DedeHai:
partial fix for #5661

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined gamma correction application for improved color rendering consistency in real-time display modes.


softhack007 (commented):
very good, i like the idea to move the checking outside of the pixel loop 👍

DedeHai:
> very good, i like the idea to move the checking outside of the pixel loop 👍

the compiler might have optimized that already, but better be explicit.
I also wanted to remove the double-checking of `gammaCorrectCol` inside the gamma functions but then I remembered that the table is also used by brightness-gamma. Could still remove it from `Correct32(uint32_t color)` as that is only used once - and is inlined so the compiler might take care of it if it is that clever.

DedeHai:
cherry picked to 16_x

PR #5672, opened 2026-06-06
Prevent arduino core from assigning default MISO pin

### LINKED ISSUE #5670: GPIO19 not working as an 2-Wire LED output in 16.0.0
### What happened?

Hi, i have this type of led:

Image

I configured it as "APA102" on my WLED 0.15.4, and all works fine.
this is my setup:

Image

and here a picture of leds on:

Image

updating to WLED 16.0.0, leds won't power on and won't works.
I tried to fix changing some options or doing a factory reset and re-configuring it.
nothing happend.

I'm using an ESP32 board and WLED 16.0.0 starts on this board, but i don't know what to do to fix this.

### To Reproduce Bug

Use this board with APA102 leds and wled 16.0.0
https://www.tindie.com/products/giovannicas/esp32-wled-controller-2chan/
https://www.amazon.it/dp/B07C979QJD?ref_=ppx_hzsearch_conn_dt_b_fed_asin_title_1&th=1

### Expected Behavior

n/a

### Install Method

Binary from WLED.me

### What version of WLED?

WLED 16.0.0

ESP32

### Relevant log/trace output

```shell

```

### Anything else?

_No response_

### Code of Conduct

DedeHai:
Adds invalid dummy pin if miso is undefined to prevent default fallback

fixes #5670

copy of comment from [#5670](https://github.com/wled/WLED/issues/5670#issuecomment-4638253129):

>I have found a fix that works around this arduino core quirk, to not say bug. In beginDotStar() adding this line fixes it:
if (miso == -1) miso = 127;
For reference:
in IDF passing a "-1" as a pin means unused
in arduino core a "-1" as apin means "use default"
the solution is to pass an invalid pin to the arduino function, it will then try to assign it, the underlaying IDF API rejects the pin, arduino core does not even check if it fails and just move on.
The default assignment by arduino may actually have some other side-effects when using clocked LEDs in combination with default MISO pins, all ESP32 flavours are affected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue on non-ESP8266 boards where DotStar LED initialization could pick an incorrect default SPI pin, causing erratic or non-working LED output. This update ensures the SPI setup respects caller intent and prevents unintended pin reassignment during DotStar startup, improving LED reliability on affected platforms.

### DIFF
--- a/wled00/bus_wrapper.h
+++ b/wled00/bus_wrapper.h
@@ -358,6 +358,7 @@ class PolyBus {
     #ifdef ESP8266
     dotStar_strip->Begin();
     #else
+    if (miso == -1) miso = 127; // note: in arduino core, -1 meaans "default" not "none", passing 127 as the MISO pin is a workaround to prevent SPI.begin() assign the default pin, see #5670
     if (sck == -1 && mosi == -1) dotStar_strip->Begin();
     else                         dotStar_strip->Begin(sck, miso, mosi, ss);
     #endif
--- a/wled00/bus_wrapper.h
+++ b/wled00/bus_wrapper.h
@@ -358,7 +358,7 @@ class PolyBus {
     #ifdef ESP8266
     dotStar_strip->Begin();
     #else
-    if (miso == -1) miso = 127; // note: in arduino core, -1 meaans "default" not "none", passing 127 as the MISO pin is a workaround to prevent SPI.begin() assign the default pin, see #5670
+    if (miso == -1) miso = 127; // note: in arduino core, -1 means "default" not "none", passing 127 as the MISO pin is a workaround to prevent SPI.begin() assign the default pin, see #5670
     if (sck == -1 && mosi == -1) dotStar_strip->Begin();
     else                         dotStar_strip->Begin(sck, miso, mosi, ss);
     #endif


wled00/bus_wrapper.h
```diff
@@ -358,6 +358,7 @@ class PolyBus {
     #ifdef ESP8266
     dotStar_strip->Begin();
     #else
+    if (miso == -1) miso = 127; // note: in arduino core, -1 means "default" not "none", passing 127 as the MISO pin is a workaround to prevent SPI.begin() assign the default pin, see #5670
```
softhack007, line 361: @DedeHai is this workaround also working / needed for esp-idf V5 (arduino-esp32 3.5.x)?
If not, then we could wrap the line in `#if ESP_IDF_VERSION_MAJOR == 4` .. `#endif` for future compatibility.

DedeHai: from what I see, yes. the arduino.core code looks the same in the V5 branch

edit:
if it works the same I can't say - depends on what the IDF core does if you assign an out of scope pin but I would assume it wont tear down the SPI - just as it is in V4

willmmiles (approved):
Translating between our convention (-1 == unassigned) and whatever the arduino core does is the right way.

That said: this will break the setup of people who didn't think to set the pin because it turned out that they'd wired it on the default, and it just worked.  I don't know if we want to think about some kind of upgrade handoff here.  At the very least we'll need a "possibly breaking change" note on this one.

DedeHai:
> Translating between our convention (-1 == unassigned) and whatever the arduino core does is the right way.
> 
> That said: this will break the setup of people who didn't think to set the pin because it turned out that they'd wired it on the default, and it just worked. I don't know if we want to think about some kind of upgrade handoff here. At the very least we'll need a "possibly breaking change" note on this one.

I do not follow (on both counts). This is only about the MISO i.e. input pin which is unused for LED purposes. Is there any dual-use of the SPI and is that even allowed? The scenario I imagine is that the busmanager claims the SPI via NPB and after that a UM also uses that same SPI - which would initialize it anew. Or what is the scenario?

wled00/bus_wrapper.h
softhack007: @coderabbitAI please check if this workaround will still solve the problem in arduino-esp32 v3.5.x / esp-idf v5.x (see `V5` branch). Review the related framework source code, don't rely solely on official espressif documentation.

DedeHai: @softhack007 don't trust the rabbit on this one. he skipped the HAL call a few lines below its quote. what it wrote is also true for V4 but its incomplete.

DedeHai:
cherry picked to 16_x

PR #5676, opened 2026-06-10
  Fix: prevent LED flash on cold boot when a boot preset is configured

leuenbert:
# Description 
When WLED is powered by a "dumb" switch that cuts power to the whole circuit (PSU + ESP together), the LEDs briefly flash their previously-restored state at full/restored brightness during boot, before the configured boot preset (e.g. a playlist) takes over. This PR removes that flash.

# Root Cause
The boot preset is queued in WLED::beginStrip() and applied later in setup() with a transition. That transition crossfades from the current (restored) segment buffer. On cold boot the segment buffer still holds the restored colors at the restored brightness, so the crossfade's source frame is bright — that's what flashes for the first fraction of a second.

# Testing performed

  - ✅  Compiles.
  - ✅  Tested on actual hardware (ESP32 + WS2812B, single-switch PSU+ESP setup, FCOB SPI CCT WS2811 IC LED Strip) with a boot playlist. Normal cold boot (power off ≥3 s, then on) is now clean — no flash.
  - ⚠️ Known limitation: a rapid off/on within <2 s can still show a brief flash. This is a hardware effect, not firmware — the PSU's bulk capacitors keep the WS2812s powered through the dip, so the
  strip holds its last latched frame until the ESP reboots and sends a black frame. Waiting a few seconds (or adding a bleeder resistor across the 5 V rail) avoids it. Firmware sends black as early as
  it can (after config load + finalizeInit).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved boot preset initialization to properly manage LED transitions during startup. When a boot preset is applied, the LED strip now initializes to a clean black state with transitions appropriately disabled, then re-enables transitions after the preset is fully applied. This ensures smoother and more predictable LED behavior during device startup.

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -507,6 +507,7 @@ void WLED::setup()
     handlePresets();  // handle boot preset
     handlePlaylist(); // handle playlist if preset queued one
     handlePresets();  // handle presets again to give a chance for anything queued by the boot preset or playlist
+    strip.setTransition(transitionDelayDefault); // restore transitions now that the boot preset applied instantly (no boot flash)
   }
   
   if (strcmp(multiWiFi[0].clientSSID, DEFAULT_CLIENT_SSID) == 0 && !configBackupExists())
@@ -617,19 +618,28 @@ void WLED::beginStrip()
   offMode = false;   // init to on state to allow proper relay init
   handleOnOff(true); // init relay and force off
 
-  if (turnOnAtBoot) {
+  // fix for #3196 and the boot flash: if a boot preset is configured, fully blank the strip and
+  // push a black frame before applying it. The boot preset is applied with a transition, which
+  // crossfades FROM the current (restored) segment buffer; without a black starting frame that
+  // restored state flashes at the restored brightness for the first fraction of a second.
+  // This must run for turnOnAtBoot too. Brightness 0 alone (the boot brightness still feeds the
+  // crossfade's source) and blanking colors[0] alone (palette/FX states still render) are each
+  // insufficient on their own, which is why earlier attempts still flashed.
+  if (bootPreset > 0) {
+    // set all segments black (no transition)
+    for (unsigned i = 0; i < strip.getSegmentsNum(); i++) {
+      Segment &seg = strip.getSegment(i);
+      if (seg.isActive()) seg.colors[0] = BLACK;
+    }
+    colPri[0] = colPri[1] = colPri[2] = colPri[3] = 0;  // needed for colorUpdated()
+    briLast = briS; bri = 0;       // stay dark; the boot preset sets its own brightness when it applies
+    strip.fill(BLACK);
+    if (rlyPin < 0)
+      strip.show();                // push a black frame so the boot preset's transition starts from black
+  } else if (turnOnAtBoot) {
     if (briS > 0) bri = briS;
     else if (bri == 0) bri = 128;
   } else {
-    // fix for #3196
-    if (bootPreset > 0) {
-      // set all segments black (no transition)
-      for (unsigned i = 0; i < strip.getSegmentsNum(); i++) {
-        Segment &seg = strip.getSegment(i);
-        if (seg.isActive()) seg.colors[0] = BLACK;
-      }
-      colPri[0] = colPri[1] = colPri[2] = colPri[3] = 0;  // needed for colorUpdated()
-    }
     briLast = briS; bri = 0;
     strip.fill(BLACK);
     if (rlyPin < 0)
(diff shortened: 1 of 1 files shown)


softhack007:
> That transition crossfades from the current (restored) segment buffer. On cold boot the segment buffer still holds the restored colors at the restored brightness,

@leuenbert what kind of segment buffer do you mean? 
When WLED is power-cut, the MCU is off. Software is off,  RAM is off. There is no memory buffer that will "hold restored colors" at the next boot.

softhack007:
I think this PR is largely "snake oil".
@coderabbitai full review.

DedeHai:
This can be fixed, in fact I did this but never PR'd. Similar but the code here is only partially correct

leuenbert:
> > That transition crossfades from the current (restored) segment buffer. On cold boot the segment buffer still holds the restored colors at the restored brightness,
> 
> @leuenbert what kind of segment buffer do you mean? When WLED is power-cut, the MCU is off. Software is off, RAM is off. There is no memory buffer that will "hold restored colors" at the next boot.

Basically what i've tested is using a "dumb switch" so PSU and MCU are connected to an Outlet that is controlled by a normal light switch.

In the current version when turning the whole thing off and on, and you have a Boot Playlist/Transition it will: flash -> turn black -> play transition/playlist

But when you turn the brightness to 0/color to BLACK right before switching off the whole thing with the "dumb switch", wait 3s and switch it on again the flash doesn't happen

Maybe my wording was wrong, I didn't mean a buffer that persists through the power loss.

I mean the buffer after WLED reloads the saved brightness and colors from flash during boot. The boot preset is applied with a transition, so the crossfade starts from whatever is already in that buffer. That's why the strip lights up bright for a split second before the preset kicks in.

The fix forces a black starting frame (and applies the boot preset with transitions off) so the crossfade has nothing bright to start from.

As @DedeHai stated, there might be a prettier solution for that though

willmmiles:
I believe the AI analysis is wrong.  @coderabbitai, can you point to where in the code it "restores flash state to segs"?  To the best of my knowledge this is not a thing -- the initial segment state is determined by only three config inputs:
- "turn on at boot"
- "create one segment per output"
- the global brightness setting, if set in the config

The initial segments are always initialized to mode solid, either orange or black, depending on the state of "turn on at at 
boot".  `color[1]` and `color[2]` are always initialized to black by the Segment constructor.

---

Applying the boot preset should execute the transition configured in the boot preset, or the default transition from the config. 

If "turn on at boot" is disabled, then the correct behaviour is to execute the boot preset with transition from an initial state of all segments black.

If "turn on at boot" is set, then the current behaviour is for the boot preset to transition from the default initial state (ie. one segment at strip length set to solid orange).  It's not very useful but it's the natural consequence of the interaction of those features.  I could accept an argument that we should ignore the "turn on at boot" flag if a boot preset is set, but it would be a breaking change.  I believe it's been discussed before.

It is definitely not correct to disable the transition to the boot preset.  (The fade-in behaviour by default is quite nice IMO.) 

As far as I can find, this is what the current code does: the first frame written to the LEDs is always either full black or full yellow, depending on the "turn on at boot" setting.  At no point in the firmware is any color data remembered from a previous boot, and the transition correctly proceeds from the initial segment setting.  All pixel buffers are fully initialized with a clean frame render before show() is called.

---

I haven't been able to reproduce the issue.  I always get exactly as described above, an initial frame of either orange or black, followed by a transition to the boot preset, as expected.  On my test bench I'm using a switch in the USB 5V line for the micro and a WS2812-compatible LED strip.

I think the implication is that the issue is hardware-ish -- I'm aware some LEDs can hold their control state for a period of time after power-down; and we've already seen that 16.0 has a different power-on timing than previous versions (cf. all the newly reported issues with poor power supply regulation needing an additional capacitor).  I note that one of the changes in this patch is that it moves the first `show()` to earlier in the setup path when a boot preset is configured and "turn on at boot" is set; I am suspecting that's the functional change.  We might want to consider an "early LED output frame" as a mitigation for hardware weirdnessess at power on, though I fear that will have even further knock-on effects on power supply regulation.

@leuenbert Have you already tried disabling "turn on at boot"?  Did that change the behaviour?

blazoncek:
Why don't you just keep `turnOnAtBoot` off and let boot preset turn LEDs on?

That was the intended behavior as not all boot presets turn WLED on. I.e. just set up segments.

However, if you want to keep `turnOnAtBoot` on and prevent *transition* from default orange color to a boot preset defined one there is a much simpler way to do it.

leuenbert:
> @leuenbert Have you already tried disabling "turn on at boot"? Did that change the behaviour?

I've just tested it again with 16.0.0 and it works. Sorry for the confusion i'll close the MR. Nevertheless thanks for having a look at it.

PR #5683, opened 2026-06-16
Fix for ESP32C3 Glitching on Long Strips

bigfella5000:
### Hardware
- Seeed XIAO ESP32-C3
- 833x WS2815 ECO LEDs

### Fix
Made wait time before calling getContiguousFreeHeap() on ESP32C3 to be a function of the number of LEDs to avoid interrupting RMT transmissions, which corrupts end of frame during animations every 5 seconds. At 800kHz with 24 bits per LED, a full frame takes N × 30µs. For strips longer than ~500 LEDs the fixed 15ms wait is insufficient, causing getContiguousFreeHeap() to interrupt an in-progress RMT transmission and corrupt the last portion of the frame (visible as a white flash every 5 seconds). This PR replaces the hardcoded value with a calculation based on actual strip length with a safety margin, floored at the original 15ms so short strips are unaffected.

### Ongoing Issue
When running the effect given in "Wavesins API Command," my solution stopped the glitching. And in all other effects I tested, the issue is fixed. However, when running the effect given in "Stream API Command," the issue is still persisting. Additionally, the glitching occurs not just in the last quarter of the frame as it did with the other effects, but instead on the whole frame. And instead of glitching every 5 seconds, it's glitching every 6 seconds. Even when I raised the margin in the waitMS calculation, the glitch still occurred the exact same way on the Stream effect.

**Wavesins API Command**
`{"on":true,"bri":255,"transition":0,"bs":0,"mainseg":0,"seg":[{"id":0,"start":0,"stop":832,"grp":1,"spc":0,"of":0,"on":true,"frz":false,"bri":255,"cct":127,"set":0,"lc":1,"n":"","col":[[115,0,255],[255,4,0],[0,0,0]],"fx":184,"sx":15,"ix":245,"pal":18,"c1":124,"c2":34,"c3":1,"sel":true,"rev":false,"mi":false,"o1":false,"o2":false,"o3":false,"si":0,"m12":0,"bm":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0}]}`

**Stream API Command**
`{"on":true,"bri":255,"transition":0,"bs":0,"mainseg":0,"seg":[{"id":0,"start":0,"stop":832,"grp":1,"spc":0,"of":0,"on":true,"frz":false,"bri":255,"cct":127,"set":0,"lc":1,"n":"","col":[[115,0,255],[255,4,0],[0,0,0]],"fx":39,"sx":255,"ix":39,"pal":48,"c1":128,"c2":128,"c3":16,"sel":true,"rev":false,"mi":false,"o1":false,"o2":false,"o3":false,"si":0,"m12":0,"bm":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0},{"stop":0}]}`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved responsiveness on ESP32C3 devices by optimizing memory health monitoring to prevent blocking during LED strip updates. The system defers heap checks when the strip is actively updating and resumes normal monitoring intervals when complete. All memory safety checks remain intact.

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -178,7 +178,10 @@ void WLED::loop()
     // calling getContiguousFreeHeap() during led update causes glitches on C3
     // this can (probably) be removed once RMT driver for C3 is fixed
     unsigned t0 = millis();
-    while (strip.isUpdating() && (millis() - t0 < 15)) delay(1);    // be nice, but not too nice. Waits up to 15ms
+    // Most strips run at 800kHz, each LED needs 24 bits = 30µs per LED
+    // Add 30% margin and convert to ms, minimum 15ms
+    uint32_t waitMs = max(15u, (unsigned)(strip.getLengthTotal() * 30 / 1000 * 1.3));
+    while (strip.isUpdating() && (millis() - t0 < waitMs)) delay(1);    // be nice, but not too nice
     #endif
     uint32_t heap = getContiguousFreeHeap(); // ESP32 family needs ~10k of contiguous free heap for UI to work properly
     #endif
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -40,7 +40,7 @@ void WLED::reset()
 
 void WLED::loop()
 {
-  static uint16_t      heapTime = 0;   // timestamp for heap check
+  static uint16_t      heapTime = millis();   // timestamp for heap check
   static uint8_t       heapDanger = 0; // counter for consecutive low-heap readings
 #ifdef WLED_DEBUG
   static unsigned long lastRun = 0;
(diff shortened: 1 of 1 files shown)


wled00/wled.cpp
```diff
@@ -178,7 +178,10 @@ void WLED::loop()
     // calling getContiguousFreeHeap() during led update causes glitches on C3
     // this can (probably) be removed once RMT driver for C3 is fixed
     unsigned t0 = millis();
-    while (strip.isUpdating() && (millis() - t0 < 15)) delay(1);    // be nice, but not too nice. Waits up to 15ms
+    // Most strips run at 800kHz, each LED needs 24 bits = 30µs per LED
+    // Add 30% margin and convert to ms, minimum 15ms
+    uint32_t waitMs = max(15u, (unsigned)(strip.getLengthTotal() * 30 / 1000 * 1.3));
```
DedeHai: in general has a point BUT:
- don't use floats
- only calculate if we actually need to wait
- getLengthTotal() is not the length of the longest strip but total _configred_ LEDs, including gaps

in general, waiting on a single core MCU is bad. the proper way to do it is to defer this to a loop when LEDs are not updating.

edit:
even the original wait is a hack and a workaround for unfixable deeper bugs.

softhack007:
@bigfella5000 I think you're only scratching on the surface - the real problem on -C3 is the shared "RMT refill" interrupt. This interrupt can be stalled by many FreeRTOS or hardware driver activities, like file access or calculating unused memory.

You can have a look at this PR for a more complete picture
* #5435

The wled dev team is aware of the problem, and the only real solution for -C3 is to rewrite the leds driver, or implement a high-priority RMT driver (similar to #4890) for RISC-V chips. Both options are ongoing experiments, but the code is not stable yet - @DedeHai and @willmmiles might be able to tell more.

Your idea to make the wait time depend on number of leds is interesting, and could be combined with #5435 for a more robust solution. However I see a fundamental flaw: `strip.getLengthTotal()` does not relate to the number of leds driven by a single RMT pin. On esp32, I can have 4096 leds distributed over 8 pins (just one example) - in this example each RMT output drives 512 leds, not 4096.

Edit: with 4096 LEDs, your patch would calculate a wait time of ~160ms, which is totally unacceptable as a delay that happens each 5 seconds.

wled00/wled.cpp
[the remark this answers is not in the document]
softhack007: @coderabbitai how is this expression evaluated according to C semantics? Does it include intermediate results in double (1000 * 1.3 = 1300.0) ?

wled00/wled.cpp
```diff
@@ -178,7 +178,10 @@ void WLED::loop()
     // calling getContiguousFreeHeap() during led update causes glitches on C3
     // this can (probably) be removed once RMT driver for C3 is fixed
     unsigned t0 = millis();
-    while (strip.isUpdating() && (millis() - t0 < 15)) delay(1);    // be nice, but not too nice. Waits up to 15ms
+    // Most strips run at 800kHz, each LED needs 24 bits = 30µs per LED
```
softhack007: This assumption is not true for 2-wire "dotstar" (SPI based) and other "not clockless" LEDs. Additionally, there are 4-color (RGBW) and 5-color(RGB-CCT) strips where "each led needs 24 bits" is incorrect.

softhack007: @coderabbitai please remember that some "clockless" led strips use 4, 5,or even 6 bytes per pixel. Assuming 800Khz 24bit per pixel is only correct for RGB LEDs strips that use the ws281x driver at 800khz, see wled00/bus_wrapper.h for details.

softhack007: Additionally, WLED supports APA106, TM18xx, UCS890x, and "400Kbps" led strips that run on different base speed, like 400Khz or 600Khz.

bigfella5000:
Thanks for pointing to #5435. The 0_15_x build solved the issue on all effects I tested including the Stream effect. One issue I'm having with it, which is likely well known, is that when the esp32c3 boots after uploading the 0_15_x build, I'm still seeing some glitching for the first few seconds. I'm assuming this is because there's plenty of other tasks interfering on boot. One other potential solution for avoiding the interference between the heap check and led updates would be the following:

```cpp
#ifdef CONFIG_IDF_TARGET_ESP32C3
// calling getContiguousFreeHeap() during led update causes glitches on C3
// this can (probably) be removed once RMT driver for C3 is fixed
// If strip updating, defer heap check until future
if (strip.isUpdating()) {
  heapTime = millis() - 4999;
} else
#endif
```
```cpp
{
  uint32_t heap = getContiguousFreeHeap(); // ESP32 family needs ~10k of contiguous free heap for UI to work properly
  #endif
  if (heap < MIN_HEAP_SIZE - 1024) heapDanger+=5; // allow 1k of "wiggle room" for things that do not respect min heap limits
  else heapDanger = 0;
  switch (heapDanger) {
    case 15: // 15 consecutive seconds
      DEBUG_PRINTLN(F("Heap low, purging segments"));
      strip.purgeSegments();
      strip.setTransition(0); // disable transitions
      for (unsigned i = 0; i < strip.getSegmentsNum(); i++) {
        strip.getSegments()[i].setMode(FX_MODE_STATIC); // set static mode to free effect memory
      }
      errorFlag = ERR_NORAM; // alert UI  TODO: make this a distinct error: segment reset
      break;
    case 30: // 30 consecutive seconds
      DEBUG_PRINTLN(F("Heap low, reset segments"));
      strip.resetSegments(); // remove all but one segments from memory
      errorFlag = ERR_NORAM; // alert UI  TODO: make this a distinct error: segment reset
      break;
    case 45: // 45 consecutive seconds
      DEBUG_PRINTF_P(PSTR("Heap panic! Reset strip, reset connection\n"));
      strip.~WS2812FX();      // deallocate strip and all its memory
      new(&strip) WS2812FX(); // re-create strip object, respecting current memory limits
      if (!Update.isRunning()) forceReconnect = true; // in case wifi is broken, make sure UI comes back, set disableForceReconnect = true to avert
      errorFlag = ERR_NORAM; // alert UI  TODO: make this a distinct error: strip reset
      break;
    default:
      break;
  }
  heapTime = (uint16_t)millis();
}
```

However, the 0_15_x handles the interference more generally and in more locations by using waitForLEDs(), so I'm assuming that's a better solution.

wled00/wled.cpp
```diff
@@ -177,39 +177,43 @@ void WLED::loop()
     #ifdef CONFIG_IDF_TARGET_ESP32C3
     // calling getContiguousFreeHeap() during led update causes glitches on C3
     // this can (probably) be removed once RMT driver for C3 is fixed
-    unsigned t0 = millis();
-    while (strip.isUpdating() && (millis() - t0 < 15)) delay(1);    // be nice, but not too nice. Waits up to 15ms
```
softhack007, line 181: Second thought - as 15ms is just the upper limit for the waiting time - it would be sufficient here to change `15` into `50`, and drop all the other changes.

wled00/wled.cpp
```diff
@@ -177,39 +177,43 @@ void WLED::loop()
...
+    // If strip updating, defer heap check until future
+    if (strip.isUpdating()) {
+      heapTime = millis() - 4999;
```
softhack007, line 182: Not a good idea - this basicially postpones the heap check to infinite future, especially when target FPS is high.

softhack007:
Screenshot_20260617-091606_Samsung Internet

@bigfella5000  For any future PR you may do - please don't `force-push`, or add a comment that explains why you had to `push --force-with-lease`.

See https://github.com/wled/WLED?tab=contributing-ov-file#updating-your-code

wled00/wled.cpp
DedeHai: that is what I thought too, however, it may be better to give the main loop several tries to get a slot where the strip is idle. Otherwise, this can result in unnecessary FPS hickups every 5s. I don't think there is a universal solution though, if target FPS is higher than actual FPS the main loop may not/never be running at the right moment.

softhack007: I agree. A limited number of retries - maybe an extra 2seconds when a new measurement is due - could reduce hickups. If the main loops runs strip.show() in each iteration (high FPS), this will not help much. But users could still reduce target FPS to prevent glitches.

PR #5686, opened 2026-06-17
improve bootup behaviour for boot presets

DedeHai:
changes made:

- full refactor of the bootup behaviour regarding "bootup preset" and "turn on at boot" here is how it works now:
  - if turn on at boot is set, LEDs turn on unconditionally to the set default brightness
  - if no prest is applied, it still fades from black at the set default transition time instead of instantly turning on (same as it was in 0.14)
  - if a preset is applied it fades to that preset on all segments, starting off black
  - if any transition mode other than fade is set, LEDs start off at "default brightness" instead of fading from black if "turn on at boot" is set (not acutally sure why on "fade" transition it starts off black but it works)

- Segments created by a bootup presets no longer use the DEFAULT_COLOR but start out black: segments created within the first 5s after bootup default to black
- DEFAULT_COLOR used for new segments is now the same as the bootup color (AA which is 170 changed to A0 which is 160)

TLDR; 
no matter if "turn on at boot" is now set or not, boot up presets are applied correctly. If "turn on at boot" is set, LEDs always turn on.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved boot behavior so LEDs reliably begin from black, including when no relay is configured.
  * Streamlined startup brightness and boot preset initialization for more consistent first-frame behavior.
* **UI/UX**
  * Updated the bootup brightness field label to “Bootup brightness”.
* **Behavior / Visual Defaults**
  * Adjusted the default warm orange shade and ensured newly created auto segments and certain default segments start with the expected primary color.
* **Style**
  * Changed the initial stored primary color state to start from black.

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -617,26 +617,36 @@ void WLED::beginStrip()
   offMode = false;   // init to on state to allow proper relay init
   handleOnOff(true); // init relay and force off
   if (rlyPin < 0) strip.show(); // ensure LEDs are off if no relay is used
+  // note on how bootup behaviour:
+  // if turnOnAtBoot is false, strip is set to black. It will fade in to startup brightness and orange when turned on
+  // if a bootup preset is set, it will fade to that preset if it has on:true set (to default brightness) or to that preset's brightness if set
+  // if turnOnAtBoot is true the LEDs will fade in to orange and default brightness
+  // if a booup preset is set, it will start at the default brightness except if "fade" transition is used, then it will still fade from black
+  // there is no way to have LEDs off at boot and upon turn-on have them immediatel jump to a target brightness but users can use a playlist to do that
+
   bri = 0; // start off black by default (on a fresh install this is overruled by briS as turnOnAtBoot is true)
   if (turnOnAtBoot) {
-    bri = briS ? briS : 128; // load startup brightness (set in UI), fall back to mid if set to 0 to make sure LEDs turn on
-    // set color to warm welcoming orange (aka DEFAULT_COLOR)
-    colPri[0] = 255;
-    colPri[1] = 160;
-    colPri[2] = colPri[3] = 0;
+    bri = briS; // load startup brightness (set in UI), 0 is not allowed in UI
   }
+  else briLast = briS; // go to startup brightness (set in UI) when turning on (can be overruled by a preset)
 
   if (bootPreset > 0) {
     colPri[0] = colPri[1] = colPri[2] = colPri[3] = 0;  // needed for colorUpdated()
     applyPreset(bootPreset, CALL_MODE_INIT);
-    // set all segments black (no transition), their default creation color is orange (which is needed for creating new segments in the UI)
+    // note: if turn on at boot is disabled, preset will fade in from black, even if swipe is used
+    // if fade is used, it will always fade from black.
     for (unsigned i = 0; i < strip.getSegmentsNum(); i++) {
       Segment &seg = strip.getSegment(i);
       if (seg.isActive()) seg.colors[0] = BLACK;
     }
   }
-  else
+  else {
+    // set color to warm welcoming orange (aka DEFAULT_COLOR) if no preset loaded (will fade to this color once turned on)
+    colPri[0] = 255;
+    colPri[1] = 160;
+    colPri[2] = colPri[3] = 0;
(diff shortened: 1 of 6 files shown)


DedeHai:
@coderabbitai please check for any edge cases that may have a changed behaviour and come up with a test plan to verify this PR works as intended.

DedeHai:
I tested all the suggested cases and made some improvements. this is how it works now, copy from source code comment for reference:

- if turnOnAtBoot is false, strip is set to black. It will fade in to startup brightness and orange when turned on
- if a bootup preset is set, it will fade to that preset if it has on:true set (to default brightness) or to that preset's brightness if set
- if turnOnAtBoot is true the LEDs will fade in to orange and default brightness
- if a booup preset is set, it will start at the default brightness except if "fade" transition is used, then it will still fade from black
- there is no way to have LEDs off at boot and upon turn-on have them immediatel jump to a target brightness but users can use a playlist to do that

DedeHai:
@coderabbitai full review, take the test plan above into account and check if there is inconsistencies for edge cases.

DedeHai:
note: the rabbit is wrong about most things except the default transition time.

DedeHai:
TODO: remove the 2s boot timeout and move the orange color to set.cpp

DedeHai:
@coderabbitai if I change the segment constructor to use all black i.e. `colors{BLACK,BLACK,BLACK}` but still want newly created segments that are created from UI and not a preset to be orange, what is the best way to set colors[0] = DEFAULT_COLOR in that newly created segment? I think this also might apply to to `makeAutoSegments()`.

wled00/wled.cpp
```diff
@@ -616,31 +616,34 @@ void WLED::beginStrip()
...
+  colorUpdated(CALL_MODE_INIT); // set bootup brightness immediately, do not send notification (brightness is also set for preset if used, useful for swipe etc.)
+
   if (bootPreset > 0) {
     applyPreset(bootPreset, CALL_MODE_INIT);
   }
+  else {
+    // set color to warm welcoming orange (aka DEFAULT_COLOR) if no preset loaded (will fade to this color once turned on)
+    colPri[0] = 255;
+    colPri[1] = 160;
+    colPri[2] = colPri[3] = 0;
```
willmmiles: Should this be `DEFAULT_COLOR`? I know it wasn't previously, but it feels like it ought to be.

DedeHai: Yes, correct. Need a pointer cast for that and I guess I was just lazy to convert it

DedeHai: actually, the array is not defined the same way, so cant do pointer cast. Should replace that array with a CRGBW color but that is a cleanup for another day.

DedeHai:
good to merge from my side. @willmmiles you mentioned breaking changes but I can't really think of anything bad...

willmmiles:
> good to merge from my side. @willmmiles you mentioned breaking changes but I can't really think of anything bad...

Sorry I didn't get to testing it!  The potential breaking case was the "on + boot preset" where the boot preset did not set the effect state.  (Maybe they're just configuring some usermod state - who knows.)   With the new logic, the outputs will end up stuck as black.

willmmiles:
> > good to merge from my side. @willmmiles you mentioned breaking changes but I can't really think of anything bad...
> 
> Sorry I didn't get to testing it! The potential breaking case was the "on + boot preset" where the boot preset did not set the effect state. (Maybe they're just configuring some usermod state - who knows.) With the new logic, the outputs will end up stuck as black.

To be completely clear: I don't think there's a coherent way to handle "turn on at boot" in the general case where a preset can do (or not do) anything.  Making up some heuristic to guess at how to harmonize that setting with arbitrary preset code is a recipe for disaster.

So, with the new approach, we are switching to a paradigm of "assigning a boot preset overrides the 'turn on LEDS on at boot' setting", which is the correct behaviour for 99.9% of real world use cases.  I think this is the right call, but I feel obligated to point out that it is technically a behaviour change, and that 0.01% user with some crazy preset logic may need to fix their boot preset after updating.

shoelzer:
I just got the nightly specifically for this change. I use a switched outlet to keep my lights + controller fully powered off when not in use, and it's always bothered me that they would flash orange in a dark room when turned on. Now they fade in from black. So nice!

PR #5687, opened 2026-06-17
bugfix in segment deconstructor

### LINKED ISSUE #4304: Deleting segments is not handled properly
### What happened?

When having multiple segments, deleting one of them results in broken segment configuration with one exception: if it is the last segment that is deleted.

### To Reproduce Bug

I tested like this on a 16x16 matrix, setting-up 3 overlapping segments:
- Segment 0: initial segment is 16x16
- Segment 1: overlapping segment 8x16 (start at x=0) 
- Segment 2: overlapping segment 8x16 (start at x=8) 

setup any FX you like, I used "Black Hole" on all three segments
delete the middle segment -> the last segment gets broken and has to be deleted before it can work again

### Expected Behavior

deleting a segment should not mess up other segments, it should be properly handled.

### Install Method

Binary from WLED.me

### What version of WLED?

Not working on 14.4 and 15 B7

ESP32-C3

### Relevant log/trace output

_No response_

### Anything else?

_No response_

### Code of Conduct

DedeHai:
fix memory leak as discussed in #4304

fixes #4304

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of segment transitions by ensuring any active transition is properly terminated when segments are removed, reassigned, moved, or their geometry changes—helping prevent lingering transition state, resource issues, and inconsistent behavior during dynamic updates.

### DIFF
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -618,6 +618,7 @@ class Segment {
       DEBUGFX_PRINTLN();
       #endif
       clearName();
+      stopTransition();   // deallocate "_t" (transition) and with it "_segOld" note: _segOld has _t=null, see copy constructor
       #ifdef WLED_ENABLE_GIF
       endImagePlayback(this);
       #endif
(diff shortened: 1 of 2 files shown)


softhack007 (approved):
looks good.

willmmiles (approved):
LGTM

(though I wish we could be using smart pointers instead.  Every `new` and `delete` is a mistake waiting to happen.)

DedeHai:
cherry picked to 16_x

PR #5690, opened 2026-06-18
Minimal C3 bitbanging fix

willmmiles:
This is the simplest, lamest solution to C3 glitches: replace it with the bitbanging driver.  I am presenting this here as a draft for reference if we decide we need an immediate solution, no matter the performance cost.  This is not a good solution by any means - merely a working one on my test system.

Sadly it turned out that NPB's bitbanging driver also wasn't completely stable on RISCV systems, I am guessing due to cache issues returning from ISR handling.   Lowering the latch time estimates and prefetching the pixel outputs on to the stack seemed to help on my test system.  (Probably I should've added a "ISR return fudge factor" instead of permuting the timings directly; that could be done if there's interest in pursing this.

## Summary by CodeRabbit

* **Improvements**
  * Added dedicated support for ESP32-C3 boards with improved addressable LED driver compatibility, ensuring proper control and reliability on this microcontroller variant.

* **Chores**
  * Updated NeoPixel library dependency to the latest version.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -130,7 +130,7 @@ upload_speed = 115200
 lib_compat_mode = strict
 lib_deps =
     IRremoteESP8266 @ 2.8.2
-    https://github.com/Makuna/NeoPixelBus.git#a0919d1c10696614625978dd6fb750a1317a14ce
+    https://github.com/willmmiles/NeoPixelBus.git#71112f0627e1db397f1770881a4cffc6c067a420
     https://github.com/Aircoookie/ESPAsyncWebServer.git#v2.4.2
     marvinroger/AsyncMqttClient @ 0.9.0
   # for I2C interface
--- a/wled00/bus_wrapper.h
+++ b/wled00/bus_wrapper.h
@@ -248,6 +248,15 @@
 #if !defined(WLED_USE_SHARED_RMT)  && !defined(__riscv)
 #include <NeoEsp32RmtHIMethod.h>
 #define NeoEsp32RmtMethod(x) NeoEsp32RmtHIN ## x ## Method
+#elif defined(CONFIG_IDF_TARGET_ESP32C3)
+// We're going to cheat and use bitbanging
+// Wrapper type strips the 'channel' argument
+template<typename T> class NPBMethodStripChannelWrapper : public T {
+public:
+    NPBMethodStripChannelWrapper(uint8_t pin, uint16_t pixelCount, size_t elementSize, size_t settingsSize, NeoBusChannel channel) : T(pin, pixelCount, elementSize, settingsSize) {
+    };
+};
+#define NeoEsp32RmtMethod(x) NPBMethodStripChannelWrapper<NeoEsp32BitBang ## x ## Method>
 #else
 #define NeoEsp32RmtMethod(x) NeoEsp32RmtN ## x ## Method
 #endif


DedeHai:
as per my tests: to fully fix glitching on C3, do not allow any ISRs during BB from start to finish.

edit: 
for reference:
https://github.com/DedeHai/WLED/blob/5860b73b09f273fa4b7bd5475b675a2339611f61/wled00/src/WLEDpixelBus/WLEDpixelBus_BitBang.cpp#L331

DedeHai:
@willmmiles one thing I recently thought of but did not check yet:
on the S2 we allow I2S output even though it clashes with AR. We could do the same for C3, maybe there is a minimal check that can be done through an IDF call to check if the I2S is initialized - although beginStrip() is called before UM-setup, so this check would need to be done in AR usermod... aaand we are back to a hardware manager ;)

willmmiles:
> as per my tests: to fully fix glitching on C3, do not allow any ISRs during BB from start to finish.

I don't think that's acceptable in the general case.  It'll risk watchdog failures and other panics from the wifi system when under load.  As far as I'm aware, ESP-IDF can only reliably handle interrupt gaps of a few ms.

The NPB patch here doesn't include latch time corrections for all LED types - I only tuned the one on my test bench as a proof-of-concept.  If you tested this patch, what flavour were you using?

DedeHai:
> I don't think that's acceptable in the general case

what would be the worst case scenario to test if its viable? I did test with 1000 LEDs so ~30ms blocking and saw no issue with that, heavy UI reload and AR audio incoming stream. We could put a 512 LED limit on BB outputs.

willmmiles:
> @willmmiles one thing I recently thought of but did not check yet: on the S2 we allow I2S output even though it clashes with AR. We could do the same for C3, maybe there is a minimal check that can be done through an IDF call to check if the I2S is initialized - although beginStrip() is called before UM-setup, so this check would need to be done in AR usermod... aaand we are back to a hardware manager ;)

So we are!

I'm posting this for posterity as a starting place if we decide we need a quick drop-in workaround for 16.x.   For 17.0 I think we should merge your new driver framework with the expanded hardware options including parallel bitbanging, SPI, and the like, and offer direct selection of driver and options via the JSON config.

willmmiles:
> > I don't think that's acceptable in the general case
> 
> what would be the worst case scenario to test if its viable? I did test with 1000 LEDs so ~30ms blocking and saw no issue with that, heavy UI reload and AR audio incoming stream. We could put a 512 LED limit on BB outputs.

That's interesting - I was under the impression the IWDT default was 20ms.  Are you explicitly disabling it?

I do agree we'll want to put a cap on bitbanging lengths regardless.

willmmiles:
> what would be the worst case scenario to test if its viable?

`autocannon` targeting `/presets.json` hits both bandwidth and flash IO.  Another thing to check is enabling wifi sleep, and then triggering a lot of broadcast traffic.  I'm not sure what happens if wifi sleep is enabled and you miss the collection window.

DedeHai:
>I'm posting this for posterity as a starting place if we decide we need a quick drop-in workaround for 16.x. 

we can definitely add it as an option, in my WLEDbus branch I alread added it to the UI selection menu (well, claude did, I only glanced over that part and it looked legit and it works). In 16.0 ESP32, S2, S3 already offer that selection menu (RMT / I2S) so adding BB to it and restricting to 512 or 1024 LEDs if selected should be doable, although it may be a bit of a painful chore.

willmmiles:
> > I'm posting this for posterity as a starting place if we decide we need a quick drop-in workaround for 16.x.
> 
> we can definitely add it as an option, in my WLEDbus branch I alread added it to the UI selection menu (well, claude did, I only glanced over that part and it looked legit and it works). In 16.0 ESP32, S2, S3 already offer that selection menu (RMT / I2S) so adding BB to it and restricting to 512 or 1024 LEDs if selected should be doable, although it may be a bit of a painful chore.

Don't spend too much time on the integration -- it's still my intention to replace the bus configuration backend with a proper JSON configuration structure which allows unique properties for each hardware interface.  (I still really, really hate the existing solution of mapping bus hardware types to integers.  It leaves us stuck with either a global mapping for all hardware for all platforms -- so anything new requires a core change -- or worse, a unique string mapping for each platform in the UI.  Hardware type selection needs to be strings.)

DedeHai:
I tested the absolute worst case: 2048 LEDs, 400kHz bitbang output. this blocks for 125ms from enter to exit critical. It somehow still works just fine. UI is quite responsive albeit not blazing fast. DDP streaming from VideoLab through websocket does not work anymore though, I do get one update every 10 seconds tops so something gets congested - but no crashes. FPS drop to 7 but that is expected. Does not matter if I have wifi sleep enabled or disabled, its about the same. 

edit:
sync sendout also works from the BB unit. Sync receive on the BB uint is sluggish, it works if I set the sender to repeat 3 times but there is a significant delay of up to 1s.

softhack007:
> > as per my tests: to fully fix glitching on C3, do not allow any ISRs during BB from start to finish.
> 
> I don't think that's acceptable in the general case. It'll risk watchdog failures and other panics from the wifi system when under load. As far as I'm aware, ESP-IDF can only reliably handle interrupt gaps of a few ms.

@DedeHai it means you were able to "hold" a critical section up for more than 100ms? Wow, that's quite long.

And it's much more than what I would have guessed from reading the FreeRTOS docs https://m.freertos.org/Documentation/02-Kernel/04-API-references/04-RTOS-kernel-control/01-taskENTER_CRITICAL_taskEXIT_CRITICAL

> Critical sections must be kept very short, otherwise they will adversely affect interrupt response times. Every call to taskENTER_CRITICAL() must be closely paired with a call to taskEXIT_CRITICAL().
> FreeRTOS API functions must not be called from within a critical section. taskENTER_CRITICAL() and taskEXIT_CRITICAL() must not be called from an interrupt service routine (ISR).

DedeHai:
> it means you were able to "hold" a critical section up for more than 100ms? Wow, that's quite long.

I have no idea how the ESP framework handles watchdogs exactly, never looked into that. But from my debug sessions where an ISR gets stuck it looks like it triggers after a very long time, like 1 or 2 seconds. What may be an issue with keeping interrupts from firing may be IR receivers, don't have one so could not test.

willmmiles:
I did a bit of digging and found out that the default interrupt watchdogs are absurdly long (300ms)!

Yes, interrupt latencies on the >1 ms scale will pretty much kill IR reception, I think - the library we're currently using depends on software timing.   Tradeoffs, tradeoffs, everywhere...

softhack007:
(a bit off-topic) I'm curious if the rabbit can dig up some constraints 🤔 

@coderabbitai please check if the official documentation for esp-idf v4.4.x and arduino-esp32 v2.0.x has some statements / hints on how long a critical section (taskENTER_CRITICAL()) can be held open before the system becomes unreliable.

Also check the esp-idf and arduino-esp32 source code for additional information - assume a standard arduino-esp32 configuration of esp-idf, especially a task scheduler period of 1ms.

Would flash filesystem access, wifi, or PSRAM access be degraded when the looptask holds a critical section open for 100ms?

DedeHai:
IR reception is a valid point and on the ESP8266 there is not RMT receiver. I will update my driver to allow interrupts during BB output again, maybe gate it by a bool so it can be disabled if no IR is being used.
BTW I just had an idea: this "missed time window" gating could also be used for RMT output: in the translator, check CPU cycles between translations, if the window was missed (which may be a bit hard to threshold, maybe 1.5x the time it take to send out half a memory block size?) I would just detach the GPIO, set it low and let the RMT state machine play out, then re-attach the pin to RMT when the next frame is sent out. I do the same thing in the parallel SPI if a buffer underrun is detected and it successfully prevents white flashes that can happen otherwise.

DedeHai:
I had a go at this. I have a strip that flickers when using BB even if I set the "latch watch period" as low as 15µs, it probably latches below 10µs, did not test the limit. Interestingly it does flicker only for a few seconds and then stops so the IC seems to have some auto-adjust circuit for latch period, it would start to flicker again immediately if I unplug then reconnect it to the running controller.

Having such a short "latch window" means virtually any interrupt may halt the BB output. This can get critical on longer strips as the end of it may "starve" the end of the strip. My test with 512 LEDs and WS video streaming show significant drop in FPS towards the end of the strip, it was quite choppy, but no flickering. So "do not allow interrupts" should be a config option for people who want to use BB outputs and do not need IR or other time critical interrupts.

willmmiles:
> I had a go at this. I have a strip that flickers when using BB even if I set the "latch watch period" as low as 15µs, it probably latches below 10µs, did not test the limit. Interestingly it does flicker only for a few seconds and then stops so the IC seems to have some auto-adjust circuit for latch period, it would start to flicker again immediately if I unplug then reconnect it to the running controller.

I found my test strip definitely seemed to need a *much* shorter latch time estimate than I'd've expected to be stable.  IIRC I recall seeing a source that suggested some designs have strange gap handling that latches on short and long gaps, but there's a middle range that resumes passing through.  (eg. 0-10us pass, 10-25us latch, 25-50us pass, >50us latch).

I don't think it's worth our time adding that sort of handling though.

> Having such a short "latch window" means virtually any interrupt may halt the BB output. This can get critical on longer strips as the end of it may "starve" the end of the strip. My test with 512 LEDs and WS video streaming show significant drop in FPS towards the end of the strip, it was quite choppy, but no flickering.

The NPB implementation won't accept another frame until the current one has been fully output.  I do think that's the correct protocol; better a low overall FPS than effects "tearing".

>  So "do not allow interrupts" should be a config option for people who want to use BB outputs and do not need IR or other time critical interrupts.

Agreed.   Maybe something as simple as setting a latch time of 0 tells the driver to skip the interrupt polling.

willmmiles:
> BTW I just had an idea: this "missed time window" gating could also be used for RMT output: in the translator, check CPU cycles between translations, if the window was missed (which may be a bit hard to threshold, maybe 1.5x the time it take to send out half a memory block size?) I would just detach the GPIO, set it low and let the RMT state machine play out, then re-attach the pin to RMT when the next frame is sent out. I do the same thing in the parallel SPI if a buffer underrun is detected and it successfully prevents white flashes that can happen otherwise.

I tried this approach very early in the RMT fix development.   The problem is that RMT blocks don't always align to the LED word size, so sometimes the timing error ends up halfway through some LED.  You can detect, abort, wait the latch delay, and retry the output at that point; but you still end up with that LED flickering.

Re DMA-based output drivers, it might be wise to adjust the DMA block size to always output an integer number of LEDs -- in the event of an underrun, then, you're at least guaranteed no LED will show a wildly wrong value.

DedeHai:
> I tried this approach very early in the RMT fix development. The problem is that RMT blocks don't always align to the LED word size, so sometimes the timing error ends up halfway through some LED. You can detect, abort, wait the latch delay, and retry the output at that point; but you still end up with that LED flickering.

one led flickering is still an improvement over half the strip flashing even it its not a proper solution.

>Re DMA-based output drivers, it might be wise to adjust the DMA block size to always output an integer number of LEDs -- in the event of an underrun, then, you're at least guaranteed no LED will show a wildly wrong value.

that's a good idea, need to revisit the implementation. what I saw is that with 16 parallel outputs (which just shreds through the DMA buffer fast as that is 8bytes per LED bit) along with RMT or 8 parallel + 8 RMT I had to increase buffer size and use triple buffering to get it flicker free. I am quite sure it can still happen under additional heavy load it could be made to underrun. Need to think about how to detect an underrun or if triple buffering should always be used: since its a linked DMA list, if one buffer is not refilled in time, double buffering will just send the "old" buffer again.

edit:
RE BB stalls: repeating the same buffer is doable, I checked the NPB implementation and I could do it the same, i.e. run a repeat loop if it fails.
The "problem" I see with setting the latch-detect time as low as 10us is that strips that actually use 50us or more would suffer from many (unnecessarily) dropped frame updates - but we also do not want to concern our average user with that.

PR #5691, opened 2026-06-19
[16.0.x port] prevent glitches on -C3 by waiting for LEDs updates to finish

softhack007:
16.0.0 port of #5435 

temporary solution for LEDs glitches on RISC-V boards, as discussed in #5683

previous discussion see #5683 and #5688

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LED output synchronization during uploads, file operations, GIF image loading, WebSocket info sending, and OTA update flows to reduce timing conflicts, flicker, and related glitches.
  * Ensured LED activity completes before sensitive actions like retrieving free memory and closing/resuming files, improving stability during transfers and updates.

* **Improvements**
  * Added a standardized “wait for LED idle/settle” mechanism with selectable very-short/short/medium timing, improving consistency across common device workflows.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1845,21 +1845,21 @@ void WS2812FX::waitForIt() {
  **/
 bool WS2812FX::waitForLEDs(unsigned maxWaitMS, bool always) const {
   #if !defined(WLED_USE_SHARED_RMT) && !defined(__riscv) && !defined(ESP8266)
-    if (!always) return strip.isUpdating(); // no waiting needed if we have the flicker-free RMTHI driver
+    if (!always) return isUpdating(); // no waiting needed if we have the flicker-free RMTHI driver
   #endif
   #ifdef ARDUINO_ARCH_ESP32
     unsigned long waitStart = millis();
-    while (strip.isUpdating() && (millis() - waitStart < maxWaitMS)) delay(1);
+    while (isUpdating() && (millis() - waitStart < maxWaitMS)) delay(1);
   #else
     if (can_yield()) {
       // If we are in a yieldable context (main loop), wait until the LEDs output finishes
       yield();
       unsigned long waitStart = millis();
-      while (strip.isUpdating() && (millis() - waitStart < maxWaitMS)) delay(1);
+      while (isUpdating() && (millis() - waitStart < maxWaitMS)) delay(1);
       yield();
     }
   #endif
-  return strip.isUpdating();
+  return isUpdating();
 }
 
 void WS2812FX::setTargetFps(unsigned fps) {
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1838,11 +1838,15 @@ void WS2812FX::waitForIt() {
 };
 
 /**
- * Be nice, but not too nice - wait until LEDs are idle, or maxWaitMS milliseconds have passed
- * On 8266 this call will _not_ wait outside the main loop context
- *   Function returns isUpdating() status after waiting
- */
-bool WS2812FX::waitForLEDs(unsigned maxWaitMS) const {
+ * Be nice, but not too nice - wait until LEDs are idle, or maxWaitMS milliseconds have passed. 
+ * always=true enforces waiting even when using the RMTHI driver.
+ * On 8266 this call will _not_ wait outside the main loop context.
+ *  Function returns isUpdating() status after waiting.
+ **/
+bool WS2812FX::waitForLEDs(unsigned maxWaitMS, bool always) const {
+  #if !defined(WLED_USE_SHARED_RMT) && !defined(__riscv) && !defined(ESP8266)
+    if (!always) return strip.isUpdating(); // no waiting needed if we have the flicker-free RMTHI driver
+  #endif
   #ifdef ARDUINO_ARCH_ESP32
     unsigned long waitStart = millis();
     while (strip.isUpdating() && (millis() - waitStart < maxWaitMS)) delay(1);
@@ -1855,7 +1859,7 @@ bool WS2812FX::waitForLEDs(unsigned maxWaitMS) const {
       yield();
     }
   #endif
-  return isUpdating();
+  return strip.isUpdating();
 }
 
 void WS2812FX::setTargetFps(unsigned fps) {
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1844,10 +1844,12 @@ void WS2812FX::waitForIt() {
  *  Function returns isUpdating() status after waiting.
  **/
 bool WS2812FX::waitForLEDs(unsigned maxWaitMS, bool always) const {
-  #if !defined(WLED_USE_SHARED_RMT) && !defined(__riscv) && !defined(ESP8266)
-    if (!always) return isUpdating(); // no waiting needed if we have the flicker-free RMTHI driver
-  #endif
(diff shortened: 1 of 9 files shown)


wled00/FX_fcn.cpp
```diff
@@ -1837,6 +1837,31 @@ void WS2812FX::waitForIt() {
...
+    if (can_yield()) {
+      // If we are in a yieldable context (main loop), wait until the LEDs output finishes
+      yield();
+      unsigned long waitStart = millis();
+      while (strip.isUpdating() && (millis() - waitStart < maxWaitMS)) delay(1);
+      yield();
+    }
+  #endif
+  return strip.isUpdating();
+}
```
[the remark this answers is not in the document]
softhack007: this is not critical because WLED only has one global WS2812FX instance. I'll still fix it for "clean code" reasons.

wled00/wled.cpp
```diff
@@ -494,6 +491,7 @@ void WLED::setup()
 
   DEBUG_PRINTLN(F("Initializing strip"));
   beginStrip();
+  strip.waitForLEDs(STRIP_WAIT_MEDIUM); // prevent flickering - beginStrip() calls strip.show()
```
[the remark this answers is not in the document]
softhack007: this is intended. I want to wait _after_ strip.show() has triggered a LEDs update, not before.

softhack007:
@bigfella5000 this is the -C3 flicker fix, brought to `main` (mentioned during discussion of your PR #5683). Can you check if this fixes the flickering you observed on your -C3 setup?

wled00/image_loader.cpp
```diff
@@ -28,10 +28,7 @@ int fileReadCallback(void) {
 }
 
 int fileReadBlockCallback(void * buffer, int numberOfBytes) {
-  #ifdef CONFIG_IDF_TARGET_ESP32C3
-  unsigned t0 = millis();
-  while (strip.isUpdating() && (millis() - t0 < 150)) yield(); // be nice, but not too nice. Waits up to 150ms to avoid glitches
-  #endif
+  strip.waitForLEDs(STRIP_WAIT_SHORT);
```
DedeHai: why short? this is executed during LED update on each frame and gifs are not particularly fast updating

wled00/wled.cpp
```diff
@@ -598,7 +596,7 @@ void WLED::setup()
   #endif
 
   #if defined(ARDUINO_ARCH_ESP32) && defined(WLED_DISABLE_BROWNOUT_DET)
-  WRITE_PERI_REG(RTC_CNTL_BROWN_OUT_REG, 1); //enable brownout detector
+  WRITE_PERI_REG(RTC_CNTL_BROWN_OUT_REG, 1); //disable brownout detector
```
DedeHai: should be "re-enable": it is disabled at the top, not sure what exactly the disabling during setup only does, but it has been this way for 5 years.

wled00/FX_fcn.cpp
```diff
@@ -1837,6 +1837,31 @@ void WS2812FX::waitForIt() {
   #endif
 };
 
+/**
+ * Be nice, but not too nice - wait until LEDs are idle, or maxWaitMS milliseconds have passed. 
+ * always=true enforces waiting even when using the RMTHI driver.
+ * On 8266 this call will _not_ wait outside the main loop context.
+ *  Function returns isUpdating() status after waiting.
+ **/
+bool WS2812FX::waitForLEDs(unsigned maxWaitMS, bool always) const {
```
DedeHai, line 1846: add (or replace always) "bool suspend" and put the suspend code in here?

wled00/file.cpp
```diff
@@ -438,6 +449,7 @@ bool handleFileRead(AsyncWebServerRequest* request, String path){
   }
   #endif
   if(WLED_FS.exists(path) || WLED_FS.exists(path + ".gz")) {
+    strip.waitForLEDs(STRIP_WAIT_SHORT); // wait for LEDs before file access (not using strip.suspend(), to avoid effect stuttering)
```
DedeHai: must go before `WLED_FS.exists()` as that also causes glitches

wled00/file.cpp
```diff
@@ -34,13 +34,24 @@ static File f; // don't export to other cpp files
...
+  if (!doCloseFile || !f) { doCloseFile = false; return; }  // file not open, or no request to close -> nothing to do, nothing to wait
   #ifdef WLED_DEBUG_FS
     DEBUGFS_PRINT(F("Close -> "));
     uint32_t s = millis();
   #endif
+  doCloseFile = false; // consume flag early, to reduce the time window for concurrent closing attempts from several tasks.
+
+  // f.close() may enter flash critical sections (interrupts/cache paused), so we wait for LED transmission to finish first to avoid WS281x glitches
+  // This is most relevant on ESP32-C3/C5/C6, where the RMT driver is very sensitive to interrupt timing.
+  bool haveSuspended = false;
```
DedeHai, line 46: suspend will skip this frame if it already `isServicing()` but maybe that is the price to pay, I don't see an easy way out.

wled00/FX_fcn.cpp
```diff
@@ -1837,6 +1837,31 @@ void WS2812FX::waitForIt() {
...
+  #if !defined(WLED_USE_SHARED_RMT) && !defined(__riscv) && !defined(ESP8266)
+    if (!always) return isUpdating(); // no waiting needed if we have the flicker-free RMTHI driver
```
DedeHai: move into #ifdef ARDUINO_ARCH_ESP32 and drop the esp8266

wled00/image_loader.cpp
softhack007: you're right, this should be the 150ms wait, not the short one.

wled00/FX_fcn.cpp
softhack007: that's a bit tricky ... need to think about it.

wled00/file.cpp
softhack007: good point ... its really a hard measure for every file update (saving presets and saving cfg.json).
The problem is that the main loop might run the next strip.service() when closefile() gets called from the `async_tcp` task context. This could create new flickering.

wled00/FX_fcn.cpp
softhack007: thanks, will do.

softhack007:
esp01_1m_full_160 build has failed - need to check why we're now 3 bytes over program size 🤔 

```
Error: The program size (892915 bytes) is greater than maximum allowed (892912 bytes)
RAM:   [======    ]  56.8% (used 46504 bytes from 81920 bytes)
Flash: [==========]  100.0% (used 892915 bytes from 892912 bytes)
```


edit: my last commit seems to have added 16 bytes that are now pushing us over-the-limit.
```
Checking size .pio/build/esp01_1m_full_160/firmware.elf
RAM:   [======    ]  56.8% (used 46504 bytes from 81920 bytes)
Flash: [==========]  100.0% (used 892899 bytes from 892912 bytes)
```

softhack007:
It looks like the `esp01_1m_full_160` buildenv is totally closed to the flash size limit. All other `esp01_1m_full` buildenvs don't have AR support which gives them around 7Kbytes or safety margin.

I've removed AR from `esp01_1m_full_160` in `main`. Wondering if we should also remove it in `0_16` - its a regression for this specific build compared to the 16.0.0 release.

@DedeHai @netmindz @willmmiles what do you think? OK to remove AR from  `esp01_1m_full_160` also in the  `0_16` branch?

PR #5693, opened 2026-06-20
Fix ledmap parser reading past end of map array

evanc:
## Summary

`WS2812FX::deserializeMap()` (`wled00/FX_fcn.cpp:2160-2178`) reads the `map` array one comma-delimited token at a time, breaking out only when `customMappingSize >= getLengthTotal()`. When `getLengthTotal()` exceeds the actual `map[]` entry count (e.g. a trailing PWM/OnOff bus placed after a 2D matrix), and the JSON has `"map"` before `"width"/"height"`, the loop continues past the array's closing `]` and reads the next JSON key. `atoi("\"width\":23")` returns 0, which gets stored at `customMappingTable[matrixSize]`. The trailing strip's segment writes are then silently rerouted to physical pixel 0.

The existing `if (!foundDigit) break` handles the case where `]` appears alone in a chunk. This adds the symmetric case for when `]` follows the last valid digit in the same chunk — store the entry, then exit.

## How it works

```cpp
customMappingTable[customMappingSize++] = index;
if (end != nullptr) break;  // ← new line
if (customMappingSize >= getLengthTotal()) break;
```

`end` is already computed earlier as `strchr(number, ']')`. If non-null, the chunk we just processed contained the array terminator, so there are no more entries to read regardless of what `getLengthTotal()` is.

## Reproduction (pre-fix)

- ESP32-C3, WLED 16.0.0 (parser code is identical on `main` and `16_x`)
- Digital strip on one pin: 55 LEDs at start=0
- PWM White (TYPE_ANALOG_1CH) on a separate pin at start=529 (post-matrix trailing strip)
- `ledmap.json` defining a 23×23 matrix with the WLED-MM key order: `{"map":[...],"width":23,"height":23}`
- A 1D segment at `[529, 530)`

Writes to segment 1 are routed to physical pixel 0 of the digital bus instead of the PWM bus. Pin 9 receives no PWM updates.

## Why JSON key order matters

Commit 336e074b ("fix for 0byte size files, also made reading ledmaps more efficient", Nov 2025) changed WLED's UI to generate `"width"`/`"height"` before `"map"`. That implicitly dodges this bug because when the parser reads past the array, it hits the closing `}` — `strchr` finds `]`-or-nothing patterns it already handles via the no-digit-found path. The original PR framed the change as a load-time optimization; it also turns out to be a quiet workaround. WLED-MM and pre-`336e074b` ledmaps still trigger the bug.

## Testing

- Compiled clean on `esp32c3dev` and `esp32dev` (`pio run -e esp32c3dev`, `pio run -e esp32dev`).
- `npm test` passes.
- Tested on actual hardware (ESP32-C3 / Seeed Xiao, SK6812 RGBW matrix + PWM White via MOSFET):
  - With current `main`, broken-order ledmap, trailing PWM bus → PWM dead, matrix pixel 0 driven by segment 1. ✗
  - With this patch, same ledmap, same buses → PWM driven by segment 1, matrix pixel 0 untouched. ✓
- No behavior change for ledmaps where `"width"`/`"height"` come first (the `strchr(']')`-with-no-digit path still handles them).

## Risk

One line added, scoped to the trailing-strip + map-first JSON ordering case. The new break only fires on chunks that already contained the array terminator, so well-formed JSON with width/height first is unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed LED map parsing to correctly stop at the end of the mapping array, preventing misinterpretation of trailing JSON data.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -2173,6 +2173,7 @@ bool WS2812FX::deserializeMap(unsigned n) {
         int index = atoi(number);
         if (index < 0 || index > 65535) index = 0xFFFF; // prevent integer wrap around
         customMappingTable[customMappingSize++] = index;
+        if (end != nullptr) break; // array closing ']' was in this chunk; stop before atoi() coerces trailing JSON keys into bogus entries
         if (customMappingSize >= getLengthTotal()) break;
       } else break; // there was nothing to read, stop
     }


softhack007:
While the explanation ("why JSON key order matterrs") is full of wrong factual claims, the fix may still be reasonable. 

There is actually no relation to leading width/height fields (old MM format). Line 2159 `f.find("\"map\":[")` makes sure that array reading always starts at the right position, so any tags before that won't matter. Also the statement that the UI writes a ledmap.json file is wrong. But it might be possible that a short ledmap is read one element past the end of the list - i'll double-check this.

Silently re-routing to pixel 0 is possibly related to how the ledmap array gets initialized - a customMappingsTable entry that's not present in the ledmap.json file would retain the previous value - initially 0.

This will get solved with another PR
* https://github.com/wled/WLED/pull/5242

softhack007:
@coderabbitai please verify if the problem described by the user is actually caused by reading past the end of array in ledmap.json. you can assume that the input is a sized-optimized JSON file without additional whitespace. 

The source code to check starts here
https://github.com/wled/WLED/blob/3a46f444957f010f788cf27255204ab1ac08530e/wled00/FX_fcn.cpp#L2154

Make sure you ignore the explanations provided in the PR description, as the explanations provided in the PR description are full of wrong assumptions.

Also check if #5242 alone would fix the problem.

DedeHai:
its a valid fix, definitely a bug but also only shows on "sloppy"  incomplete ledmaps. 
Also saving the ledmap in the editor will fix it (I think this is what is meant by "the UI saves the ledmap") as the editor parses a ledmap and puts the width and height in front.
A ledmap must contain all entries, if entries are skipped, there is no guarantee it will map correctly: this was discussed in length and I think there is a pending (but unifinshed?) PR to the docs stating this.

softhack007:
quick comparison with WLED-MM code confirms this fix is good. 
In fact WLED-MM has a cleaner logic for reading the map array, and it also ends reading when "]" was found.

https://github.com/MoonModules/WLED-MM/blob/43daffb1ded0796fefcf4182afc6b9075bb7aac2/wled00/FX_fcn.cpp#L2739-L2740

PR #5695, opened 2026-06-21
align with WLEDMM extended error codes, small improvement for brownout detection

softhack007:
Backport and alignment of error codes between upstream WLED and WLED-MM
Fixes some inconsistencies introduced with 56b3345f6f6ffad93fc55cb8a7e566b54cba1773, and improves brownout detection by first checking RTC restart reason.

* Align ERR_REBOOT_NEEDED and ERR_POWEROFF_NEEDED with WLEDMM (prepare for AR out-of-tree)
* use new error codes in AR and in HUB75 driver
* add new error codes to WebUI
* improved brownout detection on ESP32

open ends:
* "persistent errors stay in the UI until restart" is not implemented in this PR - will come as a follow-on for 17.0.0-dev only
* brownouts that cause a reboot loop will still not show in the WebUI, same about crash restart loops.

This change is low-risk and fixes a few inconsistencies introduced in a previous commit. I'd say its good for 16.0.1.

## Summary by CodeRabbit

* **Bug Fixes**
  * Audio-reactive microphone and I2S configuration changes now retain existing settings and reliably require a reboot or power-cycle when relevant hardware settings are altered.
  * Improved system reboot/bootloop reporting distinguishes brownout resets from reboots.
  * On ESP32-S3, display cleanup now correctly initiates a reboot when needed.

* **New Features**
  * Added new system error/brownout and reboot-related toast messages, and refreshed error-code handling and “low memory” messaging behavior.

### DIFF
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -481,6 +481,8 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
 #define ERR_SYS_BROWNOUT  91  // reboot after brownout alert
 #define ERR_PERSISTENT_THRESHOLD 100 // (WLEDMM compatibility) errors below this value are non-persistent; persistent errors stay in the UI until restart
 // ERR_PERSISTENT_THRESHOLD is a threshold value only - never assign directly to errorFlag
+// Reserved threshold for WLEDMM-compatible persistent error handling.
+// Current runtime behavior still clears errorFlag after serialization.
 #define ERR_REBOOT_NEEDED 100 // reboot needed after changing hardware setting
 #define ERR_POWEROFF_NEEDED 101 // power-cycle needed after changing hardware setting
 
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -472,8 +472,17 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
 #define ERR_OVERTEMP    30  // An attached temperature sensor has measured above threshold temperature (not implemented)
 #define ERR_OVERCURRENT 31  // An attached current sensor has measured a current above the threshold (not implemented)
 #define ERR_UNDERVOLT   32  // An attached voltmeter has measured a voltage below the threshold (not implemented)
-#define ERR_REBOOT_NEEDED 98 // reboot needed after changing hardware setting
-#define ERR_POWEROFF_NEEDED 99 // power-cycle needed after changing hardware setting
+//#define ERR_LOW_MEM     33  // WLEDMM: low memory (RAM)
+//#define ERR_LOW_SEG_MEM 34  // WLEDMM: low memory (segment data RAM)
+//#define ERR_LOW_WS_MEM  35  // WLEDMM: low memory (ws)
+//#define ERR_LOW_AJAX_MEM  36 // WLEDMM: low memory (oappend)
+//#define ERR_LOW_BUF     37  // WLEDMM: low memory (LED buffer from allocLEDs)
+#define ERR_SYS_REBOOT    90  // reboot after error, trying to roll back
+#define ERR_SYS_BROWNOUT  91  // reboot after brownout alert
+#define ERR_PERSISTENT_THRESHOLD 100 // (WLEDMM compatibility) errors below this value are non-persistent; persistent errors stay in the UI until restart
+// ERR_PERSISTENT_THRESHOLD is a threshold value only - never assign directly to errorFlag
+#define ERR_REBOOT_NEEDED 100 // reboot needed after changing hardware setting
+#define ERR_POWEROFF_NEEDED 101 // power-cycle needed after changing hardware setting
 
 // JSON buffer lock owners
 #define JSON_LOCK_UNKNOWN        255
--- a/wled00/const.h
+++ b/wled00/const.h
@@ -472,18 +472,16 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
 #define ERR_OVERTEMP    30  // An attached temperature sensor has measured above threshold temperature (not implemented)
 #define ERR_OVERCURRENT 31  // An attached current sensor has measured a current above the threshold (not implemented)
 #define ERR_UNDERVOLT   32  // An attached voltmeter has measured a voltage below the threshold (not implemented)
-//#define ERR_LOW_MEM     33  // WLEDMM: low memory (RAM)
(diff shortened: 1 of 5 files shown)


wled00/util.cpp
```diff
@@ -998,6 +999,11 @@ RTC_NOINIT_ATTR static uint32_t bl_crashcounter;
 RTC_NOINIT_ATTR static uint32_t bl_actiontracker;
 
 static inline ResetReason rebootReason() {
+  // check RTC restart reason first - brownout is not reliably reported by esp_reset_reason()
+  if (rtc_get_reset_reason(0) == RTCWDT_BROWN_OUT_RESET) return ResetReason::Brownout; // core0 brownout
```
DedeHai, line 1003: were you able to test this? I did never get a brownout in my tests, if I lowered the voltage it either worked or resulted in a "powerup" as a reasaon. 
also according to https://esp32.com/viewtopic.php?t=34938 the rtc variant is lower level and I am not sure it has an advantage. Did you investigate?

softhack007: Yes the RTC variant is running in MM for years, and I've seen it report some brownouts over that time (however mainly with the old framework, not  tested with tasmota). Also some users have reported they saw the brownout message with MM. 

Good point about testing - I can try to dig out one esp32 that sometimes browns out, and see what each crash reason function says.

softhack007: just to be sure, i'll ask the rabbit to dig through the esp-idf code ....

@coderabbitAI please compare the esp-idf 4.4.8 implementations of `rtc_get_reset_reason()` and `esp_reset_reason()`. focus on esp32, esp32-c3 and esp32-s3, and look for the origins where the "brownout" restart reasons codes come from. Apart from `rtc_get_reset_reason()` allowing to query the status of core0 and core1 separately, are there other differences in esp-idf source code that make detecting a previous "brownout restart" more reliable with one or the other function call?

wled00/const.h
```diff
@@ -472,8 +472,19 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
 #define ERR_OVERTEMP    30  // An attached temperature sensor has measured above threshold temperature (not implemented)
 #define ERR_OVERCURRENT 31  // An attached current sensor has measured a current above the threshold (not implemented)
 #define ERR_UNDERVOLT   32  // An attached voltmeter has measured a voltage below the threshold (not implemented)
-#define ERR_REBOOT_NEEDED 98 // reboot needed after changing hardware setting
-#define ERR_POWEROFF_NEEDED 99 // power-cycle needed after changing hardware setting
+//#define ERR_LOW_MEM     33  // WLEDMM: low memory (RAM)
+//#define ERR_LOW_SEG_MEM 34  // WLEDMM: low memory (segment data RAM)
+//#define ERR_LOW_WS_MEM  35  // WLEDMM: low memory (ws)
+//#define ERR_LOW_AJAX_MEM  36 // WLEDMM: low memory (oappend)
+//#define ERR_LOW_BUF     37  // WLEDMM: low memory (LED buffer from allocLEDs)
```
willmmiles: Why leave these commented out?  Is there any reason not to accept these definitions as a formal part of the API?

wled00/const.h
```diff
@@ -472,8 +472,19 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
...
+#define ERR_SYS_REBOOT    90  // reboot after error, trying to roll back
+#define ERR_SYS_BROWNOUT  91  // reboot after brownout alert
+#define ERR_PERSISTENT_THRESHOLD 100 // (WLEDMM compatibility) errors below this value are non-persistent; persistent errors stay in the UI until restart
+// ERR_PERSISTENT_THRESHOLD is a threshold value only - never assign directly to errorFlag
+// Reserved threshold for WLEDMM-compatible persistent error handling.
+// Current runtime behavior still clears errorFlag after serialization.
```
willmmiles: I disagree with the framing of these comments -- I think your AI may be taking your prompt too literally instead of considering the context.   While I strongly believe that this project should be accepting features and standards that originated in WLEDMM, at no point should this project be seeking "compatibility" with another fork, and the comments should never imply that.

Instead, where there's clear benefits to harmonization, we should be adopting the improved features fully as part of this project - not treating them as foreign.

(To put it another way: seeing 'WLEDMM' here is a code smell.  As the "root fork", this project sets the standard.  We should accept improvements as they are meant to be: improvements to the standard.  We shouldn't be enshrining things in the standard that are explicitly labelled as weird-fork-isms.)

willmmiles (changes requested):

wled00/const.h
softhack007: good point 👍 i'l simplify the comment. My intention was "this is not working yet" (will port it later), creating a weird-fork-smell is weird, indeed :-)

wled00/const.h
softhack007: the only reason was that index.js still has them commented out, so using them would only show as "raw numbers" in an error message. I'l enable them. They might come handy when we proceed with the AR out-of-tree integration.

wled00/const.h
```diff
@@ -472,8 +472,17 @@ static_assert(WLED_MAX_BUSSES <= 32, "WLED_MAX_BUSSES exceeds hard limit");
...
+#define ERR_LOW_SEG_MEM 34  // low memory (effect data RAM)
+#define ERR_LOW_WS_MEM  35  // low memory (ws)
+//#define ERR_LOW_AJAX_MEM 36 // (not used any more) low memory (oappend)
+#define ERR_LOW_BUF     37  // low memory (LED pixels buffer)
+#define ERR_SYS_REBOOT      90  // reboot after error, trying to roll back
+#define ERR_SYS_BROWNOUT    91  // reboot after brownout alert
+#define ERR_PERSISTENT_THRESHOLD 100 // ToDO: errors below this value are non-persistent; persistent errors stay in the UI until restart
+// ERR_PERSISTENT_THRESHOLD is a threshold value only - never assign directly to errorFlag
+#define ERR_REBOOT_NEEDED   100 // reboot needed after changing hardware setting
+#define ERR_POWEROFF_NEEDED 101 // power-cycle needed after changing hardware setting
```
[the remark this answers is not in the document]
softhack007: the comment in line 483 should clarify this.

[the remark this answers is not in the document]
softhack007: HUE_ERROR_PUSHLINK = 101 is for `hueError` not for `errorFlag`

willmmiles (approved):

PR #5698, opened 2026-06-23
use NPB version with fixed buffers

### LINKED ISSUE #5378: LEDs excluded with gap file flash randomly
### What happened?

When using more than 1 output and 2D gap file to exclude some LEDs from display, the excluded LEDs on 2nd (and 3rd) output flash with random colors.
Non-excluded LEDs work normally.

The flashing may look different on each reboot.

### To Reproduce Bug

Create 2 or 3 outputs, set up 2D matrix (doesn't matter if you use 1 panel or more), upload 2D gap file that excludes some LEDs on the matrix (use 0, not -1).
Start any effect and observe actual LEDs (peek will not show flashes).

### Expected Behavior

No flashes occur and excluded LEDs remain off (black).

### Install Method

Binary from WLED.me

### What version of WLED?

WLED 0.15.3

ESP32

### Relevant log/trace output

```shell
N/A
```

### Anything else?

Appears to be a bug/feature of NeoPixelBus and "inconsistent" `Show()` call on double buffered bus (RMT, Async UART). "Inconsistent" `Show()` is used for speed.

The easiest solution is to call "consistent" `Show()` in bus wrapper but if you want to maintain speed benefit then the solution is to manually clear LED buffers during bus initialization (two `Show()` calls necessary with all pixels painted black between calls).

The fix has been included in recent (but not yet released) NeoPixelBus. NeoPixelBus 2.8.5 will include the fix and no code modification will be necessary in WLED.

### Code of Conduct

DedeHai:
Bumps up NPB by one commit which fixes unitialized buffers causing garbage output when using gap files that do not send data to all LEDs. 

thx @blazoncek for fixing it and pointing this out.

fixes #5378

## Summary by CodeRabbit

* **Chores**
  * Updated NeoPixelBus library dependency to the latest version.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -130,7 +130,7 @@ upload_speed = 115200
 lib_compat_mode = strict
 lib_deps =
     IRremoteESP8266 @ 2.8.2
-    https://github.com/Makuna/NeoPixelBus.git#a0919d1c10696614625978dd6fb750a1317a14ce
+    https://github.com/Makuna/NeoPixelBus.git#1d7ff38f14d04a976f9c6e365c83a232bcba04fc
     https://github.com/Aircoookie/ESPAsyncWebServer.git#v2.4.2
     marvinroger/AsyncMqttClient @ 0.9.0
   # for I2C interface


softhack007 (approved):
looks good for me 👍

DedeHai:
cherry picked to 16_x

PR #5701, opened 2026-06-25
Bugfix in cfg loading in settings_leds.htm

DedeHai:
a const can not be incremented, it throws an error.

remove dead code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined LED count and memory threshold behavior so submission is consistently blocked at the highest limit, while the higher warning level shows only the reboot recommendation.
  * Improved LED output form behavior for more consistent per-output labeling and limit calculations across multiple outputs.
  * Updated WiFi scan UI so SSID options are no longer cleared before refresh, preventing disappearing/duplicated entries during scans.

* **Chores**
  * Made small internal cleanup in the LED settings UI and configuration template handling for maintainability.

### DIFF
--- a/wled00/data/settings_wifi.htm
+++ b/wled00/data/settings_wifi.htm
@@ -65,10 +65,6 @@
 						select.setAttribute("onchange", "T(this)");
 						preScanSSID = input.value;
 
-						for (let i = 0; i < select.children.length; i++) {
-							select.removeChild(select.children[i]);
-						}
-
 						for (let i = 0; i < networks.length; i++) {
 							const option = cE("option");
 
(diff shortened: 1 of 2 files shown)


DedeHai:
@coderabbitai please thoroughly check all htm files in the /data folder for other such small bugs and dead or unreachable code.

wled00/data/settings_wifi.htm
```diff
@@ -65,10 +65,6 @@
 						select.setAttribute("onchange", "T(this)");
 						preScanSSID = input.value;
 
-						for (let i = 0; i < select.children.length; i++) {
-							select.removeChild(select.children[i]);
-						}
```
softhack007, line 70: Was this dead code?

softhack007 (approved):
Overall looks reasonable (but I did not test).

wled00/data/settings_wifi.htm
DedeHai: not dead per se but not needed: the "select" is created just a few lines above, it does not have any children as none were added.

edit: I did not find out when this code was added and what purpose it served in the past.

PR #5704, opened 2026-06-25
New LED hardware driver: WLED Pixel Bus

DedeHai:
This is a major update adding WLED Pixel Bus written from scratch to replace NeoPixelBus and it packs a lot of useful features. From a user perspective the main improvements are less memory use and dynamic adjustment of LED timing to eliminate flickering by fine-tuning the timing from -30% to + 30% in 10% steps (dropdown menu).

The new hardware driver structure allows for fully dynamic updates of LED outputs and LED timings. It also adds glitch free parallel SPI output support on the C3 as well as parallel bit-banging on all platforms including ESP8266.
All parallel outputs (except bit-bang) use ping-pong DMA buffers instead of fully pre-filled buffers - memory usage is optimized and DMA buffers are independent of number of LEDs. Also the calculation from LED colors to DMA buffers is highly speed optimized, making the driver blazing fast - I saw 30% FPS improvements in some cases.
The fully dynamic nature of the driver allows for a "Custom digital bus" in the LED settings to support virtually any LED bus type out there - specify timing, number of color channels, invert any color channel or set any color channel combination. The driver also supports inverting the output signal on any pin (ESP32 only).

The LED config UI was updated to fully support the new driver options.

A lot more testing is required for all different kind of LEDs on all platforms. I am sure there is a ton of bugs and kinks to iron out.

I would like to take this opportunity to thank @Makuna  for his outstanding work on NeoPixelBus which served me well as a reference on hardware configurations and special LED types.

- saves over 10k of heap on ESP32 and about 3k on ESP8266
- increases available LED outputs on ESP32 C3 from 2 to 14 (2x RMT, 4x parallel SPI, 8x bit-bang)
- increases available LED outputs on ESP32 S2 with 8x bit-bang if I2S is not available
- increases available LED outputs on ESP8266 3 to 6 (1x UART, 1xI2S DMA, 4x bit-bang)
- allows over 1000 LEDs again on ESP8266 thanks to reduced RAM footprint, it does not use a global buffer but writes to LED buffers directly which is slower but uses less RAM.
- adds support for TM1815
- improved low-brightness color accuracy to TM1814, TM1815 and APA102 by utilizing LED current settings
- improved color resolution on 16bit LED types (UCS8903, UCS8904, SM16825)
- if needed I2S (ESP32 and ESP32 S2) and LCD (ESP32 S3) driver can output on 16 parallel pins changing a #define
- automatic calculation and optimized use of multiple RMT memory blocks, reducing ISR frequency when not using all RMT outputs and thus increases FPS

ESP32 flash and RAM usage comparison (bit bang disabled)

using NeoPixelBus
RAM:   [==        ]  24.9% (used 81576 bytes from 327680 bytes)
Flash: [========  ]  83.1% (used 1306493 bytes from 1572864 bytes)
**Free Heap: (6 outputs, 256 each): 79.8k**

using WLEDpixelBus
RAM:   [==        ]  24.9% (used 81680 bytes from 327680 bytes)
Flash: [========  ]  83.0% (used 1306053 bytes from 1572864 bytes)
**Free Heap: (6 outputs, 256 each): 92.8k**

## Summary by CodeRabbit

* **New Features**
  * Added per-bus **Bus Speed Factor (SF/bsf)** settings.
  * Added **Custom Bus** support with channel mapping, inversion, custom timing, and per-channel colors.
  * Expanded supported output driver/type options (bit-bang, parallel SPI, RMT, I2S, LCD where available) and added new LED type options (e.g., **TM1815** / custom bus).
* **Bug Fixes**
  * Improved **ESP8266** direct-to-bus pixel handling and refined **gamma/brightness** behavior across realtime and particle rendering.
* **Chores**
  * Updated capability limits and the LED settings UI + config import/export for the new bus features.


DedeHai:
compiles just fine locally, need to look into those errors
edit: fixed for ESP32, S3 still has some issue

DedeHai:
@coderabbitai the compiled file shows over 200 bytes of additional static RAM use. Please explain where this is consumed and if it can be optimized.

DedeHai:
@coderabbitai why does the build on S3 fail? I do see that define in the file, it still sais not defined.

wled00/FX_fcn.cpp
```diff
@@ -1452,11 +1441,32 @@ void WS2812FX::blendSegment(const Segment &topSegment) const {
   const size_t  stopIndx   = startIndx + length;
   uint8_t       opacity    = topSegment.currentBri(); // returns transitioned opacity for style FADE
   uint8_t       cct        = topSegment.currentCCT();
-  if (gammaCorrectCol) opacity = gamma8inv(opacity); // use inverse gamma on brightness for correct color scaling after gamma correction (see #5343 for details)
+  opacity = gamma8inv(opacity); // use inverse gamma on brightness for correct color scaling after gamma correction (see #5343 for details)
 
   const Segment *segO = topSegment.getOldSegment();
   const bool hasGrouping = topSegment.groupLength() != 1;
 
+  #ifdef ESP8266
```
willmmiles: This should be a feature flag (`WLED_ENABLE_DIRECT_RENDER` or the like?), not an architecture gate.   If we're going to allow this, we should also bring back the "Enable global buffer" option on devices that enable it so it can be optional -- a lot of the relevant code paths already seem to have a `_pixels == nullptr` gate still anyways.

Personally I want gamma correct blending on my low pixel count devices, and I think it's important to allow 100% compatible rendering mode with ESP32 devices.  I don't agree that the RAM tradeoff is always worth it.

wled00/src/WLEDpixelBus/WLEDpixelBus_ESP8266.h
```diff
@@ -0,0 +1,180 @@
...
+  static uint16_t  s_numPixels[WLED_MAX_BB_CHANNELS];
+  static uint8_t*  s_pixelData[WLED_MAX_BB_CHANNELS];
+  static uint8_t   s_channelCount;
+  static uint32_t  s_allMask;       // GPIO bitmask of all registered output pins
+  static uint8_t   s_stagedCount;   // how many channels have called show() this frame
+  // Timing in CPU cycles — identical across all channels
+  static uint32_t  s_t0h;
+  static uint32_t  s_t1h;
+  static uint32_t  s_period;
+  static uint8_t   s_pixelBytes;    // bytes per encoded pixel
```
willmmiles: These should be in a singleton that's heap allocated when a relevant bus is constructed for the first time.  There's no need to permanently waste RAM if they're not used.

DedeHai: thats a good point, I had to look up what a singleton is but makes perfect sense. Should do the same in ESP32 bit bang driver, maybe other places too.

wled00/FX_fcn.cpp
DedeHai: I intentionally did this for ESP8266 only as I did not want to drag along the "no buffer" code everywhere as it adds conditionals in the render hot path but making it a feature flag that is only supported on ESP8266 makes sense if this path is needed in the future.

>Personally I want gamma correct blending on my low pixel count devices, and I think it's important to allow 100% compatible rendering mode with ESP32 devices. I don't agree that the RAM tradeoff is always worth it.

Ah so this is the "something is still missing" in the back of my head. Should be easy to apply gamma in 
scaleAll().
In fact I was thinking about moving brightness and gamma down to the bus level, I have a rough sketch of code that adds "per color scaling" as well but I did not want to introduce more changes than needed and make this a "change everything at once" PR.

edit:
checking the code, the gamma and brightness handling is still a bit of a mess, I did add a TODO but never done it. As it is, gamma is now applied before brightness which results in color loss. Should really move brightness and gamma down to bus level but that messes with CCT.

DedeHai:
@coderabbitai please check the ESP8266 path and verify CCT is handled the same as in the ESP32 path - highlight any discrepancies and suggest a fix. Also suggest an efficient fix for the missing gamma correction in the ESP8266 path.

DedeHai:
@coderabbitai your suggestion to fix gamma for ESP8266 is flawed, if there are multiple overlapping segments, gamma will be applied more than once.

wled00/src/WLEDpixelBus/WLEDpixelBus_ESP8266.h
DedeHai: done

wled00/FX_fcn.cpp
willmmiles: > I intentionally did this for ESP8266 only as I did not want to drag along the "no buffer" code everywhere as it adds conditionals in the render hot path but making it a feature flag that is only supported on ESP8266 makes sense if this path is needed in the future.

I absolutely agree it should be a compile time feature gate so we can keep the hot path conditional-free in most cases.

My biggest concern is being able to get strict compatibility.  If we can't 100% guarantee identical blending results, then I think we should have the ability to turn the feature on and off regardless of architecture.   I might want my short-strip ESP8266es to match my ESP32s, but someone else who has a bunch of old 8266es with long strips deployed might want to go the other way around and have their new ESP32s match their 8266es.

I'm happy to take "on by default in our 8266 envs, off by default on ESP32" via our platformio environments (or platform HAL later), but I don't think the algorithm choice should be tied to the processor type in the code.  If/when we support new platforms, we might face similar tradeoffs in the future.

> edit: checking the code, the gamma and brightness handling is still a bit of a mess, I did add a TODO but never done it. As it is, gamma is now applied before brightness which results in color loss. Should really move brightness and gamma down to bus level but that messes with CCT.

Yeah, it's not easy to untangle.  I'm happy letting them be different for the time being as long as we can choose which path is taken on a per-build basis.

wled00/src/WLEDpixelBus/WLEDpixelBus_ESP8266.h
willmmiles: > thats a good point, I had to look up what a singleton is but makes perfect sense. Should do the same in ESP32 bit bang driver, maybe other places too.

Yes, IMO there's way too much in common between the two versions of the bit bang driver - they should probably be sharing code.

I haven't thoroughly reviewed all of the other drivers yet but I rather suspect the same basic issue applies in many places.  Todays AIs are just like junior devs, they absolutely love to copy-and-paste or write from scratch when they should be building re-usable functions and data structures.  (Offhand: the replicated DMA encoding loops are a prime example.)

willmmiles:
I'll be reviewing slowly and in stages - there's a lot here and it's difficult to get through all at once.  We don't necessarily have to address everything before we land this; it's big enough that we can expect some ongoing development.

Timeline-wise, I think we want to land V5 first and take this immediately after.

wled00/FX_fcn.cpp
DedeHai: I agree, when I looked through the code today I noticed some discrepancies (see also my hidden enquiry to the rabbit below). There are some downsides not using a global buffer, it can diminish color quality when using overlapping segments - I did not test yet if it even looks right, there may be bugs. So being able to disable it as a feature flag makes sense.

edit:
we could even make the no-buffer path the opt-in define since there were no complaints I know of in 16.0

DedeHai:
> I'll be reviewing slowly and in stages - there's a lot here and it's difficult to get through all at once. We don't necessarily have to address everything before we land this; it's big enough that we can expect some ongoing development.
> 
> Timeline-wise, I think we want to land V5 first and take this immediately after.

there is no rush - I am still testing cases and find bugs all the time. I did test each and every driver and every LED type at one point but changed it later again so not sure everything is in order, for example the LCD driver, I did many tests during development using the 16-parallel output but it's possible it is now broken when I changed to the 8-parallel version. I made this PR to get some feedback but it is by no means finished and ready to merge.

I agree that the V5 should be done first - it will probably break all the drivers though as some API calls won't be available or changed names. It will take some work to get them running again but the hard part was the structure and the correct init sequences - API calls are more semantics at this point.

wled00/src/WLEDpixelBus/WLEDpixelBus_ESP8266.h
DedeHai: I am sure there is room for optimization, I had claude come up with the structure for each driver seperately and changed stuff over and over (as you can see in the commit history) to distill the current version - it was not clear to me in the beginning what the best structure will be or what will even work so did it step by step with no grand master plan.
The color encoding was a bit of a head-scratcher in the beginning but I am pretty happy how that turned out.

edit: I first did the ESP8266 BB code and once that worked well had claude derive the ESP32 version from that and then fixed it until working. Apart from a few details they are pretty much identical.

Merikei:
This is great! Thanks so much!

I built for `esp32c3dev` (no build option edits)
at commit b8e9b53f330235979b93bb4d556e595bed4ef7eb
Date:   Sat Jul 11 15:38:13 2026 +0200
bugfix in UI for ESP8266

then flashed OTA.
Non of the usual glitches, even when stress-tested (4 browsers actively using the web interface, peeking, transferring data over the WiFi it's using),
however it does keep crashing and restarting, even when only Home Assistant is connected to it, between every 15~45 minutes, based on a few hours of testing.

- LED outputs
  - Driver: RMT
  - Timing: default
  - 1: WS281x RGB
    - Len: 1
    - GPIO: 8
    - (LED built in to the dev board, didn't glitch even with glitchy firmwares)
  - 2: APA106/PL9823 RGB
    - Len: 100
    - (sold as “WS2812B… WS2812… BTF-LIGHTING… Epistar”, but was trying different options to see if that improved things in the previous glitchy firmware (tested recently: `0.15.3`, `16.0.1`,`17.0.0-dev (2607060)`) where the first few pixels glitched)
    - GPIO: 4

Bord sold as “TENSTAR ESP32-C3 Supermini Plus Development Board WiFi Bluetooth ESP32 C3 Super mini V2.0 Red Board”
```
esptool v5.1.0
Chip type:          ESP32-C3 (QFN32) (revision v0.4)
Features:           Wi-Fi, BT 5 (LE), Single Core, 160MHz, Embedded Flash 4MB (XMC)
Crystal frequency:  40MHz
```

Not a power supply issue: reboots while LEDs are off, but can turn to 255/255 white for 5 min and it's fine & wasn't happening on glitchy firmware that used the same PSU.
Not a Max. TX power issue: turned down to 8.5 dBm VS 15 was fine on glitchy firmware.

Happy to perform tests or get serial logs if you need?
But this is also a draft PR and there have been commits since I built it yesterday so it may be a known issue.

DedeHai:
thanks for testing. Yes I am aware of this bug but I was not able to track it down, it's something to do with low level RMT stuff from espressif, maybe a buffer race condition or whatnot. If you use the SPI driver instead of RMT there should be no crashes.

DedeHai:
@willmmiles I made some progress and think this PR is coming along nicely. Still a lot of work left and I hit a bit of a road-bump on the C3: for some reason it crashes when using RMT. I tried to track it down and it all works up to the point where rmt_write_sample() is called in show(). Under heavy WiFi traffic I can easily get it to stall within a minute, everything freezes up for a few seconds, then reboots, similar to the issue with using the high priority driver but I am not using that on the C3.

willmmiles:
> @willmmiles I made some progress and think this PR is coming along nicely. Still a lot of work left and I hit a bit of a road-bump on the C3: for some reason it crashes when using RMT. I tried to track it down and it all works up to the point where rmt_write_sample() is called in show(). Under heavy WiFi traffic I can easily get it to stall within a minute, everything freezes up for a few seconds, then reboots, similar to the issue with using the high priority driver but I am not using that on the C3.

I certainly wouldn't rule it out.   IIRC the signature to look for is a watchdog timeout in `vPortYield()`.  So far my suspicion remains that there's a hardware race with the interrupt controller handling nested interrupts when the software interrupt is pending -- there's some path that results in failing to clear the pending bit, so `vPortYield()` waits forever.  The high priority RMT driver just hits it routinely, the bug is always lurking.

DedeHai:
ok, I will leave it "as is" for now, it just was strange that it crashes quite soon (minutes) when stressing it and I can not get 16.0 to crash at all (have tried for hours under full load). Maybe it has also something to do with RMT init sequence, I did not check if  its done the very same way as NPB.

DedeHai:
@willmmiles I made a lot of updates and it is now in a pretty good shape. I compiled a TODO list in WLEDpixelBus.cpp with the important things that still should be updated / checked / fixed, all of them are improvements that do not need immediate fixing.
If you find the time, please have another look at it and check if you agree with the general approach of handling things WRT future compatibility. The drivers themselves are (or were at one point) tested. ESP8266 slipped my testing recently so I can't vouch for it not having issues, I focused on the V5 updates mostly.

willmmiles:
OK! I should be able to get some time to do a detailed review on the weekend.

willmmiles:
I'm finding it challenging to review this properly - I just keep exploding in a hundred nitpicks.  It's hard to know where to start and I feel bad pushing suggestions in dribs and drabs.

Bigger things I've noticed so far:
- `WLEDpixelBus.h` has duplicate copies of code from `colors.h`.  This component already isn't isolated from WLED internals (it's importing `const.h` for WLED's type mapping).  If we want to maintain the conceit that this is an independent library, pull out `colors.h` to its own "library" that both WLED and WLEDpixelBus can depend on.   (wled::colors ?? )  Otherwise we should use the definitions for `colors.h` directly.  (It makes sense in its way: one task performed here is to map between WLED's internal color format to the output pixel format.)
- The split between `BusDigital` and `PixelBus` is costly -- calls like `setPixelColor` invoke two vtable lookups.  (And a third dynamic dispatch for the color encoding -- `switch` and `virtual` are conceptually similar.)   I don't have an immediate suggestion; this is mostly a performance tuning concern and we don't necessarily have to address it up front.   From an API perspective, the key change would be moving to a factory function (eg replacing `new BusDigital(...)` to `makeDigitalBus(...)`) -- this would let the `PixelBus` system someday return a fully qualified `Bus` type with fewer dynamic dispatches.
- `CustomBusConfig` -- Good call to factor this separately from the common `BusConfig`, but directly embedding it in `BusConfig` gives up all the advantages.  I suggest using a type-erased bus-specific-settings slot in the `BusConfig` level (eg. instead of `CustomBusConfig custom`, do `std::shared_ptr<void> customBusSettings` or the like), and then cast it to the bus-specific struct during construction.  This lets each `Bus` implement its own custom settings struct and doesn't require sharing the type outside of the local bus implementation and local bus settings parser.  (The next step will be to factor the config parser on a per-`TYPE_` basis, leveraging `getLEDTypes()`, but that's a plan for another day.)

DedeHai:
Thanks for taking a dive.

> I'm finding it challenging to review this properly - I just keep exploding in a hundred nitpicks. It's hard to know where to start and I feel bad pushing suggestions in dribs and drabs.

Fully understand, its a major overhaul and a "diamond in the rough", I also did not start out with a full concept but kept adding and replacing things (as you can see in the PR having over 200 commits). Each time I read through the code I find things that could be done better or more clearly but let's do it in increments, its just too much code to have a clean overview and its easy to get lost in details. 

> 
> Bigger things I've noticed so far:
> 
> * `WLEDpixelBus.h` has duplicate copies of code from `colors.h`.  This component already isn't isolated from WLED internals (it's importing `const.h` for WLED's type mapping).  

This was actually intentional - I did try to keep current code & structure intact as much as possible, also to not change things and loose backwards compatibility. I am not saying it's the way it has to be in the end. 

>If we want to maintain the conceit that this is an independent library, pull out `colors.h` to its own "library" that both WLED and WLEDpixelBus can depend on.   (wled::colors ?? )  Otherwise we should use the definitions for `colors.h` directly.  (It makes sense in its way: one task performed here is to map between WLED's internal color format to the output pixel format.)

I did not build it up to be an independent library - but ultimately that should be the goal. I did focus on getting the drivers to work and somehow integrate them and less on making it as clean - as I am sure you can tell ;)

> * The split between `BusDigital` and `PixelBus` is costly -- calls like `setPixelColor` invoke two vtable lookups.  (And a third dynamic dispatch for the color encoding -- `switch` and `virtual` are conceptually similar.)   I don't have an immediate suggestion; this is mostly a performance tuning concern and we don't necessarily have to address it up front.   From an API perspective, the key change would be moving to a factory function (eg replacing `new BusDigital(...)` to `makeDigitalBus(...)`) -- this would let the `PixelBus` system someday return a fully qualified `Bus` type with fewer dynamic dispatches.

Agreed, the point from above applies here as well: I kept the old structure to not get tangled up and keep my focus on the pixel-pushing driver. I am not sure on where to make the split i.e. should the WLEDpixelBus replace bus digital or be more of an independent library.
edit:
it is still (way) faster than NPB - apart from the lower level driver being faster I think the main speedup is from not doing as many color conversions, those are expensive, the color order lookup is also much faster.

> * `CustomBusConfig` -- Good call to factor this separately from the common `BusConfig`, but directly embedding it in `BusConfig` gives up all the advantages.  I suggest using a type-erased bus-specific-settings slot in the `BusConfig` level (eg. instead of `CustomBusConfig custom`, do `std::shared_ptr<void> customBusSettings` or the like), and then cast it to the bus-specific struct during construction.  This lets each `Bus` implement its own custom settings struct and doesn't require sharing the type outside of the local bus implementation and local bus settings parser.  (The next step will be to factor the config parser on a per-`TYPE_` basis, leveraging `getLEDTypes()`, but that's a plan for another day.)

sounds like a good idea, I have no real grasp on that concept though - it's a bit above my every-day C++ knowledge but I am sure I can learn that abstraction level. 
I at one point had custom bus config as it's own config but then also wanted to be able to modify "standard" configs, i.e. take any of the supported bus formats and reorder the output format. If I understand correctly, your suggestion gets the best of both worlds.

Maybe we should have a call to discuss some of this?

DedeHai:
@willmmiles  one issue that remains is the default color order (I do have a solution for color order override but need to review and test it). I had an AI summarise the color orders used by the NPB buses in the original buswrapper. Its pretty inconsistent, assuming the summary is correct.
The question is: do we add a translation table so the new driver uses the same default color orders for the different LED types? It is not pretty carrying them over but otherwise users updating will have their color order changed, causing frustration.

This is what the AI spit out, I checked some of the values and the ones I checked were assessed correctly:

`AI START`
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-CH style='font-family:Calibri;font-size:11.0pt'>


<p style='margin:0in;font-family:Calibri;font-size:11.0pt' lang=de-CH>&nbsp;</p>

<p style='margin:0in;font-family:Calibri;font-size:11.0pt'><span lang=de-CH>WLED</span><span
lang=en-CH>’</span><span lang=de-CH>s UI/config default is always COL_ORDER_GRB
unless overridden by a color-order mapping. This is defined in
bus_manager.h:472 and const.h:394-399. With the legacy driver, PolyBus first
applies co, then NeoPixelBus applies the feature</span><span lang=en-CH>’</span><span
lang=de-CH>s native order in bus_wrapper.h:791-813.</span></p>

<div style='direction:ltr'>


WLED type | NeoPixelBus feature | Native order with default co=GRB
-- | -- | --
TYPE_WS2812_RGB | NeoGrbFeature | GRB
TYPE_WS2812_1CH_X3 | NeoGrbFeature | GRB   carrier; custom single-channel mapping
TYPE_WS2812_WWA | NeoGrbFeature | GRB   carrier; special WWA packing
TYPE_WS2811_400KHZ | NeoGrbFeature | GRB
TYPE_TM1829 | NeoBrgFeature | BRG
TYPE_UCS8903 | NeoRgb48Feature | RGB,   16-bit words
TYPE_UCS8904 | NeoRgbw64Feature | RGBW,   16-bit words
TYPE_SK6812_RGBW | NeoGrbwFeature | GRBW
TYPE_TM1814 | NeoWrgbTm1814Feature | WRGB
TYPE_APA106   ESP32 | NeoGrbFeature | GRB
TYPE_APA106   ESP8266 UART | NeoRbgFeature | RBG
TYPE_APA106   ESP8266 DMA/BitBang | NeoGrbFeature | GRB
TYPE_FW1906 | NeoGrbcwxFeature | G,R,B,CW,WW,0
TYPE_WS2805 | NeoGrbwwFeature | G,R,B,W1,W2   = GRB,WW,CW
TYPE_TM1914   ESP32 | NeoGrbTm1914Feature | GRB
TYPE_TM1914   ESP8266 | NeoRgbTm1914Feature | RGB
TYPE_SM16825 | NeoRgbcwSm16825eFeature | RGB,CW,WW,   16-bit words
TYPE_WS2801 | NeoRbgFeature | RBG
TYPE_APA102 | DotStarBgrFeature | BGR
TYPE_LPD8806 | Lpd8806GrbFeature | GRB
TYPE_LPD6803 | Lpd6803GrbFeature | GRB
TYPE_P9813 | P9813BgrFeature | BGR



</div>

<p style='margin:0in;font-family:Calibri;font-size:11.0pt' lang=de-CH>The exact
feature selections are in bus_wrapper.h:159-209 and bus_wrapper.h:281-315. The
NeoPixelBus feature definitions confirm the channel sequences, including
NeoBrgFeature, NeoGrbcwxFeature, and the SM16825 aliases.</p>

<p style='margin:0in;font-family:Calibri;font-size:11.0pt' lang=de-CH>TYPE_WS2812_1CH
and TYPE_GS8608 are declared but unused/unmapped in this legacy dispatch.
Analog, network, on/off, and HUB75 types do not use NeoPixelBus color-order
features.</p>

<p style='margin:0in;font-family:Calibri;font-size:11.0pt' lang=de-CH>For the
new driver, the existing TODO is accurate: WLEDpixelBus.cpp needs chip-specific
native-order handling. Pay special attention to TM1829, APA106, TM1914, FW1906,
WS2805, SM16825, WS2801, APA102, and P9813.</p>


</body>

</html>

`AI END`

willmmiles:
Hmm, that is tricky.

Probably what I'd suggest is to change the config schema (though I think we should consider that anyways).   If the new (and accurate) schema is present, use it; if only the old keys are present, adapt them using a compatibility table.

If it were up to me I'd get rid of the `TYPE` numeric mapping entirely in favor of driver names, and have the common "digital" api always accept raw timings and explicit color order (eg. a string "RGBWC0" or the like -- defining the whole mapping order).  The predefined types/timings table would be left to the UI.

DedeHai:
I like the idea. The LED types could be put into a LEDtypes.json file and served as a gzip to the UI to read, just like we do for the htm pages. New/custom types could live in userLEDtypes.json on the file system.

PR #5274, opened 2026-01-05
Implement combined factory bin creation with filesystem support

spiro-c:
This modified version of the tasmota [post_esp32.py](https://github.com/arendst/Tasmota/blob/development/pio-tools/post_esp32.py)  script for building combine binary file that is including `bootloader.bin` `partition.bin` `boot_app0.bin` `firmware.bin` and `littlefs.bin` if the ` custom_files_upload` is used to build the filesystem and uploaded on the esp
also there is helper script to pull the factory bin file in the `build_output\factory` folder 
I  test it on standard esp32 4M esp32-s3-8MB-8MB psram and esp32s3-8MB no psram wall was working good 
I also build the environments using all the esp32 `board_build.partitions` from default to extriime it correctly read the offsets 

## Summary by CodeRabbit

* **Chores**
  * Automated copying of factory firmware into build outputs after compilation.
  * New ESP32 combined-image generation: automatic flash-size detection, partition handling, and optional bundled filesystem.
  * Added a factory-focused build environment sample to simplify creating builds that include a LittleFS filesystem and custom post-build steps.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #5712, opened 2026-06-30
Add factory binary generation for release builds

netmindz:
Post-build script uses esptool merge_bin to produce a combined bootloader + partition table + firmware image (.factory.bin) alongside the existing OTA bin, enabling single-file flashing of blank devices. Only runs when platformio_release.ini is active (release builds).

## Summary by CodeRabbit

* **New Features**
  * Added automated creation of a factory firmware binary during the build process.
  * Release builds now use a clearer naming scheme for the generated output file.
* **Bug Fixes**
  * Improved handling of build offsets and flash parameters to make packaged firmware generation more reliable.
  * Added safeguards so the factory binary step is skipped on unsupported targets.

### DIFF
--- a/pio-scripts/factory_bin.py
+++ b/pio-scripts/factory_bin.py
@@ -1,0 +1,69 @@
+Import("env")
+import csv
+import json
+import os
+import subprocess
+
+OUTPUT_DIR = "build_output{}".format(os.path.sep)
+
+def _get_cpp_define_value(env, define):
+    define_list = [item[-1] for item in env["CPPDEFINES"] if item[0] == define]
+    if define_list:
+        return define_list[0]
+    return None
+
+def _get_app_offset(env):
+    try:
+        with open(env.BoardConfig().get("build.partitions")) as f:
+            for row in csv.reader(f):
+                row = [c.strip() for c in row]
+                if len(row) >= 4 and row[0] == "app0":
+                    return int(row[3], 16)
+    except Exception:
+        pass
+    return 0x10000
+
+def create_factory_bin(source, target, env):
+    if env["PIOPLATFORM"] == "espressif8266":
+        return
+
+    app_bin = os.path.normpath(str(target[0]))
+    chip    = env.get("BOARD_MCU")
+    flash_size = env.BoardConfig().get("upload.flash_size", "4MB")
+    flash_mode = env["__get_board_flash_mode"](env)
+    flash_freq = env["__get_board_f_flash"](env)
+    app_offset = _get_app_offset(env)
+
+    release_name_def = _get_cpp_define_value(env, "WLED_RELEASE_NAME")
+    if release_name_def:
+        release_name = release_name_def.replace("\\\"", "")
+        with open("package.json") as f:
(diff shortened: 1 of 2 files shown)


pio-scripts/factory_bin.py
```diff
@@ -0,0 +1,69 @@
...
+            for row in csv.reader(f):
+                row = [c.strip() for c in row]
+                if len(row) >= 4 and row[0] == "app0":
+                    return int(row[3], 16)
+    except Exception:
+        pass
+    return 0x10000
+
+def create_factory_bin(source, target, env):
+    if env["PIOPLATFORM"] == "espressif8266":
```
willmmiles, line 27: Might be better to run this gate the other way around (ie. ```if env["PIOPLATFORM"] != "espressif32": return```).

If we add support for any other platforms in the future, they'll almost certainly need a different approach.

pio-scripts/factory_bin.py
```diff
@@ -0,0 +1,69 @@
...
+
+    release_name_def = _get_cpp_define_value(env, "WLED_RELEASE_NAME")
+    if release_name_def:
+        release_name = release_name_def.replace("\\\"", "")
+        with open("package.json") as f:
+            version = json.load(f)["version"]
+        out_file = os.path.join(OUTPUT_DIR, "release", f"WLED_{version}_{release_name}.factory.bin")
+    else:
+        out_file = os.path.join(OUTPUT_DIR, "firmware", f"{env['PIOENV']}.factory.bin")
+
```
willmmiles, line 45: This code is duplicated from `output_bins.py`.  Normally I'd suggest making a common function to compute the filename, but I've found it kind of a pain to do with platformio scripts - it is possible to make a shared import from the `pio-scripts` folder, but it's a bit cumbersome; you have to play some games with the import paths.

The technique I ended up using with usermods was to store the relevant value in the environment.   (`WLED_MODULES` exists to pass the resolved usermod list from `load_usermods.py` to `validate_usermods.py`).

Probably the thing to do here is to have `output_bins.py` calculate the path at script execution time, and save it for usage in both action functions.

softhack007:
is this similar to
* #5274 ?

softhack007:
@coderabbitai please compare this PR to #5274. Are there things we can learn & improve when comparing both implementations?

pio-scripts/factory_bin.py
```diff
@@ -0,0 +1,69 @@
...
+def create_factory_bin(source, target, env):
+    if env["PIOPLATFORM"] == "espressif8266":
+        return
+
+    app_bin = os.path.normpath(str(target[0]))
+    chip    = env.get("BOARD_MCU")
+    flash_size = env.BoardConfig().get("upload.flash_size", "4MB")
+    flash_mode = env["__get_board_flash_mode"](env)
+    flash_freq = env["__get_board_f_flash"](env)
```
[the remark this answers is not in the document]
softhack007: Are these hooks available in pioarduino and newer  tasmota framework releases, as we use them in our `V5` branch?

PR #5713, opened 2026-07-01
Clean up setGeometry (main)

willmmiles:
Reorganize setGeometry to perform all checks first before applying the values.  This lets us accurately detect changes before starting a transition.

Forward port from #5474.

Sorry for the noise, I accidentally pushed this directly to main while setting up the branch. :(

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved segment geometry handling by sanitizing/clamping all geometry inputs (including 1D/2D bounds) before updating state, with additional safeguards for invalid configurations.
  * Enhanced “no-change” detection so geometry updates are skipped when nothing relevant changed.
  * Improved render/pixel buffer lifecycle: allocation is performed only when needed, and if allocation fails the segment is cleanly disabled to avoid inconsistent playback/rendering.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -484,7 +484,11 @@ void Segment::setGeometry(uint16_t i1, uint16_t i2, uint8_t grp, uint8_t spc, ui
 
   #ifndef WLED_DISABLE_2D
   if (Segment::maxHeight>1) { // 2D
-    if (i1Y >= Segment::maxHeight) i1Y = 0;   // Unlike i1, doesn't inherit old value
+    if (i1Y >= Segment::maxHeight) {
+      // Unlike i1 (X), Y values don't inherit old values if invalid
+      // This behaviour preserved for backwards compatibility
+      i1Y = 0;   
+    }
     if (i2Y > Segment::maxHeight) {
       i2Y = Segment::maxHeight;
     } else if (i2Y < 1) {
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -458,28 +458,33 @@ void Segment::handleRandomPalette() {
 // strip must be suspended (strip.suspend()) before calling this function
 // this function may call fill() to clear pixels if spacing or mapping changed (which requires setting _vWidth, _vHeight, _vLength or beginDraw())
 void Segment::setGeometry(uint16_t i1, uint16_t i2, uint8_t grp, uint8_t spc, uint16_t ofs, uint16_t i1Y, uint16_t i2Y, uint8_t m12) {
-  // Sanitise inputs  
+  // Sanitise inputs
   if (i2 <= i1) { // For any values, this means deactivate the segment; we check i2 before i1 for this case
     i2 = 0;
-  } else {
-    // Clamp i2 to maximum length
-    if (i2 > Segment::maxWidth*Segment::maxHeight) {
-      i2 = MIN(i2,strip.getLengthTotal());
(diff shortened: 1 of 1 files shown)


wled00/FX_fcn.cpp
```diff
@@ -458,83 +458,103 @@ void Segment::handleRandomPalette() {
...
-  // return if neither bounds nor grouping have changed
-  bool boundsUnchanged = (start == i1 && stop == i2);
+  // Sanitise inputs  
+  if (i2 <= i1) { // For any values, this means deactivate the segment; we check i2 before i1 for this case
+    i2 = 0;
+  } else {
+    // Clamp i2 to maximum length
+    if (i2 > Segment::maxWidth*Segment::maxHeight) {
+      i2 = MIN(i2,strip.getLengthTotal());
+    } else if (i2 > Segment::maxWidth) {
```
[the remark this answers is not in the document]
willmmiles: This is a legitimate concern and I'm working on it - it's not quite a trivial fix, there's a couple of chained concerns that follow.

wled00/FX_fcn.cpp
```diff
@@ -458,28 +458,33 @@ void Segment::handleRandomPalette() {
...
+      i2Y = 1;
+    } else if (i2 > Segment::maxWidth) {
+      i2 = Segment::maxWidth;
+    }
+  }
+
   #ifndef WLED_DISABLE_2D
   if (Segment::maxHeight>1) { // 2D
-    if (i1Y >= Segment::maxHeight) i1Y = startY;
+    if (i1Y >= Segment::maxHeight) i1Y = 0;   // Unlike i1, doesn't inherit old value
```
DedeHai: can you explain why one does inherit and the other does not? in what scenario does this even happen? I imagine if loading a preset with oob segment settings?

DedeHai (approved):

wled00/FX_fcn.cpp
willmmiles: Historical compatibility, ie. that's what the code prior to this PR did.  I agree it's a strange semantic.  Like so many API quirks I imagine it was an early implementation decision that stuck around.  The tweak here was because I misread the original implementation - in the original code, Y was always reset to 0 if invalid, while X keeps the old value.

I ended up with a little test script to validate that the API was consistent with main, except fixing the crash and notification bugs.  Sadly there is one incompatibility I left in here as I judged it wasn't worth the effort to fix: if you have a system with a matrix + trailing strip, and you send an update to a segment that was in the trailing strip with a trailing-strip stop, but matrix start index; then the original code would disable the segment, but this code keeps the old start value. 

eg. for a system with a 10x10 matrix followed by a 100 LED strip:
original segment: `{ "start": 120, "len": 60 }` (stop = 180)
is sent a bad update: `{ "start": 2 }`  (unlike every other value, omitting "len" doesn't keep the length the same, it keeps the stop coord the same...!!!)

The old code would produce:
`{ "start": 120, "len": 0 }` (disabled)
while the new code will still land at:
`{ "start": 120, "len": 60}`

DedeHai: thanks for the explanation, makes sense to me. Matrix with trailing strip is a bit of a special case I assume not many people use. 
could you please update the comment to ` // Unlike i1, doesn't inherit old value (legacy)` for clarity?

softhack007:
@coderabbitai is this PR compatible with #5729 ?

softhack007:
@willmmiles @DedeHai are we ready to merge?

DedeHai:
I have no objections.

Issue #5206, opened 2025-12-18
Bug in blending styles


DedeHai:
### What happened?

When any blending style other than "Fade" is used changing the brightness first turns all pixels black and then blends in with the selected style, using a fade for the brightness.

### To Reproduce Bug

Select any blending style other than "Fade", change the brightness.

### Expected Behavior

Ideally, the original segment should stay at the current brightness and the new segment should blend in with the new brightness. But at least it should not turn all pixels black and then blend the segment in with a fading brightness. This gets more tricky if multiple parameters change: for exampel a preset that changes the solid FX from one color to another and also changing brightness at the same time. 

### Install Method

Binary from WLED.me

### What version of WLED?

nightly build

### Which microcontroller/board are you seeing the problem on?

ESP32-C3

### Relevant log/trace output

```shell
-
```

### Anything else?

a simple solution would be to keep the "Fade" for brightness and just apply the blending in the same way as its done for an FX change or a color change i.e. simply dont turn the old segment to black.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

blazoncek:
The fix (if transition style is to be ignored during global brightness change) is rather simple. See my commit.

DedeHai:
thanks, this did the trick. now transitions work as intended and brightness is fading at the same time.

DedeHai:
can you please also add the same fix for 2D?

PR #5726, opened 2026-07-09
Fix whole-segment blackout frame at the start of on/off transitions with non-fade blending styles

npabisz:
## Symptom

With a transition time set and any blending style other than *Fade*, toggling power **off** causes a visible flick: the whole segment goes black for a single frame, jumps back to full brightness, and only then the transition animation plays. Toggling **on** has the same defect, but it is invisible in practice (the LEDs are still dark at that moment).

Reproducible on v16.0.1 and current `main` (ESP32, 313× WS2814, styles tested: Inside-out, Outside-in; transition 1.5–5 s). Not reproducible with *Fade*.

## Root cause

A race between the request context and the render loop.

`stateUpdated()` (led.cpp) sets `bri = 0` first and only calls `strip.setTransitionMode(true)` further down - after `notify()` / `sendSysInfoUDP()`, which can take several milliseconds of WiFi I/O. On ESP32 the render loop runs on a different task, so a frame can be serviced inside that window.

In that window the on/off workaround in `blendSegment()`:
```cpp
if ((briOld == 0 || bri == 0) && ((!clipped && (bri != briT) && !bri) || (clipped && (bri != briT) && bri))) c_a = BLACK;
```
already sees `bri != briT` (power change pending), but no segment transition exists yet, so `isPixelClipped()` returns `false` for every pixel. As a result the `!clipped && !bri` arm blanks every pixel of the segment for that frame. One frame later the transition has started and the animation begins from full brightness - hence the flick.

## Fix

Gate the blackout workaround on the segment transition actually having started (`topSegment.isInTransition()`), in both the 1D and 2D paths. Outside the race window the condition is unchanged: in steady state `bri == briT` (and `isPixelClipped()` needs an active transition anyway), and at the end of the transition `applyFinalBri()` makes `bri == briT`, so behavior there is identical.

## Testing

- ESP32 (esp32dev), 313× WS2814, single segment, transition 1.5–5 s.
- Before: dark flash right after turning off (Inside-out / Outside-in / Swipe).
- After: transition starts cleanly from the current frame; on/off, interrupted transitions and Fade behave as before.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LED transition handling so pixels are no longer briefly blanked before a segment transition actually starts.
  * Applied the fix consistently across both matrix and standard blending paths, resulting in smoother on/off brightness changes.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1642,7 +1642,9 @@ void WS2812FX::blendSegment(const Segment &topSegment) const {
         // workaround for On/Off transition
         // (bri != briT) && !bri => from On to Off
         // (bri != briT) &&  bri => from Off to On
-        if ((briOld == 0 || bri == 0) && ((!clipped && (bri != briT) && !bri) || (clipped && (bri != briT) && bri))) c_a = BLACK;
+        // note: only blank pixels once the segment transition has actually started; bri changes before
+        // startTransition() is called (stateUpdated()) and a frame rendered in that window would blank the whole segment
+        if (topSegment.isInTransition() && (briOld == 0 || bri == 0) && ((!clipped && (bri != briT) && !bri) || (clipped && (bri != briT) && bri))) c_a = BLACK;
       }
       // map it into frame buffer
       x = c;  // restore coordiates if we were PUSHing
@@ -1714,7 +1716,9 @@ void WS2812FX::blendSegment(const Segment &topSegment) const {
         // workaround for On/Off transition
         // (bri != briT) && !bri => from On to Off
         // (bri != briT) &&  bri => from Off to On
-        if ((briOld == 0 || bri == 0) && ((!clipped && (bri != briT) && !bri) || (clipped && (bri != briT) && bri))) c_a = BLACK;
+        // note: only blank pixels once the segment transition has actually started; bri changes before
+        // startTransition() is called (stateUpdated()) and a frame rendered in that window would blank the whole segment
+        if (topSegment.isInTransition() && (briOld == 0 || bri == 0) && ((!clipped && (bri != briT) && !bri) || (clipped && (bri != briT) && bri))) c_a = BLACK;
       }
       // map into frame buffer
       i = k; // restore index if we were PUSHing


softhack007:
> `stateUpdated()` (led.cpp) sets `bri = 0` first and only calls `strip.setTransitionMode(true)`

If the root cause is correct, would it be sufficient to change the order? Like `strip.setTransitionMode(true); bri = 0;` ?

npabisz:
> > `stateUpdated()` (led.cpp) sets `bri = 0` first and only calls `strip.setTransitionMode(true)`
> 
> If the root cause is correct, would it be sufficient to change the order? Like `strip.setTransitionMode(true); bri = 0;` ?

Unfortunately a reorder alone can't fully close the race, for two reasons:

1. `bri` isn't actually written inside `stateUpdated()` (my description was imprecise there - sorry). It is already changed by the *callers* before `stateUpdated()` is even entered: `toggleOnOff()` / `handleSet()` for the HTTP API, `deserializeState()` for JSON, nightlight, IR, UDP notifications, etc. `stateUpdated()` then relies on `bri != briOld` to decide whether a transition should be started at all, so "start the transition first, then change `bri`" would have to be implemented at every call site (or via a pending-brightness indirection). That changes the contract of the whole state-update path and touches many more places than this fix.

2. Even a perfect reorder only shrinks the window, it doesn't remove it: on ESP32 the render loop runs on a different task than the web/UDP handlers, so a frame can be serviced between *any* two instructions of the request handler. `deserializeState()` for example does quite a bit of work between writing `bri` and reaching `stateUpdated()`.

The blanking workaround in `blendSegment()` is the one place that *assumes* "a power transition is pending (`bri != briT`), therefore the segment transition has already started". The guard just makes that assumption explicit, so it holds no matter which caller changed `bri` and how long it takes to reach `startTransition()`.

That said, moving `strip.setTransitionMode(true)` earlier inside `stateUpdated()` - before `notify()`/`sendSysInfoUDP()`, which are the slowest part of the window - would be a reasonable additional hardening; on its own it would make the flash rarer, but not impossible.

DedeHai:
good catch, strange this went unnoticed for so long.  I checked and the bug was introduced with the fix for https://github.com/wled/WLED/issues/5206 before that commit it worked (but brightness change suffered from the same issue)

I confirmed the fix is correct and working, thanks!

DedeHai (approved):

blazoncek:
@npabisz thanks for digging into this. It baffled me for some time and couldn't pinpoint where the root cause was. Good job!

EDIT: BTW I have since rewritten On/Off transition handling in my fork.

PR #5734, opened 2026-07-13
prepare WS DDP for ESP8266 compatibility, fix compiler warnings in pacman FX, remove omggif.js for ESP8266

DedeHai:
- use smaller packet size in WS DDP send on ESP8266 so sending does not fail
- fix the flooding warnings in PacMan
- remove omggif.js from ESP8266 builds as it is not useable (no image support) - accommodated with fixes in pixelforge, this saves roughly 2.5k of flash.

note: pixelpaint and videolab tools will get an update too

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Pac‑Man effect stability by hardening position math and preventing out-of-range LED indexing, including stricter ghost/power-dot state handling.
  * Enhanced reset behavior when blue ghosts return to the start for more consistent gameplay.
  * Fixed DDP WebSocket streaming chunk sizing for ESP8266 devices to improve update reliability.
* **UI Improvements**
  * Updated PixelForge image/segment tab behavior to better handle ESP8266 limitations, including conditional resource loading and clearer availability messaging.
* **Compatibility**
  * GIF-related assets are now only served when GIF support is enabled.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -3150,8 +3150,8 @@ static const char _data_FX_MODE_ROLLINGBALLS[] PROGMEM = "Rolling Balls@!,# of b
 *   aux1 is the main counter for timing.
 */
 typedef struct PacManChars {
-  int32_t   pos;
-  int32_t   topPos;     // LED position of farthest PacMan has moved
+  int       pos;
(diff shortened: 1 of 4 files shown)


DedeHai:
cherry picked to 16_x

PR #5736, opened 2026-07-15
Create small default segment if out of heap

DedeHai:
Fallback path was missing. If out of heap and no segments are created the UI shows "no segments" error and disables most control elements which is confusing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved segment initialization when automatic segment creation fails.
  * Added a fallback to ensure a default segment is available when no active segments can be created.
  * Improved handling for matrix and one-dimensional LED configurations.


DedeHai:
cherry picked to 16_x

PR #5739, opened 2026-07-17
Fix a frame bug with the effect Stream that after a certain time will…

Ryrumeli:
Tested in my ESP32 Wroom.

Basically, there was an oversight where the 32 bits float is tested against a temporary 16 bit variable. As soon as it extrapolates (In the highest speed for the effect in about over an hour) there is an erratic shit that happens instead of the desired effect. Just applying a similar solution to the one in Stream 2 fixed it completely, by using & 0xFFFF.

Working Version:
ezgif-3633caa18a4d4298

Glitch happening:
ezgif-37a0ab7589585766

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved zone-change tracking in the Stream/random-chase lighting effect.
  * Kept zone transitions synchronized during longer-running animations.
  * Reduced the risk of unexpected or delayed pattern changes as the effect continues running.


wled00/FX.cpp
```diff
@@ -1198,7 +1198,7 @@ void mode_running_random(void) {
     z++;
   }
 
-  SEGENV.aux1 = it;
+  SEGENV.aux1 = it & 0xFFFF;
```
DedeHai, line 1201: this is redundant, aux1 is uint16_t

DedeHai:
thanks for finding and fixing this. looks good, the last line is unnecessary though.

softhack007 (approved):
Fix looks good to me.

As a small correction to the description: there is no `float` involved. The problem is a comparison `it  != aux1` that is performed in full 32bit, while `aux1` only has 16bit, and the effect behaves erratically after some time. The patch fixes it by masking out the upper 16bits so the comparison continues to work as expected.

DedeHai:
cherry picked to 16_x

PR #5744, opened 2026-07-18
Use global Buffer for HUB75

DedeHai:
Saves on RAM and is faster.
Dirty buffer is kept as it is faster if there are black pixels and not much slower if there are not.
Also changed PSRAM allocation rule to use PSRAM for large buffers which allows larger setups to run. Tested this part on S3 only using 128x128. 64x64 still uses DRAM mostly. Did not check what is possible on ESP32 with PSRAM.
Also fixed a bug perventing to actually use MAX_LEDS (off by 1)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved segment data handling to prevent stale values when memory is reused.
  - HUB75 matrix displays now update pixels immediately and more reliably.
  - Improved color-state reporting for HUB75 matrix pixels.
  - Added safer handling for large LED configurations on supported hardware.

- **Performance**
  - Improved memory allocation behavior for large buffers, helping maintain stability when PSRAM is available.

- **Documentation**
  - Clarified supported LED limits for ESP32-S3 devices.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -150,11 +150,9 @@ bool Segment::allocateData(size_t len) {
   if (len == 0) return false;    // nothing to do
   if (data && _dataLen >= len) { // already allocated enough (reduce fragmentation)
     if (call == 0) {
-      if (_dataLen < FAIR_DATA_PER_SEG) { // segment data is small
         //DEBUG_PRINTF_P(PSTR("--   Clearing data (%d): %p\n"), len, this);
         memset(data, 0, len);  // erase buffer if called during effect initialisation
         return true; // no need to reallocate
-      }
     }
     else
       return true;
@@ -198,7 +196,7 @@ void Segment::deallocateData() {
     DEBUG_PRINTF_P(PSTR("---- Released data (%p): inconsistent UsedSegmentData (%d/%d), cowardly refusing to free nothing.\n"), this, _dataLen, Segment::getUsedSegmentData());
   }
   data = nullptr;
-  Segment::addUsedSegmentData(_dataLen <= Segment::getUsedSegmentData() ? -_dataLen : -Segment::getUsedSegmentData());
+  Segment::addUsedSegmentData(-_dataLen); // addUsedSegmentData guards agains negative and sets zero if underrun so this is safe
   _dataLen = 0;
 }
 
--- a/wled00/bus_manager.cpp
+++ b/wled00/bus_manager.cpp
@@ -1033,7 +1033,6 @@ BusHub75Matrix::BusHub75Matrix(const BusConfig &bc) : Bus(bc.type, bc.start, bc.
     display->clearScreen();   // initially clear the screen buffer
     DEBUGBUS_PRINTLN("MatrixPanel_I2S_DMA clear ok");
 
-    if (_ledBuffer) d_free(_ledBuffer);                 // should not happen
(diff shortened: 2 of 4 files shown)


DedeHai:
@softhack007 is there a reson we do not use the NO_FAST_FUNCTIONS buildflag? see https://github.com/tidbyt/ESP32-HUB75-MatrixPanel-I2S-DMA/blob/master/doc/BuildOptions.md

softhack007:
> @softhack007 is there a reson we do not use the NO_FAST_FUNCTIONS buildflag? see https://github.com/tidbyt/ESP32-HUB75-MatrixPanel-I2S-DMA/blob/master/doc/BuildOptions.md

@DedeHai  I think we can enabled that, WLED-MM builds already use the NO_FAST_FUNCTIONS buildflag.

softhack007:
@DedeHai I'll take a closer look in the next days. 

Question: did you test with a larger panel, like 128x64 or 64x64? 

One reason for the HUB75 driver-level buffer was to prevent redraw flickering - if the show() loop ist faster than the screen draw "beam", we can be sure that tearing/flickering will only happen rarely. Maybe it's only visible with effects that have lots of motion and color changes.

The other reason - and we surely don't need that any more - was to support getPixelColor on the driver level.

DedeHai:
I did test on 64x64 and saw no tearing, but I also did not push it. Wheter the DMA driver is fed from global buffer or a local buffer should not make any difference should it? at the point where the (old) local buffer was pushed, the contents of that buffer and global buffer is identical and not changing. Or what am I missing that gets "out of sync"?

softhack007:
> Wheter the DMA driver is fed from global buffer or a local buffer should not make any difference should it?

True, that's not the difference. My concern was more on how fast can we update the DMA driver data? A local loop (in `hub75.show()`) that operates only on the local buffer is perfect for CPU caches, that's why this should be the fastest possible option. With wled v16 we already push all pixels down to the driver in one loop, so the difference is becoming very small - the remaining overhead may just be some function calls and the "find the right bus" loop in busmanager that runs on each pixel. Maybe the impact is not that severe any more. It was a huge problem in V0.14.x.

DedeHai:
valid point, I see the difference now. Could maybe grant show() access to the _pixels[] buffer (if it does not already have it) and push pixels from there. That would break ledmaps though.

DedeHai:
@softhack007 any proposal on how to test the possible frame tearing? I did not see any on 64x64 but have no larger panel to test.

softhack007:
@DedeHai my largest physical panel is 192x64 (i.e. 3 chained 64x64). I can check during this week if there is is any flickering/frame tearing added by this PR.

netmindz:
> @softhack007 any proposal on how to test the possible frame tearing? I did not see any on 64x64 but have no larger panel to test.

It would certainly be interesting to know if the issue with "stray" pixels is resolved with this. One of the existing issues was that if you used 3rd party software to send realtime data like scrolling text, you would sometimes get random pixels that were left still on

netmindz:
> > @softhack007 any proposal on how to test the possible frame tearing? I did not see any on 64x64 but have no larger panel to test.
> 
> It would certainly be interesting to know if the issue with "stray" pixels is resolved with this. One of the existing issues was that if you used 3rd party software to send realtime data like scrolling text, you would sometimes get random pixels that were left still on

Trying to find this issue it appears it might have been an issue with realtime rather than hub75, just more likely to be seen when higher pixel count

PR #5749, opened 2026-07-20
Fix bootloops, remove stray flag, rescue DeviceID and audioreactive V5 builds

DedeHai:
C3 and S3 would just bootloop. There are two issues:

- the ADC calibration used to "salt" the device hash is no longer available in V5 -  While it works on ESP32 and S2 it's best to remove it completely from V5 builds. The issue is that this pulls in the legacy driver which V5 has a guard for: if both legacy and new drivers are used, it will "abort" during boot.

edit: adding `-march=rv32imc` is not needed, it was related to local PIO breaking when switching from V4 to V5 without deleting the .pio folder:
- On the C3 there is a secondary issue: the build uses rv32imac by default for RISC V builds. The C3 being the first one using this architecture does not have the atromic hardware support (tha "a" in rv32imac). Using `-march=rv32imc` forces the compiler to use software "atomic" -> no more "invalid instruction"

I also removed the stray "-DCO" flags that have no meaning and are probably truncated `DCORE_DEBUG_LEVEL` flags

edit: also fixes bootloop on ESP32 when using AR

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ESP32-family build compatibility by removing an incorrect compile flag and applying explicit USB-mode settings for supported variants.
  * Updated ESP32-family device fingerprint generation to use the correct ADC-calibration approach based on the detected ESP-IDF version and chip type.
  * Enhanced ESP32 audio-reactive stability on ESP-IDF 5 with a targeted compatibility shim to avoid boot-time issues.


softhack007 (approved):
Looks good to me 👍 
I was always wondering what `-DCO` is about, it was probably an ancient copy&paste from tasmota

DedeHai:
@netmindz this changes the "device ID" for the stats, at least for C3 and S3 - just a heads-up

platformio.ini
```diff
@@ -385,11 +385,11 @@ build_flags = -g
   -DARDUINO_ARCH_ESP32
   -DARDUINO_ARCH_ESP32C3
   -DCONFIG_IDF_TARGET_ESP32C3=1
-  -DCO
   -DARDUINO_USB_MODE=1 ;; this flag is mandatory for ESP32-C3
   ;; please make sure that the following flags are properly set (to 0 or 1) by your board.json, or included in your custom platformio_override.ini entry:
   ;; ARDUINO_USB_CDC_ON_BOOT
   ${esp32_idf_V5.build_flags}
+  -march=rv32imc ; non-atomic RISC-V architecture (default is rv32imac, C3 does not support the 'a'tomic operations)
```
softhack007: IMC confirmed ✅
It's strange that pioarduino / esp-idf don't set this correctly ... in theorie this should hit any firmware that needs mutex/semaphores/queues/spin locks or any other atomic operation.

softhack007: @coderabbitai please verify where the -march= (or -mcpu=) parameter for -C3 comes from - pioarduino, arduino-esp32 v3, or esp-idf v5.

softhack007: @coderabbitai what's the purpose of the "_zicsr_zifencei" extensions?

willmmiles: I'm not sure what's up here?  I don't think we should need this.   Out of the box, `main` builds both C3 envs with `-march=rv32imc_zicsr_zifencei` for me.

I do need the `util.cpp` patch to avoid bootlooping, though, but that's all?

DedeHai: @willmmiles you are right, once I did a full re-install of PIO it works without `-march=rv32imc`  I need to check if it comes back if I switch to a V4 build

wled00/util.cpp
```diff
@@ -1343,7 +1339,7 @@ String generateDeviceFingerprint() {
   fp[0] ^= chip_info.revision | (chip_info.model << 16);
 #endif
 
-#if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32C3 || CONFIG_IDF_TARGET_ESP32S3
+#if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 0, 0) && (CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32C3 || CONFIG_IDF_TARGET_ESP32S3)
```
netmindz: This will mean that every unique deviceId will change. Is this really our only option? 

If so then we need to start rolling out a dual deviceid approach into current firmware so we can try and migrate

netmindz (changes requested):

wled00/util.cpp
DedeHai: I think we could go back to my initial approach of reading the values directly from the fuses - which was replaced with this as it is "dangerous low level stuff" and better to use a higher level API but since it is no longer available that would now be our only option.

softhack007: > If so then we need to start rolling out a dual deviceid approach into current firmware

I think our DeviceIDs will definitely change with V5 - the old API is gone; it gave us some calibration data to salt the DeviceID with. We cannot even re-calculate the old IDs in V5 firmware.

DedeHai: maybe I can salvage it, the ESP32 and the S2 still support the old API (at least in the V5 version we are using, AI said it got removed in 5.6 or something). For the C3 and S3 we may be able to replicate the value from the fuse value, need to check.

softhack007:
@DedeHai i've just fixed some merge conflicts with my latest platformio.ini changes, hope this doesn't screw up any uncommitted changes you may have.

DedeHai:
@softhack007 I have no uncommited changes yet. I am just running V5 on some of my test devices - C3 and S2 are completely screwed using RMT - not sure we are missing some parameters for NPB or if it just plain is not working. Symptoms: frames get pushed out while the last frame is still sending (FPS limiter disabled). When trying to save any LED settings, it hangs at `while(!canAllShow())`. S2 is slightly better than C3 it seems but can not save LED settings as well.

willmmiles:
> @softhack007 I have no uncommited changes yet. I am just running V5 on some of my test devices - C3 and S2 are completely screwed using RMT - not sure we are missing some parameters for NPB or if it just plain is not working. Symptoms: frames get pushed out while the last frame is still sending (FPS limiter disabled). When trying to save any LED settings, it hangs at `while(!canAllShow())`. S2 is slightly better than C3 it seems but can not save LED settings as well.

@DedeHai I'd suggest fixing up pixelbus and merging it instead of debugging NPB.   We can continue to iterate on it in main.

softhack007:
> - C3 and S2 are completely screwed using RMT

Wow, that's heavy. It might be due to using the "core3" neopixelbus branch that uses the new RMT channelizer API 🤔 

https://github.com/wled/WLED/blob/f7cff3db55dfbe4e2aec10a33186c130d212d58b/wled00/bus_wrapper.h#L254-L256

Also the V4 builds now seem to use NPB "core3", I'll try to separate that out so V5 and V4 can have different NPB versions. 

In principle - if nothing in wled references the new RMT api - we could even continue to use the "normal" NPB version with the old RMT driver.

Edit: or following the suggestion by @willmmiles, we wait a bit until WLEDpixelBus is ready for merging (must run on esp-idf V5 and V4 !), and then completely swap out NPB for all esp32 variants.

DedeHai:
> I'd suggest fixing up pixelbus and merging it instead of debugging NPB. We can continue to iterate on it in main.

exactly my thought, already updated the branch to main but oh so many compile errors ;)

DedeHai:
@softhack007 what is the reason we need everything to run on V4 still?

softhack007:
> @softhack007 what is the reason we need everything to run on V4 still?

Right now I still keep V4 as a "fallback" option, for back-to-back compare and for users that need features which are currently not working with V5 (analog mic, DMX serial, RMTHI flicker fixer, etc.). Its just a complication in platformio.ini, and some `#if ESP_IDF_VERSION ` guards.

If it gets too complicated to handle, we can drop the V4 support.

softhack007:
@DedeHai @netmindz as this is currently a show-stopper for S3 and C3 🤔 how about limiting the deviceID workaround to these two MCUs and then merge to main, and we'll try to rescue the old deviceID code in a separate PR?

It means we will collect some duplicate/bad IDs in the usage database. I guess there will just be a few reports that could be removed later, by searching for date + version. Additionally, nightly builds are off, and I hope that only a few users will install from `main`.

DedeHai:
@softhack007 I have a suggested fix from claude but did not check if it is legit and actually working.

DedeHai:
And I have the new WLEDpixelbus running on an S2 - RMT, I2S and BitBang appear to be working but RMT is quite fragile without the high priority interrupt: using more than two RMT outputs gives me horrible flickering but at least it works. I2S seems glitch free and bit bang as well so good progress on that front

DedeHai:
@softhack007 I don't know why it worked before but now I can not get classic ESP32 to boot, seems to be the same legacy adc driver conflict. If I do not compile AR UM it runs. `#include <driver/i2s.h>` pulls in the legacy driver. Does it boot for you?
edit: AI found a fix but I have no idea if it is legit: it adds dummy stubs to override the functions that pull in the legacy ADC but if this breaks something else in I2S audio input I don't know. Let me know if I should push it to this PR for testing.

softhack007:
Well, I got AR to compile with V5 yesterday - all V5 builds before  yesterday did not include AR. I need to cross-check. If driver/i2s.h is enough to create the legacy driver conflict, it means we need to exclude AR from esp32. I'll check during the day.

In fact classic esp32 is the only boards where i2s supported adc-over-i2s, so it makes sense that classic esp32 now has the adc driver conflict.

DedeHai:
@softhack007 if you can also check if S3 works, I also got a bootloop with AR but not sure I was not on the new LED driver branch which may not be fully up to date with main

DedeHai:
regarding device ID: I have a fix for the C3, ESP32 and S2 still work correctly using the deprecated function (it does not pull in the legacy driver) BUT the legacy adc stuff is no longer available in newer IDFs so I will try if I can do it without that function. S3 I could not get to produce the same ID as in 16.0. @netmindz how many S3 are in the stats and how bad would it be if those changed ID?

softhack007:
@DedeHai bootloop on classic esp32 confirmed - it happens when audioreactive is included

```gdb
Backtrace: 0x40082f81:0x3ffe3b30 0x4008d651:0x3ffe3b50 0x40090d89:0x3ffe3b70 0x4017a80f:0x3ffe3be0 0x4016e01e:0x3ffe3c00 0x40082501:0x3ffe3c30 0x400792bb:0x3ffe3c80 |&lt;-CORRUPTED

ELF file SHA256: b149e4e6e

Rebooting...
```

## Stack Trace
|  | Address | Function | Location|
|-- | -- | -- | --|
|0 | 0x40082f81 | panic_abort | ??|
|1 | 0x4008d651 | esp_system_abort | ??|
|2 | 0x40090d89 | abort | ??|
|3 | 0x4017a80f | check_adc_oneshot_driver_conflict | adc_legacy.c.obj|
|4 | 0x4016e01e | start_cpu0_default | startup.c.obj|
|5 | 0x40082501 | call_start_cpu0 | ??|

softhack007:
> @softhack007 if you can also check if S3 works, I also got a bootloop with AR but not sure I was not on the new LED driver branch which may not be fully up to date with main

S3 works for me, also with AR. But I'll still push a change that avoids including adc.h

bjoernh:
I hit also the boot loop. on C3 and classic ESP32, put the discovery (from claude) as a reference here:


Confirming this fixes the C3 here, and flagging that the classic ESP32 has a **second, independent path to the same `abort()`** that this PR doesn't cover.

Re: "While it works on ESP32 and S2" — the `esp32dev` env still boot-loops with this patch applied, because `default_usermods = audioreactive` pulls in the legacy ADC driver via legacy I2S rather than via `esp_adc_cal`:

```
audio_reactive.cpp (i2s_set_clk)
  -> libdriver.a(i2s_legacy.c.obj)
  -> adc_i2s_deprecated.c.obj   (adc_set_i2s_data_source)
  -> adc_legacy.c.obj           (adc_common_gpio_init)
```

`adc_legacy.c.obj` is what carries the `check_adc_oneshot_driver_conflict()` constructor, so it aborts as soon as anything also links the oneshot driver — which Arduino's `analogRead()` always does. Removing the `esp_adc_cal` usage doesn't break this chain.

### Reproduction

Board: ESP32-D0WD-V3 rev v3.0, 4MB flash. Tree at `17.0.0-devV5` with this PR's `wled00/util.cpp` diff applied, `pio run -e esp32dev`:

```
rst:0x1 (POWERON_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
...
entry 0x40080554

abort() was called at PC 0x40150263 on core 0

Backtrace: 0x40082edd:0x3ffe3b30 0x4008d569:0x3ffe3b50 0x40090ca1:0x3ffe3b70 0x40150263:0x3ffe3be0 0x40142bd2:0x3ffe3c00 0x4008245d:0x3ffe3c30 0x400792bb:0x3ffe3c80 |<-CORRUPTED
```

147 crash cycles in 12 s. Decoded:

```
0x40082edd: panic_abort
0x4008d569: esp_system_abort
0x40090ca1: abort
0x40150263: check_adc_oneshot_driver_conflict at adc_legacy.c.obj
0x40142bd2: start_cpu0_default at startup.c.obj
0x4008245d: call_start_cpu0
```

i.e. it aborts in a C constructor during `start_cpu0_default`, before `app_main`. The linker map confirms both drivers are present (`adc_oneshot_new_unit` from `esp32-hal-adc.c`, `check_adc_oneshot_driver_conflict` from `adc_legacy.c.obj`, 88 references).

### Confirmation

Rebuilding `esp32dev` with `custom_usermods =` (audioreactive removed, nothing else changed) drops `adc_legacy.c.obj`, `i2s_legacy.c.obj` and the constructor from the link entirely, and the board boots normally:

```
rst:0x1 (POWERON_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
...
entry 0x40080554
Ada
```

One reset, zero aborts.

### Notes

- The middle link, `adc_i2s_deprecated.c.obj`, is the I2S-ADC path that only exists on classic ESP32, so I'd expect **S3 to be genuinely fixed by this PR** — this second issue looks ESP32-only. I have not verified S2/S3 on hardware.
- On the C3 the chain is the one you describe (`esp_adc_cal_characterize` -> `esp_adc_cal_common_legacy.c.obj` -> `adc_legacy.c.obj`), and this PR does fix it — verified on an ESP32-C3 rev v0.4: boot-loop before, clean boot after.
- Worth noting the C3 gives no diagnostic at all for this: the constructor's `ESP_EARLY_LOGE` is compiled out at the default log level and targets UART0, not USB CDC, so you only see a bare `abort()`.


Hope it is useful, otherwise I can delete it.

softhack007:
> Hope it is useful, otherwise I can delete it.

Thanks, this confirms our observation (The tech-babble from claude is not adding anything useful, unfortunately; esp-idf is written in plain C, there is no such thing as a "C constructor". The abort is simply from the esp-idf startup code that checks if a weak symbol is not NULL).

It means that for classic esp32 only, audioreactive does not work any more in V5. It needs a full port to the new APIs.  
Hint: don't ask any AI for attempting to port AR to V5, it will just burn through your tokens without any useful result.

DedeHai:
@softhack007  in general you can check the .map file: if it contains "adc_legacy" it wont boot and something pulled that in.
I will try and fix the "salt"ing without the use of calling adc calibration stuff so it won't break again immediately if we need to move to 5.6 for some reason.

DedeHai:
> It means that for classic esp32 only, audioreactive does not work any more in V5. It needs a full port to the new APIs.

@softhack007 eventually yes but as I noted above, I do have a "hack fix" but I have no hardware to test if the I2S mic still works. Since the other option is to disable AR completely, I will add it to this PR for you to test with actual audio hardware.

i.e. this:
>>AI found a fix but I have no idea if it is legit: it adds dummy stubs to override the functions that pull in the legacy ADC but if this breaks something else in I2S audio input I don't know. Let me know if I should push it to this PR for testing.

usermods/audioreactive/audio_reactive.cpp
```diff
@@ -5,8 +5,48 @@
 
 #include <driver/i2s.h>
 #include <driver/adc.h>
+#endif
+
+// AI: below section was generated by an AI
+#if defined(CONFIG_IDF_TARGET_ESP32) && defined(ARDUINO_ARCH_ESP32) && (ESP_IDF_VERSION_MAJOR >= 5)
```
softhack007: can be shortened a bit, because IDF_TARGET_ESP32 implies ARDUINO_ARCH_ESP32

`#if defined(CONFIG_IDF_TARGET_ESP32) && (ESP_IDF_VERSION_MAJOR > 4)`

softhack007:
> i.e. this:
> 
> > > AI found a fix but I have no idea if it is legit: it adds dummy stubs to override the functions that pull in the legacy ADC but if this breaks something else in I2S audio input I don't know. Let me know if I should push it to this PR for testing.

@DedeHai thanks, I'll check on my boards. I can do a quick test today, and tomorrow test with real I2S and PDM hardware.

Edit: first test without mic: SUCCESS (esp32 boot complete)

usermods/audioreactive/audio_reactive.cpp
```diff
@@ -5,8 +5,48 @@
...
+  return ESP_ERR_NOT_SUPPORTED;
+}
+
+// Stub: would acquire ADC1 for DMA (built-in ADC mode only). Report failure if ever called.
+extern "C" esp_err_t adc1_dma_mode_acquire(void) {
+  return ESP_ERR_NOT_SUPPORTED;
+}
+
+// Stub: would release the legacy ADC1 lock. No lock is ever taken, so nothing to do.
+extern "C" esp_err_t adc1_lock_release(int adc_unit) {
```
softhack007: @coderabbitai please crosscheck if this function is also used by arduino-esp32 `analogRead()` or `touchRead()`. Follow the call chains into esp-idf to be sure.

https://github.com/wled/WLED/blob/a8ae8afa6be0a4ad9f5c460333b74d0998d4802b/wled00/button.cpp#L116

https://github.com/wled/WLED/blob/a8ae8afa6be0a4ad9f5c460333b74d0998d4802b/wled00/button.cpp#L182

usermods/audioreactive/audio_reactive.cpp
```diff
@@ -5,8 +5,48 @@
...
+  (void)adc_unit;
+  return ESP_OK;
+}
 #endif
+// AI: end
```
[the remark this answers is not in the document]
DedeHai: so what would be the correct signatures? propose a fix.

DedeHai:
after a lot of trial and error I found a solution to preserve the device ID's on all platforms. It is quite an elaborate solution to dance around the different APIs but it works, I got the same ID on all devices I tested (about 8 of all varieties ESP32, S2, S3, C3). I do not have an S2 with the "version 1" ADC calibration so could not test that but I think it is implemented correctly, worst case we get a few S2 that change IDs. 
I also added calibration entropy for all new chips for a truly untraceable ID by just hashing 32 bytes of the efuse BLOCK2 which according to documentation may or may not contain a unique ID but for sure contains calibration data (except for the S31 where this is not documented yet). Getting it directly from the fuses this time to not repeat the same mistake. @willmmiles what's your opinion on the get_calib_entropy() function? Can this swing around and bite us again in the future?

DedeHai:
@willmmiles after a long debug session I found the new compiler version is quite a bit different: I had flickering issues on BB output and it turned out that the IRAM function was inlined and put into flash. using `IRAM_ATTR __attribute__((noinline))` finally fixed the issue. This could be a problem throughout the codebase.

willmmiles:
> what's your opinion on the get_calib_entropy() function? Can this swing around and bite us again in the future?

Yes and no.  In an "is the design future proof" sense, then yes: doing our own low-level efuse reads and making whatever hash suits us best from it is guaranteed to be stable.   In an "is the code future proof" sense, I don't think ESP-IDF guarantees stability of the macro APIs for doing low-level access, so there'll always be the possibility that the next IDF version means we have to track down the new macro names.

From a code wrangling point of view, I'd say it's OK for now -- later we'll pull it in to the HAL.

> @willmmiles after a long debug session I found the new compiler version is quite a bit different: I had flickering issues on BB output and it turned out that the IRAM function was inlined and put into flash. using IRAM_ATTR __attribute__((noinline)) finally fixed the issue. This could be a problem throughout the codebase.

Oh, that'd do it!  Yes, IRAM and inlining have a bit of a history -- if I recall correctly, NPB has some comments in places about various less-than-ideal workarounds to get the right code in the right places.

That said, outside of our custom hardware drivers, we're usiing IRAM mostly as a performance tuning tool, and realistically we're going to end up re-doing a lot of the performance tuning for the new compiler anyways.  In most places, I expect that if the compiler can inline functions, it'll often be faster even if that code doesn't end up in IRAM.  The inlined code will be loaded from flash along with the rest of the surrounding function, so it'll "effectively" be IRAM but without the function call overhead .... right up to when the calling function gets to be so big it spans flash segments.   So "often" is doing all the heavy lifting there -- we'll have to benchmark things.

(This does point towards dropping the V4 envs sooner rather than later - trying to tune for both compiler versions might be tricky.)

softhack007:
> IRAM function was inlined and put into flash. using `IRAM_ATTR __attribute__((noinline))` finally fixed the issue

That's a very interesting one, well spotted 😃. I was always wondering what happens if the optimizer runs into "inline this function" together with "but put it into the IRAM section", Now we know - it seems that inlining "wins". The new compiler has LTO (link-time optimization) enabled by default, so i'd expect that _a lot_ more functions will become candidates for inlining, because the link-time optimizer sees the whole program.

DedeHai:
Once things are running more smoothly we should definitely revisit optimisations. 
I have a partially working PARLIO driver (AI converted the I2S driver still need to fully review and test), a somewhat working non legacy RMT driver, a slightly broken parallel SPI driver and a fixed and working parallel BB driver. I2S out works but is still legacy, did not test S3 LCD driver but it at least compiles. BTW C61 has no RMT, no PARLIO, no I2S.

Regarding this PR: please test some of your devices to confirm ID at /json/info is the same in 16.0, I could only test a few. It should be merged asap.

netmindz:
> Regarding this PR: please test some of your devices to confirm ID at /json/info is the same in 16.0, I could only test a few. It should be merged asap.

As you have tested and I trust your testing, confirmation by others is a nice to have not blocker to merge

PR #5752, opened 2026-07-21
remove unmaintained WLED_SAVE_RAM

softhack007:
On 8266, compiling with WLED_SAVE_RAM increases flash program size by 1.2KB, while reducing static RAM by 20 bytes only. Plus it reduces the size of the Segment::Transition struct by 48 bytes (from 80 to 32), by avoiding a separate palette for blending.

On the other side, the WLED_SAVE_RAM special code is not maintained (corner case, needs a special build flag to activate), and its largely untested with the latest codebase.

I suggest removing this micro-optimization, for the sake of maintainability, and for reducing special cases alternate implementations.

### DIFF
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -350,38 +350,6 @@ WLED_GLOBAL std::vector<WiFiConfig> multiWiFi;
 WLED_GLOBAL IPAddress dnsAddress _INIT_N(((  8,   8,  8,  8)));   // Google's DNS
 WLED_GLOBAL char cmDNS[33]       _INIT(MDNS_NAME);                // mDNS address (*.local, replaced by wledXXXXXX if default is used)
 WLED_GLOBAL char apSSID[33]      _INIT("");                       // AP off by default (unless setup)
-#ifdef WLED_SAVE_RAM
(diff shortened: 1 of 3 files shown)


DedeHai (approved):

wled00/wled.h
```diff
@@ -776,46 +682,13 @@ WLED_GLOBAL byte effectPalette _INIT(0);
...
-      Port = p1;
-      Port2 = p2;
-      RgbPort = p3;
-      NumRetries = r;
-      Connected = c1;
-      Connected2 = c2;
-      RgbConnected = c3;
-    }
-} __attribute__ ((aligned(1), packed)) udp_port_t;
-WLED_GLOBAL udp_port_t udp _INIT_N(({21234, 65506, 19446, 0, false, false, false}));
```
softhack007, line 802: Just to prove me right 😜 , this line actually has a bug - `21234` (udp port) should be `21324` (middle 2<->3 swapped)

willmmiles:
I'm late to the party on this one, but 100% agree - saving these few bytes are not worth the code duplication.

PR #5754, opened 2026-07-23
Add WS live stream

Antonin1106:
# Add new WebSocket Live Stream interface 
> This new feature provides low-latency LED streaming over WebSocket for environments where native real-time protocols such as DDP, E1.31, Art-Net, or DMX are unavailable or cannot be used (browser-based applications).

## How it work ?
The implementation reuses the existing WebSocket infrastructure to minimize changes and avoid duplicating logic.

1. Streaming data is received through the existing `wsEvent()` handler in [`ws.cpp`](https://github.com/Antonin1106/WLED/blob/main/wled00/ws.cpp).
2. A new function, `handleRawRGB()`, is called from a new `switch` case inside `wsEvent()`.
3. `handleRawRGB()` processes the incoming frame and updates the LED strip by calling `setRealtimePixel()`.

## Binary Protocol

Streaming uses a compact binary packet.

Each frame begins with a single command byte (`0x10`) followed by the RGB values of every LED, such as : `0x10 R G B R G B R G B ...`.
Each LED therefore occupies three bytes (1 per colors). 
For N LEDs, the packet size is `1 + (N × 3) bytes`.

### Example
For three LEDs (first red, second green, third blue), the transmitted bytes are `10 FF 00 00 00 FF 00 00 00 FF`.

> [!NOTE]
> RGBW is not currently supported.
>
> For RGBW LED strips, the white channel is automatically set to `0`.
> Sending RGBW data over this protocol will result in undefined behavior.

## Testing

- Verified to work as expected.
- Tested on an `esp32dev` board with a WS2812B LED strip.

> You can try it with this browser-based application : [WSync LED](https://github.com/Antonin1106/WSync-LED).

> [!NOTE]
> If this PR is accepted, we'll submit a follow-up PR to `wled-docs` documenting the protocol and its usage.

## Summary by CodeRabbit

- **New Features**
  - Added WebSocket realtime synchronization for incompatible broadcast sources.
  - Added support for receiving raw RGB data over WebSocket to directly control LED colors.
  - Documented browser-based WebSocket integration as an example use case.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -56,6 +56,7 @@ Originally created by [Aircoookie](https://github.com/Aircoookie), now maintaine
 - [Adalight / TPM2](https://kno.wled.ge/interfaces/serial/) (PC ambilight via serial)
 - [Infrared remote control](https://kno.wled.ge/interfaces/infrared/) (24-key RGB, receiver required)
 - Timers and schedules (NTP time sync, full timezone and DST support)
+- **WebSocket** realtime sync for other incompatible broadcast sources (browser-based applications such as [WSync LED](https://github.com/Antonin1106/WSync-LED)).
 
 ### Developer-Friendly
 - **Usermod system** — extend WLED with community or custom modules without modifying core code
--- a/wled00/ws.cpp
+++ b/wled00/ws.cpp
@@ -13,12 +13,39 @@ constexpr uint8_t BINARY_PROTOCOL_GENERIC = 0xFF; // generic / auto detect NOT I
 constexpr uint8_t BINARY_PROTOCOL_E131    = P_E131; // = 0, untested!
 constexpr uint8_t BINARY_PROTOCOL_ARTNET  = P_ARTNET; // = 1, untested!
 constexpr uint8_t BINARY_PROTOCOL_DDP     = P_DDP; // = 2
+constexpr uint8_t BINARY_PROTOCOL_RAW_RGB = 0x10; // WS reception
 
 static uint16_t wsLiveClientId = 0;
 static unsigned long wsLastLiveTime = 0;
 //static uint8_t* wsFrameBuffer = nullptr;
 
 #define WS_LIVE_INTERVAL 40
+
+void handleRawRGB(const uint8_t *data, size_t packetLength)
+{
+  realtimeLock(
+      realtimeTimeoutMs,
+      REALTIME_MODE_GENERIC);
+
+  if (realtimeOverride)
+    return;
+
+  // Number of leds
+  const uint16_t ledCount = min(strip.getLengthTotal(), (uint16_t)(packetLength / 3)); // R, G, B so /3
+
+  // Set all leds to the color received
+  for (uint16_t i = 0, o = 0; i < ledCount; i++, o += 3)
+  {
+    setRealtimePixel(
+        i,
+        data[o],
+        data[o + 1],
+        data[o + 2],
+        0);
+  }
+
+  e131NewData = true;
+}
 
 void wsEvent(AsyncWebSocket * server, AsyncWebSocketClient * client, AwsEventType type, void * arg, uint8_t *data, size_t len)
 {
@@ -94,7 +121,11 @@ void wsEvent(AsyncWebSocket * server, AsyncWebSocketClient * client, AwsEventTyp
             break;
           case BINARY_PROTOCOL_DDP:
             handleE131Packet((e131_packet_t*)&data[offset], client->remoteIP(), P_DDP, len - offset);
-        }
+          case BINARY_PROTOCOL_RAW_RGB:
+            realtimeIP = client->remoteIP();
+            handleRawRGB(&data[offset], len - offset);
+            break;
+          }
       }
     } else {
       DEBUG_PRINTF_P(PSTR("WS multipart message: final %u index %u len %u total %u\n"), info->final, info->index, len, (uint32_t)info->len);


DedeHai:
thanks for contributing. can you explain why not just use DDP?

softhack007:
@coderabbitai please compare this PR against the existing live preview implementation, which also uses websockets for communication.
https://github.com/wled/WLED/blob/a8ae8afa6be0a4ad9f5c460333b74d0998d4802b/wled00/ws.cpp#L186

https://github.com/wled/WLED/blob/main/wled00/data/liveviewws2D.htm

readme.md
```diff
@@ -56,6 +56,7 @@ Originally created by [Aircoookie](https://github.com/Aircoookie), now maintaine
 - [Adalight / TPM2](https://kno.wled.ge/interfaces/serial/) (PC ambilight via serial)
 - [Infrared remote control](https://kno.wled.ge/interfaces/infrared/) (24-key RGB, receiver required)
 - Timers and schedules (NTP time sync, full timezone and DST support)
+- **WebSocket** realtime sync for other incompatible broadcast sources (browser-based applications such as [WSync LED](https://github.com/Antonin1106/WSync-LED)).
```
softhack007, line 59: this should be put into the wled-docs online documentation, rather than the main README file.

Antonin1106: Ok, I understand
I'm making a md file for wled-docs

Antonin1106:
Hello @DedeHai !
The idea wasn't to replace DDP, but to provide a protocol specifically designed for WebSocket transport.

The protocol is intentionally minimal: a single command byte followed directly by the RGB data. While DDP's 10 bytes header is negligible over UDP, much of it (addressing, pixel offsets, fragmentation, sequencing... exists to compensate for UDP limitations. But with WebSocket the connection is already ok and each message is delivered reliably, in order, and as a complete frame. To avoid unnecessary network saturation, I also wanted to keep each WebSocket frame as small as possible, which is why the protocol only contains the data required to update the LEDs.

Since WebSocket already provides a reliable and ordered with message-oriented connection, many of the features implemented by DDP don't provide much value in this context. Instead, the protocol simply consists of a command byte followed by the RGB data, keeping each frame as small as possible to avoid unnecessary network overhead (30 fr/s !) and making the implementation very lightweight by reusing the existing WebSocket infrastructure.

If that's considered more valuable than keeping the protocol minimal, I'd be happy to adapt the implementation accordingly.
It was just an idea that I use with [WSync](https://github.com/Antonin1106/WSync-LED) on mobile browsers, it allows you to stream such as Prismatik or Hyperion directly on mobile.

softhack007:
> So the live-preview implementation itself is a complementary outbound observer, but the existing `sendDDP()` implementation is strong evidence that a new inbound raw protocol is unnecessary unless there is a clearly stated interoperability requirement that prevents the sender from emitting the small DDP header.

@Antonin1106 can you explain what your PR adds on top of the existing DDP-over-ws streaming feature? DDP itself is a rather lightweight protocol, that should be simple to implement in a browser (for example, we use it in the [PixelForge `videolab`](https://github.com/DedeHai/WLED-Tools/tree/main/WLED-VideoLab/Source) add-on tool);  [`<wled-ip>/common.js`](https://github.com/wled/WLED/blob/a8ae8afa6be0a4ad9f5c460333b74d0998d4802b/wled00/data/common.js#L216) already provides a JS helper that is ready-to-use for emitting the necessary headers.

Edit:

> keeping each frame as small as possible to avoid unnecessary network overhead (30 fr/s !) 

Just an idea - if your goal is to have an optimized stream: you may want to look into [RLE](https://en.wikipedia.org/wiki/Run-length_encoding), or other lightweight compression methods. RLE compression is very simple to implement; LEDs data usually has lots of repeating or similar colors, so i'd expect that ~50% stream compression can be achieved easily.

Antonin1106:
Thanks, I wasn't aware there was already DDP over WS support (is that documented on wled-docs ?). If that's the case, then my PR may indeed overlap with existing functionality.

My intention wasn't to introduce an alternative to DDP but to provide a minimal binary streaming interface over the existing WebSocket connection. The main advantage is its simplicity: clients don't need to implement the DDP protocol or construct DDP headers, they can simply send a buffer containing the RGB values (a very basic solution :)). I deliberately kept the protocol as simple as possible rather than introducing additional complexity for compression or other optimizations. I've also been using this protocol with other ESP-based projects that don't run WLED, where having such a simple interface has proven very convenient.

That said, if DDP overWS already covers this use case, I'd be interested to understand whether this implementation still brings any value.

DedeHai:
> Thanks, I wasn't aware there was already DDP over WS support (is that documented on wled-docs ?). If that's the case, then my PR may indeed overlap with existing functionality.

I mean its RIGHT THERE above the first line you added, it kind of tells me you did not look at the code and had an AI do this?

Antonin1106:
Sorry, but I did write the code myself.

Maybe I'm misunderstanding something: is it actually DDP over WebSocket, or is it DDP support exposed through the WebSocket interface? I'm still fairly new to LED streaming protocols and, to be honest, I've never worked with DDP before, so I may have misunderstood how the existing implementation works.

Sorry.... :)

Antonin1106:
But as you say, now I don't think this PR adds enough value on top of the existing DDP-over-WebSocket streaming feature. If the RAW RGB native streaming protocol is considered too close to DDP existing features, do you want to close the PR ?

softhack007:
@DedeHai I think the PR code is very simple, however appears quite fragile, and we already have a DDP helper for web clients.   Also there is no performance benefit for users (could be if RLE was implemented). 

Shortcomings I see with this PR:
 * most likely cannot handle multi-part messages, or more than 480 LEDs
 * No recovery from lost ws messages in a stream (will stay out-of-sync forever?)
 * Implementation assumes that the ws packet matches the physical LEDs count
 
So I'd suggest to close this PR, and improve the existing docs (wled-docs) instead.

wled00/ws.cpp
```diff
@@ -94,7 +121,11 @@ void wsEvent(AsyncWebSocket * server, AsyncWebSocketClient * client, AwsEventTyp
             break;
           case BINARY_PROTOCOL_DDP:
             handleE131Packet((e131_packet_t*)&data[offset], client->remoteIP(), P_DDP, len - offset);
-        }
+          case BINARY_PROTOCOL_RAW_RGB:
```
softhack007, line 124: critical bug: `break` missing before `case`.

softhack007: @coderabbitai why did this slip through your review?

DedeHai:
> Maybe I'm misunderstanding something: is it actually DDP over WebSocket

You did modify the ws.cpp which is all about websocket, anything in there is ws related. if you need an implementation example, check my videolab tool available in 16.0 through pixelfoge, see my wled tools repo.

PR #5758, opened 2026-07-28
add multiple meteors and ability to slow meteors

benjamw:
I really like the "Meteor" effect, and wanted to have multiple meteors going across my house. So I added a slider that adds meteors (up to 8, default 1).

The meteors also moved very quickly across my house, so I added a checkbox function to slow the meteors down (default off).

Built and tested locally and works well.

AI was used to assist me coding this.

## Summary

* **New Features**
  * Enhanced the Meteor effect to display multiple evenly spaced meteors.
  * Added a **Slow** mode for a more relaxed animation pace.
  * Added controls for the **# of Meteors** and **Slow** settings.

* **Improvements**
  * Improved meteor timing and trail behavior for smoother, more consistent rendering.
  * Adjusted animation handling for short segments to prevent overlapping meteor positions.
  * Preserved coordinated progression across all meteors.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -2424,12 +2424,11 @@ void mode_meteor() {
     unsigned start = (meteorstart + m * SEGLEN / numMeteors) % SEGLEN;
     for (unsigned j = 0; j < meteorSize; j++) {
       unsigned index = (start + j) % SEGLEN;
-      if(meteorSmooth) {
-          trail[index] = max;
-          uint32_t col = SEGMENT.check1 ? SEGMENT.color_from_palette(index, true, false, 0, trail[index]) : SEGMENT.color_from_palette(trail[index], false, true, 255);
-          SEGMENT.setPixelColor(index, col);
-      }
-      else{
+      if (meteorSmooth) {
+        trail[index] = max;
+        uint32_t col = SEGMENT.check1 ? SEGMENT.color_from_palette(index, true, false, 0, trail[index]) : SEGMENT.color_from_palette(trail[index], false, true, 255);
+        SEGMENT.setPixelColor(index, col);
+      } else {
         int idx = 255;
         int i = trail[index] = max;
         if (!SEGMENT.check1) {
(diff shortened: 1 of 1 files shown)


DedeHai:
thanks, sounds like good addition - similar to multi comet. You can also try PS Sparkler, it is different though but also can do something similar.

can you show what it looks like? there is an FX recorder tool in my wled tools repo.

benjamw:
FX_076(2)

Palette: Fire
Effect Speed: 23
Trail: 245
\# of Meteors: 128
Gradient: on
Slow: on
Smooth: off

FX_076(3)

Original for comparison:
FX_076(4)

wled00/FX.cpp
```diff
@@ -2417,29 +2419,31 @@ void mode_meteor() {
     }
   }
 
-  // draw meteor
-  for (unsigned j = 0; j < meteorSize; j++) {
-    unsigned index = (meteorstart + j) % SEGLEN;
-    if(meteorSmooth) {
+  // draw meteor(s), evenly spaced along the strip
+  for (unsigned m = 0; m < numMeteors; m++) {
+    unsigned start = (meteorstart + m * SEGLEN / numMeteors) % SEGLEN;
```
softhack007, line 2428: @coderabbitai i'm seeing a jump in the example shown by the PR author. Please check if the math is correct here.

benjamw: The jump may also be due to my failed attempt at trying to get a nice looping gif.

DedeHai:
I ran a test, the multi meteor breaks existing presets. Other than that it looks fine.

benjamw:
So the existing meteor presets break because unused custom values in the presets assume a default value of `[max]/2` which in this case is 128 for a 256-point slider.

I've added a secondary "marker" value that will tell us if it's a new preset or a legacy preset that needs to be migrated before displaying.

wled00/FX.cpp
```diff
@@ -2383,6 +2383,10 @@ void mode_meteor() {
   byte* trail = SEGENV.data;
 
   const unsigned meteorSize = 1 + SEGLEN / 20; // 5%
+  if (SEGMENT.custom3 != 1) {
+    SEGMENT.custom1 = 0;
+    SEGMENT.custom3 = 1;
```
DedeHai, line 2388: while this works, it is a hack that may bite us in the future. I don't have a better solution though. add comment so the intent is clear.

PR #5761, opened 2026-07-29
Don't restrict pins from non pin dropdowns in UM settings

### LINKED ISSUE #5759: UM settings non-pinlists still reserve pins
### What happened?

WLED Version 16.0.1

When I go to the Config/Usermods settings

and switch AGC from Normal to Vivid, my "Pin I2S WS" changes from 2 to "2 config".
This doesn't happen when I switch from Normal to Lazy and back to Normal.

If you then click "Save" after selecting "Vivid" and leaving the I2S WS pin set to "2 config," the I2S WS pin 2 disappears and is replaced by 32. Of course, the 32 then disappears from "I2S SCK pin" and is replaced by "unused."
If you then reboot and check the settings again, the “Type” is no longer set to “Generic I2S” but to “Generic PDM.”

This only happens if you configure “Generic I2S” as follows:

"Pin I2S SD": 12
"Pin I2S WS": 2
"Pin I2S SCK": 32

@dedehai was able to reproduce the issue in the #german channel on the WLED Discord server.

I'm using a Dig Uno V3.

Flashed under Linux:

Image
Image

sudo esptool.py erase-flash
sudo esptool write-flash 0x0 ./esp32_bootloader_v4.bin
sudo esptool.py write-flash 0x10000 ./WLED_16.0.1_ESP32.bin

sha256sum esp32_bootloader_v4.bin WLED_16.0.1_ESP32.bin
694d36d82db55b39060f2e0a2ab348d665d8a0d15659b2f5864f315b8d33eeb6  esp32_bootloader_v4.bin
628917b0753edcfc9a8408e6387c6d1ace6a360d315441e9563d60299fef8594  WLED_16.0.1_ESP32.bin

### To Reproduce Bug

Change Digmic Settings to: "Generic I2S"
"Pin I2S SD": 12
"Pin I2S WS": 2
"Pin I2S SCK": 32

Save, reboot, change to AGC Vivid.

### Expected Behavior

The "Pin I2S WS" setting does not change when I switch AGC from Normal to Vivid.

### Install Method

Binary from WLED.me

### What version of WLED?

16.0.1

ESP32

### Relevant log/trace output

```shell

```

### Anything else?

_No response_

### Code of Conduct

DedeHai:
Fix for #5759

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented non-pin controls from unintentionally changing pin selection options in the usermod settings interface.

### DIFF
--- a/wled00/data/settings_um.htm
+++ b/wled00/data/settings_um.htm
@@ -161,6 +161,7 @@
 	}
 	function UI(e) {
 		// update changed select options across all usermods
+		if (!e.classList.contains("pin")) return; // not a pin select, ignore. fix for #5759
 		let oldV = parseInt(e.dataset.val);
 		e.dataset.val = e.value;
 		let txt = e.name.split(":")[e.name.split(":").length-2];


softhack007 (commented):
looks good to me 👍 

maybe also check for "gpio" to keep support for the deep slepp usermod (if the rabbit statement is correct)
> One detail to handle: DeepSleep:gpio is a real GPIO dropdown, but its field name is gpio, not something containing pin. It currently does not receive the .pin class.

DedeHai:
the rabbit was wrong about deep sleep, I made it gpio intentionally exactly for the reason to be able to select a pin that is for example used by a button in normal use.

DedeHai:
I tested this using AR, multirely and deep sleep UMs and all dropdowns are now correct

netmindz:
Does this need porting back to 16_x ?

DedeHai:
it's already cherry picked.

PR #5774, opened 2026-08-06
Split udp.cpp into per-protocol files, encapsulate two low-fanout globals

netmindz:
## Summary

`udp.cpp` mixed at least 6 unrelated wire protocols in one 987-line file, mostly interleaved inside a single `handleNotifications()` dispatcher with nested early-returns: WLED's own proprietary "Notifier" state-sync protocol, node discovery, Hyperion (raw RGB), TPM2.NET, legacy UDP realtime (WARLS/DRGB/DRGBW/DNRGB/DNRGBW), the Art-Net/DDP/E1.31 *sender* (the receiver side already lives separately in `e131.cpp`), and the ESP-NOW transport used as an alternate carrier for the Notifier protocol.

This splits each concern into its own file, leaving `handleNotifications()` as a thin dispatcher:

- `sync_notifier.cpp` — `notify()` / `parseNotifyPacket()`, WLED's own state-sync protocol
- `sync_nodes.cpp` — node discovery (`refreshNodeList`, `sendSysInfoUDP`, node-info packet parsing)
- `realtime.cpp` — the shared realtime lock/unlock state machine (`realtimeLock`, `exitRealtime`, `setRealtimePixel`)
- `realtime_udp.cpp` — Hyperion / TPM2.NET / legacy UDP realtime packet parsing
- `realtime_broadcast.cpp` — outbound Art-Net/DDP/E1.31 sender (`realtimeBroadcast`)
- `espnow_sync.cpp` — ESP-NOW transport (`espNowSentCB`/`espNowReceiveCB`), which just reassembles fragments and hands off to `sync_notifier.cpp`'s `parseNotifyPacket()`
- `sync.h` — small constants shared across the above (`WLEDPACKETSIZE`, `UDP_SEG_SIZE`, `partial_packet_t`, ...)
- `udp.cpp` — `handleNotifications()` only, now ~85 lines

Every extraction preserves the original control flow exactly — no logic changes. Bare `return;` statements inside the old inline blocks map 1:1 to `return true;`/`return false;` in the new boolean-returning helper functions, matching the original branch semantics (verified by re-reading each original early-return path against its replacement).

### Also encapsulated two clusters of global state

Splitting the file made it possible to measure exactly how many other files reference each global that `udp.cpp` used. Two clusters turned out to be referenced *only* within this subsystem:

- `tpmPacketCount` / `tpmPayloadFrameSize` were referenced only in `realtime_udp.cpp` — now file-local `static` variables instead of `WLED_GLOBAL`.
- `notificationCount` / `notificationSentCallMode` / `notificationSentTime` were referenced only in `udp.cpp` + `sync_notifier.cpp` — now a `NotifierSendState` struct private to `sync_notifier.cpp`, exposed to `udp.cpp`'s dispatcher via a single `notifyRetryIfNeeded()` call instead of raw field access.

No behavior change intended anywhere in this PR — this is a pure reorganization plus removing state that didn't need to be global.

## Test plan

- [x] `nodemcuv2` (ESP8266): builds and links cleanly, including the `wled_espnow` usermod linking successfully against the relocated `espNowSentCB`/`espNowReceiveCB` — this exercises the ESP-NOW path even on an ESP8266 build.
- [ ] `esp32dev`: **not verified** — blocked by a sandbox networking issue (broken IPv6 route to a GitHub-hosted asset) unrelated to this change, not a compile problem. Opening as draft until this is confirmed on a real machine/CI.
- [ ] Runtime testing on hardware (WLED-to-WLED sync, Hyperion, TPM2.NET, Art-Net/DDP/E1.31, ESP-NOW sync) not performed — this PR is a pure refactor with no intended behavior change, but the protocol code is security/timing-sensitive enough to want a maintainer's eyes and CI's full build matrix before merge.

## Summary by CodeRabbit

- **New Features**
  - Added synchronization over ESP-NOW and UDP, including retries and packet validation.
  - Added automatic discovery and status updates for connected WLED devices.
  - Added support for Hyperion, TPM2.NET, WARLS, DRGB, DRGBW, DNRGB, and DNRGBW realtime protocols.
  - Added realtime broadcasting through DDP and Art-Net.
  - Improved realtime mode handling, including brightness restoration, timeouts, and RGBW pixel updates.
- **Bug Fixes**
  - Improved handling of fragmented, incomplete, unsupported, and out-of-order synchronization packets.


netmindz:
Would this kind of refactor be helpful for long term support?

willmmiles:
> Would this kind of refactor be helpful for long term support?

YES.  Thank you very much!  We might want to go one step further and file them in to a subfolder.

PR #5777, opened 2026-08-08
Encapsulate hue.cpp's own runtime state as static, not global

netmindz:
## Summary

11 `WLED_GLOBAL` variables declared in `wled.h` turned out to be referenced only in `hue.cpp` — nowhere else in `wled00/` or `usermods/` touches them: `hueXLast`, `hueYLast`, `hueHueLast`, `hueCtLast`, `hueSatLast`, `hueBriLast`, `hueLastRequestSent`, `hueAuthRequired`, `hueReceived`, `hueNewKey`, `hueClient`. They were global purely because everything in `wled.h` lives in one big `extern` block regardless of who actually uses it.

Converted all 11 to file-local `static` in `hue.cpp`, with the same types and initial values as before. `hueStoreAllowed`, which was declared on the same source line as `hueNewKey`, stays global since `set.cpp` also writes it — verified by grep before splitting that declaration.

No behavior change — purely a storage-class change, nothing else touched.

This is a small, mechanical first step from a broader pass identifying which of WLED's ~270 `WLED_GLOBAL` declarations are actually only used in one file (this is the largest single-file cluster).

## Test plan

- [x] `esp32dev`: builds and links cleanly via `pio run -e esp32dev` — 1,320,323 bytes flash, no warnings from either changed file.
- [x] Confirmed via repo-wide grep (`wled00/`, `usermods/`) that none of the 11 converted variables are referenced outside `hue.cpp`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Refactor**
  * Improved internal handling of Hue synchronization state.
  * Reduced exposure of Hue-related implementation details while preserving existing behavior.
  * Hue synchronization continues to operate without changes to its user-facing functionality or configuration.
  * Internal state management is now more clearly separated from publicly accessible application interfaces.

PR #5778, opened 2026-08-08
Encapsulate wled.cpp's own runtime state as static, not global

netmindz:
## Summary

Part of an ongoing pass identifying `WLED_GLOBAL` declarations that are actually only referenced in one file (see #5777 for the first of these). 8 more turned out to be private to `wled.cpp` itself: `ddp`, `e131` (the `ESPAsyncE131` sender objects), `lastMqttReconnectAttempt`, `lastReconnectAttempt`, `ledStatusLastMillis`, `ledStatusState`, `ledStatusType`, `udpRgbPort`.

Converted all 8 to file-local `static`:
- The three `ledStatus*` variables keep their original `#if defined(STATUSLED)` guard.
- `ddp`/`e131` use direct constructor syntax (`static ESPAsyncE131 e131(handleE131Packet);`), equivalent to what `WLED_GLOBAL ESPAsyncE131 e131 _INIT_N(((handleE131Packet)));` expanded to when `wled.cpp` itself was the definition translation unit (`WLED_DEFINE_GLOBAL_VARS`).

No behavior change — purely a storage-class change, same types and initial values as before.

## Test plan

- [x] `esp32dev`: builds and links cleanly via `pio run -e esp32dev` — 1,320,323 bytes flash, no warnings.
- [x] Confirmed via repo-wide grep (`wled00/`, `usermods/`) that none of the 8 converted variables are referenced outside `wled.cpp`.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal state management for MQTT and WiFi reconnections, UDP communication, lighting protocol handling, and status LED tracking.
  * No user-facing features or behavior changes were introduced.

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -25,6 +25,19 @@
 #endif
 extern "C" void usePWMFixedNMI();
 
+// Runtime state private to this file - previously WLED_GLOBAL, a leftover from
+// when all state lived in one big extern block regardless of who used it.
+static unsigned long lastMqttReconnectAttempt = 0;  // used for other periodic tasks too
+static unsigned long lastReconnectAttempt = 0;
+static uint16_t udpRgbPort = 19446; // Hyperion port
+static ESPAsyncE131 e131(handleE131Packet);
+static ESPAsyncE131 ddp(handleE131Packet);
+#if defined(STATUSLED)
+static unsigned long ledStatusLastMillis = 0;
+static uint8_t ledStatusType = 0; // current status type - corresponds to number of blinks per second
+static bool ledStatusState = false; // the current LED state
+#endif
+
 /*
  * Main WLED class implementation. Mostly initialization and connection logic
  */
--- a/wled00/wled.h
+++ b/wled00/wled.h
@@ -476,7 +476,7 @@ WLED_GLOBAL bool e131SkipOutOfSequence _INIT(false);              // freeze inst
 WLED_GLOBAL uint16_t pollReplyCount _INIT(0);                     // count number of replies for ArtPoll node report
 
 // mqtt
-WLED_GLOBAL unsigned long lastMqttReconnectAttempt _INIT(0);  // used for other periodic tasks too
+// lastMqttReconnectAttempt is private to wled.cpp - see there.
 #ifndef WLED_DISABLE_MQTT
   #ifndef MQTT_MAX_TOPIC_LEN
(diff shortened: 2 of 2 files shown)


wled00/wled.cpp
```diff
@@ -25,6 +25,19 @@
 #endif
 extern "C" void usePWMFixedNMI();
 
+// Runtime state private to this file - previously WLED_GLOBAL, a leftover from
+// when all state lived in one big extern block regardless of who used it.
+static unsigned long lastMqttReconnectAttempt = 0;  // used for other periodic tasks too
```
softhack007, line 30: @coderabbitai please double-check if any of these now static variables is used in any usermod/

wled00/wled.h
```diff
@@ -476,7 +476,7 @@ WLED_GLOBAL bool e131SkipOutOfSequence _INIT(false);              // freeze inst
 WLED_GLOBAL uint16_t pollReplyCount _INIT(0);                     // count number of replies for ArtPoll node report
 
 // mqtt
-WLED_GLOBAL unsigned long lastMqttReconnectAttempt _INIT(0);  // used for other periodic tasks too
```
willmmiles, line 479: We should update the agents file to avoid clutter comments like this -- there's no need to document private variables in global headers, even if they were misplaced there before.  It creates a maintenance burden that will inevitably fall out of date.

softhack007: @willmmiles I agree 👍

softhack007: implemented in b8233121f19bf930a7ad1ea4008970a9159e8949

willmmiles: Thank you!  That's maybe more specific to this specific task than I was thinking, but still good.

I'd been thinking something like:
"Document the code as it is, not as it was.  When refactoring, write as if the code has always been the better way.  We do not need to clutter the code with comments referencing past mistakes or describing why best practices are being followed.  Instead, comment when the code seems to go against best practices -- that's when the rationale is most important.  You do not need to 'leave tracks' in the code; good architecture makes relevant logic easy to find, and the commit history fills in the gaps."

softhack007: good points about "Dokument the code as it is, not as it was" and " comment when the code seems to go against best practices -- that's when the rationale is most important.". 

I'll think about how to integrate them. Actually, AIs are like children - if you talk too much, they get confused 😉. So i'd need something in the style of "do this, don't that".

PR #5796, opened 2026-08-10
Fixed nullptr crash in ETH IP comparing code

Tarik2142:
This PR fixes a nullptr crash when using ethernet.
The current implementation treats the passed value as a pointer, and since it is 0, it causes a crash.
image
Сomparing with `IPAddress()` is a more reliable option.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Ethernet static IP configuration validation.
  * Static IP and gateway settings are now applied only when valid network configuration data is available.

### DIFF
--- a/wled00/network.cpp
+++ b/wled00/network.cpp
@@ -295,7 +295,7 @@ bool initEthernet()
   }
 
   // https://github.com/wled/WLED/issues/5247
-  if (multiWiFi[0].staticIP != (uint32_t)0x00000000 && multiWiFi[0].staticGW != (uint32_t)0x00000000) {
+  if (multiWiFi.size() && multiWiFi[0].staticIP != IPAddress() && multiWiFi[0].staticGW != IPAddress()) {
     ETH.config(multiWiFi[0].staticIP, multiWiFi[0].staticGW, multiWiFi[0].staticSN, dnsAddress);
   } else {
     ETH.config(INADDR_NONE, INADDR_NONE, INADDR_NONE);


softhack007:
@Tarik2142 can you explain a bit more on which build this crash is happening? 
* WLED version
* using a "V4" build (16.0.1) or a "V5" (17.0.0-devV5) ?

The cast to (uint32_t) was added to enforce that the uint32_t overload of "==" gets used, this overload should also handle nullptr properly - at least in the new framework.

wled00/network.cpp
```diff
@@ -295,7 +295,7 @@ bool initEthernet()
   }
 
   // https://github.com/wled/WLED/issues/5247
-  if (multiWiFi[0].staticIP != (uint32_t)0x00000000 && multiWiFi[0].staticGW != (uint32_t)0x00000000) {
+  if (multiWiFi.size() && multiWiFi[0].staticIP != IPAddress() && multiWiFi[0].staticGW != IPAddress()) {
```
softhack007, line 298: It might be that the addition of `if (multiWiFi.size()` is the actual fix.

@coderabbitai how is `IPAddress()` defined? Is it the same as 0.0.0.0 ?

Tarik2142:
> @Tarik2142 can you explain a bit more on which build this crash is happening?
> 
> * WLED version
> * using a "V4" build (16.0.1) or a "V5" (17.0.0-devV5) ?
> 
> The cast to (uint32_t) was added to enforce that the uint32_t overload of "==" gets used, this overload should also handle nullptr properly - at least in the new framework.

I am working on a fork for ESP32-S3 + W5500 ethernet https://github.com/smlight-tech/A1-SLWF-09-WLED
And I found this crash on the S3 variant, I didn't check for the regular ESP32.
Crash dump looks like this:
image

I can say for sure that `multiWiFi.size()` is not the fix. I added that first but it didn't work. Then I replaced `(uint32_t)0x00000000` and it fixed.

wled00/network.cpp
softhack007: would `IPAddress(0)` be equvalent to `IPAddress()` ? I think it makes the intention clearer.

softhack007:
@coderabbitai please scan the WLED source code for other locations where an `IPAddress` object is directly compared to 0, [0,0,0,0], or 0U. List all locations that should get the same fix as proposed in this PR.

softhack007:
> wled00/network.cpp Line 789 compares uint32_t(multiWiFi[...].staticIP) and uint32_t(...staticGW) with 0U

@coderabbitAI can the condition also be expressed by comparing to `ipAddress(0,0,0,0)` ?

softhack007:
@coderabbitai are there recommendations in case we decided to use IPv6 (future support)?

softhack007:
cherry picked to 16_x

PR #5803, opened 2026-08-17
Add paired WS281x RGB+WW bus type

timle2:
## Summary
- add a new `WS281x RGB+WW Pair` digital LED type for strings where each logical RGBW bulb is backed by two WS281x pixels
- map logical RGB to raw physical `2n` and logical white to raw physical `2n+1`
- keep WLED effects, segments, presets, JSON API, and Home Assistant operating on 30 logical RGBW bulbs instead of 60 physical WS281x pixels

## Working Settings Used For Hardware Validation
Tested on Govee S14 Outdoor String Lights 2, two 15-bulb strings connected end-to-end:

```text
Controller: DOM-WLE-FM_A1.9 / ESP32 / 4 MB flash
Firmware target: ESP32 / esp32dev
LED type: WS281x RGB+WW Pair
Length: 30
LED GPIO: 16
Color order: BRG
Hardware Reversed: enabled for the tested installation orientation
Skip first LEDs: 0
Auto-calculate W from RGB: manual-only for RGB/manual-W tests; Auto Accurate for auto-white test
Relay GPIO: 12, rev:false on the tested custom-firmware config
```

A normal single segment should use logical bulb indexes:

```text
Segment start: 0
Segment stop: 30
Grouping: 1
Spacing: 0
Offset: 0
```

## Empirical Basis
The tested Govee S14 strings are not native one-address RGBW pixels. Each visible bulb is two sequential WS281x pixels. With WLED hardware `Reversed` off, raw physical order was verified as:

```text
raw physical 2n     = RGB side
raw physical 2n + 1 = warm-white side
```

Earlier stock-WLED workaround notes that described even indexes as warm-white were observed with WLED hardware `Reversed` enabled, which flips logical segment parity before writing to the strip. The new bus type keeps reverse operating on logical bulb order while preserving RGB/W order inside each two-pixel pair.

## How This Differs From RGBCCT/FW1906
The payoff of `WS281x RGB+WW Pair` is that WLED keeps one visible bulb as one logical RGBW pixel even though the string consumes two ordinary 3-channel WS281x addresses per bulb.

This PR exposes the tested fixture as 30 logical RGBW bulbs:

```text
logical bulb n RGB -> raw physical 2n
logical bulb n W   -> raw physical 2n + 1, written as W,W,W on the second RGB WS281x pixel
```

In plain terms, the warm-white side is driven by setting all three RGB channels equally on the second physical address. It is not a native WW/CW/CCT channel on one multi-channel LED IC.

| Area | Existing RGBCCT/FW1906-style types | `WS281x RGB+WW Pair` in this PR |
|---|---|---|
| Physical model | One native multi-channel IC per logical LED | Two sequential RGB WS281x addresses per logical bulb |
| WLED color model | RGB plus CCT/WW/CW handling | RGB plus one warm-white channel; no CCT channel |
| Driver/backend | RGBCCT-specific feature/protocol path | Existing 3-channel WS281x path, with pair mapping in `BusDigital` |
| `hasCCT()` | true for FW1906/WS2805/SM16825 | false |
| `hasWhite()` | true | true |
| Logical length for the tested S14 string | Not the intended abstraction for this hardware | 30, matching the visible bulbs |
| Physical allocation/current accounting | One native IC per logical LED | Doubled to 60 physical WS281x pixels |

FW1906 is the closest existing analogue because it also carries multiple color channels in one logical LED. I have not hardware-tested FW1906 as a stock workaround yet. From the current code path, though, FW1906 is modeled as RGBCCT (`hasCCT() == true`) and writes native RGB/WW/CW data. The S14 fixture tested here is not a native CCT IC: it is two separately addressable RGB WS281x pixels, one used for RGB and one used as the warm-white emitter. The goal of this PR is therefore not to add another RGBCCT protocol, but to keep effects, palettes, presets, JSON API, and Home Assistant operating on the visible bulb count while the bus layer expands each logical RGBW bulb to two physical pixels.

That distinction also explains the ABL changes. ABL is WLED's automatic brightness/current limiting. Existing digital bus types still report their normal `_len` from `getPhysicalLength()`. Only `TYPE_WS2812_RGBW_PAIR` reports `_len * 2`, because ABL standby/current accounting needs to count physical WS281x chips, not only logical bulbs. For all other digital types, the ABL math remains 1:1 with the previous bus length.

For PR #5704, I agree the new driver may be the better future path. I am treating that as the next plan after this PR-specific cleanup: evaluate whether the new custom channel mapping can represent one logical bulb as RGB plus W,W,W over two WS281x addresses, or whether the same pair-mapping concept still needs a small `BusDigital` layer there. This PR remains scoped to making the existing-driver implementation correct, narrow, and reviewable.

## Validation
- [x] Stock `esp32dev` build succeeded before changes
- [x] Prototype `esp32dev` build succeeded after changes
- [x] `npm test` passed
- [x] Manual W-only produced warm whitish-yellow output
- [x] RGB-only red, green, and blue produced correct colors with no warm-white contamination
- [x] RGB-only white stayed on the RGB side; on this fixture it appears cool blue/purple
- [x] Auto-white accurate mode made RGB white warmer by using the warm-white side
- [x] Rainbow effect ran smoothly across 30 logical bulbs with no alternating RGB/W artifacts
- [x] Chase effect ran as a clean red chase by visible bulb
- [x] Logical segment boundary `0-15` lit exactly 15 visible bulbs while `15-30` stayed off
- [x] Hardware reverse preserved RGB/W pairing for W-only and RGB-only output
- [x] Preset save/recall worked with a 30-bulb red chase test preset
- [ ] Boot preset behavior still needs paired-native validation
- [ ] Home Assistant behavior still needs validation

This remains a draft for maintainer/bot feedback while the remaining boot preset and Home Assistant checks are validated.

## Summary by CodeRabbit

* **New Features**
  * Added support for paired WS281x pixels that represent a single RGBW pixel.
  * RGB and warm-white channels are output separately and correctly reconstructed when read.
  * Physical pixel counts are now reflected in brightness limits, memory sizing, and current estimates.
  * Added LED type recognition and support across compatible hardware platforms.
  * Improved configuration, color ordering, reporting, and output reliability for paired RGBW arrangements.


timle2:
@coderabbitai review

timle2:
@coderabbitai review

timle2:
@coderabbitai review

softhack007:
@coderabbitai what is the difference between this PR and the RGBCCT bus types already supported in WLED?

https://github.com/wled/WLED/blob/8584eb3cf069448f368ffccdda433b9c69891b41/wled00/bus_manager.cpp#L363-L365

https://github.com/wled/WLED/blob/8584eb3cf069448f368ffccdda433b9c69891b41/wled00/bus_manager.cpp#L631

softhack007:
@coderabbitAI we are currently re-writing the bus driver; intending to replace NPB with our own driver. Is this PR compatible with https://github.com/wled/WLED/pull/5704 ?

wled00/bus_manager.cpp
```diff
@@ -1494,7 +1532,7 @@ void BusManager::initializeABL() {
       for (auto &bus : busses) {
         if (bus->isDigital() && bus->isOk()) {
           BusDigital &busd = static_cast<BusDigital&>(*bus);
-          uint32_t busLength = busd.getLength();
+          uint32_t busLength = busd.getPhysicalLength();
```
softhack007, line 1535: why? This seems to affect all busses, not just the new `paired` type.

@coderabbitai does this change create side-effects for other LED drivers?

wled00/bus_manager.cpp
```diff
@@ -1518,7 +1556,7 @@ void BusManager::applyABL() {
         if (_gMilliAmpsMax == 0)
           busd.applyBriLimit(0); // apply per bus ABL limit, updates _milliAmpsTotal if limit reached
         milliAmpsSum += busd.getUsedCurrent();
-        totalLEDs += busd.getLength(); // sum total number of LEDs for global Limit
+        totalLEDs += busd.getPhysicalLength(); // sum total number of physical LEDs for global Limit
```
softhack007, line 1559: same here - this change affects all LED drivers

DedeHai:
@softhack007 TYPE_FW1906 could be identical to this PR, did not check the details.
with my new driver, this is totally obsolete - support for new types is just a matter of defining the channels and a new type. It also supports reconfiguring channels for any of the LED protocols (currently up to 6 color channels supported, any order, any color, so you could do RRGGBB or RGB WW CW none etc. - this can be done in LED settings at runtime if a pre-defined type is missing. That (plus the flickering issue) was the whole point of the new driver.

timle2:
@softhack thanks for the feedback! 

I'll look at integrating with the new driver as an alternative pass on this next. Thank you for letting me know about that. 

Additionally, I'll do a pass on this PR to validate our changes are isolated to our WS281x RGB+WW Pair. This will be just to close out the PR with complete code, since it sounds like the new driver is true future path. 

As a side note - It's quite an interesting light in that it has two physical chips per bulb (my 30 lights have 60 addressable chips). There's a dedicated RGB led, then a second dedicated warm-to-cool light in each bulb. I think this is the first example in the configs of such a layout, but I could be wrong! Because this is a new layout style, it's inevitable that some code outside of the 'boxed' WS281x RGB+WW Pair config will be touched in this PR

timle2:
Thanks again for the review notes. I updated the PR body with a dedicated explanation of why this is separate from the existing RGBCCT/FW1906-style types.

Short version: `WS281x RGB+WW Pair` is not a native RGBCCT protocol. It is a logical-bulb mapping for the tested Govee S14 strings, where one visible bulb consumes two ordinary 3-channel WS281x addresses:

```text
logical bulb n RGB -> raw physical 2n
logical bulb n W   -> raw physical 2n + 1, written as W,W,W on the second RGB WS281x pixel
```

So WLED exposes 30 logical RGBW bulbs for effects/presets/API/HA, while the bus layer allocates and writes 60 physical WS281x pixels.

I also clarified the ABL/current-limit concern. Existing digital bus types still return `_len` from `getPhysicalLength()`. Only `TYPE_WS2812_RGBW_PAIR` returns `_len * 2`, because automatic brightness/current limiting needs to count physical WS281x chips for this paired layout. That should leave all current 1:1 digital bus behavior unchanged.

On FW1906 specifically: I agree it is the closest current analogue, but the current WLED path models it as RGBCCT/CCT (`hasCCT() == true`) with native RGB/WW/CW handling. I have not hardware-tested FW1906 as a stock workaround yet, so I am not claiming it cannot be made useful. The tested S14 behavior is simply different: two separately addressable RGB WS281x pixels, one RGB side and one warm-white side, presented as one logical RGBW bulb.

For PR #5704, I am keeping that as the next plan after this PR-specific cleanup. The question there is whether the new custom channel mapping can express one logical bulb as RGB plus W,W,W across two WS281x addresses, or whether the same small pair-mapping concept still belongs in `BusDigital` against the new driver API.

timle2:
@coderabbitai review

timle2:
@coderabbitai review

timle2:
@coderabbitai review

wled00/bus_manager.cpp
```diff
@@ -202,7 +207,12 @@ void BusDigital::estimateCurrent() {
   }
   // _colorSum has all the values of color channels summed, max would be getLength()*(3*255 + (255 if hasWhite()): convert to milliAmps
   uint32_t clrUnitsPerChannel = hasWhite() ? 4*255 : 3*255;
-  _milliAmpsTotal = ((uint64_t)_colorSum * actualMilliampsPerLed) / clrUnitsPerChannel + getLength(); // add 1mA standby current per LED to total (WS2812: ~0.7mA, WS2815: ~2mA)
+  uint16_t standbyMilliamps = getPhysicalLength(); // add 1mA standby current per LED to total (WS2812: ~0.7mA, WS2815: ~2mA)
+  if (_type == TYPE_WS2812_RGBW_PAIR && _milliAmpsPerLed < 255) {
+    actualMilliampsPerLed *= 2;
```
DedeHai, line 212: why?

softhack007:
@coderabbitai full review

PR #5804, opened 2026-08-18
fix(watchdog): use esp_task_wdt_config_t struct API for IDF 5.x

aenertia:
esp_task_wdt_init(timeout, panic) was removed in ESP-IDF 5.x. The new API takes an esp_task_wdt_config_t struct. Without this fix, the watchdog silently fails to initialise on IDF 5.x builds — the device will not reset on a hung loop task.

Fix: #if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 0) branch using esp_task_wdt_config_t, with fallback to the old two-argument API for IDF 4.x.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved watchdog compatibility and reliability on newer platform versions.
  * Added safer watchdog behavior, including configurable timeout and panic handling.
  * Ensured already active watchdogs are correctly updated when needed.
  * Preserved support for older platform versions.
  * Reduced unnecessary monitoring while maintaining expected system protection.

### DIFF
--- a/wled00/wled.cpp
+++ b/wled00/wled.cpp
@@ -361,7 +361,21 @@ void WLED::loop()
 #if WLED_WATCHDOG_TIMEOUT > 0
 void WLED::enableWatchdog() {
   #ifdef ARDUINO_ARCH_ESP32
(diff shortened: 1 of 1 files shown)


softhack007:
@coderabbitai full review

PR #5805, opened 2026-08-18
fix(segment): alloc-fill-swap-free in setName() to prevent race conditions between parallel tasks

aenertia:
Segment::setName() previously freed the old name buffer before allocating the new one. On ESP32 dual-core, the effects service loop (Core 1) reads segment.name while the main loop (Core 0) calls setName(). The free-then- alloc pattern leaves a window where name points to freed heap.

Fix: allocate the new buffer first, fill it, atomically swap the pointer, then free the old buffer. The effects loop always sees either the old valid pointer or the new valid pointer — never a freed one.

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal documentation around segment name updates and transition handling.
  * Preserved existing behavior when changing segment names, including retaining the current name if memory allocation fails.
  * Continued to support smooth display transitions while names are updated.

### DIFF
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -632,25 +632,12 @@ Segment &Segment::setName(const char *newName) {
   if (newName) {
     const int newLen = min(strlen(newName), (size_t)WLED_MAX_SEGNAME_LEN);
     if (newLen) {
-      // allocate and fill new name BEFORE freeing old to avoid race condition:
-      // the effect service loop (Core 1) may read SEGMENT.name while this runs (Core 0).
-      // Old code freed name first, leaving a window where name pointed to heap garbage.
       char *newBuf = static_cast<char*>(allocate_buffer(newLen+1, BFRALLOC_PREFER_PSRAM));
       if (newBuf) {
         strlcpy(newBuf, newName, newLen+1);
-        // start transition BEFORE swapping — the copy constructor deep-copies the current
-        // (still valid) name, so the old segment gets the correct previous text for blending.
-        if (mode == FX_MODE_2DSCROLLTEXT) startTransition(strip.getTransition(), true);
+        if (mode == FX_MODE_2DSCROLLTEXT) startTransition(strip.getTransition(), true); // if the name changes in scrolling text mode, we need to copy the segment for blending
         char *oldName = name;
-        name = newBuf;  // atomic pointer store — no torn reads on Xtensa
-        // Note: callers hold strip.suspend()+waitForIt() (see deserializeState),
-        // but waitForIt() has a timeout (see #4779).  If it expires while the
-        // effect loop is mid-read (e.g. mode_2Dscrolltext iterating name[]),
-        // freeing oldName here is a use-after-free.  This is a pre-existing
-        // upstream issue; alloc-before-swap narrows the window vs the old code
-        // (which freed first, leaving a dangling `name` pointer for the entire
-        // allocation+copy duration).  A proper fix requires waitForIt() to
-        // guarantee quiescence — tracked upstream as #4779.
+        name = newBuf;
         if (oldName) p_free(oldName);
       }
       return *this;
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -632,10 +632,27 @@ Segment &Segment::setName(const char *newName) {
   if (newName) {
     const int newLen = min(strlen(newName), (size_t)WLED_MAX_SEGNAME_LEN);
     if (newLen) {
-      if (name) p_free(name); // free old name
-      name = static_cast<char*>(allocate_buffer(newLen+1, BFRALLOC_PREFER_PSRAM));
-      if (mode == FX_MODE_2DSCROLLTEXT) startTransition(strip.getTransition(), true); // if the name changes in scrolling text mode, we need to copy the segment for blending
-      if (name) strlcpy(name, newName, newLen+1);
+      // allocate and fill new name BEFORE freeing old to avoid race condition:
+      // the effect service loop (Core 1) may read SEGMENT.name while this runs (Core 0).
+      // Old code freed name first, leaving a window where name pointed to heap garbage.
+      char *newBuf = static_cast<char*>(allocate_buffer(newLen+1, BFRALLOC_PREFER_PSRAM));
+      if (newBuf) {
+        strlcpy(newBuf, newName, newLen+1);
+        // start transition BEFORE swapping — the copy constructor deep-copies the current
+        // (still valid) name, so the old segment gets the correct previous text for blending.
+        if (mode == FX_MODE_2DSCROLLTEXT) startTransition(strip.getTransition(), true);
+        char *oldName = name;
+        name = newBuf;  // atomic pointer store — no torn reads on Xtensa
+        // Note: callers hold strip.suspend()+waitForIt() (see deserializeState),
+        // but waitForIt() has a timeout (see #4779).  If it expires while the
+        // effect loop is mid-read (e.g. mode_2Dscrolltext iterating name[]),
+        // freeing oldName here is a use-after-free.  This is a pre-existing
+        // upstream issue; alloc-before-swap narrows the window vs the old code
+        // (which freed first, leaving a dangling `name` pointer for the entire
+        // allocation+copy duration).  A proper fix requires waitForIt() to
+        // guarantee quiescence — tracked upstream as #4779.
+        if (oldName) p_free(oldName);
+      }
       return *this;
     }
   }


wled00/FX_fcn.cpp
```diff
@@ -632,10 +632,19 @@ Segment &Segment::setName(const char *newName) {
...
+      char *newBuf = static_cast<char*>(allocate_buffer(newLen+1, BFRALLOC_PREFER_PSRAM));
+      if (newBuf) {
+        strlcpy(newBuf, newName, newLen+1);
+        // start transition BEFORE swapping — the copy constructor deep-copies the current
+        // (still valid) name, so the old segment gets the correct previous text for blending.
+        if (mode == FX_MODE_2DSCROLLTEXT) startTransition(strip.getTransition(), true);
+        char *oldName = name;
+        name = newBuf;  // atomic pointer swap — effect now reads valid new name
+        if (oldName) p_free(oldName);
+      }
```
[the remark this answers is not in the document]
aenertia: Good catch on the analysis, but p_free(oldName) is safe here. After name = newBuf, no reader holds oldName — they all see newBuf. On Xtensa LX6, word-aligned pointer stores are single-instruction and atomic (no torn reads possible across cores). The old pointer is unreachable by any reader after the store completes.

The waitForIt() timeout concern is valid in general, but doesn't apply to this specific free — oldName is a different pointer from what name now points to. Even if isServicing() is still true and the effect loop reads SEGMENT.name, it reads newBuf, not oldName.

Added an explanatory comment at the p_free call site documenting the Xtensa atomicity guarantee. Pushed.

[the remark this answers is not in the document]
aenertia: You're right — I was wrong about the reclamation being safe. The pointer store is atomic (no torn reads), but a reader on Core 1 can load the old pointer into a local (e.g. strlen(SEGMENT.name) at line 6404) before the store, and then dereference it after p_free(oldName) completes on Core 0. Lines 6412, 6417, and 6456 iterate through indexed reads on that cached pointer.

The callers do hold strip.suspend() + strip.waitForIt() (see deserializeState() at json.cpp:472-473), but waitForIt() has a timeout (2 * getFrameTime() + 100 ms) — there's an upstream TODO noting this: // TODO: this needs a proper fix for timeout! see #4779. If it expires while the effect loop is mid-iteration through name[], the free is a use-after-free.

This is a pre-existing upstream issue — the old code was strictly worse (freed name first, then allocated and wrote the new buffer, leaving a dangling name pointer visible to readers for the entire alloc+copy duration). This PR narrows the window to only the timeout-expiry case, but doesn't eliminate it.

Updated the comment to document the residual hazard with a cross-reference to #4779. A proper fix requires waitForIt() to guarantee quiescence. Pushed.

DedeHai:
> main loop (Core 0)

main loop runs on core1 but wifi async task does not. there are plenty of these race conditions. what were the effects of this one you observed? the window given for the service() to finish is plenty so this race should be inexistent in practice - at least from my understanding.

aenertia:
Good catch on the core numbering — you're right, the main loop is Core
1, not Core 0. I had the cores backwards in the PR description. The
race window is between the JSON deserializer (which can run from the
async web server task on Core 0, or from UDP on the lwIP task) and the
effect service loop on Core 1.

You're right that the waitForIt() window is generous enough that this
is essentially theoretical on standard WiFi setups.

Where I hit it was on a PPP-over-serial transport I've been
developing. PPP runs its own lwIP task on Core 0 with different timing
characteristics to WiFi — the tcpip_thread mailbox contention pattern
is different, and waitForIt() timeouts were hitting more frequently
under sustained DDP + HTTP activity over the serial link. The visible
symptom was garbage appearing in the scrolling text effect — strings
like "null &&" rendering on-screen because mode_2Dscrolltext was
iterating through SEGMENT.name[] after the old buffer had been freed
and the heap had partially reused it. Once I switched to
alloc-before-swap the artifacts stopped completely.

The fix also corrects the startTransition() ordering — the old code
called it after the free but before strlcpy, so the Segment copy
constructor deep-copied an unfilled buffer into the old segment,
giving it garbage text for blending.

As the updated comment notes, a proper fix for the underlying
waitForIt() timeout issue is tracked in #4779 — this PR just narrows
the window for setName() specifically.

DedeHai:
we'd appreciate it if you would not use AI to communicate. Also your PR's have a stron AI smell without any AI markings: it's ok to remove AI markings if you reviewed each and every line but please cleanup the AI comments and write them in your own words, this goes for other PR's too. 

I am not sure what the plan is to untangle the race conditions and whether we should still do workarounds like this or if it will be tackled properly (as discussed in #4779)

aenertia:
Yup — everything I submit goes through a local Forgejo instance and
hardware test cycle before it reaches the GitHub mirror. From there it goes
into a tracking queue where I assess PR eligibility, and I go through each
submission with a fine tooth comb before opening it upstream.

Not trying to hide anything here — yes, I use LLM tooling in my workflow.
It's significantly better than my kak-hand at precision edits across a
codebase this size. I use a mix of frontier and local models alongside my
own code, and the output still goes through my review and hardware
validation before submission. The tooling helps me work faster and more
accurately, but the engineering decisions and testing are mine.

wled00/FX_fcn.cpp
```diff
@@ -632,10 +632,27 @@ Segment &Segment::setName(const char *newName) {
...
-      if (name) strlcpy(name, newName, newLen+1);
+      // allocate and fill new name BEFORE freeing old to avoid race condition:
+      // the effect service loop (Core 1) may read SEGMENT.name while this runs (Core 0).
+      // Old code freed name first, leaving a window where name pointed to heap garbage.
+      char *newBuf = static_cast<char*>(allocate_buffer(newLen+1, BFRALLOC_PREFER_PSRAM));
+      if (newBuf) {
+        strlcpy(newBuf, newName, newLen+1);
+        // start transition BEFORE swapping — the copy constructor deep-copies the current
+        // (still valid) name, so the old segment gets the correct previous text for blending.
+        if (mode == FX_MODE_2DSCROLLTEXT) startTransition(strip.getTransition(), true);
```
softhack007: original comment lost - please restore

wled00/FX_fcn.cpp
```diff
@@ -632,10 +632,27 @@ Segment &Segment::setName(const char *newName) {
   if (newName) {
     const int newLen = min(strlen(newName), (size_t)WLED_MAX_SEGNAME_LEN);
     if (newLen) {
-      if (name) p_free(name); // free old name
-      name = static_cast<char*>(allocate_buffer(newLen+1, BFRALLOC_PREFER_PSRAM));
-      if (mode == FX_MODE_2DSCROLLTEXT) startTransition(strip.getTransition(), true); // if the name changes in scrolling text mode, we need to copy the segment for blending
-      if (name) strlcpy(name, newName, newLen+1);
+      // allocate and fill new name BEFORE freeing old to avoid race condition:
+      // the effect service loop (Core 1) may read SEGMENT.name while this runs (Core 0).
```
softhack007: please remove these AI generated phantasies. I agree the fix would help, but the explaining comments are somewhere between questionale and plain wrong.

wled00/FX_fcn.cpp
```diff
@@ -632,10 +632,27 @@ Segment &Segment::setName(const char *newName) {
...
+      // Old code freed name first, leaving a window where name pointed to heap garbage.
+      char *newBuf = static_cast<char*>(allocate_buffer(newLen+1, BFRALLOC_PREFER_PSRAM));
+      if (newBuf) {
+        strlcpy(newBuf, newName, newLen+1);
+        // start transition BEFORE swapping — the copy constructor deep-copies the current
+        // (still valid) name, so the old segment gets the correct previous text for blending.
+        if (mode == FX_MODE_2DSCROLLTEXT) startTransition(strip.getTransition(), true);
+        char *oldName = name;
+        name = newBuf;  // atomic pointer store — no torn reads on Xtensa
+        // Note: callers hold strip.suspend()+waitForIt() (see deserializeState),
```
softhack007: same as above - remove the comment block (AI chatter), the code speaks for itself.

wled00/FX_fcn.cpp
```diff
@@ -632,10 +632,27 @@ Segment &Segment::setName(const char *newName) {
...
+        // but waitForIt() has a timeout (see #4779).  If it expires while the
```
softhack007: #4779 is not related to this problem.

wled00/FX_fcn.cpp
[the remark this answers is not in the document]
softhack007: @aenertia @coderabbitai no need to further discuss this. We know that there is still a potential that the main loop (strip.service) reads segment data while the webserver task (async_tcp) tries to update segments.

Its good to have some improvements, but we are just talking about one string that is read by a few effects only (scrolling text, image). So let's not over-fix it 😅 

A real solution would be to add semaphores around strip.service(). This is a planned improvement, but it goes together with a complete redesign of the segment architecture - will take some time until this lands in the code.

softhack007:
As a general comment - esp32 uses a realtime OS (FreeRTOS) - discussing "core0 vs. core1" scenarios is less than half of the truth. 

FreeRTOS can interrupt/change active tasks at any time. Having two cores is just a technical nuance, the real driver for race conditions is tasks switching performed by FreeRTOS.

softhack007:
> the tcpip_thread mailbox contention pattern is different, and waitForIt() timeouts were hitting more frequently under sustained DDP + HTTP activity over the serial link.

That makes sense, however in your case I'd suggest to 
* review all timeouts (`waitForIt()` but also other timeouts, for example in `requestJSONBufferLock()` or the UDP sync interval in udp.cpp, and timeouts in file.cpp)
* measure real delays that happen in your setup (some additional DEBUG_PRINT statements could be useful)
* adjust timeouts and check if problems remain

in a nutshell: understanding the problem's cause is better than fixing minor symptoms.

softhack007:
> [aenertia](https://github.com/aenertia) [force-pushed](https://github.com/wled/WLED/compare/4cfade9fc61a31838dc7089a05e912853e1d88d3..b639a52dc22c604093a2e1abc7af01b30cd21f30) the upstream-pr/segment-name-race-fix branch from [4cfade9](https://github.com/wled/WLED/commit/4cfade9fc61a31838dc7089a05e912853e1d88d3) to [b639a52](https://github.com/wled/WLED/commit/b639a52dc22c604093a2e1abc7af01b30cd21f30) 
[13 hours ago](https://github.com/wled/WLED/pull/5805#event-29650837895)

@aenertia important: _please_ _please_ stop using force-push while your PR is open under review. 

see https://github.com/wled/WLED/tree/main?tab=contributing-ov-file#updating-your-code

softhack007:
@coderabbitai full review

wled00/FX_fcn.cpp
```diff
@@ -632,10 +632,27 @@ Segment &Segment::setName(const char *newName) {
...
+        // effect loop is mid-read (e.g. mode_2Dscrolltext iterating name[]),
+        // freeing oldName here is a use-after-free.  This is a pre-existing
+        // upstream issue; alloc-before-swap narrows the window vs the old code
+        // (which freed first, leaving a dangling `name` pointer for the entire
+        // allocation+copy duration).  A proper fix requires waitForIt() to
+        // guarantee quiescence — tracked upstream as #4779.
+        if (oldName) p_free(oldName);
```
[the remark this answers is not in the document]
softhack007: @coderabbitai see my previous over-engineering comment.

In the current architecture, there are plenty of possibilities for creating use-after-free scenarios even with an atomic pointer swap. For example, the scrolling text effect may cache the pointer in a local variable and iterate over it, then the swap happens, and access via the old pointer reads from old memory locations. The only solution would be to implement a garbage collector - totally overkill for a few strings.

If this PR can reduce the likelyhood of strings appearing gabled because the memory is update by a parallel task, that is a good improvement. Going for 100% safe will need a lot more changes.

aenertia:
> > [aenertia](https://github.com/aenertia) [force-pushed](https://github.com/wled/WLED/compare/4cfade9fc61a31838dc7089a05e912853e1d88d3..b639a52dc22c604093a2e1abc7af01b30cd21f30) the upstream-pr/segment-name-race-fix branch from [4cfade9](https://github.com/wled/WLED/commit/4cfade9fc61a31838dc7089a05e912853e1d88d3) to [b639a52](https://github.com/wled/WLED/commit/b639a52dc22c604093a2e1abc7af01b30cd21f30)
> > [13 hours ago](https://github.com/wled/WLED/pull/5805#event-29650837895)
> 
> @aenertia important: _please_ _please_ stop using force-push while your PR is open under review.
> 
> see https://github.com/wled/WLED/tree/main?tab=contributing-ov-file#updating-your-code

Will do

wled00/FX_fcn.cpp
```diff
@@ -632,10 +632,14 @@ Segment &Segment::setName(const char *newName) {
...
-      name = static_cast<char*>(allocate_buffer(newLen+1, BFRALLOC_PREFER_PSRAM));
-      if (mode == FX_MODE_2DSCROLLTEXT) startTransition(strip.getTransition(), true); // if the name changes in scrolling text mode, we need to copy the segment for blending
-      if (name) strlcpy(name, newName, newLen+1);
+      char *newBuf = static_cast<char*>(allocate_buffer(newLen+1, BFRALLOC_PREFER_PSRAM));
+      if (newBuf) {
+        strlcpy(newBuf, newName, newLen+1);
+        if (mode == FX_MODE_2DSCROLLTEXT) startTransition(strip.getTransition(), true); // if the name changes in scrolling text mode, we need to copy the segment for blending
+        char *oldName = name;
+        name = newBuf;
+        if (oldName) p_free(oldName);
```
softhack007, line 641: The comment "free old name" is still lost.

softhack007 (approved):
Looks much clearer after removing all the AI chatter 👍 

I think it's good to merge now.

willmmiles (approved):

softhack007:
post-merge note: segment.name is also updated in the segment move- and copy-constructors. To be really thread-safe with exchanging the name string, it might be necessary to also change these to use the alloc-copy-swap pattern.

https://github.com/wled/WLED/blob/42da20ac3fe66a0ec5ee8d5228ba8368db8154f4/wled00/FX_fcn.cpp#L72

https://github.com/wled/WLED/blob/42da20ac3fe66a0ec5ee8d5228ba8368db8154f4/wled00/FX_fcn.cpp#L98

https://github.com/wled/WLED/blob/42da20ac3fe66a0ec5ee8d5228ba8368db8154f4/wled00/FX_fcn.cpp#L115

https://github.com/wled/WLED/blob/42da20ac3fe66a0ec5ee8d5228ba8368db8154f4/wled00/FX_fcn.cpp#L131

@coderabbitai please check if the segment constructors are "safe by usage pattern" considering the usage in `deserializeSegment()`, which indirectly activates segment via the `setGeometry()` method (json.cpp). `deserializeSegment()` is called from async_tcp context (webUI callback), while `strip::service()` runs in the loopTask context.

softhack007:
@aenertia if you still have problems in your low-bandwidth PPP-over-UART setup, I'd recommend increasing the timeout value in `WS2812FX::waitForIt()` - "100" is an empirical/measured value that works for WiFi, but might be too short for your use case. Alternatively, try to use fewer segments, for example `-D MAX_NUM_SEGMENTS=20 -DWLED_MAX_DIGITAL_CHANNELS=4`.

aenertia:
It's actually been pretty solid -i'm easily achieving 6000 + px runs
without touching segment setups so far. Most of the slow link issues are
only hit in streaming situations; so DDP rather than WebUI issues even,
56kbps baud seems to be relatively usable for full webui.

Given how many LORA devices are cheap and easily available + the various
meshnet projects are down around that rate, my primary USB/USB-CDC path is
easily serviceable in comparison.

WS/PPP interactions are a pretty niche case; and are still a bit
problematics - I'm not 100% sure why you would use WS DDP over the UDP path
that works pretty solidly for non broadband class link capacities anyway.

blazoncek:
FYI solution provided in this PR is still inadequate.
The (currently) only effect uses `SEGMENT.name` and index counter for access. This PR does not deal with OOB access if name changes to a shorter one and index is already past its length during execution.

aenertia:
Yeah the scrolling text FX as is - has caused me quite a few issues. I've actually got several separate implementations on the go atm to cover different display / use cases. Probably Having a Single Scrolling Text FX vs two or three variants with fewer knobs and i.e  'Wraping Static Text etc etc'' would be useful.  I for example I  started adding drop-shadow to it as is and quickly ran out structs to manipulate/overload without hitting a lot of quirks.

softhack007:
> FYI solution provided in this PR is still inadequate.
> The (currently) only effect uses `SEGMENT.name` and index counter for access. This PR does not deal with OOB access if name changes to a shorter one and index is already past its length during execution.

@blazoncek yes you're right, thanks for pointing this out. Especially scrolling text does lots of things in between accessing `SEGMENT.name`, so the likelihood of OOB access is still high. 

I think it will be safer to make "Scrolling Text" and "image" initially copy `SEGMENT.name` to a local char[] (on stack), and then only operate on the local copy. At least until we have a thread-safe redesign of segments that include semaphores or atomic access.

I'll tinker out a follow-up PR to at least make scolling text more robust 👍

blazoncek:
@softhack007 no need for elaborate or sophisticated approach. A single pointer backup and `if` comparing backup with name pointer is enough to break out of loop and stop processing text.
Making a name copy would only waste resources. After all it is only a read access and next frame will deal with new name.

However IMO this (root cause) is a very unlikely situation in any case. Possible, but unlikely.

willmmiles:
In the general case, it is never going to be safe to update live `Segment` objects while a render pass is in progress.   IMO there's no point in adding complex code to try to "fix" one site -- the issue will crop up somewhere else.

I merged the PR not because it closed all race conditions -- it does not -- but because it was a strict improvement over the previous code: ensuring the new allocation succeeded before releasing the old, and (admittedly) in the hopes of forestalling a bunch of well-meaning people trying to plug a leaky dam by sticking a finger in every hole.

In the context of the current code, the race should already be a difficult condition to trigger: the segment is already "locked" by `strip.suspend(); strip.waitForIt();` in the JSON path, and can't be modified in the legacy API.   It's not perfect but it gets us by.

I think we should instead be focusing our efforts on ensuring thread safety in a broad way: either globally by always handling state updates on the loop task (#4808), or by placing the live `Segment`s behind a stronger mutex.  The challenge with traditional mutexing is that the async web task is expected to be "nonblocking", so if we introduce a formal "render lock", it'd need to either be taken early in the handler processing and support back-off if unavailable (like the current JSON lock), or we'd need to be able to defer Segment changes to the loop task later (like `BusConfig` does for part of the hardware configuration).

As I'm sure you're all aware, I lean strongly towards the #4808 solution now as we have many, many places that are unsafe all over the code -- it's not just the members of `Segment`.

TL;DR: we really don't want to be trying to make every individual variable change thread safe -- it'd be effectively impossible to catch everything.

PR #5808, opened 2026-08-18
Potential fix(json): use Content-Length instead of chunked transfer for /json/effects and /json/fxdata

aenertia:
`respondModeData()` uses `sendChunked()`. The truncation is in ESPAsyncWebServer's `_ack()` — on the chunked path (`_sendContentLength=false`), `RESPONSE_WAIT_ACK` exits immediately because `!_sendContentLength` is unconditionally true. `close()` fires as soon as the final chunk is written to the lwIP send buffer, not when it's ACKed. On fast WiFi the write-to-ACK gap is negligible; on PPP/UART at 1.5Mbaud a 3KB response takes ~16ms to drain, and congested/distant WiFi retransmits create the same window.

`/json/effects` went through `LockedJsonResponse`, which holds the JSON buffer lock for the duration of the async send. Under concurrent WebSocket state pushes on constrained hardware, `sendDataWs()` blocks on the same lock.

Fix: two-pass Content-Length for both endpoints. `measureJSONStringElement()` pre-computes exact payload size; `request->send()` with that length sets `_sendContentLength=true`, so `RESPONSE_WAIT_ACK` gates on `_ackedLength >= _writtenLength` before close. `respondModeNames()` replaces the `LockedJsonResponse` path for `/json/effects` using the same streaming callback pattern as `respondModeData()` — no lock held during send, 256-byte stack buffer per callback.

The `_ack()` early-exit on chunked responses regardless of ACK state is an ESPAsyncWebServer bug — filing separately.

Tested on M5StickC (ESP32-PICO-D4) over PPP at 1.5Mbaud: `/json/fxdata` completes without truncation, `/json/effects` no longer deadlocks under concurrent WebSocket pushes. WiFi unaffected.

### DIFF
--- a/wled00/json.cpp
+++ b/wled00/json.cpp
@@ -1287,9 +1287,7 @@ static size_t writeJSONStringElement(uint8_t* dest, size_t maxLen, const char* s
   return 1 + n;
 }
 
-// Measure bytes that writeJSONStringElement would produce, without writing.
-// Used for Content-Length pre-computation to avoid chunked transfer truncation
-// on slow links (PPP/serial) where Connection:close races with final chunk.
+// Dry-run counterpart to writeJSONStringElement -- counts bytes without writing.
 static size_t measureJSONStringElement(const char* src) {
   size_t len = 1; // leading comma
   len += 1; // opening quote
@@ -1301,94 +1299,90 @@ static size_t measureJSONStringElement(const char* src) {
   return len;
 }
 
-// Two-pass streamed JSON response for mode data: first measures total payload
-// size, then sends with Content-Length via request->send().  The content callback
-// fills each TCP buffer in turn, using writeJSONStringElement for each entry.
+// Streamed JSON response for mode data. Two-pass: measure then send with Content-Length.
 void respondModeData(AsyncWebServerRequest* request) {
-  // Pass 1: measure total payload size so we can send with Content-Length.
-  // Pass 2: the content callback fills each TCP send buffer on demand.
   char lineBuffer[256];
-  size_t totalLen = 1; // ']' only — first element's comma becomes '['
+  size_t totalLen = 1; // ']' only -- first element's comma becomes '['
   for (size_t i = 0; i < strip.getModeCount(); i++) {
     strncpy_P(lineBuffer, strip.getModeData(i), sizeof(lineBuffer)-1);
     lineBuffer[sizeof(lineBuffer)-1] = '\0';
     if (lineBuffer[0] != 0) {
-      const char* dp = strchr(lineBuffer, '@');
-      totalLen += measureJSONStringElement(dp ? dp + 1 : "");
+      const char* dataPtr = strchr(lineBuffer, '@'); // Find '@', if there is one
+      totalLen += measureJSONStringElement(dataPtr ? dataPtr + 1 : "");
     }
   }
+
   size_t fx_index = 0;
-  bool firstEmitted = false;
   request->send(FPSTR(CONTENT_TYPE_JSON), totalLen,
-    [fx_index, firstEmitted](uint8_t* data, size_t len, size_t) mutable -> size_t {
+    [fx_index](uint8_t* data, size_t len, size_t) mutable -> size_t {
       size_t bytes_written = 0;
       char lineBuffer[256];
       while (fx_index < strip.getModeCount()) {
-        strncpy_P(lineBuffer, strip.getModeData(fx_index), sizeof(lineBuffer)-1);
+        strncpy_P(lineBuffer, strip.getModeData(fx_index), sizeof(lineBuffer)-1); // Copy to stack buffer for strchr
         if (lineBuffer[0] != 0) {
-          lineBuffer[sizeof(lineBuffer)-1] = '\0';
-          const char* dp = strchr(lineBuffer, '@');
-          size_t mode_bytes = writeJSONStringElement(data, len, dp ? dp + 1 : "");
-          if (mode_bytes == 0) break;
-          if (!firstEmitted) { *data = '['; firstEmitted = true; }
+          lineBuffer[sizeof(lineBuffer)-1] = '\0'; // terminate string (only needed if strncpy filled the buffer)
+          const char* dataPtr = strchr(lineBuffer,'@'); // Find '@', if there is one
+          size_t mode_bytes = writeJSONStringElement(data, len, dataPtr ? dataPtr + 1 : "");
+          if (mode_bytes == 0) break;  // didn't fit; break loop and try again next packet
+          if (fx_index == 0) *data = '[';
           data += mode_bytes;
           len -= mode_bytes;
           bytes_written += mode_bytes;
         }
         ++fx_index;
       }
-      if (fx_index >= strip.getModeCount() && len >= 1) {
+
+      if ((fx_index == strip.getModeCount()) && (len >= 1)) {
         *data = ']';
         ++bytes_written;
-        ++fx_index;
+        ++fx_index; // we're really done
       }
+
       return bytes_written;
   });
 }
 
-// Stream effect names as JSON array without holding the JSON buffer lock.
-// Eliminates the deadlock between /json/effects HTTP response (LockedJsonResponse
-// holds lock during async TCP send) and WebSocket state push (sendDataWs needs lock).
-// Pattern mirrors respondModeData() — zero heap allocation for response body.
+// Effect names only (no @-suffix data), without holding JSON buffer lock.
 void respondModeNames(AsyncWebServerRequest* request) {
-  // Two-pass: measure then send with Content-Length (same as respondModeData).
   char lineBuffer[256];
-  size_t totalLen = 1; // ']' only — first element's comma becomes '['
+  size_t totalLen = 1; // ']' only -- first element's comma becomes '['
   for (size_t i = 0; i < strip.getModeCount(); i++) {
     strncpy_P(lineBuffer, strip.getModeData(i), sizeof(lineBuffer)-1);
     lineBuffer[sizeof(lineBuffer)-1] = '\0';
     if (lineBuffer[0] != 0) {
-      char* dp = strchr(lineBuffer, '@');
-      if (dp) *dp = 0;
+      char* dataPtr = strchr(lineBuffer, '@'); // Find '@', if there is one
+      if (dataPtr) *dataPtr = 0;
       totalLen += measureJSONStringElement(lineBuffer);
     }
   }
+
   size_t fx_index = 0;
-  bool firstEmitted = false;
   request->send(FPSTR(CONTENT_TYPE_JSON), totalLen,
-    [fx_index, firstEmitted](uint8_t* data, size_t len, size_t) mutable -> size_t {
+    [fx_index](uint8_t* data, size_t len, size_t) mutable -> size_t {
       size_t bytes_written = 0;
       char lineBuffer[256];
       while (fx_index < strip.getModeCount()) {
-        strncpy_P(lineBuffer, strip.getModeData(fx_index), sizeof(lineBuffer)-1);
-        lineBuffer[sizeof(lineBuffer)-1] = '\0';
+        strncpy_P(lineBuffer, strip.getModeData(fx_index), sizeof(lineBuffer)-1); // Copy to stack buffer for strchr
         if (lineBuffer[0] != 0) {
-          char* dp = strchr(lineBuffer, '@');
-          if (dp) *dp = 0;
+          lineBuffer[sizeof(lineBuffer)-1] = '\0'; // terminate string (only needed if strncpy filled the buffer)
+          char* dataPtr = strchr(lineBuffer,'@'); // Find '@', if there is one
+          if (dataPtr) *dataPtr = 0;
           size_t mode_bytes = writeJSONStringElement(data, len, lineBuffer);
-          if (mode_bytes == 0) break;
-          if (!firstEmitted) { *data = '['; firstEmitted = true; }
+          if (mode_bytes == 0) break;  // didn't fit; break loop and try again next packet
+          if (fx_index == 0) *data = '[';
           data += mode_bytes;
           len -= mode_bytes;
           bytes_written += mode_bytes;
         }
         ++fx_index;
       }
-      if (fx_index >= strip.getModeCount() && len >= 1) {
+
+      if ((fx_index == strip.getModeCount()) && (len >= 1)) {
         *data = ']';
         ++bytes_written;
-        ++fx_index;
+        ++fx_index; // we're really done
       }
+
       return bytes_written;
   });
 }
--- a/wled00/json.cpp
+++ b/wled00/json.cpp
@@ -1287,38 +1287,108 @@ static size_t writeJSONStringElement(uint8_t* dest, size_t maxLen, const char* s
   return 1 + n;
 }
 
-// Generate a streamed JSON response for the mode data
-// This uses sendChunked to send the reply in blocks based on how much fit in the outbound
-// packet buffer, minimizing the required state (ie. just the next index to send).  This
-// allows us to send an arbitrarily large response without using any significant amount of
-// memory (so no worries about buffer limits).
+// Measure bytes that writeJSONStringElement would produce, without writing.
+// Used for Content-Length pre-computation to avoid chunked transfer truncation
+// on slow links (PPP/serial) where Connection:close races with final chunk.
+static size_t measureJSONStringElement(const char* src) {
+  size_t len = 1; // leading comma
+  len += 1; // opening quote
+  for (const char* p = src; *p; ++p) {
+    char esc = ARDUINOJSON_NAMESPACE::EscapeSequence::escapeChar(*p);
+    len += esc ? 2 : 1;
+  }
+  len += 1; // closing quote
+  return len;
+}
+
+// Two-pass streamed JSON response for mode data: first measures total payload
+// size, then sends with Content-Length via request->send().  The content callback
+// fills each TCP buffer in turn, using writeJSONStringElement for each entry.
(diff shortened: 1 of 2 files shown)


wled00/json.cpp
```diff
@@ -1287,38 +1287,110 @@ static size_t writeJSONStringElement(uint8_t* dest, size_t maxLen, const char* s
...
 // Generate a streamed JSON response for the mode data
 // This uses sendChunked to send the reply in blocks based on how much fit in the outbound
 // packet buffer, minimizing the required state (ie. just the next index to send).  This
 // allows us to send an arbitrarily large response without using any significant amount of
 // memory (so no worries about buffer limits).
 void respondModeData(AsyncWebServerRequest* request) {
+  // Two-pass: measure then send with Content-Length. Avoids chunked transfer
+  // truncation on slow links (PPP/serial) where Connection:close + TCP FIN
+  // races with final chunk delivery. Content-Length tells TCP the exact
+  // payload size, ensuring complete delivery before connection teardown.
```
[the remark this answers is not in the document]
aenertia: Right — the old block comment was leftover from the sendChunked implementation this PR replaces. Updated both the function-level and inline comments to describe the actual two-pass measure+send behaviour. Pushed.

wled00/json.cpp
```diff
@@ -1287,38 +1287,110 @@ static size_t writeJSONStringElement(uint8_t* dest, size_t maxLen, const char* s
...
         if (lineBuffer[0] != 0) {
-          lineBuffer[sizeof(lineBuffer)-1] = '\0'; // terminate string (only needed if strncpy filled the buffer)
-          const char* dataPtr = strchr(lineBuffer,'@'); // Find '@', if there is one
-          size_t mode_bytes = writeJSONStringElement(data, len, dataPtr ? dataPtr + 1 : "");
-          if (mode_bytes == 0) break;  // didn't fit; break loop and try again next packet
+          lineBuffer[sizeof(lineBuffer)-1] = '\0';
+          const char* dp = strchr(lineBuffer, '@');
+          size_t mode_bytes = writeJSONStringElement(data, len, dp ? dp + 1 : "");
+          if (mode_bytes == 0) break;
           if (fx_index == 0) *data = '[';
```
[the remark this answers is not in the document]
aenertia: Valid point. While mode 0 ("Solid") always has data in practice, using fx_index == 0 to detect the first emitted element is fragile if a leading entry were ever empty. Replaced with an explicit bool firstEmitted flag in both respondModeData() and respondModeNames(). Pushed.

softhack007:
@coderabbitai full review

softhack007:
@aenertia it might be a minor thing, but please teach your AI how to not delete existing source code comments.
Lost comments will make it harder for future maintainers to understand the source code. That's why we want them preserved, or adjusted. But not deleted.

wled00/json.cpp
```diff
@@ -1287,38 +1287,108 @@ static size_t writeJSONStringElement(uint8_t* dest, size_t maxLen, const char* s
...
       while (fx_index < strip.getModeCount()) {
-        strncpy_P(lineBuffer, strip.getModeData(fx_index), sizeof(lineBuffer)-1); // Copy to stack buffer for strchr
+        strncpy_P(lineBuffer, strip.getModeData(fx_index), sizeof(lineBuffer)-1);
         if (lineBuffer[0] != 0) {
-          lineBuffer[sizeof(lineBuffer)-1] = '\0'; // terminate string (only needed if strncpy filled the buffer)
-          const char* dataPtr = strchr(lineBuffer,'@'); // Find '@', if there is one
-          size_t mode_bytes = writeJSONStringElement(data, len, dataPtr ? dataPtr + 1 : "");
-          if (mode_bytes == 0) break;  // didn't fit; break loop and try again next packet
-          if (fx_index == 0) *data = '[';
+          lineBuffer[sizeof(lineBuffer)-1] = '\0';
```
softhack007: comment lost

wled00/json.cpp
```diff
@@ -1287,38 +1287,108 @@ static size_t writeJSONStringElement(uint8_t* dest, size_t maxLen, const char* s
...
+          const char* dp = strchr(lineBuffer, '@');
```
softhack007: not clear why you renamed "dataPtr" to "dp" ?

wled00/json.cpp
```diff
@@ -1287,38 +1287,108 @@ static size_t writeJSONStringElement(uint8_t* dest, size_t maxLen, const char* s
...
+          size_t mode_bytes = writeJSONStringElement(data, len, dp ? dp + 1 : "");
+          if (mode_bytes == 0) break;
```
softhack007: comment lost

wled00/json.cpp
```diff
@@ -1287,38 +1287,108 @@ static size_t writeJSONStringElement(uint8_t* dest, size_t maxLen, const char* s
...
-  request->sendChunked(FPSTR(CONTENT_TYPE_JSON),
-    [fx_index](uint8_t* data, size_t len, size_t) mutable {
+  bool firstEmitted = false;
+  request->send(FPSTR(CONTENT_TYPE_JSON), totalLen,
+    [fx_index, firstEmitted](uint8_t* data, size_t len, size_t) mutable -> size_t {
       size_t bytes_written = 0;
       char lineBuffer[256];
       while (fx_index < strip.getModeCount()) {
-        strncpy_P(lineBuffer, strip.getModeData(fx_index), sizeof(lineBuffer)-1); // Copy to stack buffer for strchr
+        strncpy_P(lineBuffer, strip.getModeData(fx_index), sizeof(lineBuffer)-1);
```
softhack007: comment lost

wled00/json.cpp
```diff
@@ -1287,38 +1287,108 @@ static size_t writeJSONStringElement(uint8_t* dest, size_t maxLen, const char* s
...
-        ++fx_index;        
+        ++fx_index;
       }
-
-      if ((fx_index == strip.getModeCount()) && (len >= 1)) {
+      if (fx_index >= strip.getModeCount() && len >= 1) {
         *data = ']';
         ++bytes_written;
-        ++fx_index; // we're really done
+        ++fx_index;
```
softhack007: comment lost

wled00/json.cpp
```diff
@@ -1287,38 +1287,108 @@ static size_t writeJSONStringElement(uint8_t* dest, size_t maxLen, const char* s
...
           data += mode_bytes;
           len -= mode_bytes;
           bytes_written += mode_bytes;
         }
-        ++fx_index;        
+        ++fx_index;
       }
-
-      if ((fx_index == strip.getModeCount()) && (len >= 1)) {
+      if (fx_index >= strip.getModeCount() && len >= 1) {
```
softhack007: why did you remove brace levels?

wled00/json.cpp
```diff
@@ -1287,38 +1287,108 @@ static size_t writeJSONStringElement(uint8_t* dest, size_t maxLen, const char* s
...
       char lineBuffer[256];
       while (fx_index < strip.getModeCount()) {
-        strncpy_P(lineBuffer, strip.getModeData(fx_index), sizeof(lineBuffer)-1); // Copy to stack buffer for strchr
+        strncpy_P(lineBuffer, strip.getModeData(fx_index), sizeof(lineBuffer)-1);
         if (lineBuffer[0] != 0) {
-          lineBuffer[sizeof(lineBuffer)-1] = '\0'; // terminate string (only needed if strncpy filled the buffer)
-          const char* dataPtr = strchr(lineBuffer,'@'); // Find '@', if there is one
-          size_t mode_bytes = writeJSONStringElement(data, len, dataPtr ? dataPtr + 1 : "");
-          if (mode_bytes == 0) break;  // didn't fit; break loop and try again next packet
-          if (fx_index == 0) *data = '[';
```
softhack007, line 1308: did you delete this line on purpose?

wled00/json.cpp
```diff
@@ -1287,38 +1287,108 @@ static size_t writeJSONStringElement(uint8_t* dest, size_t maxLen, const char* s
   return 1 + n;
 }
 
-// Generate a streamed JSON response for the mode data
-// This uses sendChunked to send the reply in blocks based on how much fit in the outbound
-// packet buffer, minimizing the required state (ie. just the next index to send).  This
-// allows us to send an arbitrarily large response without using any significant amount of
-// memory (so no worries about buffer limits).
+// Measure bytes that writeJSONStringElement would produce, without writing.
```
softhack007: is this function needed for "production" builds ? If not, maybe encapsulate it with `#ifdef WLED_DEBUG` ... `#endif`

aenertia:
Yeah - it's annoying I know, - there are multiple harness rules I have for
that already ; new project rules/subtlety of conventions, need a few PR's
to get it aligned. Since I have a bunch of stuff relating to the ppp
transport these first few small fixups will hopefully make future
submissions better aligned.

wled00/json.cpp
aenertia: measureJSONStringElement is critical for nondebug -- it computes the Content-Length for request->send(). Without it there's no way to know the payload size before sending.

softhack007:
> The final chunk arrives after the connection is torn down, truncating the response.

I have the feeling that this is not the root cause of the problem.

What @aenertia assumes is basically a violation of the TCP protocol, because a TCP connection is always held until the full communication is finished, or one of the enpoints prematurely aborts communication (`$ kill -9 -1`).  A broken connection would always happen also when using WiFi, however we never experienced that. 

My guess would be on the JSON buffer locking mechanism - `requestJSONBufferLock()` and `class LockedJsonResponse`. The JSON buffer is protected by a semaphore, but the semaphore timeout is relatively short, and the main JSON buffer is a ressource needed by many operations. It might be that another tasks takes over the buffer ownership too early, then initially clears it for another purpose. This scenario could also lead to a truncated response, especially on slow connections.

> json/effects routes directly to this handler without the shared JSON buffer lock.

without needing the shared lock, there is no contention on the lock any more. 

I'm not sure if my scenario is the correct explanation. But if it is, the whole PR could be reduced to just changing `serializeModeNames()` so it directly writes to `response`, instead of going via a JSON array. This would be faster and more memory-efficient, and the separate run to gauge the response size could be skipped, too.

aenertia:
TCP is correct — the protocol itself won't deliver FIN before data. The truncation is happening a layer up, in ESPAsyncWebServer.

In AsyncAbstractResponse::_ack() (WebResponses.cpp), the chunked path sets _chunked=true, _sendContentLength=false. When the final zero-length chunk fires, state goes to RESPONSE_WAIT_ACK. The exit condition there is !_sendContentLength || _ackedLength >= _writtenLength — since _sendContentLength is false, that's immediately true regardless of ACK state. close() fires as soon as the last chunk lands in the lwIP send buffer, not when it's ACKed.

The Content-Length path (_sendContentLength=true) gates that transition on _ackedLength >= _writtenLength — actual lwIP ACKs have to catch up before close. That's the window.

PPP makes it deterministic because ACKs transit the same UART byte stream as data — on a 3KB response at 1.5Mbaud there's ~16ms of UART drain time between write and ACK. I've also seen it on badly congested or distant WiFi where retransmissions push the ACK latency out far enough to hit the same window. Less frequent, but same root cause.

On the semaphore hypothesis — LockedJsonResponse::_fillBuffer() releases the lock as soon as serialisation is done, before the async send completes, so the buffer isn't held across the wire transfer. Lock elimination in respondModeNames() is worth keeping as a separate benefit, but it's not what's causing the fxdata truncation.

The actual bug I think, is in ESPAsyncWebServer — RESPONSE_WAIT_ACK should gate on _ackedLength >= _writtenLength for chunked responses the same way it does for Content-Length responses. This PR fixes it at the WLED level by switching to Content-Length, which is the right thing to do here regardless. I'll update the PR description to cite the actual mechanism rather than "TCP FIN races" — that framing was imprecise. Filing the ESPAsync root cause separately.

willmmiles:
I concur this problem analysis is AI crap.  LwIP holds the connection after `close()` until all bytes have been sent.  The closed TCP connection will be held in the TIME_WAIT state for up to 20 seconds -- this ought to be long enough to complete pretty much any transaction, especially given that the maximum window size is only two packets.   This is TCP 101, and frankly it's a waste of our time that we have to put up with this.

I'm also pretty confident in the JSON lock - the guard object guarantees that the lock isn't released until the transaction has been completely queued at the transport layer.  I've pretty thoroughly gone through all of the other use cases.

I certainly wouldn't rule out other web server bugs though.   A more productive approach might be to test with the upstream AsyncWebServer for comparison.   Ideally I'd like to migrate to the upstream fork at some point, but there's a fair amount of work to be PR'd there to make that possible (request queuing, memory safety on platforms that don't support C++ exceptions, and code size are all open concerns).

Re migrating more paths to the streaming approach: I'd been working on some general tooling for this to support all of the semi-static HTTP endpoints - sorry I've been slow getting this done.  There were still some issues with code size that I wanted to shake out before PR'ing.  ( https://github.com/willmmiles/WLED/tree/stream-json )

Re precomputing content length: IMO that's a nonstarter; the whole point of the chunked approach is to avoid that.  I would veto this PR on that basis alone.  CPU is cheap, but not that cheap.

aenertia:
The mechanism was wrong, I'll own that -- and appreciate the correction. The truncation is real and I'd been chasing the root cause before submitting — the explanation in the PR description was my best working theory at the time. Content-Length consistently avoids it (and I did try a number of approaches before the PR), sendChunked() doesn't, reproducible on both PPP and congested WiFi. I'll test against upstream AsyncWebServer and come back with something more concrete.

softhack007:
> Re precomputing content length: IMO that's a nonstarter; the whole point of the chunked approach is to avoid that. I would veto this PR on that basis alone. CPU is cheap, but not that cheap.

I agree with @willmmiles here - changing the /json/fxdata endpoint from chunked to content-length based transfer seems premature while a clear root cause is not yet established:

* Bug in AsyncWebser is not (yet) identified https://github.com/ESP32Async/ESPAsyncWebServer/discussions/472#discussioncomment-18121098
* it appears like the PPP-over-UART implementation from @aenertia has some potential defects https://github.com/wled/WLED/issues/5813#issuecomment-5386109207 that could cause the observed behaviour.



@aenertia I'm leaning towards closing this PR, because it tries to fix a problem where the root cause is still unclear. The core of this PR - changing from chunked to content-length based transfer - is not a good solution in general.

aenertia:
Yeah, fair -- I hadn't done enough isolation before writing this up although I thought I had, sorry about that.

The PPP patches are also rougher than I'd like, and am still working on them actively, so I can't rule out interplay -- #5813 may well be a PPP artifact. I'll get that sorted before pushing further here. Appreciate the patience. 

The lock-elimination piece -- softhack007 flagged that path earlier. Worth splitting standalone, or better to wait?

willmmiles:
> The lock-elimination piece -- softhack007 flagged that path earlier. Worth splitting standalone, or better to wait?

No need - the streaming JSON branch handles that as well as many other endpoints; it'd just create more merge conflicts.

aenertia:
This can be closed -- the Content-Length approach was wrong, as @willmmiles and @softhack007 correctly identified.

**Probable root cause found** -- I believe the issue is in how the `sendChunked` callback signals "buffer too small" vs "I'm done", but I haven't verified this against the upstream ESPAsyncWebServer fork (only against the WLED-bundled copy), so there may be nuance I'm missing.

ESPAsync's chunked response path in `_ack()` treats `readLen == 0` from `_fillBuffer` as "no more data, response complete" and transitions to `RESPONSE_WAIT_ACK`. But there's a separate return value for "I have more data but couldn't fit anything in this buffer": `RESPONSE_TRY_AGAIN` (`0xFFFFFFFF`). When the callback returns `RESPONSE_TRY_AGAIN`, ESPAsync retries on the next ACK when TCP send buffer space frees up.

Our streaming callback was returning 0 when the available buffer was too small to fit the current fxdata entry (some entries are 100+ bytes with slider/parameter metadata). On WiFi the TCP window is large enough that this never happens -- the buffer always fits at least one item. On PPP at 1.5Mbaud, the TCP send buffer (`TCP_SND_BUF=5744`) fills and ACKs arrive slowly via the same UART byte stream. When `space` drops below a single item size, the callback returns 0, ESPAsync ends the response, and `/json/fxdata` truncates at 6933B of 9208B.

I'm fairly confident in this analysis because:
- The truncation point is deterministic (always 6933B, consistent with TCP_SND_BUF=5744 plus HTTP framing on the first window fill-and-drain cycle)
- `curl` reproduces it (not browser-specific)
- WiFi never truncates (same firmware, same endpoint, larger TCP window)
- The one-line fix below resolves it completely on device (5/5 runs return full 9208B valid JSON over PPP, UI loads effect icons correctly)

What I haven't ruled out: whether there's also a PPP transport contribution (the UART RX path had real bugs i've since fixed -- `uart_write_bytes` corruption, LCP echo teardown -- and I can't fully isolate whether those interacted). The fix works with the current PPP code; I don't know if the old PPP code would have truncated even with this fix applied.

**Fix** (one line in the sendChunked callback adapter):

```cpp
// Before:
return r.count;

// After:
return (r.count == 0 && !r.done) ? RESPONSE_TRY_AGAIN : r.count;
```

This isn't PPP-specific -- any link slow enough to shrink the TCP window below a single fxdata entry size would hit it. PPP just makes it deterministic because data and ACKs share the same serial wire.

Crediting @willmmiles for correcting my earlier wrong theories about `tcp_close()` and AsyncTCP internals -- those corrections are what eventually pointed me toward the actual callback contract. The fix lives in my fork's streaming JSON adapter for now. @willmmiles' `stream-json` branch will likely supersede that code, but the `RESPONSE_TRY_AGAIN` pattern applies to any `sendChunked` callback that can legitimately produce 0 bytes without being done.

The lock-elimination piece from the earlier version of this PR is already covered by @willmmiles' streaming JSON work, so nothing from this PR needs to land separately.
````

PR #5814, opened 2026-08-23
Fix dynarray support for multiple arrays

willmmiles:
The dynarray linker script was incorrectly merging multiple arrays.   Use a strict alphanumeric sort to ensure entries are correctly ordered.

Note this adds a hard requirement that the numeric section keys must always be exactly 5 digits (ie. zero-padded).   Since this cannot be done with a macro, a validation is added instead.

(Also I once again incorrectly pushed directly to main - sorry for the noise!  Can we maybe disable that?)

## Summary by CodeRabbit

- **Improvements**
  - Improved reliability and consistency when registering dynamic-array entries and user modules.
  - Standardized five-digit ordering values to ensure entries are processed predictably.
  - Added C++ validation to detect invalid dynamic-array ordering values earlier.
  - Updated ordering behavior and documentation to use consistent lexicographic sorting.
  - Updated module validation to recognize the standardized dynamic-array boundaries.

### DIFF
--- a/pio-scripts/dynarray.py
+++ b/pio-scripts/dynarray.py
@@ -11,7 +11,7 @@ from pathlib import Path
 DYNARRAY_INJECTION = (
     "\n    /* dynarray: WLED dynamic module arrays */\n"
     "    . = ALIGN(0x10);\n"
-    "    KEEP(*(SORT_BY_INIT_PRIORITY(.dynarray.*)))\n"
+    "    KEEP(*(SORT(.dynarray.*)))\n"
     "    "
 )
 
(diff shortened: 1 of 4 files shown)


lost-hope:
Tested the multi-array setup using an ESP32S3 with multiple usermods (Audioreactive, SensorHub, 2 providers, and usermod effect). The sensor Hub uses an dynarray for the sensors, which brought up the issue in the previous script.
The board boots and runs without crashes. the usermods work like intended.

the diagnostic script fix [395abfa](https://github.com/wled/WLED/pull/5814/commits/395abfad0dd0497543c4f3ef3f9ae9915a8b8e9c) correctly reports the usermod number.


I had AI inspect the linker map:
> sensorSlots and usermods blocks are cleanly separated, correct 5-digit zero-padded keys (00000/00001/99999), no interleaving. The static_assert validation works correctly (fails as expected on non-padded keys; our own sensor_bus.h was updated to 00001 accordingly).

Issue #5810, opened 2026-08-18
DDP Compression Extension — Gauging Upstream Interest


aenertia:
# DDP Compression Extension — Gauging Upstream Interest

I've been running WLED on an M5StickC over a PPP serial link (USB at 1.5Mbaud,
~172 KB/s effective) and hit the obvious bandwidth wall. To make it work I
implemented a compression extension for DDP — it uses the reserved bits in the
existing header so it's backwards compatible with standard senders and receivers.

Posting here to share what I've built and see if there's any appetite for
carrying something like this upstream. Happy to submit PRs if it's useful to
others, or keep it as a fork-only thing if not.

---

## The problem

DDP is great for LED streaming. The bandwidth requirements are not:

| Transport | Effective BW | 800px RGB @ 30fps | 12,800px RGB @ 30fps |
|-----------|-------------|-------------------|----------------------|
| WiFi (20 Mbps) | 2.3 MB/s | fine | fine |
| PPP/USB serial (1.5 Mbaud) | **172 KB/s** | 71 fps — tight | **4.5 fps — unusable** |
| Serial (115200 baud) | **11.5 KB/s** | **4.7 fps** | impossible |
| LoRa (250 kbps) | **~25 KB/s** | **10 fps** | impossible |

WiFi is fine. Anything slower gets painful quickly.

---

## What I built

A delta+RLE compression extension that uses the reserved bits in the DDP header:

- **Flags byte (byte 0), bit 5**: `COMPRESSED` flag (0x20)
- **Sequence byte (byte 1), upper nibble**: compression type
  - `0x10` = delta+RLE (XOR against previous frame, then PackBits RLE)
  - `0x20` = RLE only (keyframe — no delta, for resync after packet loss)
  - `0x30` = transform (global fade/scale + sparse explicit writes — decoder only so far)

Standard DDP senders never set bit 5, so existing receivers are unaffected.
A receiver that doesn't understand compression will display noise on compressed
packets — which is the right behaviour for an opt-in extension.

### Measured compression ratios (real hardware, M5StickC, 800 LEDs, 40fps)

| Pattern | Raw | Compressed | Ratio | Wire bandwidth |
|---------|-----|-----------|-------|----------------|
| Rainbow (worst case) | 2,400 B | 2,400 B | 1:1 | 93.7 KB/s |
| Solid colour pulse | 2,400 B | ~1,900 B | 1.2:1 | 77.8 KB/s |
[...]

PR #5817, opened 2026-08-25
Per-segment realtime eligibility mask

aenertia:
# Per-segment realtime eligibility mask

Replaces `useMainSegmentOnly` with a `ddpEligibleMask` bitmask. Segments marked
eligible accept DDP/E1.31/Art-Net realtime data; the rest keep running effects.

## Motivation

Multi-segment setups (PC ARGB with fans + matrix, retail video panel + accent
lighting, DJ booth with hub75 + LED strips) need some segments on realtime while
others run local effects. The old boolean was main-segment-only or the entire strip.

Some additional comments from separate DDP Compression conversation #5810.

## Changes

- `ddpEligibleMask` (uint32): per-segment eligibility, cfg key `ddpelig`
- `ddpSlots[]`: pre-computed offset table for flat-stream-to-segment routing
- `rtFrozenSegs` (uint32): tracks which segments are frozen by realtime
- `showFrozenSegs()`: renders frozen segment pixels to bus without re-running effects
- `service()` show gate: skips `show()` when frozen segs exist, prevents race
- `realtimeLock()`/`exitRealtime()`: freeze/unfreeze lifecycle
- `rebuildDdpSlots()` called from config load, segment API changes, and strip init

Files: wled.h, udp.cpp, FX.h, FX_fcn.cpp, e131.cpp, cfg.cpp, json.cpp, wled.cpp,
set.cpp, xml.cpp + audioreactive and EleksTube_IPS usermods (replaced references).

## Backwards compatibility

- `mso=true` in cfg.json migrates to `ddpEligibleMask = (1 << mainSegId)`
- `ddpelig=0` (default) preserves legacy full-strip pixel indexing
- `MO` settings checkbox maps to single-segment mask, same UI behaviour
- cfg.json writes both `mso` (bool) and `ddpelig` (bitmask)
- `liveseg` in `/json/info` now reports the first frozen segment index (via
  `__builtin_ctz(rtFrozenSegs)`) rather than always the main segment ID. When
  `ddpelig` matches the main segment (the common case), the value is identical.
  When `ddpelig` targets a different segment, the reported value reflects the
  actual realtime target -- more useful for diagnostics but a visible change.
- No change for existing senders (OpenRGB, xLights, LedFX, HyperHDR)

## Overhead

- **RAM**: +164B static (ddpSlots[32] = 160B, masks + counters = 4B). Not heap-allocated
  because rebuildDdpSlots() runs from packet handlers.
- **Flash**: ~750B (showFrozenSegs ~70 lines, slot functions ~30 lines)
- **CPU when ddpelig=0**: zero -- all paths gated on ddpSlotCount/rtFrozenSegs

## Testing

Tested on ESP32-PICO-D4 (M5StickC), vanilla esp32dev build, WS2812B 8x32 panel:
- 3 segments: seg0 Rainbow, seg1 DDP-eligible, seg2 Ghost Rider
- During DDP to seg1: `frozensegs=2`, seg0+seg2 effects undisturbed
- Realtime enter/exit cycling: heap stable
- Legacy mode (ddpelig=0): unchanged behaviour
- Audioreactive: audio continues for non-frozen segments

Build: esp32dev SUCCESS, Flash 85.0%, RAM 26.6%.

## Future direction

The slot table and frozen-segment rendering enable a follow-up: per-packet segment
targeting via the DDP destination byte, for video wall compositors that need to route
different content to different segments within a single frame.

Assisted-by: OpenCode/OhMyOpenCode agentic harness with multiple LLM providers

## Summary by CodeRabbit

* **New Features**
  * Added support for routing DDP streams across multiple eligible LED segments.
  * Improved realtime rendering with per-segment freezing and optimized output.
  * Added DDP eligibility tracking and expanded status reporting.
* **Bug Fixes**
  * Preserved compatibility with existing “main segment only” configurations.
  * Kept local audio processing active when only selected segments are frozen.
  * Improved realtime timeout handling and restoration of segment states.

### DIFF
--- a/usermods/audioreactive/audio_reactive.cpp
+++ b/usermods/audioreactive/audio_reactive.cpp
@@ -1579,8 +1579,8 @@ class AudioReactive : public Usermod {
       if (strip.isUpdating() && (millis() - lastUMRun < 2)) return;   // be nice, but not too nice
 
       // suspend local sound processing when "real time mode" is active (E131, UDP, ADALIGHT, ARTNET, DDP, DMX)
-      //  exception: sound input is still needed when useMainSegmentOnly - other segments are still running with local input.
-      if (realtimeMode && !realtimeOverride && !useMainSegmentOnly) {
+      //  exception: sound input is still needed when only some segments are frozen - other segments are still running with local input.
+      if (realtimeMode && !realtimeOverride && !rtFrozenSegs) {
         #if defined(ARDUINO_ARCH_ESP32) && defined(WLED_DEBUG)
         if ((disableSoundProcessing == false) && (audioSyncEnabled == 0)) {  // we just switched to "disabled"
           DEBUG_PRINTLN(F("[AR userLoop]  realtime mode active - audio processing suspended."));
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1871,8 +1871,21 @@ void WS2812FX::showFrozenSegs() {
 
 void WS2812FX::setRealtimePixelColor(unsigned i, uint32_t c) {
   if (rtFrozenSegs) {
-    const Segment &seg = getMainSegment();
-    if (seg.isActive() && i < seg.length()) seg.setPixelColorRaw(i, c);
+    if (ddpSlotCount > 0) {
+      // route through slot table so non-main eligible segments get pixels
+      for (uint8_t s = 0; s < ddpSlotCount; s++) {
+        const DdpSegSlot &slot = ddpSlots[s];
+        if (i < slot.globalStart + slot.length) {
+          Segment &seg = getSegment(slot.segId);
+          if (seg.isActive()) seg.setPixelColorRaw(i - slot.globalStart, c);
+          return;
+        }
+      }
+    } else {
+      // legacy: no slot table, write to main segment
+      const Segment &seg = getMainSegment();
+      if (seg.isActive() && i < seg.length()) seg.setPixelColorRaw(i, c);
+    }
   } else {
     setPixelColor(i, c);
   }
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1821,19 +1821,28 @@ void WS2812FX::showFrozenSegs() {
   const bool is2D = (Segment::maxHeight > 1);
   const uint16_t mw = Segment::maxWidth;
 
+  // row-stride aware paint: flat segPixStart+i is wrong when seg.width() < maxWidth
+  const auto paintSeg = [&](const Segment &seg, unsigned maxLen) {
+    const unsigned w = seg.stop  - seg.start;
+    const unsigned h = seg.stopY - seg.startY;
+    unsigned i = 0;
+    for (unsigned r = 0; r < h && i < maxLen; r++) {
+      const unsigned rowStart = is2D ? (unsigned)(seg.startY + r) * mw + seg.start : seg.start;
+      for (unsigned x = 0; x < w && i < maxLen; x++, i++) {
+        uint32_t c = seg.pixels[i];
+        if (c && useGamma) c = gamma32(c);
+        BusManager::setPixelColor(getMappedPixelIndex(rowStart + x), c);
+      }
+    }
+  };
+
   if (__builtin_popcount(rtFrozenSegs) == 1) {
     // Case B: single frozen segment -- walk seg.pixels[] directly
     unsigned segIdx = __builtin_ctz(rtFrozenSegs);
     if (segIdx < _segments.size()) {
       const Segment &seg = _segments[segIdx];
       if (seg.isActive() && seg.pixels) {
-        uint16_t segPixStart = is2D ? (uint16_t)seg.startY * mw + seg.start : seg.start;
-        unsigned segLen = seg.length();
-        for (unsigned i = 0; i < segLen; i++) {
-          uint32_t c = seg.pixels[i];
-          if (c && useGamma) c = gamma32(c);
-          BusManager::setPixelColor(getMappedPixelIndex(segPixStart + i), c);
-        }
+        paintSeg(seg, seg.length());
       }
     }
   } else {
@@ -1844,13 +1853,7 @@ void WS2812FX::showFrozenSegs() {
       if (slot.segId >= _segments.size()) continue;
       const Segment &seg = _segments[slot.segId];
       if (!seg.isActive() || !seg.pixels) continue;
-      uint16_t segPixStart = is2D ? (uint16_t)seg.startY * mw + seg.start : seg.start;
-      unsigned len = min((unsigned)slot.length, (unsigned)seg.length());
-      for (unsigned i = 0; i < len; i++) {
-        uint32_t c = seg.pixels[i];
-        if (c && useGamma) c = gamma32(c);
-        BusManager::setPixelColor(getMappedPixelIndex(segPixStart + i), c);
-      }
+      paintSeg(seg, min((unsigned)slot.length, (unsigned)seg.length()));
     }
   }
 
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -1801,12 +1801,12 @@ void WS2812FX::showFrozenSegs() {
   if (!_pixels || !rtFrozenSegs) { show(); return; }
 
   uint32_t allActiveMask = 0;
-  for (unsigned i = 0; i < _segments.size(); i++)
+  for (unsigned i = 0; i < _segments.size() && i < 32; i++)
     if (_segments[i].isActive()) allActiveMask |= (1u << i);
 
   // Case D: effect segments running -- service() calls show(); calling it here too
   // would double-show per iteration, hanging TFT SPI DMA.
-  for (unsigned i = 0; i < _segments.size(); i++) {
+  for (unsigned i = 0; i < _segments.size() && i < 32; i++) {
     if (!(allActiveMask & (1u << i))) continue;
     if (rtFrozenSegs & (1u << i)) continue;
     if (_segments[i].on && !_segments[i].freeze) { show(); return; }
--- a/wled00/FX_fcn.cpp
+++ b/wled00/FX_fcn.cpp
@@ -632,10 +632,14 @@ Segment &Segment::setName(const char *newName) {
   if (newName) {
     const int newLen = min(strlen(newName), (size_t)WLED_MAX_SEGNAME_LEN);
     if (newLen) {
-      if (name) p_free(name); // free old name
-      name = static_cast<char*>(allocate_buffer(newLen+1, BFRALLOC_PREFER_PSRAM));
-      if (mode == FX_MODE_2DSCROLLTEXT) startTransition(strip.getTransition(), true); // if the name changes in scrolling text mode, we need to copy the segment for blending
-      if (name) strlcpy(name, newName, newLen+1);
+      char *newBuf = static_cast<char*>(allocate_buffer(newLen+1, BFRALLOC_PREFER_PSRAM));
+      if (newBuf) {
+        strlcpy(newBuf, newName, newLen+1);
+        if (mode == FX_MODE_2DSCROLLTEXT) startTransition(strip.getTransition(), true); // if the name changes in scrolling text mode, we need to copy the segment for blending
+        char *oldName = name;
+        name = newBuf;
+        if (oldName) p_free(oldName);
+      }
       return *this;
     }
   }
@@ -1359,7 +1363,7 @@ void WS2812FX::service() {
   #ifdef WLED_DEBUG
   if ((_targetFps != FPS_UNLIMITED) && (millis() - nowUp > _frametime)) DEBUG_PRINTF_P(PSTR("Slow effects %u/%d.\n"), (unsigned)(millis()-nowUp), (int)_frametime);
   #endif
-  if (doShow && !_suspend) {
+  if (doShow && !_suspend && !rtFrozenSegs) {
     yield();
     Segment::handleRandomPalette(); // slowly transition random palette; move it into for loop when each segment has individual random palette
     _lastServiceShow = nowUp; // update timestamp, for precise FPS control
@@ -1741,7 +1745,7 @@ void WS2812FX::show() {
     _pixelCCT = static_cast<uint8_t*>(allocate_buffer(totalLen * sizeof(uint8_t), BFRALLOC_PREFER_PSRAM)); // allocate CCT buffer if necessary, prefer PSRAM
   if (_pixelCCT) memset(_pixelCCT, 127, totalLen); // set neutral (50:50) CCT
 
-  if (realtimeMode == REALTIME_MODE_INACTIVE || useMainSegmentOnly || realtimeOverride > REALTIME_OVERRIDE_NONE) {
+  if (realtimeMode == REALTIME_MODE_INACTIVE || rtFrozenSegs || realtimeOverride > REALTIME_OVERRIDE_NONE) {
     // clear frame buffer
     memset(_pixels, 0, sizeof(uint32_t) * totalLen);
     // blend all segments into (cleared) buffer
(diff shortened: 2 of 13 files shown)


wled00/set.cpp
```diff
@@ -472,7 +472,7 @@ void handleSettingsSet(AsyncWebServerRequest *request, byte subPage)
     nodeBroadcastEnabled = request->hasArg(F("NB"));
 
     receiveDirect = request->hasArg(F("RD")); // UDP realtime
-    useMainSegmentOnly = request->hasArg(F("MO"));
+    ddpEligibleMask = request->hasArg(F("MO")) ? (1UL << strip.getMainSegmentId()) : 0; rebuildDdpSlots();
```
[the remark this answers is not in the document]
aenertia: This preserves existing UI behavior currently - discussion required as to how to expose. PR was not designed to resolve UI considerations, only implement machinery.

wled00/FX_fcn.cpp
```diff
@@ -1794,8 +1794,80 @@ void WS2812FX::show() {
...
+      if (!seg.isActive() || !seg.pixels) continue;
+      uint16_t segPixStart = is2D ? (uint16_t)seg.startY * mw + seg.start : seg.start;
+      unsigned len = min((unsigned)slot.length, (unsigned)seg.length());
+      for (unsigned i = 0; i < len; i++) {
+        uint32_t c = seg.pixels[i];
+        if (c && useGamma) c = gamma32(c);
+        BusManager::setPixelColor(getMappedPixelIndex(segPixStart + i), c);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
aenertia: Yup - more UI consideration that I missed and valid omission

softhack007:
> [aenertia](https://github.com/aenertia) [force-pushed](https://github.com/wled/WLED/compare/37079db7900cf526d2c85c114b07150496dee1a6..e9507b99cdf2e22c8dd1bcce0e3e3ffb189e5232) the upstream-pr/segment-eligibility branch from [37079db](https://github.com/wled/WLED/commit/37079db7900cf526d2c85c114b07150496dee1a6) to [e9507b9](https://github.com/wled/WLED/commit/e9507b99cdf2e22c8dd1bcce0e3e3ffb189e5232) 
[1 hour ago](https://github.com/wled/WLED/pull/5817#event-29973608763)

@aenertia we already talked about force-push. "No force-push" means "never ever". I'll restart the review bot on this PR, be ready to adress already asked questions.

Edit: "Review failed - The head commit changed during the review" - this is one thing we wanted to avoid with the "don't force-push" rule.

wled00/set.cpp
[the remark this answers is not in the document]
softhack007: > PR was not designed to resolve UI considerations, only implement machinery.

@aenertia @coderabbitai not so fast ;-) we can't merge a PR that breaks the settings UI. It might be ok if "advanced stuff" is ony available by json API commands, but backwards compatibility with existing settings and with options showing in the WebUI is needed.

softhack007:
@coderabbitai full review

softhack007:
@coderabbitai full review

wled00/FX_fcn.cpp
```diff
@@ -1794,10 +1794,98 @@ void WS2812FX::show() {
...
+    // Case C: multiple frozen segments -- sparse paint via ddpSlots[]
+    for (uint8_t s = 0; s < ddpSlotCount; s++) {
+      const DdpSegSlot &slot = ddpSlots[s];
+      if (!(rtFrozenSegs & (1u << slot.segId))) continue;
+      if (slot.segId >= _segments.size()) continue;
+      const Segment &seg = _segments[slot.segId];
+      if (!seg.isActive() || !seg.pixels) continue;
+      paintSeg(seg, min((unsigned)slot.length, (unsigned)seg.length()));
+    }
+  }
```
softhack007, line 1858: @coderabbitai would it be possible to integrate `WS2812FX::showFrozenSegs()` into the normal `WS2812FX::show()` ? 

Apart from the claimed speedup, how is this function different? Is it still supporting segment layering/blending and fade transitions?

wled00/FX_fcn.cpp
```diff
@@ -1794,10 +1794,98 @@ void WS2812FX::show() {
...
+
+  // _pixels[] zeroed -- fast path writes bus buffers directly, not _pixels[].
+  if (_pixels) memset(_pixels, 0, sizeof(uint32_t) * getLengthTotal());
```
softhack007, line 1861: i don't think this is any faster than the original `memset(_pixels, 0, sizeof(uint32_t) * totalLen);` - or maybe I misunderstood the comment in line 1860 ?

wled00/FX_fcn.cpp
softhack007: Is this a breaking change, compared to the previous behaviour with "use main segment only" ?

wled00/udp.cpp
```diff
@@ -404,21 +404,51 @@ static void parseNotifyPacket(const uint8_t *udpIn) {
...
+    freezeSegForRealtime(ddpSlots[s].segId);
+}
+
 // realtimeLock() is called from UDP notifications, JSON API or serial Ada
 void realtimeLock(uint32_t timeoutMs, byte md)
 {
   if (!realtimeMode && !realtimeOverride) {
-    if (useMainSegmentOnly) {
-      Segment& mainseg = strip.getMainSegment();
-      mainseg.clear(); // clear entire segment (in case sender transmits less pixels)
```
softhack007, line 413: > ` // clear entire segment (in case sender transmits less pixels)`

looks like this step got lost? @coderabbitai what happens if the sender transmits fewer pixels than what's available in a segment? Likewise, what if too many pixels arrive for a segment?

wled00/udp.cpp
```diff
@@ -592,7 +625,7 @@ void handleNotifications()
       }
       if (tpmPacketCount == numPackets) { //reset packet count and show if all packets were received
         tpmPacketCount = 0;
-        if (useMainSegmentOnly) strip.trigger();
+        if (rtFrozenSegs) strip.showFrozenSegs();
```
softhack007, line 628: i'm not sure this is a safe solution. The original code uses strip.trigger() to serialize calls to strip.show() when useMainSegmentOnly is set. Without serialization, how can we be sure that several tasks will not call the show() function in parallel? And how to prevent excessive framerates?

wled00/udp.cpp
```diff
@@ -637,7 +670,7 @@ void handleNotifications()
           setRealtimePixel(id, udpIn[i], udpIn[i+1], udpIn[i+2], udpIn[i+3]);
         }
       }
-      if (useMainSegmentOnly) strip.trigger();
+      if (rtFrozenSegs) strip.showFrozenSegs();
```
softhack007, line 673: see previous comment.

softhack007:
@coderabbitai will other LED streaming protocols like WARLS, E1.31 DMX, and Art-Net still work as before?

wled00/wled.h
```diff
@@ -725,7 +725,23 @@ WLED_GLOBAL IPAddress realtimeIP _INIT_N(((0, 0, 0, 0)));
 WLED_GLOBAL unsigned long realtimeTimeout _INIT(0);
 WLED_GLOBAL uint8_t tpmPacketCount _INIT(0);
 WLED_GLOBAL uint16_t tpmPayloadFrameSize _INIT(0);
-WLED_GLOBAL bool useMainSegmentOnly _INIT(false);
+// DDP per-segment targeting (replaces useMainSegmentOnly)
+WLED_GLOBAL uint32_t ddpEligibleMask _INIT(0);                           // bitmask of segments accepting concatenated DDP
+WLED_GLOBAL uint32_t rtFrozenSegs    _INIT(0);                           // which segments are currently frozen by realtime
```
softhack007, line 730: what happens when WLED is build with  MAX_NUM_SEGMENTS > 32 ?

For example on boards with PSRAM

https://github.com/wled/WLED/blob/c456457717f55f78f8efb7fc7ba3540cd8c7b9f4/wled00/FX.h#L94

wled00/wled.cpp
```diff
@@ -149,13 +149,27 @@ void WLED::loop()
...
+        const Segment &_seg = strip.getSegment(_i);
+        if (_seg.isActive() && _seg.on) activeMask |= (1u << _i);
+      }
+      allSegsFrozenByDDP = activeMask && ((rtFrozenSegs & activeMask) == activeMask);
+    }
+    if (!allSegsFrozenByDDP && (!offMode || strip.isOffRefreshRequired() || strip.needsUpdate()))
       strip.service();
     #ifdef ESP8266
-    else if (!noWifiSleep)
+    else if (!noWifiSleep && !allSegsFrozenByDDP)
```
softhack007, line 166: why did you add `&& !allSegsFrozenByDDP` in the 8266 wifi sleep code?

wled00/wled.cpp
```diff
@@ -149,13 +149,27 @@ void WLED::loop()
     handlePresets();
     yield();
 
-    if (!offMode || strip.isOffRefreshRequired() || strip.needsUpdate())
```
softhack007, line 152: @coderabbitai please check if the previous condition ` if (!offMode || strip.isOffRefreshRequired() || strip.needsUpdate())` is faithfully represented in this PRs changes.

wled00/wled.cpp
```diff
@@ -149,13 +149,27 @@ void WLED::loop()
...
+    if (!allSegsFrozenByDDP && (!offMode || strip.isOffRefreshRequired() || strip.needsUpdate()))
       strip.service();
     #ifdef ESP8266
-    else if (!noWifiSleep)
+    else if (!noWifiSleep && !allSegsFrozenByDDP)
       delay(1); //required to make sure ESP enters modem sleep (see #1184)
     #endif
   }
+  // Safety net: exit realtime if timeout expired (backup for handleNotifications check)
+  if (realtimeMode && millis() > realtimeTimeout) exitRealtime();
```
softhack007, line 171: @coderabbitai the previous implementation had some "magic timeout" values that were interpreted as "infinite timeout" (I'm not saying its a good design, but its the legacy behaviour). Is this behaviour preserved?

https://github.com/wled/WLED/blob/c456457717f55f78f8efb7fc7ba3540cd8c7b9f4/wled00/udp.cpp#L430

https://github.com/wled/WLED/blob/c456457717f55f78f8efb7fc7ba3540cd8c7b9f4/wled00/json.cpp#L451-L459

usermods/audioreactive/audio_reactive.cpp
```diff
@@ -1579,8 +1579,8 @@ class AudioReactive : public Usermod {
       if (strip.isUpdating() && (millis() - lastUMRun < 2)) return;   // be nice, but not too nice
 
       // suspend local sound processing when "real time mode" is active (E131, UDP, ADALIGHT, ARTNET, DDP, DMX)
-      //  exception: sound input is still needed when useMainSegmentOnly - other segments are still running with local input.
-      if (realtimeMode && !realtimeOverride && !useMainSegmentOnly) {
+      //  exception: sound input is still needed when only some segments are frozen - other segments are still running with local input.
+      if (realtimeMode && !realtimeOverride && !rtFrozenSegs) {
```
netmindz, line 1583: This is a big no no. Totally breaking breaking responsibility boundaries, changing behaviour of unrelated code for questionable benefits, total spaghetti through bad use of globals

wled00/udp.cpp
```diff
@@ -404,21 +404,51 @@ static void parseNotifyPacket(const uint8_t *udpIn) {
   stateUpdated(CALL_MODE_NOTIFICATION);
 }
 
+void rebuildDdpSlots() {
```
netmindz, line 407: Nope! UDP should not be doing anything with segment behaviour

softhack007:
> wled00/FX_fcn.cpp (1)
> > `1366-1371`: _🎯 Functional Correctness_ | _🟠 Major_ | _🏗️ Heavy lift_
> > **Restore one normal show for mixed frozen and effect segments.**
> > When a frozen segment and an active non-frozen segment coexist, `WLED::loop()` runs `strip.service()`. Line 1366 then suppresses `show()` because `rtFrozenSegs` is nonzero. The local effect updates its buffer but does not render until another realtime packet calls `showFrozenSegs()`.
> > Make the mixed-mode path perform one normal `show()` from `service()`. Then make the Case D path avoid a second show for the same frame.
> > Also applies to: 1807-1813

@aenertia please also check this finding. It was a bit hidden in the rabbit's "outside of diff range" comments, but the concern should be addressed.

netmindz:
@aenertia please do not invest any further time in this pr. We will not be able to accept this work as it currently stands, nor any further incremental changes

wled00/wled.cpp
softhack007: in other words
* ignores strip.trigger() requests
* ignores strip.needsUpdate() status
* breaks behaviour of "Off Refresh" in LEDs settings, possibly causing TM1814 to flash into "demo mode"

softhack007: just from my curiosity - what happens if some segments are not DDP-frozen, however frozen by the "image" effect, frozen by the user (segment UI), or frozen because they are under JSON individual pixel control (json.cpp line 314) ?

aenertia:
> > [aenertia](https://github.com/aenertia) [force-pushed](https://github.com/wled/WLED/compare/37079db7900cf526d2c85c114b07150496dee1a6..e9507b99cdf2e22c8dd1bcce0e3e3ffb189e5232) the upstream-pr/segment-eligibility branch from [37079db](https://github.com/wled/WLED/commit/37079db7900cf526d2c85c114b07150496dee1a6) to [e9507b9](https://github.com/wled/WLED/commit/e9507b99cdf2e22c8dd1bcce0e3e3ffb189e5232)
> > [1 hour ago](https://github.com/wled/WLED/pull/5817#event-29973608763)
> 
> @aenertia we already talked about force-push. "No force-push" means "never ever". I'll restart the review bot on this PR, be ready to adress already asked questions.
> 
> Edit: "Review failed - The head commit changed during the review" - this is one thing we wanted to avoid with the "don't force-push" rule.

ACK. Sorry I work with another codebase who has the exact opposite rule (ONLY collapsed in openPR) and the original submit had cruft in it from local repo that had no business being here so thought I was being helpful.

wled00/wled.h
aenertia: I think ArduinoJson v6 on ESP32 rejects `uint64_t` without `ARDUINOJSON_USE_LONG_LONG=1` -- I believe I hit this during the rework prototype. So a 64-segment bitmask would need either that build flag project-wide, or splitting to two uint32s in the JSON serialization. For now I'm using `uint32_t` which I believe covers the non-PSRAM case (MAX_NUM_SEGMENTS=32) worth flagging.

wled00/wled.cpp
aenertia: The service gate was the most damaging change -- trigger, needsUpdate, isOffRefreshRequired all dead. I don't have TM1814 so that was invisible to me. Will be gone in the rework -- `seg.freeze` already handles effect suppression without gating service.

I'll be on the lookout for other chipsets that need the off-refresh path besides TM1814

aenertia:
@softhack007 @netmindz -- thanks for the feedback, this was what I needed to highlight my blind spots around the implementation and make this into something more robust. I appreciate the time.

The code was originally designed around a DDP/UDP optimized path, and that meant I was blind to several of the pathways highlighted here -- TM1814 off-refresh, WARLS routing, the `strip.trigger()` serialization contract, the service gate breaking `needsUpdate()`. The existing `service()` / `show()` / `blendSegment()` pipeline already handles mixed frozen+effect segments correctly when `useMainSegmentOnly` is true. I should have traced that path before building a parallel one.

I am reading the code, but WLED has gotten substantially larger since I was last hacking at it 7-8 years ago and the rendering pipeline goes deeper than I'd traced. On the hardware side I've got two or three test bench boards set up with a mix of WS2812B, SK6812, some PWM channels -- but no TM1814, no WARLS sender. How are y'all keeping track of all the board/strip/protocol combinations at this point to stay sane? Any tips appreciated -- I'd like to at least be checking the right code paths even when I can't test the hardware.

I've got enough from this review to work on a rework. I'll close this and test a more generic version on the hardware I do have before resubmitting.

If there's a better review/cycle procedure for something like this ; or a different workflow - llet me know. I don't want to be putting work on you more than needs to be, and do put considerable effort into manual pipeline hygiene before opening these.

Issue #4448, opened 2025-01-02
IPv6 Support is not planned yet


apalrd:
**Is your feature request related to a problem? Please describe.**
WLED should support IPv6 completely

**Describe the solution you'd like**
- Receive IPv6 RAs and autoconfigure addresses accordingly (LLA + GUA if available)
- Support IPv6 sockets in any place an IPv4 socket is used
- Broadcast IPv6 AAAA records over mDNS (#3861 requests this)

**Describe alternatives you've considered**
In general almost all IPv6 features required are already provided by arduino_esp32 v3, however, they are not all present in the older versions currently used by WLED. It looks like version updates have been attempted in #4045 and therefore this can't be implemented until the version update is completed. 

Minor updates may be required to WLED logic which deals directly with IPAddress items.

netmindz:
As ipv6 usage in domestic settings is very low and ipv6 only rather than dual stack almost unheard of and without a killer feature of ipv6, I don't see this being added anytime soon.

Thunderbolt1003usa:
IPv6 usage is now over 50% and it's not higher because projects like WLED haven't implemented it. So what would be an disadvantage of adding IPv6 support?

softhack007:
> IPv6 usage is now over 50% and it's not higher because projects like WLED haven't implemented it. So what would be an disadvantage of adding IPv6 support?

Can you name some "killer feature" in ipv6 that wled would need, especially in a domestic setting behind a firewall, with less than 32767 clients in the local network? 

Dual stack is development and maintenance effort, and it increases flash size (bad for 4MB device owners). Besides satisfying the principle and "arriving in the new century", what would be the added value for wled?

netmindz:
I don't know where you are getting that 50% from @Thunderbolt1003usa 

I would expect usage for domestic networks to be substantially lower. You can't measure rollout based on the traffic percentage to high volume websites as this will be impacted by mobile phone networks moving from CGN to IPV6

apalrd:
The 50% number probably comes from Google who estimates that 50% of users support ipv6. APNIC is in the same ballpark, their methodology is buying mobile ads which are only available over ipv6 and seeing how many load successfully. Those are both counts of users, not bandwidth. 

[...]

PR #5822, opened 2026-08-28
Add read-only support for IPv6 addresses using the built-in LwIP features

RedVyper:
First, narrowly-scoped step toward wled/WLED#4448 (IPv6 support).

The ESP-IDF v5 build targets already compile LWIP with IPv6 support, it just needed someone to actually enable it with `WiFi.enableIPv6()`/`ETH.enableIPv6()`.

Deliberately out of scope: no socket/mDNS/dual-stack changes, no static IPv6 config, no changes to the ESP-IDF v4 (Tasmota-fork, `esp32dev_V4`/`esp32_eth_V4` envs) build targets. Those still blackhole router advertisements and are untouched here.

Tested on ESP32-C3

wled-ipv6-ll

## Summary by CodeRabbit

* **New Features**
  * Added IPv6 support for compatible ESP32 devices using Wi-Fi or Ethernet.
  * IPv6 link-local and global addresses are now available through device information responses.
  * Added IPv6 address display to the Wi-Fi settings page, preferring the global address when available.


softhack007:
Maintainers: this is just a proposal to get started with IPv6, see discussions in https://github.com/wled/WLED/issues/4448.

T-X:
The code reads a bit as if only a single IPv6 address of LLA/GUA/ULA (each?) was expected/supported. Am I misreading this? Or is this somehow a limitation of the ESP-IDF?

I've come across IPv6 networks (e.g. Freifunk) where multiple gateways were present (multihoming). There you'd potentially get multiple IPv6 GUAs, one from each gateway router. Potentially with some tuning of the RA's preference field to make a host prefer a specific gateway.

Also with IPv6 privacy extensions ([RFC4941](https://www.rfc-editor.org/info/rfc4941/)) you'll get/generate multiple addresses (but likely the privacy extension is less useful for WLED devices).

And many thanks for looking into this, @RedVyper!

RedVyper:
@T-X you're right, the expectation from IPv6 is that a host can and will have multiple IP addresses and potentially (actually even probably) multiple ULAs and GUAs. This work is just the beginning and I will be happy to add support for multiple ULAs and GUAs, even if read-only for now. My goal was to break the initial inertia on the matter and start working. With time I will address all of your concerns (not a promise but certainly a hope).

The privacy extensions, unfortunately, seem not to be included in LwIP (yet?).

PR #5837, opened 2026-09-07
warn when effect registration exceeds the list limit

### LINKED ISSUE #5827: addEffect() fails silently when the effect list is full
Summary
When the effect vector is full, WS2812FX::addEffect() returns 255 and no caller checks the return value. The effect is never registered, and there is no log line, no warning, and nothing in the UI to say so. From the user's side the effect simply is not in the build.

This is separate from the effect-count ceiling itself (PR [#5108](https://github.com/Aceftwspades/WLED/pull/5108)) — it is about the ceiling being hit invisibly. Whatever happens with the limit long-term, hitting it should be diagnosable.

Details:
`uint8_t WS2812FX::addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name) {
  ...
  } else if (_mode.size() < 255) { // 255 is reserved for indicating the effect wasn't added
    _mode.push_back(mode_fn);
    ...
  } else {
    return 255; // The vector is full so return 255
  }
}`

Every usermod I can find calls it as strip.addEffect(255, &mode_x, _data_x); and discards the result, so that branch is invisible at runtime.

With MODE_COUNT at 220, the ceiling is reachable in practice: 255 slots minus 220 built-ins leaves ~35 for all usermods on the device combined — audioreactive, any effect-providing usermod, and anything the user wrote.

Why it is worse than a plain "no space left"
Which effects go missing is not stable. Usermods register from static initialisers across translation units, and the relative order of those is unspecified per-file, so which registrations land in the surviving slots varies with toolchain and link order. The user-visible symptom is "some of my effects are missing, and which ones changes between builds", with nothing anywhere pointing at the cause.

Repro
Build with usermods registering enough effects to exceed 255 total.
Boot. Some effects are absent from the effect list.
No log output, no error, no indication of how many were dropped.
Suggested fix
Small and self-contained:

Log a warning in addEffect() when it returns 255 because the vector is full, including the effect name that was dropped.
Optionally surface remaining free slots (or a dropped-effect count) in the Info page, so it can be seen without a serial console.

Environment
main @ 700c0a42 (17.0.0-devV5); also verified identical on 16.0.1
ESP32, esp32dev, PlatformIO
Hit while developing a usermod that registers ~36+ effects (since trimmed to 33 but would like to add more without adjusting base wled)

keeltrace:
AI-generated contribution. I am the human operator responsible for follow-up with maintainers. Please feel free to close this PR or request any changes.

Fixes #5827.

When `WS2812FX::addEffect()` reaches the 254-effect capacity, it currently returns `255` silently. Most usermods discard that return value, so dropped effects are difficult to diagnose.

This adds a debug warning only on the full-list failure path and includes the effect metadata string that could not be registered. It intentionally does not change the 8-bit effect ID limit or overlap with the separate larger-effect-ID work.

Validation:
- `git diff --check`
- reviewed against current `main`

I could not run a firmware build locally because PlatformIO is not installed in this environment.

## Summary by CodeRabbit

- **Diagnostics**
  - Added diagnostic messages when the effect list reaches its maximum capacity, including a warning and the affected effect name. This makes failed effect registration easier to identify during troubleshooting.

- **Documentation**
  - Clarified that effect names supplied when registering effects must be stored in program memory, helping developers provide effect metadata in the required format.

### DIFF
--- a/wled00/FX.cpp
+++ b/wled00/FX.cpp
@@ -10965,6 +10965,8 @@ uint8_t WS2812FX::addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name)
     if (_modeCount < _mode.size()) _modeCount++;
     return _mode.size() - 1;
   } else {
+    DEBUG_PRINT(F("WARNING: effect list full, could not add effect: "));
+    DEBUG_PRINTLN(FPSTR(mode_name));
     return 255; // The vector is full so return 255
   }
 }
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -929,7 +929,7 @@ class WS2812FX {
     uint8_t getFirstSelectedSegId() const;
     uint8_t getLastActiveSegmentId() const;
     uint8_t getActiveSegsLightCapabilities(bool selectedOnly = false) const;
-    uint8_t addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name);         // mode_name must point to PROGMEM data; defined in FX.cpp;
+    uint8_t addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name);         // add effect to the list - mode_name must point to PROGMEM; defined in FX.cpp;
 
     inline uint8_t getBrightness() const    { return _brightness; }       // returns current strip brightness
     inline static constexpr unsigned getMaxSegments() { return MAX_NUM_SEGMENTS; }  // returns maximum number of supported segments (fixed value)
--- a/wled00/FX.h
+++ b/wled00/FX.h
@@ -929,7 +929,7 @@ class WS2812FX {
     uint8_t getFirstSelectedSegId() const;
     uint8_t getLastActiveSegmentId() const;
     uint8_t getActiveSegsLightCapabilities(bool selectedOnly = false) const;
-    uint8_t addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name);         // add effect to the list; defined in FX.cpp;
+    uint8_t addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name);         // mode_name must point to PROGMEM data; defined in FX.cpp;
 
     inline uint8_t getBrightness() const    { return _brightness; }       // returns current strip brightness
     inline static constexpr unsigned getMaxSegments() { return MAX_NUM_SEGMENTS; }  // returns maximum number of supported segments (fixed value)


wled00/FX.cpp
```diff
@@ -10965,6 +10965,8 @@ uint8_t WS2812FX::addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name)
     if (_modeCount < _mode.size()) _modeCount++;
     return _mode.size() - 1;
   } else {
+    DEBUG_PRINT(F("WARNING: effect list full, could not add effect: "));
+    DEBUG_PRINTLN(FPSTR(mode_name));
```
softhack007, line 10969: @coderabbitai please check if `FPSTR()`is the correct macro here to assure that `mode_name` stays in PROGMEM on 8266.

softhack007: what is the difference between `PSTR()` and `FPSTR()` on 8266?

softhack007: > The `addEffect()` API accepts `const char*`. A third-party usermod could pass an SRAM string. That would not be safe with `FPSTR()`.

1. is the added line `DEBUG_PRINTLN(FPSTR(mode_name));` kind of nailing the function to flash string pointers, or was this an implicit need already before this PR?
2. is there an alternative way to DEBUG_PRINT `mode_name` without creating trouble for SRAM strings?
3. If not, would it be possible to add two interfaces to this function? maybe making the parameter a `String` would be sufficient? Or we need to have a second instance of the function that would handle any `const char *`? 
4. Please find all call sites of `WS2812FX::addEffect()` in the WLED core and inside all usermods. Is there a need to have a generic  `const char *` interface, or can we restrict mode_name to `const __FlashStringHelper *` without breaking existing call sites?

keeltrace:
> @keeltrace yes that was also my conclusion, especially after looking at https://github.com/wled/WLED/pull/5837#discussion_r3969038273. My question was if we are introducing a new (implied) constraint on `mode_name`. I think that's not the case. Just wondering if a small comment should be added to the declaration of `WS2812FX::addEffect()` in fx.h.

Sorry that was an AI comment you responded too, my be l bad I will make sure that gets a label in the future. 

I would defer to your judgment or preference. Pretty please tell my clanker the changes you'd like. I'll have it do another pass

Thanks - keel trace handler

wled00/FX.h
```diff
@@ -929,7 +929,7 @@ class WS2812FX {
     uint8_t getFirstSelectedSegId() const;
     uint8_t getLastActiveSegmentId() const;
     uint8_t getActiveSegsLightCapabilities(bool selectedOnly = false) const;
-    uint8_t addEffect(uint8_t id, mode_ptr mode_fn, const char *mode_name);         // add effect to the list; defined in FX.cpp;
```
softhack007, line 932: minor: please preserve the existing comment, for example `//add effect to the list - mode_name must point to PROGMEM; defined in FX.cpp;`

softhack007 (approved):
looks good to me now - ready to merge.

PR #5862, opened 2026-09-19
Fix PWM for new chips on V5

DedeHai:
update to PWM bus so it works properly in V5.
It does basically the same thing as before just using the new functions and structs. 
for some reason espressif decided to not name the registers the same on all new ESP types.

Also improved dead-time handling for 2-pin CCT strips to prevent shoot-through (increased dead time as it was kind of useless before).

Tested on ESP32 classic, C6 and C5

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved LED output compatibility across supported ESP32 hardware and Arduino-ESP32 versions.
  - Fixed PWM timer initialization for more reliable startup behavior.
  - Improved two-channel CCT lighting output with better phase alignment and consistent brightness control.
  - Added controlled output dead time to reduce switching overlap and improve signal stability.
  - Improved recovery when PWM channel setup fails, helping prevent partially initialized outputs.


wled00/bus_manager.cpp
```diff
@@ -558,59 +561,51 @@ void BusPwm::show() {
...
-      if (_bri >= 254 && duty >= maxBri / 2 && duty < maxBri) {
-        duty -= deadTime << 1; // shorten duty of larger signal except if full on
+      // add dead time between signals to prevent shoot-through +2 ensures proper spacing in dithering as the signal jitters by 1 (8bit) tick
+      // a tick when dithering can be higher than 500ns (1/4.8kHz/256 = 800ns), the +2 adds less than 50ns when not dithering
+      deadTime = (2 << bitShift) + (uint64_t(DEADTIME_NS) * _frequency * maxBri) / 1000000000ULL; // _frequency * maxBri is 1/pwmtick,
+      int dutyOvershoot = (pwmBri + (deadTime * 2)) - maxBri;
+      // if total pwmBri plus two dead-time zones exceed maxBri, shorten the duty proportionally except if one is at 0%
+      if (duty < maxBri && dutyOvershoot > 0 && _data[0] > 0 && _data[1] > 0) {
+        const unsigned totalDuty = ((_data[0] + _data[1]) * pwmBri) / 255; //duty[0] + duty[1];
+        if (totalDuty > 0) duty -= (dutyOvershoot * duty) / totalDuty; // subtract overshoot proportionally to signals contribution
```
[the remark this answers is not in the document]
DedeHai: why not just skip the whole dead time block if cctICused?

wled00/bus_manager.cpp
```diff
@@ -558,59 +569,51 @@ void BusPwm::show() {
...
+    constexpr unsigned DEADTIME_NS = 500; // 500ns should be safe for most drivers that do not have internal shoot-through protection, with dithering this is a min value
+    if (_type == TYPE_ANALOG_2CH && Bus::_cctBlend <= 0 && !cctICused) {
+      // add dead time between signals to prevent shoot-through +2 ensures proper spacing in dithering as the signal jitters by 1 (8bit) tick
+      // a tick when dithering can be higher than 500ns (1/4.8kHz/256 = 800ns), the +2 adds less than 50ns when not dithering
+      deadTime = (2 << bitShift) + (uint64_t(DEADTIME_NS) * _frequency * maxBri) / 1000000000ULL; // _frequency * maxBri is 1/pwmtick,
+      int dutyOvershoot = (pwmBri + (deadTime * 2)) - maxBri;
+      // if total pwmBri plus two dead-time zones exceed maxBri, shorten the duty proportionally except if one is at 0%
+      if (duty < maxBri && dutyOvershoot > 0 && _data[0] > 0 && _data[1] > 0) {
+        const unsigned totalDuty = ((_data[0] + _data[1]) * pwmBri) / 255; //duty[0] + duty[1];
+        if (totalDuty > 0) duty -= (dutyOvershoot * duty) / totalDuty; // subtract overshoot proportionally to signals contribution
```
willmmiles, line 591: Do we need to clamp the correction to make sure `duty` never goes negative?

DedeHai: There is actually a logic flaw in the math: 
`int dutyOvershoot = (pwmBri + (deadTime * 2)) - maxBri;` only works correctly for white channel = 100% (which is what I tested with). if not at full white, the code is overly cautious and the duty negative clamp is needed. the overshoot should be calculated using `totalDuty` instead. If that is done, resulting duty can never become negative. Still won't hurt to add a guard to prevent future bugs if this code is adjusted.

DedeHai:
@coderabbitai I discovered an issue, probably unrelated to this PR:
setting cctBlend to a positive value for example 50, full brightness. Using the CCT slider generates a discontinuity at the switch over point. I.e. going from full warm towards cold white: ww should be 255 up to the switch point, then start to decrease. However this is not what I see at the pwm output pins. Can you elaborate on what the current code is expected to output? If you can find this behaviour in the code: when was this issue introduced?

DedeHai:
@willmmiles I did a full rewrite on how the phase-shifts are handled. The issue I asked the rabbit about above is actually caused by hardware: on the C5 there is no support for "wrap around" like on the older chips. i.e. if duty+hpoint exceed the max pwm value (maxBri) the pulse is not carried over to the next period but output set to zero which is why I saw signals being cut. I fixed that by making sure all pulses always fit one period. There was another quirk when using dithering: the hardware extends the dutycycle by one pulse, so if duty+hpoint were close to maxBri (closer than one full pulse i.e. 16 dither bits) it would overflow causing the hardware pin to latch high until updated in the next period(s). 
I thoroughly tested this and did not find any additional edge cases. Tested on ESP32 classic, C3, C5, C6 and ESP8266 although not all possible combinations of outputs but CCT and RGBW.

